PR #3138, opened 2026-07-09
feat(sc): add checkpoint save/restore to SingleController

haitian-nvidia:
## What does this PR do?

Adds checkpoint save/restore to the single-controller (SC) RL loop, so a run can
resume from the latest checkpoint instead of starting over. It reuses the legacy
GRPO checkpoint machinery (`CheckpointManager`, `GRPOSaveState`,
`load_dataloader_state`) and keeps the same on-disk layout, extending it to the
state SC owns:

- **Model weights + optimizer + training state**: setup builds a
  `CheckpointManager`, loads the latest `GRPOSaveState` (or defaults),
  propagates `pretrained_checkpoint`, and passes the weights/optimizer resume
  paths into `_build_trainer`. The actor restores
  `_trainer_version`/`_train_steps`/`_max_rollout_version` and
  `consumed_samples`/`total_valid_tokens`, and after each train step saves on
  `save_period`, the last step, or a `checkpoint_must_save_by` timeout (top-k
  `metric_name` supported), stopping early on timeout.
- **Dataloader position + epoch**: `train_dataloader.pt` is written on save and
  restored via `load_dataloader_state` on setup; `current_epoch` is persisted
  and restored.
- **Replay buffer** (`TQReplayBuffer`): `state_dict` fetches each ready group's
  rows from the DataPlane into an envelope; `load_state_dict` validates and
  truncates to the current capacity before any DataPlane write, then re-puts
  each group. `run()` restores it before the pumps start.

## Notable design points

- **The replay buffer is saved only when `over_sampling=True`.** Under the
  strict quota (`over_sampling=False`) each weight version gets exactly one
  dispatched batch and the pump never re-dispatches. A checkpoint can only hold
  *committed* groups, so a partially-committed window could never be completed
  on resume — the train pump would wait forever for groups that will never
  arrive (an immediate stall under `force_in_order`). Instead, resume restores
  `_max_rollout_version = current_step - 1` and regenerates the whole window
  from the restored dataloader position. With `over_sampling=True` there is no
  quota: restored groups are surplus, missing ones are regenerated by the
  free-running pump, and stale ones are evicted — so persisting the buffer is
  safe and saves wasted generation.
- **Restore never blocks.** `run()` acquires one `_buffer_capacity` permit per
  restored group before the pumps start; if the restored count exceeded
  capacity, that acquisition would hang `run()` forever (no pump is running yet
  to release a permit). `load_state_dict`'s truncation to the current
  `max_buffered_rollouts` guarantees `K <= capacity`, asserted before acquiring
  — this closes a resume deadlock when resuming onto a smaller-capacity config.
- **The `state_dict` snapshot is taken synchronously on the event loop** before
  any `await`, so it cannot be torn by the rollout pump committing into the same
  buffer concurrently.
- **In-flight rollouts are not persisted** (reserved-but-uncommitted slots),
  matching legacy — they are regenerated on resume.
- **Backward compatible**: a checkpoint written before dataloader / replay-buffer
  support (missing `train_dataloader.pt` / `replay_buffer.pt`) resumes with a
  fresh dataloader position / empty buffer and a warning, not a crash.

## Testing

- **Unit**: `test_sc_checkpointing.py` + `test_tq_replay_buffer.py` — counter
  restore, save triggers, dataloader position round-trip, and replay-buffer
  round-trip / preflight validation / permit accounting (incl. the
  `K == capacity` no-hang case).
- **Functional**: `tests/functional/grpo_checkpoint_single_controller.sh` — a
  two-run save→restart on 2 GPUs (Qwen3-0.6B, Megatron), verifying
  weights/dataloader/replay-buffer restore and resume at the correct step
  (step 2 → resume → step 4).

### DIFF
--- a/nemo_rl/algorithms/async_utils/replay_buffer.py
+++ b/nemo_rl/algorithms/async_utils/replay_buffer.py
@@ -701,6 +701,194 @@ class TQReplayBuffer:
 
         return len(drop_idxs)
 
+    async def state_dict(self, *, saved_capacity: int) -> dict[str, Any]:
+        """Serialize ready groups (meta + DataPlane payloads) for checkpointing.
+
+        Snapshots the ready slots synchronously on the event loop first, then
+        fetches each group's rows from the DataPlane. Unready reservations are
+        in-flight rollouts and are dropped, matching legacy semantics. The
+        snapshot stays consistent during the async fetch: concurrent commits
+        only append/flip *other* slots, and the train pump — the only
+        remover — is the caller itself; groups committed mid-save land in the
+        next checkpoint.
+
+        Args:
+            saved_capacity: max_buffered_rollouts at save time, recorded so
+                load_state_dict can report capacity changes across restarts.
+
+        Returns:
+            Envelope: ``{"partition_id": ..., "saved_capacity": ...,
+            "groups": [{"meta", "start_weight", "end_weight", "target_step",
+            "group_id", "fields_data"}, ...]}``.
+        """
+        snapshot: list[tuple[KVBatchMeta, int, int, Optional[int], str]] = []
+        for i, ready in enumerate(self.ready_list):
+            if not ready:
+                continue
+            meta = self.meta_list[i]
+            assert meta is not None  # commit sets meta before ready=True
+            snapshot.append(
+                (
+                    meta,
+                    self.start_weight_list[i],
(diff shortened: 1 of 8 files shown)


yuki-97 (approved):
thanks for supporting this, LGTM and merged into SC branch.

PR #3144, opened 2026-07-09
fix: router-replay startup validation and NeMo-Gym trace alignment

zyzhou5:
# What does this PR do ?

**Startup validation** (`router_replay.py`, `grpo.py`) — new `validate_router_replay_startup(...)` in `grpo.setup()` rejects, up front, configs that previously no-op'd or crashed mid-run: non-Megatron backend (routes captured but never replayed), MTP, NeMo-Gym over sync-TQ, and dense (non-MoE) models.

**NeMo-Gym trace alignment** (`nemo_gym.py`) — exact route-row count check (`<`→`!=`); splice each turn's real final-token route from the next turn's re-forward (vLLM leaves it a placeholder); sentinel-fill a single trace-less rollout instead of killing the run (only a fully trace-less batch raises); read the sentinel shape via `len()` instead of a redundant tensor conversion.
# Validation

We didn't observe any regression.

Screenshot 2026-07-09 at 4 08 36 PM

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -104,6 +104,7 @@ from nemo_rl.models.generation.vllm import VllmConfig, VllmGeneration
 from nemo_rl.models.megatron.router_replay import (
     configure_vllm_for_router_replay,
     router_replay_enabled,
+    validate_router_replay_startup,
 )
 from nemo_rl.models.policy import PolicyConfig
 from nemo_rl.models.policy.interfaces import ColocatablePolicyInterface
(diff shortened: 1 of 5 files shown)


zyzhou5:
/ok to test [93e6e16](https://github.com/NVIDIA-NeMo/RL/commit/93e6e16f79994797592dddd0d4feb7912f7b2aa1)

zyzhou5:
/ok to test [acdd038](https://github.com/NVIDIA-NeMo/RL/commit/acdd03812cc5e316149c91b16c592886dc51abe5)

PR #3148, opened 2026-07-09
fix: skip RayVirtualCluster.shutdown when Ray already torn down

ZhiyuLi-Nvidia:
## Issue

MOPD nightly (`llm_mopd_qwen3_1_7b_3n8g_megatron_pack`) exits 1 after training completes cleanly:

```
[C] core_worker_process.cc:88: Check failed: !core_worker_process
```

Ray's atexit tears down `CoreWorker` first, then Python's GC runs `__del__` on `RayVirtualCluster` instances, which call `remove_placement_group` on the dead runtime — CoreWorker re-init fails the CHECK and the process aborts.

## Fix

Return early from `RayVirtualCluster.shutdown()` when `ray.is_initialized()` is False. The guard only fires from `__del__` during `_Py_Finalize`; placement groups are already gone with Ray at that point.

**5 lines, 1 file.** No new imports, no semantic change when Ray is alive, no reference retention.

Verified twice end-to-end on the `mopd-qwen3-1.7b-3n8g-megatron-pack` nightly recipe:

### DIFF
--- a/nemo_rl/distributed/virtual_cluster.py
+++ b/nemo_rl/distributed/virtual_cluster.py
@@ -1018,6 +1018,11 @@ class RayVirtualCluster:
 
         This method is idempotent and can be safely called multiple times.
         """
+        # Skip if Ray is already gone (typically from __del__ during _Py_Finalize
+        # after Ray's atexit teardown). Any Ray API call would trigger a fatal
+        # core_worker_process.cc:88 CHECK. Placement groups die with Ray.
+        if not ray.is_initialized():
+            return True
         if self._node_placement_groups is not None:
             # Remove all placement groups
             for pg in self._node_placement_groups:


nemo_rl/distributed/virtual_cluster.py
```diff
@@ -1042,3 +1043,31 @@ def __del__(self) -> None:
         user calls shutdown().
         """
         self.shutdown()
+
+
+@contextmanager
+def ray_resources(
```
yuki-97: Can we scope this down to teachers only?

`common.env` sets `set -eou pipefail`, so every nightly's `uv run ... | tee $RUN_LOG` already fails on non-zero exit — MOPD isn't the first to check exit code end-to-end. And prior 2-cluster GRPO nightlies have been passing for months with `RayVirtualCluster.__del__ -> remove_placement_group` running after Ray's atexit, so that path is de facto silent.

What's actually new in MOPD is `TeacherWorkerGroup.shutdown` (`teacher_worker_group.py:294`) -> `RayWorkerGroup.shutdown(cleanup_method="shutdown")` -> `ray.get(...)` + `ray.kill(worker)` at `worker_groups.py:1075/1091`. Those need a live CoreWorker; `remove_placement_group` empirically doesn't trip the CHECK.

wdyt trying a narrower fix — just teacher shutdown inside `async_grpo_train`'s existing finally, dropping the `ray_resources` helper and both runner wraps:

```python
# nemo_rl/algorithms/grpo.py — async_grpo_train, existing finally
finally:
    ...
    for _name, _twg in (teacher_worker_groups or {}).items():
        try:
            _twg.shutdown()
        except Exception as e:
            print(f"Error shutting down teacher worker group '{_name}': {e}")
```

If MOPD still passes with that, `virtual_cluster.py` and both runners stay untouched, and the PR body's "wired into `async_grpo_train` — teachers cleaned up at the algorithm level" ends up matching what the code does. If MOPD still crashes, then the cluster wrapping is load-bearing and this PR shape is right.

yuki-97 (commented):
thanks for root cause and helping fix!

I have some thoughts on whether we can scope down the fix, since other recipes run with `run_grpo_nemo_gym.py` doesn't meet this error. could you help to take a look?

nemo_rl/distributed/virtual_cluster.py
ZhiyuLi-Nvidia: Tried both variants of the teacher-only scope-down. Both crashed with the same core_worker_process.cc:88 signature.

Why the two teacher-only fixes don't work:

- @yuki-97 exact snippet, teacher loop in async_grpo_train.finally. Algorithm layer. Same crash.
- Fix inside `TeacherWorkerGroup` (algorithm-level): defensive if not ray.is_initialized(): return guard in TeacherWorkerGroup.shutdown(). Class layer. Same crash.

"remove_placement_group is de facto silent" seems not hold for MOPD. Tried tear down both the teacher worker_group and cluster while still got the same failure.

Proposed new fix: return early from RayVirtualCluster.shutdown() when ray.is_initialized() is False. The guard only fires from __del__ during _Py_Finalize; placement groups are already gone with Ray at that point. 5 lines, 1 file. No new imports, no semantic change to shutdown() when Ray is alive, no reference retention.

ZhiyuLi-Nvidia:
> thanks for root cause and helping fix!
> 
> I have some thoughts on whether we can scope down the fix, since other recipes run with `run_grpo_nemo_gym.py` doesn't meet this error. could you help to take a look?

That's consistent with what we observed. The failure fires only when a Ray-touching del runs after Ray's atexit tears down CoreWorker, and whether that condition is reached seems to depend on cluster count and topology. Colocated GRPO has 1 RayVirtualCluster and stays silent. Non-colocated 2-cluster async GRPO stays silent for the topologies exercised so far. MOPD with teacher_cluster runs with 3 clusters and tips Ray past whatever it silently tolerates, so the CHECK fires.

ZhiyuLi-Nvidia:
/ok to test 5dacef40df6257992cd898fe26d192c3ad2567a9..c247a8fd7f4c9f0376fd2b7b3bc4e95c69ccc4de

yuki-97 (approved):

PR #3152, opened 2026-07-10
fix(async): Always honor recompute_kv_cache_after_weight_updates

jepio:
# What does this PR do ?

KV cache invalidation is not specific to in-flight updates. If KV cache (eg. same prompt) is reused across steps with different weights we can keep a stale KV cache around for a long time which can manifest as compounding policy KL divergence and logprob errors.

Make recompute_kv_cache_after_weight_updates evaluation independent of in_flight_weight_updates and honor it in async grpo.

# Issues

# Usage

Use this config:
```yaml
grpo:
   async_grpo:
     enabled: true
     max_trajectory_age_steps: 1
     in_flight_weight_updates: false
     recompute_kv_cache_after_weight_updates: true

```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/async_utils/trajectory_collector.py
+++ b/nemo_rl/algorithms/async_utils/trajectory_collector.py
@@ -554,16 +554,20 @@ class AsyncTrajectoryCollector:
         async_cfg = self.master_config.grpo.get("async_grpo", {})
         if async_cfg.get("recompute_kv_cache_after_weight_updates", False):
             try:
-                print("🔄 Invalidating vLLM prefix/KV caches after weight update")
+                print(
+                    "🔄 Invalidating generation backend KV caches after weight update"
+                )
                 invalidated = self.policy_generation.invalidate_kv_cache()
                 if invalidated:
-                    print("✅ Invalidated vLLM prefix/KV caches after weight update")
+                    print(
+                        "✅ Invalidated generation backend KV caches after weight update"
+                    )
                 else:
                     print(
-                        "⚠️ vLLM cache invalidation reported partial/unsuccessful on some workers"
+                        "⚠️ KV cache invalidation not supported or only partially applied by the generation backend"
                     )
             except Exception as e:
-                print(f"⚠️ Failed to invalidate vLLM caches: {e}")
+                print(f"⚠️ Failed to invalidate generation backend KV caches: {e}")
 
         self._refit_pause_cleared.set()
 
--- a/nemo_rl/algorithms/async_utils/trajectory_collector.py
+++ b/nemo_rl/algorithms/async_utils/trajectory_collector.py
@@ -547,13 +547,12 @@ class AsyncTrajectoryCollector:
         """Resume new generation starts after refit is complete."""
         print("🔄 Resuming generation starts after refit")
 
-        # Invalidate&recompute vLLM caches after the in-flight weight updates if
+        # Invalidate&recompute vLLM caches after the weight updates (in-flight or not) if
         # recompute_kv_cache_after_weight_updates is True (AREAL-style implementation).
         # Otherwise, keep using the stale KV caches (Magistral-style implementation).
+        # Not invalidating KV cache can result in compounding policy KL errors across steps.
         async_cfg = self.master_config.grpo.get("async_grpo", {})
-        if async_cfg.get("in_flight_weight_updates", False) and async_cfg.get(
-            "recompute_kv_cache_after_weight_updates", False
-        ):
+        if async_cfg.get("recompute_kv_cache_after_weight_updates", False):
             try:
                 print("🔄 Invalidating vLLM prefix/KV caches after weight update")
                 invalidated = self.policy_generation.invalidate_kv_cache()
--- a/tests/unit/algorithms/test_async_utils.py
+++ b/tests/unit/algorithms/test_async_utils.py
@@ -1341,6 +1341,30 @@ class TestAsyncTrajectoryCollector:
         ray.kill(buffer)
         ray.kill(mock_env)
 
+    def test_resume_after_refit_invalidates_cache_without_in_flight_updates(self):
+        """Test resume after refit invalidates cache without in-flight updates."""
+        collector = self.create_local_collector()
+        async_cfg = collector.master_config.grpo["async_grpo"]
+        async_cfg["in_flight_weight_updates"] = False
+        async_cfg["recompute_kv_cache_after_weight_updates"] = True
+        collector.policy_generation.invalidate_kv_cache = mock.Mock(return_value=True)
+
+        collector.resume_after_refit()
+
+        collector.policy_generation.invalidate_kv_cache.assert_called_once_with()
+
+    def test_resume_after_refit_skips_cache_invalidation_when_recompute_disabled(self):
+        """Test resume after refit skips cache invalidation when recompute is disabled."""
+        collector = self.create_local_collector()
+        async_cfg = collector.master_config.grpo["async_grpo"]
+        async_cfg["in_flight_weight_updates"] = True
+        async_cfg["recompute_kv_cache_after_weight_updates"] = False
+        collector.policy_generation.invalidate_kv_cache = mock.Mock(return_value=True)
+
+        collector.resume_after_refit()
+
+        collector.policy_generation.invalidate_kv_cache.assert_not_called()
+
     def test_calculate_target_weights(self):
         """Test target weight calculation logic."""
         buffer = ReplayBuffer.remote(max_size=10)
--- a/docs/guides/async-grpo.md
+++ b/docs/guides/async-grpo.md
@@ -42,7 +42,7 @@ grpo:
     enabled: true
     max_trajectory_age_steps: 1  # Maximum age, in training steps, for trajectories
     in_flight_weight_updates: false  # Enable for faster weight synchronization
-    recompute_kv_cache_after_weight_updates: false # Invalidates kv cache after in-flight-weight-updates
+    recompute_kv_cache_after_weight_updates: false # Invalidates kv cache after weight-updates
 ```
 
 ### Complete Example Config
@@ -68,7 +68,7 @@ grpo:
     enabled: true
     max_trajectory_age_steps: 1
     in_flight_weight_updates: false  # Enable for faster weight synchronization
-    recompute_kv_cache_after_weight_updates: false # Invalidates kv cache after in-flight-weight-updates
+    recompute_kv_cache_after_weight_updates: false # Invalidates kv cache after weight-updates
 
 cluster:
   num_nodes: 2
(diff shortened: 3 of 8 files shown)


nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -521,13 +521,12 @@ def resume_after_refit(self) -> None:
...
-        # Invalidate&recompute vLLM caches after the in-flight weight updates if
+        # Invalidate&recompute vLLM caches after the weight updates (in-flight or not) if
         # recompute_kv_cache_after_weight_updates is True (AREAL-style implementation).
         # Otherwise, keep using the stale KV caches (Magistral-style implementation).
+        # Not invalidating KV cache can result in compounding policy KL errors across steps.
         async_cfg = self.master_config.grpo.get("async_grpo", {})
-        if async_cfg.get("in_flight_weight_updates", False) and async_cfg.get(
-            "recompute_kv_cache_after_weight_updates", False
-        ):
+        if async_cfg.get("recompute_kv_cache_after_weight_updates", False):
```
terrykong, line 555: [`nemo_rl/algorithms/async_utils/trajectory_collector.py:529`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/nemo_rl/algorithms/async_utils/trajectory_collector.py#L529)

Now that this condition no longer requires `in_flight_weight_updates`, a few docs/comments still describe the flag as in-flight-only — one of them with a functional flag-name typo. It would be great to update them in this PR:

1. [`docs/guides/async-grpo.md:192-193`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/docs/guides/async-grpo.md#L192-L193) — "While using in-flight weight update, user can choose whether to recompute KV caches after weight udpate by configuring `recompute_kv_cache_after_weight_update` configuration." Besides the now-stale in-flight premise and the "udpate" typo, the flag name is missing the trailing `s` — a user copy-pasting `recompute_kv_cache_after_weight_update` sets a nonexistent key that [`async_cfg.get(..., False)`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/nemo_rl/algorithms/async_utils/trajectory_collector.py#L529) silently ignores, so the feature stays off with no warning. This one is the highest-value fix.
2. [`nemo_rl/algorithms/grpo.py:152`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/nemo_rl/algorithms/grpo.py#L152) — `AsyncGRPOConfig` field comment: "Recomputes the KV cache after the in-flight weight updates."
3. Inline YAML comments "after in-flight-weight-updates" (optional): [`docs/guides/async-grpo.md:45`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/docs/guides/async-grpo.md#L45), [`:71`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/docs/guides/async-grpo.md#L71), [`examples/configs/grpo_math_1B.yaml:53`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/examples/configs/grpo_math_1B.yaml#L53), [`examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml:51`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L51), [`tests/unit/reference_configs/grpo_math_1B.yaml:53`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/tests/unit/reference_configs/grpo_math_1B.yaml#L53), [`research/template_project/configs/grpo_math_1B.yaml:47`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/research/template_project/configs/grpo_math_1B.yaml#L47)

**Action:** reword to "after weight updates (in-flight or not)" (or similar) at sites 1-2 and fix the flag name/typo in `docs/guides/async-grpo.md:193`; the YAML comment updates are nice-to-have.

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -521,13 +521,12 @@ def resume_after_refit(self) -> None:
...
             try:
                 print("🔄 Invalidating vLLM prefix/KV caches after weight update")
                 invalidated = self.policy_generation.invalidate_kv_cache()
```
terrykong, line 560: [`nemo_rl/algorithms/async_utils/trajectory_collector.py:535-538`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/nemo_rl/algorithms/async_utils/trajectory_collector.py#L535-L538)

With the relaxed condition, the `else` warning a few lines below becomes newly reachable for a documented, supported config: megatron backend + `recompute_kv_cache_after_weight_updates=true` without in-flight updates. Megatron generation doesn't override `invalidate_kv_cache()`, so the [base implementation returns `False`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/nemo_rl/models/generation/interfaces.py#L274), and every refit would print "⚠️ vLLM cache invalidation reported partial/unsuccessful on some workers" — misleading twice over: the backend isn't vLLM, and `False` here means "unsupported/no-op" rather than a partial failure (megatron recompute is already handled natively via the [`kv_cache_management_mode="recompute"` override](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/nemo_rl/algorithms/grpo.py#L917-L928)).

**Action:** reword the message in `resume_after_refit()` to cover the unsupported-backend case, e.g. `"⚠️ KV cache invalidation not supported (or only partially applied) by the generation backend"`. Non-blocking.

tests/unit/algorithms/test_async_utils.py
```diff
@@ -1236,6 +1236,19 @@ def test_async_trajectory_collector_prepare_for_refit(self):
...
+    ) -> None:
+        collector = self.create_local_collector()
+        async_cfg = collector.master_config.grpo["async_grpo"]
+        async_cfg["in_flight_weight_updates"] = False
+        async_cfg["recompute_kv_cache_after_weight_updates"] = True
+        collector.policy_generation.invalidate_kv_cache = mock.Mock(return_value=True)
+
+        collector.resume_after_refit()
+
+        collector.policy_generation.invalidate_kv_cache.assert_called_once_with()
```
terrykong, line 1354: [`tests/unit/algorithms/test_async_utils.py:1250`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/tests/unit/algorithms/test_async_utils.py#L1250)

Nice regression test for the fixed positive path. The skip path is still unasserted, though — nothing checks that `invalidate_kv_cache` is NOT called when `recompute_kv_cache_after_weight_updates=false`. The test below deliberately sets `in_flight_weight_updates=True` to pin that invalidation is gated solely on `recompute` (verified passing locally on this PR head):

```suggestion
        collector.policy_generation.invalidate_kv_cache.assert_called_once_with()

    def test_resume_after_refit_skips_cache_invalidation_when_recompute_disabled(
        self,
    ) -> None:
        """KV cache is not invalidated when recompute_kv_cache_after_weight_updates is off."""
        collector = self.create_local_collector()
        async_cfg = collector.master_config.grpo["async_grpo"]
        async_cfg["in_flight_weight_updates"] = True
        async_cfg["recompute_kv_cache_after_weight_updates"] = False
        collector.policy_generation.invalidate_kv_cache = mock.Mock(return_value=True)

        collector.resume_after_refit()

        collector.policy_generation.invalidate_kv_cache.assert_not_called()
```

terrykong (commented):
Thanks for the fix — the team-review (4 agents) confirms the core change is correct and needed: `resume_after_refit()` runs on every refit regardless of `in_flight_weight_updates` ([`grpo.py:4187`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/nemo_rl/algorithms/grpo.py#L4187)), so previously `recompute_kv_cache_after_weight_updates=true` was silently ignored without in-flight updates. The change also makes the vLLM path consistent with the megatron path, which already honors the flag independently of in-flight ([`grpo.py:918-928`](https://github.com/NVIDIA-NeMo/RL/blob/695d4eb0dd6a8884fe04aac44958c757d043f8f1/nemo_rl/algorithms/grpo.py#L918-L928)). Invalidation ordering is safe (before `_refit_pause_cleared.set()`), and the new unit test is a genuine regression test (fails pre-fix, passes post-fix — verified locally).

3 low-severity, non-blocking suggestions inline: stale doc wording (incl. a functional flag-name typo in the async-grpo guide), a now-reachable misleading warning message, and a negative-path test.

Linters: all pre-commit hooks pass.

_Generated by Claude Code_

terrykong:
thanks @jepio for the fix. do you have an runs demonstrating the benefit (gen kl or token mult prob diff)?

FYI @yuki-97 @mehraakash

jepio:
> thanks @jepio for the fix. do you have an runs demonstrating the benefit (gen kl or token mult prob diff)?
> 
> FYI @yuki-97 @mehraakash

I do, though i don't have wandb, so here are some plots. There are other backend probability mismatches for the model i'm using but KV cache is a big contributor:
image

tests/unit/algorithms/test_async_utils.py
jepio: added this test

nemo_rl/algorithms/async_utils/trajectory_collector.py
jepio: i've reworded all the messages in this function to talk about "generation backend KV cache"

nemo_rl/algorithms/async_utils/trajectory_collector.py
jepio: Changed all comments i could find so that they talk about "weight updates" and no only "in-flight weight updates".

yuki-97 (dismissed):
thanks @jepio , the fix and the plot LGTM.

yuki-97:
hi @jepio , there seems a lint fail, could you help to fix by running `pre-commit run --all-files`?

jepio:
> hi @jepio , there seems a lint fail, could you help to fix by running `pre-commit run --all-files`?

done

yuki-97 (approved):

PR #3116, opened 2026-07-08
fix: un-blocklist deepseek_v3 in transformers tokenizer class set

ZhiyuLi-Nvidia:
## Summary
https://github.com/NVIDIA-NeMo/RL/issues/2764
Two coordinated fixes so Moonlight-16B-A3B (and any future `deepseek_v3` model) can load its tokenizer offline.

### Bug 1 — transformers blocks the tokenizer

transformers 5.4–5.11 lists `"deepseek_v3"` in two internal registries that together force the fast tokenizer backend and suppress `trust_remote_code`:

| Registry | Effect |
|---|---|
| `MODELS_WITH_INCORRECT_HUB_TOKENIZER_CLASS` (set) | Suppresses the model's `tokenizer_class` / `auto_map` |
| `TOKENIZER_MAPPING_NAMES` (dict) | Pins `deepseek_v3` to `TokenizersBackend` (fast) |

Under `HF_HUB_OFFLINE=1` the fast backend can't enumerate the repo to find `tiktoken.model` and fails. Moonlight ships no `tokenizer.json`, only a slow `TikTokenTokenizer` reachable via `auto_map` — so the online fallback also can't help.

Fix: `nemo_rl/__init__.py` removes both entries at import time. `discard` / `pop-with-default` are no-ops if the entries are absent, so the patch is safe on any transformers version.

### Bug 2 — dead code in Megatron tokenizer setup

`nemo_rl/models/megatron/setup.py:setup_model_and_optimizer` used to mutate `hf_tokenizer_kwargs["trust_remote_code"] = True` on the Megatron `TokenizerConfig`. This was dead code:

| Field | Snapshot at `__post_init__` (empty dict) | Original code's intent | Actual runtime | This PR |
|---|---|---|---|---|
| `trust_remote_code` | `dict.get("trust_remote_code", False)` = **False** | True (dead) | **False** — dead | **True** (attribute set directly) |
| `tokenizer_hf_no_use_fast` | `not dict.get("use_fast", True)` = **False** (i.e. `use_fast=True`) | True (dead) | **`use_fast=True`** via default | Unchanged, `use_fast=True` |

Root cause of the deadness: `TokenizerConfig.__post_init__` (see `Megatron-Bridge/src/megatron/bridge/training/tokenizers/config.py`) snapshots `hf_tokenizer_kwargs` into plain attributes *once* at construction time; from then on `build_tokenizer` reads the attribute, not the dict.

[...]

PR #3154, opened 2026-07-10
fix: fix dapo-deepseek-v3-64n8g.v2 and dsv3 tokenizer

yuki-97:
## 1. Fix `dapo-deepseek-v3-64n8g.v2`

Update `num_prompts_per_step * num_generations_per_prompt` from 64 * 8 to 32 * 16.

The dynamic-sampling fix in #2478 filters more samples than before. With `num_generations_per_prompt=8`, samples are easily filtered and most runs approach the max allowed number of batches (10), triggering the 4h timeout. Some runs also fail with:

```
ValueError: Dynamic sampling has reached the maximum allowed number of batches (10). Consider evaluating the complexity of your data or adjusting the num_prompts_per_step or num_generations_per_prompt parameters to enhance the diversity of the samples.
```

Raising `num_generations_per_prompt` to 16 gives enough within-group diversity.

## 2. Fix DSV3 tokenizer

The Moonlight tokenizer fix in #3116 breaks DSV3. Narrow the scope of the Moonlight patch so it only kicks in when the default `AutoTokenizer.from_pretrained` path fails.

## Tests
Validated below tests passed.
- `dapo-deepseek-v3-64n8g.v2`
- `grpo-moonlight-16b-automodel-1n8g-ep8`

## Note
`dapo-deepseek-v3-64n8g.v2` is fixed in `r0.7.0` branch, but will easily meet the below error in main branch. tracked in https://github.com/NVIDIA-NeMo/RL/issues/3155.

`(MegatronPolicyWorker pid=1391870) [2026-07-10 06:19:36] pool0-00008:1391870:1393239 [7] init.cc:904 NCCL WARN Duplicate GPU detected : rank 15 and rank 7 both on CUDA device e4000`

### DIFF
--- a/nemo_rl/models/policy/__init__.py
+++ b/nemo_rl/models/policy/__init__.py
@@ -55,13 +55,34 @@ def _patch_transformers_tokenizer_class_set():
         "Check if the upstream fix now applies and remove this patch if so."
     )
 
+    from transformers import AutoTokenizer
     from transformers.models.auto.tokenization_auto import (
         MODELS_WITH_INCORRECT_HUB_TOKENIZER_CLASS,
         TOKENIZER_MAPPING_NAMES,
     )
 
-    MODELS_WITH_INCORRECT_HUB_TOKENIZER_CLASS.discard("deepseek_v3")
-    TOKENIZER_MAPPING_NAMES.pop("deepseek_v3", None)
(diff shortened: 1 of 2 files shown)


ZhiyuLi-Nvidia (approved):
Thank you LGTM.

PR #3156, opened 2026-07-10
fix(automodel): avoid fp32 weight load when optimizer holds master

### LINKED ISSUE #2865: [Automodel]: Hard-coded fp32 model weights doubles memory with FusedAdam
## Summary

The automodel backend hard-codes the model load dtype to **fp32** in order to support master weights [here](https://github.com/NVIDIA-NeMo/RL/blob/1ff6e117a41e37fdd80401ae375ec9f23e8d28ea/nemo_rl/models/automodel/setup.py#L305). 
However, the optimizer is configurable and supports both `torch.optim.AdamW` and TE `FusedAdam`. When **FusedAdam** is used, the fp32 load is unnecessary and causes **~2× optimizer master-weight memory** (and ~2× model-weight memory), because FusedAdam already maintains its own fp32 master internally.

## Details

With the TE FusedAdam optimizer, the fp32 master is kept **inside the optimizer**, via:

```yaml
optimizer:
  name: transformer_engine.pytorch.optimizers.fused_adam.FusedAdam
  kwargs:
    master_weights: true
    store_param_remainders: true
```

`store_param_remainders: true` lets FusedAdam split the fp32 master into **bf16 (high 16 bits) + int16 remainder (low 16 bits)** — `16 + 16 = 32 = fp32`. This depends on the parameters being **bf16**: once a parameter is fp32, TE detects `param.dtype != bf16` and **skips the remainder path, storing a separate full fp32 master** instead. Ref: [https://github.com/NVIDIA/TransformerEngine/blob/42b840051647eef89761a16dfdff87e82bb253ab/transformer_engine/pytorch/optimizers/fused_adam.py#L97](https://github.com/NVIDIA/TransformerEngine/blob/42b840051647eef89761a16dfdff87e82bb253ab/transformer_engine/pytorch/optimizers/fused_adam.py#L97)

So when the load dtype is hard-coded to fp32, with FusedAdam:

- **Model weights**: bf16 → fp32 = **2× memory**.
- **Optimizer master**: `bf16 + int16 remainder` → **full fp32** = **2× memory**.

## Reproduction

Run a grpo training, toggling only the model load dtype (`AutoConfig(torch_dtype=...)` in `nemo_rl/models/automodel/setup.py`) between
`float32` and `bfloat16`, and sum the `optimizer.state` tensor bytes after the first `optimizer.step()`.

## Observed (rank 0)

| metric                    | fp32 load (current)              | modified to bf16 load |
| ------------------------- | -------------------------------- | -------------------- |
| `master_param`            | **fp32 7.32 GB** + int16 0.60 GB | int16 — 4.26 GB      |
| optimizer state /rank     | **16.43 GB**                     | 12.77 GB             |
| optimizer state all-ranks | 131.5 GB                         | 102.2 GB             |
| model params all-ranks    | 66.6 GB                          | 35.7 GB              |

`exp_avg` / `exp_avg_sq` are unchanged (bf16); the entire increase is the master copy.

## Suggested fix

Choose the model load dtype based on the optimizer (AdamW vs FusedAdam) instead of hard-coding fp32: load fp32 only for optimizers without an internal master, and keep the compute dtype (bf16) when FusedAdam holds the master.

For reference, I have a candidate fix in: [https://github.com/NVIDIA-NeMo/RL/commit/e968d90f72925637e8834f81510b5f7a181b4093](https://github.com/NVIDIA-NeMo/RL/commit/e968d90f72925637e8834f81510b5f7a181b4093)

lonexreb:
## What
The automodel (DTensor) backend hard-coded the model load dtype to `float32` in `nemo_rl/models/automodel/setup.py` so the optimizer could keep fp32 master weights. When the optimizer maintains its **own** fp32 master — TE `FusedAdam` with `master_weights=True` (bf16 params + int16 remainder) — that fp32 load is redundant and roughly **doubles** both model-weight and optimizer master-weight memory.

Fixes #2865.

## Change
Pick the load dtype from the optimizer config in `validate_and_prepare_config`:
- `master_weights=True` (optimizer holds the master) → load in the **compute dtype**.
- plain AdamW / no `master_weights` / no optimizer configured → keep the **fp32** load (unchanged behavior).

Only the load dtype changes; the FSDP mixed-precision policy and everything downstream are untouched.

## Impact
No behavior change for the exemplar configs (they use `torch.optim.AdamW`, which has no internal master → still fp32). The savings apply only to the FusedAdam+`master_weights` path the issue reports (~131GB → ~102GB optimizer state, ~67GB → ~36GB model params in the reporter's run).

## Test plan
- [x] New unit tests in `tests/unit/models/automodel/test_automodel_setup.py` asserting the `torch_dtype` passed to `AutoConfig.from_pretrained`:
  - plain AdamW → `float32`
  - FusedAdam + `master_weights=True` → compute dtype (bf16)
  - optimizer absent → `float32`
- [ ] CI (`/ok to test`)

_Note: local test run blocked by an unrelated `uv`/Python `mac_ver()` issue on the dev box; logic validated in isolation and syntax-checked. CI will exercise the full suite._

### DIFF
--- a/nemo_rl/models/automodel/setup.py
+++ b/nemo_rl/models/automodel/setup.py
@@ -324,10 +324,20 @@ def validate_and_prepare_config(
         else ("sdpa" if cp_size_cfg > 1 else None)
     )
 
+    # Load weights in float32 so the optimizer can keep fp32 master weights.
+    # Exception: TE FusedAdam with master_weights=True keeps its own fp32 master
+    # internally (bf16 params + int16 remainder), so an fp32 load is redundant and
+    # roughly doubles both model- and master-weight memory. In that case load in the
+    # compute dtype instead. See https://github.com/NVIDIA-NeMo/RL/issues/2865.
+    optimizer_cfg = config.get("optimizer") or {}
+    optimizer_kwargs = optimizer_cfg.get("kwargs") or {}
+    optimizer_keeps_fp32_master = bool(optimizer_kwargs.get("master_weights", False))
+    model_load_dtype = dtype if optimizer_keeps_fp32_master else torch.float32
+
     # Load model config
     model_config = AutoConfig.from_pretrained(
         model_name,
-        torch_dtype=torch.float32,  # Always load in float32 for master weights
+        torch_dtype=model_load_dtype,
         trust_remote_code=True,
         attn_implementation="flash_attention_2" if enable_seq_packing else None,
         **hf_config_overrides,
--- a/tests/unit/models/automodel/test_automodel_setup.py
+++ b/tests/unit/models/automodel/test_automodel_setup.py
@@ -126,6 +126,73 @@ class TestValidateAndPrepareConfig:
     @patch("nemo_rl.models.automodel.setup.AutoConfig")
     @patch("nemo_rl.models.automodel.setup.resolve_model_class")
     @patch("nemo_rl.models.automodel.setup.configure_dynamo_cache")
+    def test_load_dtype_fp32_for_plain_adamw(
+        self,
+        mock_dynamo,
+        mock_resolve_class,
+        mock_autoconfig_class,
+        mock_config,
+        mock_autoconfig,
+    ):
+        """Optimizers without an internal master (e.g. AdamW) load weights in fp32."""
+        mock_autoconfig_class.from_pretrained.return_value = mock_autoconfig
+        mock_resolve_class.return_value = Mock
+
+        validate_and_prepare_config(config=mock_config, processor=None, rank=0)
+
+        call_kwargs = mock_autoconfig_class.from_pretrained.call_args.kwargs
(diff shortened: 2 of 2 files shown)


terrykong (commented):
@hemildesai could you review?

lonexreb:
Gentle ping @hemildesai @terrykong — this has been green and mergeable since mid-July (all checks passing, no conflicts). Happy to address any feedback or rebase if you'd prefer it freshened before review.

lonexreb:
@hemildesai @yuki-97 — gentle ping when you have bandwidth. This fixes #2865: the automodel path loads weights in fp32 even when the precision-aware optimizer already holds fp32 master weights, doubling load-time memory for no benefit. All pre-CI checks are green and the branch is mergeable against current main; it still needs a `CI:*` label to run the full tier. Happy to address any feedback or rebase again if it goes stale.

PR #3157, opened 2026-07-10
feat: fault-tolerance checkpoint retention (ft_save_period / ft_keep_latest_k)

ananthsub:
# What does this PR do ?

Adds two optional checkpointing knobs for crash recovery/resiliency, independent of the existing metric-based `keep_top_k` retention:

- **`ft_save_period`**: save a checkpoint every N steps for fault tolerance (separate from `save_period`).
- **`ft_keep_latest_k`**: retain the K most recent checkpoints (recency only, metric-agnostic).

This enables saving after each step for recovery purposes and subsequently clearing them without affecting checkpoints taken by `save_period` that are intended to be retained for downstream usage.

## Retention policy (`CheckpointManager.remove_old_checkpoints`)

Retention is now a union. A checkpoint survives if any rule keeps it:

- **Periodic**: `save_period`-aligned checkpoints. When `keep_top_k` is set, only the top-k best of these are kept (by metric if `metric_name` is set, else by recency); otherwise all periodic checkpoints are kept.
- **`ft_keep_latest_k`**: the K most recent checkpoints (pure recency), regardless of `save_period` alignment.
- The latest checkpoint (when `exclude_latest=True`).

`init_tmp_checkpoint` also clears a stale `tmp_step_N` left behind by an interrupted save so the next save starts clean.

## Driver wiring

`ft_save_period` is wired into the `should_save_by_step` gate of **every** training driver (grpo, grpo_sync, sft, dpo, rm, ppo, distillation, xtoken). `ft_keep_latest_k` retention is centralized in `CheckpointManager`, so it applies to all drivers automatically.

## Tests

- Extended the checkpoint-manager unit tests with the `ft_keep_latest_k` / periodic / metric-top-k union behaviors.
- Added per-driver tests asserting `ft_save_period` triggers saves at the expected cadence for sft, dpo, rm, distillation, and xtoken.

**Pre checks**:

* [x] Make sure you read and followed Contributor guidelines
* [x] Did you write any new necessary tests?
* [ ] Did you add or update any necessary documentation?
* [x] Did you run the unit tests and functional tests locally?

### DIFF
--- a/nemo_rl/algorithms/dpo.py
+++ b/nemo_rl/algorithms/dpo.py
@@ -570,6 +570,8 @@ def dpo_train(
 
     policy.prepare_for_training()
 
+    ft_save_period = master_config.checkpointing.get("ft_save_period")
+
     while (
         current_epoch < max_num_epochs and total_steps < master_config.dpo.max_num_steps
     ):
@@ -643,7 +645,6 @@ def dpo_train(
                 ]
                 timeout.mark_iteration()
 
-                ft_save_period = master_config.checkpointing.get("ft_save_period")
                 should_save_by_step = (
                     is_last_step
                     or (total_steps + 1) % master_config.checkpointing["save_period"]
@@ -792,9 +793,11 @@ def dpo_train(
             total_steps += 1
 
             if should_save_by_timeout:
+                checkpointer.shutdown()
                 print("Timeout has been reached, stopping training early", flush=True)
                 return
             if total_steps >= master_config.dpo.max_num_steps:
+                checkpointer.shutdown()
                 print(
                     "Max number of steps has been reached, stopping training early",
                     flush=True,
@@ -803,3 +806,10 @@ def dpo_train(
 
         current_epoch += 1
         current_step = 0  # Reset step counter for new epoch
+
+    # Flush the last checkpoint's background finalization on an epoch-bounded
+    # exit. Reaching max_num_epochs falls through the while loop and bypasses
+    # the inline shutdown() calls at the max_num_steps / timeout early returns,
+    # so without this the daemon finalization thread could be killed before the
+    # final tmp_step_N is renamed.
+    checkpointer.shutdown()
--- a/nemo_rl/algorithms/dpo.py
+++ b/nemo_rl/algorithms/dpo.py
@@ -643,10 +643,15 @@ def dpo_train(
                 ]
                 timeout.mark_iteration()
 
+                ft_save_period = master_config.checkpointing.get("ft_save_period")
                 should_save_by_step = (
                     is_last_step
                     or (total_steps + 1) % master_config.checkpointing["save_period"]
                     == 0
+                    or (
+                        ft_save_period is not None
+                        and (total_steps + 1) % ft_save_period == 0
+                    )
                 )
                 # +1 because step is 0-indexed
                 # Check if timeout-based checkpointing is enabled in config.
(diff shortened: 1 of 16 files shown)


nemo_rl/algorithms/sft.py
```diff
@@ -520,10 +520,15 @@ def sft_train(
                     "train_global_batch_size"
                 ]
                 timeout.mark_iteration()
+                ft_save_period = master_config.checkpointing.get("ft_save_period")
```
ashors1: [`sft.py:523`](https://github.com/NVIDIA-NeMo/RL/blob/d0e0317436bddbcaf5fdd4756bbda35d1f19baaf/nemo_rl/algorithms/sft.py#L523) — same pattern in `dpo.py`, `rm.py`, `ppo.py`, `grpo_sync.py`, `distillation.py`, `xtoken_off_policy_distillation.py`

Side note while reviewing: the other drivers call `begin_finalization()` but never call `checkpointer.shutdown()` at training exit, so the last checkpoint's daemon rename thread can be killed before completing on async-save backends. `grpo.py` handles this correctly — see [grpo.py:3300](https://github.com/NVIDIA-NeMo/RL/blob/d0e0317436bddbcaf5fdd4756bbda35d1f19baaf/nemo_rl/algorithms/grpo.py#L3300), [3305](https://github.com/NVIDIA-NeMo/RL/blob/d0e0317436bddbcaf5fdd4756bbda35d1f19baaf/nemo_rl/algorithms/grpo.py#L3305), [3321](https://github.com/NVIDIA-NeMo/RL/blob/d0e0317436bddbcaf5fdd4756bbda35d1f19baaf/nemo_rl/algorithms/grpo.py#L3321). This predates this PR but is worth calling out.

nemo_rl/algorithms/ppo.py
```diff
@@ -1493,10 +1493,15 @@ def ppo_train(
                 consumed_samples += master_config.ppo["num_prompts_per_step"]
                 timeout.mark_iteration()
 
+                ft_save_period = master_config.checkpointing.get("ft_save_period")
```
ashors1: [`ppo.py:1496`](https://github.com/NVIDIA-NeMo/RL/blob/d0e0317436bddbcaf5fdd4756bbda35d1f19baaf/nemo_rl/algorithms/ppo.py#L1496)

The PR adds `test_ft_save_period_triggers_periodic_saves` for sft, dpo, rm, distillation, and xtoken — but not for `ppo`, `grpo` (two training paths), or `grpo_sync`. Given `grpo.py` received the most structural changes, a unit test covering the `ft_save_period` save cadence there would be helpful.

nemo_rl/algorithms/dpo.py
```diff
@@ -643,10 +643,15 @@ def dpo_train(
                 ]
                 timeout.mark_iteration()
 
+                ft_save_period = master_config.checkpointing.get("ft_save_period")
```
ashors1: [`dpo.py:646`](https://github.com/NVIDIA-NeMo/RL/blob/d0e0317436bddbcaf5fdd4756bbda35d1f19baaf/nemo_rl/algorithms/dpo.py#L646) (same in all drivers)

`ft_save_period` is fetched from the config dict on every training step. It's immutable for the duration of training — consider reading it once before the loop alongside `save_period`.

ashors1 (commented):
Reviewed `ft_save_period` / `ft_keep_latest_k` changes. The `CheckpointManager` retention logic and `TestFTKeepLatestK` suite are solid. One bug in the async-finalization lifecycle across non-grpo drivers, one test gap, one minor style nit.

_Generated by Claude Code_

nemo_rl/algorithms/sft.py
ananthsub: this is safe as other algorithms wrap the train function with the checkpointer context, which implements shutdown() in `__exit__`, but i'll update to be explicit in the other algorithms too

ashors1 (approved):

PR #3103, opened 2026-07-07
fix: Port contention issues between Ray/vLLM/Gym and sandbox workers

saumishr:
# What does this PR do ?

Eliminates port contention between all NeMo RL services (Ray, vLLM, SGLang, Gym, Sandbox) and the kernel's ephemeral port range by pinning every service port below the observed ephemeral floor.

### Fix
Pin all service ports below 9000. Final layout (documented in `ray.sub` and `nemo_rl/distributed/virtual_cluster.py`):

- **Ray worker gRPC** moved to 2000-2999 (`MIN_WORKER_PORT`/`MAX_WORKER_PORT` in `ray.sub`), and `--min-worker-port`/`--max-worker-port` added to `ray start --head` so the head node is constrained to the same range as workers.
- **Ray GCS + management ports** moved below the floor: GCS/client 1200-1201, node/object/runtime-env/dashboard-agent/metrics ports to 1301-1312, dashboard-agent-listen to 1311.
- **Master address / TCPStore** moved to 1400-1999 (`DEFAULT_MASTER_PORT_RANGE_*` in `virtual_cluster.py`).
- **NeMo RL generation HTTP** moved to 3000-4999 (`port_range_low/high` in the example configs and `DEFAULT_GENERATION_PORT_RANGE_*` in `virtual_cluster.py`).
- **NeMo Gym HTTP** moved to 5000-5999 (example configs and `DEFAULT_GYM_PORT_RANGE_*`, referenced from `nemo_gym.py`).
- **Sandbox** Nginx stays at 6000; uWSGI workers moved to 6001+ by exporting `SANDBOX_WORKER_BASE_PORT=6001` and `NGINX_PORT=6000` into the sandbox srun (the container's `start-with-nginx.sh` already supports these env vars — no image change needed).
- **vLLM engine TP/DP rendezvous** moved from 20001 to 7000 (`DEFAULT_VLLM_PORT_RANGE_LOW` in `virtual_cluster.py`, consumed by `vllm_worker.py`).
- **SGLang engine ports** (server / nccl / dist_init) moved from 15000 to the same 7000-8999 band (vLLM and SGLang are mutually exclusive backends); free-port helper defaults lowered from 10000. The SGLang router HTTP port already allocated from 3000-4000 and is unchanged.
- Propagated the new bands to all example configs that hardcoded the old ranges, kept the frozen `tests/unit/reference_configs/*` copies in sync, and refreshed stale port-range comments.

# Issues

None

# Usage

None

PR #3089, opened 2026-07-06
perf: tune Ray GCS server configs for burst tolerance at scale

ananthsub:
## What

Tunes the Ray GCS server on the head node to survive the actor-registration burst during RL job init. All settings are `RAY_<config>` env-var exports with `${VAR:-default}` fallbacks, so they stay overridable. Ported from internal GitLab MR 199.

## Why

During init, rapid Ray actor creation generates thousands of `RegisterWorker` RPCs against the single GCS server. With Ray's defaults the RPC queue backs up, workers hit registration timeouts, the raylet kills them, and NCCL collectives inside vLLM can hang permanently.

## Changes

- **GCS gRPC thread pools**: `RAY_gcs_server_rpc_server_thread_num=64`, `RAY_num_server_call_thread=32` so the head drains registration bursts faster.
- **Timeouts → 120s**: `worker_register_timeout_seconds`, `gcs_rpc_server_connect_timeout_s`, `gcs_rpc_server_reconnect_timeout_s`, `gcs_server_request_timeout_seconds` (from 5–60s defaults) to survive the burst window while keeping failure detection under ~2 min.
- **Less background RPC noise during init**: `raylet_report_resources_period_milliseconds` 100ms→500ms and `task_events_report_interval_ms` 1s→5s.
- **Cap idle task-worker pool**: `RAY_num_workers_soft_limit=16` (default is CPU count) — task workers are short-lived (port discovery, GPU id queries), so we don't need one idle worker per core each holding a GCS connection.

Config reference: https://github.com/ray-project/ray/blob/master/src/ray/common/ray_config_def.h

PR #2916, opened 2026-06-24
feat: ThreadSafeTimer, container init timing, and Ray telemetry

saumishr:
# What does this PR do ?
Improves observability for large-scale Slurm/Ray GRPO runs

## Timer infrastructure
Extends Timer with optional context labels (rank, worker, hostname) and structured start/stop/record/mark logging for timeline debugging.
Adds ThreadSafeTimer for concurrent use in AsyncTrajectoryCollector (refit waits, generation-limit pauses, buffer-full backoff, failed trajectories).
Adds mark() for point-in-time events (e.g. worker crashes) correlatable across Ray actors.
## Container init breakdown
ray.sub records NRL_JOB_START_EPOCH at job submit and NRL_RAY_READY_EPOCH when Ray is ready on the head node.
Fixes propagation of NRL_RAY_READY_EPOCH into the driver container so the driver sees the same timestamp.
Adds log_container_init_timing(), called at the start of GRPO entrypoints, printing:
Slurm prologue (optional)
Cluster startup (orchestrator + container + Ray)
Driver startup (launch + Python imports)
Total wall time to first training code
## Async GRPO efficiency accounting
Driver and collector track idle categories: init/total, idle/buffer_starvation, idle/buffer_full_backoff, idle/refit_bubble, idle/refit_event_wait, idle/generation_limit_pause, idle/validation, wasted/failed_trajectory.
Each step prints and logs an efficiency summary (% wall time productive vs wasted) to stdout and WandB.
Policy workers emit timer spans around refit/prepare paths.
Safe when env vars are unset: init timing is skipped with a warning; timer behavior is unchanged for sync training.
# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
[...]

PR #3158, opened 2026-07-10
perf(ray.sub): tune GCS for large-scale multi-node init bursts

ananthsub:
## What

Extends the existing GCS tuning block in `ray.sub` with additional Ray configuration knobs that help the single head-node GCS survive the actor-registration burst during init at large node counts.

At large scale, the burst of `RegisterWorker` RPCs during init can overload the head GCS and cause cascading failures (GCS RPC queue backs up → workers time out during registration → raylet kills them). These settings widen the GCS reply path, add handler headroom, and trim steady background traffic during the init window.

## Changes

- **`gcs_server_rpc_client_thread_num`** — widen the GCS→raylet reply (gRPC client) pool to match the inbound server pool. Both scale from `CPUS_PER_WORKER`, consistent with the existing `gcs_server_rpc_server_thread_num`.
- **`gcs_max_active_rpcs_per_handler`** — set to 2× the auto default (`gcs_server_rpc_server_thread_num * 100`) so the handler queue doesn't cap out when thousands of actors register without staggering.
- **`grpc_keepalive_time_ms` / `grpc_keepalive_timeout_ms`** — 10s→60s / 20s→60s to cut keepalive ping load on the GCS during the burst.
- **`enable_timeline` / `event_stats`** — disable unused observability that otherwise emits per-actor/task writes to the GCS.
- **`ray_syncer_message_refresh_interval_ms` / `core_worker_internal_heartbeat_ms` / `health_check_period_ms`** — relax background syncer/heartbeat/health-check intervals during init.

## Notes

- Config reference: https://github.com/ray-project/ray/blob/master/src/ray/common/ray_config_def.h

### DIFF
--- a/ray.sub
+++ b/ray.sub
@@ -323,8 +323,10 @@ export RAY_event_stats="${RAY_event_stats:-false}"
 # core_worker_internal_heartbeat_ms (default 1s): worker->raylet heartbeat.
 #   Thousands of workers at 1s add indirect GCS load via raylet aggregation;
 #   5s is sufficient.
-# health_check_period_ms (default 3s): node->GCS health checks. 10s cuts the
-#   aggregate rate; slower node-failure detection is acceptable during init.
+# health_check_period_ms (default 3s): GCS->node health probes. 10s applies for
+#   the whole job, not just init: with defaults (threshold 5, timeout 10s),
+#   dead-node detection slows ~3x (~15s -> ~50s; up to ~100s if probes hang to
+#   the timeout) — accepted in exchange for GCS headroom during the burst.
 export RAY_ray_syncer_message_refresh_interval_ms="${RAY_ray_syncer_message_refresh_interval_ms:-10000}"
 export RAY_core_worker_internal_heartbeat_ms="${RAY_core_worker_internal_heartbeat_ms:-5000}"
 export RAY_health_check_period_ms="${RAY_health_check_period_ms:-10000}"
--- a/ray.sub
+++ b/ray.sub
@@ -300,9 +300,10 @@ export RAY_task_events_report_interval_ms="${RAY_task_events_report_interval_ms:
 export RAY_num_workers_soft_limit="${RAY_num_workers_soft_limit:-16}"
 
 # --- Reduce gRPC keepalive overhead ---
-# grpc_keepalive_time_ms (default 10s): interval between keepalive pings to the
-#   GCS. At many nodes these pings alone are a steady GCS load; 60s cuts the
-#   rate 6x during the init burst.
+# grpc_keepalive_time_ms (default 10s): interval at which the GCS gRPC server
+#   sends keepalive pings on its client connections (client->GCS pings are a
+#   separate knob, grpc_client_keepalive_time_ms, default 5min). At many nodes
+#   these pings are a steady GCS-side load; 60s cuts the rate 6x.
 # grpc_keepalive_timeout_ms (default 20s): keepalive ACK timeout. Keep it >= the
 #   ping interval so GCS load spikes don't trip false dead-connection detection.
 export RAY_grpc_keepalive_time_ms="${RAY_grpc_keepalive_time_ms:-60000}"
--- a/ray.sub
+++ b/ray.sub
@@ -251,9 +251,20 @@ fi
 # GB200/Grace).
 # gcs_server_rpc_server_thread_num: threads that poll for incoming gRPC requests.
 # num_server_call_thread: threads that send gRPC replies.
-# gcs_max_active_rpcs_per_handler auto-scales to rpc_thread_num * 100.
+# gcs_server_rpc_client_thread_num (default max(1, cores/4)): threads the GCS
+#   uses as a gRPC client to send replies back to raylets. Match it to the
+#   inbound server pool so the reply path does not become the bottleneck once
+#   the request pool is widened.
 export RAY_gcs_server_rpc_server_thread_num="${RAY_gcs_server_rpc_server_thread_num:-$CPUS_PER_WORKER}"
 export RAY_num_server_call_thread="${RAY_num_server_call_thread:-$(( CPUS_PER_WORKER / 2 > 0 ? CPUS_PER_WORKER / 2 : 1 ))}"
+export RAY_gcs_server_rpc_client_thread_num="${RAY_gcs_server_rpc_client_thread_num:-$CPUS_PER_WORKER}"
+
+# --- GCS active RPC headroom ---
+# gcs_max_active_rpcs_per_handler (default gcs_server_rpc_server_thread_num*100):
+#   max concurrent in-flight RPCs per GCS handler. Give 2x the auto default so
+#   the handler queue does not cap out when thousands of actors register
+#   simultaneously without staggering.
+export RAY_gcs_max_active_rpcs_per_handler="${RAY_gcs_max_active_rpcs_per_handler:-$(( CPUS_PER_WORKER * 200 ))}"
 
 # --- Registration and RPC timeouts ---
 # With thread tuning, GCS should drain even the worst burst in ~60s. 120s gives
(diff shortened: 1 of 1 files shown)


ray.sub
```diff
@@ -251,9 +251,20 @@ fi
 # GB200/Grace).
 # gcs_server_rpc_server_thread_num: threads that poll for incoming gRPC requests.
 # num_server_call_thread: threads that send gRPC replies.
-# gcs_max_active_rpcs_per_handler auto-scales to rpc_thread_num * 100.
+# gcs_server_rpc_client_thread_num (default max(1, cores/4)): threads the GCS
+#   uses as a gRPC client to send replies back to raylets. Match it to the
+#   inbound server pool so the reply path does not become the bottleneck once
+#   the request pool is widened.
```
terrykong: [`ray.sub:254-257`](https://github.com/NVIDIA-NeMo/RL/blob/67785be29386c82df932e43542a556a11044e5b3/ray.sub#L254-L257)

Wording fix: these threads don't send replies — they poll for replies to the GCS's own **outbound** client RPCs. Sending replies to inbound requests is `num_server_call_thread`, which line 253 already documents correctly ("threads that send gRPC replies"), so the new text contradicts the line above it.

Evidence (ray 2.55.1):
- Definition, [ray_config_def.h#L386-L390](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/common/ray_config_def.h#L386-L390): "Number of polling threads for raylet + worker clients on the GCS. These threads **poll for replies** and copy from the socket buffer to create the proto Reply object."
- Consuming code, [gcs_server.cc#L74-L82](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/gcs/gcs_server.cc#L74-L82): the value sizes the `ClientCallManager` used by `raylet_client_pool_` — the GCS's outbound-RPC machinery.

The tuning itself makes sense (the reply-polling pool for GCS→raylet calls also benefits from widening); only the rationale is inverted.

**Action:** reword to describe the outbound reply-polling role:
```suggestion
# gcs_server_rpc_client_thread_num (default max(1, cores/4)): threads that poll
#   replies to the GCS's own outbound RPCs to raylets/workers. Match it to the
#   inbound server pool so the outbound reply path does not become the
#   bottleneck once the request pool is widened.
```

ray.sub
```diff
@@ -288,6 +299,35 @@ export RAY_task_events_report_interval_ms="${RAY_task_events_report_interval_ms:
 #   concurrent per node is plenty.
 export RAY_num_workers_soft_limit="${RAY_num_workers_soft_limit:-16}"
 
+# --- Reduce gRPC keepalive overhead ---
+# grpc_keepalive_time_ms (default 10s): interval between keepalive pings to the
+#   GCS. At many nodes these pings alone are a steady GCS load; 60s cuts the
+#   rate 6x during the init burst.
```
terrykong: [`ray.sub:303-305`](https://github.com/NVIDIA-NeMo/RL/blob/67785be29386c82df932e43542a556a11044e5b3/ray.sub#L303-L305)

Direction fix: `grpc_keepalive_time_ms` is the interval at which the GCS gRPC **server sends** keepalive pings on its connections — not pings "to the GCS" from nodes. Client→GCS pings are a separate, untouched knob, `grpc_client_keepalive_time_ms`, which already defaults to 5 minutes.

Evidence (ray 2.55.1):
- [ray_config_def.h#L825-L827](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/common/ray_config_def.h#L825-L827): "grpc keepalive sent interval for server. This is only configured in GCS server now."
- [grpc_server.h#L213-L216](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/rpc/grpc_server.h#L213-L216): "The interval to send a new gRPC keepalive timeout from **server -> client**" (applied to the server builder at [grpc_server.cc#L78](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/rpc/grpc_server.cc#L78)).
- [ray_config_def.h#L836-L840](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/common/ray_config_def.h#L836-L840): `grpc_client_keepalive_time_ms` default 300000 (5 min), `grpc_client_keepalive_timeout_ms` 120000.

The conclusion (less GCS keepalive overhead) still holds — the GCS server pings every connection, so 10s→60s cuts that 6x — but as written the comment attributes the load to node-side pings that only happen every 5 min.

**Action:**
```suggestion
# grpc_keepalive_time_ms (default 10s): interval at which the GCS gRPC server
#   sends keepalive pings on its client connections (client->GCS pings are a
#   separate knob, grpc_client_keepalive_time_ms, default 5min). At many nodes
#   these pings are a steady GCS-side load; 60s cuts the rate 6x.
```

ray.sub
```diff
@@ -288,6 +299,35 @@ export RAY_task_events_report_interval_ms="${RAY_task_events_report_interval_ms:
...
+export RAY_event_stats="${RAY_event_stats:-false}"
+
+# --- Reduce syncer, heartbeat, and health-check background traffic ---
+# ray_syncer_message_refresh_interval_ms (default 3s): periodic cross-node state
+#   refresh via ray_syncer. 10s reduces this 3x.
+# core_worker_internal_heartbeat_ms (default 1s): worker->raylet heartbeat.
+#   Thousands of workers at 1s add indirect GCS load via raylet aggregation;
+#   5s is sufficient.
+# health_check_period_ms (default 3s): node->GCS health checks. 10s cuts the
+#   aggregate rate; slower node-failure detection is acceptable during init.
```
terrykong: [`ray.sub:325-326`](https://github.com/NVIDIA-NeMo/RL/blob/67785be29386c82df932e43542a556a11044e5b3/ray.sub#L325-L326)

Two accuracy points on this comment:

1. **Direction:** these are GCS→node probes, not "node->GCS": the health-check manager runs on the GCS and sends gRPC health `Check`s to each raylet ([gcs_health_check_manager.cc#L122-L171](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/gcs/gcs_health_check_manager.cc#L122-L171)).
2. **Scope:** the env var persists for the whole job, so the relaxed detection isn't init-scoped. Per the failure loop ([gcs_health_check_manager.cc#L200-L224](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/gcs/gcs_health_check_manager.cc#L200-L224): a node is declared dead after `failure_threshold`=5 consecutive failed probes, each rescheduled `period_ms` after completion) and the defaults ([ray_config_def.h#L929-L936](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/common/ray_config_def.h#L929-L936)), dead-node detection slows ~3x for the entire run: ~15s → ~50s when probes fail fast, up to ~65s → ~100s when probes hang to the 10s `health_check_timeout_ms` deadline. (`initial_delay` only applies when a node first registers.)

If the whole-job tradeoff is intended (probably fine), worth stating it so a future reader doesn't assume it reverts after init — the same whole-job scope applies to the syncer/heartbeat knobs above.

**Action:** reword to reflect direction + whole-job effect:
```suggestion
# health_check_period_ms (default 3s): GCS->node health probes. 10s applies for
#   the whole job, not just init: with defaults (threshold 5, timeout 10s),
#   dead-node detection slows ~3x (~15s -> ~50s; up to ~100s if probes hang to
#   the timeout) — accepted in exchange for GCS headroom during the burst.
```

terrykong (commented):
Team review (5 agents). Overall this is a clean tuning PR — nice follow-up to #3089. We verified all 10 new `RAY_*` config names exist verbatim in the pinned ray 2.55.1 and every "(default ...)" claim in the comments matches [`ray_config_def.h`](https://github.com/ray-project/ray/blob/ray-2.55.1/src/ray/common/ray_config_def.h); the block follows the #3089 conventions (`CPUS_PER_WORKER`-derived scaling, `${VAR:-}` overridability). Disabling `enable_timeline`/`event_stats` was also checked against NeMo-RL's own Ray telemetry (#2916) — no dependency.

Two non-blocking notes:
- Consider copying the commit message's "validated at 3K-node scale" note into the PR description — it helps future maintainers know the regime these values target.
- No conflict with open #3103 (port relayout): it touches different regions of `ray.sub`.

Inline comments are all low-severity wording/consistency suggestions.

_Generated by Claude Code_

saumishr (approved):
LGTM!

PR #3162, opened 2026-07-11
feat(tq): train prefetch microbatches

zyzhou5:
# What does this PR do ?

Synchronous TQ training currently fetches, materializes, and distributes the complete DP shard before training begins. This creates an upfront delay and requires workers to retain the full shard in CPU memory. This PR adds opt-in, depth-one microbatch prefetching. Tensor payloads are retrieved one packed microbatch at a time, while preparation of microbatch N+1 overlaps with training microbatch N.

The existing full-shard path remains unchanged when prefetching is disabled.

  ## Configuration
```
  policy:
    train_microbatch_prefetch:
      enabled: true
      depth: 1
```


nemo_rl/algorithms/grpo_sync.py
```diff
@@ -115,6 +115,52 @@ def _raise_if_message_level_advantage_penalties_enabled(
     )
 
 
+def _summarize_train_microbatch_prefetch_source_metrics(
```
ZhiyuLi-Nvidia, line 118: For better reuse in grpo async, grpo etc, is there a better place to track these metrics?

Considering in prefetch class or nemo_rl/data_plane/observability.py

ZhiyuLi-Nvidia: Oh, you already tracked metrics in prefetch class. Feel free to skip.

nemo_rl/models/megatron/train_microbatch_prefetch.py
```diff
@@ -0,0 +1,728 @@
...
+
+def _tensor_bytes(data: BatchedDataDict[Any]) -> int:
+    return sum(
+        value.numel() * value.element_size()
+        for value in data.values()
+        if isinstance(value, torch.Tensor)
+    )
+
+
+class TrainMicrobatchPrefetcher(Iterator[PrefetchedMicrobatch]):
```
ZhiyuLi-Nvidia: have we considered use a customized pytorch dataloader and get prefetch (and other optimization) for free? Is there like any blocker?

nemo_rl/models/megatron/train_microbatch_prefetch.py
```diff
@@ -0,0 +1,728 @@
...
+            for batch in self.plan.global_batches
+            for mb_idx, sample_ids in enumerate(batch.microbatch_sample_ids)
+        )
+        self._queue: queue.Queue[PrefetchedMicrobatch | _PrefetchFailure] = queue.Queue(
+            maxsize=1
+        )
+        # One initial permit prepares MB0. Depth one releases after consuming
+        # MB N; depth zero releases on the source only when MB N+1 is requested.
+        self._producer_permit = threading.Semaphore(1)
+        self._stop_event = threading.Event()
```
ZhiyuLi-Nvidia, line 415: would that be easier to use thread safe queue like multiprocessing.Queue and better for readbility?

nemo_rl/models/megatron/train_microbatch_prefetch.py
```diff
@@ -0,0 +1,728 @@
...
+                f"{item.sample_ids})"
+            )
+        if item.data.size != len(expected_ids):
+            raise TrainMicrobatchPrefetchError(
+                f"prefetched microbatch has {item.data.size} rows for "
+                f"{len(expected_ids)} keys"
+            )
+
+        attach_message_log_view(item.data)
+        trace_tq_prefetch_payload(keys=item.sample_ids, data=item.data)
```
ZhiyuLi-Nvidia, line 830: which part is overlapped in prefetch?
* tq get 
* sequence packing
* broadcast

ZhiyuLi-Nvidia:
Awesome change, thanks for the change.

My main suggestion is to simplify the change by isolating the prefetch Iterator/DataLoader abstraction from the TQ-based “next batch” logic. That separation would make the code easier to maintain, read, and extend.

Can we also cover a profile (nsys) to show prefetching provides better overlap for optimization?

nemo_rl/models/megatron/train_microbatch_prefetch.py
zyzhou5: tq get and broadcast

nemo_rl/models/megatron/train_microbatch_prefetch.py
zyzhou5: would that require using a new process for prefetching?

nemo_rl/models/megatron/train_microbatch_prefetch.py
zyzhou5: made some update regarding this issue

zyzhou5:
/ok to test [43df55e](https://github.com/NVIDIA-NeMo/RL/pull/3162/commits/43df55ec6d3a16041aba1b309bb77491b6398842)

zyzhou5:
/ok to test [3f8625e](https://github.com/NVIDIA-NeMo/RL/pull/3162/commits/3f8625e3db2c4ee938a070c18db4171509b585d0)

PR #3166, opened 2026-07-11
feat(megatron): add nccl refit backend for non-colocated generation

shanmugamr1992:
## Summary

Adds an `nccl` option for `refit_backend` (megatron generation), alongside the existing `gloo` and `nvshmem` copy services used to stream training weights into the Megatron inference engine in non-colocated setups.

- Megatron-Core already ships an `NCCLCopyService` (moves weight bytes via CUDA-tensor `batch_isend_irecv`), but NeMo-RL only wired up `gloo`/`nvshmem`. This wires up `nccl` too.
- The `nccl` path registers an NCCL/CUDA backend on the manually-built cross-world refit `ProcessGroup` (in addition to the GLOO/CPU backend that the object collectives in `prepare_swap_model_weights` rely on), so the CUDA P2P sends/recvs dispatch to NCCL. GLOO remains the default backend.
- `refit_backend` config `Literal` widened to `["gloo", "nccl", "nvshmem"]`; docstrings updated.

## Motivation

On the nano-3.5 30B-A3B MoE (non-colocated, `transformer_impl=inference_optimized`, `num_cuda_graphs=-1`), the `nvshmem` copy service's per-step refit is slow **and degrades step-over-step**, while `nccl` refit is fast and flat — and needs no `NVSHMEM_MAX_TEAMS` tuning.

Measured over 10 GRPO steps (per-step `transfer_and_update_weights`, steps 2→10):

| refit backend | per-step refit | trend | warm step time |
|---|---|---|---|
| `nccl` | ~16–19s | flat | ~37–40s |
| `nvshmem` | ~95s → 135s+ | grows steadily | ~120–170s |

## Test plan

- [x] Validated end-to-end on a GB200 cluster: 10-step non-colocated GRPO run with `refit_backend=nccl` completed successfully; refit stable at ~16–19s/step vs nvshmem ~135s-and-growing. Learning dynamics equivalent (reward, gen-KL error).
- [ ] CI (`/ok to test`)


tdene (approved):

terrykong:
@yuki-97 to take a look

terrykong:
@shanmugamr1992 is there any test path that exercises this? is this a common configuration we expect users to run?

shanmugamr1992:
@terrykong So nccl is performing well now for refit. NVShmem is old and has not been maintained. So Peter once he is back from vacation will be taking a look into it so in the meanwhile we want to have this as an option I could add a test with NCCL refit however I would defer to after a couple of weeks once we decide between NCL and NVSHMEM it should be fairly simple to add that test.

terrykong (approved):

PR #3168, opened 2026-07-11
fix: add vLLM config context during collective refit

snowmanwwg:
## What changed

Wrap the collective-refit call to `process_weights_after_loading()` in
`set_current_vllm_config(self.model_runner.vllm_config)`.

Extend the focused unit test to verify that post-load processing runs inside
that context and that the context exits before FP8 KV-cache post-processing.

## Why

A Nano-v3 async GRPO run using current `main` (`ad23de8da`) and NeMo RL's latest
nightly container initialized all 256 Ray actors, both 128-worker model groups,
and all 13 NeMo Gym servers. The initial policy-to-generation refit then failed
on every vLLM worker with:

```
Current vLLM config is not set.
```

This originates from `process_weights_after_loading()`, which newer vLLM
versions require to execute under `set_current_vllm_config()`. Other post-load
paths in the same extension already establish this context; the collective
refit path did not.

Reproduction:

- Slurm job: `13777880`
- W&B: https://wandb.ai/nvidia/grpo-dev-wenweng/runs/88atirl1
- Runtime error: `Updating weights for the generation policy failed during refit`

## Impact

Collective Megatron-to-vLLM weight synchronization can complete with newer
vLLM containers instead of returning `False` from every generation worker
before the first rollout.

## Validation

- `git diff --check`
- `python -m py_compile nemo_rl/models/generation/vllm/vllm_backend.py tests/unit/models/generation/test_vllm_backend.py`
- Focused test updated to assert config-context enter/process/exit ordering.
- The prescribed vLLM test invocation was attempted, but dependency resolution
  is currently blocked by HTTP 403 while fetching metadata for the pinned
  FlashAttention aarch64 wheel:
  `flash_attn-2.8.1+cu13torch2.10cxx11abiTRUE-cp313-cp313-linux_aarch64.whl`.

### DIFF
--- a/nemo_rl/models/generation/vllm/vllm_backend.py
+++ b/nemo_rl/models/generation/vllm/vllm_backend.py
@@ -465,9 +465,12 @@ class VllmInternalWorkerExtension:
                 process_weights_after_loading,
             )
 
-            process_weights_after_loading(
-                self.model_runner.model, self.model_config, self.device
-            )
+            from vllm.config import set_current_vllm_config
+
+            with set_current_vllm_config(self.model_runner.vllm_config):
+                process_weights_after_loading(
+                    self.model_runner.model, self.model_config, self.device
+                )
             self._maybe_process_fp8_kv_cache()
 
         except Exception as e:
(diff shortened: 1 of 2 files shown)


macandro96 (dismissed):
lgtm!

terrykong (approved):

PR #3171, opened 2026-07-12
fix(ppo): restore optimizer state on checkpoint resume

bg51717:
# What does this PR do ?

Fixes PPO checkpoint resume so both the policy and value models restore their own weights and optimizer state from the checkpoint.

## Bug

Megatron-Core stores optimizer and scheduler state as part of the distributed checkpoint under `weights/iter_0000000/`, rather than in standalone `policy/optimizer` and `value/optimizer` directories. `common.pt` contains the optimizer marker/common metadata and scheduler state, while sharded optimizer tensors such as Adam moments are stored in the checkpoint's `.distcp` payloads.

PPO previously determined whether optimizer state was available by checking only for those standalone directories. For an MCore checkpoint, that check returned false even though the weights checkpoint contained valid optimizer and scheduler state. As a result, PPO restored model weights but passed `optimizer_path=None`, silently initializing fresh optimizers and schedulers instead of restoring Adam moments and scheduler progress.

For post-warmup checkpoints, this reset both policy and value optimizer/scheduler state. For checkpoints saved during critic warmup, the policy had not yet been trained or saved, but the already-trained value optimizer and scheduler were still reset. This broke optimizer-state continuity across PPO restarts and could change the subsequent optimization trajectory.

PPO also resolved policy and value checkpoint paths independently instead of using the shared checkpoint resolver, whose MCore path already knows how to detect optimizer state embedded in `common.pt`.

## Fix

The main changes are:

- Extend `CheckpointManager.get_resume_paths()` with a model-component selector, while preserving the policy component as the default for existing callers such as GRPO.
- Restore policy and value weights and optimizer states through the normal `Policy` and `Value` initialization paths.
- Always save policy weights during critic warmup, while saving policy optimizer state only after policy training has started.
- Reject `ckpt_assume_constant_structure=true` when PPO critic warmup and optimizer checkpointing are both enabled, because the policy checkpoint structure changes when optimizer state first appears.
- Add checkpoint-path and Megatron value-worker checkpoint round-trip coverage.

**Pre checks**:

- [x] No documentation changes are required.

# Test

- PPO algorithm, checkpoint-path resolver, and Megatron value-worker unit tests: **49 passed**. The MCore value checkpoint round-trip verifies exact equality after save → resume → re-save for model tensors, Adam `exp_avg`/`exp_avg_sq`, optimizer metadata, and scheduler state.
- PPO Megatron functional test: **9/9 checks passed**.


bg51717:
/ok to test [10aaaa0](https://github.com/NVIDIA-NeMo/RL/pull/3171/commits/10aaaa02f58ad49b60d7619a009cf87eb5de0901)

nemo_rl/algorithms/ppo.py
```diff
@@ -230,6 +229,26 @@ def setup(
...
+        assert not (
+            ppo_config["policy_training_start_step"] > 0
+            and master_config.checkpointing["enabled"]
+            and master_config.checkpointing["save_optimizer"]
+            and "checkpoint" in policy_megatron_config
+            and policy_megatron_config["checkpoint"].get(
+                "ckpt_assume_constant_structure"
+            )
+        ), (
+            "policy.megatron_cfg.checkpoint.ckpt_assume_constant_structure=true "
```
yuki-97, line 246: am I understand correctly that value ckpt is using `policy.megatron_cfg.checkpoint.ckpt_assume_constant_structure` instead of `value.megatron_cfg.checkpoint.ckpt_assume_constant_structure`? and `value.megatron_cfg.checkpoint` is not respected to?

tests/unit/models/value/test_megatron_value_worker.py
```diff
@@ -675,25 +676,17 @@ def test_value_worker_train_decreases_loss(value_setup):
     ids=["2gpu_dp2"],
 )
 def test_value_worker_checkpoint_save_and_load(value_setup, tmp_path):
-    """Full save → shutdown → restore round-trip with state correctness checks.
+    """Round-trip model, optimizer, and scheduler state through an MCore checkpoint.
```
yuki-97, line 679: could you also help to update the test in `tests/unit/models/value/test_dtensor_value_worker.py`?

tests/unit/utils/test_checkpoint.py
```diff
@@ -409,16 +412,79 @@ def test_get_resume_paths_missing_optimizer(checkpoint_manager, checkpoint_dir):
...
+    expected_optimizer_path = checkpoint_path / model_component / "optimizer"
+    assert not expected_optimizer_path.exists()
+
+    weights_path, optimizer_path = CheckpointManager.get_resume_paths(
+        checkpoint_path,
+        model_component=model_component,
+    )
+
+    assert weights_path == expected_weights_path
+    assert optimizer_path == expected_optimizer_path
```
yuki-97, line 485: Missing coverage for MCore + `save_optimizer=False` — `common.pt` exists without an `"optimizer"` key.

That branch falls through `checkpoint.py:194-200` to warn + `optimizer_path = None`. Neither `test_get_resume_paths_missing_optimizer` (no `common.pt`) nor `test_get_resume_paths_embedded_megatron_optimizer` (always writes `"optimizer"`) covers it, and it's exactly the config `ppo.py:249` recommends.

yuki-97 (commented):
nice catch and thanks for fixing! left some comments.

tests/unit/models/value/test_megatron_value_worker.py
bg51717: Updated the existing DTensor value test to use the shared resume resolver and verify optimizer/scheduler state across save → resume → re-save. It reuses the existing `2gpu_dp2` case.

tests/unit/utils/test_checkpoint.py
bg51717: With `save_optimizer=False`, returning `(weights_path, None)` is intentional: weights resume while optimizer/scheduler start fresh. The existing missing-optimizer test covers this contract, so we won’t add another fixture.

bg51717:
/ok to test [5de1a5b](https://github.com/NVIDIA-NeMo/RL/pull/3171/commits/5de1a5bc33e1966047b5afaa6c632c3fde142b99)

nemo_rl/algorithms/ppo.py
bg51717: No, value does not use the policy checkpoint config. `make_policy_like_config()` preserves `value.megatron_cfg`, so value respects its own `value.megatron_cfg.checkpoint`. The guard is policy-only because, during critic warmup, policy checkpoints change from weights-only to weights plus optimizer/scheduler once policy training starts. Value trains and saves optimizer state from step 0, so its checkpoint structure remains constant.

yuki-97: make sense, thanks for the explanation.

yuki-97 (approved):

PR #2917, opened 2026-06-24
fix(async-grpo): starvation diagnostics, gym logging, and dataloader exhaustion

saumishr:
# What does this PR do ?

Addresses async GRPO runs that appear hung during buffer fill or between steps: adds actionable diagnostics, reduces logging overhead, and fails fast when the dataloader is exhausted instead of spinning forever.

## Buffer starvation diagnostics
ReplayBuffer.get_debug_info() reports per-buffer stats: trajectory duration, max generation tokens per turn, and turn depth (mean/median/max/p95).
Rollout paths record max_gen_tokens_per_turn and turn-count percentiles for long multi-turn Gym trajectories.
Collector records trajectory_duration_s per buffered group.
Driver logs starvation context (buffer state + diagnostics) while waiting for trajectories.

## Gym logging cost
Skips writing train_data_step*.jsonl in async GRPO when env.should_log_nemo_gym_responses is true (NeMo Gym handles full response logging).
Avoids serializing very large Gym payloads every step when that flag is set.

## Dataloader exhaustion fix
Collector sets data_exhausted when the train iterator completes without an explicit stop.
Exposes get_status() for driver-side checks.
Driver raises RuntimeError with collector status during initial buffer fill and mid-training starvation waits, instead of an infinite loop when max_num_epochs is too low for the buffer size.

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

PR #3173, opened 2026-07-13
fix: fix rollout metrics fail in unit test

yuki-97:
fix rollout metrics fails in `CI:L1` introduced in https://github.com/NVIDIA-NeMo/RL/pull/2917.

### DIFF
--- a/nemo_rl/experience/metric_utils.py
+++ b/nemo_rl/experience/metric_utils.py
@@ -24,6 +24,16 @@ from wandb import Histogram
 def calculate_single_metric(
     values: Sequence[float | int], batch_size: int, key_name: str
 ) -> dict:
+    """Compute summary statistics for a metric as slash-prefixed keys.
+
+    Args:
+        values: Per-sample metric values to aggregate.
+        batch_size: Denominator for the mean (sum(values) / batch_size, not len(values)); stddev still uses len(values).
+        key_name: Prefix for the returned metric keys (e.g. "total_reward").
+
+    Returns:
+        Dict mapping "{key_name}/{stat}" to its value for stat in mean, max, min, median, stddev (nan for a single value), and histogram (a wandb.Histogram).
+    """
     return {
         f"{key_name}/mean": sum(values) / batch_size,
         f"{key_name}/max": max(values),
--- a/nemo_rl/experience/metric_utils.py
+++ b/nemo_rl/experience/metric_utils.py
@@ -1,0 +1,43 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Shared aggregation helpers for rollout metrics."""
+
+import math
+import statistics
+from collections.abc import Sequence
+
+from wandb import Histogram
+
+
+def calculate_single_metric(
+    values: Sequence[float | int], batch_size: int, key_name: str
+) -> dict:
+    return {
+        f"{key_name}/mean": sum(values) / batch_size,
+        f"{key_name}/max": max(values),
+        f"{key_name}/min": min(values),
+        f"{key_name}/median": statistics.median(values),
+        f"{key_name}/stddev": statistics.stdev(values) if len(values) > 1 else math.nan,
+        f"{key_name}/histogram": Histogram(values),
+    }
+
+
+def pct(values: Sequence[float | int], p: float) -> float:
+    """Percentile helper for buffer starvation diagnostics."""
+    if not values:
+        return 0.0
+    sorted_v = sorted(values)
+    idx = min(int(len(sorted_v) * p / 100), len(sorted_v) - 1)
+    return float(sorted_v[idx])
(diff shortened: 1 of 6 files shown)


nemo_rl/experience/metric_utils.py
```diff
@@ -0,0 +1,43 @@
...
+import math
+import statistics
+from collections.abc import Sequence
+
+from wandb import Histogram
+
+
+def calculate_single_metric(
+    values: Sequence[float | int], batch_size: int, key_name: str
+) -> dict:
```
terrykong, line 26: [`nemo_rl/experience/metric_utils.py:24`](https://github.com/NVIDIA-NeMo/RL/blob/ca42c2dc83ed531956251211faac0b758a21897c/nemo_rl/experience/metric_utils.py#L24)

Nice clean extraction — the move is byte-for-byte faithful and the test key-set fix checks out. One small ask: `calculate_single_metric` is now a public cross-file interface (imported by [`rollout_manager.py`](https://github.com/NVIDIA-NeMo/RL/blob/ca42c2dc83ed531956251211faac0b758a21897c/nemo_rl/experience/rollout_manager.py#L28) and the unit test), but it has no docstring, while its sibling `pct` got one in this PR. It would also help to document the non-obvious denominator quirk: `/mean` divides by `batch_size`, not `len(values)`, while `/stddev` uses `len(values)`.

**Action:** add a Google-style docstring, e.g.:

```suggestion
def calculate_single_metric(
    values: Sequence[float | int], batch_size: int, key_name: str
) -> dict:
    """Compute summary statistics for a metric as slash-prefixed keys.

    Args:
        values: Per-sample metric values to aggregate.
        batch_size: Denominator for the mean. Note the mean is
            ``sum(values) / batch_size`` (NOT ``len(values)``), while
            ``stddev`` uses ``len(values)``.
        key_name: Prefix for the returned metric keys (e.g. ``"total_reward"``).

    Returns:
        Dict mapping ``"{key_name}/{stat}"`` to its value for ``stat`` in
        ``mean``, ``max``, ``min``, ``median``, ``stddev`` (``nan`` for a
        single value) and ``histogram`` (a ``wandb.Histogram``).
    """
```

nemo_rl/experience/metric_utils.py
```diff
@@ -0,0 +1,43 @@
...
+        f"{key_name}/mean": sum(values) / batch_size,
+        f"{key_name}/max": max(values),
+        f"{key_name}/min": min(values),
+        f"{key_name}/median": statistics.median(values),
+        f"{key_name}/stddev": statistics.stdev(values) if len(values) > 1 else math.nan,
+        f"{key_name}/histogram": Histogram(values),
+    }
+
+
+def pct(values: Sequence[float | int], p: float) -> float:
```
terrykong, line 47: [`nemo_rl/experience/metric_utils.py:37`](https://github.com/NVIDIA-NeMo/RL/blob/ca42c2dc83ed531956251211faac0b758a21897c/nemo_rl/experience/metric_utils.py#L37)

`pct` is now a public helper but has no direct unit test — it is only exercised through GPU-gated integration tests, whereas `calculate_single_metric` is covered by [`TestCalculateSingleMetric`](https://github.com/NVIDIA-NeMo/RL/blob/ca42c2dc83ed531956251211faac0b758a21897c/tests/unit/experience/test_rollouts.py#L71). Its edge cases (empty list → `0.0`, index clamping) are pure Python and cheap to cover.

**Action:** consider adding a `TestPct` class next to `TestCalculateSingleMetric` in `tests/unit/experience/test_rollouts.py` (verified passing locally, 7 passed):

```python
from nemo_rl.experience.metric_utils import pct


class TestPct:
    """Unit tests for the pct percentile helper."""

    def test_empty_returns_zero(self):
        assert pct([], 95) == 0.0

    def test_returns_float_for_int_input(self):
        result = pct([3, 1, 2], 50)
        assert isinstance(result, float)
        assert result == 2.0

    def test_sorts_unsorted_input(self):
        assert pct([5, 1, 3], 95) == 5.0

    def test_p95_small_list_clamps_to_max(self):
        assert pct([1, 2, 3, 4, 5], 95) == 5.0

    def test_p99_clamps_to_last_index(self):
        assert pct([1, 2, 3, 4, 5], 99) == 5.0

    def test_single_value(self):
        assert pct([42], 95) == 42.0

    def test_median_like_p50(self):
        assert pct([10, 20, 30, 40], 50) == 30.0
```

Not a merge blocker.

terrykong (commented):
Reviewed with a team of agents (rl-expert, test-agent, bug-finder, devil's advocate). Summary:

- The fix is verified: on `main`, `run_async_nemo_gym_rollout` already emits `max_gen_tokens_per_turn/*` and `turns_per_sample/p95`,`/p99` (from #2917) but the test's expected key-set was never updated, so `test_run_async_nemo_gym_rollout` fails; this PR adds exactly the missing keys and restores manager/original parity for `test_async_rollout_manager_matches_original`.
- The helper extraction into `nemo_rl/experience/metric_utils.py` is byte-for-byte faithful, with no stale references to the deleted `_calculate_single_metric`/`_pct` anywhere in the repo, no circular imports, and correctly registered in `pyrefly.toml`. `TestCalculateSingleMetric` passes locally; all pre-commit hooks pass.
- Informational, no action needed in this PR: `max_gen_tokens_per_turn` intentionally has different key shapes across the two manager paths because each mirrors its `rollouts.py` original (3 keys for `run_async_multi_turn_rollout`, 7 for `run_async_nemo_gym_rollout`) — unifying them would break the parity test. Also, `nemo_rl/algorithms/async_utils/replay_buffer.py` still carries its own identical `_pct`; could be deduped onto `metric_utils.pct` in a follow-up.

Two optional low-severity suggestions inline (docstring, CPU-only test for `pct`).

_Generated by Claude Code_

terrykong:
aside from ai review do you think we can add the test that was skipped in the "FAST" suite? since we know it crashed in L1, might be good to promote it to fast

nemo_rl/experience/metric_utils.py
yuki-97: 34376b1

nemo_rl/experience/metric_utils.py
yuki-97: 34376b1

yuki-97:
> aside from ai review do you think we can add the test that was skipped in the "FAST" suite? since we know it crashed in L1, might be good to promote it to fast

updated in 570d3b8

terrykong (approved):

PR #2827, opened 2026-06-15
perf: reduce srun overhead in ray.sub and gate driver on sandbox readiness

ananthsub:
# What does this PR do ?

`ray.sub` previously scales srun/slurmctld RPCs with the node count: it launches one srun per worker (each followed by a 3s sleep), slept a fixed 120s before workers, polls head readiness via `srun --overlap test -f`, and polls cluster status via repeated `srun ... ray status`. For larger jobs, this dominates cluster bringup time and risks throttling slurmctld.

This PR replaces the per-node fan-out with shared-filesystem file signaling so srun calls stay constant relative to the node count:

1. Launch all workers with a single batched srun (--nodes/--ntasks=N-1, --ntasks-per-node=1, --exclude=head); workers self-identify via SLURM_PROCID / SLURMD_NODENAME. This change drops the per-worker loop, the 3s stagger, and the 120s sleep.
2.  Start the head without --block and touch STARTED_RAY_HEAD only after the GCS is listening, so workers connect on the first try; workers poll that file before `ray start --address` and retry the GCS connect.
3. A head-side ray-status sidecar publishes the live worker_units count to a file; the submit host reads it instead of issuing `srun ... ray status` RPCs.
4. The head container runs the driver inline (via a driver_command.sh file) and the submit host just waits on the head srun, removing the dedicated driver srun.
5. Sandbox readiness becomes a blocking gate: each per-node sandbox task polls its local port and touches `SANDBOX_READY_<host>`; the head waits for all instances before launching the driver. The `SANDBOX_PORTS_DIR` signal dir is always mounted and exported, independent of the `SANDBOX_EXTRA_MOUNTS` / `SANDBOX_ENV_VARS` knobs.

This change assumes that the LOG_DIR is on a shared filesystem visible across compute nodes.

This change also does the following:
1. cleans stale Ray session state inside the head and worker retry loops
2. Adds 0-indexed, zero-padded worker logs for natural sorting
3. Suffix LOG_DIR with SLURM_RESTART_COUNT so a requeue does not clobber the previous attempt's logs and signal files

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
[...]

PR #3178, opened 2026-07-13
fix: ensure unique TOPO_RANK per node

### LINKED ISSUE #3155: `dapo-deepseek-v3-64n8g.v2` fail in main branch
`dapo-deepseek-v3-64n8g.v2` is fixed in #3154 in `r0.7.0` branch, but will easily meet the below error in main branch.

`(MegatronPolicyWorker pid=1391870) [2026-07-10 06:19:36] pool0-00008:1391870:1393239 [7] init.cc:904 NCCL WARN Duplicate GPU detected : rank 15 and rank 7 both on CUDA device e4000`

yuki-97:
## Issue
closes #3155

## Summary
Under Slurm, the head srun and the batched worker srun each start their own `SLURM_PROCID` counter at 0, so the head node and worker[0] both computed `TOPO_RANK=0`. **Ray silently drops value-0 custom resources**, so those two nodes lost their `topo_rank` and collapsed into `virtual_cluster.py`'s UNKNOWN topology bucket.

**This latent bug was invisible until #2827.** The pre-#2827 worker srun started workers serially (with a `sleep 3` stagger), so bundles landed on the alive-nodes in a blocked order and Megatron's rank→node mapping happened to be per-node blocked. After #2827 workers register with Ray in parallel; on the two UNKNOWN nodes the unified-PG placement path now produces an **interleaved** bundle→node mapping, and Megatron-based recipes with cross-node model parallelism (e.g. DAPO DSV3 64n8g) crash at NCCL init with `Duplicate GPU detected`.

## `topo_rank` in the Ray node dump

| Node | Before (crashed) | After (this fix) |
|---|---|---|
| head | **None** (Ray dropped value=0) | 1 |
| worker[0] | **None** (Ray dropped value=0) | 2 |
| worker[1..62] | 1..62 | 3..64 |

## Consequence on rank→node mapping (before fix)

On the **two nodes that both lost their `topo_rank`** (head + worker[0]), the unified PG placed bundles in an interleaved order, so the ranks Megatron assigned to those two physical machines were striped like this:

| Physical node | Ranks landed on it |
|---|---|
| Node A | 0, 2, 4, 6, 8, 10, 12, 14 |
| Node B | 1, 3, 5, 7, 9, 11, 13, 15 |

Every rank *R* and rank *R+8* in the same NCCL comm ends up on the same physical machine, and NCCL reports them on the same physical GPU → `Duplicate GPU detected: rank R and rank R+8` at NCCL init.

After the fix all 64 nodes have distinct `topo_rank` → `virtual_cluster` sees a full topology → blocked mapping (ranks 0-7 on one node, 8-15 on next, ...) → no collision.

## Fix
`ray.sub` `topology_probe.sh`: nest a `SLURMD_NODENAME == $head_node` check inside the `SLURM_PROCID` fallback branch. Head hardcodes to 1; workers get `SLURM_PROCID + 2`. `$head_node` is submit-time interpolated into the generated script; `SLURMD_NODENAME` is runtime.

### DIFF
--- a/ray.sub
+++ b/ray.sub
@@ -399,7 +399,7 @@ ip_head=$head_node_ip:$PORT
 # TOPO_RANK: Topology-aware infrastructure rank.
 #   Fallback chain:
 #   1. SLURM_TOPOLOGY_ADDR (block.node format) -> block_num * 10^10 + node_num
-#   2. SLURM_PROCID (SLURM without topology plugin)
+#   2. SLURM_PROCID + 2 on workers, head pinned to 1 (SLURM without topology plugin)
 #   3. Hostname digits (non-SLURM fallback)
 #
 # TODO(ansubramania): Harden the SLURM_TOPOLOGY_ADDR digit-extraction logic for open source.
--- a/ray.sub
+++ b/ray.sub
@@ -426,7 +426,14 @@ if [[ -n "\${SLURM_TOPOLOGY_ADDR:-}" && "\${SLURM_TOPOLOGY_ADDR_PATTERN:-}" == "
     TOPO_RANK=\$(( 10#\$_block_digits * 10000000000 + 10#\$_node_digits ))
   fi
 elif [[ -n "\${SLURM_PROCID:-}" ]]; then
-  TOPO_RANK=\$(( 10#\$SLURM_PROCID ))
+  # Head srun and worker srun each start their own SLURM_PROCID counter at 0.
+  # Head hardcoded to 1; workers get +2 so worker[0]=2, worker[62]=64 -- keeps
+  # all nodes at unique values >=1 (Ray drops value-0 custom resources).
+  if [[ "\$SLURMD_NODENAME" == "$head_node" ]]; then
+    TOPO_RANK=1
+  else
+    TOPO_RANK=\$(( 10#\$SLURM_PROCID + 2 ))
+  fi
 else
   _hostname_digits=\$(hostname | tr -dc '0-9')
   if [[ -n "\$_hostname_digits" ]]; then
--- a/nemo_rl/distributed/virtual_cluster.py
+++ b/nemo_rl/distributed/virtual_cluster.py
@@ -122,7 +122,9 @@ Nodes sharing the same key belong to the same NVLink switch fabric (e.g. one GB2
 TOPO_RANK_KEY = "topo_rank"
 """Ray resource key for the SLURM topological rank.
 Derived from ``SLURM_TOPOLOGY_ADDR`` (when ``SLURM_TOPOLOGY_ADDR_PATTERN=block.node``),
-falling back to ``SLURM_PROCID`` or hostname digits.
+falling back to ``SLURM_PROCID + 2`` on worker nodes (head node is pinned to ``1``),
+then to hostname digits when SLURM is unavailable. Values are always ``>= 1`` so that
+Ray does not drop the custom resource (Ray drops value-0 custom resources).
 Used to sort nodes within and across NVLink domains so rank assignment follows physical topology."""
 
 NVLINK_DOMAIN_UNKNOWN = "unknown"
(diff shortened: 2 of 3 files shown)


ray.sub
```diff
@@ -426,7 +426,14 @@ if [[ -n "\${SLURM_TOPOLOGY_ADDR:-}" && "\${SLURM_TOPOLOGY_ADDR_PATTERN:-}" == "
...
   fi
 elif [[ -n "\${SLURM_PROCID:-}" ]]; then
-  TOPO_RANK=\$(( 10#\$SLURM_PROCID ))
+  # Head srun and worker srun each start their own SLURM_PROCID counter at 0.
+  # Head hardcoded to 1; workers get +2 so worker[0]=2, worker[62]=64 -- keeps
+  # all nodes at unique values >=1 (Ray drops value-0 custom resources).
+  if [[ "\$SLURMD_NODENAME" == "$head_node" ]]; then
+    TOPO_RANK=1
+  else
+    TOPO_RANK=\$(( 10#\$SLURM_PROCID + 2 ))
```
terrykong, line 435: [`ray.sub:435`](https://github.com/NVIDIA-NeMo/RL/blob/fe2776c5b95e6c9a11be4079dddc763da4926f9b/ray.sub#L435)

The new `+2` worker offset makes an existing test comment stale: [`test_topology_placement.py:121`](https://github.com/NVIDIA-NeMo/RL/blob/fe2776c5b95e6c9a11be4079dddc763da4926f9b/tests/unit/distributed/test_topology_placement.py#L121) in `test_slurm_procid_value_as_float` says `SLURM_PROCID fallback: $(( 10#7 )) -> TOPO_RANK=7`, which describes the pre-fix mapping — after this change a worker with `SLURM_PROCID=7` gets TOPO_RANK **9**, and the head is pinned to 1. The assertion itself still passes (pure `int(7.0)` round-trip); only the comment misdocuments the probe.

**Action:** update that comment to match the new mapping, e.g.:
```python
# SLURM_PROCID fallback (worker): $(( 10#7 + 2 )) -> TOPO_RANK=9, stored as 9.0 by Ray
# (head node is pinned to TOPO_RANK=1)
```

terrykong (commented):
Reviewed with a multi-agent pass (rl-expert, bug-finder, test-agent, adversarial verifier). **The fix is correct, minimal, and verified end-to-end** — thanks for the excellent root-cause writeup, the before/after `topo_rank` table made this easy to check.

Verified premises:
- Head srun ([`ray.sub:904`](https://github.com/NVIDIA-NeMo/RL/blob/fe2776c5b95e6c9a11be4079dddc763da4926f9b/ray.sub#L904)) and worker srun ([`ray.sub:916-924`](https://github.com/NVIDIA-NeMo/RL/blob/fe2776c5b95e6c9a11be4079dddc763da4926f9b/ray.sub#L916-L924)) are separate job steps with independent `SLURM_PROCID` counters, so head and worker[0] both computed 0 pre-fix.
- Ray 2.55.1 silently drops zero-value custom resources: [`resource_and_label_spec.py:91-95`](https://github.com/ray-project/ray/blob/ray-2.55.1/python/ray/_private/resource_and_label_spec.py#L91-L95) (`if resource_quantity != 0`).
- Missing `topo_rank` → [`TOPO_RANK_UNKNOWN=-1`](https://github.com/NVIDIA-NeMo/RL/blob/fe2776c5b95e6c9a11be4079dddc763da4926f9b/nemo_rl/distributed/virtual_cluster.py#L131) → UNKNOWN bucket → interleaved placement, matching the NCCL `Duplicate GPU` crash in #3155.
- head=1 / workers `SLURM_PROCID+2` is unique and ≥1 for any node count; the `SLURMD_NODENAME == head_node` guard compares within the same scontrol NodeName namespace and the head task is pinned with `-w`, so the match is reliable.

Two optional suggestions inline (a stale test comment and a ready-made regression test). Full `pre-commit` suite passes at this head.

_Generated by Claude Code_

terrykong (approved):

ray.sub
terrykong: fixed in 160753ad4e923c2dd63758603c28ee0804138129

PR #3091, opened 2026-07-07
feat(sglang): Megatron+SGLang + MXFP8

xiuhu17:
## Summary

Ports the megatron + sglang weight-update scheme into the main-tracking branch: Megatron trainers export HF-named tensors through AutoBridge, optionally quantize them to MXFP8 online, and push them to colocated SGLang engines via Ray CUDA-IPC (`weight_transfer_mode: ipc`) or to disaggregate engines via a trainer-rank-0 NCCL broadcast group (`weight_transfer_mode: broadcast`). The FSDP colocated path is unified onto the same connect/send machinery, replacing the previous HTTP streaming design.

mxfp8 checkpoint conversion, weight update and fine-grained layer precision is adopted from: 
https://github.com/radixark/miles/issues/615 which is done by @zianglih. 

## What changed

**New scheme components**
- `mxfp8_quantization_core.py` — single source of truth for per-tensor MXFP8 decisions (`should_quantize` / `quantize_mxfp8`, flashinfer swizzle-free kernel, UE8M0 scales, [1,32] blocks), shared by offline conversion and online refit
- `mxfp8_setup.py` — offline HF→MXFP8 checkpoint conversion with hash-qualified caching; `ensure_mxfp8_checkpoint` bootstraps SGLang from a quantized checkpoint at setup
- `megatron_sglang_weight_iterator.py` — walks `bridge.export_hf_weights`, applies the MXFP8 policy per tensor (expanding `(weight, weight_scale_inv)` pairs), and packs buckets sized by post-quantization bytes
- `fault_tolerance.py` — `RolloutHealthMonitor` daemon that health-checks engines and restarts dead actors; recovered engines are reconnected on the next refit via `get_updatable_engines_and_lock` / `recover_updatable_engines`

**Refit plumbing (`policy/utils.py`)**
- `connect_colocate_topology` — layout-keyed per-engine Gloo gather subgroups, supporting explicit `engine_gpu_counts` / `engine_gpu_offsets` and rebuild-on-recover
- `send_hf_buckets_via_ipc_actor_impl` — gather-to-src + Ray actor RPC with per-chunk `ray.get` (IPC lifetime safety) and engine-result validation
- `broadcast_hf_buckets_via_distributed_impl` + NCCL group connect/disconnect helpers — trainer-rank-0 broadcast, serialized against engine recovery by a Ray `Lock` actor

**Drivers and dispatch**
[...]

PR #2997, opened 2026-06-29
refactor(sglang): allocate ports from reserved ranges

xiuhu17:
## Summary

The SGLang backend allocated ports with its own helpers (`find_available_port`,
`is_port_available`, `get_free_port`) that probe OS-ephemeral ports, instead of
the shared range-based primitives in `virtual_cluster.py` already used by vLLM
and NeMo-Gym. This re-introduced the TOCTOU-prone allocation that #2380 removed,
just in a different place.

This PR rebases SGLang onto the shared helpers and reserved port map, while
keeping the SGLang-specific bits the shared helpers don't cover (contiguous
dp_attention port blocks, a per-node cursor, IPv6 bracketing, non-Ray host lookup).

Closes #2874

## What changed

- **`virtual_cluster.py`**: added `_get_free_consecutive_ports_local()` (returns a
  contiguous block of bindable ports in a reserved range — the one capability the
  existing helpers lacked) and a dedicated router/Prometheus band (28001–30000).
- **`ray_utils.py`**: removed `find_available_port` / `is_port_available` /
  `get_free_port`; `get_current_node_ip` now delegates to `_get_node_ip_local()`.
  `get_host_info` and the NOSET env list are unchanged.
- **`ip_port_utils.py` / `sglang_worker.py`**: allocate through the shared helper
  with `port_range_low/high` + a per-node cursor instead of a hard-coded `base_port=15000`.
- **`sglang_generation.py`**: reads `port_range_low/high` from the generation config.
- **`sglang_router.py`**: router/Prometheus ports from `_get_free_port_local` over their band.
- **Tests**: added `TestGetFreeConsecutivePortsLocal`; updated SGLang tests/helpers
  that used the removed functions.
- **patch**: This patch rebases `_patch_sglang_custom_all_reduce_v2_tms_cudagraph` onto the final form of [sgl-project/sglang#27948](https://github.com/sgl-project/sglang/pull/27948) (tracking [radixark/miles#1176](https://github.com/radixark/miles/issues/1176)): instead of the earlier revision's `set_cuda_graph_register_inputs` toggle threaded through five files, it applies the upstream Python-only fix to `custom_all_reduce_v2.py` — adding the `envs` import and `self.tms_cudagraph` flag, and gating both `set_cuda_graph_capture(True)` calls on `not self.tms_cudagraph` — so the kernel's capture flag stays off under TMS and we no longer touch the fragile jit_kernel C++/`.cuh` sources.

## Port layout

| Port class | Band |
|---|---|
[...]

PR #3187, opened 2026-07-14
feat(sglang): engine fault tolerance for rollouts

xiuhu17:
## What

Adds `RolloutHealthMonitor`: a daemon thread that health-checks each SGLang engine and restarts hung or dead actors during rollout, gated by `policy.generation.sglang_cfg.use_fault_tolerance` (off by default).

- `fault_tolerance.py`: monitor lifecycle (start/stop/pause/resume), `health_generate` probing, engine kill/restart bookkeeping
- `SGLangGeneration`: engine recovery (`_recover`, `recover_updatable_engines`, `get_updatable_engines_and_lock`), `num_new_engines` tracking, monitor wiring
- `ray_utils.Lock`: cooperative Ray lock serializing weight refits against engine recovery
- `SGLangGenerationWorker._simulate_crash`: test-only crash injection
- config + exemplar YAML keys for the health checker

Ft design is adapted from [miles](https://github.com/radixark/miles)

## Stack

Stacked on #2997 (reserved-band port allocation) — **review only the last commit** (`feat(sglang): engine fault tolerance for rollouts`); earlier commits belong to #2997. Follow-ups: #3190 (megatron refit) → #3188 (mxfp8) → #3189 (comm-split fix).

Split out of #3091.


nemo_rl/models/generation/sglang/fault_tolerance.py
```diff
@@ -0,0 +1,200 @@
...
+        logger.info("Stopping RolloutHealthMonitor...")
+        assert self._stop_event is not None
+        self._stop_event.set()
+        # Also clear pause to let the thread exit
+        if self._pause_event:
+            self._pause_event.clear()
+        timeout = self._check_timeout + self._check_interval + 5
+        self._thread.join(timeout=timeout)
+        if self._thread.is_alive():
+            logging.warning(
```
yuki-97: **[BUG]** bypasses the module logger set up at line 22 and used everywhere else in this file.

```suggestion
            logger.warning(
```

nemo_rl/models/generation/sglang/utils/ray_utils.py
```diff
@@ -32,6 +34,28 @@
 ]
 
 
+@ray.remote
```
yuki-97: nit: `@ray.remote` classes are not tracked by coverage — please add `# pragma: no cover` per convention (e.g. `replay_buffer.py:633`, `math_environment.py:61`, `virtual_cluster.py:153`).

```suggestion
@ray.remote  # pragma: no cover
```

nemo_rl/models/generation/sglang/fault_tolerance.py
```diff
@@ -0,0 +1,200 @@
...
+    def _check_engine_health(self, rollout_engine_id, engine) -> None:
+        if engine is None:
+            logger.info(f"Skipping health check for engine {rollout_engine_id} (None)")
+            return
+
+        try:
+            ray.get(engine.health_generate.remote(timeout=self._check_timeout))
+        except Exception as e:
+            logger.error(
+                f"Health check failed for rollout engine {rollout_engine_id} (ray timeout or error). Killing actor. Exception: {e}"
```
yuki-97: the `timeout=` here goes to `requests.get` inside `health_generate` (see `sglang_worker.py:200-202`), not to `ray.get`. the log message says "(ray timeout or error)" which suggests otherwise. was a ray-level timeout intended (to defend against actor-side python hangs), or should the wording be updated? if the former, `ray.get(future, timeout=...)` would need to be added.

nemo_rl/models/generation/sglang/config.py
```diff
@@ -67,6 +67,13 @@ class SglangSpecificArgs(TypedDict):
     # Nested server/router configs. Kept under ``sglang_cfg`` so YAML and call
     # sites have a single sglang namespace instead of three sibling fields.
     sglang_server_config: SGLangServerConfig
+
+    # Fault tolerance (RolloutHealthMonitor). Off by default; when enabled, a
+    # daemon thread health-checks each engine and restarts hung/dead actors.
+    use_fault_tolerance: NotRequired[bool]
+    rollout_health_check_interval: NotRequired[int]
+    rollout_health_check_timeout: NotRequired[int]
+    rollout_health_check_first_wait: NotRequired[int]
```
yuki-97, line 83: let's move this to under `sglang_router_config` to keep the config order more clear.

nemo_rl/models/generation/sglang/fault_tolerance.py
```diff
@@ -0,0 +1,200 @@
...
+import threading
+
+import ray
+
+from nemo_rl.models.generation.sglang.config import SGLangConfig
+
+logger = logging.getLogger(__name__)
+
+
+class RolloutHealthMonitor:
```
yuki-97, line 25: could you add some unit tests to test the lifecycle of this?

nemo_rl/models/generation/sglang/config.py
```diff
@@ -67,6 +67,13 @@ class SglangSpecificArgs(TypedDict):
     # Nested server/router configs. Kept under ``sglang_cfg`` so YAML and call
     # sites have a single sglang namespace instead of three sibling fields.
     sglang_server_config: SGLangServerConfig
+
+    # Fault tolerance (RolloutHealthMonitor). Off by default; when enabled, a
+    # daemon thread health-checks each engine and restarts hung/dead actors.
+    use_fault_tolerance: NotRequired[bool]
```
yuki-97: this one seems always required.
```suggestion
    use_fault_tolerance: bool
```

nemo_rl/models/generation/sglang/sglang_generation.py
```diff
@@ -292,7 +310,83 @@ def check_weights(self, action: str):
...
+        """Return engines eligible for weight updates."""
+        return (
+            self.engines,
+            self.rollout_engine_lock,
+            self.num_new_engines,
+            self.engine_gpu_counts,
+            self.engine_gpu_offsets,
+        )
+
+    def recover_updatable_engines(self):
```
yuki-97, line 370: I think this function should be used in fault tolerance, but I don't seem to see it be called int this PR?

also I'm curious don't we need some related changes in `generate/generate_async` and also somewhere else if workers crash? so that the worker crash won't cause the whole training crash, which is fault tolerance for.

nemo_rl/models/generation/sglang/sglang_generation.py
```diff
@@ -292,7 +310,83 @@ def check_weights(self, action: str):
...
+        """Restart any dead rollout engines and update ``num_new_engines``."""
+        self.health_monitoring_pause()
+
+        self._recover()
```
yuki-97, line 375: don't we need to call `health_monitoring_resume` after here? otherwise once we call `recover_updatable_engines` the monitor is paused forever.

nemo_rl/models/generation/sglang/sglang_generation.py
```diff
@@ -292,7 +310,83 @@ def check_weights(self, action: str):
...
+            self.engines,
+            self.rollout_engine_lock,
+            self.num_new_engines,
+            self.engine_gpu_counts,
+            self.engine_gpu_offsets,
+        )
+
+    def recover_updatable_engines(self):
+        """Restart any dead rollout engines and update ``num_new_engines``."""
+        self.health_monitoring_pause()
```
yuki-97: I tend to think we can just call `self._health_monitor.pause()` here instead of adding those two functions (and `self._health_monitor.resume()` after _recover).
and also don't need check if `_health_monitor` is not None, since recover will only be used when enable `use_fault_tolerance`, and in this case `_health_monitor` is always not None.

nemo_rl/models/generation/sglang/sglang_generation.py
```diff
@@ -292,7 +310,83 @@ def check_weights(self, action: str):
...
+
+        return (
+            self.engines,
+            self.rollout_engine_lock,
+            self.num_new_engines,
+            self.engine_gpu_counts,
+            self.engine_gpu_offsets,
+        )
+
+    def clear_updatable_num_new_engines(self):
```
yuki-97, line 385: where will this function be called?

nemo_rl/models/generation/sglang/sglang_worker.py
```diff
@@ -304,6 +304,17 @@ def resume_memory_occupation(self, tags: list[str] | None = None):
     def check_weights(self, action: str):
         return self._make_request("weights_checker", {"action": action})
 
+    def _simulate_crash(self):
```
yuki-97, line 378: where will this function be called?

nemo_rl/models/generation/sglang/sglang_generation.py
```diff
@@ -292,7 +310,83 @@ def check_weights(self, action: str):
...
+                ]
+            )
+            ray.get(
+                [
+                    engine.resume_memory_occupation.remote(tags=["weights"])
+                    for engine in new_engines
+                ]
+            )
+
+    def get_updatable_engines_and_lock(self):
```
yuki-97, line 360: where will this function be called?

nemo_rl/models/generation/sglang/sglang_generation.py
```diff
@@ -122,6 +130,14 @@ def __init__(
         if init_handles:
             ray.get(init_handles)
 
+        # Serializes weight refits against engine recovery across processes.
+        self.rollout_engine_lock = Lock.options(num_cpus=1, num_gpus=0).remote()
```
yuki-97, line 141: looks it only be used in fault tolerance case, so shall we also init with None and do real init under `if sglang_cfg["sglang_cfg"].get("use_fault_tolerance"):` like `self._health_monitor`?

nemo_rl/models/generation/sglang/sglang_generation.py
```diff
@@ -122,6 +130,14 @@ def __init__(
...
+
+        if sglang_cfg["sglang_cfg"].get("use_fault_tolerance"):
```
yuki-97, line 143: could you add some unit tests for sglang generation for use_fault_tolerance=True and force a crash to see if fault tolerance works fine?

yuki-97 (commented):
thanks @xiuhu17 , I rebased main for easily review and left some comments.

nemo_rl/models/generation/sglang/sglang_generation.py
xiuhu17: Definitely, that makes sense

nemo_rl/models/generation/sglang/fault_tolerance.py
xiuhu17: Thanks for suggestion, fixed

nemo_rl/models/generation/sglang/sglang_generation.py
xiuhu17: It is inside the https://github.com/NVIDIA-NeMo/RL/pull/3187/changes#diff-ccc7bd9f2ab92a1a2c34c1899dda3e50ebf3558d605ffebf359a29e5db25418aR141

nemo_rl/models/generation/sglang/sglang_worker.py
xiuhu17: It is for testing purpose and used in test: test_crashed_engine_is_detected_recovered_and_generation_survives

nemo_rl/models/generation/sglang/fault_tolerance.py
xiuhu17: Added to tests/unit/models/generation/sglang/test_fault_tolerance.py

nemo_rl/models/generation/sglang/sglang_generation.py
xiuhu17: Added to tests/unit/models/generation/sglang/test_fault_tolerance_real.py

nemo_rl/models/generation/sglang/sglang_generation.py
xiuhu17: It is now used in https://github.com/NVIDIA-NeMo/RL/pull/3187/changes#diff-ccc7bd9f2ab92a1a2c34c1899dda3e50ebf3558d605ffebf359a29e5db25418aR133

yeah, it is a relative simple ft. It only recover rollout engines during weight update, recover it with newer weight from trainer

nemo_rl/models/generation/sglang/sglang_generation.py
xiuhu17: https://github.com/NVIDIA-NeMo/RL/pull/3187/changes#diff-8a10a17c21c663421694b10784c45d12ceb9ef1408ce4ecac14d391bbb549025R894

Thanks for pointing out, self._health_monitor.resume() is now called in prepare_for_generation

nemo_rl/models/generation/sglang/sglang_generation.py
xiuhu17: It is now called inside the refit. https://github.com/NVIDIA-NeMo/RL/pull/3187/changes#diff-ccc7bd9f2ab92a1a2c34c1899dda3e50ebf3558d605ffebf359a29e5db25418aR126

nvpohanh:
@xiuhu17 could you fix the DCO issue? thanks

PR #3189, opened 2026-07-14
fix(sglang): avoid ncclCommSplit hang in the refit weight-update group

xiuhu17:
## What

When the trainer's default process group is eager-initialized with a bound device id (`device_id=` passed to `torch.distributed.init_process_group`), torch's `_new_process_group_helper` silently sets `Options.split_from` to the default group's NCCL communicator, and the first collective on the side weight-update group then issues `ncclCommSplit`. The split is collective over the **parent (trainer) communicator**, but only trainer rank 0 is a member of the refit group, so rank 0 blocks forever in the split bootstrap all-gather (`commGetSplitInfo`, NCCL 2.28.9) while the remaining trainer ranks hit the watchdog.

Fix: clear `default_pg.bound_device_id` for the duration of the `_new_process_group_helper` call (restored in `finally`) so the helper's split predicate (`is_initialized() and _get_default_group().bound_device_id`) sees a lazily-initialized default group and never selects the split path.

Current main initializes the Megatron default PG without `device_id`, so this is latent hardening: any trainer base that binds the default group would otherwise trip it silently.

Root cause analysis by @Kh4L (co-authored).

## Verification

- torch 2.10.0 and 2.11.0 have the identical split predicate; `setBoundDeviceId` is a pure field assignment accepting `None` in both (no propagation to live backends)
- gloo interop smoke (this helper ↔ sglang `init_custom_process_group` replica): broadcast + destroy OK
- split-predicate regression: with a device-bound default group, the raw helper path consults `_get_split_source` (positive control); the fixed path never does, and `bound_device_id` is restored

## Stack

Stacked on #3188 (mxfp8) on #3190 (megatron refit) on #3187 (fault tolerance) on #2997 (ports) — **review only the last commit** (`fix(sglang): avoid ncclCommSplit hang...`).

### DIFF
--- a/examples/configs/grpo_math_1B_sglang.yaml
+++ b/examples/configs/grpo_math_1B_sglang.yaml
@@ -38,6 +38,8 @@ policy:
         pause_generation_mode: retract
         num_gpus: 2
         num_gpus_per_engine: ${policy.generation.sglang_cfg.tp_size}
+        # "ipc" for colocated inference, "broadcast" for disaggregate GPUs.
+        weight_transfer_mode: ipc
       sglang_router_config:
         use_external_router: false
     colocated:
--- a/examples/configs/grpo_math_1B_sglang.yaml
+++ b/examples/configs/grpo_math_1B_sglang.yaml
@@ -30,6 +30,11 @@ policy:
       rollout_health_check_interval: 60
       rollout_health_check_timeout: 60
       rollout_health_check_first_wait: 60
+      # Weight precision for rollout/refit. scheme=bf16 (default) sends BF16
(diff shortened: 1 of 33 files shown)


yuki-97 (commented):
hi @xiuhu17 @Kh4L, thanks for fixing! we also meet similar issue in https://github.com/NVIDIA-NeMo/RL/pull/3073.

curious are we able to split this fix out instead of stack on previous PRs since it looks not based on the new features. so that we can get it in more quickly.

xiuhu17:
Hi @yuki-97 the init process group is used in disaggregated mode(which is supported for megatron+Sglang backend). We could fix it first here but it might be a dead code.

PR #3190, opened 2026-07-14
feat(sglang): megatron backend weight refit for sglang rollouts

xiuhu17:
## What

Replaces the HTTP weight-streaming refit with two first-class SGLang refit paths driven from grpo via `_refit_sglang_dispatch`:

- **colocated** (`weight_transfer_mode: ipc`): Ray CUDA-IPC buckets via `send_hf_buckets_via_ipc_actor_impl` + `SGLangColocatedWeightSynchronizer`, generalized Gloo gather topology (`connect_colocate_topology`) supporting FSDP and Megatron layouts
- **disaggregate** (`weight_transfer_mode: broadcast`): trainer-rank-0 NCCL weight-update group (side-by-side `init_process_group`, `connect/disconnect_rollout_engines_from_distributed`) broadcasting AutoBridge-restored HF buckets; engines join via `init_weights_update_group`
- `MegatronSGLangHfWeightIterator`: AutoBridge export walk bucketed by post-transformation size
- engine-side worker endpoints (`update_weights_from_distributed`, `pause/continue_generation`, `post_process_weights`, weight versioning)
- `NCCL_CUMEM_ENABLE=0` alignment between trainer and sglang scheduler
- drops `stream_weights_via_http` / `set_rollout_num_gpus_per_engine`
- fix: clear default_pg.bound_device_id for the duration of the _new_process_group_helper call (restored in finally) so the helper's split predicate (is_initialized() and _get_default_group().bound_device_id) sees a lazily-initialized default group and never selects the split path. Root cause analysis by @Kh4L (co-authored).

Quantization-related parameters (`target_precision`, `sglang_quantization_cfg`) are inert bf16 plumbing here; the mxfp8 implementation lands in #3188.

Megatron broadcast design is adapted from [miles](https://github.com/radixark/miles)

## Result
image
Screenshot 2026-08-08 at 10 48 46 AM
Screenshot 2026-08-08 at 10 47 43 AM

## Stack

Stacked on #3187 (fault tolerance) on #2997 (ports) — **review only the last commit** (`feat(sglang): megatron backend weight refit for sglang rollouts`). Follow-ups: #3188 (mxfp8)

Supersedes the refit part of #2446; split out of #3091.


nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -2135,6 +2165,162 @@ def _expert_idx(megatron_name: str) -> str:
...
+            return
+
+        if group_name is not None:
+            self._sglang_dist_group_name = group_name
+
+        if self._sglang_dist_group is not None:
+            disconnect_rollout_engines_from_distributed(
+                group_name=self._sglang_dist_group_name,
+                model_update_group=self._sglang_dist_group,
+                rollout_engines=self._sglang_dist_engines,
```
yuki-97, line 2258: `update_weights_to_sglang_colocated` and `update_weights_to_sglang_distributed` (2286) accept `target_precision` and `sglang_quantization_cfg` kwargs but neither reads them.

The DTensor V2 sibling at `dtensor_policy_worker_v2.py:1171-1176` raises `NotImplementedError` for non-bf16 and `del sglang_quantization_cfg` with an "accepted for dispatch parity, bf16-only" comment, giving callers a truthful contract.

**Suggest fix:** mirror the DTensor pattern — raise on `target_precision != "bf16"` and `del sglang_quantization_cfg` with the parity comment; or actually wire them through so the caller can trust the args.

nemo_rl/models/policy/utils.py
```diff
@@ -521,230 +521,520 @@ def rebuild_cuda_tensor_from_ipc(
...
+        for engine in rollout_engines
+    ]
+    try:
+        dist.destroy_process_group(model_update_group)
+    except Exception:
+        pass
+    try:
+        ray.get(refs)
+    except Exception:
+        pass
```
yuki-97: `disconnect_rollout_engines_from_distributed` swallows every exception from both `destroy_process_group` and the engine-side `destroy_weights_update_group` RPCs with bare `except Exception: pass`.

Partial teardown failures (leaked NCCL groups, orphaned engine state) surface later as ghost groups on the next `connect` attempt with no signal at teardown time.

**Suggest fix:** narrow to the specific expected exception (e.g. `RuntimeError` from torch when the group is already dead) and at minimum `logger.warning(...)` the swallowed error.

nemo_rl/models/generation/sglang/config.py
```diff
@@ -31,6 +31,10 @@ class SGLangServerConfig(TypedDict):
     pause_generation_mode: str
     # Total number of GPUs allocated to inference across all engines.
     num_gpus: NotRequired[int]
+    # "ipc" -> CUDA-IPC to the colocated SGLang HTTP server (default for
+    # colocated inference). "broadcast" -> NCCL broadcast over a shared
+    # weight-update group (used when SGLang engines run on disaggregate GPUs).
+    weight_transfer_mode: NotRequired[str]
```
yuki-97: `weight_transfer_mode: NotRequired[str]` (and `pause_generation_mode: str` at line 31) admit any string, but only `"ipc"`/`"broadcast"` and `"retract"`/`"in_place"` are meaningful. Runtime rejection lives inline at `grpo.py:1402` for the first — static callers can pass an unknown value undetected.

**Suggest fix:** tighten to `Literal["ipc", "broadcast"]` and `Literal["retract", "in_place"]`.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -513,6 +520,29 @@ def __init__(
...
+            # tensor is serialized.
+            from nemo_rl.models.generation.sglang.utils.train_utils import (
+                monkey_patch_torch_reductions,
+            )
+
+            monkey_patch_torch_reductions()
+        self._sglang_dist_group: Any = None
+        self._sglang_dist_group_name: str = "nemo_rl_sglang"
+        self._sglang_dist_engines: list = []
+        self._sglang_weight_version: int = 0
```
yuki-97: SGLang state (`_sglang_ipc_state`, `_sglang_dist_group`, `_sglang_dist_group_name`, `_sglang_dist_engines`, `_sglang_weight_version`) is initialized on every Megatron worker regardless of backend.

The DTensor V2 sibling at `dtensor_policy_worker_v2.py:377-388` wraps the parallel init in `if generation_backend == "sglang"`, keeping the "these attributes exist iff SGLang is the rollout backend" contract explicit. Divergence here wastes 5 attributes on every non-SGLang Megatron run and weakens the SGLang-only boundary that the sibling makes obvious.

**Suggest fix:** nest the five assignments under the same `generation_cfg.get("backend") == "sglang"` guard used for the monkey-patch.

nemo_rl/weight_sync/megatron_sglang_refit.py
```diff
@@ -0,0 +1,146 @@
...
+        engine_gpu_offsets,
+    ) = fetch_updatable_engines_with_recover(policy_generation)
+
+    if num_new_engines > 0:
+        policy.connect_sglang_rollout_engines(
+            engine_gpu_counts=engine_gpu_counts,
+            engine_gpu_offsets=engine_gpu_offsets,
+        )
+        policy_generation.clear_updatable_num_new_engines()
+        assert policy_generation.num_new_engines == 0, (
```
yuki-97: `assert policy_generation.num_new_engines == 0` immediately after `policy_generation.clear_updatable_num_new_engines()` asserts the previous line's own write on the driver-side facade (`clear_updatable_num_new_engines` just sets `self.num_new_engines = 0`, `sglang_generation.py:384-387`). Same pattern at line 122 in this file and line 57 of `dtensor_sglang_refit.py`.

**Suggest fix:** drop the three asserts.

nemo_rl/models/generation/sglang/config.py
yuki-97: Related: `weight_transfer_mode` has no declared default anywhere in the schema. Its effective default lives inline at `grpo.py:1399` as `.get("weight_transfer_mode", "ipc" if colocated else "broadcast")`.

Per config-conventions, defaults should live in the exemplar YAMLs (or on the BaseModel when this type is migrated).

**Suggest fix:** set `weight_transfer_mode` explicitly in every sglang exemplar YAML (e.g. `examples/configs/grpo_math_1B_sglang.yaml` and the two `grpo-qwen2.5-math-1.5b-instruct-{1n8g,2n8g}-fsdp2tp1-sglang.yaml` recipes).

nemo_rl/models/policy/workers/megatron_sglang_weight_iterator.py
```diff
@@ -0,0 +1,92 @@
...
+        for hf_param_name, tensor in self._bridge.export_hf_weights(
+            self._models,
+            show_progress=False,
+            conversion_tasks=self._conversion_tasks,
+        ):
+            # AutoBridge yields plain ``torch.Tensor`` for Megatron (no
+            # DTensor / async-collective wrapping), so no ``.wait()`` is
+            # needed here. The previous ``hasattr(tensor, "wait")`` check
+            # was a copy-from-FSDP residue.
+            yield [(hf_param_name, tensor)]
```
yuki-97, line 96: `_iter_finalized_hf_named_tensors` return type is `Iterator[list[tuple[str, torch.Tensor]]]` but the body always yields a length-1 list, and the docstring says "each item produces exactly one finalized pair" — which contradicts the returned-list-of-pairs shape. The only caller (`iter_hf_weight_buckets`) then unpacks with `for name, tensor in finalized:` as if the list may have >1 entry.

**Suggest fix:** either flatten to yield `(name, tensor)` directly (drop the list wrapping) or add an assert that the list is length-1 and update the docstring to explain what the future non-length-1 case would be.

nemo_rl/weight_sync/interfaces.py
```diff
@@ -73,20 +74,19 @@ def sync_weights(
...
-        errors (matching the existing grpo.py behavior).
+        Step 4 (verification) is performed by every transport: each checks
+        its ``update_success`` signal and raises on failure.
 
         Args:
             timer: Optional Timer for profiling individual phases.
             kv_scales: Optional KV cache scales for FP8 quantization.
                 Honored by the IPC/ZMQ and NCCL collective transports. The
-                HTTP transport ignores this parameter.
+                SGLang transports ignore this parameter.
```
yuki-97: The docstring says "The SGLang transports ignore this parameter", but `_SGLangWeightSynchronizer._reject_kv_scales` (`sglang_weight_synchronizer.py:108`) raises when it is non-None — "ignore" implies silent drop, the code does the opposite.

**Suggest fix:** rewrite the sentence, e.g. "The SGLang transports do not support this parameter and raise if it is set."

nemo_rl/models/policy/utils.py
```diff
@@ -521,230 +521,520 @@ def rebuild_cuda_tensor_from_ipc(
...
+                    names=names,
+                    dtypes=dtypes,
+                    shapes=shapes,
+                    group_name=group_name,
+                    weight_version=str(weight_version),
+                )
+                for engine in rollout_engines
+            ]
+            handles = []
+            for i, (_, tensor) in enumerate(bucket):
```
yuki-97: In `broadcast_hf_buckets_via_distributed_impl`, `bucket_idx` (initialized at line 1003, incremented at 1008) is written but never read, and the `enumerate` indices in the inner loops here (`for i, (_, tensor)`) and at line 1036 (`for i, handle`) are unused.

**Suggest fix:** drop `bucket_idx` (three lines) and switch the two enumerates to plain `for` loops.

nemo_rl/algorithms/ppo.py
```diff
@@ -655,10 +656,22 @@ def initialize_generation_with_policy(
     # during setup to free GPU for value model initialization).
     policy.prepare_for_training()
 
-    # prepare refit info
-    state_dict_info = policy.prepare_refit_info()
-    if policy_generation is not None:
-        policy_generation.prepare_refit_info(state_dict_info)
+    if isinstance(policy_generation, SGLangGeneration):
```
yuki-97: I think we don't need to update ppo in this PR, we can just focus on grpo, so we can revert the changes of this file, wdyt?

we can do the support of ppo after https://github.com/NVIDIA-NeMo/RL/pull/3262 in, cc @bg51717

examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-2n8g-fsdp2tp1-sglang.yaml
yuki-97, line 1: could you help to add a nightly test based on this one, but change to use mcore + non-colcated? and also share the result curve in the PR description so that others can take as a reference.

examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-1n8g-fsdp2tp1-sglang.yaml
```diff
@@ -29,7 +29,7 @@ policy:
       max_running_requests: null
       mem_fraction_static: 0.6
       skip_server_warmup: true
-      disable_piecewise_cuda_graph: true
+      cuda_graph_backend_prefill: breakable
```
yuki-97, line 32: curious what's this change for and why we haven't it before?

nemo_rl/algorithms/grpo.py
```diff
@@ -1392,6 +1391,21 @@ def init_trtllm():
             )  # type: ignore
             ray.get(futures_train + futures_inference)
         worker_init_timing_metrics["collective_init_time_s"] = time.perf_counter() - t0
+
+    if backend == "sglang":
```
yuki-97: can we move this check to under https://github.com/xiuhu17/RL/blob/3e010c04f0419137627f927cbe08159abbfffbb6/nemo_rl/algorithms/grpo.py#L1262?

nemo_rl/algorithms/grpo.py
```diff
@@ -1431,6 +1445,21 @@ def init_trtllm():
             f"Using checkpoint-engine refit backend: {checkpoint_engine_config['backend']}",
             flush=True,
         )
+    elif backend == "sglang":
+        t0 = time.perf_counter()
+        policy_generation.weight_synchronizer = create_weight_synchronizer(
+            policy=policy,
+            generation=policy_generation,
+            generation_backend=backend,
+            colocated=colocated_inference,
```
yuki-97, line 1452: why we also need `create_weight_synchronizer` for SGL colocated case? I remember we didn't do this before and colocated could work in your previous PR.

nemo_rl/weight_sync/factory.py
```diff
@@ -95,9 +101,22 @@ def create_weight_synchronizer(
...
+            # lazily on the first refit, so it needs neither cluster handle.
+            from nemo_rl.weight_sync.sglang_weight_synchronizer import (
+                SGLangDisaggregatedWeightSynchronizer,
             )
+
+            return SGLangDisaggregatedWeightSynchronizer(
+                policy=policy,
+                generation=generation,
+                refit_buffer_size_gb=refit_buffer_size_gb,
+            )
```
yuki-97: here is the same with the colocated case of sglang below.

could you move them to above `if not colocated:` then colocated / non-colocated sglang can share one.

nemo_rl/algorithms/grpo.py
```diff
@@ -2260,7 +2301,6 @@ def refit_policy_generation(
         # update weights
         update_success = False
```
yuki-97, line 2298: let's keep the comment

nemo_rl/models/generation/sglang/sglang_generation.py
```diff
@@ -228,6 +235,10 @@ def _start_engines(
             # Explicitly pass CUDA_VISIBLE_DEVICES through to the engine actor so
             # all engines see the same global value (Ray would otherwise remap it
             # because we set the NOSET_* flags above).
+            # Trainer and engine must agree on the NCCL transport; sglang's
+            # scheduler subprocess defaults to NCCL_CUMEM_ENABLE=0.
+            env_vars["NCCL_CUMEM_ENABLE"] = "0"
+
```
yuki-97, line 232: let's move it upper or lower, otherwise the `CUDA_VISIBLE_DEVICES` notes seems not match the place.

nemo_rl/models/generation/sglang/sglang_worker.py
```diff
@@ -304,6 +304,87 @@ def resume_memory_occupation(self, tags: list[str] | None = None):
     def check_weights(self, action: str):
         return self._make_request("weights_checker", {"action": action})
 
+    def get_weight_version(self):
```
yuki-97: will does this function be used? if not can we remove it from this PR and add it back when needed so that it won't become a dead code.

nemo_rl/models/megatron/setup.py
```diff
@@ -292,8 +292,16 @@ def destroy_parallel_state():
...
-def setup_distributed() -> None:
+def setup_distributed(config: Optional[PolicyConfig] = None) -> None:
     """Handle NCCL settings, dtype mapping, and basic config setup."""
+    if (
+        config is not None
+        and "generation" in config
+        and config["generation"] is not None
+        and config["generation"].get("backend") == "sglang"
+    ):
+        os.environ["NCCL_CUMEM_ENABLE"] = "0"
```
yuki-97: looks you will set it in `validate_and_set_config` which is a better place to set, why we need set it here again?

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuki-97, line 1: looks after this PR sglang will support dtensorv2+mcore, can we add some assert for dtensorv1 + sglang in setup?

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -2135,6 +2165,162 @@ def _expert_idx(megatron_name: str) -> str:
...
+
+        return MegatronSGLangHfWeightIterator(
+            megatron_bridge=self.megatron_bridge,
+            models=[self.model],
+            conversion_tasks=self.refit_conversion_tasks,
+        )
+
+    @torch.no_grad()
+    @wrap_with_nvtx_name("megatron_policy_worker/connect_sglang_rollout_engines")
+    def connect_sglang_rollout_engines(
```
yuki-97: does this function the same with the dtensor v2 one? also are there other same functions?

can we move these to a shared place `nemo_rl/models/policy/workers/base_policy_worker.py`? so that once we need update the functions, we won't miss to update another side.

nemo_rl/models/policy/workers/megatron_sglang_weight_iterator.py
```diff
@@ -0,0 +1,92 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Any, Iterator
+
+import torch
+
+
+class MegatronSGLangHfWeightIterator:
```
yuki-97, line 28: how is this different from `_iter_params_with_optional_kv_scales` in `nemo_rl/models/policy/workers/megatron_policy_worker.py` and can we reuse that and drop the file?

since they are both transfer from mcore to hf using `bridge.export_hf_weights`, I don't think there will be huge difference.
if there's some feature there you don't need (draft model, kv_scales), you can just skip or add a sglang path there.

nemo_rl/models/policy/utils.py
```diff
@@ -521,230 +521,520 @@ def rebuild_cuda_tensor_from_ipc(
...
+    finally:
+        if saved_bound_device_id is not None:
+            default_pg.bound_device_id = saved_bound_device_id
+
+    # Map identity ranks so collective ops can resolve member ranks for ``pg``.
+    _world.pg_group_ranks[pg] = {i: i for i in range(world_size)}
+    return pg
+
+
+def connect_rollout_engines_from_distributed(
```
yuki-97, line 880: is it possible to reuse `from nemo_rl.distributed.stateless_process_group import StatelessProcessGroup` or just reuse `init_collective` in `nemo_rl/models/policy/workers/base_policy_worker.py`?

vllm and trtllm are both using that, so curious why sglang don't use that?

nemo_rl/models/policy/utils.py
```diff
@@ -521,230 +521,520 @@ def rebuild_cuda_tensor_from_ipc(
...
+        dist.destroy_process_group(model_update_group)
+    except Exception:
+        pass
+    try:
+        ray.get(refs)
+    except Exception:
+        pass
+
+
+def get_sglang_quantization_cfg(policy_generation: Any) -> dict:
```
yuki-97: I think don't need a function for this, we can just use `policy_generation.sglang_cfg["sglang_cfg"]["quantization"]` at the calling point.

nemo_rl/models/policy/utils.py
```diff
@@ -521,230 +521,520 @@ def rebuild_cuda_tensor_from_ipc(
...
+    """Read the active SGLang quantization block from the generation handle.
+
+    Returns an empty dict when no quantization config is set, so callers can
+    treat the result as a stable mapping without ``None`` checks.
+    """
+    return dict(policy_generation.sglang_cfg["sglang_cfg"].get("quantization") or {})
+
+
+def fetch_updatable_engines_with_recover(policy_generation: Any) -> tuple:
```
yuki-97: I tend to think let's remove fault tolerance related code in this PR and add it to #3187 after this PR merged first.

nemo_rl/models/policy/utils.py
```diff
@@ -521,230 +521,520 @@ def rebuild_cuda_tensor_from_ipc(
...
+    """
+    use_ft = bool(
+        policy_generation.sglang_cfg["sglang_cfg"].get("use_fault_tolerance", False)
+    )
+    if use_ft:
+        policy_generation.recover_updatable_engines()
+    return policy_generation.get_updatable_engines_and_lock()
+
+
+def broadcast_hf_buckets_via_distributed_impl(
```
yuki-97, line 960: out of curious how's this function different from the one vllm/trtllm using and is it possible to reuse?

nemo_rl/weight_sync/dtensor_sglang_refit.py
```diff
@@ -0,0 +1,102 @@
...
+        ray.get(futures)
+    finally:
+        # Close the session and resume on every path, so a failed refit
+        # leaves the engine usable instead of wedged in the update state.
+        policy_generation.end_weight_update()
+        policy_generation.continue_generation()
+    return True
+
+
+def refit_sglang_distributed(
```
yuki-97: I think this function and the colocated one from mcore are not different? since we both call self._policy to get the lm_policy then doing things.

the only difference is dtensor not support `refit_sglang_distributed`, but we can just assert it at the beginning.

so I feel we can remove this file and don't need to distinguish between dtensor or mocre.

nemo_rl/weight_sync/sglang_weight_synchronizer.py
```diff
@@ -0,0 +1,209 @@
...
+            sync_succeeded = True
+        finally:
+            self._policy.offload_after_refit()
+            self._generation.prepare_for_generation(tags=["kv_cache"])
+
+        self._stale = not sync_succeeded
+        return None
+
+
+class SGLangDisaggregatedWeightSynchronizer(_SGLangWeightSynchronizer):
```
yuki-97, line 254: I think `refit_sglang_distributed` can be a member function of this class, also for another function and the class. then we can remove the two `xxx_sglang_refit.py` file

yuki-97 (commented):
thanks for supporting this @xiuhu17 , I took a first round review and will review again after your updates.

Two scope notes before the per-line comments:

1. **Skipping `b09309918` per the stack note** — that commit duplicates #3187 (fault tolerance); everything below reviews `26e1d60b..HEAD`.

2. **`fetch_updatable_engines_with_recover` should move to #3187.** `nemo_rl/models/policy/utils.py:959` is fault-tolerance glue — it reads `sglang_cfg.use_fault_tolerance` and gates the call to `recover_updatable_engines`. Its natural home is #3187; the three refit-driver callers (`policy/utils.py`, `dtensor_sglang_refit.py`, `megatron_sglang_refit.py`) can call `get_updatable_engines_and_lock` directly here, and the recover gate lives with the FT feature it belongs to.

Other refit-side uses of FT-provided primitives (`pause_generation`, `num_new_engines`, `rollout_engine_lock` acquire/release, `clear_updatable_num_new_engines`) are legitimate consumers of the FT interface and stay in this PR.

nemo_rl/models/policy/utils.py
xiuhu17: fixed

nemo_rl/models/policy/workers/megatron_policy_worker.py
xiuhu17: _build_sglang_hf_iterator now reads and check with target_precision. Now it only accept bf16, and in later #3188, it could accept mxfp8 precision

examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-2n8g-fsdp2tp1-sglang.yaml
xiuhu17: Just added: vllm+fsdp vs sglang+megatron on 4xb200, colocated, qwen2.5

Screenshot 2026-08-03 at 5 36 30 PM

https://wandb.ai/xiuhu17-bytedance/nemo-rl-sglang-1n4g?nw=nwuserxiuhu17

nemo_rl/models/generation/sglang/config.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/weight_sync/interfaces.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/weight_sync/dtensor_sglang_refit.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/weight_sync/sglang_weight_synchronizer.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/weight_sync/megatron_sglang_refit.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/weight_sync/factory.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/algorithms/grpo.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/algorithms/grpo.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/algorithms/ppo.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/models/generation/sglang/sglang_generation.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/models/policy/utils.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/models/policy/utils.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/models/policy/workers/megatron_policy_worker.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

nemo_rl/models/generation/sglang/sglang_worker.py
yuki-97: fixed in 9194ded10539592b37654d4e0f9614951f929168

examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-2n8g-fsdp2tp1-sglang.yaml
yuki-97: nightly registered in `nightly.txt:16` and `nightly_mcore.txt:29`. Could you also paste the convergence curve (reward + loss + `token_mult_prob_error`) in the PR description?

examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-1n4g-fsdp2tp1-sglang.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+data:
+  max_input_seq_length: 512
+
+logger:
+  log_dir: logs/grpo-qwen2.5-math-1.5b-instruct-1n4g-fsdp2tp1-sglang
+  wandb_enabled: true
+  tensorboard_enabled: true
+  wandb:
+    project: nemo-rl-sglang-1n4g
```
yuki-97, line 58: `project: nemo-rl-sglang-1n4g` (and `name` below) are shadowed at runtime by the `.sh` CLI overrides `logger.wandb.project=nemo-rl` / `logger.wandb.name=$EXP_NAME` — Hydra CLI overrides beat YAML.

Sibling `1n8g-fsdp2tp1-sglang.yaml` uses `project: nemo-rl` to match its `.sh`.

```suggestion
    project: nemo-rl
```

tests/test_suites/llm/grpo-qwen2.5-math-1.5b-instruct-1n4g-fsdp2tp1-sglang.sh
yuki-97, line 1: thanks for adding the two 1n4g tests, will also need to add them to `tests/test_suites/nightly_gb200.txt`.

yuki-97 (commented):
thanks for the updates! some comments on the new added tests, and you'll need a rebase since there's some conflict with main branch.

nemo_rl/algorithms/grpo.py
xiuhu17: After the 5f5438b95d701234648b1b92e58968ac1f58a6b0, we use weight synchronizer for sglang colocated and sglang disaggregated

nemo_rl/models/megatron/setup.py
xiuhu17: I just moved all settings into `setup_distributed`, therefore `validate_and_set_config` will not set anymore.

nemo_rl/models/policy/workers/megatron_policy_worker.py
xiuhu17: connect_sglang_rollout_engines now shared between fsdp and megatron 

Since we have not supported fsdp with non colocate mode, therefore there are no others share the same.

nemo_rl/models/policy/utils.py
yuki-97: fixed in d33bb462fd72616a41cf2dc0942a1a6ef6590b04

nemo_rl/models/policy/workers/megatron_policy_worker.py
yuki-97: fixed in 1686ab1ddca93f919aaa5982c44ee61429003dd6

nemo_rl/models/policy/utils.py
yuki-97: fixed in 1686ab1ddca93f919aaa5982c44ee61429003dd6

nemo_rl/models/policy/workers/megatron_policy_worker.py
yuki-97: fixed in 1686ab1ddca93f919aaa5982c44ee61429003dd6

nemo_rl/models/megatron/setup.py
```diff
@@ -294,8 +292,18 @@ def destroy_parallel_state():
...
     """Handle NCCL settings, dtype mapping, and basic config setup."""
+    generation_cfg = (config or {}).get("generation") or {}
+    rollout_backend = generation_cfg.get("backend")
+    is_generation_colocated = (generation_cfg.get("colocated") or {}).get("enabled")
+
+    # SGLang's scheduler subprocess defaults to NCCL_CUMEM_ENABLE=0 and the
+    # trainer has to agree with it, so this one has to land before the process
+    # group below picks a transport.
+    if rollout_backend == "sglang":
+        os.environ["NCCL_CUMEM_ENABLE"] = "0"
```
yuki-97, line 305: for other rollout backends which need to set NCCL_CUMEM_ENABLE=1 for non-colocatred refit, you have the comment "Deliberately after init_process_group: it targets the refit communicator, not the training process group."
why sglang need to put it before init_process_group? isn't sglang also only target the refit communicator?

if sglang should also do it after, I think let's still put it at `validate_and_set_config` since we set config there.

nemo_rl/algorithms/grpo.py
yuki-97: I see. currently seems other rollout backends only go here in disaggregated case so just curious with this.

I'm ok with this as is, cc @terrykong @saumishr could you help take a look on whether the implementation match your new `weight_synchronizer` design?

nemo_rl/models/policy/workers/megatron_sglang_weight_iterator.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
yuki-97, line 1: nit: copyright header year lags. This is a new file added by this PR, and per `.claude/skills/copyright/SKILL.md` (always use the current year) it should say 2026 — the sibling new file `nemo_rl/weight_sync/sglang_weight_synchronizer.py` in this same PR already uses 2026.

```suggestion
# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```

examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-1n4g-megatrontp1-sglang.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+data:
+  max_input_seq_length: 512
+
+logger:
+  log_dir: logs/grpo-qwen2.5-math-1.5b-instruct-1n4g-megatrontp1-sglang
+  wandb_enabled: true
+  tensorboard_enabled: true
+  wandb:
+    project: nemo-rl-sglang-1n4g
```
yuki-97, line 67: Same shadow issue as [the fsdp2 sibling](https://github.com/NVIDIA-NeMo/RL/pull/3190#discussion_r3713052524): the launcher `.sh` overrides `logger.wandb.project=nemo-rl` on the CLI, so this YAML value never takes effect. The other sglang recipes (1n8g/2n8g fsdp2, 2n8g megatron-noncolocated) all use `nemo-rl`.

```suggestion
    project: nemo-rl
```

tests/test_suites/llm/grpo-qwen2.5-math-1.5b-instruct-1n4g-megatrontp1-sglang.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
yuki-97, line 1: For the megatron variant, also register in `tests/test_suites/nightly_mcore_gb200.txt` — mirrors how `grpo-qwen2.5-math-1.5b-instruct-2n8g-megatrontp1-sglang-noncolocated.sh` lands in both `nightly.txt` and `nightly_mcore.txt` in this PR.

yuki-97 (commented):
Follow-up on post-review commits (866087916 build + eac58ed3c 1n4g recipes): 3 nits — copyright year on new .py, wandb project shadow on the mcore sibling of an already-flagged yaml, and mcore nightly manifest registration for the mcore 1n4g sh.

PR #2980, opened 2026-06-28
refactor(ppo): reuse LossPostProcessor in value worker + add offload_to_cpu

bg51717:
# What does this PR do ?

1. **`policy.offload_to_cpu()`** — a clearer name for `offload_after_refit` at the PPO call sites that offload *outside* the weight-refit lifecycle (e.g. freeing GPU after policy init so the value model can initialize). Added as a concrete method on `ColocatablePolicyInterface` that delegates to `offload_after_refit`, and adopted in `ppo.py`. `offload_after_refit` is kept for the weight-sync refit path and the other algorithms.

2. **Reuse `LossPostProcessor` in the Megatron value worker** — the value worker had a bespoke `forward_step_value` that duplicated the CP / num-microbatches loss scaling already in the policy's `LossPostProcessor`. The value train step now runs through `megatron_forward_backward(post_processing_fn=LossPostProcessor(...))` like the policy, with a `prepare_fn` hook for the value-specific scalar-head right-shift + CP all-gather. `forward_step_value` is deleted, so the scaling lives in exactly one place. Mirrors how the DTensor value worker reuses its `LossPostProcessor`.

Both are behaviour-preserving refactors (no convergence impact — see Test).

# Issues
Closes #2785.

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [x] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Test
Tests run (all pass):

- **Unit** (`tests/unit/models/value/test_megatron_value_worker.py`, mcore, 2 GPU): **21 passed** — `get_values` / `train` / checkpoint round-trip across DP / TP / SP / dynamic batching / sequence packing / PP2 / CP2, SP & CP numerical-equivalence parity, plus a new CPU unit test for the value `prepare_fn` (right-shift + truncate). `tests/unit/algorithms/test_ppo.py`: **20 passed**.
[...]

PR #3195, opened 2026-07-14
cp: cherry-pick #3171 - fix(ppo): restore optimizer state on checkpoint resume

yuki-97:
manually cp #3171 since conflict of `offload_after_refit` -> `offload_to_cpu` in `ppo.py` with #2980 which is not in r0.7.0 branch.

### DIFF
--- a/nemo_rl/algorithms/ppo.py
+++ b/nemo_rl/algorithms/ppo.py
@@ -15,7 +15,6 @@ import gc
 import os
 import time
 import warnings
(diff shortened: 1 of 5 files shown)


bg51717 (approved):
LGTM. I confirmed that retaining `offload_after_refit()` is the correct adaptation for `r0.7.0`, where #2980 is not present. The rest of the backport matches #3171.
Thanks for your work @yuki-97.

PR #3197, opened 2026-07-14
refactor(policy): migrate common config leaves to BaseModel

NolenLiang:
## Summary

  This is the first PR in the policy config migration series.

  - Migrate common policy config leaves from `TypedDict` to Pydantic `BaseModel` with `extra="allow"`:
    - sequence packing
    - dynamic batching
    - tokenizer
    - reward model
    - draft model
    - router replay
    - PyTorch optimizer and scheduler
  - Update Policy, Value, DTensor, Automodel, Megatron, TQ, and reward-model environment call sites to
  use attribute access.
  - Update example configs, reference configs, and test fixtures.
  - Support both direct and sequential PyTorch scheduler configurations.

  ## Compatibility

  This PR intentionally does not migrate top-level `PolicyConfig` / `ValueConfig` or backend-specific DTensor, Automodel, and Megatron configs. Those will be handled in follow-up PRs.

  ## Validation

  - `tests/unit/test_config_v2.py` and config validation: 509 passed, 6 skipped
  - Targeted MCore tests: 30 passed
  - Targeted Automodel tests: 44 passed
  - Common policy leaf tests: 12 passed
  - Teacher, DPO, and RM regression tests: 14 passed
  - Ruff format and lint checks passed
  - Python compilation and `git diff --check` passed


nemo_rl/models/policy/__init__.py
```diff
@@ -443,72 +446,71 @@ class MegatronConfig(TypedDict):
...
     name: str
     kwargs: dict[str, Any]
 
 
-class SinglePytorchMilestonesConfig(TypedDict):
+class SinglePytorchMilestonesConfig(BaseModel, extra="allow"):
     milestones: list[int]  # Used in SequentialLR configuration
 
 
 SchedulerMilestones = dict[str, list[int]]
```
jinglinglingling, line 528: PolicyConfig.scheduler still includes SchedulerMilestones, but runtime parsing paths (automodel / dtensor workers) and validation tests now only cover SinglePytorchSchedulerConfig and sequential list[...].
Could we either remove SchedulerMilestones from the union or add explicit normalization for that shape to keep schema and runtime behavior aligned?

jinglinglingling (changes requested):

nemo_rl/models/policy/__init__.py
NolenLiang: Good catch. I removed the standalone SchedulerMilestones shape from the scheduler union while keeping the legacy symbol importable, introduced a shared PytorchSchedulerConfig used by both PolicyConfig and ValueConfig, and added regression coverage that accepts the single-scheduler and valid SequentialLR-list forms while rejecting a bare {"milestones": [5]} config. The targeted config/scheduler/packing suite passes (62 tests) on 9779ad5c8.

jinglinglingling (approved):

PR #3201, opened 2026-07-14
fix: honor NeMo Gym's mask_sample in the Gym rollout path

linj-glitch:
# What does this PR do ?

**Honors NeMo Gym's `mask_sample` signal in the Gym rollout path so rollouts with unreliable rewards are excluded from the gradient instead of trained on (or crashing the step).**

Gym agent servers (e.g. `swe_agents`, `anyterminal_agent`) set `mask_sample: true` on the verify response when a rollout's reward is unreliable — agent/eval timeout, OOM-killed container, failed token-trace capture — and document the contract as "downstream RL drops the gradient for this rollout." Nothing in NeMo RL consumed it, so today:

- masked rollouts **train anyway** with garbage rewards contaminating GRPO group advantages — the worst sub-case being `reward=1.0` accidental resolves (patch passed but the agent died on max-iteration/context-window) reinforced as positives;
- a masked rollout carrying **no token-annotated generations** (e.g. a failed trace capture on the Gym↔Switchyard path) raises `ValueError` in `_postprocess_nemo_gym_to_nemo_rl_result` and fails the whole rollout batch.

Changes (both mirror existing conventions):

- `run_async_nemo_gym_rollout` zeroes the sample's `loss_multiplier` when `full_result["mask_sample"]` is truthy — the same mechanism used for truncated samples (`loss_multiplier[truncated] = 0`).
- The converter emits a prompt-only message log for **masked** rollouts with no generations, so tensorization proceeds and the zeroed loss multiplier excludes the sample; **unmasked** empty rollouts still raise the existing actionable `ValueError`.

Producer side: NeMo Gym promotes `mask_sample` to a top-level `BaseVerifyResponse` field (NVIDIA-NeMo/Gym#2026), so it also lands in the per-agent auto-aggregated metrics (`rollouts.py` scalar aggregation) — masked-sample rates become visible in training metrics for free.

# Issues

None filed; found while validating the Gym↔Switchyard token-capture integration end-to-end (NVIDIA-NeMo/Gym#2026).

# Usage

No API change. Any Gym rollout whose verify response carries `mask_sample: true` is now excluded from the gradient:

```python
# Gym agent server (producer side)
return MyVerifyResponse(..., reward=reward, mask_sample=True)  # unreliable reward
# → NeMo RL zeroes that sample's loss_multiplier; no training signal from it.
```

**Pre checks**:
- [x] Did you write any new necessary tests? (`test_nemo_gym_postprocess_masked_rollout_without_generations`, `test_nemo_gym_postprocess_unmasked_rollout_without_generations_raises`)
- [x] Did you run the unit tests and functional tests locally? (converter paths + loss-multiplier zeroing exercised directly on CPU; `ruff@0.9.9 check`/`format` clean)

# Additional Information

The end-to-end verification that surfaced this: Switchyard token capture → Gym rollout reconstruction → this converter → GRPO tensors, with the masked-sample and non-contiguous-token failure modes reproduced against real code.

### DIFF
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -400,17 +400,36 @@ Output prompt token IDs: {output_item_dict["prompt_token_ids"]}
 
         if not nemo_rl_message_log:
             input_messages = nemo_gym_result["responses_create_params"]["input"]
-            prompt_token_ids = tokenizer.apply_chat_template(
-                input_messages, tokenize=True
-            )
-            raise ValueError(
-                f"NeMo Gym returned a result with no generation data. "
-                f"This typically means the prompt for the first turn already exceeds the vLLM max_model_len, "
-                f"so vLLM rejected the request before any tokens could be generated.\n"
-                f"  Prompt length: {len(prompt_token_ids)} tokens.\n"
-                f"  → Fix: increase `policy.max_total_sequence_length` and `policy.generation.vllm_cfg.max_model_len` "
-                f"to a value larger than {len(prompt_token_ids)}."
-            )
+            if nemo_gym_result.get("mask_sample"):
+                # A masked rollout (unreliable reward: agent/eval timeout, OOM
+                # kill, failed trace capture) may legitimately carry no
+                # token-annotated generations. Its loss multiplier is zeroed
+                # downstream, so emit a prompt-only message log to keep
+                # tensorization working instead of failing the batch.
+                prompt_token_ids = (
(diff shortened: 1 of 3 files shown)


terrykong:
@ananthsub @bxyu-nvidia to review

jepio:
@linj-glitch Gym's `mask_sample` is not granular enough to always honor, for example if rollouts hit max iteration but solves the task, Gym sets "mask_sample=True" but that is a trainable sample and we want to train our workload on such samples.

I would suggest adding a config setting that let's users decide whether to honor `mask_sample` or not.

PR #3203, opened 2026-07-14
feat(sc): add native and NeMo-Gym validation

haitian-nvidia:
# What does this PR do ?

Adds initial, periodic, and final validation to SingleController for both native environments and NeMo-Gym.

- Pauses rollout admission and drains already-dispatched rollouts before validation weight sync, while preserving in-flight weight updates on non-validation steps.
- Uses a dedicated validation dataloader and rollout manager to report accuracy, average response length, sample count, JSONL outputs, and current-step checkpoint metrics.
- Reuses the training NeMo-Gym actor for validation and evaluates the full prepared Gym validation dataset when `max_val_samples` is unset.
- Keeps the rollout gate closed when drain, sync, or validation fails.

# Additional Information

NeMo-Gym validation does not yet support `env.nemo_gym.effort_levels`; setup raises an explicit `NotImplementedError` when it is configured.


yuki-97:
@RayenTian could you help review? it's not target main branch so we can merge it if there's no major issues.

RayenTian (approved):

RayenTian:
LGTM! Approved!

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -444,19 +518,35 @@ def setup_single_controller(
         use_nemo_gym=use_nemo_gym,
         tq_buffer=tq_buffer,
     )
+    validation_rollout_manager: Optional[RolloutManager] = None
+    if validation_enabled:
+        validation_rollout_manager = RolloutManager(
+            tokenizer=tokenizer,
+            env_handles=val_env_handles,
```
yuki-97: can we merge `val_env_handles` into `env_handles` (just add those in `val_env_handles` but not in `env_handles`) and use one rollout manager?

haitian-nvidia: We should keep `val_env_handles` separate because these dictionaries are task-name-to-environment routing maps, not just collections of handles. The same task can use different environments for training and validation; merging only missing keys would silently retain the training environment and route validation to the wrong actor.

We cannot use one rollout manager because the generation for training and validation has different `num_generations_per_prompt`.

yuki-97 (approved):
thanks for the updates!

PR #14, opened 2025-03-21
fix: disable mixed precision training until #13 is resolved

parthchadha:
# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

# Changelog 
- Please update the [CHANGELOG.md](/CHANGELOG.md) under next version with high level changes in this PR.

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation? Make sure to also update the [NeMo Framework User Guide](https://docs.nvidia.com/nemo-framework/user-guide/latest/index.html) which contains the tutorials

# Checklist when contributing
- [ ] TBD

# Additional Information
* Related to # (issue)

PR #3208, opened 2026-07-15
feat(megatron): Zero train-gen mismatch training with megatron inference

YigongQin:
# What does this PR do ?

Add `policy.megatron_cfg.zero_train_gen_mismatch` — a single flag that pins Megatron train and generation onto the same batch-invariant kernel paths so training and generation achieve bitwise parity on the certified TP=1 beds.

Works for colocated `transformer_engine` generation and non-colocated `inference_optimized` generation. This PR already lands TOKEN_ROUNDER alignment from [#3848](https://github.com/NVIDIA-NeMo/RL/pull/3848) (packed totals in `data.py`, unpacked per-sequence pad in `lm_policy.py`) so TE native RMSNorm sees the same token-count class as generation CUDA-graph buckets.

**Limitation.** Train and gen logprobs are bitwise the same; they still call different log-softmax functions (train: `_compute_distributed_log_softmax_with_grad`; gen: fused `F.log_softmax`). Unifying that function — and TP>1 full-vocab gather for train scoring — is follow-up [#3848](https://github.com/NVIDIA-NeMo/RL/pull/3848). The TP>1 gather is not required for recipes in this PR (`tensor_model_parallel_size=1`).

**Validation module.** When the flag is set, `nemo_rl/models/megatron/zero_train_gen_mismatch.py` resolves recipe defaults, runs collect-all validation (one `ValueError` listing every violation), and optionally enables MCore batch-invariant kernels. `setup.enable_zero_train_gen_kl()` is a thin wrapper around `configure_zero_train_gen_mismatch()`; `enable_batch_invariant_mode()` delegates BI checks and kernel enable to that module.

**`batch_invariant.py`** (unchanged scope): only `batch_invariant_token_multiple()` for TOKEN_ROUNDER alignment in `data.py` / `lm_policy.py`. It is not part of zero-KL validation; keep it separate so policy/data code does not import the full zero-KL gate module.

# Issues

- https://github.com/NVIDIA-NeMo/RL/issues/3743 — zero train/gen mismatch

# Usage

**BF16** — `transformer_engine` or `inference_optimized`. Set the flag. For `inference_optimized`, put `transformer_impl` under `policy.generation.mcore_generation_config` (typically non-colocated):

```yaml
policy:
  megatron_cfg:
    zero_train_gen_mismatch: true   # the switch
  generation:
    backend: megatron
    mcore_generation_config:
      transformer_impl: inference_optimized   # or transformer_engine
```

Shipped BF16 infopt recipes:

- `examples/configs/recipes/llm/grpo-qwen3-1.7b-megatron-zero-train-gen-kl-infopt-noncolocated.yaml` (4 train TE + 4 dedicated infopt gen on one 8-GPU node)
- `examples/configs/recipes/llm/grpo-dapomath17k-qwen-30ba3b-megatron-zero-train-gen-kl-infopt-noncolocated.yaml` (2 nodes × 4 GPUs: train + dedicated infopt gen, BF16, `inference_grouped_gemm_backend: vllm`)

**MXFP8** — colocated `transformer_engine` (accuracy beds) or experimental non-colocated `inference_optimized` + native TE MoE MXFP8 (requires newer Megatron-Core; see support table).

Colocated TE + mxfp8 (FC1/FC2):

```yaml
policy:
  megatron_cfg:
    zero_train_gen_mismatch: true
    fp8_cfg:
      enabled: true
      fp8_recipe: mxfp8
      fp8_param: false   # optional; BF16 primary weights + mxfp8 compute
  generation:
    backend: megatron
    mcore_generation_config:
      transformer_impl: transformer_engine
```

Experimental infopt + TE grouped MXFP8 MoE: `grpo-dapomath17k-qwen-30ba3b-megatron-zero-train-gen-kl-mxfp8-te-infopt-noncolocated.yaml` + `examples/configs/te_precision/qwen30ba3b-moe-only-mxfp8.yaml`.

# Support status (`zero_train_gen_mismatch=true`)

Validation runs on the driver (config + BI parity, no GPU packages) and on workers (`check_packages` + `check_platform` when `apply_kernels=True`). All hard failures are collected into one error.

| Gate | Supported | Not supported / notes |
|------|-----------|------------------------|
| **(0) Backend** | `policy.generation.backend=megatron`; merged gen `transformer_impl` ∈ `{inference_optimized, transformer_engine}` | vLLM / other backends |
| | `transformer_engine` gen: **allowed**, **warn** to use `inference_optimized` for performance | |
| **(1) Platform** | Blackwell (SM100+); `flash_attention_version=4` (flash-attn-4 + CuteDSL via `[mcore]` extra) | Hopper (SM90) / FA3; pre-Blackwell GPUs |
| **(2) Packages** (workers) | MCore git checkout ≥ `MEGATRON_CORE_MIN_COMMIT_SHA` (`4e603390d…`); `transformer_engine>=2.18`; `flash_attn>=2.8.1`; `flash-attn-4>=4.0.0b20`; `nvidia-cutlass-dsl>=4.6.0.dev0` | MCore wheel-only installs without git HEAD; Hopper-only FA3 envs |
| **(3) Architecture** | Standard flash attention; dense + MoE linears (Qwen-style beds) | MLA (`multi_latent_attention`, `use_mla`); linear-attention / SSM hybrid keys (denylist) |
| **(4) Precision** | **BF16**: `policy.precision=bfloat16` and `fp8_cfg.enabled=false`. **MXFP8**: `fp8_cfg.enabled=true` and `fp8_recipe=mxfp8` (including MoE-only mxfp8 compute with `fp8_param=false`). Train and merged gen must classify the same way. | blockwise/other FP8 recipes; non-`bfloat16` policy precision without mxfp8 |
| **(5) Topology** | `tensor_model_parallel_size=1`, `context_parallel_size=1`; train/gen 

Constants and floors: `MEGATRON_CORE_MIN_COMMIT_SHA`, `TRANSFORMER_ENGINE_MIN_VERSION`, etc. in `zero_train_gen_mismatch.py` — bump when taking new MCore inference commits.

# Knobs

`configure_zero_train_gen_mismatch()` (via `enable_zero_train_gen_kl()` in `setup.py` when `zero_train_gen_mismatch=true`):

**Set defaults** (`resolve_zero_train_gen_mismatch`)

| Config path | Key | Value |
|-------------|-----|-------|
| `policy.megatron_cfg` | `batch_invariant_mode` | `true` |
| | `moe_permute_fusion` | `false` |
| | `attention_backend` | `flash` |
| | `flash_attention_version` | `4` |
| | `batch_invariant_backend` | `te_native` |
| | `batch_invariant_collective` | `ordered` |
| `policy.generation.mcore_generation_config` | `logprobs_mode` | `raw_logprobs` |
| | `enable_chunked_prefill` | `false` |

**Conflicts:** if yaml already sets any default key to a different value, NeMo-RL warns and overrides it.

**Validation** (`validate_zero_train_gen_mismatch` + `validate_batch_invariant_mode`):

- Gates in support table above
- `use_fused_linear_logprobs=false`
- Train and generation share the same BI fields (TP, CP, `batch_invariant_*`, `attention_backend`, `flash_attention_version`)
- `inference_optimized` gen requires `policy.precision=bfloat16`

**Not set by zero-KL:** colocated vs non-colocated; `transformer_impl` (`transformer_engine` or `inference_optimized`); `fp8_param` (recipe/YAML).

**Runtime** (worker init, `apply_kernels=True`): `enable_batch_invariant_kernels()` calls Megatron `assert_te_supports_batch_invariant_attention()` (TE ≥ 2.18) and MCore BIK enable; `_skip_megatron_moe_bi_fp8_assert()` in `setup.py` lets **training** workers use MoE+BI+mxfp8 without full `te_mxfp8_inference` on `TransformerConfig`, while **infopt** gen keeps upstream gates. `lm_policy.py` sets `CUBLASLT_WORKSPACE_SIZE=0` and `CUBLAS_WORKSPACE_CONFIG=:0:0`. TOKEN_ROUNDER padding is applied automatically for packed and unpacked scoring.

**Tests:** `tests/unit/models/megatron/test_zero_train_gen_mismatch.py` (validation/resolve); existing `test_megatron_setup.py` zero-KL and BI cases.

# Accuracy

Metric: W&B `train/gen_kl_error` = 0 (bitwise). Train and gen still use different log-softmax functions; that is a code-path follow-up, not a numerical gap.

### Qwen2.5-1.5B (dense, colocated TE + mxfp8, packing on)

Config: `examples/configs/recipes/llm/grpo-qwen1.5b-megatron-zero-train-gen-kl-mxfp8.yaml`

Train/gen mismatch — Qwen2.5-1.5B

### Qwen3-30B-A3B (MoE, colocated TE + mxfp8, packing off)

Config: `examples/configs/recipes/llm/grpo-dapomath17k-qwen-30ba3b-megatron-zero-train-gen-kl-mxfp8.yaml`

Train/gen mismatch — Qwen3-30B-A3B

This recipe also enables long CoT for the DAPO math verifier, to reproduce convergence in https://humansand.ai/blog/nvfp4-rl.

Non-colocated infopt 30B (`grpo-dapomath17k-qwen-30ba3b-megatron-zero-train-gen-kl-infopt-noncolocated.yaml`) and TE MXFP8 infopt (`…-mxfp8-te-infopt-noncolocated.yaml`) are wired and validated by `zero_train_gen_mismatch.py`; bitwise KL figures for those beds are follow-up once generation paths are stable.

# Performance

No tok/s table in this PR. Zero-KL does **not** require graphs off: TOKEN_ROUNDER pads train scoring to the same token-count class as generation CUDA-graph buckets, so `train/gen_kl_error` stays bitwise 0 with graphs. Train stays `policy.megatron_cfg.cuda_graph_impl: none` in all shipped recipes.

| Recipe | Generation | KV |
|--------|------------|----|
| Qwen2.5-1.5B mxfp8 (colocated TE) | inherits `cuda_graph_impl: local` / `scope: block`; colocated managers attached at worker init (`_setup_colocated_cuda_graph_managers`) | `persist` |
| Qwen3-1.7B infopt (non-colocated) | same parent `local` / `block` on dedicated gen ranks | `persist` |
| Qwen3-30B-A3B infopt (non-colocated) | explicit `local` / `block`, `use_cuda_graphs_for_non_decode_steps: true` | `persist` |
| Qwen3-30B-A3B mxfp8 (colocated TE, accuracy figure) | explicit `cuda_graph_impl: none` | `recompute` |

The 30B mxfp8 accuracy bed is the eager/`recompute` exception (KV rebuilt after each refit). Infopt and 1.5B keep `persist` buffers and captured graphs across weight refit on generation.

**Pre checks**:

# Additional Information


shanmugamr1992:
@YigongQin  can you elaborate more on this MR ? maybe attach some sample runs that you did , how the performance changes etc.

rybakov:
zero mismatch between training and rollouts is important feature for RL debugging, please submit for review

nemo_rl/models/megatron/moe_routing_record.py
```diff
@@ -0,0 +1,96 @@
...
+            "routing_indices must be present for every sample when router replay is enabled"
+        )
+    aligned = [
+        align_routing_rows_to_token_count(
+            coerce_routing_to_3d(r), int(seq_lengths[i].item())
+        )
+        for i, r in enumerate(routing_per_sample)
+    ]
+    num_layers, topk = aligned[0].shape[1], aligned[0].shape[2]
+    out = torch.zeros(
```
ZhiyuLi-Nvidia: padding zeros is wrong

 moe_routing_record.py:88 pads routed_experts with torch.zeros. The repo's own R3 validator (router_replay.py:308-320) rejects exactly this, verbatim: "padded routed_experts rows must use a valid dummy top-k route, not repeated zeros." The vLLM producer pads with arange(topk) + the -1 sentinel (vllm/utils.py:218-234). every pad token replays top-k → expert 0.

nemo_rl/models/megatron/moe_routing_record.py
```diff
@@ -0,0 +1,96 @@
...
+        len(aligned), seq_dim, num_layers, topk, dtype=aligned[0].dtype
```
ZhiyuLi-Nvidia: Question can we double check if we respect `resolve_routed_experts_dtype(...)` at vllm_worker.py:493.
dtype=aligned[0].dtype just passes through whatever the engine handed us. 

```
def resolve_routed_experts_dtype(num_experts: Optional[int]) -> torch.dtype:
    """Return the narrowest signed dtype that fits expert ids and the -1 sentinel."""
    if num_experts is None:
        return ROUTED_EXPERTS_FALLBACK_DTYPE
    if num_experts - 1 <= torch.iinfo(torch.int8).max:
        return torch.int8
    if num_experts - 1 <= torch.iinfo(torch.int16).max:
        return torch.int16
    return torch.int32
```

This function is dtype aware pretty important to ensure ckpt and communication efficiency.

nemo_rl/models/policy/workers/moe_zero_kl_patches.py
```diff
@@ -0,0 +1,1012 @@
...
+        "(gather combine), and rejected fused MoE paths."
+    )
+
+
+# ---------------------------------------------------------------------------
+# Router replay inference
+# ---------------------------------------------------------------------------
+
+
+def _nrl_dynamic_step_context_bookkeeping(self: "TextGenerationController") -> Dict[str, Any]:
```
ZhiyuLi-Nvidia: Hi, is there a reason to override the whole function with the customized `_nrl_dynamic_step_context_bookkeeping`? And possible to minimize the gap if possible?

Like only touch part of the function `DynamicInferenceContext.update_requests` or reuse the function and do some post processing?

```
    _DYNAMIC_STEP_BOOKKEEPING_ORIG = TextGenerationController._dynamic_step_context_bookkeeping
    TextGenerationController._dynamic_step_context_bookkeeping = (
        _nrl_dynamic_step_context_bookkeeping
    )
```

From AI:

Finding: this forks routing reconstruction that Megatron already performs, disables the original, and the copy dropped one of upstream's guards.

Risk: upstream only reconstructs when len(record.requests) == 1. A >1 record means the request was preempted and resumed, so its KV blocks cover only the last stretch. Without that check, partial routing gets replayed as if it were the whole sequence — wrong replayed experts on those samples, i.e. exactly the bit-exactness this PR exists to deliver. Two smaller drifts compound it: the gate is moe_enable_routing_replay instead of kv_block_allocator.block_routing, and _apply_stop_word_finished_ids is inlined rather than called, so upstream fixes there never reach us.

Evidence:
- Upstream reconstruction: dynamic_engine.py:1414-1428 — same reconstruct_routing_from_blocks call, same request.routing_indices assignment
- _nrl_async_bookkeep pops finished_routing_block_ids, so that upstream path never runs
- Guard: dynamic_engine.py:1419; DynamicInferenceRequestRecord is documented as "history of DynamicInferenceRequest objects over multiple request checkpoints" (inference_request.py:616-618)
- House style: _install_missing_route_fallback_patch (router_replay.py:255-361) wraps, delegates, and raises if the upstream signature moves; mamba_zero_kl_patches.py::_validate_method_signature does the same. This patch does neither.

Recommendation: wrap DynamicInferenceContext.update_requests instead of forking this function. It's the call that frees the blocks, and its active_requests_mask already identifies finishing requests (mask == 0) — reconstruct before delegating, stash on the context, apply from the existing async_bookkeep wrapper. ~15 lines instead of 70, nothing popped, upstream keeps owning the mask/stop-word logic. If the fork stays: carry the == 1 guard, restore the upstream gate, call the stop-word helper, and name the upstream commit it's based on.

Question: do these recipes actually produce preempted/resumed requests? If not the guard is moot today, but it's a live trap for anyone who enables pausing later.

nemo_rl/models/policy/workers/moe_zero_kl_patches.py
```diff
@@ -0,0 +1,1012 @@
...
+) -> Tuple[list, list[int]]:
+    """Keep only 64-token-multiple CUDA graph decode buckets."""
+    from megatron.core.inference.batch_dimensions_utils import InferenceBatchDimensions
+
+    if not ladder:
+        return [], []
+
+    seen: set[int] = set()
+    new_list: list[InferenceBatchDimensions] = []
+    cap = min(max_tokens, (max_requests // 64) * 64)
```
ZhiyuLi-Nvidia: Hi here's a cap with magic number 64, why do we need this one?

Some feedbacks from AI:

Question first: where does 64 come from?

It's hardcoded six times in this function with no comment. Nothing in the config, the docstring, or the module header says why 64 rather than 32 or 128, or what breaks if it's wrong. Megatron does the same job with a named constant and an explanation — CUDA_GRAPH_ROUNDER = 2, commented "Constant for rounding token counts when generating CUDA graph batch dimensions" (batch_dimensions_utils.py:215-216). Please name it and say what it's tracking. If it's a GEMM tile size or a kernel-selection boundary, that's exactly the kind of fact a future reader can't recover.

One concrete thing the number ignores: upstream deliberately aligns every bucket to tp_size (round_up_to_nearest_multiple(cuda_graph_step_size, tp_size), and again at line 306). This function re-rounds to 64 afterward and drops that alignment. Fine for TP 1/2/4/8, wrong for any TP that doesn't divide 64.

Then the risks.

Buckets get silently dropped. seen is keyed on token_count alone, but ladder entries are (token_count, prefill_req_count, decode_req_count) triples (batch_dimensions_utils.py:21-36). Upstream generates token counts as multiples of 2, so everything from 2 to 64 rounds into the same bucket — and only the first survives, whichever that happens to be. If the survivor isn't decode-only, decode steps can't match a graph: is_applicable_for_batch_dim requires prefill_req_count == 0 for decode-only batches, commented "keep decode only property" (:56-61). No match means eager, and eager-vs-graphed is the divergence this PR exists to remove. Fix: key seen on the full triple.

The whole patch silently no-ops below 64 requests. cap = min(max_tokens, (max_requests // 64) * 64) is 0 when max_requests < 64. Every bucket then clamps to 0, fails the < 64 check, and the function returns [], []. The caller falls back to the original unfloored ladder and only logs on success (:600-607) — so the 64-alignment quietly doesn't happen and nothing says so. Same for the max_tokens is None or max_requests is None early return. Fix: raise, or log the fallback.

Positional-arg fallback reads the wrong slots. args[5]→max_requests, args[6]→max_tokens, but the real order is args[4]=max_requests, args[5]=max_tokens, args[6]=max_sequence_length. @classmethod doesn't put cls in *args, so nothing absorbs the shift. Dead today — the only call site passes everything by keyword (dynamic_context.py:688-699) — so simplest fix is to delete it rather than correct it. Also: upstream is @staticmethod; the patch converts it to @classmethod with no stated reason.

No tests. It's a pure function whose only dependency is a dataclass import, and grep finds no reference in tes

nemo_rl/models/policy/workers/moe_zero_kl_patches.py
```diff
@@ -0,0 +1,1012 @@
...
+
+Wired from ``setup._apply_zero_train_gen_mismatch`` for recipes such as
+Qwen3-30B-A3B DAPO with ``zero_train_gen_mismatch: true``:
+
+1. **MoE routing + combine** — ``topk_routing_with_score_function`` builds sparse
+   routing maps with ``index_put_`` (not ``scatter``); ``unpermute`` uses gather on
+   train and gather+droppad on CUDA-graphed decode. Fused TE router/unpermute paths
+   are rejected (``moe_permute_fusion=false`` required).
+2. **Router replay** — reconstruct finished-request routing before KV blocks release.
+3. **CUDA graphs** — 64-token graph bucket floor, runtime decode padding alignment,
```
ZhiyuLi-Nvidia: Emm I am just curious how's this 64 token graph relevant to batch invariant feature?

My point is like: batch-invariant means the result doesn't depend on batch size. If the kernels were genuinely batch-invariant, forcing M into 64x would be unnecessary by definition.

nemo_rl/models/policy/workers/moe_zero_kl_patches.py
```diff
@@ -0,0 +1,1012 @@
...
+        impl = _COMBINE_IMPL_OVERRIDE
+    else:
+        impl = os.environ.get("NRL_COMBINE_IMPL", "gather").strip().lower()
+    return impl if impl in ("segmented", "gather") else "gather"
+
+
+def _combine_gather_droppad() -> bool:
+    if _COMBINE_GATHER_DROPPAD_OVERRIDE is not None:
+        return _COMBINE_GATHER_DROPPAD_OVERRIDE
+    return os.environ.get("NRL_COMBINE_GATHER_DROPPAD", "0") == "1"
```
ZhiyuLi-Nvidia: These global variables are super hacky.

Do you think this would be better:


Finding: this block exists to choose among three combine kernels, but the correct choice is fully determined by the tensor layout, which _patched_unpermute already receives as arguments. The globals are a second, independent source of truth for the same decision — and they can disagree with the layout.

Risk: when they disagree, batch invariance silently breaks. Trace the dispatch:

impl = _combine_impl()                 # always "gather"
if impl == "gather" and routing_map is not None and not drop_and_pad:      → gather_combine
elif impl == "gather" and drop_and_pad and routing_map is not None and _combine_gather_droppad():
                                                                           → gather_combine_droppad
else:                                                                      → segmented_combine

gather_droppad is only set True by apply_cuda_graph_inference_determinism_patches(), which _apply_zero_train_gen_mismatch calls only when cuda_graphs_on. So in a graphs-off run the flag stays False, and a drop_and_pad=True call falls through to segmented_combine. But _unpermute_gather_combine's docstring says it "matches droppad expert-id order" — that shared ordering is the whole basis of the train/gen parity. Segmented ordering is not that order. So the one configuration where the flag differs from the layout is exactly the configuration that voids the guarantee.

Latent today, not active: both MoE recipes set cuda_graph_impl: local, and the 1.5B recipe is dense. But it's a trap armed for the first graphs-off MoE run.

Recommendation: delete the whole block. Dispatch on the layout, since the layout is what determines the correct reduction order:

if routing_map is None:
    out = _unpermute_segmented_combine(permuted_tokens, sorted_indices, restore_shape)
elif drop_and_pad:
    out = _unpermute_gather_combine_droppad(permuted_tokens, sorted_indices, restore
else:
    out = _unpermute_gather_combine(permuted_tokens, sorted_indices, restore_shape,

That removes configure_moe_combine_for_zero_kl, configure_moe_combine_for_cuda_graphombine_gather_droppad, _COMBINE_IMPL_OVERRIDE, _COMBINE_GATHER_DROPPAD_OVERRIDE, both env branches, the two call sites (:888, :955), and the restore bookkeeping (:1011-1012). The invariant "same layout → same reduction order" becomes structural rather than dependent on setup
ordering.

Two more globals in the same block worth dropping while you're there. _NRL_UNPERMUTEte_path is a dedup set that emits three WARNINGs per process saying which kernel ran— that's debug scaffolding at the wrong level. _NRL_DET_COMBINE_BANNER is a second one-shot flag guarding a print(). Both need reset logic in restore_moe_determinism_patches (:1007-1010). One
_NRL_LOGGER.debug(...) at the dispatch site replaces both, with no dedup state neede

Net: 3 module globals and 5 functions collapse to a 6-line if/elif/else.

The one thing to confirm with the author before proposing this: whether _unpermute_g outside CUDA-graph capture. If the flag exists to gate an unvalidated kernel ratherthan to select a phase, that's legitimate — but then it should be named for that and documented, because as written it reads as a phase selector and behaves as a correctness hazard.

nemo_rl/models/policy/workers/moe_zero_kl_patches.py
```diff
@@ -0,0 +1,1012 @@
...
+    permuted_tokens: torch.Tensor,
+    sorted_indices: torch.Tensor,
+    restore_shape: torch.Size,
+) -> torch.Tensor:
+    """Deterministic combine via segmented sum (no dense [T, max_slots, H] buffer)."""
+    num_tokens, hidden = restore_shape
+    num_permuted = permuted_tokens.size(0)
+    out = torch.zeros(num_tokens, hidden, dtype=permuted_tokens.dtype, device=permuted_tokens.device)
+    if num_permuted == 0:
+        return out
```
ZhiyuLi-Nvidia: I think we should minimize the patch if possible.

Do you know why we need this batch invariant or determinism ops in RL repo instead of mcore?

nemo_rl/models/policy/lm_policy.py
```diff
@@ -850,10 +850,23 @@ def generate(
...
+        if self.cfg["generation"]["backend"] == "megatron":
+            # Coordinator-based Megatron inference: only the DP=0 submitter returns
+            # tensors; other ranks return empty shells. Take the submitter's result
+            # directly — from_batches would also flatten any ndim>3 output tensors
+            # (e.g. routed_experts [B, S, L, K] for router replay).
+            result: BatchedDataDict[GenerationOutputSpec] = next(
+                wb
+                for wb in worker_batches
+                if wb.get("output_ids") is not None and wb["output_ids"].numel() > 0
+            )
```
ZhiyuLi-Nvidia: > only the DP=0 submitter returns tensors

Do you by the way know if there's alternative?

This is not good for scale at all.

Say if we can't do dp sharding then it is easily get OOM:

Per sample 128k * 35 * 2 * 4B = 35.84 MB
Per dp say 4 samples  per dp 4 x 128k * 35 * 2 * 4B = 143.36 MB
Total 8192 * 128k * 35 * 2 * 4B = 293 GB per step <===== DP0 without sharding.

ZhiyuLi-Nvidia:
There are lots of patches do you think we can minimize these and only keep the ones are really necessary?

Otherwise it would be really hard to maintain in the long run.
```
┌─────┬──────────────────────────────────────────────────────────────────────────┬─────────┬────────────────────────────────────────────────────────────────┐
│  #  │                                  Target                                  │  Owner  │                              Goal                              │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 1   │ te…cpp_extensions.gemm.get_cublas_workspace_size_bytes → 4               │ TE      │ Starve cuBLAS workspace → force splitK=1 (fixed reduction)     │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 2   │ model_utils._compute_distributed_log_softmax_with_grad                   │ NeMo-RL │ TP=1 use F.log_softmax to match gen raw_logprobs               │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 3   │ moe_utils.unpermute + token_dispatcher.unpermute                         │ mcore   │ Batch-invariant gather combine (the core idea)                 │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 4   │ moe_utils.topk_routing_with_score_function + moe_router.…                │ mcore   │ index_put_ routing maps instead of scatter                     │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 5   │ TextGenerationController._dynamic_step_context_bookkeeping               │ mcore   │ Reconstruct routing before update_requests frees KV blocks     │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 6   │ DynamicInferenceEngine.async_bookkeep                                    │ mcore   │ Apply that routing; drop upstream's finished_routing_block_ids │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 7   │ CUDAGraphBatchDimensionBuilder.generate_cuda_graph_batch_dimensions_list │ mcore   │ Force graph buckets to 64-multiples                            │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 8   │ DynamicInferenceContext.__setattr__                                      │ mcore   │ Round runtime decode padding to the same 64 quantum            │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 9   │ DynamicInferenceContext.initialize_attention_state                       │ mcore   │ Track graph-capture window                                     │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 10  │ CUDAGraphBatchDimensionBuilder.match_graph_config                        │ mcore   │ Suppress bucket match when graphs are off                      │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 11  │ MegatronGenerationMixin.prepare_for_generation / .finish_generation      │ NeMo-RL │ Track when inference graphs toggle on/off                      │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 12  │ MegatronGenerationMixin._sleep                                           │ NeMo-RL │ Clear expert padding before scoring                            │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 13  │ MambaMixer._ssm_prefill / ._ssm_decode + 5 injected helpers              │ mcore   │ Align train/prefill/decode SSM scan                            │
├─────┼──────────────────────────────────────────────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────┤
│ 14  │ megatron.core.utils.get_mamba_version                                    │ mcore   │ Avoid importing mamba_ssm (tvm crash on py3.13)                │
└─────┴──────────────────────────────────────────────────────────────────────────┴─────────┴────────────────────────────────────────────────────────────────┘
```

Plus non-patch global mutations: RouterReplay.global_router_replay_instances (rebuild), metadata.routing_indices_buffer = None, cg_utils.cuda_graph_attr_cache.pop(), and three env vars (CUBLAS_WORKSPACE_CONFIG=:0:0, CUBLASLT_WORKSPACE_SIZE=0, MAMBA_DETERMINISTIC=1).

Grouped by goal:
- Batch invariance (1, 2, 3, 4) — mcore's batch_invariant_kernels.py already owns 1 and 2; 3–4 are the genuine gap
- CUDA-graph shape pinning (7, 8, 9, 10) — all downstream of the unexplained 64
- Router replay plumbing (5, 6) — works around blocks being freed before serialization
- Mamba (13, 14) — 14 is an unrelated py3.13 import workaround riding along
- First-party lifecycle (11, 12) — patching our own code; should be hooks

Three observations that fall out of the table: 3 of 14 patch NeMo-RL's own code (2, 11, 12), which should be plain hooks or edits; #14 is unrelated to zero-KL and belongs in its own PR; and only 3, 4, 5 are genuinely novel — the rest is scaffolding around them.

YigongQin:
@wdykas please review. This PR enables bitwise parity for the two paths in M-inf. For low precision, it stays on `transformer_impl=transformer_engine`, and CUDA Graph is off for MOE models. For BF16, `inference_optimized` mode can be enabled. Fixes on logsoftmax calculations will be added in a following up PR.

Issue #2625, opened 2026-05-29
[Single Controller / Async RL] cleanup tracking issue


yuki-97:
Tracking issue for loose ends in the Single Controller / Async RL / Streaming.
Items will be checked off or split into separate issues as they land.

### Known cleanup items

Drop after new implementation aligned with old implementation.
- Async RL
  - [ ] Drop `nemo_rl/algorithms/async_utils/trajectory_collector.py` .
  - [ ] Drop old implementation in `nemo_rl/algorithms/async_utils/replay_buffer.py`.
- Streaming
  - [ ] Drop `nemo_rl/experience/rollouts.py`, new implementation is in #2567 and we'll no longer maintain sync rollout path.
  - [ ] Cleanup sync engine in inference backend, since we'll drop sync rollout path, so sync engine will no longer be used.

### Known missing items

<details>
<summary>✅ Resolved</summary>

- [ ] ~~Epoch semantics in the Single Controller train loop: `_train_pump` was bounded only by `max_train_steps`, with no `max_num_epochs`/`current_epoch` notion of full dataset passes.~~ — ✅ **Fixed in #2700** (`feat(sc): epoch-bounded rollout passes (max_num_epochs parity)`, commit `4c2bf5b6c`): SC now has `max_num_epochs` + a `_current_epoch` counter with epoch-bounded rollout dispatch, surfaced in `run()`/`ping()`.
- [x] ~~**Add a retry policy to the Single Controller train loop.** `_train_pump`'s per-cycle error path ([single_controller.py L437-L573](https://github.com/NVIDIA-NeMo/RL/blob/88e840731149a80138408c09e6c858851b451771/nemo_rl/algorithms/single_controller.py#L437)) aborts the worker step and re-raises on any mid-cycle failure (microbatch OOM, NaN-guard trip, prepare_logprobs / begin / finish failure), so a transient error kills the run — there is no retry. Flagged by the `TODO(sc): retry policy is a follow-up` at [L441](https://github.com/NVIDIA-NeMo/RL/blob/88e840731149a80138408c09e6c858851b451771/nemo_rl/algorithms/single_controller.py#L441). (Flagged during #2700 review.)~~ — Not planned.
[...]

PR #3219, opened 2026-07-15
feat(sc): rollout path — TQReplayBuffer, rollout_manager, _rollout_pump

yuki-97:
Part 1/4 of the #2819 split. See #2819 (reference) or #3267 (4/4) for the full code and context.

## Summary

Rollout-side infrastructure for the SC (Single Controller) async-GRPO training path.

- **`TQReplayBuffer`** (`nemo_rl/algorithms/async_utils/replay_buffer.py`): group-granular replay buffer with reserve/commit slot accounting; producer-side tensorization writes N training-shaped rows per prompt group. Replaces the WIP `ReplayBufferNew`.
- **`RolloutManager.generate_and_push`** (`nemo_rl/experience/rollout_manager.py`): reserve a buffer slot, run one prompt rollout, commit with start/end weight versions. Also renames `task_to_env → env_handles` across both `AsyncRolloutImpl` and `AsyncNemoGymRolloutImpl`.
- **`payload.py`** (`nemo_rl/experience/payload.py`): new `pack_payload` + `record_to_train_batch` helpers used by `TQReplayBuffer.commit`.
- **`SingleControllerActor._rollout_pump`**: rewritten to dispatch prompts through the new `RolloutManager`, with per-weight-version quota (`over_sampling=false`) and exact target-step matching (`force_in_order=true`). New `SingleControllerConfig` fields `over_sampling` / `force_in_order`; actor `__init__` gains `rollout_manager` / `dataloader` for driver-side wiring.
- **Tests**: new `test_tq_replay_buffer` and `test_rollout_pump`; new `test_rollout_manager` replaces the old `test_rollouts.py`; `TestReplayBufferNew` removed from `test_async_utils`.

**Note**: `_train_pump` is left in its pre-refactor form in this PR (still uses `dp_client.claim_meta` + the old top-level `StalenessSampler`); PR#2 switches it to the new `sampler.evict/select` API. SC becomes end-to-end runnable only after PR#3 lands the driver-side setup and entrypoint.

## Usage
`min_groups_for_streaming_train` (current called `min_groups_per_batch`, will be renamed in later PR)
| | min_groups_for_streaming_train | over_sampling | force_in_order |
|-|-|-|-|
| non-streaming (same as legacy) | num_prompts_per_step | False | True |
| streaming 1 | x (<= num_prompts_per_step) | False | True/False |
| streaming 2 | x (<= num_prompts_per_step) | True | False |

## Known Missing Features (Tracked in #2625)
1. Generation-level backpressure in the Single Controller: `max_inflight_prompts` -> `max_inflight_rollouts`
2. Add `over_sampling_ratio` (>1) to avoid over sample to much and waste data.


nemo_rl/algorithms/single_controller.py
```diff
@@ -321,7 +340,6 @@ def __init__(
 
         self._trainer_version: int = 0
         self._train_steps: int = 0
-        self._rollout_done: bool = False
```
yuki-97, line 324: for reviewers:
dropped this is to let `max_num_epochs` and `max_num_steps` control when to stop to avoid complexity.

`max_num_epochs` and `max_num_steps` will be aligned in the PR, ref: https://github.com/NVIDIA-NeMo/RL/blob/yukih/sc-entrypoint/nemo_rl/algorithms/single_controller_utils/setup.py#L226-L237

nemo_rl/algorithms/single_controller.py
```diff
@@ -397,74 +428,83 @@ async def _call_dp(self, method_name: str, **kwargs) -> Any:
...
-                await asyncio.gather(*tasks)
-                dispatched += k
-                self._current_epoch += 1
-                print(
-                    f"rollout_pump: epoch {self._current_epoch}/{max_epochs} "
-                    f"complete ({dispatched}/{n} prompts)",
-                    flush=True,
-                )
+                    self._dispatched_rollouts.add(task)
+                    task.add_done_callback(self._dispatched_rollouts.discard)
```
terrykong: [`single_controller.py:499-503`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/single_controller.py#L499-L503)

**[BUG] Uncaught rollout exceptions are silently swallowed and leak a `_buffer_capacity` permit — repeated failures deadlock the pump.**

The done-callback only discards the task; [`task.exception()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.exception) is never retrieved (the only trace is asyncio's GC-time "Task exception was never retrieved" log line — the run keeps going). If [`generate_and_push`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/experience/rollout_manager.py#L672-L697) raises anywhere in reserve→rollout→commit — the unprotected [env step](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/experience/rollout_manager.py#L191-L193) or [multi-reward assert](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/experience/rollout_manager.py#L198), a gym/DataPlane actor death, the [`assert n > 0` in `record_to_train_batch`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/experience/payload.py#L55), or tensorization on a malformed record — then:

1. [`_dispatch_one_prompt`'s `finally`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/single_controller.py#L467-L469) releases `sem` and `_inflight_rollouts`, but **not** `_buffer_capacity` — that permit is only released by the train pump after the group's *committed* rows are claimed and cleared. A failed rollout never commits, so each failure permanently loses one permit; after `max_buffered_rollouts` (default 8) failures the pump blocks forever on [`_buffer_capacity.acquire()`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/single_controller.py#L492) with no surfaced error.
2. The slot appended by [`TQReplayBuffer.reserve`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/async_utils/replay_buffer.py#L663-L688) stays `ready=False` forever — this PR removed `_flush_incomplete_groups`, so there is no reaper.
3. The old pump `await asyncio.gather(*tasks)`-ed the rollouts, so exceptions propagated to `run()`; this rewrite regresses to silent failure.

Cross-checked against the rest of the stack (#3220/#3266/#3267): #3220's `sampler.evict` does later reap slots whose *start weight* falls out of the staleness window (regardless of readiness) and releases their permits, which softens the windowed-streaming case — but an orphan still inside the window is untouched, and under `strict_on_policy` / `force_in_order` the trainer is waiting on that exact slot, so the version never advances and the run still deadlocks silently. The swallowed exception is never surfaced anywhere in the stack.

**Action:** clean up at the scope that owns each resource, then fail fast. Sketch (fail-fast default; a skip/retry policy for windowed streaming can layer on later under #2625):

In [`generate_and_push`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/experience/rollout_manager.py#L684-L697) — drop the reserved slot on failure and re-raise (a small `TQReplayBuffer.remove_group(group_id)` helper wrapping [`remove`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/async_utils/replay_buffer.py#L748-L787) keeps `_group_ids` private):

```diff
         start_version = self._weight_version
         group_id = self._tq_buffer.reserve(
             weight_version=start_version, target_step=target_step
         )
-
-        record = await self.run_rollout(input_sample)
-        end_version = self._weight_version
-
-        await self._tq_buffer.commit(
-            group_id,
-            record,
-            start_weight_version=start_version,
-            end_weight_version=end_version,
-        )
+        try:
+            record = await self.run_rollout(input_sample)
+            end_version = self._weight_version
+            await self._tq_buffer.commit(
+                group_id,
+                record,
+                start_weight_version=start_version,
+                end_weight_version=end_version,
+            )
+        except BaseException:
+            # Drop the reserved slot so it can't wedge force_in_order /
+            # strict_on_policy consumers; re-raise so the pump sees it.
+            await self._tq_buffer.remove_group(group_id)
+            raise
```

In [`_dispatch_one_prompt`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/single_controller.py#L452-L469) — release the permit paired with this dispatch on the error path, and surface the exception instead of discarding it (raising inside a done-callback only reaches the loop's exception handler, so store it and re-raise from the pump/`run()`):

```diff
         async def _dispatch_one_prompt(
             prompt: DatumSpec, target_step: Optional[int]
         ) -> None:
             self._inflight_rollouts += 1
             try:
                 await self._rollout_manager.generate_and_push(
                     prompt, target_step=target_step
                 )
                 ...
+            except BaseException:
+                self._buffer_capacity.release()  # permit acquired for this dispatch
+                raise
             finally:
                 self._inflight_rollouts -= 1
                 sem.release()
```
```diff
+        def _on_rollout_done(task: asyncio.Task) -> None:
+            self._dispatched_rollouts.discard(task)
+            if not task.cancelled() and task.exception() is not None:
+                self._pump_error = task.exception()  # run()/pump checks & raises
+
         ...
                     self._dispatched_rollouts.add(task)
-                    task.add_done_callback(self._dispatched_rollouts.discard)
+                    task.add_done_callback(_on_rollout_done)
```

(Alternative that gets the surfacing half for free: dispatch through an `asyncio.TaskGroup` (py≥3.11) wrapping the pump loop — the first child failure cancels siblings and propagates out of the pump to `run()`; the try/except for the permit/slot cleanup is still needed either way.) Related minor note: [`run()` cancels only the pump task](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/single_controller.py#L388-L392) — the children in `_dispatched_rollouts` are never cancelled/awaited on shutdown; draining them in the same fix would close both gaps.

tests/unit/single_controller/test_rollout_pump.py
```diff
@@ -0,0 +1,318 @@
...
+        out: dict[str, torch.Tensor] = {}
+        for k in td.keys():
+            v = td.get(k)
+            if isinstance(v, torch.Tensor) and v.is_nested:
+                v = torch.nested.to_padded_tensor(v, padding=0)
+            out[k] = v
+        return TensorDict(out, batch_size=td.batch_size)
+
+
+def test_rollout_pump_writes_expected_tq_data(
```
terrykong, line 357: [`test_rollout_pump.py:154`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/tests/unit/single_controller/test_rollout_pump.py#L154)

**[TEST] Missing `@pytest.mark.vllm` — this test drives a real vLLM engine but is collected in the default (non-vLLM) mode.**

The test uses the vLLM-backed `multi_step_setup_vllm_async`/`rollout_cluster` fixtures, yet has no marker (verified via `--collect-only`: it is collected with no `--vllm-only`). `L0_Unit_Tests_Other.sh` runs `unit/` without the vllm extra, so it will collect and attempt this test there. Its direct sibling [`test_async_rollout_manager` is marked](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/tests/unit/experience/test_rollout_manager.py#L342-L343).

**Action:** add the marker (and the missing `import pytest`):

```diff
 import time
 from typing import Any

+import pytest
 import ray
 import torch
```
```diff
+@pytest.mark.vllm
 def test_rollout_pump_writes_expected_tq_data(
```

tests/unit/experience/test_rollout_manager.py
```diff
@@ -0,0 +1,805 @@
...
+        "message_log": message_log,
+        "extra_env_info": metadata,
+        "task_name": "multi_step_calculator_game",
+        "stop_strings": ["<call: calculator>"],
+        "idx": 0,
+    }
+
+
+@pytest.mark.vllm
+def test_async_rollout_manager(
```
terrykong, line 365: [`test_rollout_manager.py:343`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/tests/unit/experience/test_rollout_manager.py#L343)

**[TEST] Moving these tests silently un-excludes them from the fast CI shards — `excluded_unit_tests.sh` still points at the old path.**

[`tests/unit/excluded_unit_tests.sh:221-223`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/tests/unit/excluded_unit_tests.sh#L221-L223) deselects `test_async_rollout_manager`, `test_async_rollout_manager_truncation`, and `test_async_nemo_gym_rollout_manager` under `unit/experience/test_rollouts.py::`. pytest silently ignores unknown `--deselect` nodeids, so after this move those ~67-91s tests now **run** in the FAST vLLM/nemo_gym shards that previously excluded them.

**Action:** repoint the three deselect paths in `tests/unit/excluded_unit_tests.sh` to this file:

```diff
-    --deselect=tests/unit/experience/test_rollouts.py::test_async_rollout_manager
-    --deselect=tests/unit/experience/test_rollouts.py::test_async_rollout_manager_truncation
-    --deselect=tests/unit/experience/test_rollouts.py::test_async_nemo_gym_rollout_manager
+    --deselect=tests/unit/experience/test_rollout_manager.py::test_async_rollout_manager
+    --deselect=tests/unit/experience/test_rollout_manager.py::test_async_rollout_manager_truncation
+    --deselect=tests/unit/experience/test_rollout_manager.py::test_async_nemo_gym_rollout_manager
```

(The other five `test_rollouts.py::` deselects at [L216-220](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/tests/unit/excluded_unit_tests.sh#L216-L220) are still valid — those tests remain in that file.)

tests/unit/single_controller/test_rollout_pump.py
```diff
@@ -0,0 +1,318 @@
...
+    cfg = SingleControllerConfig.model_construct(
+        batch_selection_strategy="staleness_window",
+        max_weight_staleness_versions=1,
+        min_groups_per_batch=1,
+        group_size=num_generations,
+        max_inflight_prompts=max_rollout_prompts,
+        max_buffered_rollouts=max_rollout_prompts,
+        max_train_steps=1,
+        max_num_epochs=None,
+        over_sampling=True,
```
terrykong, line 390: [`test_rollout_pump.py:187`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/tests/unit/single_controller/test_rollout_pump.py#L187)

**[TEST] The default dispatch path is untested — this test sets `over_sampling=True`, so the `over_sampling=False` batch gate is exercised by no test.**

With `over_sampling=True` the pump skips the [batch-level gate on `_max_rollout_version`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/single_controller.py#L475-L481) entirely; `over_sampling` defaults to `False`, so the *default* branch — and `force_in_order=True` → `target_step` stamping through [`reserve(target_step=...)`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/async_utils/replay_buffer.py#L663-L688) — has no coverage. (#3220 adds sampler-level `force_in_order` selection tests, but the pump-side gate and target-step stamping remain untested through the stack.)

**Action:** add a pump test with `over_sampling=False` (and one with `force_in_order=True` asserting `target_step_list` is populated) — or a lightweight unit test of the gate/target-step logic if a second vLLM E2E is too heavy.

pyrefly.toml
```diff
@@ -170,7 +171,6 @@ project-includes = [
   "nemo_rl/models/generation/vllm/vllm_backend.py",
   "nemo_rl/models/huggingface/__init__.py",
   "nemo_rl/models/megatron/__init__.py",
```
terrykong, line 185: [`pyrefly.toml:173`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/pyrefly.toml#L173)

**[GUIDELINE] This PR removes `nemo_rl/models/megatron/draft/__init__.py` from `project-includes`, but the file still exists** — dropping it silently exempts it from type-checking, and the change looks unrelated to the rollout path (possibly a rebase artifact).

**Action:** restore the entry (or confirm the removal is intentional):

```suggestion
  "nemo_rl/models/megatron/__init__.py",
  "nemo_rl/models/megatron/draft/__init__.py",
```

tests/unit/single_controller/test_rollout_pump.py
```diff
@@ -0,0 +1,318 @@
...
+from tests.unit.experience.test_rollouts import (
+    initial_multi_step_calculator_batch,  # noqa: F401
+    multi_step_calculator_environment,  # noqa: F401
+    multi_step_setup_vllm_async,  # noqa: F401
+    rollout_cluster,  # noqa: F401
+    rollout_tokenizer,  # noqa: F401
+)
+
+_PARTITION_ID = "rollout_data"
+# TQReplayBuffer.add tensorizes each PromptGroupRecord and writes
```
terrykong: [`test_rollout_pump.py:48`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/tests/unit/single_controller/test_rollout_pump.py#L48)

**[DOC] `TQReplayBuffer.add` doesn't exist — the API is `reserve` → `commit`** (tensorization happens in [`commit`](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/async_utils/replay_buffer.py#L690-L746)). Same stale reference at [L165](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/tests/unit/single_controller/test_rollout_pump.py#L165). Heads-up: #3220's rewording of this comment block also says `TQReplayBuffer.add`, so fixing the name here will want a matching touch-up when the stack rebases.

```suggestion
# TQReplayBuffer.commit tensorizes each PromptGroupRecord and writes
```

nemo_rl/algorithms/single_controller.py
```diff
@@ -321,7 +353,6 @@ def __init__(
 
         self._trainer_version: int = 0
         self._train_steps: int = 0
-        self._rollout_done: bool = False
```
terrykong, line 324: Anchoring the [self-note thread](https://github.com/NVIDIA-NeMo/RL/pull/3219#discussion_r3604314259) with pinned permalinks, since its branch ref moves:

- The dropped termination path this note refers to: [`self._rollout_done` + `_flush_incomplete_groups` at the pre-PR base](https://github.com/NVIDIA-NeMo/RL/blob/bed45e606d051c6e5667d36b372db5d8bb11dc4b/nemo_rl/algorithms/single_controller.py#L324).
- The `max_num_epochs`/`max_num_steps` alignment the note's `yukih/sc-entrypoint` link points to is [`_clamp_max_num_steps` at #3266's current head](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller_utils/setup.py#L211-L222) (`max_num_steps = min(max_num_steps, max_num_epochs * len(dataloader))`, applied in [`setup_single_controller`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller_utils/setup.py#L319)).

Context for other reviewers: at *this* PR's head, a finite `max_num_epochs` has no termination signal (the [`rollout_exhausted` branches](https://github.com/NVIDIA-NeMo/RL/blob/bf73365a86a6e07b58d6dfb64e2dea4da1cec3f8/nemo_rl/algorithms/single_controller.py#L550) are never set `True`, so the train pump would spin after the dataset drains) — that intermediate state is resolved within the stack: #3220 deletes the dead `rollout_exhausted` code in its `_train_pump` rewrite, and #3266's clamp above makes finite epochs terminate via the step budget.

tests/unit/experience/test_rollout_manager.py
```diff
@@ -0,0 +1,805 @@
...
+            "loss_multiplier": [1.0] * num_generations,
+        }
+    )
+
+    vllm_generation.prepare_for_generation()
+    original_batch, original_metrics = run_async_multi_turn_rollout(
+        policy_generation=vllm_generation,
+        input_batch=batch,
+        tokenizer=tokenizer,
+        task_to_env=env_handles,
```
terrykong: nit: do you think we should revert `env_handles` renaming of `task_to_env`? the renaming is purely cosmetic and you kind of lose the info of what this holds while having to learn a new naming

terrykong (commented):
Team review of the rollout-path infrastructure (PR 1/4 of the #2819 split). The split is clean: parity coverage from `test_rollouts.py` moved 1:1 into `test_rollout_manager.py` (both `*_matches_original` guards survive), removed symbols (`ReplayBufferNew`, `_rollout_done`, `_flush_incomplete_groups`) have no dangling references, and the producer→consumer DataPlane contract (sample-id scheme + `weight_version` tag) stays compatible with the untouched `_train_pump`. Linters pass; 18 of the PR's unit tests were run locally on GPU (all pass), including `test_async_rollout_manager_matches_original`.

Findings were cross-checked against the rest of the stack (#3220 / #3266 / #3267) — items already resolved downstream (e.g. the dead `max_rollout_prompts` field, deleted in #3220; the finite-`max_num_epochs` termination gap, closed by #3220's dead-code removal plus #3266's `max_num_steps ≤ max_num_epochs × len(dataloader)` clamp) are intentionally NOT flagged inline. The remaining inline comments cover only what survives the full stack: a silent-failure/permit-leak path in the new fire-and-forget rollout dispatch, a missing `@pytest.mark.vllm`, stale `excluded_unit_tests.sh` deselect paths after the test move, an untested default dispatch branch, a likely-rebase-artifact `pyrefly.toml` removal, and a doc nit.

Non-blocking observations (no action needed in this PR):
- `RolloutManager`'s `max_rollout_turns` default changed from `None` (coerced to 999999 on the native path) to `1`, and the coercion was removed. No production caller exists until PR#3, but a future native multi-turn caller that omits the arg now silently gets single-turn rollouts — worth keeping in mind for the PR#3 wiring.
- The new config-validation raises in `SingleControllerActor.__init__` live under `# pragma: no cover`; factoring them into a module-level helper would make them unit-testable (optional).

_Generated by Claude Code_

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [e48e80b](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/e48e80b81987028373c5fd4c22e3581d13c3d5c5). 

- Use Asyncio.TaskGroup to manage all rollout tasks.
- `SingleControllerActor._dispatched_rollouts` won't be used in this PR, but will be accessed in [PR! 3266](https://github.com/NVIDIA-NeMo/RL/pull/3266/changes#diff-02264c7e627a91fcb07f656f3aba4dd1a0d1d3604e685efeba666cf9080cb194R382), so leave it as is.

tests/unit/single_controller/test_rollout_pump.py
RayenTian: Done in [e48e80b](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/e48e80b81987028373c5fd4c22e3581d13c3d5c5)

tests/unit/experience/test_rollout_manager.py
RayenTian: Done in [e48e80b](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/e48e80b81987028373c5fd4c22e3581d13c3d5c5).

tests/unit/single_controller/test_rollout_pump.py
RayenTian: Done in [e48e80b](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/e48e80b81987028373c5fd4c22e3581d13c3d5c5)

tests/unit/single_controller/test_rollout_pump.py
RayenTian: Done in [e48e80b](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/e48e80b81987028373c5fd4c22e3581d13c3d5c5).

tests/unit/experience/test_rollout_manager.py
RayenTian: Done in [e48e80b](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/e48e80b81987028373c5fd4c22e3581d13c3d5c5)

pyrefly.toml
RayenTian: Done in [e48e80b](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/e48e80b81987028373c5fd4c22e3581d13c3d5c5)

RayenTian:
/ok to test [efc7d31](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/efc7d3179f6c8c49278b43c918f6a6dedd327c54)

RayenTian:
/ok to test [c454acf](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/c454acf9fb9af06122f61dbad535af31d1bcd0c7)

RayenTian:
/ok to test [f929aa0](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/f929aa0091cc4f38474f19a6abc7cb428b36a16c)

nemo_rl/algorithms/single_controller.py
terrykong: fixed in f929aa0091cc4f38474f19a6abc7cb428b36a16c

tests/unit/single_controller/test_rollout_pump.py
terrykong: fixed in f929aa0091cc4f38474f19a6abc7cb428b36a16c

tests/unit/experience/test_rollout_manager.py
terrykong: fixed in f929aa0091cc4f38474f19a6abc7cb428b36a16c

tests/unit/single_controller/test_rollout_pump.py
terrykong: fixed in f929aa0091cc4f38474f19a6abc7cb428b36a16c

pyrefly.toml
terrykong: fixed in f929aa0091cc4f38474f19a6abc7cb428b36a16c

tests/unit/single_controller/test_rollout_pump.py
terrykong: fixed in f929aa0091cc4f38474f19a6abc7cb428b36a16c

tests/unit/experience/test_rollout_manager.py
terrykong: fixed in f929aa0091cc4f38474f19a6abc7cb428b36a16c

nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -629,93 +635,206 @@ class ReplayBuffer(ReplayBufferImpl):
...
-
-            if len(all_indices) < num_prompt_groups:
-                print(
-                    f"Insufficient trajectories: have {len(all_indices)}, "
-                    f"need {num_prompt_groups}. Waiting."
+            except BaseException as rollback_error:
+                raise BaseExceptionGroup(
+                    f"commit and rollback both failed for group_id={group_id!r}",
+                    [commit_error, rollback_error],
                 )
```
terrykong, line 763: [`replay_buffer.py:757-761`](https://github.com/NVIDIA-NeMo/RL/blob/f929aa0091cc4f38474f19a6abc7cb428b36a16c/nemo_rl/algorithms/async_utils/replay_buffer.py#L757-L761)

**[MINOR] The rollback path can wrap a `CancelledError` in a `BaseExceptionGroup`, which breaks cancellation semantics.**

`except BaseException as commit_error` at [L748](https://github.com/NVIDIA-NeMo/RL/blob/f929aa0091cc4f38474f19a6abc7cb428b36a16c/nemo_rl/algorithms/async_utils/replay_buffer.py#L748) also catches `CancelledError` (e.g. `run()`'s `finally` [cancels `rollout_task`](https://github.com/NVIDIA-NeMo/RL/blob/f929aa0091cc4f38474f19a6abc7cb428b36a16c/nemo_rl/algorithms/single_controller.py#L400-L402) while a child is parked at the `put_samples` await). If the `clear_samples` rollback then *also* fails, the raised `BaseExceptionGroup([CancelledError, rollback_error])` is no longer a bare `CancelledError` — asyncio only marks a task as cancelled when `CancelledError` itself propagates out ([docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel)), so the enclosing `TaskGroup` misclassifies an expected teardown cancel as a genuine error. Impact is low (teardown-only, and `run()`'s `gather(return_exceptions=True)` usually swallows it), but preserving cancellation keeps shutdown reports clean.

**Action:** in `TQReplayBuffer.commit`, re-raise the cancel directly instead of grouping it:

```suggestion
            except BaseException as rollback_error:
                if isinstance(commit_error, asyncio.CancelledError):
                    raise commit_error from rollback_error
                raise BaseExceptionGroup(
                    f"commit and rollback both failed for group_id={group_id!r}",
                    [commit_error, rollback_error],
                )
```

terrykong (commented):
Follow-up review of the commits addressing the earlier round (bf73365 → f929aa0). The fixes hold up well under an adversarial concurrency pass: permit accounting is exactly-once on every interleaving we could construct (including cancel-before-start via `task_started_event`), `run()` now propagates rollout failures without orphaning tasks, and `commit()`'s rollback has no window against the train pump's claim path. The new tests genuinely exercise these paths, and the added `@pytest.mark.vllm` also fixes a latent lane bug (the test previously ran GPU work in the CPU-only Other shard). One low-severity note inline.

_Generated by Claude Code_

nemo_rl/algorithms/async_utils/replay_buffer.py
RayenTian: Done in [48184ad](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/48184ad00ccd68474ac2dd4118052309876d156f)

RayenTian:
/ok to test [48184ad](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/48184ad00ccd68474ac2dd4118052309876d156f)

terrykong (dismissed):

RayenTian:
/ok to test [62a759a](https://github.com/NVIDIA-NeMo/RL/pull/3219/commits/62a759a065b3a53447b9fe9e99422fe4edce9974)

terrykong (approved):

PR #3226, opened 2026-07-15
fix: forward calibrated KV cache scales on colocated ZMQ IPC refit path

### LINKED ISSUE #3225: Calibrated FP8 KV cache scales are dropped on the colocated ZMQ-IPC refit path (silently fall back to 1.0)
# Issue body

**Describe the bug**

On **colocated** generation (the common co-located GRPO/PPO setup where the policy
and vLLM share GPUs and weights are refit over ZMQ IPC), the trainer-calibrated
per-tensor FP8 KV cache scales from `calibrate_qkv_fp8_scales` never reach the
generation worker — every per-tensor K/V scale silently falls back to `1.0`.

The calibration runs correctly every step; its output is just discarded at the
refit call site because `kv_scales` is not passed to `stream_weights_via_ipc_zmq`.
Only the non-colocated NCCL path (`broadcast_weights_for_collective`) forwards
`kv_scales`, so the bug is specific to the colocated IPC transport.

Root cause (paths at base commit `4d756fe39`): the colocated IPC refit calls
`stream_weights_via_ipc_zmq` **without** `kv_scales`, while the collective path
passes it.

1. `nemo_rl/algorithms/grpo.py`, `refit_policy_generation` — the ZMQ IPC branch:
   ```python
   # Original ZMQ IPC path for vLLM
   futures_train = policy.stream_weights_via_ipc_zmq(
       buffer_size_bytes=buffer_size_bytes          # <-- kv_scales dropped
   )
   ```
   vs. the collective branch just below, which does pass it:
   ```python
   futures_train = policy.broadcast_weights_for_collective(kv_scales=kv_scales)
   ```

2. `nemo_rl/weight_sync/ipc_weight_synchronizer.py`, `IPCWeightSynchronizer.sync_weights`
   — the method accepts `kv_scales` but drops it at the call:
   ```python
   futures_train = self._policy.stream_weights_via_ipc_zmq(
       buffer_size_bytes=buffer_size_bytes          # <-- kv_scales dropped
   )
   ```

Downstream, the worker defaults to `1.0` when `kv_scales` is `None`
(`nemo_rl/models/policy/workers/megatron_policy_worker.py`,
`_iter_params_with_optional_kv_scales`):
```python
if kv_scales and param_name in kv_scales:
    scale_value = kv_scales[param_name]
else:
    scale_value = 1.0
```

**Steps/Code to reproduce bug**

1. Run GRPO with a **colocated** vLLM generation backend and FP8 KV cache enabled
   (`kv_cache_dtype=fp8`, KV-scale sync enabled) on a Megatron policy.
2. After a refit, inspect the per-tensor K/V scales landed in the generation
   worker's attention layers (`_k_scale` / `_v_scale`).
3. Observe they are all `1.0`, even though `calibrate_qkv_fp8_scales` produces
   non-1.0 per-layer values each step. Switching to the non-colocated (NCCL) path
   with the same config yields the calibrated (non-1.0) scales — confirming the
   drop is specific to the colocated IPC transport.

**Expected behavior**

The calibrated `kv_scales` should reach the worker on the colocated IPC path,
exactly as they do on the collective path, so the per-tensor scales used by vLLM
match what `calibrate_qkv_fp8_scales` computed (non-1.0, updated each step).

**Additional context**

- Impact: all colocated FP8 KV cache runs effectively used scale `1.0` instead of
  the calibrated per-tensor scales — i.e. "FP8 KV @ 1.0", not calibrated FP8 KV.
  No crash or warning, since `1.0` is a valid scale; FP8 activations simply
  clamp/round against a mis-scaled range, degrading KV-cache fidelity silently.
  Any A/B of "calibrated FP8 KV vs baseline" run on the colocated path is really
  "FP8 KV @ 1.0 vs baseline" and should be re-interpreted. The NCCL path is
  unaffected.
- Proposed fix: forward `kv_scales` on both colocated IPC call sites to match the
  collective path — two one-line additions, precision-independent, no new
  machinery. PR: 
- Environment: NeMo-RL `main` @ `4d756fe39`; Megatron policy worker + vLLM
  generation, colocated (ZMQ IPC refit).

babyplutokurt:
Fixes NVIDIA-NeMo/RL#3225

## What

Forward the trainer-calibrated per-tensor FP8 KV cache scales (`kv_scales`)<br>on the **colocated ZMQ-IPC** refit path, so they actually reach the generation<br>worker.

## Why

`calibrate_qkv_fp8_scales` computes per-tensor K/V scales every step, but they<br>were only forwarded on the **non-colocated NCCL** path<br>(`broadcast_weights_for_collective(kv_scales=...)`). The colocated ZMQ-IPC path<br>called `stream_weights_via_ipc_zmq(buffer_size_bytes=...)` **without**<br>`kv_scales`, so `_iter_params_with_optional_kv_scales` received `None` and<br>defaulted every per-tensor K/V scale to **1.0**<br>(`megatron_policy_worker.py`, `scale_value = 1.0`).

Net effect: in the common **colocated** GRPO/PPO setup, FP8 KV cache runs<br>silently used scale 1.0 instead of the calibrated values. The calibration ran<br>each step but its output was discarded at the refit call site — no error, just<br>uncalibrated KV cache.

## Change

Two one-line additions, passing `kv_scales` on both colocated IPC call sites so<br>they match the collective path:

* `nemo_rl/algorithms/grpo.py` — `refit_policy_generation`, the ZMQ IPC branch.
* `nemo_rl/weight_sync/ipc_weight_synchronizer.py` — `IPCWeightSynchronizer.sync_weights`<br>(the method already accepted `kv_scales` but dropped it at the call).

No new machinery; precision-independent; the receiving<br>`stream_weights_via_ipc_zmq` → `_iter_params_with_optional_kv_scales` chain<br>already threads `kv_scales` end-to-end.

## Verification

On a colocated run, the streamed `kv_scales` now populate the worker's<br>per-tensor scales each step (calibrated, non-1.0, drifting with activations)<br>instead of defaulting to 1.0.

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -2222,7 +2222,8 @@ def refit_policy_generation(
             else:
                 # Original ZMQ IPC path for vLLM
                 futures_train = policy.stream_weights_via_ipc_zmq(
-                    buffer_size_bytes=buffer_size_bytes
+                    buffer_size_bytes=buffer_size_bytes,
+                    kv_scales=kv_scales,
                 )
                 futures_inference = policy_generation.update_weights_via_ipc_zmq()
                 # wait for all futures to complete
--- a/nemo_rl/models/policy/interfaces.py
+++ b/nemo_rl/models/policy/interfaces.py
@@ -187,7 +187,9 @@ class ColocatablePolicyInterface(PolicyInterface):
 
     @abstractmethod
     def stream_weights_via_ipc_zmq(
-        self, *args: Any, **kwargs: Any
+        self,
+        buffer_size_bytes: int,
+        kv_scales: Optional[dict[str, float]] = None,
     ) -> list[ray.ObjectRef]:
         pass
 
--- a/nemo_rl/weight_sync/http_weight_synchronizer.py
+++ b/nemo_rl/weight_sync/http_weight_synchronizer.py
@@ -66,6 +66,12 @@ class HTTPWeightSynchronizer(WeightSynchronizer):
         timer: Optional[Timer] = None,
         kv_scales: Optional[dict[str, float]] = None,
     ) -> None:
+        assert kv_scales is None, (
+            "HTTP transport (SGLang) does not support FP8 KV cache scale sync. "
+            "`kv_scales` is accepted only for interface uniformity; if this "
+            "assertion fires, calibrated scales were routed through a path that "
+            "would otherwise silently ignore them."
+        )
         self._policy.offload_before_refit()
         self._generation.prepare_for_generation(tags=["weights"])
 
--- a/nemo_rl/weight_sync/interfaces.py
+++ b/nemo_rl/weight_sync/interfaces.py
@@ -85,9 +85,8 @@ class WeightSynchronizer(ABC):
         Args:
             timer: Optional Timer for profiling individual phases.
             kv_scales: Optional KV cache scales for FP8 quantization.
-                **Note**: Only honored by the NCCL collective transport,
-                which forwards them to ``policy.broadcast_weights_for_collective()``.
-                IPC and HTTP transports ignore this parameter.
+                Honored by the IPC/ZMQ and NCCL collective transports. The
+                HTTP transport ignores this parameter.
 
         Returns:
             Optional transport-specific scalar metrics for the current sync.
--- a/nemo_rl/weight_sync/ipc_weight_synchronizer.py
+++ b/nemo_rl/weight_sync/ipc_weight_synchronizer.py
@@ -83,7 +83,8 @@ class IPCWeightSynchronizer(WeightSynchronizer):
                 buffer_size_bytes = self._compute_buffer_size()
 
                 futures_train = self._policy.stream_weights_via_ipc_zmq(
-                    buffer_size_bytes=buffer_size_bytes
+                    buffer_size_bytes=buffer_size_bytes,
+                    kv_scales=kv_scales,
                 )
                 futures_inference = self._generation.update_weights_via_ipc_zmq()
 
--- a/tests/unit/algorithms/test_grpo.py
+++ b/tests/unit/algorithms/test_grpo.py
@@ -42,6 +42,7 @@ from nemo_rl.algorithms.grpo import (
     compute_and_apply_seq_logprob_error_masking,
     dynamic_sampling,
     grpo_train,
+    refit_policy_generation,
     validate,
 )
 from nemo_rl.algorithms.loss import ClippedPGLossConfig, ClippedPGLossFn
@@ -71,6 +72,33 @@ def _mock_policy_generation() -> MagicMock:
     policy_generation.requires_kv_scale_sync = False
     policy_generation.get_logger_metrics.return_value = {}
     return policy_generation
+
+
+@patch("nemo_rl.algorithms.grpo.ray")
+def test_refit_policy_generation_forwards_kv_scales_on_colocated_ipc(
+    mock_ray: MagicMock,
+) -> None:
+    mock_ray.get.return_value = [True]
+    policy = MagicMock()
+    policy_generation = MagicMock()
+    # Match VllmGeneration's default; a bare MagicMock would auto-create a truthy
+    # weight_synchronizer and refit_policy_generation would delegate to it instead
+    # of taking the colocated IPC path under test.
+    policy_generation.weight_synchronizer = None
+    kv_scales = {"layer.0": 0.5}
+
+    refit_policy_generation(
+        policy,
+        policy_generation,
+        colocated_inference=True,
+        _refit_buffer_size_gb=1.0,
+        kv_scales=kv_scales,
+    )
+
+    policy.stream_weights_via_ipc_zmq.assert_called_once_with(
+        buffer_size_bytes=1024**3,
+        kv_scales=kv_scales,
+    )
 
 
 class TestMaskSampleFilter:
--- a/tests/unit/weight_sync/test_weight_synchronizer.py
+++ b/tests/unit/weight_sync/test_weight_synchronizer.py
@@ -120,6 +120,19 @@ class TestIPCWeightSynchronizer:
         gen.prepare_for_generation.assert_any_call(tags=["kv_cache"])
 
     @patch("nemo_rl.weight_sync.ipc_weight_synchronizer.ray")
+    def test_sync_weights_passes_kv_scales(self, mock_ray: MagicMock) -> None:
+        mock_ray.get.return_value = [True]
+        policy = _mock_policy()
+        gen = _mock_generation()
+        sync = IPCWeightSynchronizer(policy, gen)
+        kv_scales = {"layer.0": 0.5}
+
+        sync.sync_weights(kv_scales=kv_scales)
+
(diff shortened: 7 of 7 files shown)


terrykong (commented):
@sharonyu-115 to review

ZhiyuLi-Nvidia:
Thanks a lot @babyplutokurt. 

Fix works — median max_seq_mult_prob_error drops 20× (132 → 6.5), median policy_kl_error drops and median policy_kl_error drops 0.27 → 0.05 (5×).

ZhiyuLi-Nvidia:
LGTM.

Just 2 nits:

1. Tighten stream_weights_via_ipc_zmq interface signature (structural root cause)

nemo_rl/models/policy/interfaces.py:189 currently uses *args, **kwargs, which is why the type checker couldn't warn the two IPC call sites that dropped kv_scales. All four concrete implementations (lm_policy, MegatronPolicyWorker, DTensorPolicyWorker, DTensorPolicyWorkerV2) already have the concrete typed signature, and both callers (grpo.py::refit_policy_generation, IPCWeightSynchronizer.sync_weights) pass compatible kwargs — so tightening the abstract is a no-op at runtime and safe against every existing call:

```
--- a/nemo_rl/models/policy/interfaces.py
+++ b/nemo_rl/models/policy/interfaces.py
@@ -186,9 +186,11 @@
     ) -> list[ray.ObjectRef]:
         pass

     @abstractmethod
     def stream_weights_via_ipc_zmq(
-        self, *args: Any, **kwargs: Any
+        self,
+        buffer_size_bytes: int,
+        kv_scales: Optional[dict[str, float]] = None,
     ) -> list[ray.ObjectRef]:
         pass
```

2. Loud-guard the HTTP synchronizer's silent kv_scales swallow (defensive)

HTTPWeightSynchronizer is wired exclusively for SGLang (nemo_rl/weight_sync/factory.py:91), and SGLangGeneration never overrides the default requires_kv_scale_sync = False, so refit_policy_generation never sends non-None kv_scales down this path today. Not a live bug. But HTTPWeightSynchronizer.sync_weights still accepts and silently drops the kv_scales kwarg (http_weight_synchronizer.py:67) — exactly the pattern this PR is fixing on the IPC path. A one-line assert keeps it interface-uniform without adding real behavior, and makes any future misconfig loud:

```
--- a/nemo_rl/weight_sync/http_weight_synchronizer.py
+++ b/nemo_rl/weight_sync/http_weight_synchronizer.py
@@ -63,6 +63,10 @@
     def sync_weights(
         self,
         *,
         timer: Optional[Timer] = None,
         kv_scales: Optional[dict[str, float]] = None,
     ) -> None:
+        assert kv_scales is None, (
+            "HTTP transport (SGLang) does not support FP8 KV cache scale sync. "
+            "`kv_scales` is accepted only for interface uniformity; if you're "
+            "seeing this, somethiated scales down "
+            "a path that will silently ignore them."
+        )
         self._policy.offload_before_refit()
         self._generation.prepare_"])
```

ZhiyuLi-Nvidia:
By the way, could you also rebase onto main branch? [CICD NeMo RL / Check if PR branch is up to date (push)](https://github.com/NVIDIA-NeMo/RL/actions/runs/29887909616/job/88822271996?pr=3226)Failing after 3s

babyplutokurt:
@ZhiyuLi-Nvidia Cool, added your suggested changes and rebased.

ZhiyuLi-Nvidia (dismissed):

babyplutokurt:
@ZhiyuLi-Nvidia Weird, one of the presubmit failed, but I don't think it's related to my code changes.

`L1_Functional_Tests_Megatron_1.sh` skipped `grpo_megatron.sh` and `grpo_megatron_mbridge_restore.sh` and ran only:
  1. `tests/functional/audio_grpo_megatron.sh`
  2. `tests/functional/grpo_megatron_eagle3_online.sh`
  
And the job dies at mid-run, can you take a look?

ZhiyuLi-Nvidia:
Sure. Let me rerun might be transition error.

babyplutokurt:
From logs, looks like a huggingface flaky outage issue

babyplutokurt:
``` HTTP Error 503 ... HEAD https://huggingface.co/datasets/gijs/avqa-processed/resolve/main/README.md
  Retrying in 2s [Retry 2/5]. ... Retrying in 8s [Retry 5/5].
  ...
  File "/opt/nemo-rl/nemo_rl/data/datasets/response_datasets/avqa.py", line 97, in __init__
      self.dataset = load_dataset("gijs/avqa-processed", split=split)
  FileNotFoundError: Couldn't find any data file at /opt/nemo-rl/gijs/avqa-processed.
    Couldn't find 'gijs/avqa-processed' on the Hugging Face Hub either:
    LocalEntryNotFoundError: An error happened while trying to locate the file on the Hub
```

babyplutokurt:
### 1. `L0_Unit_Tests_Algorithms` failure is on me (Now fiexed): 

  ```
  FAILED test_refit_policy_generation_forwards_kv_scales_on_colocated_ipc
  AssertionError: Expected 'stream_weights_via_ipc_zmq' to be called once. Called 0 times.
  ```

  `refit_policy_generation` opens with a synchronizer delegation guard:

  ```python
  # nemo_rl/algorithms/grpo.py:2167
  synchronizer = getattr(policy_generation, "weight_synchronizer", None)
  if synchronizer is not None:
      return synchronizer.sync_weights(timer=timer, kv_scales=kv_scales) or {}
  ```

  `VllmGeneration` declares `self.weight_synchronizer: WeightSynchronizer | None = None`
  (`vllm_generation.py:106`), so in production this is `None` and control falls through to
  the colocated IPC branch. But `MagicMock()` **auto-creates any attribute on access**, so
  `getattr(...)` returned a truthy child mock, the call returned early via
  `synchronizer.sync_weights(...)`, and the IPC branch under test never executed.

  Fix: pin the attribute to the real default:

  ```diff
   mock_ray.get.return_value = [True]
   policy = MagicMock()
   policy_generation = MagicMock()
  +# Match VllmGeneration's default; a bare MagicMock would auto-create a truthy
  +# weight_synchronizer and refit_policy_generation would delegate to it instead
  +# of taking the colocated IPC path under test.
  +policy_generation.weight_synchronizer = None
   kv_scales = {"layer.0": 0.5}
  ```
  
  
  
  ### 2. `fast_L1_Functional_Tests_Megatron_1` — unrelated HF Hub outage

  Died 1m13s into `audio_grpo_megatron.sh`, in dataset setup, before any training/refit ran:

  ```
  HTTP Error 503 ... HEAD https://huggingface.co/datasets/gijs/avqa-processed/resolve/main/README.md
  Retrying in 2s [Retry 2/5]. ... Retrying in 8s [Retry 5/5].
  ...
  File "nemo_rl/data/datasets/response_datasets/avqa.py", line 97, in __init__
      self.dataset = load_dataset("gijs/avqa-processed", split=split)
  FileNotFoundError: Couldn't find any data file at /opt/nemo-rl/gijs/avqa-processed.
  ```

  37 HTTP 503s in the log; all five `huggingface_hub` retries exhausted. Because the shard
  runs under `set -eou pipefail`, this aborted before `grpo_megatron_eagle3_online.sh` ran
  at all. No code path from this PR was reached.

  @ZhiyuLi-Nvidia  could you re-trigger when you get a chance?

terrykong (approved):
lgtm. thanks @babyplutokurt for the fix

PR #3232, opened 2026-07-16
test: unset CI env var in audiomcq VLM GRPO test

kajalj22:
## Summary

* `CI=true` is automatically set by GitLab CI in every runner shell and propagates into the SLURM job and container
* PyTorch changes behaviour when `CI` is set (disables certain optimizations), causing `vlm_vlm_grpo_qwen3_omni_30ba3b_audiomcq_4n8g_megatron_v1` to fail in CI but pass locally
* Unsetting `CI` inside the test script is the most targeted fix — only this test is affected, no infrastructure changes needed

## Test plan

- [ ] CI pipeline passes for `vlm_vlm_grpo_qwen3_omni_30ba3b_audiomcq_4n8g_megatron_v1` with this branch


yuki-97 (approved):
validated could pass using r0.7.0 branch + this PR, but failed (OOM) using main branch + this PR, tracked in https://github.com/NVIDIA-NeMo/RL/issues/3244.

Issue #3175, opened 2026-07-13
async_grpo doesn't respect to `grpo.max_num_epochs`


yuki-97:
as title.

abhinav-bellapu:
Hi @terrykong — I saw this is assigned to you. I have a draft patch and am opening it for visibility, marked as related rather than closing the issue. It bounds async training by `min(max_num_steps, max_num_epochs * len(dataloader))`, repeats collector passes up to `max_num_epochs`, and persists collector-owned epoch/dataloader state for partial-epoch resume. It also includes unit coverage and documentation.

I am happy to hand this off or close it if it overlaps with work already in progress.

yuki-97:
hi @abhinav-bellapu , thanks for the contribution! I left some comments in #3248.

bg51717:
Async PPO currently has the same limitation: the collector cycles the dataloader and training length is controlled only by `max_num_steps`, so `max_num_epochs` is not respected. In #3410, we require `max_num_epochs=-1` and document that `max_num_steps` should be used. Epoch-aware Async PPO support can be tracked here together with Async GRPO.

PR #3248, opened 2026-07-16
fix(grpo): honor max_num_epochs in async training

abhinav-bellapu:
# What does this PR do?

Makes async GRPO honor `grpo.max_num_epochs` and preserve the background collector's dataset position across checkpoints.

## Problem

Synchronous GRPO stops at the first configured training bound, but the async path only bounded its trainer loop with `max_num_steps`. Its background trajectory collector also iterated the training dataloader once and then reported exhaustion.

As a result, changing `grpo.max_num_epochs` did not reliably control async training. A run configured for multiple epochs could exhaust generation data after one pass, while a run with a small epoch bound could continue toward `max_num_steps`.

The checkpoint path adds another async-specific complication: generation can prefetch ahead of the trainer. The trainer's global step is therefore not a reliable proxy for the collector's completed epoch or partial dataloader position.

## Expected behavior

Async GRPO should use the same effective duration rule as synchronous GRPO:

```text
min(max_num_steps, max_num_epochs * batches_per_epoch)
```

The collector should provide at most `max_num_epochs` complete dataset passes, and a resumed job should continue from the exact saved dataloader/epoch position without replaying completed epochs.

### Trainer bound

- Computes `max_train_steps` from the configured step and epoch bounds.
- Uses that bound for the main loop, final-step validation, checkpointing, and completion messages.
- Returns cleanly when a restored job has already reached the effective bound.
- Keeps `total_steps` synchronized in async checkpoints.

### Collector epoch handling

- Gives `AsyncTrajectoryCollector` an explicit `start_epoch` rather than deriving its epoch from the trainer step.
- Recreates the dataloader iterator for each configured epoch.
- Distinguishes natural epoch exhaustion from a manual stop or collector error.
- Reports the collector's current epoch in driver-side status diagnostics.

### Checkpoint/resume consistency

- Snapshots the stateful dataloader and collector-owned completed-epoch count together under a lock.
- Persists that epoch in `grpo_save_state` before initializing the checkpoint.
- Restores the saved collector epoch for new checkpoints.
- Falls back to step-derived epoch recovery for legacy async checkpoints that predate collector epoch tracking.
- Supports a checkpoint taken partway through an epoch: resume consumes the remaining batches and then performs only the remaining full epochs.

# Issues

Related to #3175.

The issue is assigned to @terrykong, so this uses `Related to` rather than an auto-closing keyword. The implementation is offered for review and coordination and can be adjusted if it overlaps with work already in progress.

# Usage

```yaml
grpo:
  max_num_epochs: 2
  max_num_steps: 1000000
  async_grpo:
    enabled: true
```

With a dataloader containing 10 batches, this configuration performs at most 20 training steps. A smaller `max_num_steps` would take precedence.

# Testing

Added regression coverage for:

- Natural collector exhaustion and status reporting.
- Repeating the dataloader for multiple epochs.
- Resuming after completed epochs without replay.
- Resuming a partially consumed `StatefulDataLoader` epoch.
- Stopping async GRPO at the epoch-derived step bound.
- Saving the correct final step, collector epoch, and dataloader checkpoint.
- Existing step-bound and timeout behavior.

Local validation completed:

- Ruff lint and format checks passed.
- Python syntax/compile checks passed.
- `git diff --check` passed.
- An isolated `StatefulDataLoader` partial-epoch save/resume regression passed.

The complete Ray/CUDA unit and functional environment is not available on this macOS host and is left for NVIDIA CI.

# Documentation

Updates the async GRPO guide with the effective duration rule and clarifies that one epoch is one complete training-dataloader pass.

# Before this PR is ready to merge

- [x] Added unit regression coverage.
- [x] Updated the async GRPO documentation.
- [x] Ran available local lint, formatting, syntax, and isolated resume validation.
- [ ] Run the repository-selected Ray/CUDA CI suite.

### DIFF
--- a/nemo_rl/algorithms/async_utils/trajectory_collector.py
+++ b/nemo_rl/algorithms/async_utils/trajectory_collector.py
@@ -62,6 +62,7 @@ class AsyncTrajectoryCollector:
         master_config: MasterConfig,
         replay_buffer: Any,
         start_step: int = 0,
+        start_epoch: int = 0,
         teacher_worker_groups: Optional[dict[str, Any]] = None,
         alias_to_group_alias: Optional[dict[str, str]] = None,
         on_policy_distillation_cfg: Optional[dict[str, Any]] = None,
@@ -103,6 +104,8 @@ class AsyncTrajectoryCollector:
 
         self.current_weight_version: int = start_step
         self.initial_weight_version: int = start_step
+        self.current_epoch: int = start_epoch
+        self._dataloader_lock: _threading.Lock = _threading.Lock()
         self.dataloader: StatefulDataLoader | None = None
         self.collection_thread: _threading.Thread | None = None
 
@@ -266,6 +269,7 @@ class AsyncTrajectoryCollector:
             "data_exhausted": self.data_exhausted,
             "errored": self.collection_failed,
             "inflight_workers": inflight_workers,
+            "current_epoch": self.current_epoch,
         }
 
     def _collection_loop(self):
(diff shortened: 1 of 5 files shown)


nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -598,9 +633,17 @@ def wait_for_pending_generations(self) -> None:
 
     def get_dataloader_state(self) -> dict:
         """Get the current dataloader state for checkpointing."""
-        if hasattr(self, "dataloader") and hasattr(self.dataloader, "state_dict"):
-            return self.dataloader.state_dict()
-        return {}
+        state, _ = self.get_dataloader_checkpoint()
```
yuki-97: looks `get_dataloader_state` is just a thin wrapper after the change.

wdyt remove `get_dataloader_checkpoint` function and just move things to under `get_dataloader_state`?
we can change to still `return state, self.current_epoch` but make the change at where it is called.

nemo_rl/algorithms/grpo.py
```diff
@@ -3738,6 +3756,15 @@ def async_grpo_train(
         },
     }
 
+    # Async checkpoints created before collector epoch tracking only updated
+    # current_step. New checkpoints keep total_steps in sync and persist the
+    # collector's own completed-epoch count because generation may prefetch
+    # ahead of the trainer.
+    if grpo_save_state.get("total_steps", 0) == step:
```
yuki-97: `get("total_steps", 0) == step` is really "does this ckpt carry `current_epoch`?" but tests it via an equality that only holds because async writes `total_steps == current_step` in the same handler. If async ever adopts sync-GRPO's cumulative-`total_steps` semantics, this silently takes the wrong branch on old ckpts.

```suggestion
    if "current_epoch" in grpo_save_state:
```

nemo_rl/algorithms/grpo.py
```diff
@@ -3624,6 +3624,24 @@ def async_grpo_train(
 
     # Training state
     step = grpo_save_state["current_step"]
+    steps_per_epoch = len(dataloader)
+    max_num_epochs = master_config.grpo["max_num_epochs"]
+    configured_max_num_steps = master_config.grpo["max_num_steps"]
+    max_train_steps = min(
+        configured_max_num_steps,
+        max_num_epochs * steps_per_epoch,
+    )
```
yuki-97: wdyt mirror the logic of `_clamp_max_num_steps` in `nemo_rl/algorithms/single_controller_utils/setup.py`? by doing so we only need to focus on `max_num_steps` in train loop.

yuki-97 (commented):
@abhinav-bellapu thanks for adding this support and sorry for late reply. overall LGTM and left some comments.

abhinav-bellapu:
/ok to test [b681acf](https://github.com/NVIDIA-NeMo/RL/commit/b681acfa1aeefc97c2b22eece74c396e739d1a86)

nemo_rl/algorithms/grpo.py
abhinav-bellapu: addressed: this now checks "current_epoch" in grpo_save_state directly. I also added regression coverage for both new checkpoints and legacy checkpoints without that field

nemo_rl/algorithms/async_utils/trajectory_collector.py
abhinav-bellapu: addressed: get_dataloader_state() now returns (state, current_epoch) directly under the lock, and I removed get_dataloader_checkpoint(). I updated the call sites and tests accordingly

nemo_rl/algorithms/grpo.py
abhinav-bellapu: addressed: master_config.grpo["max_num_steps"] is now clamped once using the epoch-derived bound, and the async training loop only uses the resulting max_num_steps

abhinav-bellapu:
gentle bump @yuki-97 when you have a chance, thanks!

yuki-97 (dismissed):
hi @abhinav-bellapu , LGTM. sorry for late reply and you'll need do a rebase to solve the conflict. I've triggered the CI first to see if there's some fails.

also let me know do you mind us helping to do rebase and fix failed tests (if there's some) to get it in?  which may help to accelerate to get it in if you haven't a chance to do it.

abhinav-bellapu:
thanks @yuki-97. i rebased onto the current main and resolved the conflicts

the new head is `81573621fd5f27ac1aa7a1406a11de07b3ff7023`. could you please re-review?

yuki-97 (approved):
thanks @abhinav-bellapu , LGTM. I rebased since there's a conflict, and enqueued.

PR #3253, opened 2026-07-16
fix: preserve runtime tools after arm64 Apptainer build

kajalj22:
## Summary

- The arm64 Apptainer source-build path purged every package declared as a build dependency, including tools already supplied by the base image or installed by `docker/Dockerfile`.
- The initial cleanup removed `curl`, `git`, and `wget`, causing GB200 tests to fail at the first `git config` command.
- After preserving those tools, the image-build log showed that `build-essential` was also pre-installed. Purging it with `--auto-remove` removed GCC, G++, Make, and CUDA compiler packages, causing Triton JIT compilation to fail at runtime.
- Keep `build-essential` as an explicit runtime dependency, record which temporary build packages were absent before installation, and purge only that newly installed subset.
- Revert the earlier `BASH_ENV` workaround because it did not address the missing executables.

The amd64 Apptainer path installs a release package and does not use this source-build cleanup, which is why the corresponding H100 jobs were unaffected.

## Test plan

- [x] Run `bash -n docker/install_apptainer.sh`
- [x] Run `git diff --check`
- [ ] Confirm the GB200 image retains `git`, `curl`, `wget`, GCC, G++, Make, and the CUDA compiler packages
- [ ] Confirm GB200 functional tests proceed through Triton JIT compilation
- [ ] Confirm H100 functional tests remain unaffected

### DIFF
--- a/.github/actions/test-template/action.yml
+++ b/.github/actions/test-template/action.yml
@@ -116,6 +116,7 @@ runs:
           --env HF_DATASETS_CACHE=/home/TestData/nemo-rl/hf_datasets_cache \
           --env NEMO_RL_REPO_DIR=/opt/nemo-rl \
           --env HF_TOKEN \
+          --env BASH_ENV= \
           ${{ inputs.is_fork_pr == 'true' && '--env HF_HUB_OFFLINE=1' || '' }} \
           ${{ inputs.image-tag != '' && '--env FAST=1' || '' }} \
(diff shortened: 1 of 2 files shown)


terrykong (dismissed):
lgtm. but should we report this since arm vs x86 having this diff is surprising and seems like we should handle at the cuda base image layer

thomasdhc (dismissed):

chtruong814 (dismissed):

chtruong814 (approved):

PR #3259, opened 2026-07-17
feat(weight-sync): add ModelExpress trainer publishing

KavinKrishnan:
## Overview

Adds trainer-side publication for a ModelExpress (MX) weight-refit backend. NeMo RL's `WeightSynchronizer` remains the owner of the refit lifecycle; this PR adds the policy operation that a future `ModelExpressWeightSynchronizer` will call.

The existing DTensor worker and the Megatron worker publish rank-local tensor ownership and topology metadata. Generation-side discovery, transfer, resharding, and installation remain separate reviews.

## Changes

- Adds `publish_weights_for_model_express()` to the policy interface and trainer workers.
- Publishes native Tensor Parallelism (TP), Pipeline Parallelism (PP), and Expert Parallelism (EP) shards without gathering the full model on rank zero.
- Describes global tensor geometry, rank coordinates, expert ownership, and Megatron-to-Hugging Face name mappings.
- Requires the public ModelExpress publisher lifecycle; no private MX state is accessed.
- Fails before publication for unsupported DTensor placements or unknown Megatron placement at TP greater than one.
- Adds focused lifecycle, geometry, role-classification, and failure tests.
- Adds the trainer-publication design and current implementation status to the documentation.

## Architecture boundary

```text
WeightSynchronizer                              [follow-up]
  -> policy.publish_weights_for_model_express() [this PR]
  -> generation discovery and transfer          [follow-up]
  -> inference reshard and install               [follow-up]
  -> global completion                           [follow-up]
```

This PR does not create a second orchestration path. Existing IPC, HTTP, and NCCL synchronizers are unchanged.

## Scope

This PR does not add:

- a user-facing ModelExpress refit configuration;
- generation-worker discovery, transfer, or installation;
- `DTensorPolicyWorkerV2` / AutoModel publication;
- Dynamo integration;
- Kubernetes resources, recipes, or benchmark harnesses.

The future synchronizer will own user configuration and construct the internal publisher settings explicitly.

## Current status

**Partially aligned:** rank-local publication is implemented and unit-tested for the existing DTensor worker and the Megatron worker. End-to-end refit requires the generation-worker adapter and the corresponding ModelExpress publisher, planning, and installer APIs.

## ModelExpress dependencies

- filtered and versioned discovery: ai-dynamo/modelexpress#519 (merged);
- mapped vLLM installation: ai-dynamo/modelexpress#521;
- no-gather geometry and reshard planning: ai-dynamo/modelexpress#496;
- required publisher contract: public `reset_tensors()` plus one `add_tensor()` surface for DTensor and Megatron metadata (follow-up review).

## Relationship to #3068

This is the trainer-publication portion of #3068, separated by ownership and review scope. The umbrella remains a reference for the complete integration.

## Test plan

- [x] 14 focused publication lifecycle, DTensor geometry, and Megatron classification tests
- [x] Ruff lint and formatting
- [x] Focused pyrefly check for new modules
- [x] Python compilation and Git whitespace checks
- [x] DCO sign-off
- [ ] Full repository and GPU end-to-end tests through maintainer CI

### DIFF
--- a/docs/design-docs/model-express-refit.md
+++ b/docs/design-docs/model-express-refit.md
@@ -76,13 +76,14 @@ second synchronization lifecycle to algorithm code.
 ## Current status
 
 **Partially aligned:** trainer-side rank-local publication and metadata
-construction are implemented. Focused unit tests cover DTensor shard ranges,
-unsupported placements, replicated-tensor ownership, fused QKV classification,
-and global expert IDs.
+construction are implemented for the existing DTensor worker and the Megatron
+worker. Focused unit tests cover DTensor shard ranges, unsupported placements,
+replicated-tensor ownership, fused QKV classification, and global expert IDs.
 
 Generation-side discovery, transfer, translation, installation, and
-end-to-end GPU validation are follow-up work. Existing NeMo RL weight
-synchronizers remain unchanged.
+end-to-end GPU validation are follow-up work. The AutoModel-based
+`DTensorPolicyWorkerV2` does not yet implement MX publication. Existing NeMo RL
+weight synchronizers remain unchanged.
 
 ## Assumptions
 
@@ -115,6 +116,7 @@ synchronizers remain unchanged.
 ## Open questions
 
 - Final user-facing configuration for `ModelExpressWeightSynchronizer`.
+- Rank-local publication for `DTensorPolicyWorkerV2`.
 - The common install-plan contract between MX reshard planning and the
   inference adapter.
 - Retention and read-lease behavior for trainer buffers during long updates.
--- a/docs/design-docs/model-express-refit.md
+++ b/docs/design-docs/model-express-refit.md
@@ -1,0 +1,132 @@
+# ModelExpress Weight Refit
+
+## Problem
+
+NeMo RL periodically updates generation workers with policy weights produced by
+trainer workers. A non-colocated deployment needs to move those weights across
+different process groups and may use different parallel layouts for training
+and generation.
+
+ModelExpress (MX) provides source discovery and topology-aware transfer
+planning for this exchange. The NVIDIA Inference Xfer Library (NIXL) can move
+the resulting byte ranges directly between worker memory. The MX server stores
+metadata and lifecycle state; it does not carry model weights.
+
+## Recommended design
+
+Each trainer rank publishes the tensor shards it already owns. Generation
+workers discover a complete source set for one policy version, request the
+ranges required by their local layout, and install the received tensors through
+an inference-engine adapter.
+
+```mermaid
+flowchart TB
+    T["**1. Trainer ranks**<br/>publish local tensor ownership"]
+    C["**2. MX catalog**<br/>stores version and topology metadata"]
+    G["**3. Generation workers**<br/>discover sources and build local plans"]
+    X["**4. NIXL data plane**<br/>moves bytes directly between workers"]
+    I["**5. Inference adapter**<br/>transforms and installs local weights"]
+
+    T --> C
+    C --> G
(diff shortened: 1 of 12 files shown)


KavinKrishnan:
### Added a correctness fix to this branch: global layer indices under pipeline parallelism

Pushed `21f804e5`. Folding it in here rather than opening a follow-up, because this PR is what introduces `mx_megatron_helpers.py` and merging it as-is would upstream a defect that is invisible to every correctness gate.

**What was wrong.** `model.named_parameters()` names a layer by its index in the *local* `TransformerBlock.layers` ModuleList. Under pipeline parallelism that is not the global index, so at `pp_size=2` on a 48-layer model both stages publish `decoder.layers.0..23`. The reshard shard table is keyed by name with a first-writer-wins tiebreak, so one stage's 24 layers displace the other's. Global layers 24-47 are never refit, and where a stage-1 offer wins, destination layer N is filled from global layer N+24 — at a valid address, with a self-consistent digest.

Nothing catches this. Every gate compares arrived bytes against the publisher's digest for the same name, and a name collision makes both sides agree on the wrong tensor. Measured on Qwen3-30B-A3B at trainer TP1/PP2/EP8/DP8: coverage 51%, 15.58 GB moved against the engine's own 30.6 GB per-rank footprint. After the fix, coverage is 100.0% and `params_installed` is 435/435 on all 16 ranks.

**Why `layer_number` and not arithmetic.** `map_local_layers_to_global` reads the global index off `TransformerLayer.layer_number`, which Megatron already maintains as 1-based and global — its own checkpoint code subtracts an offset from it to recover the ModuleList index. The tempting `pp_rank * layers_per_stage` is wrong under virtual pipeline parallelism, where a rank owns several non-contiguous chunks, and under uneven stages (`num_layers_in_first_pipeline_stage`). Reading the attribute is exact for both.

It **raises** rather than falling back to the local name when pipeline parallelism is on and no layer exposes `layer_number`. That is deliberate and I would push back on softening it to a warning: the silent fallback is exactly the failure being fixed.

**Reviewer notes.**

| file | lines | what to look at |
| :--- | --: | :--- |
| `nemo_rl/distributed/mx_megatron_helpers.py` | +89 | The two new functions, and the one-line hook in `collect_megatron_publish_set`. |
| `tests/unit/distributed/test_mx_megatron_pp_layer_names.py` | +178 | 8 tests. The two worth reading are the virtual-pipeline/uneven-stage case and the raise-don't-fall-back case. |
| `nemo_rl/models/policy/workers/megatron_policy_worker.py` | +1 | Passes `pp_size`, already in scope at the call site. |

`collect_megatron_publish_set` gains a keyword-only `pp_size: int = 1`. The default is deliberate: at `pp_size=1` local index equals global index, so existing callers keep their current behavior and the guard cannot fire on them.

One thing worth deciding in review rather than pre-deciding: Bridge has `_megatron_local_name_to_global`, which is arguably the more correct dependency to lean on upstream. The `layer_number` approach shipped here needs nothing outside this module, which is why it went in first.

KavinKrishnan:
Rebased onto `main` (was 69 commits behind) and squashed the four commits into one, since the later three only reworked the first and replaying them produced conflicts against the same lines four times over.

Two files needed a real resolution rather than a clean replay. `main` grew an NCCL-reshard path (`prepare_nccl_reshard_refit_info`, `nccl_reshard_refit`, and the pipeline-stage helpers on the Megatron worker) in the same places this change adds the ModelExpress path. The two are independent, so both are kept side by side and nothing was dropped. The net diff against `main` is unchanged line for line from before the rebase.

No functional change, so this is still the same review.

PR #3262, opened 2026-07-17
feat(ppo): support non-colocated generation

bg51717:
# What does this PR do ?

Adds synchronous non-colocated PPO, with the policy and value model sharing a training cluster while vLLM generation runs on a separate inference cluster.

## Design

The main changes are:

- Create separate training and inference `RayVirtualCluster`s when `policy.generation.colocated.enabled=false`.
- Support both splitting GPUs on a single node and assigning complete nodes to generation in multi-node runs, with fail-fast validation for invalid resource configurations.
- Initialize the cross-cluster collective used to transfer updated policy weights to vLLM after training.
- Coordinate policy and value offloading so they can continue to time-share the training GPUs while generation remains resident on the inference GPUs.
- Apply topology constraints to the training and inference clusters when `cluster.segment_size` is configured.
- Keep the existing colocated PPO path supported and reject non-colocated SGLang explicitly because it does not provide the required cross-cluster refit path.
- Align PPO's sequence-level train/generation logprob-error metrics and optional masking with the shared GRPO implementation, including an explicit error when filtering leaves no valid response tokens.
- Add AutoModel and Megatron functional tests, 40-step convergence recipes, unit coverage, and PPO deployment documentation.

## Related correctness fixes

- **Filtered-sample GAE semantics:** Compute PPO advantages with `token_mask * sample_mask`, matching GRPO, so samples removed by `overlong_filtering` do not affect GAE or advantage normalization. When truncated samples are present, this changes the previous PPO advantage numerics even if the optional sequence-level threshold is disabled.
- **Reward-model resource detection for structured data configs:** Collect `env_name` from every dict or list entry under `data.train`, `data.validation`, and `data.default`. Previously, list-shaped dataset configs were skipped, which could prevent PPO from reserving resources for an active reward-model environment.

# Usage

Set `policy.generation.colocated.enabled=false` and reserve resources for vLLM generation:

```yaml
policy:
  generation:
    backend: vllm
    colocated:
      enabled: false
      resources:
        gpus_per_node: 4
        num_nodes: null
```

For a single-node run, `gpus_per_node` reserves that many GPUs for generation on the node. For a multi-node run, generation uses complete nodes, so set `num_nodes` and keep `gpus_per_node` equal to `cluster.gpus_per_node`.

**Pre checks**:

# Test

The following test matrix passed:

- Unit and regression tests passed for PPO masking and resource validation, shared async/GRPO compatibility, data/config parsing, checkpoint resolution, recipe validation, and Megatron value-worker coverage.
- AutoModel functional tests: colocated and non-colocated PPO checks passed.
- Megatron functional tests: colocated and non-colocated PPO checks passed, including critic warmup and sequence logprob-error masking.
- AutoModel 40-step non-colocated E2E: step-40 reward `0.8408`, validation accuracy `0.7500`, median token probability error `1.0076`, and median maximum sequence probability error `1.0261`.
image

- Megatron 40-step non-colocated E2E: step-40 reward `0.8467`, validation accuracy `0.7383`, median token probability error `1.0087`, and median maximum sequence probability error `1.0270`.
image


bg51717:
/ok to test [4fe31a4](https://github.com/NVIDIA-NeMo/RL/pull/3262/commits/4fe31a44aa8e4227041a1ef1f63c488ae9c08894)

bg51717:
/ok to test [6767e6d](https://github.com/NVIDIA-NeMo/RL/pull/3262/commits/6767e6df5013761d4b20eee6c9ee2af6383b33a2)

nemo_rl/algorithms/ppo.py
```diff
@@ -1253,7 +1475,7 @@ def ppo_train(
                     adv_kwargs = dict(
                         prompt_ids=prompt_ids_for_adv,
                         rewards=train_data["rewards"],
-                        mask=train_data["token_mask"],
+                        mask=advantage_mask,
```
yuki-97, line 1503: `mask=train_data["token_mask"]` → `mask=advantage_mask` at `ppo.py:1454` silently shifts advantage-computation numerics for pre-existing recipes with `overlong_filtering: true` — `advantage_mask = token_mask * sample_mask` (`ppo.py:187`) zeros filtered samples out of advantages, whereas pre-PR they entered via `token_mask` alone. Same axis on `response_advantages = torch.masked_select(flat_advantages, advantage_mask.bool())` at `ppo.py:1590`.

Two shipped recipes hit this at default `seq_logprob_error_threshold=null`: `ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp.yaml` and `…-megatron-valuetp2sp-dynbatch.yaml`.

**Suggest fix:** call out in PR description / commit body that this aligns PPO to GRPO's existing `token_mask * sample_mask` convention (`grpo.py:2864-2871`), and that the shift fires from `overlong_filtering=true` even when the new threshold stays `null`.

tests/unit/algorithms/test_ppo.py
```diff
@@ -693,3 +698,961 @@ def test_create_advantage_estimator_requires_adv_estimator_key():
...
+        match=(
+            "policy.generation.colocated.resources.gpus_per_node must be explicitly set"
+        ),
+    ):
+        ppo_mod.setup(config, MagicMock(), _setup_dataset(), None)
+
+    cluster_cls.assert_not_called()
+
+
+def test_noncolocated_inference_requires_explicit_gpus_per_node_multi_node(
```
yuki-97, line 1437: [TEST] Four assert-branches in the new non-colocated setup path have no coverage in `test_ppo.py`:

- `ppo.py:482` `train_gpus_per_node > 0` — no test triggers "Not enough GPUs for PPO training after reserving …".
- `ppo.py:491` multi-node `inference_nodes is not None and > 0` — `test_noncolocated_inference_requires_explicit_gpus_per_node_multi_node` (`test_ppo.py:1373`) actually exercises the sibling `gpus_per_node` assertion at `ppo.py:496`, not this one.
- `ppo.py:496` multi-node `inference_gpus_per_node == cluster.gpus_per_node` — `is None` sub-case covered; set-but-wrong-value (e.g. `4` vs `cluster=8`) is not.
- `ppo.py:561` non-divisible `nodes_per_instance` warning branch — the divisible IF is tested (`test_ppo.py:1488`); the silent-warning ELIF isn't.

**Suggest fix:** one parametrized case per branch mirroring the shape of `test_noncolocated_inference_requires_explicit_gpus_per_node_multi_node`.

tests/unit/algorithms/test_ppo.py
```diff
@@ -693,3 +698,961 @@ def test_create_advantage_estimator_requires_adv_estimator_key():
...
+        {"domain-1": 0.001},
+        {"domain-1": 0.001},
+    ]
+    assert inference_cluster.kwargs["segment_size"] == 2
+    generation_factory.init_cluster_placement_groups.assert_called_once_with(
+        inference_cluster, config.policy["generation"]
+    )
+
+
+def test_noncolocated_vllm_builds_separate_clusters_and_collective(monkeypatch):
```
yuki-97, line 1688: The "noncolocated" test suite MagicMocks `refit_policy_generation`, `VllmGeneration`, and `Policy`, so it verifies argument wiring but never exercises the cross-cluster collective handshake — the PR's headline feature.

wdyt adding one lighter-weight integration test that uses fake Policy/Vllm whose `init_collective` actually completes a handshake between two `DummyCluster`s, or otherwise explicitly scope these tests' docstrings as "wiring-only" so a real cross-cluster test doesn't get skipped by mistake?

nemo_rl/algorithms/ppo.py
```diff
@@ -393,14 +425,16 @@ def setup(
...
-        "PPO currently requires colocated generation (vLLM / SGLang sharing GPUs "
-        "with the policy worker). Set policy.generation.colocated.enabled=true. "
-        "Non-colocated PPO is not yet supported."
-    )
-    reward_model_enabled = (
-        "env_name" in data_config and data_config["env_name"] == "reward_model"
-    )
+    backend = generation_config["backend"]
+    if not colocated_inference:
+        assert backend == "vllm", (
```
yuki-97: `assert backend == "vllm"` gets stripped under `python -O`, and diverges from grpo.py's `raise NotImplementedError` for a semantically identical case (SGLang not-yet-implementing the cross-cluster refit path).

Suggest converting so the guard survives optimized runs and matches the sibling-backend rejection style:
```suggestion
    if not colocated_inference and backend != "vllm":
        raise NotImplementedError(
            "Non-colocated PPO generation currently supports only vLLM; "
            f"got backend={backend!r}. SGLang does not yet implement the "
            "cross-cluster collective weight update path."
        )
```

nemo_rl/data/datasets/utils.py
```diff
@@ -302,14 +302,13 @@ def extract_necessary_env_names(data_config: dict) -> list[str]:
...
     necessary_env_names = set()
-    keys = ["train", "validation", "default"]
-    for key in keys:
-        if (
-            key in data_config
-            and data_config[key] is not None
-            and "env_name" in data_config[key]
-        ):
-            necessary_env_names.add(data_config[key]["env_name"])
+    for key in ("train", "validation", "default"):
```
yuki-97, line 304: thanks for adding the fix here, and could you help to mention this in PR description?

tests/functional/L1_Functional_Tests_PPO.sh
```diff
@@ -36,6 +36,8 @@ run_test() {
 
 run_test fast uv run --no-sync bash ./tests/functional/ppo_automodel.sh
 run_test fast uv run --no-sync bash ./tests/functional/ppo_megatron.sh
+run_test      uv run --no-sync bash ./tests/functional/ppo_non_colocated.sh
+run_test      uv run --no-sync bash ./tests/functional/ppo_megatron_non_colocated.sh
```
yuki-97: let's mark fast to let `CI:Lfast` cover it.
```suggestion
run_test fast uv run --no-sync bash ./tests/functional/ppo_non_colocated.sh
run_test fast uv run --no-sync bash ./tests/functional/ppo_megatron_non_colocated.sh
```

nemo_rl/algorithms/ppo.py
yuki-97: #3185 is still using `train_data["token_mask"]`, could you check with @HeyyyyyyG to see which one should be used?

yuki-97 (commented):
Overall LGTM on the non-colocated surface after auditing offload/onload sequencing and node allocation against the GRPO reference on `main`.

**Offload/onload — parity verified.** All 3 refit sites (`ppo.py:1163-1170` initial-val, `1227-1269` loop, `1574-1587` val-in-loop) correctly add an explicit `if not colocated_inference: policy.offload_to_cpu()` after `refit_policy_generation` returns, mirroring the fact that the shared refit helper intentionally skips policy offload in non-colocated mode (weights transfer via NCCL, vLLM stays resident). Value ↔ policy time-share the training GPUs via explicit `finish_training` / `finish_inference` boundaries — no double-resident window. Checkpoint save at `ppo.py:1796-1831` does the same policy → save → offload → value → save dance.

**Node allocation — parity verified.** RayVirtualCluster construction, all 4 assert guards (train-gpus-per-node > 0, inference_nodes explicit for multi-node, gpus_per_node full-node for multi-node, nodes_per_instance divisibility), and `cluster.segment_size` handling all mirror `grpo.py` on `main`, with two intentional PPO-specific divergences:
1. `max_colocated_worker_groups=2` on `train_cluster` — necessary since PPO hosts both `policy` and `value_model` worker groups on the same train cluster (GRPO has policy only, uses `=1`).
2. `required_nodes` for the segment_size check uses `train_nodes` when shared-node non-colocated (single node hosts both train and inference), `train_nodes + inference_nodes` otherwise. Stricter-correct than GRPO's always-sum for the rare single-node + segment_size combo.

Comments below are the remaining findings.

nemo_rl/algorithms/ppo.py
bg51717: Confirmed with him that filtered samples should be masked out of GAE. I kept the combined `token_mask * sample_mask`, aligned the implementation with GRPO, and will document that this also changes existing `overlong_filtering=true` PPO numerics when the sequence-level threshold is unset.

tests/unit/algorithms/test_ppo.py
bg51717: Thanks, I added a targeted unit test for each of the four uncovered branches you listed, covering the expected assertion or warning/fallback behavior. All four tests pass.

tests/unit/algorithms/test_ppo.py
bg51717: Thanks, I clarified in the unit-test docstring that it covers wiring only. I kept the real cross-cluster `init_collective` coverage in the AutoModel and Megatron non-colocated functional tests. Both functional tests passed, and I marked them as `fast` so they are included in `CI:Lfast`.

nemo_rl/algorithms/ppo.py
bg51717: Thanks, I replaced the assertion with `NotImplementedError` so the guard remains active under `python -O`, and updated the corresponding unit test. The test passes.

nemo_rl/data/datasets/utils.py
bg51717: OK. I will do this.

tests/functional/L1_Functional_Tests_PPO.sh
bg51717: Thanks, I marked both non-colocated PPO functional tests as `fast`. This adds the real cross-cluster path to `CI:Lfast`, and both tests pass.

bg51717:
/ok to test [e4e64e8](https://github.com/NVIDIA-NeMo/RL/pull/3262/commits/e4e64e8722adf37d82ec4643f14d3aef3ef0b48e)

bg51717:
/ok to test [08c888c](https://github.com/NVIDIA-NeMo/RL/pull/3262/commits/08c888c73bd632eb89394564ef81d035875db550)

bg51717:
/ok to test [fbcfbe3](https://github.com/NVIDIA-NeMo/RL/pull/3262/commits/fbcfbe327da6c58c3b5d95881c1f54b6a4c16cc6)

nemo_rl/algorithms/ppo.py
```diff
@@ -420,39 +456,185 @@ def setup(
...
+            )
+            node_resource_constraints, remaining_node_ids, topology = (
+                prepare_segment_topology(
+                    segment_size,
+                    train_nodes,
+                    topology=topology,
+                    role="training",
+                )
+            )
+            if node_resource_constraints is not None:
```
yuki-97, line 565: This segment_size + topology block duplicates ~55 lines of `grpo.py:770-862` and drops GRPO's `⚠ N training nodes have NVLink domain info but no topo_rank` diagnostic (grpo.py:798-812).

Two copies will drift: a future non-vLLM PPO backend enablement would silently miss the per-backend `gpus_per_instance` branch GRPO already has for trtllm/sglang (grpo.py:820-834), and non-colocated PPO users on mixed-topo clusters lose the placement-suboptimal warning today.

**Suggest fix:** extract a shared helper (e.g. in `nemo_rl/distributed/virtual_cluster.py` alongside `prepare_segment_topology`) called by both algorithms; at minimum, port the topo_rank warning here.

yuki-97: let's copy the dropped things to here.

the shared helper is not a blocker of this PR but feels worth to do later. could unify grpo / ppo / distillation / single-contoller for this.
cc @terrykong do you know if someone is working one this? if not @bg51717 do you mind helping to cleanup this later?

yuki-97 (commented):
overall LGTM, just one small comment. @terrykong could you take a review as well?

nemo_rl/algorithms/ppo.py
bg51717: I could help that. You could open an issue and assign it to me.

yuki-97: thanks @bg51717 , tracked at https://github.com/NVIDIA-NeMo/RL/issues/3464 and assigned to you.

> This segment_size + topology block duplicates ~55 lines of grpo.py:770-862 and drops GRPO's ⚠ N training nodes have NVLink domain info but no topo_rank diagnostic (grpo.py:798-812).

for this PR, could you help adding the dropped things?

bg51717:
/ok to test [6f176b3](https://github.com/NVIDIA-NeMo/RL/pull/3262/commits/6f176b3b4899325c20022630b27b2973d8273544)

nemo_rl/algorithms/ppo.py
bg51717: I added back the missing `no-topo_rank` warning and `gpus_per_instance`. I’ll complete a more thorough refactor in a follow-up PR.

nemo_rl/algorithms/ppo.py
```diff
@@ -393,14 +428,16 @@ def setup(
...
-    assert colocated_inference, (
-        "PPO currently requires colocated generation (vLLM / SGLang sharing GPUs "
-        "with the policy worker). Set policy.generation.colocated.enabled=true. "
-        "Non-colocated PPO is not yet supported."
-    )
-    reward_model_enabled = (
-        "env_name" in data_config and data_config["env_name"] == "reward_model"
-    )
+    backend = generation_config["backend"]
+    if not colocated_inference and backend != "vllm":
```
terrykong, line 432: [`ppo.py:432`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/nemo_rl/algorithms/ppo.py#L432)

**1 action item.**

### AI-1 — this block makes the error text at L285-289 false

PR-adjacent: those lines aren't in the diff, but this PR is what invalidates them.

[`ppo.py:284-290`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/nemo_rl/algorithms/ppo.py#L284) still raises:

> Checkpoint-engine refit requires non-colocated generation, but **PPO currently requires colocated generation**. Non-colocated PPO support is tracked in issue #3275.

The guard itself is still correct — PPO genuinely doesn't wire the `nccl_reshard` / `remote_transport` / checkpoint-engine transports that [`grpo.py:1373-1409`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/nemo_rl/algorithms/grpo.py#L1373) branches on, so rejecting is right. Only the stated *reason* is now false. A user who sets `refit_transport=nccl_reshard` **and** `colocated.enabled=false` is told to satisfy a condition they already satisfied, and pointed at an issue this PR closes the premise of.

**Action:** reword the `ValueError` in `setup` (L284-290) — one way to do it:

```python
        if refit_transport is not None:
            raise ValueError(
                f"policy.generation.refit_transport={refit_transport!r} is not yet "
                "supported by PPO (colocated or non-colocated); PPO refits over the "
                "default collective path. Set policy.generation.refit_transport=null. "
                "Tracked in https://github.com/NVIDIA-NeMo/RL/issues/3275."
            )
```

nemo_rl/algorithms/ppo.py
```diff
@@ -393,14 +428,16 @@ def setup(
...
+        raise NotImplementedError(
+            "Non-colocated PPO generation currently supports only vLLM; "
+            f"got backend={backend!r}. SGLang does not yet implement the "
+            "cross-cluster collective weight update path."
+        )
+
+    reward_model_enabled = "reward_model" in extract_necessary_env_names(data_config)
+    segment_size = cluster_config.get("segment_size")
```
terrykong, line 440: [`ppo.py:440`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/nemo_rl/algorithms/ppo.py#L440)

### AI-1 — three `cluster.*` keys PPO now reads aren't in the exemplar

PR-introduced. Class: config-conventions / discoverability.

This line makes `ppo.py` read `cluster.segment_size` for the first time, and the new non-colocated clusters also read `cluster.master_port_range_low` / `_high`. None of the three appear in `examples/configs/ppo_math_1B.yaml`'s `cluster:` block, which has only `gpus_per_node` and `num_nodes`. [`grpo_math_1B.yaml:486-494`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/examples/configs/grpo_math_1B.yaml#L486) documents all three with comments.

Since the exemplar YAML is the source of truth for defaults, PPO's new topology support is currently undiscoverable — a user has to read `setup()` to learn the knob exists.

**Action:** add to the `cluster:` block of `examples/configs/ppo_math_1B.yaml` (not in this diff, so as prose):

```yaml
cluster:
  gpus_per_node: 1
  num_nodes: 1
  # Port range for the distributed master address (TCPStore / NCCL rendezvous)
  # and per-worker available ports. Kept below the OS ephemeral range.
  master_port_range_low: 1400
  master_port_range_high: 1999
  segment_size: null  # Nodes per NVLink domain segment for topology-aware alignment; null to disable
```

### Follow-up — not for this PR

`segment_size` and the port range are threaded into the two non-colocated clusters but not into the colocated one at [L471-477](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/nemo_rl/algorithms/ppo.py#L471), where `grpo.py:700-713` passes all four. So after this PR the same key is honored or silently ignored depending on `colocated.enabled`. Wiring it up means changing placement on PPO's default, most-used path, which this PR neither touches nor tests — so it belongs with the shared grpo/ppo cluster-setup cleanup in #3464 rather than here. Flagging only so it's on the record.

tests/unit/algorithms/test_ppo.py
```diff
@@ -693,3 +698,1041 @@ def test_create_advantage_estimator_requires_adv_estimator_key():
...
+            "policy.generation.colocated.resources.gpus_per_node must be "
+            "explicitly set and equal to cluster.gpus_per_node"
+        ),
+    ):
+        ppo_mod.setup(config, MagicMock(), _setup_dataset(), None)
+
+    cluster_cls.assert_not_called()
+
+
+def test_noncolocated_inference_requires_full_node_gpus_multi_node(monkeypatch):
```
terrykong, line 1464: [`test_ppo.py:1406`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/tests/unit/algorithms/test_ppo.py#L1406)

**2 action items.**

_Context so this doesn't read as re-opening a settled thread: the four branches from the earlier review are all genuinely covered now — I checked each one hits its intended assert rather than a sibling. These are **two different asserts that the earlier pass never named**, both added by this PR._

### AI-1 — `inference_nodes` must be 1 on a single node ([`ppo.py:501`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/nemo_rl/algorithms/ppo.py#L501))

Every single-node test leaves `inference_nodes=None`, so the `== 1` half never executes. Triggering config: `cluster.num_nodes=1, gpus_per_node=8` with `resources={num_nodes: 2, gpus_per_node: 2}` — the `gpus_per_node` check at L494 passes first, so L501 is what actually stops it.

### AI-2 — `train_nodes > 0` ([`ppo.py:535`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/nemo_rl/algorithms/ppo.py#L535))

The only guard catching rollout consuming every node. Triggering config: `cluster.num_nodes=2, gpus_per_node=8` with `resources={num_nodes: 2, gpus_per_node: 8}` → `train_nodes = 2-2 = 0`, with L519 (`2 > 0`) and L523 (`8 == 8`) both passing first. Without it you build a `train_cluster` with `bundle_ct_per_node_list=[]` and world size 0. Note this is only reachable through the multi-node `else` branch — in the shared-node branch `train_nodes` and `inference_nodes` are both fixed at 1.

**Action:** add one test per assert, mirroring `test_noncolocated_inference_requires_full_node_gpus_multi_node`:

```python
def test_noncolocated_inference_rejects_multiple_nodes_single_node(monkeypatch):
    """A one-node GPU split cannot span more than one inference node."""
    ppo_mod = _patch_ppo_setup_prerequisites(monkeypatch)
    cluster_cls = MagicMock()
    monkeypatch.setattr(ppo_mod, "RayVirtualCluster", cluster_cls)
    config = _make_noncolocated_setup_config(
        total_nodes=1, total_gpus_per_node=8,
        inference_nodes=2, inference_gpus_per_node=2,
    )
    with pytest.raises(
        AssertionError,
        match="policy.generation.colocated.resources.num_nodes must be 1 or set to null",
    ):
        ppo_mod.setup(config, MagicMock(), _setup_dataset(), None)
    cluster_cls.assert_not_called()


def test_noncolocated_inference_must_leave_training_nodes_multi_node(monkeypatch):
    """Rollout cannot consume every node in a multi-node PPO allocation."""
    ppo_mod = _patch_ppo_setup_prerequisites(monkeypatch)
    cluster_cls = MagicMock()
    monkeypatch.setattr(ppo_mod, "RayVirtualCluster", cluster_cls)
    config = _make_noncolocated_setup_config(
        total_nodes=2, total_gpus_per_node=8,
        inference_nodes=2, inference_gpus_per_node=8,
    )
    with pytest.raises(
        AssertionError,
        match="Non-colocated PPO requires both training and inference resources",
    ):
        ppo_mod.setup(config, MagicMock(), _setup_dataset(), None)
    cluster_cls.assert_not_called()
```

_Both are syntax-checked and the assert ordering traced by hand, but **not executed** — the lockfile is Linux-only and this review ran on macOS, so no pytest run was possible._

docs/guides/ppo.md
```diff
@@ -279,6 +297,9 @@ PPO logs all the same metrics as GRPO (see [GRPO Metrics](grpo.md#metrics)). In
 | `critic/values_max` | Maximum predicted value |
 | `critic/returns_mean` | Mean of GAE returns |
 | `critic/explained_var` | Explained variance: $1 - \text{Var}(R - V) / \text{Var}(R)$. Higher is better; values near 1.0 indicate the critic accurately predicts returns. |
+| `max_seq_mult_prob_error` | Maximum sequence-level multiplicative probability error between generation and training logprobs before optional masking. |
+| `mean_seq_mult_prob_error` | Mean sequence-level multiplicative probability error before optional masking. |
+| `num_masked_seqs_by_logprob_error` | Number of sequences excluded by `ppo.seq_logprob_error_threshold`. |
```
terrykong, line 306: [`ppo.md:300-302`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/docs/guides/ppo.md#L300)

**1 action item.**

### AI-1 — the table lists 3 of the 8 metrics that are actually logged

PR-introduced doc gap.

`compute_and_apply_seq_logprob_error_masking` returns 8 keys and all reach the logger via `metrics.update(seq_logprob_error_metrics)` at [`ppo.py:1747`](https://github.com/NVIDIA-NeMo/RL/blob/6f176b3b4899325c20022630b27b2973d8273544/nemo_rl/algorithms/ppo.py#L1747). The four omitted ones are the feedback loop for the knob this PR adds: you set `seq_logprob_error_threshold`, then read `*_after_mask` to see whether the tail actually dropped below it, and `masked_correct_pct` to see whether you're discarding *correct* answers. Documenting the three easy keys while omitting the four that make the knob tunable is the gap worth closing.

**Action:**

```suggestion
| `max_seq_mult_prob_error` | Maximum sequence-level multiplicative probability error between generation and training logprobs before optional masking. |
| `mean_seq_mult_prob_error` | Mean sequence-level multiplicative probability error before optional masking. |
| `min_seq_mult_prob_error` | Minimum sequence-level multiplicative probability error before optional masking. |
| `max_seq_mult_prob_error_after_mask` | Maximum sequence-level error over the sequences kept by `ppo.seq_logprob_error_threshold`. Equal to the pre-mask value when the threshold is `null`. |
| `mean_seq_mult_prob_error_after_mask` | Mean sequence-level error over the kept sequences. |
| `min_seq_mult_prob_error_after_mask` | Minimum sequence-level error over the kept sequences. |
| `num_masked_seqs_by_logprob_error` | Number of sequences excluded by `ppo.seq_logprob_error_threshold`. |
| `masked_correct_pct` | Fraction of excluded sequences that had reward = 1. A high value means the threshold is discarding good samples. |
```

terrykong (commented):
Re-review at `6f176b3b4`. Six agents (RL/domain, bug-finder, test, design, comment, adversarial) went over this.

**All 7 prior review threads are fixed.** I verified each against the code at head rather than on the replies — including that the four assert branches each hit their intended branch (not a sibling), and that the topology port-back in `6f176b3b4` is faithful to `grpo.py:816-878`. Replies are on the individual threads.

**No correctness bug was found in the new non-colocated path.** Three agents diffed it line-by-line against `grpo.py`: the GPU-resource arithmetic (including the reward-model x node-count boundaries), the policy/value offload coordination, and the `token_mask * sample_mask` change all came back clean — every mask consumer is consistent, with no sibling left on the old mask, and PPO is actually *more* consistent here than GRPO, which still uses bare `token_loss_mask` for its own `response_advantages` metric. The `required_nodes` shared-node refinement is a genuine improvement over grpo's version. Nice work; the five comments below are all low-severity.

Everything here is **static analysis only**. `pre-commit` and `pytest` could not be run (the lockfile supports Linux `x86_64`/`aarch64` only and this review ran on macOS), so nothing below is a passing result and the suggested tests are syntax-checked, not executed.

Twelve candidate findings were dropped during the adversarial pass, several of which are worth naming so they don't get re-raised later:
- A proposed `.get()` change to the `seq_logprob_error_threshold` read — PPO's bare subscript is convention-correct for a required TypedDict field, and `.get(key, default)` is the forbidden hidden-default pattern.
- A suggestion to delete the unreachable trtllm/sglang branches in the topology block — those were added at a reviewer's explicit request in the previous round.
- A `gpu_memory_utilization: 0.4` concern, withdrawn once the right peer group was checked: 0.4 is what every 1.5B PPO recipe uses, colocated included, so the new recipes follow house style rather than diverging from it.
- The new `RuntimeError` on a fully-masked batch: it does change PPO's prior behavior (previously a benign zero-gradient no-op — `masked_mean` divides by `sum(mask) + 1e-8`, so there was never a NaN), but firing it needs all 1024 samples in a step truncated simultaneously, and the 40-step runs on both recipes are empirical evidence it doesn't happen. Fail-loud there is a defensible choice.

Two design observations belong with the already-tracked #3464 rather than this PR, and I'll add them there: the non-colocated layout arithmetic is inlined in `setup()`, which is why ~450 of the 1043 new test lines are Ray mock scaffolding (a pure `plan_noncolocated_layout(...)` helper would make those validation tests ~3 lines each); and the post-refit `policy.offload_to_cpu()` is now a caller obligation duplicated at three sites.

_Generated by Claude Code_

nemo_rl/algorithms/ppo.py
terrykong: Documented in the PR description under "Related correctness fixes → Filtered-sample GAE semantics".

tests/unit/algorithms/test_ppo.py
terrykong: fixed in e4e64e8722adf37d82ec4643f14d3aef3ef0b48e

tests/unit/algorithms/test_ppo.py
terrykong: fixed in e4e64e8722adf37d82ec4643f14d3aef3ef0b48e

nemo_rl/algorithms/ppo.py
terrykong: fixed in e4e64e8722adf37d82ec4643f14d3aef3ef0b48e

nemo_rl/data/datasets/utils.py
terrykong: Mentioned in the PR description under "Related correctness fixes → Reward-model resource detection for structured data configs".

tests/functional/L1_Functional_Tests_PPO.sh
terrykong: fixed in e4e64e8722adf37d82ec4643f14d3aef3ef0b48e

nemo_rl/algorithms/ppo.py
terrykong: fixed in 6f176b3b4899325c20022630b27b2973d8273544

nemo_rl/algorithms/ppo.py
bg51717: Good catch. I updated the error message to clarify that explicit refit transports are not supported by PPO yet and that PPO uses the default collective path.

nemo_rl/algorithms/ppo.py
bg51717: Added the three cluster fields to the PPO exemplar config and synchronized the reference config. The colocated-path follow-up will remain tracked in #3464.

tests/unit/algorithms/test_ppo.py
bg51717: Added both targeted tests. They cover the single-node inference_nodes constraint and the requirement to leave training nodes in the multi-node layout.

docs/guides/ppo.md
bg51717: Updated the PPO metrics table to document all eight sequence-level logprob error metrics.

yuki-97 (approved):
LGTM, thanks @bg51717 !

PR #3263, opened 2026-07-17
feat(sc): abort stale in-flight rollouts early in non Gym path

RayenTian:
## Summary

- Support aborting stale in-flight prompt groups when they fall outside the sampler selection window.
- Add W&B metrics for evicted stale prompt groups (`train/evicted_stale_prompt_groups`) and aborted stale in-flight groups (`train/aborted_stale_inflight_groups`).
- Fix the legacy GRPO tokenizer decode error by explicitly disabling tokenization-space cleanup.

## Tests
Recipe: [examples/configs/recipes/llm/performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml](https://github.com/NVIDIA-NeMo/RL/blob/55296257c96d49cd95c7d77613cb0f36bd4a4dc7/examples/configs/recipes/llm/performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml#L4)
### disable kv cache recompute
image

### enable kv cache recompute
image


RayenTian:
/ok to test [cc02515](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/cc02515a3fa1ee0d6e0e0bc092b80367ddac6479)

nemo_rl/experience/rollouts.py
```diff
@@ -330,7 +330,11 @@ def generate_responses(
         generated_part = full_output[input_len:total_length]
         generated_ids.append(generated_part)
 
-    generated_texts = tokenizer.batch_decode(generated_ids, skip_special_tokens=True)
+    generated_texts = tokenizer.batch_decode(
+        generated_ids,
+        skip_special_tokens=True,
+        clean_up_tokenization_spaces=False,
```
RayenTian: If not set `clean_up_tokenization_spaces=False`, we'll meet ` Error generating response for sample 1864: TokenizersBackend has no attribute clean_up_tokenization_spaces_for_bpe_even_though_it_will_corrupt_output` in main `run_grpo.py` path(not single controller path), further lead to reward & grad_norm = 0


Some Transformers/tokenizer combinations enable cleanup by default when `clean_up_tokenization_spaces `is not explicitly set. Explicitly setting `clean_up_tokenization_spaces=False` disables that cleanup path and avoids accessing the missing attribute.

RayenTian:
/ok to test [728de3f](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/728de3fcbfa61ff433522b68613574410dd72596)

RayenTian:
/ok to test [2019c54](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/2019c54292467e9daec2378db6ee3114e44fef81)

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -1167,17 +1167,23 @@ async def process_single_sample(sample_idx):
...
+                try:
+                    result = await completed_task
+                    yield result
+                except Exception as e:
+                    # Cancel remaining tasks
+                    for task in sample_tasks:
+                        if not task.done():
+                            task.cancel()
+                    await asyncio.gather(*sample_tasks, return_exceptions=True)
+                    raise e
```
yuki-97: After the new outer `try/finally` at :1170-1186 covers cancel-on-exit, this inner `except Exception` cancel/gather block (:1175-1181) is redundant — it cancels + gathers the same `sample_tasks` the outer `finally` already handles.

Suggest fix: drop the inner cancel loop + gather and just `raise` (or `raise e`); the `finally` covers success, exception, and cancellation paths uniformly. Same duplication at :1272-1281 (token API).

tests/unit/experience/test_rollouts.py
```diff
@@ -141,7 +141,12 @@ def test_median_like_p50(self):
...
     pad_token_id = 0
 
-    def batch_decode(self, generated_ids, skip_special_tokens=True):
+    def batch_decode(
+        self,
+        generated_ids,
+        skip_special_tokens=True,
+        clean_up_tokenization_spaces=False,
+    ):
         return ["ok" for _ in generated_ids]
```
yuki-97, line 145: **[TEST]** `_DummyTokenizer.batch_decode` accepts `clean_up_tokenization_spaces=False` as a *default kwarg* but the body never asserts the caller actually passed `False` — a regression at `rollouts.py:333` / `:478` that drops the kwarg would still pass this test.

Mirror `test_rollout_manager_router_replay.py:48` which does `assert clean_up_tokenization_spaces is False` explicitly.

tests/unit/single_controller/test_train_pump.py
```diff
@@ -363,6 +363,8 @@ def test_train_pump_drives_mcore_training_step(
         for metrics in train_metrics:
             assert math.isfinite(metrics["reward"])
             assert math.isfinite(metrics["advantages/mean"])
+            assert metrics["evicted_stale_prompt_groups"] == 0
+            assert metrics["aborted_stale_inflight_groups"] == 0
```
yuki-97, line 374: **[TEST]** Trivially-zero assertion: the new metric keys are checked only equal to `0` — presence check without exercising the non-zero path. A regression that drops the `evicted_stale_prompt_groups += evicted` accumulation at `single_controller.py:386` or the metric assignment at `:530-537` would still pass here.

Suggest a fixture that seeds `_inflight_by_group_id` with a stale task (so `_abort_stale_inflight` cancels ≥1) or stubs `_sampler.evict` to return >0, so the non-zero path is exercised.

nemo_rl/experience/rollouts.py
yuki-97: **[BUG]** Sibling decode sites still missing this `clean_up_tokenization_spaces=False` fix:

- `nemo_rl/environments/nemo_gym.py:469, 472` — `batch_decode` of generated tokens for JSONL logging.
- `nemo_rl/models/generation/trtllm/trtllm_http_server.py:248` — `tokenizer.decode(gen_token_ids, skip_special_tokens=False)` on the client return path (result then passed to the reasoning parser + tool-call parser).

If the fix here + at `rollout_manager.py:309` + at `:478` was needed to unblock a specific `TokenizersBackend` error (per the author's comment on r3694869704), these sibling sites will hit the same error on the same tokenizers. Suggest adding the flag consistently.

nemo_rl/algorithms/single_controller.py
```diff
@@ -573,11 +587,46 @@ async def _train_pump(self) -> None:
...
+            flush=True,
+        )
+        return len(stale_tasks)
+
     async def _sync_weights(
         self,
         *,
         calibration_data: Optional[BatchedDataDict[Any]] = None,
-    ) -> None:
+    ) -> int:
```
yuki-97, line 636: nit: let's add args and return in docstring.

nemo_rl/models/generation/vllm/vllm_generation.py
```diff
@@ -776,34 +776,40 @@ async def _async_generate_base(
             os.environ.get("NRL_VLLM_ASYNC_TIMEOUT_SECONDS", "900")
         )  # Default 15 minutes
 
+        # Propagate cancellation to the Ray worker and its vLLM request.
         try:
```
yuki-97, line 780: curious have you tested whether nemo-gym path works? since the changes in vLLM all seem to be native async path.

yuki-97 (commented):
thanks for supporting this, changes overall LGTM. left some comments and one q for nemo-gym path.

nemo_rl/experience/rollouts.py
RayenTian: Split the tokenizer fix into https://github.com/NVIDIA-NeMo/RL/pull/3486

tests/unit/experience/test_rollouts.py
RayenTian: Split the tokenizer fix into https://github.com/NVIDIA-NeMo/RL/pull/3486

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [43baf36](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/43baf361f67cfc583bdc704312cc9244313b5af4)

tests/unit/single_controller/test_train_pump.py
RayenTian: Done in [43baf36](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/43baf361f67cfc583bdc704312cc9244313b5af4)

nemo_rl/models/generation/vllm/vllm_worker_async.py
RayenTian: Done in [43baf36](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/43baf361f67cfc583bdc704312cc9244313b5af4)

RayenTian:
/ok to test [0510d09](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/0510d09b912bb4a15a0eb7473f8dfdcd5cf46b99)

nemo_rl/models/generation/vllm/vllm_generation.py
RayenTian: Will do in next PR. Trace in Issue[# 2625](https://github.com/NVIDIA-NeMo/RL/issues/2625#issuecomment-5105095632)

parthmannan:
Had a quick question on the graphs (haven't read the changes yet) - Why do we not see any evicted stale groups when not aborting? `staleness_window` would still evict samples even when we don't abort in-flight samples?

RayenTian:
> Had a quick question on the graphs (haven't read the changes yet) - Why do we not see any evicted stale groups when not aborting? `staleness_window` would still evict samples even when we don't abort in-flight samples?

Yes, we also evict groups when we don't  abort in-flight samples. Because the two panels are added in this PR, and no-abort version runs are running on main, so there're no panels for the runs. In theory, in scenarios without abort, the evict count should be the sum of the values from the two current panels.

nemo_rl/algorithms/single_controller.py
```diff
@@ -590,9 +639,19 @@ async def _sync_weights(
...
+            calibration_data: Optional data used to calibrate FP8 KV-cache
+                scales before synchronizing weights.
+
+        Returns:
+            The number of stale in-flight rollout groups aborted before the
+            weight synchronization.
         """
         self._rollout_permitted.clear()
 
+        aborted_stale_inflight_groups = await self._abort_stale_inflight()
```
yuki-97: since gym path isn't validated and may still need some develops, can we only call `_abort_stale_inflight` only for non-gym path for now and also add a TODO here?

RayenTian:
/ok to test [bec1089](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/bec1089e9ec1472ea151e6a1b57275f7e4b8e0c1)

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [bec1089](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/bec1089e9ec1472ea151e6a1b57275f7e4b8e0c1)

RayenTian:
/ok to test [d1a1bbe](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/d1a1bbe726b75ca51f58922618efe008cbe221ec)

RayenTian:
/ok to test [132f9d9](https://github.com/NVIDIA-NeMo/RL/pull/3263/commits/132f9d918a286668016765ecd774220abaf93507)

yuki-97 (approved):

PR #3266, opened 2026-07-17
feat(sc): setup + entrypoint

yuki-97:
Part 3/4 of the #2819 split. See #2819 (reference) or #3267 (4/4) for the full code and context. Stack on #3220.

**Note**: NeMo-Gym support goes in PR#4; this PR wires up the non-gym SC path.

## Summary

Driver-side setup and CLI entrypoint that make the SC (Single Controller) async-GRPO path end-to-end runnable.

- **`single_controller_utils` package** (`setup.py` / `config.py` / `utils.py`): `setup_single_controller` builds the worker groups, virtual cluster, dp client, dataloader, TQReplayBuffer, RolloutManager, and weight synchronizer on the driver and returns a `SingleControllerActorArgs`. `config.py` centralizes `AsyncRLConfig` / `AdvantageConfig` / `WeightSyncConfig` / `MasterConfig`. `utils.py` hosts metric-reduction rules and `KVBatchMeta` helpers.
- **`SingleControllerActor` slimdown**: takes a prebuilt `SingleControllerActorArgs` and does no construction work — heavy objects are cloudpickled in. Running setup inside a nested Ray actor breaks runtime_env resolution (see #2692). Pump crashes propagate and in-flight rollouts drain on exit.
- **`AsyncRLConfig`**: dropped `batch_selection_strategy` in favor of explicit `max_staleness_versions` + `over_sampling`; renamed `min_prompt_groups_per_batch` → `min_groups_for_streaming_train`; defaults aligned with the exemplar and each field has a doc comment.
- **`examples/run_grpo_single_controller.py`**: SC CLI entrypoint, mirrors `run_grpo.py` for config loading. Requires `data_plane.enabled=true`. Drains env actors on the error path before vLLM shutdown.
- **Recipes**: `grpo_math_1B_megatron_single_controller.yaml` exemplar plus streaming2 and megatron-sync variants under `examples/configs/recipes/llm/`.
- **CI**: adds `L1_Functional_Tests_SingleController.sh` to `cicd-main.yml`; adds SC recipes to `nightly.txt` and paired `*-single-controller-{streaming2,sync}.sh` launchers.
- **Tests**: `test_single_controller_setup.py` covers the new setup module (with an `init_communicator()` assertion). New `tests/functional/grpo_dp_single_controller.sh` gives the entrypoint a 2-GPU e2e smoke. Rollout/train pump tests migrated to the split configs.

## Test Results
1. **sync mode**: `grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync`
    baseline (DTensor + colocated), SC (Megatron + non-colocated), so only compare convergence.
    image
3. **async mode**: `grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2`
    SC run uses `min_groups_for_streaming_train=num_prompts_per_step`, `over_sampling=false`, `force_in_order=true` (different from the script in this PR) to keep using the same setting as legacy code for compare.
    - yellow: baseline (old code path), didn't enable validate and checkpointing to keep the same test settings.
    - blue: this PR #3266, `grad_norm` is ~1/2x smaller and reward is a bit lower.
    - red: fix grad_norm in #3282. (see more details there)
    image
    image

## Known Issues (From the above tests results)
1. `grad_norm` is ~1/2x smaller: fixed in #3282.
2. `invalidate_kv_cache` only enabled in sync mode, which will cause SC async mode never reset vLLM's prefix cache, tracked in "Known Missing Features" below and #2625.
3. `total_step_time` is longer (since `policy_and_reference_logprobs` is longer), it's because SC's `prev_logprobs_required` / `reference_logprobs_required` gating is provisional and didn't skip calculating `prev_logprobs` in this run, already tracked in #2625.

## Known Missing Features (Tracked in #2625 or have PRs)
1. Checkpointing not wired: https://github.com/NVIDIA-NeMo/RL/pull/3138, currently landed in 2819, need to merge into main after this PR (3266) go in.
2. Validation not wired: https://github.com/NVIDIA-NeMo/RL/pull/3203, currently landed in 2819, need to merge into main after this PR (3266) go in.
4. Abort stale in-flight rollouts in refit instead of evicting in train: https://github.com/NVIDIA-NeMo/RL/pull/3263, need this PR (3266) get into main first.
5. Drain gate in refit is not supported yet.
6. `invalidate_kv_cache` only enabled in sync mode (`max_staleness_versions=1`), need to respect to `recompute_kv_cache_after_weight_updates`.


terrykong:
The `async_rl` knob set this PR exposes ([`over_sampling`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/configs/grpo_math_1B_megatron_single_controller.yaml), `force_in_order`, `max_staleness_versions`, `min_groups_for_streaming_train`, `max_buffered_rollouts`) is powerful, but figuring out which *combination* yields which async-RL behavior currently requires piecing together comments scattered across the three exemplar YAMLs plus the usage table in #3219's description — and some combinations are invalid (`force_in_order=true` + `over_sampling=true`) or silently ignored (`force_in_order` in sync mode).

Could this PR (or an immediate follow-up in the stack) add a user-facing markdown that maps configurations to knob settings? A natural home is a "Single-controller async GRPO" section in [`docs/guides/async-grpo.md`](https://github.com/NVIDIA-NeMo/RL/blob/main/docs/guides/async-grpo.md) (or a new guide wired into `docs/index.md`). Suggested skeleton, seeded from this PR's own exemplars:

| Mode | `max_staleness_versions` | `over_sampling` | `force_in_order` | `min_groups_for_streaming_train` | Exemplar config |
|---|---|---|---|---|---|
| Sync / on-policy | 0 | false | true (ignored in sync mode) | `num_prompts_per_step` | [`...-single-controller-sync.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync.yaml) |
| Async, exact batch→step matching (legacy `target_weight` semantics) | ≥1 | false | true | `num_prompts_per_step` | [`grpo_math_1B_megatron_single_controller.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/configs/grpo_math_1B_megatron_single_controller.yaml) |
| Streaming, gated dispatch ("streaming 1" in #3219) | ≥1 | false | false | x ≤ `num_prompts_per_step` | — |
| Streaming, over-sampled ("streaming 2") | ≥1 | true | false | x ≤ `num_prompts_per_step` | [`...-async-1off-single-controller-streaming2.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml) |

…plus the invariants that today live only in YAML comments and setup-time raises: `over_sampling=false` requires `max_buffered_rollouts == num_prompts_per_step × (max_staleness_versions + 1)`; `force_in_order=true` requires `over_sampling=false`; sync mode (`max_staleness_versions: 0`) rejects `over_sampling=true` and ignores `force_in_order`; over-sampled data aged past the staleness window is evicted (wasted compute — see the `over_sampling_ratio` follow-up tracked in #2625).

Asking here rather than on #3219/#3220 because this is the PR where the knobs get their final names and become user-settable (the `async_rl` block + entrypoint), and the three exemplar YAMLs the table should reference land here.

_Generated by Claude Code_

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -0,0 +1,400 @@
...
+    if backend == "vllm":
+        generation_config["vllm_kwargs"]["hf_overrides"] = master_config.policy.get(
+            "hf_config_overrides", {}
+        )
+        gen = VllmGeneration(cluster=inference_cluster, config=generation_config)
+    elif backend == "sglang":
+        generation_config["sglang_cfg"].setdefault(
+            "model_path", master_config.policy["model_name"]
+        )
+        gen = SGLangGeneration(cluster=inference_cluster, config=generation_config)
```
terrykong: [`setup.py:157`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller_utils/setup.py#L157)

[`SGLangGeneration.__init__`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/models/generation/sglang/sglang_generation.py#L70-L74) has no `config` parameter — its signature is `(self, cluster: RayVirtualCluster, sglang_cfg: SGLangConfig)`, so this call raises `TypeError` (unexpected keyword `config`, missing `sglang_cfg`) whenever `backend == "sglang"`. The canonical call in [`grpo.py:974-977`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/grpo.py#L974-L977) uses `sglang_cfg=generation_config`.

**Action:** pass `sglang_cfg=`:

```suggestion
        gen = SGLangGeneration(cluster=inference_cluster, sglang_cfg=generation_config)
```

tests/functional/L1_Functional_Tests_SingleController.sh
```diff
@@ -0,0 +1,43 @@
...
+        time "$@"
+    elif [[ "${FAST:-0}" == "1" ]]; then
+        echo "FAST: Skipping: $*"
+    else
+        time "$@"
+    fi
+}
+
+run_test fast uv run --no-sync bash ./tests/functional/grpo_dp_single_controller.sh
+run_test fast uv run --no-sync bash ./tests/functional/grpo_async_gym_single_controller.sh
```
terrykong: [`L1_Functional_Tests_SingleController.sh:38`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/tests/functional/L1_Functional_Tests_SingleController.sh#L38)

`grpo_async_gym_single_controller.sh` doesn't exist in this PR — it lands with the gym path in #3267. Since this PR also registers `L1_Functional_Tests_SingleController` in [`cicd-main.yml`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/.github/workflows/cicd-main.yml#L720) and `run_test fast` executes in both modes under `set -xeuo pipefail`, the missing file exits 127 and keeps this L1 job red until #3267 merges.

**Action:** drop this line here and let #3267 add it together with the script:

```suggestion
```

nemo_rl/algorithms/single_controller.py
```diff
@@ -338,49 +191,47 @@ def __init__(
...
-            f"{cfg.max_weight_staleness_versions} "
-            f"buffer={cfg.max_buffered_rollouts} "
-            f"inflight={cfg.max_inflight_prompts} "
-            f"transport={cfg.refit_cfg.impl}",
+            f"SingleControllerActor: "
+            f"staleness_cap={self._async_cfg.max_staleness_versions} "
+            f"buffer={self._async_cfg.max_buffered_rollouts} "
+            f"inflight={self._async_cfg.max_inflight_prompts} "
+            f"over_sampling={self._async_cfg.over_sampling} "
+            f"transport={self._weight_sync_cfg.transport}",
```
terrykong: [`single_controller.py:223`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller.py#L223)

This is a **diagnostics bug** — the banner's `transport=` field never reflects the real transport. It reads `self._weight_sync_cfg.transport`, where [`self._weight_sync_cfg = WeightSyncConfig()`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller.py#L95) is constructed with hardcoded defaults and never populated, and [`WeightSyncConfig.transport` defaults to the literal `"stub"`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/config.py#L83). The *actual* weight synchronizer is a different object — [`self._weight_synchronizer`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller.py#L103) — chosen by [`create_weight_synchronizer`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/setup.py#L352) from backend/colocated (NCCL, IPC, …).

So the startup banner **always prints `transport=stub`**, even when real NCCL weight sync is running — anyone reading the log to confirm which weight-transport is active (a normal check when debugging async weight sync) is misled into thinking nothing real is wired. On top of that, `WeightSyncConfig` is entirely dead: `transport` is read only by this line, and `nccl_addr`/`nccl_port` have **zero readers repo-wide** (grep-confirmed).

**Action:** print the real synchronizer's concrete class, and drop the dead [`WeightSyncConfig`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/config.py#L82) (and its export in [`single_controller_utils/__init__.py`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/__init__.py#L33)):

```suggestion
            f"weight_sync={type(self._weight_synchronizer).__name__}",
```

nemo_rl/algorithms/single_controller.py
```diff
@@ -235,100 +82,106 @@ class SingleControllerActor:
...
-            cfg.over_sampling = False
-            print(
-                "Using strict_on_policy, auto setting max_weight_staleness_versions "
-                "to 0 and over_sampling to False.",
-                flush=True,
-            )
-        if cfg.max_weight_staleness_versions == 0 and cfg.over_sampling:
+        num_prompts_per_step = self._master_config.grpo["num_prompts_per_step"]
+        if num_prompts_per_step < self._async_cfg.min_groups_for_streaming_train:
             raise ValueError(
```
terrykong: [`single_controller.py:119-175`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller.py#L119-L175)

These six config invariants run in `SingleControllerActor.__init__` — after [`setup_single_controller`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/examples/run_grpo_single_controller.py#L118) has already built the clusters, vLLM workers, and policy, so a one-character config typo costs the whole multi-minute setup before raising. Keeping a check in `__init__` is reasonable (the actor takes prebuilt, cloudpickled args and shouldn't assume the caller validated them) — the ask is just to *also* fail fast, and to give the branches test coverage (none of the six is currently exercised; the pump tests only build valid configs and `test_single_controller_setup.py` covers `setup()`, not the actor).

**Action:** extract the six checks into a `@staticmethod` (e.g. `SingleControllerActor.validate_async_rl_config(async_cfg, num_prompts_per_step, train_global_batch_size)`), call it from `__init__` (replacing the inline block, so there's a single source of truth that can't drift) **and** at the top of `setup_single_controller()` before `_build_clusters`. That fails fast on misconfig and makes each branch unit-testable directly. If you'd rather it be impossible to forget a call site, a [`@model_validator(mode="after")`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/config.py#L53) on `MasterConfig` (it's a pydantic `BaseModel`) is an even-earlier alternative that fires at config-load time.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -0,0 +1,400 @@
...
+    master_config: MasterConfig, dataloader: StatefulDataLoader
+) -> None:
+    """Set megatron_cfg.train_iters; must run before _build_trainer."""
+    policy_config = master_config.policy
+    if not policy_config.get("megatron_cfg", {}).get("enabled", False):
+        return
+    grpo_config = master_config.grpo
+    policy_config["megatron_cfg"]["train_iters"] = min(
+        grpo_config["max_num_steps"],
+        grpo_config["max_num_epochs"] * len(dataloader),
```
terrykong: [`setup.py:235`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller_utils/setup.py#L235)

`_maybe_inject_megatron_train_iters` computes `grpo_config["max_num_epochs"] * len(dataloader)` with no `None`-guard, while its sibling [`_clamp_max_num_steps`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller_utils/setup.py#L217) returns early on `max_num_epochs is None`. With `megatron_cfg.enabled=true` and `grpo.max_num_epochs: null`, this raises `TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'` during setup. Latent today (shipped recipes set `max_num_epochs: 1`), but the two siblings should handle the same field consistently.

**Action:** mirror the sibling's guard (early-return / assert), and add a `max_num_epochs=None` regression case to `test_single_controller_setup.py` (the existing tests cover the clamp both ways but never the null case).

examples/run_grpo_single_controller.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+
+    init_ray()
+
+    tokenizer = get_tokenizer(config.policy["tokenizer"])
+    assert config.policy["generation"] is not None, (
+        "A generation config is required for SC-driven async GRPO"
+    )
+    has_refit_draft_weights = bool(config.policy["draft"]["enabled"])
+    config.policy["generation"] = configure_generation_config(
```
terrykong, line 113: [`run_grpo_single_controller.py:111`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/run_grpo_single_controller.py#L111)

This `configure_generation_config(...)` call omits the `trains_mtp=` argument that the [`run_grpo.py:116-121`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/run_grpo.py#L116-L121) analog passes (`trains_mtp = bool(megatron_cfg.get("mtp_num_layers"))`); it therefore defaults to `False`. Since the SC path is megatron-centric, a megatron+MTP model would get the wrong vLLM generation config. Latent (the shipped SC config has `mtp_num_layers=0`).

**Action:** mirror `run_grpo.py` — pass `trains_mtp=bool(megatron_cfg.get("mtp_num_layers"))`.

nemo_rl/models/policy/lm_policy.py
```diff
@@ -1141,19 +1141,21 @@ def shutdown(self) -> bool:
...
 
-        This is an extra safety net in case the user forgets to call worker_group.shutdown() and the pointer to
+        This is an extra safety net in case the user forgets to call shutdown() and the pointer to
         the object is lost due to leaving a function scope. It's always recommended that the
-        user calls worker_group.shutdown().
+        user calls shutdown().
         """
-        if hasattr(self, "worker_group"):
-            self.worker_group.shutdown(cleanup_method="shutdown")
+        self.shutdown()
```
terrykong, line 1160: [`lm_policy.py:1158`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/models/policy/lm_policy.py#L1158)

`__del__` changed from `if hasattr(self, "worker_group"): self.worker_group.shutdown(...)` to a bare `self.shutdown()`. If `Policy.__init__` raises before `worker_group` is assigned, GC-time `__del__` → `self.shutdown()` → `self.worker_group` `AttributeError`, now swallowed by `shutdown()`'s generic `except Exception` and surfaced as a misleading `Error during policy shutdown: ... no attribute 'worker_group'`. The old `hasattr` guard prevented that.

**Action:** restore the guard — e.g. early-return in `shutdown()` when `not hasattr(self, "worker_group")`.

examples/run_grpo_single_controller.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    actor_args = setup_single_controller(config, tokenizer)
+
+    print("🚀 Launching SingleControllerActor")
+    sc = SingleControllerActor.remote(master_config=config, actor_args=actor_args)
+    try:
+        result = ray.get(sc.run.remote())
+        print(f"SC run complete: {result}")
+    finally:
```
terrykong, line 127: [`run_grpo_single_controller.py:124-127`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/run_grpo_single_controller.py#L124-L127)

If one env's `shutdown` raises inside this `finally`, it replaces the original `run()` traceback and skips the remaining envs; the generation and trainer worker groups are also never shut down on the error path.

**Action:** make the drain best-effort per env:

```suggestion
        for env_name, handle in actor_args.env_handles.items():
            try:
                ray.get(handle.shutdown.remote())
            except Exception as e:
                print(f"Env {env_name!r} shutdown failed: {e}")
```

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -0,0 +1,400 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
terrykong, line 1: [`setup.py:1`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller_utils/setup.py#L1)

The new `single_controller_utils/{__init__,config,setup}.py` aren't in `pyrefly.toml` `project-includes`, while their siblings [`single_controller.py` and `single_controller_utils/utils.py`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/pyrefly.toml#L55-L56) (from earlier PRs in this stack) are — so the new files silently escape type-checking. Adding them locally surfaces 12 errors: one is the real `SGLangGeneration` bug flagged above; the rest are typing-only mismatches that mirror patterns in the excluded `grpo.py`/`run_grpo.py` (e.g. the 4-tuple unpack at [L307](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller_utils/setup.py#L307) is runtime-safe because `env_configs` is non-None, same as [`run_grpo.py:126`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/run_grpo.py#L126)).

**Action:** opt the three files into `pyrefly.toml` `project-includes` and resolve the typing errors (the `sglang_cfg=` fix clears the real one); `uv run --group dev pre-commit run --all-files` verifies.

tests/functional/grpo_dp_single_controller.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+# Lightweight e2e for examples/run_grpo_single_controller.py — exercises
+# setup_handle + setup_single_controller_component + SingleControllerActor
```
terrykong: [`grpo_dp_single_controller.sh:3`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/tests/functional/grpo_dp_single_controller.sh#L3)

`setup_handle` / `setup_single_controller_component` don't exist — the API this script exercises is [`setup_single_controller`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller_utils/setup.py#L240) + `SingleControllerActor`.

**Action:**

```suggestion
# setup_single_controller + SingleControllerActor
```

tests/test_suites/nightly.txt
```diff
@@ -160,6 +160,10 @@ tests/test_suites/llm/grpo-nanov3-30BA3B-2n8g-fsdp2-tq_mooncake.sh
 tests/test_suites/llm/grpo-qwen3-8B-base-1n8g-fsdp2-lora-tq_mooncake.sh
 tests/test_suites/llm/prorlv2-qwen2.5-math-1.5b-instruct-1n8g-fsdp2tp1.v2-tq_mooncake.sh
 
+# Single Contoller (SC)
```
terrykong: [`nightly.txt:162`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/tests/test_suites/nightly.txt#L162)

```suggestion
# Single Controller (SC)
```

examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml
```diff
@@ -0,0 +1,33 @@
+# SingleController variant of grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml.
```
terrykong, line 1: [`grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml#L1)

The convergence evidence in the description is much appreciated — sync parity and the legacy-matching async comparison (with the #3282 grad_norm fix) cover the two most important modes. One residual gap: the async run was done with `over_sampling=false, force_in_order=true` ("different from the script in this PR"), so the configuration this recipe actually ships — streaming2 (`over_sampling=true`, windowed `strict_weight_fifo=false` selection) — has no curve yet, and that's also the mode where the new `StalenessSampler` window/eviction logic (from #3220) does the most work.

**Action:** when convenient, attach a convergence curve for this recipe as-shipped (streaming2 settings) plus a rough throughput number (tokens/sec or step time) vs the legacy async path — that plus the existing curves would complete the evidence users need before switching. (Moved here from the #3220 review since the stack becomes e2e-runnable in this PR.)

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -0,0 +1,400 @@
...
+    Passed as a single arg to SingleControllerActor.remote so the actor's __init__ does
+    no construction work — every heavy object is cloudpickled in.
+    """
+
+    gen_handle: Any
+    trainer_handle: Any  # driver-side TQPolicy
+    env_handles: dict[str, EnvironmentInterface]
+    train_cluster: RayVirtualCluster
+    inference_cluster: RayVirtualCluster
+    dp_client: Any
```
terrykong: [`setup.py:63`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/setup.py#L63)

`dp_client: Any` can be narrowed to `DataPlaneClient` — that's the declared return type of [`build_data_plane_client(...) -> DataPlaneClient`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/data_plane/factory.py#L21-L23) that populates this field at [L349](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/setup.py#L349), and [`DataPlaneClient`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/data_plane/interfaces.py#L248) (an `ABC`) declares every method the actor invokes — `put_samples` / `get_samples` / `clear_samples`. It's already exported from `nemo_rl.data_plane.__all__` and this file already imports from that package at [L40](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/setup.py#L40), so it's a one-line import extension. The dynamic [`getattr(self._dp_client, method_name)`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller.py#L276) is unaffected — `getattr(obj, str)` returns `Any` regardless.

Unlike the sibling `Any` fields (`gen_handle`/`trainer_handle` are Ray actor handles, `advantage_estimator` is a callable union), `dp_client` has one unambiguous concrete type — the clear one to fix.

**Action:** narrow the field and extend the import:
```suggestion
    dp_client: DataPlaneClient
```
and at L40: `from nemo_rl.data_plane import DataPlaneClient, build_data_plane_client`

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -0,0 +1,400 @@
...
+            num_gpus_per_node=gpus_per_node,
+            max_colocated_worker_groups=1 if backend == "megatron" else 2,
+            port_range_low=port_range_low,
+            port_range_high=port_range_high,
+        )
+        return cluster, cluster
+
+    # Non-colocated: split node into train + inference clusters.
+    assert backend != "megatron", (
+        "Non-colocated inference is not supported for Megatron generation backends."
```
terrykong: [`setup.py:101`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller_utils/setup.py#L101)

This assert makes the Single Controller reject the **Megatron *generation* backend** (generating with mcore itself) whenever inference is non-colocated. Legacy `grpo.py` **supports** exactly this: [`init_megatron_generation`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/grpo.py#L1037-L1055) has an explicit non-colocated `else` branch that builds a dedicated `MegatronGeneration(cluster=inference_cluster, ..., weights_path=...)`, and grpo has no equivalent assert — so SC is stricter than its direct analog here (`distillation.py` carries the same assert, but grpo does not).

Scope, to be clear: this does **not** affect Megatron *training* + vLLM *generation* non-colocated (the common case, and what every SC recipe uses via `backend: vllm`) — that's fully supported. It only blocks mcore-as-generation-engine in non-colocated mode, which is latent today (no SC recipe uses it) and fail-loud.

**Action:** if this is an intentional not-yet-wired gap, keep the guard but reword the message to name the *generation backend* (it currently reads like it's about Megatron training). Tracked as a cleanup item in #2625. If it should work, mirror grpo's `init_megatron_generation` non-colocated branch.

nemo_rl/algorithms/single_controller.py
```diff
@@ -538,240 +389,245 @@ def _release_permits_if_task_not_started(
...
+        # print(
+        #     f"  _sync_weights: drained in {drain_elapsed:.3f}s, "
+        #     f"syncing weights v{self._trainer_version}",
+        #     flush=True,
+        # )
 
         t0 = time.monotonic()
-        await self._weight_synchronizer.sync_weights(self._trainer_version)
+        await asyncio.to_thread(self._weight_synchronizer.sync_weights)
+        if self._async_cfg.max_staleness_versions == 0:
```
terrykong: [`single_controller.py:625`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller.py#L625)

KV-cache invalidation in the SC path happens **only** here, gated on `max_staleness_versions == 0` (sync mode). I checked every invalidation site: the regular synchronizers (Collective / IPC / HTTP / CheckpointEngine) don't invalidate internally, so for the shipped recipes this gate is the only one; only [`VllmRemoteSparseWeightSynchronizer`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/weight_sync/vllm_remote_sparse_weight_synchronizer.py#L133) invalidates on its own (unconditionally — delta/sparse refit requires it).

The correct semantics are being finalized in **[#3152 "Always honor `recompute_kv_cache_after_weight_updates`"](https://github.com/NVIDIA-NeMo/RL/pull/3152)** (open, not yet merged): it decouples invalidation from `in_flight_weight_updates` and honors the flag on its own — per that PR, *"if KV cache (e.g. same prompt) is reused across steps with different weights we can keep a stale KV cache around for a long time which can manifest as compounding policy KL divergence and logprob errors."* So invalidation should be driven by `recompute_kv_cache_after_weight_updates`, **independent of** the sync/async or in-flight distinction.

Two gaps in SC vs. that:

1. **The knob is gone.** SC's new `async_rl` config doesn't carry `recompute_kv_cache_after_weight_updates`, and SC never reads `async_grpo`/that flag (grep-confirmed) — so an SC user has no way to opt into KV recompute at all. Legacy honors it in both the vLLM path ([`trajectory_collector.resume_after_refit`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/async_utils/trajectory_collector.py#L545), the code #3152 fixes) and the megatron path ([grpo.py](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/grpo.py#L967), `kv_cache_management_mode="recompute"`).
2. **Async mode never invalidates.** With `max_staleness_versions >= 1` and a regular synchronizer, vLLM serves KV computed under old weights across weight versions — exactly the stale-KV / compounding-KL situation #3152 targets (the concept originally landed in [#1455](https://github.com/NVIDIA-NeMo/RL/pull/1455)).

Overlaps Known Issue #2 / Known Missing Feature #6 (tracked in #2625), so non-blocking — but the **fail-silent** part is worth fixing now: nothing tells the user async KV recompute is unavailable.

**Action:** carry `recompute_kv_cache_after_weight_updates` into `async_rl` and gate `invalidate_kv_cache()` on it independent of the sync/async split, matching #3152's semantics. Until then, warn/assert at setup when recompute is requested but unhonored, so it fails loud instead of silently serving stale KV.

_(Minor: Known Missing #6's note says "sync mode (`max_staleness_versions=1`)", but sync mode is `== 0` here and in the `__init__` validation — worth fixing the tracking text.)_

nemo_rl/algorithms/single_controller.py
```diff
@@ -538,240 +389,245 @@ def _release_permits_if_task_not_started(
...
+        # drain_elapsed = time.monotonic() - drain_start
+        # print(
+        #     f"  _sync_weights: drained in {drain_elapsed:.3f}s, "
+        #     f"syncing weights v{self._trainer_version}",
+        #     flush=True,
+        # )
 
         t0 = time.monotonic()
-        await self._weight_synchronizer.sync_weights(self._trainer_version)
+        await asyncio.to_thread(self._weight_synchronizer.sync_weights)
```
terrykong: [`single_controller.py:624`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/single_controller.py#L624)

`_sync_weights` calls `sync_weights()` with **no `kv_scales`**, and SC never computes them — so **FP8 KV-cache scale sync isn't implemented in the SC path**. grpo's train loop does it: when the vLLM backend reports [`requires_kv_scale_sync`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/models/generation/interfaces.py#L329) (FP8 KV cache), it computes scales via [`policy.calibrate_qkv_fp8_scales(...)`](https://github.com/NVIDIA-NeMo/RL/blob/d4a9d0f7c1686d936556441174c6519933aaddea/nemo_rl/algorithms/grpo.py#L2626) and forwards them through [`sync_weights(..., kv_scales=...)`](https://github.com/NVIDIA-NeMo/RL/blob/d4a9d0f7c1686d936556441174c6519933aaddea/nemo_rl/algorithms/grpo.py#L2636) (feature at [grpo.py#L2475](https://github.com/NVIDIA-NeMo/RL/blob/d4a9d0f7c1686d936556441174c6519933aaddea/nemo_rl/algorithms/grpo.py#L2475)).

The pieces all exist here — the [synchronizer interface accepts `kv_scales`](https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/weight_sync/interfaces.py#L65) (defaults `None`), SC's trainer is a `TQPolicy(Policy)` which has `calibrate_qkv_fp8_scales`, and `self._gen` exposes `requires_kv_scale_sync` — only the driver-side compute+forward is missing. Effect: an FP8-KV-cache rollout under SC keeps **stale FP8 quantization scales** after each weight update → degraded generation (the class of bug [#3226](https://github.com/NVIDIA-NeMo/RL/pull/3226) fixed for the colocated IPC path).

Latent (no shipped SC recipe enables FP8 KV cache), so non-blocking; tracked in #2625.

**Action:** in `_sync_weights`, when `self._gen.requires_kv_scale_sync` is True, compute scales via `self._trainer.calibrate_qkv_fp8_scales(...)` and pass `kv_scales=` to `sync_weights`, mirroring grpo.py. At minimum, warn/assert at setup if an FP8-KV-cache generation config is used with SC until this is wired.

terrykong (commented):
Team review of the stack's part 3/4 (setup + entrypoint). The split is clean and the convergence evidence in the description is appreciated — sync and async curves both look reasonable. 12 inline comments below; the two to look at first are the `SGLangGeneration` call signature in `setup.py` and the CI registration of `L1_Functional_Tests_SingleController.sh`, which references a script that only lands in #3267.

**Re-staged against `850b58377e`.** The branch was rebased during review (from `eaca7ee8`); every finding was re-verified at the new head and all still reproduce — only line numbers shifted. Inline permalinks intentionally pin to the `eaca7ee8` snapshot (immutable, identical code).

**Stack / rebase gap:** this branch is stacked on a *pre-fix* version of #3220 — the fix commits `09a88ff` and `d3188b2` that resolve several #3220 review threads (e.g. the empty `rewards`/`masked_advantages` accumulators that silently drop those metrics) are **not** in this head. Rebasing onto the latest #3220 before merge will pull them in so the inherited train-path code matches #3220's reviewed state.

**Reconciled against the PR description + #3267:** the provisional `prev_logprobs_required`/`reference_logprobs_required` gating is the author-documented Known Issue #3 (tracked in #2625), so it is deliberately not flagged; every remaining comment was cross-checked against #3267 and none is resolved there.

Non-blocking notes (no inline comment):
- [`single_controller.py:255`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller.py#L255): `self._logger.finish()` is skipped when a pump crash re-raises at [L251](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/nemo_rl/algorithms/single_controller.py#L251) (still true at the new head, now L246 inside the `try` before the `finally`); an optional `try/finally` would flush buffered metrics on the crash path too.
- `config.py`'s `AsyncRLConfig`/`MasterConfig`/`AdvantageConfig` classes could use short class docstrings (the retired `SingleControllerConfig` had one).
- [`test_rollout_pump.py`](https://github.com/NVIDIA-NeMo/RL/blob/eaca7ee86dbd32787941293b477b70d0f2159d4f/tests/unit/single_controller/test_rollout_pump.py#L51) uses a real `VllmGeneration` but carries no `@pytest.mark.vllm`, so per `tests/unit/conftest.py` routing it runs in the vllm-less default shard and is excluded from `--vllm-only`. This pre-dates this PR, but since the file is already touched, adding the marker here would be cheap.
- Stack-level follow-up (pre-existing from #3220, untouched by this diff): `_train_pump`'s `if not step_open: break` guard sits after the inner dispatch loop, which spins on `sampler.select()` returning `None` — if the rollout pump ever finishes while the buffer can't supply a final step, the trainer spins forever. Might be worth folding into the #2625 tracking.

Checks run locally (at review time, `eaca7ee8`): `pre-commit run --all-files` passes; `tests/unit/single_controller/test_single_controller_setup.py` passes (12/12). mcore/vLLM-venv suites deferred to CI.

_Generated by Claude Code_

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [470e7e821](https://github.com/NVIDIA-NeMo/RL/commit/470e7e82115f7682bba736c9ee4822c04a917d63)

examples/run_grpo_single_controller.py
RayenTian: Done in [e536bf5c7](https://github.com/NVIDIA-NeMo/RL/commit/e536bf5c7523b55ff9ee95bc92cde38332b77e22)

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [ed283dfcb](https://github.com/NVIDIA-NeMo/RL/commit/ed283dfcb31f31cefd32f5c691c19d50f188110a)

nemo_rl/algorithms/single_controller_utils/setup.py
RayenTian: Done in [e536bf5c7](https://github.com/NVIDIA-NeMo/RL/commit/e536bf5c7523b55ff9ee95bc92cde38332b77e22)

nemo_rl/algorithms/single_controller_utils/setup.py
RayenTian: Done in [ed283dfcb](https://github.com/NVIDIA-NeMo/RL/commit/ed283dfcb31f31cefd32f5c691c19d50f188110a)

nemo_rl/models/policy/lm_policy.py
RayenTian: Done in [e536bf5c7](https://github.com/NVIDIA-NeMo/RL/commit/e536bf5c7523b55ff9ee95bc92cde38332b77e22)

tests/functional/L1_Functional_Tests_SingleController.sh
RayenTian: Done in [ed283dfcb](https://github.com/NVIDIA-NeMo/RL/commit/ed283dfcb31f31cefd32f5c691c19d50f188110a)

nemo_rl/algorithms/single_controller_utils/setup.py
RayenTian: Done in [1cb29d175](https://github.com/NVIDIA-NeMo/RL/commit/1cb29d1755b318f0fe945ef5d62a32bfe8a6df5d)

examples/run_grpo_single_controller.py
RayenTian: Done in [1cb29d175](https://github.com/NVIDIA-NeMo/RL/commit/1cb29d1755b318f0fe945ef5d62a32bfe8a6df5d)

tests/test_suites/nightly.txt
RayenTian: Done in [fe65cd6d2](https://github.com/NVIDIA-NeMo/RL/commit/fe65cd6d2254204ea0d0a3108857ff9eb6e0573a)

nemo_rl/algorithms/single_controller_utils/setup.py
RayenTian: Done in [fe65cd6d2](https://github.com/NVIDIA-NeMo/RL/commit/fe65cd6d2254204ea0d0a3108857ff9eb6e0573a)

nemo_rl/algorithms/single_controller_utils/setup.py
RayenTian: Done in [fe65cd6d2](https://github.com/NVIDIA-NeMo/RL/commit/fe65cd6d2254204ea0d0a3108857ff9eb6e0573a)

tests/functional/grpo_dp_single_controller.sh
RayenTian: Done in [fe65cd6d2](https://github.com/NVIDIA-NeMo/RL/commit/fe65cd6d2254204ea0d0a3108857ff9eb6e0573a)

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [e706b52b4](https://github.com/NVIDIA-NeMo/RL/commit/e706b52b417759ee98463907c100f7fd871d5659)

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [9288d1a34](https://github.com/NVIDIA-NeMo/RL/commit/9288d1a343e867c63222206aea1fb7edcb4a62de)

RayenTian:
/ok to test [9288d1a](https://github.com/NVIDIA-NeMo/RL/pull/3266/commits/9288d1a343e867c63222206aea1fb7edcb4a62de)

nemo_rl/algorithms/single_controller_utils/setup.py
terrykong: fixed in 1cb29d1755b318f0fe945ef5d62a32bfe8a6df5d

tests/functional/L1_Functional_Tests_SingleController.sh
terrykong: fixed in ed283dfcb31f31cefd32f5c691c19d50f188110a

nemo_rl/algorithms/single_controller.py
terrykong: fixed in ed283dfcb31f31cefd32f5c691c19d50f188110a

nemo_rl/algorithms/single_controller.py
terrykong: fixed in 470e7e82115f7682bba736c9ee4822c04a917d63

nemo_rl/algorithms/single_controller_utils/setup.py
terrykong: fixed in e536bf5c7523b55ff9ee95bc92cde38332b77e22

examples/run_grpo_single_controller.py
terrykong: fixed in e536bf5c7523b55ff9ee95bc92cde38332b77e22

nemo_rl/models/policy/lm_policy.py
terrykong: fixed in e536bf5c7523b55ff9ee95bc92cde38332b77e22

examples/run_grpo_single_controller.py
terrykong: fixed in 1cb29d1755b318f0fe945ef5d62a32bfe8a6df5d

nemo_rl/algorithms/single_controller_utils/setup.py
terrykong: fixed in 1cb29d1755b318f0fe945ef5d62a32bfe8a6df5d

tests/functional/grpo_dp_single_controller.sh
terrykong: fixed in fe65cd6d2254204ea0d0a3108857ff9eb6e0573a

tests/test_suites/nightly.txt
terrykong: fixed in fe65cd6d2254204ea0d0a3108857ff9eb6e0573a

nemo_rl/algorithms/single_controller_utils/setup.py
terrykong: fixed in fe65cd6d2254204ea0d0a3108857ff9eb6e0573a

nemo_rl/algorithms/single_controller.py
terrykong: fixed in e706b52b417759ee98463907c100f7fd871d5659

nemo_rl/algorithms/single_controller.py
terrykong: fixed in 9288d1a343e867c63222206aea1fb7edcb4a62de

nemo_rl/algorithms/single_controller_utils/setup.py
terrykong: partially addressed in fe65cd6d2254204ea0d0a3108857ff9eb6e0573a — the assert message now names the generation backend ("The Megatron generation backend does not support non-colocated inference in SingleController"). The capability gap itself remains by design (non-colocated Megatron generation is still unsupported here, unlike grpo.py's dedicated-inference-policy path); now tracked in #2625.

nemo_rl/algorithms/single_controller.py
```diff
@@ -538,240 +331,298 @@ def _release_permits_if_task_not_started(
...
+            calibration_batches: list[BatchedDataDict[Any]] = []
+
+            with self._timer.time("total_step_time"):
+                while groups_dispatched < grpo_cfg["num_prompts_per_step"]:
+                    # Wait for a selectable batch
+                    with self._timer.time("exposed_generation"):
+                        await asyncio.sleep(0)
+
+                        # Evict stale groups
+                        evicted = await self._sampler.evict(
```
terrykong, line 378: _As discussed previously, we agreed to address the premature-eviction (not-yet-ready) case in #3220 and the abort-on-outdated-version behavior in this PR (#3266)._

Tracking comment for that follow-up. Today `_train_pump` handles staleness by **evicting** stale prompt groups after the fact — [`self._sampler.evict(current_train_weight=self._trainer_version)`](https://github.com/NVIDIA-NeMo/RL/blob/9288d1a343e867c63222206aea1fb7edcb4a62de/nemo_rl/algorithms/single_controller.py#L376) — and releasing their buffer capacity. The agreed approach is for #3266 to instead **abort stale in-flight rollouts during refit** (in [`_sync_weights`](https://github.com/NVIDIA-NeMo/RL/blob/9288d1a343e867c63222206aea1fb7edcb4a62de/nemo_rl/algorithms/single_controller.py#L568)) when their weight version is outdated, rather than evicting them in the train pump.

This is Known Missing Feature #4 in the PR description ("Abort stale in-flight rollouts in refit instead of evicting in train"), tracked in #3263 / #2625.

**Action:** in a follow-up on this PR (or #3263), add a version check + abort of in-flight `generate_and_push` dispatches at refit time, and reconcile with the current evict-in-train path so a group isn't both aborted and evicted.

nemo_rl/algorithms/single_controller.py
saumishr, line 681: Could be a rebase issue but we seem to be appending rewards to the _step_log_dict twice? Once here and then at line#652.

nemo_rl/algorithms/single_controller.py
```diff
@@ -675,44 +559,70 @@ async def _train_pump(self) -> None:
...
-        while self._inflight_rollouts > 0:
-            await asyncio.sleep(0.005)
-
-        drain_elapsed = time.monotonic() - drain_start
-        print(
-            f"  _sync_weights: drained in {drain_elapsed:.3f}s, "
-            f"syncing weights v{self._trainer_version}",
-            flush=True,
-        )
+        # TODO: currently drain is not implemented, comment out for now
```
saumishr: How are we tracking this ToDo? Wanna make sure that it doesn't get lost. Is there a follow up PR already on this?
Also, it may be good to have the code removed instead of keeping it around as commented out.

nemo_rl/algorithms/single_controller.py
```diff
@@ -329,6 +195,7 @@ async def run(self) -> dict[str, Any]:
                 # rollout pump leaves the train pump to drain committed groups.
                 await rollout_task
             await train_task
+            self._logger.finish()
```
saumishr: Should this rather be in finally? Otherwise we may miss critical logs in case of exceptions in the try block

nemo_rl/algorithms/single_controller.py
```diff
@@ -177,95 +85,55 @@ class SingleControllerActor:
...
-        self._rollout_manager = rollout_manager
+        """Initialize the SingleController actor.
+
+        Args:
+            master_config: SC MasterConfig.
+            actor_args: Pre-built actor args from setup_single_controller.
+        """
+        validate_single_controller_config(master_config)
+
+        self._advantage_cfg = AdvantageConfig()
```
saumishr, line 99: This is always constructed with defaults, so configs will not be YAML-settable. Is this a WIP?

nemo_rl/models/policy/__init__.py
```diff
@@ -539,7 +539,7 @@ class PolicyConfig(TypedDict):
     max_total_sequence_length: int
     # This sets the clipping norm for the DTensorPolicyWorkers (Megatron's is called clip_grad)
     max_grad_norm: NotRequired[float | int | None]
-    refit_buffer_size_gb: NotRequired[float]
+    refit_buffer_size_gb: NotRequired[int]
```
saumishr: Curious about this change to int. Since its in GB, any issues if we keep it as `float | int`?

nemo_rl/algorithms/single_controller.py
terrykong: i think @yuki-97 / @mehraakash have been using https://github.com/NVIDIA-NeMo/RL/issues/2625. so i've been appending to that. we should make sure these TODOs are there too @RayenTian

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [31640629e](https://github.com/NVIDIA-NeMo/RL/commit/31640629e53c7980c8c5188ecb00ba268990e291).

nemo_rl/algorithms/single_controller.py
RayenTian: Removed the commented-out implementation and left a concise TODO linked to [#2625](https://github.com/NVIDIA-NeMo/RL/issues/2625), which already tracks the refit drain gate. Done in [31640629e](https://github.com/NVIDIA-NeMo/RL/commit/31640629e53c7980c8c5188ecb00ba268990e291).

nemo_rl/algorithms/single_controller.py
RayenTian: Done in [31640629e](https://github.com/NVIDIA-NeMo/RL/commit/31640629e53c7980c8c5188ecb00ba268990e291).

nemo_rl/algorithms/single_controller.py
RayenTian: In my understanding, `AdvantageConfig` is intentionally internal and remains the DataPlane field mapping. The logprob requirement is now decoupled from optional field names and derived from the existing GRPO/loss configuration in [31640629e](https://github.com/NVIDIA-NeMo/RL/commit/31640629e53c7980c8c5188ecb00ba268990e291).

nemo_rl/models/policy/__init__.py
RayenTian: Done in [31640629e](https://github.com/NVIDIA-NeMo/RL/commit/31640629e53c7980c8c5188ecb00ba268990e291).

RayenTian:
/ok to test [0629fed](https://github.com/NVIDIA-NeMo/RL/pull/3266/commits/0629fed9c1d2c409434278ba2a6f16a3e1dad9ba)

examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml
RayenTian: Before apply PR! https://github.com/NVIDIA-NeMo/RL/pull/3282. (fix grad_norm), the grad_norm has a mismatch with baseline as @yuki-97 's results pasted in the [PR doc](https://github.com/NVIDIA-NeMo/RL/pull/3266#issue-4911566326).
image

After apply PR! https://github.com/NVIDIA-NeMo/RL/pull/3282. (fix grad_norm): The discrepancy is alleviated but not fully matched
image

And in both senario, the streaming2 have higher rewards.

nemo_rl/algorithms/single_controller.py
RayenTian: #3263 will follow this comment.

examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml
RayenTian: Now this PR doesn't include PR https://github.com/NVIDIA-NeMo/RL/pull/3282 's fix. Not sure whether we should cherry-pick https://github.com/NVIDIA-NeMo/RL/pull/3282 fix to #3266 first, then merge #3266 to main together. Or  we can first merge #3266, then merge #3282. Need @terrykong or @yuki-97  's confirm.

RayenTian: Nightly test also reran. The metrics checks can pass, and the wandb results as follow: 
image

RayenTian: Update: now PR https://github.com/NVIDIA-NeMo/RL/pull/3282 's fix is cherry-picked to this branch.

RayenTian:
/ok to test [7c100ad](https://github.com/NVIDIA-NeMo/RL/pull/3266/commits/7c100ad76ac684fa844809d8be57b6f474a8674d)

terrykong (approved):

terrykong:
CI failure seemed transient (topp/topk test barely missed metric test). retrying

terrykong (approved):

PR #3267, opened 2026-07-17
feat(sc): NeMo-Gym path

yuki-97:
Part 4/4 of the #2819 split. Stack on #3266.

## Summary

Wires NeMo-Gym rollouts into the SC (Single Controller) entrypoint by introducing an SC-facing gym-actor spinup helper and wiring it into the SC setup path. The existing `grpo.py` path retains its local `_spinup_nemo_gym` implementation for now because it also supplies GRPO-specific timing, routed-experts dtype, and fastokens configuration.

- **`spinup_nemo_gym_actor`** (`nemo_rl/environments/nemo_gym.py`): factors the NeMo-Gym actor construction needed by SC (config assembly, uv/venv cache reuse, image-baked venv reuse, and soft `NodeAffinitySchedulingStrategy` when `num_gpu_nodes > 0`) into a reusable helper. Consolidating the remaining GRPO-specific startup path is left for follow-up work.
- **SC setup wiring** (`nemo_rl/algorithms/single_controller_utils/setup.py`): `setup_single_controller` branches on `_should_use_nemo_gym` — skips env spinup in `setup_response_data`, brings up generation first, then calls `spinup_nemo_gym_actor` with the vLLM OpenAI URLs. Raises `NotImplementedError` on non-vllm backends.
- **`examples/run_grpo_single_controller.py`**: restores the `setup_nemo_gym_config` call (removed in #3266 while the SC path had no gym support).
- **Tests**: `tests/unit/single_controller/test_single_controller_setup.py` gains env-handle wiring + non-vllm backend guard coverage (replaces the previous "not supported yet" trip-wire). New `tests/functional/grpo_async_gym_single_controller.sh` covers SC + NeMo-Gym end-to-end.

### DIFF
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -325,6 +325,11 @@ def setup_single_controller(
     # ==========================
     # TODO: add validate dataset wiring.
     use_nemo_gym = _should_use_nemo_gym(cast(GrpoMasterConfig, master_config))
+    if use_nemo_gym and generation_config["backend"] != "vllm":
+        raise NotImplementedError(
+            "SC NeMo-Gym integration currently supports the vllm backend "
+            f"only; got {generation_config['backend']!r}"
+        )
     if use_nemo_gym:
         # NeMo-Gym creates the env actor outside setup_response_data; we wire
         # it in after generation is up (it needs the OpenAI server URLs).
@@ -377,11 +382,6 @@ def setup_single_controller(
     # NeMo-Gym actor (after generation is up so OpenAI URLs are available)
     # ==========================
     if use_nemo_gym:
-        if generation_config["backend"] != "vllm":
-            raise NotImplementedError(
-                "SC NeMo-Gym integration currently supports the vllm backend "
-                f"only; got {generation_config['backend']!r}"
-            )
         enable_router_replay = router_replay_enabled(master_config.policy)
         env_handles["nemo_gym"] = spinup_nemo_gym_actor(
             env_configs=master_config.env,
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -382,6 +382,8 @@ def setup_single_controller(
     # NeMo-Gym actor (after generation is up so OpenAI URLs are available)
     # ==========================
     if use_nemo_gym:
+        # TODO(#2625): Mirror GRPO's deferred vLLM load so NeMo-Gym spinup
+        # overlaps model loading instead of running serially afterward.
         enable_router_replay = router_replay_enabled(master_config.policy)
         env_handles["nemo_gym"] = spinup_nemo_gym_actor(
             env_configs=master_config.env,
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -49,11 +49,13 @@ from nemo_rl.data.utils import setup_response_data
 from nemo_rl.data_plane import DataPlaneClient, build_data_plane_client
 from nemo_rl.distributed.virtual_cluster import RayVirtualCluster
 from nemo_rl.environments.interfaces import EnvironmentInterface
+from nemo_rl.environments.nemo_gym import spinup_nemo_gym_actor
 from nemo_rl.experience.rollout_manager import RolloutManager
 from nemo_rl.models.generation.sglang.config import SGLangConfig
 from nemo_rl.models.generation.sglang.sglang_generation import SGLangGeneration
 from nemo_rl.models.generation.vllm import VllmGeneration
 from nemo_rl.models.generation.vllm.config import VllmConfig
+from nemo_rl.models.megatron.router_replay import router_replay_enabled
 from nemo_rl.models.policy.tq_policy import TQPolicy
 from nemo_rl.weight_sync import WeightSynchronizer, create_weight_synchronizer
 
@@ -318,16 +320,22 @@ def setup_single_controller(
     # Setup Dataset & Environments
     # ==========================
     # TODO: add validate dataset wiring.
-    if _should_use_nemo_gym(cast(GrpoMasterConfig, master_config)):
-        raise NotImplementedError(
-            "NeMo-Gym integration for SingleController is not supported yet; "
-            "it will land in https://github.com/NVIDIA-NeMo/RL/pull/3267."
+    use_nemo_gym = _should_use_nemo_gym(cast(GrpoMasterConfig, master_config))
+    if use_nemo_gym:
+        # NeMo-Gym creates the env actor outside setup_response_data; we wire
+        # it in after generation is up (it needs the OpenAI server URLs).
+        response_data = setup_response_data(
+            tokenizer, data_config, env_configs=None
         )
-    response_data = setup_response_data(
-        tokenizer, data_config, env_configs=master_config.env
-    )
-    assert len(response_data) == 4
-    dataset, _val_dataset, env_handles, _val_env_handles = response_data
+        assert len(response_data) == 2
+        dataset, _val_dataset = response_data
+        env_handles: dict[str, EnvironmentInterface] = {}
+    else:
+        response_data = setup_response_data(
+            tokenizer, data_config, env_configs=master_config.env
+        )
+        assert len(response_data) == 4
+        dataset, _val_dataset, env_handles, _val_env_handles = response_data
     dataloader = StatefulDataLoader(
         dataset,
         batch_size=grpo_config["num_prompts_per_step"],
@@ -362,6 +370,23 @@ def setup_single_controller(
             )
             generation = gen_future.result()
             policy = policy_future.result()
+
+    # ==========================
+    # NeMo-Gym actor (after generation is up so OpenAI URLs are available)
+    # ==========================
+    if use_nemo_gym:
+        if generation_config["backend"] != "vllm":
+            raise NotImplementedError(
+                "SC NeMo-Gym integration currently supports the vllm backend "
+                f"only; got {generation_config['backend']!r}"
+            )
+        enable_router_replay = router_replay_enabled(master_config.policy)
+        env_handles["nemo_gym"] = spinup_nemo_gym_actor(
+            env_configs=master_config.env,
+            base_urls=generation.dp_openai_server_base_urls,
+            model_name=generation_config["model_name"],
+            enable_router_replay=enable_router_replay,
+        )
 
     # ==========================
     # Setup Data Plane Client & Weight Sync
@@ -403,6 +428,7 @@ def setup_single_controller(
         max_rollout_turns=grpo_config.get("max_rollout_turns"),
         policy_generation=generation,
         generation_config=generation_config,
+        use_nemo_gym=use_nemo_gym,
         tq_buffer=tq_buffer,
     )
 
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -328,9 +328,7 @@ def setup_single_controller(
     if use_nemo_gym:
         # NeMo-Gym creates the env actor outside setup_response_data; we wire
         # it in after generation is up (it needs the OpenAI server URLs).
-        response_data = setup_response_data(
-            tokenizer, data_config, env_configs=None
-        )
+        response_data = setup_response_data(tokenizer, data_config, env_configs=None)
         assert len(response_data) == 2
         dataset, _val_dataset = response_data
         env_handles: dict[str, EnvironmentInterface] = {}
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -55,7 +55,10 @@ from nemo_rl.models.generation.sglang.config import SGLangConfig
 from nemo_rl.models.generation.sglang.sglang_generation import SGLangGeneration
 from nemo_rl.models.generation.vllm import VllmGeneration
 from nemo_rl.models.generation.vllm.config import VllmConfig
-from nemo_rl.models.megatron.router_replay import router_replay_enabled
+from nemo_rl.models.megatron.router_replay import (
+    configure_vllm_for_router_replay,
+    router_replay_enabled,
+)
 from nemo_rl.models.policy.tq_policy import TQPolicy
 from nemo_rl.weight_sync import WeightSynchronizer, create_weight_synchronizer
 
@@ -169,6 +172,7 @@ def _build_generation(
         vllm_config.setdefault("vllm_kwargs", {})["hf_overrides"] = (
             master_config.policy.get("hf_config_overrides", {})
         )
+        configure_vllm_for_router_replay(master_config.policy)
         gen = VllmGeneration(cluster=inference_cluster, config=vllm_config)
     elif backend == "sglang":
         sglang_config = cast(SGLangConfig, generation_config)
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -18,12 +18,14 @@ import sys
 from collections import Counter
 from collections.abc import AsyncGenerator
 from pathlib import Path
-from typing import Any, Dict, List, NotRequired, TypedDict
+from typing import Any, Dict, List, NotRequired, Optional, TypedDict
 
 import ray
 import torch
+from ray.util.scheduling_strategies import NodeAffinitySchedulingStrategy
 from transformers import PreTrainedTokenizerBase
 
+from nemo_rl.distributed.ray_actor_environment_registry import get_actor_python_env
 from nemo_rl.distributed.virtual_cluster import (
     DEFAULT_GYM_PORT_RANGE_HIGH,
     DEFAULT_GYM_PORT_RANGE_LOW,
@@ -32,6 +34,7 @@ from nemo_rl.distributed.virtual_cluster import (
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
 from nemo_rl.utils.timer import Timer
+from nemo_rl.utils.venvs import create_local_venv_on_each_node
 
 # Kept local (not imported from models.generation) so the gym actor stays free of
 # generation-module imports. Must cover every name resolve_routed_experts_dtype
@@ -600,3 +603,66 @@ def setup_nemo_gym_config(config, tokenizer) -> None:
     # Stop strings or token ids are not supported
     generation_config["stop_strings"] = None
     generation_config["stop_token_ids"] = None
+
+
+def spinup_nemo_gym_actor(
+    env_configs: dict[str, Any],
+    base_urls: list[Optional[str]],
+    model_name: str,
+    enable_router_replay: bool,
+) -> Any:
+    """Spin up the NeMo-Gym actor against the given generation server URLs.
+
+    When ``env_configs["nemo_gym"]["num_gpu_nodes"] > 0``, the actor is
+    scheduled with soft NodeAffinity to the current Ray node so its colocated
+    GPU resources land where the caller expects.
+    """
+    nemo_gym_dict = dict(env_configs["nemo_gym"])
+
+    # NeMo-RL-side detection knobs are top-level NemoGymConfig fields
+    # (where the detector reads them), not part of Gym's global config.
+    invalid_tool_call_patterns = nemo_gym_dict.pop("invalid_tool_call_patterns", None)
+    thinking_tags = nemo_gym_dict.pop("thinking_tags", None)
+
+    # Pass prebuilt cache + venv dirs through the global config so the gym reuses
+    # image-baked venvs instead of rebuilding them.
+    uv_cache_dir = get_nemo_gym_uv_cache_dir()
+    if uv_cache_dir is not None:
+        nemo_gym_dict.setdefault("uv_cache_dir", uv_cache_dir)
+    uv_venv_dir = get_nemo_gym_venv_dir()
+    if uv_venv_dir is not None:
+        nemo_gym_dict.setdefault("uv_venv_dir", uv_venv_dir)
+
+    nemo_gym_cfg = NemoGymConfig(
+        model_name=model_name,
+        base_urls=base_urls,
+        invalid_tool_call_patterns=invalid_tool_call_patterns,
+        thinking_tags=thinking_tags,
+        require_routed_experts=enable_router_replay,
+        initial_global_config_dict=nemo_gym_dict,
+    )
+
+    nemo_gym_py_exec = get_actor_python_env("nemo_rl.environments.nemo_gym.NemoGym")
+    if nemo_gym_py_exec.startswith("uv"):
+        nemo_gym_py_exec = create_local_venv_on_each_node(
+            nemo_gym_py_exec, "nemo_rl.environments.nemo_gym.NemoGym"
+        )
+
+    nemo_gym_opts: dict[str, Any] = {}
+    if nemo_gym_dict.get("num_gpu_nodes", 0):
+        nemo_gym_opts["scheduling_strategy"] = NodeAffinitySchedulingStrategy(
+            node_id=ray.get_runtime_context().get_node_id(),
+            soft=True,
+        )
+    nemo_gym_opts["runtime_env"] = {
+        "py_executable": nemo_gym_py_exec,
+        "env_vars": {
+            **os.environ,
+            "VIRTUAL_ENV": nemo_gym_py_exec,
+            "UV_PROJECT_ENVIRONMENT": nemo_gym_py_exec,
+        },
+    }
+
+    actor = NemoGym.options(**nemo_gym_opts).remote(nemo_gym_cfg)
+    ray.get(actor._spinup.remote())
+    return actor
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -616,6 +616,19 @@ def spinup_nemo_gym_actor(
     When ``env_configs["nemo_gym"]["num_gpu_nodes"] > 0``, the actor is
     scheduled with soft NodeAffinity to the current Ray node so its colocated
     GPU resources land where the caller expects.
+
+    Args:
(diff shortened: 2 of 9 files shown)


tests/functional/grpo_async_gym_single_controller.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
+# SingleController + NeMo-Gym e2e smoke. Mirrors grpo_async_gym.sh but
```
terrykong, line 2: [`tests/functional/grpo_async_gym_single_controller.sh:2`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/tests/functional/grpo_async_gym_single_controller.sh#L2)

**This new e2e test is registered in no CI lane, so it never runs.** [`L1_Functional_Tests_SingleController.sh`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/tests/functional/L1_Functional_Tests_SingleController.sh#L37) only invokes `grpo_dp_single_controller.sh`; the gym SC test isn't listed (its sibling [`grpo_async_gym.sh` **is** registered](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/tests/functional/L1_Functional_Tests_Gym.sh#L37) in the Gym runner). At the pre-rebase head this test was registered at `L1_Functional_Tests_SingleController.sh:38`; that line was dropped from the base #3266 (correctly, since the script didn't exist there yet) and wasn't re-added here — so the PR ships a test it describes as covering "SC + NeMo-Gym end-to-end," but nothing executes it.

**Action:** register it in `tests/functional/L1_Functional_Tests_SingleController.sh` after line 37:

```bash
run_test fast uv run --no-sync bash ./tests/functional/grpo_async_gym_single_controller.sh
```

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -363,6 +373,23 @@ def setup_single_controller(
             generation = gen_future.result()
             policy = policy_future.result()
 
+    # ==========================
+    # NeMo-Gym actor (after generation is up so OpenAI URLs are available)
+    # ==========================
+    if use_nemo_gym:
+        if generation_config["backend"] != "vllm":
```
terrykong: [`nemo_rl/algorithms/single_controller_utils/setup.py:380`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L380-L384)

**This non-vLLM guard is unreachable through the real entrypoint — it's dead code, and the test that "covers" it only reaches it by mocking the two real gates away. Hoisting it makes it meaningful and fail-fast.** Tracked in #2625.

### The guard
[`nemo_rl/algorithms/single_controller_utils/setup.py:379-384`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L379-L384):
```python
if use_nemo_gym:
    if generation_config["backend"] != "vllm":          # line 380
        raise NotImplementedError(
            "SC NeMo-Gym integration currently supports the vllm backend "
            f"only; got {generation_config['backend']!r}"
        )
```
It reads like the thing that rejects sglang+gym / megatron+gym. But `use_nemo_gym` is decided back at [L327](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L327), and **both** non-vLLM backends are already rejected before line 380:

### sglang + gym → `AssertionError` at L327, inside [`_should_use_nemo_gym`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/grpo.py#L1959)
[`nemo_rl/algorithms/grpo.py:1973-1984`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/grpo.py#L1973-L1984):
```python
if generation_config["backend"] == "vllm":
    should_expose_http_server = generation_config["vllm_cfg"].get("expose_http_server")
elif generation_config["backend"] == "megatron":
    should_expose_http_server = generation_config["mcore_generation_config"].get("expose_http_server")
else:                                    # sglang lands here
    should_expose_http_server = False
assert should_expose_http_server, "... you must expose the generation server ..."   # raises for sglang
```
sglang → `else` → `False` → the [assert](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/grpo.py#L1983) throws at [L327](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L327). Never reaches 380.

### megatron + gym → `ValueError` at L187, inside [`_build_generation`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L361)
megatron has a real branch above, so with `expose_http_server: true` it clears `_should_use_nemo_gym`. Flow then reaches [`_build_generation`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L361), which only knows vllm/sglang — [`nemo_rl/algorithms/single_controller_utils/setup.py:186-188`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L186-L188):
```python
else:
    raise ValueError(
        f"single_controller_utils.setup only supports vllm or sglang generation; got {backend!r}"
    )
```
megatron → [`ValueError` at L187](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L187), before 380, with a *different* error type/message.

### So why is `test_nemo_gym_rejects_non_vllm_backend` green?
[`tests/unit/single_controller/test_single_controller_setup.py:421`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/tests/unit/single_controller/test_single_controller_setup.py#L421-L433) stubs out both gates:
```python
with (
    patch.object(sc_setup_mod, "_should_use_nemo_gym", return_value=True),  # skips the sglang assert
    patch.object(sc_setup_mod, "spinup_nemo_gym_actor") as mock_spinup,     # (_build_generation is a patched_factories mock)
    pytest.raises(NotImplementedError, match="vllm"),
):
    setup_single_controller(mc, MagicMock(pad_token_id=0))
```
With both real gates removed, execution falls through to line 380 and raises `NotImplementedError`. In production those mocks don't exist, so a non-vLLM run raises `AssertionError` (sglang) / `ValueError` (megatron) instead — never this `NotImplementedError`. The test exercises a path production can't take.

### Fix: hoist to right after L327
```python
use_nemo_gym = _should_use_nemo_gym(cast(GrpoMasterConfig, master_config))
if use_nemo_gym and generation_config["backend"] != "vllm":     # ~L328, before any build
    raise NotImplementedError(
        "SC NeMo-Gym integration currently supports the vllm backend "
        f"only; got {generation_config['backend']!r}"
    )
```
…and delete the late copy at 380. Two wins:
1. **Fail fast.** The one case that *can* clear `_should_use_nemo_gym` — **megatron+gym with `expose_http_server: true`** — today spends the full multi-minute cluster+vLLM+policy build, then dies at `_build_generation`'s `ValueError`. Hoisted, it rejects at config time.
2. **Clear message.** `_build_generation`'s "only supports vllm or sglang" reads like a generic backend problem; the hoisted `NotImplementedError` correctly names the gym constraint.

`test_nemo_gym_rejects_non_vllm_backend` stays green after the hoist (it mocks `_should_use_nemo_gym→True`, so it reaches the hoisted check the same way).

**Nuance (why this is low-severity, not a bug):** even hoisted, sglang+gym is *still* unreachable — it dies at the `_should_use_nemo_gym` assert regardless of the guard. The hoist genuinely helps only the megatron+gym path (fail-fast + clearer message). Tracked as an SC cleanup item in #2625.

tests/unit/single_controller/test_single_controller_setup.py
```diff
@@ -400,3 +391,45 @@ def test_megatron_train_iters_not_set_when_disabled(self, patched_factories):
...
+        with (
+            patch.object(sc_setup_mod, "_should_use_nemo_gym", return_value=True),
+            patch.object(
+                sc_setup_mod, "spinup_nemo_gym_actor", return_value=fake_gym_actor
+            ) as mock_spinup,
+            patch.object(sc_setup_mod, "router_replay_enabled", return_value=False),
+        ):
+            actor_args = setup_single_controller(mc, MagicMock(pad_token_id=0))
+
+        mock_spinup.assert_called_once()
```
terrykong: [`tests/unit/single_controller/test_single_controller_setup.py:417`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/tests/unit/single_controller/test_single_controller_setup.py#L417)

`assert_called_once()` doesn't pin the arguments, so a regression passing the wrong `base_urls`/`model_name`/`enable_router_replay` to [`spinup_nemo_gym_actor`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/environments/nemo_gym.py#L601) would still pass. The [call site](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L384-L389) uses all-keyword args, so the kwargs form matches:

```suggestion
        mock_spinup.assert_called_once_with(
            env_configs=mc.env,
            base_urls=patched_factories[
                "_build_generation"
            ].return_value.dp_openai_server_base_urls,
            model_name="test-model",
            enable_router_replay=False,
        )
```

nemo_rl/environments/nemo_gym.py
```diff
@@ -593,3 +596,66 @@ def setup_nemo_gym_config(config, tokenizer) -> None:
...
+    base_urls: list[Optional[str]],
+    model_name: str,
+    enable_router_replay: bool,
+) -> Any:
+    """Spin up the NeMo-Gym actor against the given generation server URLs.
+
+    When ``env_configs["nemo_gym"]["num_gpu_nodes"] > 0``, the actor is
+    scheduled with soft NodeAffinity to the current Ray node so its colocated
+    GPU resources land where the caller expects.
+    """
```
terrykong, line 632: [`nemo_rl/environments/nemo_gym.py:607`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/environments/nemo_gym.py#L601-L612)

New public helper imported by the SC setup path ([`setup.py`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L52)) — per the repo Google-style docstring convention, document its params/return:

```suggestion
    """Spin up the NeMo-Gym actor against the given generation server URLs.

    When ``env_configs["nemo_gym"]["num_gpu_nodes"] > 0``, the actor is
    scheduled with soft NodeAffinity to the current Ray node so its colocated
    GPU resources land where the caller expects.

    Args:
        env_configs: The ``master_config.env`` mapping; ``env_configs["nemo_gym"]``
            supplies the Gym global config plus NeMo-RL detection knobs
            (``invalid_tool_call_patterns``, ``thinking_tags``, ``num_gpu_nodes``).
        base_urls: Per-DP-rank OpenAI-compatible server base URLs from the
            generation backend.
        model_name: Served model name the Gym rollouts should target.
        enable_router_replay: Sets ``require_routed_experts`` on the NemoGymConfig
            (Megatron router-replay correctness).

    Returns:
        The spun-up ``NemoGym`` Ray actor handle (``_spinup`` already awaited).
    """
```

tests/functional/grpo_async_gym_single_controller.sh
```diff
@@ -0,0 +1,113 @@
...
+    ++async_rl.max_buffered_rollouts=4 \
+    $@ \
+    2>&1 | tee $RUN_LOG
+
+uv run tests/json_dump_tb_logs.py $LOG_DIR --output_path $JSON_METRICS
+
+# Observed to be between 0.8-1.3
+uv run tests/check_metrics.py $JSON_METRICS \
+    'median(data["train/gen_kl_error"]) < 1.3' \
+    'max(data["train/reward"]) > 0'
```
terrykong, line 114: [`tests/functional/grpo_async_gym_single_controller.sh:113`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/tests/functional/grpo_async_gym_single_controller.sh#L111-L113)

`max(data["train/reward"]) > 0` is a near-trivial liveness bound (passes if any single rollout in 10 steps earns nonzero reward); the sibling [`grpo_async_gym.sh`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/tests/functional/grpo_async_gym.sh) checks `validation/accuracy > 0.1`. Reasonable now since SC validation wiring is still a [TODO](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L326) — no action needed for this PR.

**Action (optional, follow-up):** once SC validation lands, tighten to a validation-accuracy / trending-mean-reward threshold to match the sibling's signal strength.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -167,6 +172,7 @@ def _build_generation(
         vllm_config.setdefault("vllm_kwargs", {})["hf_overrides"] = (
             master_config.policy.get("hf_config_overrides", {})
         )
+        configure_vllm_for_router_replay(master_config.policy)
```
terrykong, line 175: [`nemo_rl/algorithms/single_controller_utils/setup.py:175`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L175)

This PR wires the **generation** side of router replay (R3) into the SC path — [`configure_vllm_for_router_replay`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L175) here, [`require_routed_experts`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L385-L389) on the gym actor — and [`TQPolicy`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/models/policy/tq_policy.py#L121-L183) selects the routed-experts fields. But R3 isn't actually supported end-to-end on the SC (async + TransferQueue) path yet, and nothing fails loud:

- **`routed_experts` are never preserved through the SC rollout → `TQReplayBuffer` → train path.** That capture lives only in the legacy sync [`experience/rollouts.py`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/experience/rollouts.py#L103-L133) and grpo's [`_preserve_router_replay_routed_experts`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/grpo.py#L1911-L1937); the SC rollout/buffer path never writes them.
- **No SC-side guard.** The legacy path hard-blocks this combo — [grpo.py](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/grpo.py#L3742-L3748): *"Async + TQ support has not been merged yet"* — but the SC path has no equivalent, so `policy.router_replay.enabled=true` on an SC run silently trains without replayed routing (incorrect MoE logprobs).

Not blocking (no shipped SC recipe enables R3). **Action:** no change needed here — flagging that it's tracked: filed as #3327 (assigned) and added to the SC loose-ends list #2625.

terrykong: FYI @zyzhou5

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -363,6 +373,23 @@ def setup_single_controller(
...
+    # NeMo-Gym actor (after generation is up so OpenAI URLs are available)
+    # ==========================
+    if use_nemo_gym:
+        if generation_config["backend"] != "vllm":
+            raise NotImplementedError(
+                "SC NeMo-Gym integration currently supports the vllm backend "
+                f"only; got {generation_config['backend']!r}"
+            )
+        enable_router_replay = router_replay_enabled(master_config.policy)
+        env_handles["nemo_gym"] = spinup_nemo_gym_actor(
```
terrykong, line 388: [`nemo_rl/algorithms/single_controller_utils/setup.py:386`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L376-L391)

**NeMo-Gym is spun up sequentially here, after generation is fully loaded — the grpo analog overlaps it with the vLLM weight-load.** In grpo.py's vLLM path, vLLM is built with [`defer_model_load=True`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/grpo.py#L1199-L1203) to reserve the server URLs up front *without* loading weights, then a [3-way `ThreadPoolExecutor`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/grpo.py#L1237-L1245) runs `init_nemo_gym` concurrently with the vLLM weight-load and policy init — the deliberate optimization noted in its [log line](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/grpo.py#L1195-L1198) (*"Deferred model load: reserving vLLM ports for overlapped NeMo Gym init"*).

The SC path has no deferred-load path: [`_build_generation`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L172-L178) builds `VllmGeneration` non-deferred and calls `finish_generation()` (full weight load), and `spinup_nemo_gym_actor` only runs [after `generation` is up](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L376-L391). So the gym venv build ([`create_local_venv_on_each_node`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/environments/nemo_gym.py#L640-L643)) + `ray.get(actor._spinup.remote())` is fully serialized behind vLLM loading instead of hidden under it — SC setup wall-time ≈ `vLLM_load + gym_spinup` vs. grpo's `max(vLLM_load, gym_spinup)`. Correctness/training are unaffected (setup-time latency only), but this is the async/streaming perf path, so the lost overlap matters more here.

**Action:** add a `TODO` here to mirror grpo's deferred-load overlap and track it as an SC loose-end in #2625 — so the divergence from grpo is explicit and doesn't silently regress SC setup latency. Implementing it later means plumbing `defer_model_load` through [`_build_generation`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L172-L178), reserving the vLLM URLs up front, and adding gym spinup as a third task alongside the existing [generation∥policy `ThreadPoolExecutor`](https://github.com/NVIDIA-NeMo/RL/blob/dd3cb538492ec48a0ab96942e1860a6f96dc6119/nemo_rl/algorithms/single_controller_utils/setup.py#L366-L374).

tests/functional/grpo_async_gym_single_controller.sh
terrykong: maybe just add a todo to mirror and track in #2625

terrykong (commented):
Re-review at the rebased head (`dd3cb538`, now MERGEABLE/CLEAN). The `spinup_nemo_gym_actor` extraction and SC wiring check out: the 2-tuple/4-tuple `setup_response_data` unpacking is correct (now with an explicit `assert len(...) == 2`), `dp_openai_server_base_urls` is guaranteed real by the `_should_use_nemo_gym` asserts, and `configure_vllm_for_router_replay` is a safe no-op when router replay is off. Note: the rebase **dropped** the original PR's grpo.py refactor — grpo.py keeps its own inline `_spinup_nemo_gym` (still importing `NemoGym`, never referencing `spinup_nemo_gym_actor`), so the new helper is currently a near-duplicate used only by the SC path. grpo's sync path is untouched (no behavior change), but the description's "shared function so both grpo.py and the SC setup can call it" no longer holds — worth either re-landing the grpo call-site swap or noting the duplication. No correctness bugs.

Six inline comments: one **medium** (the new e2e functional test is registered in no CI lane, so it never runs — regressed during the rebase), and five low/non-blocking (guard hoist, a test-assertion tightening, a docstring, a smoke-threshold note, and a router-replay tracking note). Enabling router replay (R3) on the SC async+TransferQueue path is only half-wired and fails silently — filed as #3327 (assigned) and added to the SC loose-ends tracker #2625.

Cross-checked against merged #3219/#3220 and base #3266 threads to avoid duplication.

_Reviewed by an agent team (rl-expert, bug-finder, test-agent, comment-reviewer, devil's-advocate), re-audited after rebase and cross-referenced against the PR stack. Generated by Claude Code_

tests/functional/grpo_async_gym_single_controller.sh
RayenTian: Done in [47b097091](https://github.com/NVIDIA-NeMo/RL/commit/47b097091dadcac0c4bfe3cfd7b30e91ed67911a)

tests/unit/single_controller/test_single_controller_setup.py
RayenTian: Done in [47b097091](https://github.com/NVIDIA-NeMo/RL/commit/47b097091dadcac0c4bfe3cfd7b30e91ed67911a)

nemo_rl/environments/nemo_gym.py
RayenTian: Done in [47b097091](https://github.com/NVIDIA-NeMo/RL/commit/47b097091dadcac0c4bfe3cfd7b30e91ed67911a)

nemo_rl/algorithms/single_controller_utils/setup.py
RayenTian: Done in [47b097091](https://github.com/NVIDIA-NeMo/RL/commit/47b097091dadcac0c4bfe3cfd7b30e91ed67911a)

nemo_rl/algorithms/single_controller_utils/setup.py
RayenTian: Acknowledged

tests/functional/grpo_async_gym_single_controller.sh
RayenTian: Add a todo item #2625

RayenTian:
> Re-review at the rebased head (`dd3cb538`, now MERGEABLE/CLEAN). The `spinup_nemo_gym_actor` extraction and SC wiring check out: the 2-tuple/4-tuple `setup_response_data` unpacking is correct (now with an explicit `assert len(...) == 2`), `dp_openai_server_base_urls` is guaranteed real by the `_should_use_nemo_gym` asserts, and `configure_vllm_for_router_replay` is a safe no-op when router replay is off. Note: the rebase **dropped** the original PR's grpo.py refactor — grpo.py keeps its own inline `_spinup_nemo_gym` (still importing `NemoGym`, never referencing `spinup_nemo_gym_actor`), so the new helper is currently a near-duplicate used only by the SC path. grpo's sync path is untouched (no behavior change), but the description's "shared function so both grpo.py and the SC setup can call it" no longer holds — worth either re-landing the grpo call-site swap or noting the duplication. No correctness bugs.
> 
> Six inline comments: one **medium** (the new e2e functional test is registered in no CI lane, so it never runs — regressed during the rebase), and five low/non-blocking (guard hoist, a test-assertion tightening, a docstring, a smoke-threshold note, and a router-replay tracking note). Enabling router replay (R3) on the SC async+TransferQueue path is only half-wired and fails silently — filed as #3327 (assigned) and added to the SC loose-ends tracker #2625.
> 
> Cross-checked against merged #3219/#3220 and base #3266 threads to avoid duplication.
> 
> _Reviewed by an agent team (rl-expert, bug-finder, test-agent, comment-reviewer, devil's-advocate), re-audited after rebase and cross-referenced against the PR stack. Generated by Claude Code_

Update the PR  doc describe.

nemo_rl/algorithms/single_controller_utils/setup.py
RayenTian: Done in [1d6a09ce9](https://github.com/NVIDIA-NeMo/RL/commit/1d6a09ce9ba421580b00fba87f7c8425f32f6c0d)
Add an item to #2625

RayenTian:
/ok to test [1d6a09c](https://github.com/NVIDIA-NeMo/RL/pull/3267/commits/1d6a09ce9ba421580b00fba87f7c8425f32f6c0d)

terrykong (approved):

PR #3131, opened 2026-07-09
feat: add OpenR1-Math-220k response dataset

tianyi-zhang-02:
## Summary

Adds `open-r1/OpenR1-Math-220k` as a built-in response dataset for verifiable-answer math training.

Changes:

- add `OpenR1Math220KDataset`
- register it as `OpenR1-Math-220k` in the response `DATASET_REGISTRY`
- add a dedicated unit test file (`tests/unit/data/datasets/test_openr1_math_dataset.py`) covering the Hugging Face dataset id/subset/split and emitted `messages` schema, plus a regression test for the default subset

## Traceability

Requested and approved in #3105.
Closes #3105.

Author: @tianyi-zhang-02

cc @yuki-97

## Notes

The wrapper follows the same lightweight pattern as existing math response datasets such as `GSM8KDataset` and `DeepScalerDataset`: it maps `problem` and `answer` into the existing two-message response dataset shape and leaves reward handling to the existing math environment/data processor path.

`subset` (the HF config name, matching the `ResponseDatasetConfig` / `ResponseDataset` convention) defaults to `default`, with `split` defaulting to `train`. Users can pass another OpenR1-Math-220k config such as `all` via `subset:` in the response dataset config.

## Review follow-ups (193bd86d)

- Renamed `config_name` → `subset` so a YAML `subset:` is honored instead of being silently swallowed by `**kwargs` (per @RayenTian's review).
- Moved the OpenR1-specific test out of `test_response_dataset_registry.py` into a dedicated `test_openr1_math_dataset.py`, following the `test_mmpr_tiny.py` pattern.

## Validation

- `ruff check` / `ruff format --check` on the three touched Python files — clean
- `python -m py_compile` on the three touched Python files — OK
- Mocked unit tests cover: subset threading (`subset: all` → HF config `all`), default subset fallback, and the emitted two-message schema

PR #3071, opened 2026-07-06
fix(checkpoint): break metric ties by recency in get_best_checkpoint_path

tianyi-zhang-02:
# What does this PR do ?

`CheckpointManager.get_best_checkpoint_path()` sorted candidate checkpoints on the **metric value alone**. Because `_load_checkpoint_history` builds its list from an unordered `glob.glob(...)` and Python's sort is stable, when two or more checkpoints **tie for the best metric value** the returned one was whichever `glob` happened to list first — often an **older** checkpoint.

This PR sorts by `(metric, step)` so the **most recent** checkpoint wins on ties, matching the tie-breaking already used by `remove_old_checkpoints()` (pinned by `test_remove_old_checkpoints_topk_bias_recent_if_equal`). It also adds a regression test.

This is a user-facing, documented selection API (no internal caller; six algorithm configs point users to it for final checkpoint export/eval), so the fix affects which checkpoint a user gets when exporting the "best" model on a metric tie.

# Issues

Closes #3067.

# Usage

```python
from nemo_rl.utils.checkpoint import CheckpointManager

# Two checkpoints tie for the best (lowest) loss=0.2; step 30 is more recent.
mgr = CheckpointManager({
    "enabled": True, "checkpoint_dir": d, "metric_name": "loss",
    "higher_is_better": False, "keep_top_k": None, "save_optimizer": True,
})
for step, loss in [(10, 0.2), (20, 0.9), (30, 0.2)]:
    mgr.finalize_checkpoint(mgr.init_tmp_checkpoint(step, {"loss": loss}))

mgr.get_best_checkpoint_path()  # before: step_10 (older tied-best); after: step_30
```

# Before your PR is "Ready for review"

**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests? — added `test_get_best_checkpoint_path_bias_recent_if_equal`, which fails on the previous behavior and passes with this fix.
- [x] Did you run the unit tests and functional tests locally? — validated `tests/unit/utils/test_checkpoint.py` logic locally; the added test forces the adversarial (older-first) history order via `monkeypatch` so it is deterministic regardless of filesystem `glob` ordering.
- [ ] Did you add or update any necessary documentation? — N/A (bug fix; no doc changes required).

# Additional Information

* Behavior is unchanged when there is no tie (distinct metric values still select the true best).
[...]

Issue #2840, opened 2026-06-16
refactor eval dataset to reuse train response dataset


yuki-97:
Eval datasets and response/train datasets in NeMo-RL frequently wrap the same underlying HuggingFace datasets, duplicating loader logic, dataset IDs, input-key conventions, and per-variant branching.

It's better to unify them to use a same one, so that we don't need to support them twice.

PR #3271, opened 2026-07-17
fix(data): warn when behavioral dataset config keys are silently ignored

### LINKED ISSUE #3270: Standard dataset config keys are silently ignored by built-in response dataset loaders
## Bug

`load_response_dataset` dispatches with `dataset_class(**data_config)`, and every built-in loader accepts `**kwargs` — so any standard `ResponseDatasetConfig` key a loader doesn't support is **silently swallowed**.

Minimal repro (exactly what the dispatcher does for `dataset_name: gsm8k, subset: socratic, split_validation_size: 0.05`):

```python
GSM8KDataset(split="train", subset="socratic", split_validation_size=0.05)
# → load_dataset("openai/gsm8k", "main")  — both keys silently dropped
```

- `GSM8KDataset` hardcodes the `main` config while `openai/gsm8k` also ships `socratic` — `subset:` is ignored with no error.
- 12 built-in loaders (aime, clevr, dapo_math, deepscaler, gsm8k, helpsteer3, nemogym_dataset, geometry3k, intent, oai_format_dataset, squad, refcoco) never call `split_train_validation`, so `split_validation_size`/`seed` silently do nothing — while the SFT guide documents them as generic dataset keys.

This generalizes the failure mode @RayenTian caught in the #3131 review: *"a user who sets `subset: all` in their YAML gets no error — it's swallowed by `**kwargs` and the default config loads instead."*

## Proposed fix

A dispatcher-level warning in `load_response_dataset`: after resolving the dataset class, warn if a user-set behavioral key (`subset`, `split`, `split_validation_size`, `seed`) is not accepted anywhere in the class `__init__` MRO. MRO-aware because e.g. the `intent.py` subclasses legitimately consume keys via `**kwargs` and forward to their base; `functools.partial` registry entries (AIME variants) are unwrapped. A warning rather than an error, for backward compatibility.

Happy to send the PR (guard + unit tests).

cc @RayenTian @yuki-97

tianyi-zhang-02:
## What does this PR do ?

`load_response_dataset` instantiates with `dataset_class(**data_config)`, and every built-in loader accepts `**kwargs` — so a standard `ResponseDatasetConfig` key the class doesn't support is silently swallowed. E.g. `subset: socratic` on `gsm8k` silently loads the hardcoded `main` config, and `split_validation_size`/`seed` silently do nothing on the 12 loaders that never call `split_train_validation` (while the SFT guide documents them as generic keys).

This adds a dispatcher-level guard, `warn_on_unsupported_dataset_config_keys`, that warns when a user-set behavioral key (`subset`, `split`, `split_validation_size`, `seed`) is not declared anywhere in the resolved class's `__init__` MRO.

Design notes:
- **MRO-aware**: `kwargs`-forwarding subclasses that delegate to a declaring base (the intent datasets) are not flagged.
- **`functools.partial` registry entries** (AIME variants) are unwrapped before inspection.
- **`None` values and falsy `split_validation_size` are skipped**, so documented config boilerplate (`subset: null`, `split_validation_size: 0`) stays quiet — none of the shipped `examples/configs` emit warnings.
- **Warning, not error**, for backward compatibility.
- **Known limitation**: an *external* dotted-path dataset class that consumes one of these keys via `kwargs.get(...)` without declaring it in any `__init__` signature would see a spurious warning. None of the built-in loaders do this (verified), and it's warning-only by design — but flagging it for review. Scoped to the response dispatcher; the preference dispatcher has the same pattern and can adopt the shared helper as a follow-up.

## Issues

Closes #3270. Generalizes the failure mode @RayenTian flagged in the #3131 review.

## Usage

```yaml
data:
  train:
    dataset_name: gsm8k
    subset: socratic   # before: silently loads "main"; after: UserWarning at dispatch
```

- [x] New tests: 7 unit cases for the guard (unsupported key warns; supported key quiet; `None`/`0` quiet; MRO forwarding quiet; `partial` unwrapped; non-class input safe) plus an end-to-end dispatcher test, in `tests/unit/data/datasets/test_response_dataset_registry.py` (registry-dispatch scope).
- [x] Validation: helper cases verified locally (all pass); `ruff check` / `ruff format --check` clean; `pyrefly==0.24.2` (CI-pinned) reports no new errors on the two whitelisted files touched.
- [x] Docs: N/A (warning makes existing documented behavior observable; no doc changes required).

cc @RayenTian @yuki-97


nemo_rl/data/datasets/utils.py
```diff
@@ -182,6 +185,72 @@ def resolve_external_dataset_class(dataset_name: str) -> type:
     return dataset_class
 
 
+# Behavioral keys from ResponseDatasetConfig that change *which data* a run
+# sees. If one of these is set but the resolved dataset class does not accept
+# it, the user silently gets different data than they configured.
+_BEHAVIORAL_DATASET_CONFIG_KEYS = ("seed", "split", "split_validation_size", "subset")
```
RayenTian: OOC, how this set of keys was chosen? ResponseDatasetConfig also exposes `data_path`, `input_key`, `output_key`, and `download_dir`, which are passed to dataset constructors and may likewise be silently swallowed by **kwargs. For example, `GSM8KDataset` would ignore a configured `data_path` or `input_key`. Should some or all of these keys be checked as well?

RayenTian:
Thanks a lot for adding this, @tianyi-zhang-02 !

Could we also apply warn_on_unsupported_dataset_config_keys() in load_preference_dataset()? It uses the same dataset_class(**data_config) pattern and is vulnerable to the same silent swallowing of unsupported keys—for example,
`Tulu3PreferenceDataset` would silently ignore a configured split.

Eval datasets are not yet using a fully unified loading format: some go through load_response_dataset(), while the remaining eval-specific loaders still use the legacy dispatcher. I think it is reasonable to defer the eval-side check and track it as pending under [#2840](https://github.com/NVIDIA-NeMo/RL/issues/2840), where the eval datasets are being consolidated into the response dataset registry.

nemo_rl/data/datasets/utils.py
tianyi-zhang-02: Honest answer is the original four were just the doc-promised generic HF-loading keys. There wasn't a deeper principle and you're right that the others are equally behavioral (`data_path` silently ignored on `gsm8k` is arguably worse than `subset`). Expanded in `fae1d7c` to all constructor-consumed keys from both `ResponseDatasetConfig` and `PreferenceDatasetConfig`: `data_path`, `input_key`, `output_key`, `download_dir`, `prompt_key`, `chosen_key`, `rejected_key` plus the original four. 

Audited every shipped `examples/configs` usage against the declaring loaders. No new warnings (OpenMathInstruct-2/`output_key`, NemoGymDataset/`data_path`, mmpr-tiny/`download_dir` all declare their keys). Deliberately left out `max_samples` since it isn't part of either TypedDict. :)

tianyi-zhang-02:
Done in `fae1d7c` — `load_preference_dataset` now applies the same `warn_on_unsupported_dataset_config_keys()` guard (the expanded key set covers the preference-side keys, so `Tulu3PreferenceDataset` with a configured `split` now warns), with an end-to-end dispatcher test added. Agreed on deferring the eval side to the #2840 registry consolidation. Would be happy to extend the guard there once that lands :D

RayenTian:
LGTM! Thanks a lot for adding this. Trigger CI now.

RayenTian:
/ok to test [ce56cf9](https://github.com/NVIDIA-NeMo/RL/pull/3271/commits/ce56cf9f19a8369c4fc4ca7e4c1dea3d956b459d)

tianyi-zhang-02:
cc @RayenTian. Thanks! The `CI quality check` failure is just the label gate - `TEST_LEVEL=none` because the PR has no `CI:L*` label. Could you add one (`CI:Lfast` matched #3071/#3131) and re-run `/ok to test ce56cf9`? The label is read at workflow start.

yuki-97 (approved):
LGTM, thanks for adding this @tianyi-zhang-02

PR #2608, opened 2026-05-28
feat: add checkpoint-engine refit interface and integrate NIXL

HollowMan6:
# What does this PR do?

Adds NIXL checkpoint-engine refit for transferring policy weights to
non-colocated vLLM generation workers.

The implementation:

- Integrates NIXL with the existing `refit_transport` and weight-synchronizer
  lifecycle.
- Supports DTensor, DTensor v2, and Megatron policy backends with vLLM
  generation.
- Supports CUDA UCX/RDMA transfers and host-pinned CPU transfers.
- Sizes each of the two transfer buffers from a configurable fraction of total
  GPU memory.
- Keeps communicators and buffers across refits by default, with optional buffer
  release after each refit.
- Provides a readable full-weight loading path and an optional
  destination-layout-aware sharded-expert path.
- Validates unsupported topology, quantization, and expert-layout combinations
  before transfer.
- Makes NIXL available in isolated policy and vLLM actor environments on Linux
  x86-64 and ARM64.

## Limitations

- Only non-colocated vLLM generation is supported.
- `shard_expert_weights: true` currently supports unquantized Triton MoE
  storage.
- Sharded FP8/MXFP8 expert refit is not supported. FP8 models can use
  full-weight refit with `shard_expert_weights: false`.
- Dynamic expert load balancing and redundant expert placement are not
  supported by the sharded-expert path.
- Support for other generation backends is tracked in #3288.

# Issues

None.

# Usage

Enable NIXL for non-colocated vLLM generation:

```yaml
policy:
  generation:
    backend: vllm
    colocated:
      enabled: false
    refit_transport: nixl
    refit_cfg:
      nixl:
        # Fraction of total GPU memory allocated to each transfer buffer.
        # NIXL uses two buffers, so 0.05 reserves up to 10% in total.
        update_weights_bucket_memory_ratio: 0.05
        device: cuda
        backend_name: UCX
        release_after_refit: false
        shard_expert_weights: false
```

For MoE models, enable destination-local expert transfer with:

```yaml
policy:
  generation:
    refit_cfg:
      nixl:
        shard_expert_weights: true
```

NIC and rail selection can be configured through `backend_init_params`.
Site-specific device names are intentionally not included in the minimal
configuration.

See:

- `docs/guides/refit.md`
- `docs/guides/checkpoint-engine-refit.md`
- `docs/design-docs/checkpoint-engines.md`

# Performance

DeepSeek-V3 BF16 was benchmarked with 32 Megatron policy nodes and four vLLM
[...]

PR #3286, opened 2026-07-19
feat(xtoken): replace DP token alignment with offset-based cluster alignment

avenkateshha:
## What

Moves cross-tokenizer alignment from Needleman-Wunsch DP to an **offset-based cluster aligner**, and extends it to **chat / instruct (and "thinking") data**.

Alignment pairs tokens by their character-span offsets (`return_offsets_mapping`): consecutive same-span tokens collapse into a cluster, a strict char-end walker pairs clusters covering the same character range, and special tokens (offset `(0,0)`) pair by role. `is_correct` is recomputed from canonicalized text, so `pair_is_correct` and the gold-loss exact partition keep their meaning.

## Scope / status

- [x] **Part 1 — offset-cluster alignment for raw-text KD** (replaces the DP aligner)
- [x] **Part 2 — chat / native-template per-assistant-message alignment**

## Part 1 — offset alignment (done)

- **`nemo_rl/algorithms/x_token/token_aligner.py`** — DP kernel, anchor segmentation, post-processing, and byte/mojibake canonicalization removed; the offset-cluster kernel (`align_by_offsets_cluster`) added. `canonical_token` (still used by `tools/x_token` and the `is_correct` mask) kept.
- **`nemo_rl/data/cross_tokenizer_collate.py`** — `_tokenize_batch` emits `offset_mapping`; `align()` receives student/teacher char offsets.
- **Tests** — `test_token_aligner.py` rewritten for the offset path; the collator's fake tokenizer emits char offsets.

## Part 2 — chat / native-template alignment (this PR)

Each model renders a conversation with its **own** chat template, so the two rendered texts — and their offsets — differ, and the full sequence can't be aligned directly. This part aligns **per assistant message**: isolate each side's assistant content, rebase offsets to a shared coordinate system, align the slice with the same `align_by_offsets_cluster` kernel, then map positions back to the full sequence. "Thinking" data is split into `reasoning` / `close` / `answer` regions (+ an EOT marker).

Adds:
- **`token_aligner.py`** — `align_one_offset_per_asst()` (per-turn slice → rebase → align → map-back) and `align_one_offset()` (canonical + NFC decode-fix `is_correct`).
- **`cross_tokenizer_collate.py`** — a `chat` mode (`__call__` → `_call_text` / `_call_chat`) with render / assistant-mask / native-region / lockstep-packing helpers and new constructor knobs.
- **`chat_templates.py`** — `find_rendered_message_content_span()`.
- **`processors.py`** — `chat_kd_processor()` (reads `messages` / `conversation`).
- **Config** — `collator_mode`, `include_thinking_in_loss`, `native_thinking_alignment`, `drop_first_assistant_chunk_kl`, `kd_alignment_regions`, `num_packed_rows`.

## Testing

- **Part 1**: 16/16 x_token unit tests pass (training container); 100-step multi-teacher GPU smoke (Qwen3-4B cross-tok + Llama-3.2-3B same-tok → Llama-3.2-1B, 2 nodes, default dataset) completed and saved a checkpoint. Per-step throughput: offset ~7.5 s/step vs DP ~25 s/step (~3.3× faster) — DP's Needleman-Wunsch alignment stalls the dataloader; the O(n) offset walk keeps it ahead.
- **Part 2**: chat-mode unit + smoke coverage to be added with the chat commits.

### DIFF
--- a/examples/configs/xtoken_off_policy_distillation.yaml
+++ b/examples/configs/xtoken_off_policy_distillation.yaml
@@ -240,6 +240,23 @@ data:
   max_input_seq_length: 2048
   shuffle: true
   num_workers: 4
+  # Collator mode. "text" (default) = raw-text KD; "chat" = per-assistant-message
+  # alignment over each side's rendered chat template (SFT/instruct data). The
+  # knobs below only apply when collator_mode: chat.
+  collator_mode: "text"
+  # Drop the opener content pair per assistant message from the KD alignment
+  # (leaves the CE token mask untouched).
+  drop_first_assistant_chunk_kl: false
+  # Include the reasoning body (thinking) in the loss (chat mode).
+  include_thinking_in_loss: false
+  # Split each assistant message into native reasoning/close/answer regions and
+  # align them independently (requires include_thinking_in_loss). Not yet wired.
+  native_thinking_alignment: false
+  # Restrict KD to a subset of {reasoning, close, answer, eot} (requires
+  # native_thinking_alignment).
+  kd_alignment_regions: null
+  # Lockstep multi-doc packing per row (chat mode). Only 1 is supported for now.
+  num_packed_rows: 1
   train:
     dataset_name: "arrow_text"
     processor: "kd_data_processor"
--- a/nemo_rl/algorithms/x_token/token_aligner.py
+++ b/nemo_rl/algorithms/x_token/token_aligner.py
@@ -36,7 +36,7 @@ Public surface:
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, List, Tuple
+from typing import Any, Collection, Dict, List, Tuple
 
 import torch
 
@@ -172,17 +172,34 @@ class TokenAligner:
             their training device via
             :func:`nemo_rl.algorithms.x_token.loss_utils.get_sparse_projection_matrix`
             or :func:`nemo_rl.algorithms.x_token.loss_utils.get_topk_projection`.
+        alignment_method: Alignment strategy the chat path uses.
+            ``"offset_cluster_decode_fix"`` (default) pairs tokens by char
+            offsets; ``"same_tokenizer_identity"`` asserts a 1-1 identity
+            pairing (student and teacher share a tokenizer). The raw-text
+            :meth:`align` path always uses offset-cluster.
     """
+
+    _SUPPORTED_ALIGNMENT_METHODS = (
+        "offset_cluster_decode_fix",
+        "same_tokenizer_identity",
+    )
 
     def __init__(
         self,
         student_tokenizer,
         teacher_tokenizer,
         projection_matrix_path: str,
+        alignment_method: str = "offset_cluster_decode_fix",
     ):
+        if alignment_method not in self._SUPPORTED_ALIGNMENT_METHODS:
+            raise ValueError(
+                f"alignment_method must be one of "
+                f"{self._SUPPORTED_ALIGNMENT_METHODS!r}, got {alignment_method!r}"
+            )
         self.student_tokenizer = student_tokenizer
         self.teacher_tokenizer = teacher_tokenizer
         self.projection_matrix_path = projection_matrix_path
+        self.alignment_method = alignment_method
 
     def align(
         self,
@@ -398,6 +415,349 @@ class TokenAligner:
                 )
             )
         return out
+
+    # ------------------------------------------------------------------ #
+    # Chat / instruct path: per-assistant-message alignment.
+    # These return raw 7-tuples (s_tokens, t_tokens, s_start, s_end, t_start,
+    # t_end, is_correct); the chat collator converts them to AlignmentPair.
+    # ------------------------------------------------------------------ #
+    def align_one_offset(
+        self,
+        student_ids: List[int],
+        teacher_ids: List[int],
+        student_offsets: List[Tuple[int, int]],
+        teacher_offsets: List[Tuple[int, int]],
+    ) -> List[Tuple[Any, ...]]:
+        """Single-sample offset-cluster alignment with the decode-fix mask.
+
+        Returns the raw 7-tuple pair list. Used by the chat collator for
+        per-document alignment inside a packed row. ``is_correct`` starts from
+        the canonical-text match; pairs that fail are re-checked against their
+        NFC-decoded text (catches CJK / whitespace-split asymmetry).
+        """
+        pairs = align_by_offsets_cluster(
+            student_ids,
+            student_offsets,
+            self.student_tokenizer,
+            teacher_ids,
+            teacher_offsets,
+            self.teacher_tokenizer,
+        )
+        if pairs:
+            pairs_6 = [(p[0], p[1], p[2], p[3], p[4], p[5]) for p in pairs]
+            mask = _decode_fix_correct_mask(
+                pairs_6,
+                student_ids_seq=student_ids,
+                teacher_ids_seq=teacher_ids,
+                student_tokenizer=self.student_tokenizer,
+                teacher_tokenizer=self.teacher_tokenizer,
+            )
+            pairs = [
+                (p[0], p[1], p[2], p[3], p[4], p[5], m) for p, m in zip(pairs_6, mask)
+            ]
+        return pairs
+
+    def align_one_offset_per_asst(
+        self,
+        student_ids: List[int],
+        student_offsets: List[Tuple[int, int]],
+        student_asst_char_spans: List[Tuple[int, int]],
+        teacher_ids: List[int],
+        teacher_offsets: List[Tuple[int, int]],
+        teacher_asst_char_spans: List[Tuple[int, int]],
+        student_asst_mask: List[int] | None = None,
+        teacher_asst_mask: List[int] | None = None,
+        student_alignment_regions: List[Dict[str, Tuple[int, int]]] | None = None,
+        teacher_alignment_regions: List[Dict[str, Tuple[int, int]]] | None = None,
+        student_eot_indices: List[int] | None = None,
+        teacher_eot_indices: List[int] | None = None,
+        drop_first_content_pair: bool = False,
+        included_regions: Collection[str] | None = None,
+    ) -> List[Tuple[Any, ...]]:
+        """Per-assistant-message offset (or strict-identity) alignment.
+
+        Used by the chat collator where each side has its own rendered text,
+        so full-sequence offsets are in different coordinate systems. For each
+        assistant message:
+
+        1. Find student/teacher tokens whose offsets fall fully inside paired
+           content regions. By default there is one whole-message region;
+           native-thinking mode supplies reasoning, close, and answer regions.
+        2. Rebase each paired region independently so both token slices share
+           the same coordinate system.
+        3. Run :func:`align_by_offsets_cluster` on the slices.
+        4. Translate slice-local positions back to full-sequence positions.
+
+        Scaffold and user-content tokens are not aligned; the caller's
+        ``token_mask = attention_mask * assistant_mask`` zeros them out of loss.
+
+        ``drop_first_content_pair`` omits the first pair with both student and
+        teacher spans per message (removes cross-tokenizer KL on the opener
+        chunk while leaving the CE token mask untouched). ``included_regions``
+        optionally limits KD to a subset of ``reasoning``, ``close``,
+        ``answer``, ``eot``. Returns a flat 7-tuple list with full-sequence
+        position indices.
+        """
+        if len(student_asst_char_spans) != len(teacher_asst_char_spans):
+            raise ValueError(
+                f"asst message count mismatch: "
+                f"{len(student_asst_char_spans)} vs "
+                f"{len(teacher_asst_char_spans)}"
+            )
+        if (student_alignment_regions is None) != (teacher_alignment_regions is None):
+            raise ValueError("student/teacher alignment regions must be paired")
+        if student_alignment_regions is not None and (
+            len(student_alignment_regions) != len(student_asst_char_spans)
+            or len(teacher_alignment_regions or []) != len(teacher_asst_char_spans)
+        ):
+            raise ValueError("alignment-region count must match assistant turns")
+        if (student_eot_indices is None) != (teacher_eot_indices is None):
+            raise ValueError("student/teacher EOT index lists must be paired")
+        if student_eot_indices is not None and (
+            len(student_eot_indices) != len(student_asst_char_spans)
+            or len(teacher_eot_indices or []) != len(teacher_asst_char_spans)
+        ):
+            raise ValueError("EOT index count must match assistant turns")
+
+        included_region_set = (
+            frozenset(included_regions) if included_regions is not None else None
+        )
+        valid_region_names = {"reasoning", "close", "answer", "eot"}
+        if included_region_set is not None:
+            invalid = included_region_set - valid_region_names
+            if invalid:
+                raise ValueError(
+                    "included_regions contains unsupported values: "
+                    f"{sorted(invalid)!r}; expected a subset of "
+                    f"{sorted(valid_region_names)!r}"
+                )
+            if not included_region_set:
+                raise ValueError("included_regions must not be empty")
+            if student_alignment_regions is None:
+                raise ValueError(
+                    "included_regions requires native semantic alignment regions"
+                )
+
+        combined: List[Tuple[Any, ...]] = []
+        for turn_i, ((s_start_c, s_end_c), (t_start_c, t_end_c)) in enumerate(
+            zip(student_asst_char_spans, teacher_asst_char_spans)
+        ):
+            if student_alignment_regions is None:
+                paired_regions = [
+                    ("content", (s_start_c, s_end_c), (t_start_c, t_end_c))
+                ]
+            else:
+                student_turn_regions = student_alignment_regions[turn_i]
+                teacher_turn_regions = (teacher_alignment_regions or [])[turn_i]
+                paired_regions = []
+                for name in ("reasoning", "close", "answer"):
+                    s_region = student_turn_regions.get(name)
+                    t_region = teacher_turn_regions.get(name)
+                    if (s_region is None) != (t_region is None):
+                        raise ValueError(
+                            f"semantic region {name!r} is missing on one side"
+                        )
+                    if (
+                        s_region is not None
+                        and t_region is not None
+                        and (included_region_set is None or name in included_region_set)
+                    ):
+                        paired_regions.append((name, s_region, t_region))
+
+            turn_pairs: List[Tuple[Any, ...]] = []
+            for _region_name, (s_region_start, s_region_end), (
+                t_region_start,
+                t_region_end,
+            ) in paired_regions:
+                # Boundary-crossing tokens are deliberately omitted; in native
+                # mode this keeps template-only formatting whitespace out of KD.
+                s_idx = [
+                    i
+                    for i, (cs, ce) in enumerate(student_offsets)
+                    if cs >= s_region_start
+                    and ce <= s_region_end
+                    and ce > cs
+                    and (student_asst_mask is None or student_asst_mask[i] == 1)
+                ]
+                t_idx = [
+                    j
+                    for j, (cs, ce) in enumerate(teacher_offsets)
+                    if cs >= t_region_start
+                    and ce <= t_region_end
+                    and ce > cs
+                    and (teacher_asst_mask is None or teacher_asst_mask[j] == 1)
+                ]
+                if self.alignment_method == "same_tokenizer_identity" and (
+                    bool(s_idx) != bool(t_idx)
+                ):
+                    raise ValueError(
+                        "same_tokenizer_identity supervised-region presence "
+                        f"mismatch at assistant turn {turn_i}: "
+                        f"student_tokens={len(s_idx)}, teacher_tokens={len(t_idx)}"
+                    )
+                if not s_idx or not t_idx:
+                    continue
+
+                s_slice_ids = [student_ids[i] for i in s_idx]
+                t_slice_ids = [teacher_ids[j] for j in t_idx]
+                s_slice_off = [
+                    (
+                        student_offsets[i][0] - s_region_start,
+                        student_offsets[i][1] - s_region_start,
+                    )
+                    for i in s_idx
+                ]
+                t_slice_off = [
+                    (
+                        teacher_offsets[j][0] - t_region_start,
+                        teacher_offsets[j][1] - t_region_start,
+                    )
+                    for j in t_idx
+                ]
+
+                if self.alignment_method == "same_tokenizer_identity":
+                    if len(s_slice_ids) != len(t_slice_ids):
+                        raise ValueError(
+                            "same_tokenizer_identity supervised token-count "
+                            f"mismatch at assistant turn {turn_i}: "
+                            f"student={len(s_slice_ids)}, teacher={len(t_slice_ids)}"
+                        )
+                    slice_pairs = []
+                    for position, (
+                        student_id,
+                        teacher_id,
+                        student_offset,
+                        teacher_offset,
+                    ) in enumerate(
+                        zip(s_slice_ids, t_slice_ids, s_slice_off, t_slice_off)
+                    ):
+                        if student_id != teacher_id:
+                            raise ValueError(
+                                "same_tokenizer_identity token mismatch at "
+                                f"assistant turn {turn_i}, region position "
+                                f"{position}: student={student_id}, "
+                                f"teacher={teacher_id}"
+                            )
+                        if student_offset != teacher_offset:
+                            raise ValueError(
+                                "same_tokenizer_identity offset mismatch at "
+                                f"assistant turn {turn_i}, region position "
+                                f"{position}: student={student_offset}, "
+                                f"teacher={teacher_offset}"
+                            )
+                        slice_pairs.append(
+                            (
+                                [student_id],
+                                [teacher_id],
+                                position,
+                                position + 1,
+                                position,
+                                position + 1,
+                                True,
+                            )
+                        )
+                else:
+                    slice_pairs = align_by_offsets_cluster(
+                        s_slice_ids,
+                        s_slice_off,
+                        self.student_tokenizer,
+                        t_slice_ids,
+                        t_slice_off,
+                        self.teacher_tokenizer,
+                    )
+                if slice_pairs and self.alignment_method != "same_tokenizer_identity":
+                    pairs_6 = [
+                        (p[0], p[1], p[2], p[3], p[4], p[5]) for p in slice_pairs
+                    ]
+                    correct_mask = _decode_fix_correct_mask(
+                        pairs_6,
+                        student_ids_seq=s_slice_ids,
+                        teacher_ids_seq=t_slice_ids,
+                        student_tokenizer=self.student_tokenizer,
+                        teacher_tokenizer=self.teacher_tokenizer,
+                    )
+                    slice_pairs = [
+                        (p[0], p[1], p[2], p[3], p[4], p[5], is_correct)
+                        for p, is_correct in zip(pairs_6, correct_mask)
+                    ]
+
+                for s_toks, t_toks, s0, s1, t0, t1, ok in slice_pairs:
+                    full_s0 = s_idx[s0] if s0 != -1 else -1
+                    full_s1 = s_idx[s1 - 1] + 1 if s0 != -1 else -1
+                    full_t0 = t_idx[t0] if t0 != -1 else -1
+                    full_t1 = t_idx[t1 - 1] + 1 if t0 != -1 else -1
+                    turn_pairs.append(
+                        (s_toks, t_toks, full_s0, full_s1, full_t0, full_t1, ok)
+                    )
+
+            if drop_first_content_pair:
+                for pair_i, pair in enumerate(turn_pairs):
+                    if pair[2] != -1 and pair[4] != -1:
+                        del turn_pairs[pair_i]
+                        break
+            combined.extend(turn_pairs)
+
+            # Synthetic EOT pair: include the chat-template end-of-turn marker
+            # that terminates each assistant turn. Without it, the predictor
+            # whose label is EOT gets chunk_id=-1 after the kl shift and drops
+            # out of loss, so the model never learns to terminate its turn.
+            # Native-thinking mode passes the index explicitly because some
+            # templates insert whitespace between the answer boundary and EOT.
+            if student_eot_indices is not None:
+                s_eot_idx = student_eot_indices[turn_i]
+                t_eot_idx = (teacher_eot_indices or [])[turn_i]
+                s_eot_idx = s_eot_idx if s_eot_idx >= 0 else None
+                t_eot_idx = t_eot_idx if t_eot_idx >= 0 else None
+            else:
+                s_eot_idx = next(
+                    (
+                        i
+                        for i, (cs, ce) in enumerate(student_offsets)
+                        if cs == s_end_c and ce > cs
+                    ),
+                    None,
+                )
+                t_eot_idx = next(
+                    (
+                        j
+                        for j, (cs, ce) in enumerate(teacher_offsets)
+                        if cs == t_end_c and ce > cs
+                    ),
+                    None,
+                )
+            if self.alignment_method == "same_tokenizer_identity" and (
+                (s_eot_idx is None) != (t_eot_idx is None)
+            ):
+                raise ValueError(
+                    "same_tokenizer_identity EOT presence mismatch at "
+                    f"assistant turn {turn_i}: student={s_eot_idx}, "
+                    f"teacher={t_eot_idx}"
+                )
+            include_eot = included_region_set is None or "eot" in included_region_set
+            if include_eot and s_eot_idx is not None and t_eot_idx is not None:
+                if (
+                    self.alignment_method == "same_tokenizer_identity"
+                    and student_ids[s_eot_idx] != teacher_ids[t_eot_idx]
+                ):
+                    raise ValueError(
+                        "same_tokenizer_identity EOT token mismatch at "
+                        f"assistant turn {turn_i}: "
+                        f"student={student_ids[s_eot_idx]}, "
+                        f"teacher={teacher_ids[t_eot_idx]}"
+                    )
+                combined.append(
+                    (
+                        [student_ids[s_eot_idx]],
+                        [teacher_ids[t_eot_idx]],
+                        s_eot_idx,
+                        s_eot_idx + 1,
+                        t_eot_idx,
+                        t_eot_idx + 1,
+                        True,
+                    )
+                )
+
+        return combined
 
 
 # =====================================================================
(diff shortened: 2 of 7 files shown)


examples/configs/xtoken_off_policy_distillation.yaml
```diff
@@ -240,6 +240,23 @@ data:
...
   num_workers: 4
+  # Collator mode. "text" (default) = raw-text KD; "chat" = per-assistant-message
+  # alignment over each side's rendered chat template (SFT/instruct data). The
+  # knobs below only apply when collator_mode: chat.
+  collator_mode: "text"
+  # Drop the opener content pair per assistant message from the KD alignment
+  # (leaves the CE token mask untouched).
+  drop_first_assistant_chunk_kl: false
+  # Include the reasoning body (thinking) in the loss (chat mode).
+  include_thinking_in_loss: false
```
RayenTian, line 254: [`examples/configs/xtoken_off_policy_distillation.yaml:251`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/examples/configs/xtoken_off_policy_distillation.yaml#L251)

Would it make sense to hold back the parts that aren't reachable yet, and land them with the PR that implements them?

**Config knobs.** `native_thinking_alignment` and `num_packed_rows != 1` raise `NotImplementedError`; `kd_alignment_regions` requires the former, so it can't be set either; and `include_thinking_in_loss` is assigned to `self` but never read, so thinking tokens always enter the loss — the opposite of what the comment above suggests. That leaves `collator_mode` and `drop_first_assistant_chunk_kl` as the two that currently do something. Removing the other four would be pure deletion: the YAML keys, the `data_config.get(...)` lines, the constructor params, and the [guard block at `cross_tokenizer_collate.py:120-144`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/data/cross_tokenizer_collate.py#L120-L144). Those guards have to come out anyway when the feature lands.

**`alignment_method`.** This one I wasn't sure how to read — is it meant to be user-facing, or internal? Nothing sets it today: the only [`TokenAligner(...)` call site](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/algorithms/xtoken_off_policy_distillation.py#L305) doesn't pass it, there's no YAML key, and the only assignments anywhere are two tests patching it onto a mock. So it's always the default, and `same_tokenizer_identity` (~67 lines, no tests) can't be reached. If it's meant to be configurable it needs an entry point; if it's internal, could you say what's expected to select the identity path? Worth noting the same-tokenizer case already short-circuits earlier — `aligners[i] is None` skips projection and alignment entirely — so I may be missing the intended use.

**Unused code.** [`align_one_offset`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/algorithms/x_token/token_aligner.py#L424) has no callers either.

Whichever way `alignment_method` goes, dropping `same_tokenizer_identity` for now would also take out the comparisons at `token_aligner.py` [`:590`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/algorithms/x_token/token_aligner.py#L590), `:618`, `:668`, `:728`, `:739`, plus [`_CHAT_ALIGNMENT_METHODS`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/data/cross_tokenizer_collate.py#L49) in the collator — that set duplicates [`_SUPPORTED_ALIGNMENT_METHODS`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/algorithms/x_token/token_aligner.py#L182), so the check at [`:107-112`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/data/cross_tokenizer_collate.py#L107-L112) can't currently fail.

Optional and entirely your call: if you'd rather keep both strategies, would subclasses be easier to live with than the string flag? A base aligner with `OffsetClusterAligner` / `SameTokenizerIdentityAligner` would keep each strategy's validation in its own class and remove the five scattered comparisons. No strong opinion — just noting that adding a third strategy today means finding all five again.

examples/configs/xtoken_off_policy_distillation.yaml
```diff
@@ -240,6 +240,23 @@ data:
   max_input_seq_length: 2048
   shuffle: true
   num_workers: 4
+  # Collator mode. "text" (default) = raw-text KD; "chat" = per-assistant-message
+  # alignment over each side's rendered chat template (SFT/instruct data). The
+  # knobs below only apply when collator_mode: chat.
+  collator_mode: "text"
+  # Drop the opener content pair per assistant message from the KD alignment
+  # (leaves the CE token mask untouched).
+  drop_first_assistant_chunk_kl: false
```
RayenTian, line 252: [`examples/configs/xtoken_off_policy_distillation.yaml:249`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/examples/configs/xtoken_off_policy_distillation.yaml#L249)

Small structural suggestion, mostly for when the held-back knobs come back: would an `aligner:` block under each `teachers[i]` be a better home for them than `data:`?

The aligner is constructed per teacher, and [`projection_matrix_path`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/examples/configs/xtoken_off_policy_distillation.yaml#L143) — itself a `TokenAligner` constructor arg — already lives there. The [note at the top of `loss_fn:`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/examples/configs/xtoken_off_policy_distillation.yaml#L24) says it was moved out of the shared block for exactly that reason.

```yaml
teachers:
  - model_name: "Qwen/Qwen3-4B"
    aligner:
      projection_matrix_path: null
      drop_first_assistant_chunk_kl: false
```

`drop_first_assistant_chunk_kl` is the one that could move today — it's passed straight through to `align_one_offset_per_asst`. `collator_mode` seems right where it is under `data:`, since it's about how a datum becomes tokens.

RayenTian (commented):
Took a close look at the offset-cluster rewrite — the O(n) walk replacing Needleman-Wunsch is a nice simplification, and the ~3.3x per-step number with a 100-step 2-node smoke behind it is exactly the kind of evidence that makes this easy to reason about. Three questions below, two of them about scope rather than the algorithm.

**Could you rebase onto current `main` when you get a chance?** The branch point (`0f629c5b1`, 2026-07-18) is about 245 commits behind now.

There's no merge conflict, so this isn't about that. The reason I ask is that this PR changes what the aligner emits, while its consumers live elsewhere — [`valid_chunk_mask`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/algorithms/x_token/loss_utils.py) in `loss_utils.py` and the `exact_token_match_only` gate in [`loss_functions.py`](https://github.com/NVIDIA-NeMo/RL/blob/e2722b1737f4f8d6f4dd89659fc392083ae04ca6/nemo_rl/algorithms/loss/loss_functions.py). If either moved in the last two months, a clean merge could still end up subtly wrong.

Would you be able to re-run the 100-step 2-node smoke afterwards, just to confirm the ~3.3x still holds? Since the tooling is already set up, that might also be a convenient point to add a chat-mode run.

_Generated by Claude Code_

examples/configs/xtoken_off_policy_distillation.yaml
avenkateshha: Thanks — these config knobs are intentional scaffolding for the v6 follow-up that will be stacked on this PR. The downstream work already uses `include_thinking_in_loss`, `native_thinking_alignment`, `kd_alignment_regions`, and `alignment_method`, and builds on the same collator/aligner interface. Keeping the interface and fail-loud guards here avoids reintroducing that surface during the follow-up rebase and lets this PR stay focused on the offset-cluster and whole-message chat alignment foundation. I can clarify the YAML comments to label these as reserved for the v6 follow-up rather than supported in this PR. Would that address the scope concern?

avenkateshha: Thanks — I revisited the alignment portion and agreed with your read of the current branch. Addressed in ca3366fc5:

- removed `alignment_method`, `_SUPPORTED_ALIGNMENT_METHODS`, and the unreachable `same_tokenizer_identity` path;
- removed the duplicated `_CHAT_ALIGNMENT_METHODS` validation;
- renamed `align_one_offset()` to `_align_one_offset()` and made `align_one_offset_per_asst()` call it, so the per-region offset/decode-fix logic is no longer duplicated;
- removed test strategy setup and added focused coverage for the production helper call and same-tokenizer bypass.

Same-tokenizer teachers still use `aligner=None` and direct positional KL. The focused aligner/collator tests and real-tokenizer raw/chat parity pass.

I left the reserved v6 flags and `drop_first_assistant_chunk_kl` placement unchanged for this pass, per the agreed scope.

examples/configs/xtoken_off_policy_distillation.yaml
avenkateshha: Thanks — addressed in 45650a0946938b0a878bcfb32b67dca4510cf72f.

- Added a typed `TeacherAlignerConfig` under each `teachers[i]`, containing `projection_matrix_path` and `drop_first_assistant_chunk_kl` with centralized defaults.
- Routed the drop-first setting as a full per-teacher list indexed by the original teacher index, so interleaved same-tokenizer teachers keep their `None` alignment slot without index drift.
- Kept `collator_mode` under `data`.
- Updated the checked-in configs, recipes, CLI overrides, docs, and tests, including a cross/same/cross multi-teacher case.

Because the old root `teachers[i].projection_matrix_path` spelling shipped in v0.7.0, it has a narrow warning migration; conflicting old/new values fail clearly and serialization emits only the nested path. The shared data-level drop flag was PR-only, so that old location is rejected rather than migrated.

nemo_rl/data/cross_tokenizer_collate.py
```diff
@@ -172,23 +243,218 @@ def __call__(self, batch: List[DatumSpec]) -> BatchedDataDict[Any]:
...
+
+        # Locate each assistant message's content in the rendered text, scanning
+        # left-to-right so repeated content in later turns doesn't rematch early.
+        asst_char_spans: List[tuple[int, int]] = []
+        cursor = 0
+        for message in messages:
+            span = find_rendered_message_content_span(
+                rendered, message.get("content", ""), cursor
+            )
+            if span is None:
```
yuki-97, line 386: Locating assistant turns by substring-searching each side's own render for the raw `content` breaks on any template that *transforms* content rather than merely trimming it.

The helper only tries whitespace variants ([chat_templates.py:41](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/chat_templates.py#L41)) and the caller silently skips a miss. Qwen3-4B's template rewrites assistant content containing `</think>` and drops the reasoning from every turn at or before the last user query, so the original string is in the render under no variant. Reproduced end-to-end, driving `_render_and_tokenize_chat` unmodified on a two-turn conversation whose first assistant message carries a `<think>` block:

```
teacher Qwen/Qwen3-4B renders turn 1 as:  <|im_start|>assistant\nIt is 4.<|im_end|>   # reasoning gone
  student assistant spans: 2      teacher assistant spans: 1
  -> ValueError: asst message count mismatch: 2 vs 1     (raised inside the collate worker)

with Qwen3 on both sides the counts agree at 1 and nothing raises:
  -> turn 1 leaves the loss entirely; assistant_mask covers 12 of 50 tokens, so those
     positions get token_mask=0 -- no CE and no KD, silently
```

A third case needs no rewrite at all: when the two sides match *different* whitespace variants, each region rebases from its own origin and the turn is silently distilled against character-shifted teacher chunks, or loses its pairs outright — which of the two depends on the leading whitespace — under the shipped `exact_token_match_only: false`.

**Suggest fix:**

1. Search one canonicalized key on both sides (e.g. always `content.strip()`), so the two regions cannot rebase from different origins. The helper already returns the variant it matched and it is discarded at [cross_tokenizer_collate.py:390](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L390).
2. Raise rather than `continue` when an assistant message is not found.

nemo_rl/data/cross_tokenizer_collate.py
```diff
@@ -172,23 +243,218 @@ def __call__(self, batch: List[DatumSpec]) -> BatchedDataDict[Any]:
...
+                continue
+            cursor = span[1]
+            if message.get("role") == "assistant":
+                asst_char_spans.append((span[0], span[1]))
+
+        assistant_mask = [
```
yuki-97, line 392: The chat-template end-of-turn token is excluded from the loss mask, so the student gets no CE gradient for terminating its turn.

`assistant_mask` is 1 only strictly inside an assistant *content* span, and the EOT marker sits just past `span[1]` — which is how the aligner locates it ([token_aligner.py:632](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L632)). `token_mask` ([cross_tokenizer_collate.py:283](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L283)) then zeroes it and `ce_label_mask`'s shift drops the one prediction whose label is EOT; the same mask is the same-vocab teacher's KD reduction ([loss_functions.py:2561](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/loss/loss_functions.py#L2561)), so that path loses it too. The KL side already fixed this with the synthetic EOT pair at [token_aligner.py:617-657](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L617-L657), and elsewhere in the repo the terminator is in the mask.

**Suggest fix:** widen `assistant_mask` to include each turn's EOT token, mirroring the `cs == span[1]` lookup the aligner already does at [token_aligner.py:629-644](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L629-L644).

nemo_rl/data/cross_tokenizer_collate.py
```diff
@@ -172,23 +243,218 @@ def __call__(self, batch: List[DatumSpec]) -> BatchedDataDict[Any]:
...
+            aligner._drop_padding(
+                alignment,
+                student_attention_mask=student_attention_mask,
+                teacher_attention_mask=teacher_attention_mask,
+            )
+            out[f"teacher_{i}_input_ids"] = teacher_input_ids
+            out[f"teacher_{i}_input_lengths"] = teacher_attention_mask.sum(
+                dim=-1
+            ).long()
+            out[f"teacher_{i}_token_mask"] = teacher_attention_mask.long()
```
yuki-97, line 346: In chat mode the teacher's `token_mask` covers the whole rendered conversation while the student's covers assistant content only.

[cross_tokenizer_collate.py:283](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L283) narrows the student mask by the assistant mask; this line leaves the teacher's at the plain attention mask, and the teacher's own assistant mask is discarded in its padded form at [cross_tokenizer_collate.py:305](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L305). `_teacher_score_inputs` ([loss_functions.py:2789-2795](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/loss/loss_functions.py#L2789-L2795)) feeds this mask to a masked *mean*, so a cross-tokenizer teacher is scored over its template scaffolding while a same-vocab teacher is scored over assistant tokens only — and those scores drive `select_teacher` and the `sum_weights_metric` weights that multiply each teacher's KD term.

**Suggest fix:**

1. Capture the fourth return at [cross_tokenizer_collate.py:305](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L305): `teacher_input_ids, teacher_attention_mask, _, teacher_asst_mask = self._pad_chat_batch(...)`.
2. Emit `(teacher_attention_mask * teacher_asst_mask).long()` on this line.

nemo_rl/data/cross_tokenizer_collate.py
```diff
@@ -65,6 +66,10 @@ class CrossTokenizerCollator:
         ctx_length_student: Hard tokenization length cap on the student
             side (also the padded sequence length of the student tensor).
         ctx_length_teachers: Per-teacher tokenization length caps.
+        drop_first_assistant_chunk_kl_by_teacher: Per-teacher flags controlling
```
yuki-97, line 69: nit: the module docstring and this `Args:` block both describe only the text path.

[cross_tokenizer_collate.py:18-23](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L18-L23) says the collator tokenizes "(no chat template, no special handling)" and calls `TokenAligner.align`; chat mode calls `apply_chat_template` at [cross_tokenizer_collate.py:367](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L367) and `align_one_offset_per_asst` at [cross_tokenizer_collate.py:316](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L316).

The `Args:` block documents 8 of the 13 constructor parameters. Missing, all added by this PR:

- `mode`
- `include_thinking_in_loss`
- `native_thinking_alignment`
- `kd_alignment_regions`
- `num_packed_rows`

**Suggest fix:**

1. Make docstring items 1 and 2 mode-aware.
2. Add the five entries above.

nemo_rl/algorithms/x_token/token_aligner.py
```diff
@@ -1152,165 +730,348 @@ def canonical_token(token: str, *, enabled: bool = True) -> str:
...
 
-    Returns:
-        ``(merged, ranges)`` with one ``(orig_start, orig_end)`` entry per
-        output token. Every entry in :data:`_MULTI_TOKEN_ARTIFACT_FIXES`
-        rewrites to a single replacement token, so each merge contributes
-        exactly one range covering the matched pattern.
+# When one tokenizer has pad_token_id == eos_token_id (e.g. Llama-3.2 with the
+# collator fallback that sets pad_token = eos_token if undefined), trailing pad
+# positions get role "eos"; the other tokenizer with a separate pad_token_id
+# tags its trailing pads as "pad". Treating these roles as equivalent pairs
```
yuki-97: nit: the benefit this comment describes cannot occur — both callers wipe pad positions immediately after alignment.

`_drop_padding` resets every `attention_mask == 0` position to `chunk_id = -1` and partition `False` ([token_aligner.py:261-264](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L261-L264)), and `valid_chunk_mask` then drops any chunk that is empty on either side ([loss_utils.py:650](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/loss_utils.py#L650)). It runs on both paths — inside `align()` at [token_aligner.py:236-240](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L236-L240) and from the chat collator at [cross_tokenizer_collate.py:337](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L337) — so no pair built from padding survives to carry KL, and the student cannot gain "training signal at every trailing pad". What the set actually changes is the orphan count.

**Suggest fix:** restate the comment in terms of what the set does (pair a mismatched pad/eos role 1-1 instead of emitting two orphans), or drop the set along with it.

nemo_rl/algorithms/xtoken_off_policy_distillation.py
```diff
@@ -320,6 +401,15 @@ def setup(
         make_seq_div_by_teachers=[
             tc["make_sequence_length_divisible_by"] for tc in teacher_configs
         ],
+        drop_first_assistant_chunk_kl_by_teacher=[
+            teacher.aligner.drop_first_assistant_chunk_kl for teacher in teachers
+        ],
+        # Shared chat/instruct knobs; default to the raw-text path.
+        mode=data_config.get("collator_mode", "text"),
```
yuki-97, line 408: These five keys are xtoken-only but live under `data:`, whose schema is shared with every other algorithm — so there is nowhere to declare them, and their defaults land here instead of in YAML.

`data:` binds `DataConfig` ([__init__.py:69-103](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/__init__.py#L69-L103)), which `sft`, `dpo`, `grpo`, `ppo`, `rm`, `mpo`, `distillation` and `eval` bind too. Undeclared, each key has three default sources: the exemplar, this `.get()`, and the collator's parameter defaults ([cross_tokenizer_collate.py:89-93](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L89-L93)). Only one exemplar carries them, so `data.collator_mode=chat` against `xtoken_multiteacher_off_policy_distillation.yaml` is rejected by OmegaConf struct mode.

**Suggest fix:**

1. Move the five to a typed top-level `collator:` block with the defaults on the class, mirroring `TeacherAlignerConfig` ([xtoken_off_policy_distillation.py:162-175](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/xtoken_off_policy_distillation.py#L162-L175)).
2. Add the block to both exemplars.

tests/test_suites/llm/distillation-xtoken-off-policy-qwen3-4b-to-llama3.2-1b-1n8g-dtensor-tp4cp2.sh
```diff
@@ -39,7 +39,7 @@ fi
 uv run examples/run_xtoken_off_policy_distillation.py \
     --config $CONFIG_PATH \
     distillation.max_num_steps=$MAX_STEPS \
-    teachers.0.projection_matrix_path=$PROJ_PATH \
+    teachers.0.aligner.projection_matrix_path=$PROJ_PATH \
```
yuki-97: Both nightly recipes still gate on KD magnitudes that were measured against the DP aligner this PR removes.

`train/kl_loss_t0` and `validation/kl_loss` are computed from the chunk partition — `student_chunk_id`, `teacher_chunk_id`, `pair_valid` — that `align_by_offsets_cluster` now produces, so a different pairing moves them. The bounds at [distillation-xtoken-off-policy-qwen3-4b-to-llama3.2-1b-1n8g-dtensor-tp4cp2.sh:68-73](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/tests/test_suites/llm/distillation-xtoken-off-policy-qwen3-4b-to-llama3.2-1b-1n8g-dtensor-tp4cp2.sh#L68-L73) are unchanged, as is the multiteacher two-sided band `0.08 < mean(kl_loss_t1) < 0.17` at [distillation-xtoken-off-policy-multiteacher-qwen3-4b-llama3.2-3b-to-llama3.2-1b-1n8g-dtensor-tp4cp2.sh:88-91](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/tests/test_suites/llm/distillation-xtoken-off-policy-multiteacher-qwen3-4b-llama3.2-3b-to-llama3.2-1b-1n8g-dtensor-tp4cp2.sh#L88-L91) and the comments recording where those numbers came from. Nothing on this PR exercises them: the only xtoken run reachable from `/ok to test` is the 3-step `train/loss < 5` functional test.

**Suggest fix:** paste the 100-step run's `train/loss` (step 1 and mean of the last 5), `train/accuracy`, `kl_loss_t0`, `kl_loss_t1` and `validation/kl_loss` (steps 0 and 100) into the PR description and say which of these bounds they clear; re-center any bound and its narrating comment that moved.

nemo_rl/data/cross_tokenizer_collate.py
```diff
@@ -172,23 +243,218 @@ def __call__(self, batch: List[DatumSpec]) -> BatchedDataDict[Any]:
...
+                    ),
+                )
+                per_sample_pairs.append(
+                    [
+                        AlignmentPair(p[0], p[1], p[2], p[3], p[4], p[5], p[6])
+                        for p in raw
+                    ]
+                )
+
+            alignment = aligner._pairs_to_batch(per_sample_pairs, b=b, t_s=t_s, t_t=t_t)
```
yuki-97, line 336: nit: three small things about the chat path's seam with the aligner.

1. `align()` ends with this same `_pairs_to_batch` then `_drop_padding` pair ([token_aligner.py:235-240](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L235-L240)), and these two lines are the only callers of either private outside the aligner — a further step appended to that tail is silently skipped here.
2. The chat helpers return raw 7-tuples, unpacked positionally at [cross_tokenizer_collate.py:331](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/data/cross_tokenizer_collate.py#L331), while the guard protecting that order is `set`-based ([test_token_aligner.py:125](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/tests/unit/algorithms/x_token/test_token_aligner.py#L125)) and so blind to it.
3. `_pad_chat_batch`'s third return value has no reader.

**Suggest fix:**

1. *(fixes 1 and 2)* Move the block above behind a public `TokenAligner.align_chat(...)` returning `AlignmentBatch`, with the chat helpers returning `AlignmentPair` — this also turns the positional `p[2]` / `pairs[0][2:6]` indexing at [test_token_aligner.py:320-433](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/tests/unit/algorithms/x_token/test_token_aligner.py#L320-L433) red until it moves to attribute access.
2. *(fixes 3)* Drop the unused offsets slot from `_pad_chat_batch`.

nemo_rl/algorithms/x_token/token_aligner.py
```diff
@@ -410,417 +323,59 @@ def _pairs_to_batch(
...
+                s_toks,
+                t_toks,
+                s_start,
+                s_end,
+                t_start,
+                t_end,
+                _paired,
+            ) in raw_pairs
+        ]
+        for pair, m in zip(pairs, self._alignment_mask(pairs)):
```
yuki-97, line 376: nit: only the chat path gets the NFC / whitespace tiers, though they are a property of the tokenizer pair rather than of chat data.

[token_aligner.py:1041-1045](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L1041-L1045) repeats [token_aligner.py:383-400](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L383-L400) verbatim and adds three fallback tiers on top (CJK byte encodings, `['Ġ', '\n']` vs `['Ġ\n']`); this line is `_alignment_mask`'s only caller. `is_correct` feeds `pair_is_correct`, which drops KD pairs under `exact_token_match_only` ([loss_functions.py:2942](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/loss/loss_functions.py#L2942)).

**Suggest fix:** use the same mask here, then delete `_alignment_mask`.

```suggestion
        mask = _decode_fix_correct_mask(
            raw_pairs,
            student_ids_seq=student_ids,
            teacher_ids_seq=teacher_ids,
            student_tokenizer=self.student_tokenizer,
            teacher_tokenizer=self.teacher_tokenizer,
        )
        for pair, m in zip(pairs, mask):
            pair.is_correct = m
```

tests/test_suites/llm/distillation-xtoken-off-policy-multiteacher-qwen3-4b-llama3.2-3b-to-llama3.2-1b-1n8g-dtensor-tp4cp2.sh
yuki-97, line 1: could you paste the curve or metric result at PR description to show the two nightly tests could still pass after changing the alignment?

nemo_rl/algorithms/xtoken_off_policy_distillation.py
```diff
@@ -206,6 +272,20 @@ class MasterConfig(BaseModel, extra="allow"):
     cluster: ClusterConfig
     checkpointing: CheckpointingConfig
 
+    @model_validator(mode="before")
+    @classmethod
+    def _reject_shared_drop_first_assistant_chunk_kl(cls, value: Any) -> Any:
```
yuki-97, line 277: This guard rejects a spelling that only ever existed inside this PR.

`data.drop_first_assistant_chunk_kl` was added by 216faf105 and moved to the per-teacher field by 45650a094, both on this branch; `git log -S` over main returns nothing, so no released config can carry it. The docstring's own "unreleased" says as much.

**Suggest fix:** delete the validator and its test at [test_xtoken_off_policy_distillation.py:366](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/tests/unit/algorithms/x_token/test_xtoken_off_policy_distillation.py#L366).

nemo_rl/algorithms/xtoken_off_policy_distillation.py
```diff
@@ -181,18 +196,69 @@ class TeacherConfig(BaseModel, extra="allow"):
...
 
-    projection_matrix_path: Optional[str] = None
+    aligner: TeacherAlignerConfig = Field(default_factory=TeacherAlignerConfig)
     weight: float = 1.0
     gold_loss: Optional[bool] = None
     xtoken_loss: Optional[bool] = None
 
+    @model_validator(mode="before")
+    @classmethod
+    def _migrate_legacy_projection_matrix_path(cls, value: Any) -> Any:
```
yuki-97, line 206: A released key renamed by this PR is migrated silently here, where the repo's precedent for that case is to reject it.

`policy.generation.checkpoint_engine` — also a real released key, added in #2608 and later superseded — raises at [config.py:280-285](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/models/generation/vllm/config.py#L280-L285) rather than rewriting the value. This shim migrates and emits a `FutureWarning` instead, and it only covers the YAML entry path: the CLI form that v0.7.0's README and guide both publish (`teachers.0.projection_matrix_path=...`) is rejected by OmegaConf struct mode inside `parse_hydra_overrides`, before `MasterConfig` is built. So the two entry paths already disagree.

**Suggest fix:**

1. Raise here the way `normalize_vllm_refit_config` does, naming `teachers[i].aligner.projection_matrix_path`, and drop the migration branches.
2. Rename to `_reject_legacy_projection_matrix_path` and update its docstring.

nemo_rl/data/cross_tokenizer_collate.py
```diff
@@ -172,23 +243,218 @@ def __call__(self, batch: List[DatumSpec]) -> BatchedDataDict[Any]:
...
+                alignment,
+                student_attention_mask=student_attention_mask,
+                teacher_attention_mask=teacher_attention_mask,
+            )
+            out[f"teacher_{i}_input_ids"] = teacher_input_ids
+            out[f"teacher_{i}_input_lengths"] = teacher_attention_mask.sum(
+                dim=-1
+            ).long()
+            out[f"teacher_{i}_token_mask"] = teacher_attention_mask.long()
+            for f in dataclass_fields(alignment):
```
yuki-97, line 347: nit: these are never read by anything. All pre-existing — not introduced by this PR.

- `AlignmentBatch.student_exact_partition_mask`
- `AlignmentBatch.teacher_exact_partition_mask`
- `AlignmentBatch.num_chunks`
- `alignment_from_flat_batch` ([loss_utils.py:63](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/loss_utils.py#L63)) — no production caller

`localize_alignment` ([loss_utils.py:322](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/loss_utils.py#L322)) builds `LocalizedAlignment` from `student_chunk_id` / `teacher_chunk_id` / `pair_valid` / `pair_is_correct` only.

**Suggest fix:** delete them.

nemo_rl/algorithms/x_token/token_aligner.py
```diff
@@ -1152,165 +730,348 @@ def canonical_token(token: str, *, enabled: bool = True) -> str:
...
-        rewrites to a single replacement token, so each merge contributes
-        exactly one range covering the matched pattern.
+# When one tokenizer has pad_token_id == eos_token_id (e.g. Llama-3.2 with the
+# collator fallback that sets pad_token = eos_token if undefined), trailing pad
+# positions get role "eos"; the other tokenizer with a separate pad_token_id
+# tags its trailing pads as "pad". Treating these roles as equivalent pairs
+# those positions 1<->1, giving the student KL signal at pad positions.
+# Without this set, role-based pairing silently drops the mismatched roles and
+# the student loses training signal at every trailing pad on the mismatched
+# side.
```
yuki-97, line 752: nit: the claimed KL signal at pad positions can't happen — `_drop_padding` ([token_aligner.py:261-264](https://github.com/NVIDIA-NeMo/RL/blob/45650a0946938b0a878bcfb32b67dca4510cf72f/nemo_rl/algorithms/x_token/token_aligner.py#L261-L264)) zeroes every masked position right after alignment.

**Suggest fix:**

```suggestion
# pad_token may fall back to eos_token on one side only, so trailing pads get
# role "eos" there and "pad" on the other; pairing them 1<->1 keeps max_pairs
# from doubling on those positions.
```

yuki-97 (commented):
Nice work — the offset-cluster kernel is a real simplification over the Needleman-Wunsch path it replaces. The strict char-end walker is O(n) and every branch of both loops advances an index, so it cannot spin, and the ~3.3x per-step number has a 100-step 2-node multi-teacher smoke behind it. Routing the drop-first flag as a per-teacher list indexed by the original teacher index — so interleaved same-tokenizer teachers keep their `None` slot — is the right call.

**Scope:** the full diff at 45650a0946938b0a878bcfb32b67dca4510cf72f — the aligner rewrite, the new `chat` collator mode, the per-teacher config move, and the nightly / functional drivers.

**Checked:**

- Offset-kernel arithmetic — the cluster walker and the slice-to-full-sequence remap, driven on hand-built offsets.
- The chat path against the real Qwen3 and Llama chat templates, rendered rather than read.
- Config surface — every new key traced to a consumer, and both exemplars, the recipe YAMLs and the `.sh` overrides replayed through OmegaConf struct mode.
- Removals — every deleted symbol grepped at head, and every `AlignmentBatch` field traced to a reader.
- Loss-side coupling — `is_correct`, the chunk partition and both `token_mask` keys followed to their consumers.

**Remaining comments** are on the chat path's assistant-span matching and its two loss masks, the new `data.*` keys' schema and exemplar coverage, the two config-migration validators, and a few duplicated or stale spots in the aligner.

PR #3290, opened 2026-07-20
feat(nemotron-omni): add Megatron Nano image GRPO support

aroshanghias-nvd:
# What does this PR do?

Add Megatron-based Nemotron Nano Omni image-GRPO support using NeMo-RL's existing sequence-packing data path and the dedicated MBridge `NemotronOmniModel`.

> **Fresh review requested:** The branch is now synchronized with current `main` and uses the reviewed data-boundary packing design. The earlier model-owned-packing implementation is obsolete; please review this series from the beginning.
>
> **Cross-repository dependency:** [NVIDIA-NeMo/Megatron-Bridge#4885](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4885) has merged. The required Nemotron Omni support is included in the MBridge pin inherited from NeMo-RL `main` (`573e088c`); this PR does not change the submodule pin.

The ownership boundary is:

- NeMo-RL groups and packs the processor-expanded token sequences using its existing Megatron data pipeline and constructs `PackedSeqParams`.
- NeMo-RL passes the full packed THD row and multimodal inputs to MBridge.
- MBridge `NemotronOmniModel` creates and inserts media embeddings, then selects the context-parallel portion of the merged embedding sequence.
- Packing does not run inside model forward, and this path does not use the historical Nemotron Omni `LLaVAModel` collapse/expand behavior.

The PR includes:

- capability-based dispatch for models that consume caller-packed full THD input and slice context-parallel inputs after embedding insertion;
- a clear error for legacy LLaVA-layout Nemotron Omni checkpoints;
- dynamic-resolution and historical tiled-image metadata handling;
- generation compatibility for raw policy logprobs, context-length capping, and RADIO LayerScale initialization;
- maintained single-node Clevr and four-node MMPR Nano image-GRPO recipes.

The branch is synchronized with NeMo-RL `main`'s vLLM 0.25.1 dependency. This PR does not change external vLLM source or the vLLM dependency pin; its compatibility behavior is implemented in NeMo-RL's existing integration and runtime-patch layer.

# Issues

No public issue.

# Usage

The maintained recipes added by this PR are:

- `examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n8g-megatron-tp8ep8.v1.yaml`
- `examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-mmpr-4n8g-megatron-tp8ep16.v1.yaml`

# Validation

- A production-shaped four-node Nano image-GRPO parity run with vLLM 0.25.1 exercised generation, rewards, raw policy logprobs, training, repeated vLLM refit, and checkpointing through step 23. The [PR #3290 validation run](https://wandb.ai/joc/nemotron-omni-main-migration/runs/nemotron-omni-main-production-100-step-pr3290-vllm0251-seeded-665fd69eb-20260804T184500Z) tracks the [production packing-refactor reference](https://wandb.ai/joc/nemotron-omni-main-migration/runs/nemotron-omni-main-production-100-step-packing-refactor-rerun-20260731T095405Z) closely, without the prior TMPE spikes.
- The matched A/B used validation-only per-call deterministic seeding instrumentation; that instrumentation is not included in this PR.
- The focused distributed Megatron suite passed: CP2 checkpoint, TP/SP, TP2+CP2, padded multi-row attention, and PP2 (`5 passed`).
- Focused vLLM integration tests passed (`8 passed`).
- `uv lock --check` and `git diff --check` passed.
- Protected NeMo-RL CI is required before merge.

# Scope and follow-up

This PR covers the Nano image/text path. Caller-packed context parallelism with MTP is deliberately unsupported until the separate Super/MTP integration adds and validates the required MTP-mask alignment. Audio and video support is layered on the MBridge foundation separately.

- [x] Read and followed the [contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md).
- [x] Added the necessary unit and functional tests.
- [x] Ran focused local and distributed validation.
- [x] Updated `docs/guides/nemotron-3-nano-omni.md` with the Megatron architecture, supported scope, checkpoint compatibility, recipes, and usage.


nemo_rl/data/multimodal_utils.py
```diff
@@ -108,8 +113,51 @@ def as_tensor(
...
             return None
-        else:
-            return torch.cat(non_none_tensors, dim=self.dim_to_pack).to(device)
+
+        # Some multimodal processors produce a different shape per prompt,
+        # such as dynamic-resolution images, variable-frame videos, or audio
+        # feature sequences. Concatenation already permits the packing
+        # dimension to vary; when explicitly requested, pad every other
+        # dimension to the largest size in the batch.
+        if self.pad_to_max_shape:
```
rohitrango, line 130: The `pad_to_max_shape` can be very wasteful if images of very different sizes are batched and packed together. A much more elegant idea is to simply pre-`flatten` the tensors before packing along dim 0, storing the unpacked sizes as a metadata similar to `PackedSeqParams` and unpacking them. 

Instead of `pad_to_max_shape: bool = False,`  you can have `pre_flatten: bool = False` and
```python
if self.pre_flatten:
     shapes = [x.shape for x in non_none_tensors]
     flattened = [x.view(-1) for x in non_none_tensors]
     #  pack flattened along dim 0
```

nemo_rl/models/generation/vllm/vllm_worker.py
```diff
@@ -471,11 +484,21 @@ def _load_model(self, bundle_indices, seed):
...
+            ),
             **vllm_kwargs,
         )
 
         self._create_engine(llm_kwargs)
+        # Nemotron Omni checkpoints fold RADIO LayerScale into adjacent weights,
+        # while stock vLLM still allocates ls1/ls2 parameters. Initialize those
+        # parameters before colocated level-1 sleep releases their CUDA storage;
+        # mutating them later from prepare_refit_info corrupts sleep/wake state.
+        self.llm.collective_rpc("_initialize_nemotron_omni_radio_layerscale")
```
yfw: [`nemo_rl/models/generation/vllm/vllm_worker.py:501`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/nemo_rl/models/generation/vllm/vllm_worker.py#L501)

This `collective_rpc` runs in the shared `BaseVllmGenerationWorker._load_model`, which the async worker also inherits without overriding. On the async path `self.llm` is an `AsyncLLM`, whose `collective_rpc` is `async def` ([vLLM 0.20.0, `async_llm.py#L928`](https://github.com/vllm-project/vllm/blob/v0.20.0/vllm/v1/engine/async_llm.py#L928)) — unlike the sync `LLM.collective_rpc` ([`llm.py#L646`](https://github.com/vllm-project/vllm/blob/v0.20.0/vllm/entrypoints/llm.py#L646)). Since the call isn't awaited, with `async_engine: true` it produces a never-awaited coroutine: every async run (any model) gets a "coroutine was never awaited" warning with the RPC silently skipped, and a future async Omni recipe would run with dummy-initialized RADIO LayerScale weights — the fail-loud guard inside `_initialize_nemotron_omni_radio_layerscale` never executes. The shipped recipes use the sync engine, so nothing breaks today, but this is the first `collective_rpc` in `_load_model`, so there's no async override to catch it.

Consider dispatching per engine type — e.g. keep it in the sync path's engine setup and `await self.llm.collective_rpc(...)` in an async override (matching the existing awaited call sites in `vllm_worker_async.py`), or assert the Omni architectures are incompatible with `async_engine` for now.

nemo_rl/data/processors.py
```diff
@@ -614,15 +614,39 @@ def vlm_hf_data_processor(
...
+        )
+
+    # imgs_sizes is not always declared in model_input_names by bundled image
+    # processors, so append it explicitly when present. RADIO uses temporal
+    # patching even for still images and requires one num_frames=1 entry per
+    # image/tile.
     if "imgs_sizes" in message and "imgs_sizes" not in multimodal_keys:
         multimodal_keys.append("imgs_sizes")
+    if "imgs_sizes" in message and "num_frames" not in message:
+        message["num_frames"] = torch.ones(len(message["imgs_sizes"]), dtype=torch.long)
```
yfw, line 641: [`nemo_rl/data/processors.py:641`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/nemo_rl/data/processors.py#L641)

**Regression for the shipped AutoModel Omni recipes** (`vlm_grpo-nemotron-omni-30ba3b-{clevr,mmpr}-*-automodel-ep8.v1`, [`nightly.txt:48-49`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/tests/test_suites/nightly.txt#L48)) — this raises `TypeError` on the first forward. The chain, all from source:

1. The AutoModel path loads the HF model `nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-BF16`, whose top-level [`NemotronH_Nano_Omni_Reasoning_V3.forward`](https://huggingface.co/nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-BF16/blob/24e67ea000b7c2837fc8f9488aa2008524fac8ba/modeling.py#L172) is a plain InternVL-style signature — `pixel_values, input_ids, attention_mask, position_ids, image_flags, past_key_values, labels, inputs_embeds, use_cache, ...` — with **no `num_frames`, no `imgs_sizes`, and no `**kwargs`**. `imgs_sizes` appears nowhere in that model's `modeling.py`/`processing.py`; it and RADIO `num_frames` are Megatron-`NemotronOmniModel`-only concepts.

2. This block force-appends `imgs_sizes` (synthesized at L622 from the 4-D tiled `pixel_values`) and `num_frames` to `multimodal_keys` for **any** `_uses_image_placeholder` processor — which includes `NemotronH_Nano_Omni_Reasoning_V3Processor`, the processor the AutoModel recipe uses. Pre-PR neither key reached the forward (`num_frames` didn't exist; `imgs_sizes` wasn't in `model_input_names`), which is why the recipe worked.

3. Nothing filters them out: the DTensor worker spreads `vlm_kwargs` into `model_args` and calls [`self.model(**model_args)`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L823) with no signature/allowlist check — it only hand-strips `flash_attn_kwargs` for multimodal ([L820](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L820)). So `self.model(..., imgs_sizes=..., num_frames=...)` → `TypeError: forward() got an unexpected keyword argument`.

That existing `del model_args["flash_attn_kwargs"]` is the tell — this HF forward was already known to reject extra kwargs, and the PR adds two more without stripping them.

**Suggested fix:** filter `vlm_kwargs` down to the model's accepted forward inputs at the worker boundary (right where `flash_attn_kwargs` is already special-cased), so keys the model doesn't declare are dropped. The Megatron worker keeps `num_frames`/`imgs_sizes` (its model declares them); the DTensor/AutoModel worker drops them. Please also re-run the AutoModel CLEVR recipe to confirm — the signature mismatch is unambiguous, but a smoke step will catch anything downstream.

examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n8g-megatron-tp8ep8.v1.yaml
```diff
@@ -0,0 +1,55 @@
+defaults: ../../vlm_grpo_3B_megatron.yaml
```
yfw, line 1: [`vlm_grpo-nemotron-omni-30ba3b-clevr-1n8g-megatron-tp8ep8.v1.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n8g-megatron-tp8ep8.v1.yaml)

The two new Megatron recipes don't have matching driver scripts under `tests/test_suites/vlm/` or entries in [`tests/test_suites/nightly.txt`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/tests/test_suites/nightly.txt#L48). Could you add `vlm_grpo-nemotron-omni-30ba3b-clevr-1n8g-megatron-tp8ep8.v1.sh` and `vlm_grpo-nemotron-omni-30ba3b-mmpr-4n8g-megatron-tp8ep16.v1.sh` mirroring the AutoModel drivers and append both to `nightly.txt`?

tests/unit/data/test_multimodal_dict.py
```diff
@@ -349,3 +349,44 @@ def test_packedtensor_as_tensor_with_mixed_none_and_tensors():
     out = pt.as_tensor()
     expected = torch.cat([t1, t3], dim=0)
     assert torch.equal(out, expected)
+
+
+def test_packedtensor_pads_mixed_dynamic_resolution_images():
```
yfw, line 354: [`tests/unit/data/test_multimodal_dict.py:354`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/tests/unit/data/test_multimodal_dict.py#L354)

Nice coverage of the happy paths (image + generic audio/video ranks). Three error branches of `pad_to_max_shape` are still uncovered — the mismatched-rank `ValueError`, the out-of-range `IndexError`, negative-`dim_to_pack` normalization — plus slice-preserving the flag. These four tests were verified passing locally:

```python
def test_pad_to_max_shape_rejects_mismatched_ranks():
    with pytest.raises(ValueError, match="same rank"):
        PackedTensor([torch.ones(1, 3, 4), torch.ones(1, 3)],
                     dim_to_pack=0, pad_to_max_shape=True).as_tensor()


def test_pad_to_max_shape_rejects_out_of_range_dim():
    with pytest.raises(IndexError, match="dim_to_pack=3 is invalid"):
        PackedTensor([torch.ones(1, 3, 4), torch.ones(2, 3, 4)],
                     dim_to_pack=3, pad_to_max_shape=True).as_tensor()


def test_pad_to_max_shape_supports_negative_pack_dim():
    packed = PackedTensor([torch.ones(2, 3, 1), 2 * torch.ones(4, 3, 1)],
                          dim_to_pack=-3, pad_to_max_shape=True).as_tensor()
    assert packed.shape == (6, 3, 1)


def test_slice_preserves_pad_to_max_shape_flag():
    pt = PackedTensor([torch.ones(1, 3, 2, 4), 2 * torch.ones(1, 3, 4, 2)],
                      dim_to_pack=0, pad_to_max_shape=True)
    sliced = pt.slice([0, 1])
    assert sliced.pad_to_max_shape is True
    assert sliced.as_tensor().shape == (2, 3, 4, 4)
```

tests/unit/models/generation/test_vllm_backend.py
```diff
@@ -259,3 +259,82 @@ def test_load_mtp_weights_from_disk_raises_when_mtp_weights_missing(
...
+        vllm_config=SimpleNamespace(
+            model_config=SimpleNamespace(architectures=[architecture])
+        ),
+        model=SimpleNamespace(vision_model=_TinyRadioVisionModel()),
+    )
+    return ext
+
+
+@pytest.mark.vllm
+def test_initialize_folded_nemotron_radio_layerscale_while_awake():
```
yfw: [`tests/unit/models/generation/test_vllm_backend.py:289`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/tests/unit/models/generation/test_vllm_backend.py#L289)

The two `RuntimeError` guards in `_initialize_nemotron_omni_radio_layerscale` (missing `vision_model`, zero `ls1`/`ls2` params) are uncovered — and since the guards are the fail-loud protection against silently-corrupted RADIO inference, they're worth locking in. These two tests reuse the file's `_make_extension_with_radio` helper and were verified passing locally with `--vllm-only`:

```python
@pytest.mark.vllm
def test_initialize_raises_when_vision_model_missing():
    ext = _make_extension_with_radio("NemotronH_Nano_Omni_Reasoning_V3")
    ext.model_runner.model = SimpleNamespace(vision_model=None)
    with pytest.raises(RuntimeError, match="no vision_model"):
        ext._initialize_nemotron_omni_radio_layerscale()


@pytest.mark.vllm
def test_initialize_raises_when_no_layerscale_params():
    ext = _make_extension_with_radio("NemotronH_Nano_Omni_Reasoning_V3")

    class _VisionNoLS(torch.nn.Module):
        def __init__(self):
            super().__init__()
            self.config = SimpleNamespace(initializer_factor=1.0)
            self.projection = torch.nn.Parameter(torch.ones(4))

    ext.model_runner.model = SimpleNamespace(vision_model=_VisionNoLS())
    with pytest.raises(RuntimeError, match="no RADIO ls1/ls2"):
        ext._initialize_nemotron_omni_radio_layerscale()
```

nemo_rl/data/multimodal_utils.py
```diff
@@ -82,6 +83,8 @@ def __init__(
         self,
         tensors: Union[torch.Tensor, list[Optional[torch.Tensor]], list[None]],
         dim_to_pack: int,
+        *,
+        pad_to_max_shape: bool = False,
```
yfw, line 86: [`nemo_rl/data/multimodal_utils.py:87`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/nemo_rl/data/multimodal_utils.py#L87)

The new keyword-only `pad_to_max_shape` is non-obvious from the signature (it pads all non-pack dims to the batch max, and requires uniform rank). A short docstring on `__init__` would help:

```python
def __init__(
    self,
    tensors: Union[torch.Tensor, list[Optional[torch.Tensor]], list[None]],
    dim_to_pack: int,
    *,
    pad_to_max_shape: bool = False,
) -> None:
    """Wrap a list of per-item tensors packed along ``dim_to_pack``.

    Args:
        tensors: A single tensor or a list of per-item tensors (entries may
            be ``None`` for items lacking this modality).
        dim_to_pack: Dimension along which ``as_tensor`` concatenates.
        pad_to_max_shape: If ``True``, pad every non-packing dimension of
            each tensor up to the batch-wide maximum before concatenating
            (for variable-shape media such as dynamic-resolution images).
            All tensors must share the same rank.
    """
```

uv.lock
```diff
@@ -2662,7 +2662,7 @@ linting = [
 ]
 no-pypi-wheels = [
     { name = "emerging-optimizers", git = "https://github.com/NVIDIA-NeMo/Emerging-Optimizers.git?rev=v0.2.0" },
-    { name = "flash-mla", git = "https://github.com/deepseek-ai/FlashMLA?rev=9edee0c022cd0938148a18e334203b0aab43aa19" },
+    { name = "flash-mla", git = "https://github.com/deepseek-ai/FlashMLA?rev=nv_dev" },
```
yfw, line 3497: [`uv.lock:2665`](https://github.com/NVIDIA-NeMo/RL/blob/df0db31e6db7da15053b924b05c4a7577c6cbf14/uv.lock#L2665)

This changes the flash-mla source from an immutable commit (`?rev=9edee0c0...`) to the floating branch `?rev=nv_dev`, and the lockfile carries no resolved commit for it — so installs can drift as the branch moves, which defeats the lockfile's reproducibility. This was likely swept in with the Megatron-Bridge lock regeneration. Could you pin `nv_dev` at its current commit, or confirm the floating pin is intended? (The mamba-ssm and transformer-engine changes are commit-pinned and look fine.)

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -172,13 +165,78 @@ def prepare_refit_info(self, state_dict_info: dict[str, Any]) -> None:
                 e.g. {tensor_name: (shape, dtype)}
         """
         self.state_dict_info = state_dict_info  # pyrefly: ignore[implicitly-defined-attribute]  This class does not define __init__ so assignments like this should be ignored
+        self._validate_nemotron_omni_radio_layerscale_refit(state_dict_info)
+
+    def _is_nemotron_omni(self) -> bool:
```
yfw: Is the check we care about that it's a RADIO model? If that's the case, would it make more sense to check for RADIO model instead of a hardcoded list of architectures?

nemo_rl/models/megatron/setup.py
```diff
@@ -323,6 +328,41 @@ def _get_hf_config_overrides_hash(overrides: dict[str, Any]) -> str:
     return hashlib.sha256(canonical.encode("utf-8")).hexdigest()[:12]
 
 
+def _validate_nemotron_omni_megatron_lm_layout(
+    hf_config: Any,
+    checkpoint_dir: str,
+) -> None:
+    """Reject legacy LLaVA checkpoints before building the canonical Omni model.
```
yfw: I'm wondering if this check and the validate_model_contract below is necessary considering we previously didn't support a "legacy" path on nemo-rl `main`. That is, this is the first implementation of the nemotron omni model in `main` and since it ships with the correct version of Megatron-Bridge, it seems like this would not be an issue?

nemo_rl/models/megatron/setup.py
```diff
@@ -539,6 +580,14 @@ def setup_model_config(
         model_cfg = cfg_from_pretrained.model
         cfg_from_pretrained.logger = LoggerConfig()
 
+    # New and legacy Nemotron Omni checkpoints once serialized the same
+    # provider class name with different semantics. Providers that implement
+    # this hook must prove their explicit, versioned contract before setup
+    # applies runtime parallelism or constructs any model layers.
+    validate_model_contract = getattr(model_cfg, "validate_model_contract", None)
```
yfw: Same comment as above. Since we never had support for "legacy" Nemotron Omni checkpoints in nemo-rl `main`, is this necessary?

nemo_rl/data/multimodal_utils.py
rohitrango, line 130: The `pad_to_max_shape` does not have a corresponding unpad operation in `PackedTensor` and the radio model.

For example, if a `(128, 128)` and a `(512, 512)` images were packed together, then the first image needs to be re-cropped back to the original size. This doesn't seem to be done in the Radio model either.  Could you help me clarify this? I'm uncertain if this will lead to a silent bug where the padded parts are patchified and embedded which might fail silently and provide incorrect image patches to process.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -172,13 +165,78 @@ def prepare_refit_info(self, state_dict_info: dict[str, Any]) -> None:
...
+                "checkpoint values with the folded-checkpoint identity behavior."
+            )
 
-    def prepare_sparse_delta_refit_info(
-        self, state_dict_info: dict[str, tuple[tuple[int, ...], torch.dtype]]
-    ) -> list[str]:
-        """Reserve scratch space and report weights that require overwrite."""
-        applier = self._get_sparse_delta_applier()
-        return sorted(applier.discover_native_skips(state_dict_info))
+    def _initialize_nemotron_omni_radio_layerscale(self) -> int:
```
rohitrango: would we want to fold these functions into `vllm/model_patches/omni.py`?

nemo_rl/models/generation/vllm/vllm_worker.py
```diff
@@ -51,6 +48,19 @@
 logger = logging.getLogger(__name__)
 
 
+def _context_capped_max_new_tokens(
+    *, configured_max_new_tokens: int, input_length: int, max_model_len: int
+) -> int:
+    """Cap generation so the training prompt and response fit the context."""
+    remaining_context = max_model_len - input_length
+    if remaining_context <= 0:
+        raise ValueError(
```
rohitrango, line 73: For large multimodal datasets, the image/video prompt might be large enough to cancel the run. It might be better to mask these samples instead and throw a warning instead?

nemo_rl/models/generation/vllm/vllm_worker.py
aroshanghias-nvd: Good catch. I moved the initialization into the engine-specific post-init paths: the sync worker calls it normally, while the async worker awaits it. Both execute immediately after engine creation and before sleep, refit, or generation. I also added sync and async call-order coverage.

nemo_rl/data/processors.py
aroshanghias-nvd: Good catch. imgs_sizes and num_frames are required by the Megatron model but currently flow through the shared processor into AutoModel as well. I’ll filter multimodal kwargs at the AutoModel boundary using the underlying model’s forward() signature, while preserving all keys for models accepting **kwargs. This keeps the Megatron path unchanged and prevents MBridge-only metadata from reaching the HF Nemotron model.

examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n8g-megatron-tp8ep8.v1.yaml
aroshanghias-nvd: Done.

tests/unit/data/test_multimodal_dict.py
aroshanghias-nvd: Added coverage for mismatched tensor ranks, invalid and negative packing dimensions, and preservation of pad_to_max_shape through slicing. No production behavior changed.

tests/unit/models/generation/test_vllm_backend.py
aroshanghias-nvd: Added coverage for both fail-loud RADIO LayerScale guards: missing vision_model and missing ls1/ls2 parameters. No production behavior changed.

nemo_rl/data/multimodal_utils.py
aroshanghias-nvd: Good suggestion. I added an __init__ docstring documenting the packing dimension, padding of non-packing dimensions, support for None entries, and the uniform-rank requirement.

uv.lock
aroshanghias-nvd: This no longer appears in the PR diff after rebasing onto the latest main; main already uses the nv_dev source. No change is needed in this PR.

nemo_rl/models/generation/vllm/vllm_backend.py
aroshanghias-nvd: The condition is narrower than simply using RADIO. These specific Nemotron Omni checkpoint architectures fold RADIO LayerScale into adjacent weights and omit explicit ls1/ls2 tensors. Other RADIO models may retain meaningful LayerScale parameters, so generic RADIO detection could incorrectly overwrite them. The explicit architecture allowlist keeps this workaround fail-closed. I’ll rename the helper to make the folded-LayerScale contract clearer.

nemo_rl/models/megatron/setup.py
aroshanghias-nvd: Agreed. I removed the NeMo-RL-specific legacy layout and contract checks and will rely on the current MBridge provider and normal checkpoint validation.

nemo_rl/data/multimodal_utils.py
aroshanghias-nvd: We also pack and preserve imgs_sizes alongside pixel_values. In MBridge, NemotronOmniModel uses those original dimensions to crop each padded image before RADIO patchification, so the padded regions are never encoded. I added a test covering this behavior.

nemo_rl/models/generation/vllm/vllm_backend.py
aroshanghias-nvd: Yes, that separation makes sense. Moved the Nemotron Omni architecture detection and RADIO LayerScale validation/initialization into vllm/model_patches/omni.py. The worker extension now retains only the thin method required as the collective_rpc() entry point. Behavior is unchanged.

nemo_rl/models/generation/vllm/vllm_worker.py
aroshanghias-nvd: Masking may bias the training. It's safer to fail.

nemo_rl/models/generation/vllm/vllm_worker.py
```diff
@@ -568,7 +581,12 @@ def _load_model(self, bundle_indices, seed):
             enable_sleep_mode=True,
             # Set disable_log_stats=False so that self.llm.get_metrics() works.
             disable_log_stats=False,
-            logprobs_mode="processed_logprobs",
+            # Keep the main default, while allowing an RL recipe to request
+            # raw model logprobs. This is required when policy logprobs are
+            # compared with vLLM generation under a logits processor.
+            logprobs_mode=self.cfg["vllm_cfg"].get(
```
yfw: [`vllm_worker.py:587`](https://github.com/NVIDIA-NeMo/RL/blob/a98199335b89e24bd140f93d8304345c7cf10c5f/nemo_rl/models/generation/vllm/vllm_worker.py#L587)

Per our [config conventions](https://github.com/NVIDIA-NeMo/RL/blob/main/.agents/contributor-skills/config-conventions/SKILL.md), a default must live in one place and never as a `cfg.get("k", default)` at the call site. Two of the new `vllm_cfg` keys bury a non-`None` default in the consumer:
- `logprobs_mode` here → `.get("logprobs_mode", "processed_logprobs")`
- [`cap_max_tokens_to_context`](https://github.com/NVIDIA-NeMo/RL/blob/a98199335b89e24bd140f93d8304345c7cf10c5f/nemo_rl/models/generation/vllm/vllm_worker.py#L841) → `.get("cap_max_tokens_to_context", False)` — the exact shape the config-conventions skill lists as forbidden (`cfg.get("disable_ppo_ratio", False)`).

For these v1 `TypedDict` (`NotRequired`) fields the default belongs in the exemplar YAML (see the companion comment on `vllm/config.py`); the call site should then read the value without a fallback — or, since they're `NotRequired`, use `.get("logprobs_mode")` and handle `None` rather than substituting a value. (The pre-existing [`use_tqdm`](https://github.com/NVIDIA-NeMo/RL/blob/a98199335b89e24bd140f93d8304345c7cf10c5f/nemo_rl/models/generation/vllm/vllm_worker.py#L741) at L741/L935 has the same smell — existing debt, not a license for new instances.)

nemo_rl/models/generation/vllm/config.py
```diff
@@ -34,6 +34,15 @@ class VllmSpecificArgs(TypedDict):
     async_engine: bool
     load_format: NotRequired[str]
     precision: NotRequired[str]
+    # Whether vLLM returns logprobs before or after generation-time logit
+    # processors. RL policy recomputation uses raw model logits, so recipes
+    # with generation-time processors should request ``raw_logprobs`` when
+    # comparing generation and policy logprobs.
+    logprobs_mode: NotRequired[Literal["processed_logprobs", "raw_logprobs"]]
```
yfw, line 41: [`vllm/config.py:41`](https://github.com/NVIDIA-NeMo/RL/blob/a98199335b89e24bd140f93d8304345c7cf10c5f/nemo_rl/models/generation/vllm/config.py#L41)

These new keys — `logprobs_mode`, `cap_max_tokens_to_context` (here), and `bad_words` (on [`GenerationConfig`](https://github.com/NVIDIA-NeMo/RL/blob/a98199335b89e24bd140f93d8304345c7cf10c5f/nemo_rl/models/generation/interfaces.py#L200)) — are set only in the recipe YAMLs; none appears in the exemplar `examples/configs/vlm_grpo_3B_megatron.yaml`. For v1 `TypedDict` configs the exemplar is the source of truth for defaults, and the config skill's "Documenting New Config Keys" asks that new keys be reflected there with their documented default. Suggest adding to the exemplar's `generation` / `generation.vllm_cfg` block:

```yaml
generation:
  bad_words: null   # optional list[str] to suppress in generation
  vllm_cfg:
    logprobs_mode: processed_logprobs   # raw_logprobs when comparing gen vs policy logprobs
    cap_max_tokens_to_context: false
```

and dropping the call-site fallbacks (companion comment on `vllm_worker.py`). If the exemplar is covered by `tests/unit/test_config_v2.py` / `test_config_validation.py`, mirror the change in the matching `tests/unit/reference_configs/*.yaml` so the migration check stays green.

nemo_rl/models/generation/vllm/vllm_worker.py
aroshanghias-nvd: Fixed. I removed the non-None call-site defaults. logprobs_mode is now passed to vLLM only when explicitly configured, and cap_max_tokens_to_context uses the optional config value directly. I also simplified the optional bad_words forwarding.

nemo_rl/models/generation/vllm/config.py
aroshanghias-nvd: These settings are specific to the Nemotron Omni workflow, so I kept them in vlm_grpo-nemotron-omni-30ba3b-clevr-1n8g-megatron-tp8ep8.v1.yaml rather than adding them to the general VLM exemplar. All currently checked-in Nemotron Omni Megatron recipes either use this recipe directly or inherit from it. The fields remain NotRequired, and the worker now preserves vLLM’s defaults when they are absent.

nemo_rl/data/processors.py
rohitrango: Please also make sure that Automodel backend supports mixed-resolution inputs to match the embedding and placeholder counts. Currently, it looks like the mcore model shipped in bridge handles it but automodel backend does not.

rohitrango (changes requested):
everything else lgtm considering @yfw's review and my previous comments

nemo_rl/models/generation/vllm/vllm_worker.py
```diff
@@ -576,10 +589,13 @@ def _load_model(self, bundle_indices, seed):
             enable_sleep_mode=True,
             # Set disable_log_stats=False so that self.llm.get_metrics() works.
             disable_log_stats=False,
-            logprobs_mode="processed_logprobs",
             **vllm_kwargs,
         )
 
+        logprobs_mode = self.cfg["vllm_cfg"].get("logprobs_mode")
```
yfw, line 595: [`vllm_worker.py:595`](https://github.com/NVIDIA-NeMo/RL/blob/a7e52a398525ccf8da05ff84be2ac6bfb1831b0c/nemo_rl/models/generation/vllm/vllm_worker.py#L595)

**[medium] This silently changes the `logprobs_mode` default for existing recipes.** `main` hardcoded `logprobs_mode="processed_logprobs"` in the `LLM()` kwargs unconditionally; this PR removes it and only sets the key when a config provides it. vLLM 0.25.1's own default is `raw_logprobs` ([`vllm/config/model.py`]( https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/config/model.py )), so every recipe that doesn't set the key (only the 3 new Omni recipes do) flips `processed → raw`, in both the sync and async workers.

It's a mathematical no-op at identity sampling (temp=1 / top_p=1 / top_k=-1 / no penalties) — the shipped majority — so the real impact is narrow: recipes with non-identity sampling **and** importance-sampling or on-policy-KL correction use the wrong behavior-policy logprob in the loss (the loss treats `generation_logprobs` as the *filtered* policy). Concrete existing example: `grpo-nemotron3-super-120BA12B-...-qa` (top_p 0.95 + importance sampling). All non-identity recipes also get shifted train/inference-mismatch health metrics (`token_mult_prob_error` / gen-KL / JSD).

Fix is trivial and zero-downside — restore the default: `self.cfg["vllm_cfg"].get("logprobs_mode", "processed_logprobs")`. The Omni recipes pin `raw_logprobs` explicitly, so they're unaffected.

nemo_rl/data/processors.py
```diff
@@ -614,15 +614,39 @@ def vlm_hf_data_processor(
...
+    if "num_frames" in message and "num_frames" not in multimodal_keys:
+        multimodal_keys.append("num_frames")
     for key in multimodal_keys:
         if key in message:
             user_message[key] = PackedTensor(
-                message[key], dim_to_pack=get_dim_to_pack_along(processor, key)
+                message[key],
+                dim_to_pack=get_dim_to_pack_along(processor, key),
+                pad_to_max_shape=_uses_image_placeholder and key == "pixel_values",
```
yfw, line 649: [`processors.py:649`](https://github.com/NVIDIA-NeMo/RL/blob/a7e52a398525ccf8da05ff84be2ac6bfb1831b0c/nemo_rl/data/processors.py#L649)

**[low-medium] Mixed-resolution inputs are unguarded on the AutoModel path** (matches @rohitrango's open `CHANGES_REQUESTED`). `pad_to_max_shape=True` is set for any image-placeholder processor, which includes the processor the **AutoModel** Omni recipes use — so `pixel_values` gets padded to the batch-max on that path too. The mcore `NemotronOmniModel` crops the padded regions via `imgs_sizes` before RADIO patchification, but on the AutoModel/DTensor path [`filter_multimodal_kwargs_for_model`](https://github.com/NVIDIA-NeMo/RL/blob/a7e52a398525ccf8da05ff84be2ac6bfb1831b0c/nemo_rl/models/automodel/data.py#L54) drops `imgs_sizes` (the HF forward doesn't declare it) and there's no crop equivalent or setup guard.

It's latent — the shipped CLEVR/MMPR AutoModel recipes use fixed-size equal-H/W tiles, so `F.pad` is a strict no-op and they pass today. But for genuinely mixed-resolution batches this PR converts what was a loud `torch.cat` error on `main` into a silent pad-and-process (the terminal mode — silently-wrong pixels vs. a placeholder-count crash — depends on HF automodel behavior we didn't independently confirm). Suggest either a crop/unpad on the AutoModel boundary when `imgs_sizes` is present, or a setup-time assert that rejects mixed-resolution on AutoModel so it fails loud.

nemo_rl/models/megatron/setup.py
```diff
@@ -68,6 +68,8 @@
 
 _HF_CONFIG_PATCHED = False
 
+_NEMOTRON_OMNI_EXPANDED_SEQUENCE_CONTRACT = "expanded_sequence_v1"
```
yfw, line 71: [`setup.py:71`](https://github.com/NVIDIA-NeMo/RL/blob/a7e52a398525ccf8da05ff84be2ac6bfb1831b0c/nemo_rl/models/megatron/setup.py#L71)

**[low] Hardcoded contract string duplicates an exported MBridge constant.** This copies `"expanded_sequence_v1"` as a private constant instead of importing [`NEMOTRON_OMNI_EXPANDED_SEQUENCE_CONTRACT`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/573e088c9c6740082c39744e03dc5b009e730ed4/src/megatron/bridge/models/nemotron_omni/nemotron_omni_provider.py#L34). The equality check downstream gates flash-attention rejection + the auto attention-backend dispatch, and MBridge already defines a sibling `NEMOTRON_OMNI_LLAVA_CONTRACT`, so the value space is explicitly versioned. If MBridge bumps the string, this copy silently stops matching and the guard no-ops (flash-attn on multi-row THD, which the code itself flags as unsafe) — a silent-on-misconfiguration path. MBridge is a hard dep of the megatron path, so importing the constant (as the test file already does) makes a bump fail loud instead.

yfw: Keeping this for now

nemo_rl/models/generation/vllm/vllm_worker.py
yfw: Reverted back to having processed_logprobs in the exemplar configs per config-conventions

nemo_rl/data/processors.py
yfw: Added assert that rejects mixed-resolution

nemo_rl/data/multimodal_utils.py
yfw: added an assert for this on Automodel side

yfw (dismissed):
Approved @aroshanghias-nvd 's changes up until my last commit

nemo_rl/models/automodel/data.py
```diff
@@ -42,26 +42,44 @@ def _accepted_forward_kwargs(
...
+
+
 def filter_multimodal_kwargs_for_model(
     model: nn.Module, multimodal_kwargs: dict[str, Any]
 ) -> dict[str, Any]:
     """Drop processor metadata that is not accepted by an AutoModel forward."""
     accepted_kwargs = _accepted_forward_kwargs(type(model))
     if accepted_kwargs is None:
         return multimodal_kwargs
+    # A forward that cannot consume imgs_sizes also cannot crop the per-image
```
rohitrango, line 66: looks good. can you add the specific model class (i.e. `NemotronHForCausalLM`) or name (`nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-BF16`) since it might be unclear in the future which model this is for?

rohitrango (dismissed):

nemo_rl/models/automodel/data.py
yfw: Added in b8d75236fe0cf54b918ee16ee9e7272fa1d67684

terrykong (approved):

terrykong (approved):

PR #3292, opened 2026-07-20
fix: eliminate JSON serialization bottleneck in r3 + NeMo-Gym rollouts

zyzhou5:
# What does this PR do ?

With router replay enabled on the NeMo-Gym path, per-sample routed experts (`[tokens, layers, topk]`)
are attached to chat responses via `.tolist()` — at ~30k-token contexts that is **~40 MB of nested JSON
ints per sample**, re-parsed, pydantic-validated, and re-serialized at every gym HTTP hop (~1 s of
single-threaded CPU each). This throttles async-GRPO rollout production to a fraction of R3-off
throughput while generation GPUs sit idle.

This PR encodes the routes as a single self-describing base64 string
(`nrlre1:<dtype>:<SxLxK>:<base64>`, preserving the resolved `routed_experts_dtype`), so intermediate
hops handle one opaque object. The decoder still accepts the legacy nested-list format.

## Validation
The performance improves significantly with this optimization
Report (Test 4 - Exp3): https://api.wandb.ai/links/nvidia-nemo-fw-public/m4onwlx8
Screenshot 2026-07-20 at 3 50 33 PM

### DIFF
--- a/nemo_rl/models/generation/vllm/utils.py
+++ b/nemo_rl/models/generation/vllm/utils.py
@@ -22,6 +22,7 @@ from nemo_rl.models.generation.interfaces import (
     ROUTED_EXPERTS_FALLBACK_DTYPE,
     GenerationDatumSpec,
 )
+from nemo_rl.utils.routed_experts_codec import encode_routed_experts
 
 R3_MISSING_ROUTE_SENTINEL = -1
 
@@ -292,9 +293,13 @@ def attach_routed_experts_to_chat_response_choices(
                 r3_stats["actual_routes"],
                 r3_stats["expected_routes"],
             )
-        choice.message.routed_experts = routed_experts.to(
-            dtype=routed_experts_dtype
-        ).tolist()
+        # Base64 envelope instead of .tolist(): nested JSON int lists cost
+        # ~1s of CPU per serialize/parse hop at long context lengths and get
+        # re-validated at every gym HTTP hop; a single string passes through
+        # the gym chain opaquely.
+        choice.message.routed_experts = encode_routed_experts(
+            routed_experts.to(dtype=routed_experts_dtype)
+        )
 
     if len(attached_choice_indices) != len(choices):
         missing_choice_indices = sorted(
--- a/nemo_rl/utils/routed_experts_codec.py
+++ b/nemo_rl/utils/routed_experts_codec.py
@@ -1,0 +1,99 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Compact wire codec for router-replay routed-expert indices.
+
+Routed experts for a long-context sample are millions of ints (shape
+[tokens, num_moe_layers, topk]). Serialized as nested JSON lists they cost
+~1s of single-threaded CPU per serialize/parse hop, and every HTTP hop on
+the NeMo Gym path (model server, agent, resources server) pays that again
+for pydantic validation and re-serialization. Encoded as a single base64
+string the payload stays one opaque Python object end to end, so
+intermediate hops only pay a string copy.
+
+Envelope format (version 1):
+    "nrlre1:<dtype>:<S>x<L>x<K>:<base64 of C-contiguous array bytes>"
+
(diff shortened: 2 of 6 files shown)


zyzhou5:
/ok to test [c7f9bcf](https://github.com/NVIDIA-NeMo/RL/pull/3292/commits/c7f9bcf39849a91b6f5209e68467860e4571ef07)

zyzhou5:
/ok to test [d6be365](https://github.com/NVIDIA-NeMo/RL/pull/3292/commits/d6be3659377e9186a92fcde5ba6fc35822c141a8)

nemo_rl/utils/routed_experts_codec.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+    dtype_name = _TORCH_DTYPE_NAMES.get(routed_experts.dtype)
+    if dtype_name is None:
+        raise ValueError(
+            f"Unsupported routed_experts dtype {routed_experts.dtype}; "
+            f"expected one of {sorted(_NP_DTYPES)}."
+        )
+    arr = routed_experts.detach().cpu().numpy()
+    tokens, num_layers, topk = arr.shape
+    data = base64.b64encode(np.ascontiguousarray(arr).tobytes()).decode("ascii")
```
ZhiyuLi-Nvidia:
```suggestion
    arr = routed_experts.detach().cpu().contiguous().numpy()
    tokens, num_layers, topk = arr.shape
    encoded = base64.b64encode(memoryview(arr)).decode("ascii")
```

AI is proposing to avoid tobytes() is possible as it may create a full-size copy.

nemo_rl/utils/routed_experts_codec.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+    arr = np.frombuffer(base64.b64decode(data), dtype=_NP_DTYPES[dtype_name])
+    expected = shape[0] * shape[1] * shape[2]
+    if arr.size != expected:
+        raise ValueError(
+            f"routed_experts envelope has {arr.size} elements, expected "
+            f"{expected} for shape {shape}."
+        )
+    # frombuffer views are read-only; copy() yields a writable array.
+    return torch.from_numpy(arr.reshape(shape).copy()).to(dtype)
```
ZhiyuLi-Nvidia: `from_numpy` do we need a numpy copy or is it better with torch.frombuffer() directly with shared memory? 

AI is suggesting:
```
raw = bytearray(base64.b64decode(data, validate=True))

result = torch.frombuffer(raw, dtype=wire_dtype).reshape(shape)
```

frombuffer views are read-only. Looks like AI `frombuffer` views are not inherently read-only; writability depends on the underlying buffer.

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -292,9 +293,13 @@ def attach_routed_experts_to_chat_response_choices(
                 r3_stats["actual_routes"],
                 r3_stats["expected_routes"],
             )
-        choice.message.routed_experts = routed_experts.to(
-            dtype=routed_experts_dtype
-        ).tolist()
+        # Base64 envelope instead of .tolist(): nested JSON int lists cost
```
ZhiyuLi-Nvidia, line 296: Only r3 indexes need to bypass the json serialization. Is that right? How about other tensors?

ZhiyuLi-Nvidia:
Awesome @zyzhou5. Thank you for driving this.

Looks good to me except the above nit comments.

nemo_rl/utils/routed_experts_codec.py
```diff
@@ -0,0 +1,99 @@
...
+depend on numpy and torch.
+"""
+
+import base64
+from typing import Any, Union
+
+import numpy as np
+import torch
+
+_MAGIC = "nrlre1"
```
ZhiyuLi-Nvidia, line 36: Maybe it is over complicated. 
Is it a good idea to have a protobuf/schema for serialization/deserialization just for readability?


```
class RoutedExpertsCodec:
    @staticmethod
    def encode(routed_experts: torch.Tensor) -> str:
        # Existing encode_routed_experts implementation
        ...

    @staticmethod
    def decode(
        payload: Union[str, Any],
        dtype: torch.dtype,
    ) -> torch.Tensor:
        # Existing decode_routed_experts implementation
        ...
```

nemo_rl/models/generation/vllm/utils.py
zyzhou5: Only router indices are large enough to make a difference.

nemo_rl/utils/routed_experts_codec.py
zyzhou5: This might be a bit overly complicated? And protobuf is binary — inside a JSON response it'd still need base64, adding a layer rather than removing one.

zyzhou5:
/ok to test [79fb96b](https://github.com/NVIDIA-NeMo/RL/pull/3292/commits/79fb96bd7666db4189a5d21e12a52ae5e74c5775)

zyzhou5:
@terrykong can you review this at your convenience?

terrykong (approved):

PR #3294, opened 2026-07-20
perf: Reduce MXFP8 refit weight-transfer overheads

seonjinn:
## Summary

Reduce repeated BF16-to-MXFP8 refit work for colocated Sync RL.

- Quantize eligible trainer weights before the Sync RL transfer.
- Reuse CUDA IPC buffers, validated vLLM loader routes, and MXFP8 shuffle scratch buffers.
- Keep the existing path as the default when the optimization is disabled.

The Qwen Sync MXFP8 performance recipes enable the optimization. Async recipes explicitly leave it disabled; Async MXFP8 uses the NCCL Reshard path configured by #3865.

## Safety

The prequantized path exchanges metadata before transfer and requires each E4M3 tensor to include its matching `*_scale_from_checkpoint`. It rejects incompatible trainer FP8 storage and NCCL Reshard configurations.

## Results

Historical integrated B200 measurements:

| Model | Transfer + update | Total refit |
|---|---:|---:|
| Qwen3-30B-A3B | 9.67 to 2.98 s (-69.2%) | 17.2 to 13.1 s (-23.8%) |
| Qwen3-235B-A22B | 18.40 to 7.26 s (-60.5%) | 40.2 to 30.4 s (-24.4%) |
| Nemotron Nano 3 | 3.56 to 1.69 s (-52.6%) | 13.2 to 11.6 s (-12.0%) |

These measurements include the batched MXFP8 MoE shuffle now provided by #3478. A matched current-main A/B is still needed to isolate the remaining changes in this PR.

## Validation

- 63 targeted GB200 vLLM/MXFP8 refit tests pass
- Unit coverage for the prequantization handshake, wire format, incompatible configuration rejection, IPC buffer reuse, loader-route caching, and MXFP8 MoE layout handling


seonjinn:

Setup: Qwen3-30B-A3B GRPO (OpenMathInstruct-2), 2 nodes x 8 B200, `grpo-qwen3-30ba3b-4n4g-mxfp8-rollout.yaml`, 20 steps, means over steps 3-20. Control = current main behavior (all flags off); treatment = this branch with `refit_prequantize`, persistent IPC buffers (4 GB), slim offload, cached loaders, pinned reference swap (batched shuffle is default-on). Same cluster, back-to-back runs.

| steps 3-20 mean | flags off (main) | this PR | delta |
|---|---|---|---|
| transfer_and_update_weights (s) | 9.67 | **2.98** | **-69%** |
| prepare_for_generation total (s) | 17.2 | 13.1 | -4.1 s |
| generation / logprob / training (s) | 49.8 / 84.6 / 91.0 | 51.8 / 82.0 / 90.6 | unchanged (noise) |
| total step (s) | 256.3 | 250.9 | -2.1% |
| train reward | 0.528 | 0.529 | = |

The saving is confined to the refit phase, exactly as claimed; all compute phases and reward are untouched. `NRL_MXFP8_SHUFFLE_VERIFY=1` bit-exact assert passed on first refit.

seonjinn:
/ok to test [740c825](https://github.com/NVIDIA-NeMo/RL/pull/3294/commits/740c825ae1c1f07fb54bf60017e21f6715ea60ae)

seonjinn:
/ok to test [3f60f5d](https://github.com/NVIDIA-NeMo/RL/pull/3294/commits/3f60f5d6821655aa9e6aaf744083a1b4c1ced1fe)

seonjinn:
**Isolated reproduction at 235B scale (follow-up to the 30B numbers above).**

Setup: Qwen3-235B-A22B GRPO, 8 nodes x 8 B200 (64 GPUs), `grpo-qwen3-235b-16n4g-mxfp8-rollout.yaml`, vLLM TP 8, 20 steps, means over steps 3-20. Control = same clean branch with all refit flags off; treatment = `refit_prequantize` + persistent IPC buffers + slim offload + cached loaders + pinned reference swap (batched shuffle default-on).

| steps 3-20 mean | flags off (main) | this PR | delta |
|---|---|---|---|
| transfer_and_update_weights (s) | 21.8 | **7.4** | **-66%** |
| prepare_for_generation total (s) | 42.2 | 28.8 | -13.4 s |

The refit-transfer saving grows with model size (30B: 9.7 -> 3.0 s; 235B: 21.8 -> 7.4 s), because larger weights spend proportionally more of the refit in transport - exactly the phase this PR targets. Batched-shuffle bit-exact verify passed on first refit.

seonjinn:
/ok to test [04af034](https://github.com/NVIDIA-NeMo/RL/pull/3294/commits/04af034880e33d8f539d37c8700e76fc83f3a484)

seonjinn:
/ok to test [438e20d](https://github.com/NVIDIA-NeMo/RL/pull/3294/commits/438e20dbd0e6a4edb9f96d20f1917baa445d66dd)

seonjinn:
/ok to test [96ba299](https://github.com/NVIDIA-NeMo/RL/pull/3294/commits/96ba299969f182f3a6cbac684df61fbece3415fb)

nemo_rl/weight_sync/checkpoint_engine_weight_synchronizer.py
```diff
@@ -67,7 +67,15 @@ class CheckpointEngineWeightSynchronizer(WeightSynchronizer):
     _bucket_size_bytes: int | None = None
 
     def init_communicator(self) -> None:
-        self._generation.prepare_refit_info(self._policy.prepare_refit_info())
+        state_dict_info = self._policy.prepare_refit_info()
+        prequant_names = self._generation.prepare_refit_info(state_dict_info)
+        if prequant_names:
+            updated_info = self._policy.enable_refit_prequantize(prequant_names)
```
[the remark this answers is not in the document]
seonjinn: Mirrored the Megatron-backend guard in the checkpoint-engine handshake so DTensor policies now fail with the same actionable ValueError before calling enable_refit_prequantize. Added a regression test and made the successful checkpoint-engine fixtures explicitly Megatron-backed. Fixed in 85a6feaf5.

nemo_rl/algorithms/utils.py
```diff
@@ -1032,3 +1038,30 @@ def print_efficiency_summary(
...
+    policy_config: PolicyConfig,
+) -> None:
+    """Complete the trainer-side pre-quantized refit handshake if requested.
+
+    The generation backend's prepare_refit_info returns the fp8-eligible
+    parameter names when vllm_cfg.refit_prequantize is enabled; the trainer
+    then quantizes exactly those during refit and the receiver's unpack
+    metadata is refreshed to the quantized dtypes plus scale entries.
+    """
+    prequant_names = policy_generation.prepare_refit_info(state_dict_info)
```
terrykong: **What the handshake is** — needed to read the rest. Each step, updated weights are copied trainer→engine (the refit). Normally the trainer sends BF16 and vLLM quantizes on arrival; `refit_prequantize` moves quantization to the trainer so only E4M3 + scales cross the wire (~half the bytes). The trainer can't know *which* weights vLLM wants quantized — that depends on vLLM's own layer structure, PP sharding and `num_*_layers_in_bf16` — so the two sides negotiate once at startup, here:

1. trainer→vLLM: manifest `{name: (shape, dtype)}`; vLLM returns the names it can take pre-quantized ([`vllm_backend.py:432`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/generation/vllm/vllm_backend.py#L432-L436))
2. →trainer: [`enable_refit_prequantize(names)`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1809) records `_refit_prequant_names` and returns a revised manifest (those entries now `float8_e4m3fn`, each gaining a `*_scale_from_checkpoint` entry)
3. →vLLM: the revised manifest, so the receiver unpacks the new wire format correctly

At refit time [`_maybe_prequantize_param`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1827-L1835) passes through anything not in `_refit_prequant_names` — so skipping step 2 silently disables the feature. No error, no log line, no speedup.

**Coverage is uneven** — five call sites, four different shapes:

| path | handshake |
|---|---|
| [`ppo.py:665`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/algorithms/ppo.py#L665) | ✅ unconditional |
| [`grpo.py:1388`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/algorithms/grpo.py#L1388) — `else:` of the transport branch; covers `grpo_train`, `async_grpo_train` and `grpo_sync.grpo_train_sync` (all share this `setup()`) | ✅ |
| [`distillation.py:623`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/algorithms/distillation.py#L623) | ✅ |
| [`CheckpointEngineWeightSynchronizer`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/weight_sync/checkpoint_engine_weight_synchronizer.py#L70-L78) | ✅ hand-inlined copy, minus the Megatron guard this helper has |
| [`grpo.py:1362`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/algorithms/grpo.py#L1362) remote-transport branch (nixl / zmq-sparse / s3-sparse) | ❌ |
| [`single_controller_utils/setup.py:419`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/algorithms/single_controller_utils/setup.py#L419) → Ipc/Collective/Http synchronizers | ❌ |

Both ❌ rows call `prepare_refit_info` and discard the return, so steps 2-3 never run. `examples/run_grpo_single_controller.py` with a Megatron policy + colocated MXFP8 vLLM is a live config that hits the second one.

**Root cause:** this helper takes a `PolicyConfig`, which the synchronizers don't have — that's why they couldn't reuse it and the checkpoint-engine copy was hand-inlined instead.

**Action:** drop the `policy_config` parameter (derive the Megatron check from the policy, or pass a bool), move the helper next to the synchronizers, and call it from all four `init_communicator()` implementations — five sites collapse to one and both ❌ rows are fixed. If that's more than this PR should carry, the cheap version is a `raise` in `create_weight_synchronizer` when `refit_prequantize` is set and the resolved synchronizer can't honour it, matching the guards already at [`factory.py:71-83`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/weight_sync/factory.py#L71-L83).

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -215,6 +224,7 @@ def init_fp8(vllm_cfg, model_name, model_parallel_size):
     }
     if is_mx:
         fp8_config_kwargs["is_mx"] = True
+        fp8_config_kwargs["refit_prequantize"] = bool(vllm_cfg.get("refit_prequantize"))
```
terrykong, line 303: **Two config combinations accept `refit_prequantize: true` and silently drop it.** `fp8_config_kwargs["refit_prequantize"]` is set only inside `if is_mx:`; the [`else:` arm](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/generation/vllm/quantization/fp8.py#L234-L241) leaves it at the `FP8Config` default `False`.

1. **`precision: fp8, is_mx: false`** → dropped here.
2. **NVFP4** (`quant_cfg` + `real_quant`) → dropped earlier still: `init_fp8` never runs, because [`vllm_worker.py:453`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/generation/vllm/vllm_worker.py#L453) gates it on `precision == "fp8"`. NVFP4 has its own handshake — [`VllmQuantInternalWorkerExtension.prepare_refit_info`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/modelopt/models/generation/vllm_quant_backend.py#L513-L540) returns `None` unconditionally for real-quant models ("Real quantization owns a separate refit handshake").

Neither is harmful — NVFP4 real-quant already streams packed quantized weights, so it has the payload win by another route, and `refit_prequantize` is genuinely inapplicable there. The issue is only that both accept the flag and do nothing, with a wall-clock diff as the only way to notice. (MXFP4 isn't in the repo — `grep -rn "mxfp4" nemo_rl/` is empty — so there's nothing to guard for it.)

**Action:** put the guard where `vllm_cfg` keys are validated, **not** in `init_fp8` — a raise inside `init_fp8` can't see case 2, since that path never enters it. Rejecting `refit_prequantize: true` unless `precision == "fp8" and is_mx` covers both cases and any future format in one place. `refit_prequantize` has no meaning outside MXFP8, so this rejects nothing legitimate.

nemo_rl/models/generation/vllm/config.py
```diff
@@ -36,6 +37,11 @@ class VllmSpecificArgs(TypedDict):
     precision: NotRequired[str]
     # Use ModelOpt MXFP8 quantization when precision is fp8.
     is_mx: NotRequired[bool]
+    # With is_mx, quantize weights to MXFP8 on the trainer during refit and
+    # stream E4M3 data plus scales (~47% smaller payload) instead of BF16;
+    # the vLLM worker then skips its per-refit re-quantization. Requires the
+    # Megatron policy backend.
+    refit_prequantize: NotRequired[bool]
```
terrykong, line 73: **`refit_prequantize` is set in no YAML in the repo**, so nothing in-tree exercises the PR's headline optimization:

```
$ grep -rn "refit_prequantize" examples/ docs/
(no output)
```

The description's "Known deviations" says it *"is documented in the mxfp8 recipe YAMLs rather than the exemplar, mirroring the sibling `is_mx`/fp8 keys."* The `is_mx` half holds — it's in all six mxfp8 recipes (`grep -rn "is_mx" examples/configs/recipes/` → 6 hits under `performance/`) — but `refit_prequantize` is in none of them, nor the exemplar. The PR's other three new flags *were* added to both [`examples/configs/grpo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/examples/configs/grpo_math_1B.yaml) and [`tests/unit/reference_configs/grpo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/tests/unit/reference_configs/grpo_math_1B.yaml), so this looks like the fourth one was just missed.

**Action:** set it in the mxfp8 rollout recipes the benchmarks came from — that also makes the description's sentence true — or add it to the exemplar + reference pair like the other three.

Two smaller things while you're here:

- No `docs/` mention of any of the four new flags (`grep -rE "refit_prequantize|refit_slim_offload_after|pinned_reference_swap|refit_persistent_ipc_buffers" docs/` → empty). [`docs/fp8.md` §Usage](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/docs/fp8.md#L35-L57) is already a `vllm_cfg` block listing `precision`, `use_deep_gemm`, `pow2_*` — and `refit_prequantize` is the one flag with a prerequisite (`is_mx` + Megatron) a user can't discover anywhere today.
- The description says prepare_for_generation is "15.2 → 9.8 (-36%), now faster than BF16's 12.4 s", but the isolated 30B run reports 17.2 → 13.1 — and 13.1 > 12.4, so that headline isn't reproduced. Worth noting which stack each table came from. (The 235B row goes the other way, isolated -66% vs the body's -60%, so it's just this one claim.)

tests/unit/models/generation/test_mxfp8_prequant.py
```diff
@@ -0,0 +1,151 @@
...
+    assert (abs_err[~representable] <= block_amax[~representable] / 32).all()
+
+
+def test_last_dim_not_divisible_raises():
+    x = torch.randn(8, MXFP8_BLOCK_SIZE + 1, dtype=torch.bfloat16)
+    with pytest.raises(AssertionError):
+        _mxfp8_e4m3_quantize_torch(x)
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="requires CUDA")
```
terrykong: **Neither leg of the MXFP8 bit-exactness claim runs in CI.** No evidence of an actual defect — the equivalence checks out three ways (see review summary) — but nothing would catch a regression from a future FlashInfer or vLLM bump on a default-ON path.

**(a) This test is GPU-gated but not arch-gated.** Both sides dispatch on compute capability — [`fp8_train_utils.py:77`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py#L77) and vLLM's [`mxfp8_utils.py:92`](https://github.com/vllm-project/vllm/blob/v0.20.0/vllm/model_executor/layers/quantization/utils/mxfp8_utils.py#L92). CI unit tests run on H100: every `cicd-unit-tests` entry uses `runner_prefix` ← [`vars.DEFAULT_H100_RUNNER`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/.github/workflows/cicd-main.yml#L182), and `GB200_RUNNER` appears only on container-build jobs. So on sm90 both sides fall through to the same torch reference and the test compares it against itself.

**(b) On sm100 it would fail.** The reference here is bare `mxfp8_e4m3_quantize` + squeeze, but the real receiver path also clamps zero scales ([`fp8.py:468-471`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/generation/vllm/quantization/fp8.py#L468-L471)), and FlashInfer [can emit an E8M0 byte of 0](https://github.com/flashinfer-ai/flashinfer/blob/v0.6.8.post1/csrc/nv_internal/tensorrt_llm/kernels/quantization_utils.cuh#L499-L502) — upstream documents the underflow. A near-zero block gives `got_scale=1` vs `ref_scale=0` → `torch.equal` fails.

```suggestion
@pytest.mark.skipif(
    not torch.cuda.is_available()
    or torch.cuda.get_device_capability() < (10, 0),
    reason="requires sm100+; below it both sides fall back to the shared "
    "torch reference and the comparison is vacuous",
)
```
(and clamp the reference to match the receiver, or the test breaks the first time it runs on Blackwell)

**(c) `NRL_MXFP8_SHUFFLE_VERIFY` is set nowhere** — the only hits across `*.py/sh/yaml/yml/md/txt` are inside `fp8.py` itself. [`mxfp8_shuffle_verified`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/generation/vllm/quantization/fp8.py#L931) is also a process-global one-shot, and `is_gated` is fixed per model, so one run can only ever cover one arm — the description's "covers both … paths" is true across your two manual runs but isn't a standing guard. Consider a per-layer set plus setting the env var in a GB200 functional recipe.

**(d) Minor, same area:** the comment at [`fp8_train_utils.py:71-79`](https://github.com/NVIDIA-NeMo/RL/blob/2c147ac40fc170c60e3cd9d9429930ff86a129b1/nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py#L71-L79) says the two arms "may differ in boundary rounding". They differ by a systematic 2^8 — torch uses `2^floor(log2 amax)` ([`mxfp8_utils.py:61`](https://github.com/vllm-project/vllm/blob/v0.20.0/vllm/model_executor/layers/quantization/utils/mxfp8_utils.py#L61)) vs FlashInfer's `e8m0(amax/448)`, which costs ~8 binades of in-block dynamic range on the fallback (measured: 12-15x worse mean relative error on outlier-bearing blocks). Not reachable today — `mcore` and `vllm` pin the same flashinfer — and the torch function is copied verbatim from vLLM, so only the wording is yours. But the same comment says "validate with the parity test before relying on it", and per (a) that test can't discriminate the two arms on any single machine. Suggest making the `except ImportError: pass` loud and dropping that sentence.

terrykong (commented):
Reviewed with a team of 7 agents; 20 candidate findings, 4 survived adversarial verification.

**Please rebase** — currently `CONFLICTING` against `main`. One file conflicts, `nemo_rl/algorithms/grpo.py` (branch is 3 commits behind).

Thanks for the evidence on this one — isolated-branch reproductions at 30B and 235B with a flags-off control, posted unprompted, plus bitwise parity tests. That made the perf story easy to trust and is why the review below is about plumbing and coverage rather than the optimization itself.

**Verified correct, no action needed:**

- **The batched MoE shuffle is bit-identical to the per-expert loop** — established three ways: from FlashInfer source ([`get_shuffle_matrix_sf_a_row_indices` just returns `get_shuffle_matrix_a_row_indices`](https://github.com/flashinfer-ai/flashinfer/blob/v0.6.8.post1/flashinfer/utils.py#L861-L878), so reusing the weight permutation for the scale tensor is sound), by reimplementing the primitives on CPU across 10 shape/gated combos incl. Nemotron's 1856/2816/3712, and by your own `test_batched_moe_shuffle_matches_per_expert`. Keeping `_shuffle_mxfp8_moe_per_expert` as a live reference rather than deleting it is the right call — it's kill switch, verify oracle and test target at once.
- **Padding is safe.** Padded rows are E4M3 `0x00` with scale byte 1 → dequantize to exactly 0, and padded `w2` rows land only in hidden dims stripped by `output[..., :unpadded_hidden_size]`. Mutating `moe_config.intermediate_size_per_partition` can't corrupt later refits — vLLM's `_load_w13`/`_load_w2` take shard sizes from [`expert_data.shape[shard_dim]`](https://github.com/vllm-project/vllm/blob/v0.20.0/vllm/model_executor/layers/fused_moe/layer.py#L952-L955), never that field, and a sweep of `fused_moe/runner/`, `router/`, `prepare_finalize/` and `v1/` found no other post-load consumer.
- Also checked and clear: `refit_slim_offload_after` (grad buffers are already moved by the unconditional `move_model` above the branch), scratch-buffer aliasing (the `.clone()` in `_set_mxfp8_apply_tensor` is load-bearing), EPLB on the cached-loader path (vLLM re-resolves the global expert id per call), and `alignment=32` (FlashInfer's own default, and a no-op given the `%32` assert upstream).

**One thing the PR undersells:** removing `_maybe_process_fp8_kv_cache` is a **net bug fix**, not just a cleanup. Pre-PR that call sat after a bare `yield` with no `try/finally`, so it only ran on normal exit — and the checkpoint-engine transport ran *only* the KV helper, never a full `process_weights_after_loading`, so MXFP8 weights were never re-shuffled after a checkpoint-engine refit. Worth a line in the description.

**Do not remove the zero-scale clamp** (`fp8_train_utils.py:88-90`, `fp8.py:468-471`) — it looks like dead code from a CPU-only read, but FlashInfer [documents the E8M0 underflow-to-zero](https://github.com/flashinfer-ai/flashinfer/blob/v0.6.8.post1/csrc/nv_internal/tensorrt_llm/kernels/quantization_utils.cuh#L499-L502) and has no `clamp(min=tiny)`, so it is load-bearing on sm100.

Testing was static + CPU-only; no GPU or Blackwell hardware, so nothing needing a runtime MXFP8 kernel was executed.

_Generated by Claude Code_

nemo_rl/models/generation/vllm/quantization/fp8.py
seonjinn: Addressed in 00fa12851. `validate_vllm_quantization_config` now rejects `refit_prequantize=true` unless `precision=fp8` and `is_mx=true`, including NVFP4 and blockwise-FP8 misuse. Validation runs at the public generation-config boundary and defensively in direct `VllmGeneration` construction. Added tests for invalid combinations, non-boolean values, the valid MXFP8 case, and public-boundary integration.

tests/unit/models/generation/test_mxfp8_prequant.py
seonjinn: Addressed in 00fa12851. The quantizer parity test is now sm100+-gated and matches the receiver zero-scale clamp; missing FlashInfer on sm100+ fails loudly. Shuffle verification is tracked per layer with a `WeakSet`, with a unit test proving each layer is checked once. The GB200 suite runs the synthetic gated/non-gated MoE parity tests, and its GRPO smoke now uses a Megatron trainer with `refit_prequantize=true`, so the two-stage trainer-side path is exercised end to end.

nemo_rl/weight_sync/checkpoint_engine_weight_synchronizer.py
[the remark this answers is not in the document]
seonjinn: Addressed in 00fa12851 by replacing the checkpoint-engine copy with the shared `initialize_refit_metadata` handshake. The shared path reads the typed policy config, preserves the actionable Megatron-backend `ValueError`, rejects missing updated metadata, and is covered by synchronizer regression tests.

nemo_rl/algorithms/utils.py
seonjinn: Addressed in 00fa12851. The handshake now lives in `weight_sync/interfaces.py`, derives the backend guard from `policy.cfg`, and is used by IPC, HTTP, collective, and checkpoint-engine synchronizers plus the GRPO/PPO/distillation setup paths. This closes the SingleController and remote synchronizer gaps. Sparse delta and NCCL-Reshard continue to reject prequantized MXFP8 explicitly. Added shared-helper, IPC, collective, and checkpoint-engine coverage.

nemo_rl/models/generation/vllm/config.py
seonjinn: Addressed in 00fa12851. Added `refit_prequantize: false` to the exemplar/reference configs and enabled it in all six in-tree MXFP8 rollout performance recipes, with a recipe test requiring Megatron + MXFP8 + prequantization. Documented all four refit performance controls in `docs/fp8.md` and `docs/guides/refit.md`. The PR body now uses the isolated matched A/B results and removes the conflicting 15.2 -> 9.8 claim.

youngeunkwon0405:
@terrykong, discussed offline with @seonjinn about this PR. 
@seonjinn said he will investigate if we can split this PR per independent optimizations.

seonjinn:
/ok to test [c4a7b06](https://github.com/NVIDIA-NeMo/RL/pull/3294/commits/c4a7b06b1bb583b9571bb5f7937a1df46be1b7f8)

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -554,13 +555,33 @@ def check_nccl_reshard_refit_support(master_config: dict) -> None:
             f"policy.generation.backend must be 'vllm' (got {backend!r})."
         )
 
+    if policy_precision != "bfloat16":
```
youngeunkwon0405: This is a wrong gating. The current impl supports fp8-->fp8. This will block that path.

seonjinn: Fixed in cd7732c49. Removed the redundant policy.precision=bfloat16 gate and updated the matching blockwise FP8 test to use policy.precision=fp8. The storage-level checks remain authoritative: FP8 requires enabled fp8_param with the blockwise recipe and vLLM precision=fp8; BF16-to-FP8 and MXFP8 scale-mapping paths remain rejected. GCP-NRT unit shard: 51 passed.

seonjinn:
PR decomposition update: I restored the original integrated #3294 branch. The batched MXFP8 MoE shuffle has been extracted unchanged into #3478 for independent review. BF16 trainer storage to MXFP8 vLLM NCCL-Reshard support is tracked separately in #3477. Before merging #3294, I will drop or rebase the extracted shuffle portion onto #3478 so duplicate code does not land.

terrykong:
@youngeunkwon0405 could you help review first, i'll review afterwards

youngeunkwon0405:
> PR decomposition update: I restored the original integrated #3294 branch. The batched MXFP8 MoE shuffle has been extracted unchanged into #3478 for independent review. BF16 trainer storage to MXFP8 vLLM NCCL-Reshard support is tracked separately in #3477. Before merging #3294, I will drop or rebase the extracted shuffle portion onto #3478 so duplicate code does not land.

Looks like you have the PR decomposition plan. Then, this PR will be just for a reference? And the decomposed PR will be the actual target for merge?

seonjinn:
@youngeunkwon0405 Oh, I think this PR also need to be merged since co-located case seems to follow ipc path. I'll  drop batched moe-shuffle updates for this PR.

youngeunkwon0405:
I thought the plan was to split this PR into independent optimizations and merge them one by one. 
Then what are these PR? 

"I restored the original integrated https://github.com/NVIDIA-NeMo/RL/pull/3294 branch. The batched MXFP8 MoE shuffle has been extracted unchanged into https://github.com/NVIDIA-NeMo/RL/pull/3478 for independent review. BF16 trainer storage to MXFP8 vLLM NCCL-Reshard support is tracked separately in https://github.com/NVIDIA-NeMo/RL/pull/3477. Before merging https://github.com/NVIDIA-NeMo/RL/pull/3294, I will drop or rebase the extracted shuffle portion onto https://github.com/NVIDIA-NeMo/RL/pull/3478 so duplicate code does not land."

What is the merge order in your plan?

seonjinn:
@youngeunkwon0405  #3478 and #3477 and probably 3294 could be merged regardless of #3478 and #3477  by only remaining pre-quantization + reuse persistent buffer + skipping unnecessary cleaning up part since these optimizations are targeted for cuda ipc path (called in sync co-located case)

youngeunkwon0405:
#3477 is a totally independent PR. #3478 is a partial component of this #3294 PR, which is a kind of decomposition effort that we have discussed. 

So the expected review order is:
* #3477 (Doesn't have any dependency on the other PRs) 
* Review #3478 and merge --> rebase #3294 --> review and merge #3294

Is this a correct understanding?

Are both #3477 and #3478 in a clean state and ready for the merge? Which one do you prefer to get reviewed and merged first?

seonjinn:
@youngeunkwon0405 Yes, correct understanding.  I think #3478 would be easy for you to review first and then #3477. Meanwhile, I'll clean up overlapped part of #3294 from them. Thank you so much!

youngeunkwon0405:
Got it. Then let's focus on #3478 first. I will be reviewing that soon. Thanks.

seonjinn:
## Self-review completed

Reviewed head `7705d22a3` with the repository's `review-pr-team` workflow. The review covered RL/refit correctness, tests, design boundaries, existing review threads, and a final devil's-advocate pass.

### Findings fixed

- Rejected `refit_prequantize` with Megatron blockwise FP8 parameter storage. E4M3 plus TE `scale_inv` is not the same wire format as MXFP8 E4M3 plus E8M0 `scale_from_checkpoint`.
- Required each prequantized MXFP8 E4M3 weight to include its matching `*_scale_from_checkpoint` tensor.
- Rejected `refit_prequantize=true` with `nccl_reshard` instead of silently ignoring the option. NCCL-Reshard format conversion remains outside this PR.
- Preserved the non-MXFP8 E4M3 plus `*_scale_inv` path with a regression test.
- Removed a stale shuffle-verification test and updated the MoE layout fixture to the current #3478 contract.

### Scope review

- #3478 is merged and this branch now builds on its implementation.
- #3477 remains independent; its NCCL-Reshard changes are absent from this diff.
- Existing performance numbers are now labeled as historical integrated results because their baseline predates #3478 in `main`. A current-base A/B run is still needed to isolate the remaining diff.

### Verification

- `ruff format`: passed on changed files.
- `ruff check`: passed on changed files.
- `git diff --check`: passed.
- Python compilation of changed modules and tests: passed.
- Local pytest collection was not available because the repository lockfile supports Linux while the review host is macOS, and the host Python does not include Ray. Linux/vLLM/GB200 CI is therefore still required on this head.

One adjacent issue was confirmed but not treated as a blocker for this PR: `use_reference_model()` is not exception-safe, but that behavior already exists on current `main` and was not introduced here.

seonjinn:
Contributor self-review found this branch should not be mechanically refreshed. Its MXFP8 padding/refit behavior overlaps #3630/#3638, rejects a valid aligned modular FlashInfer TRTLLM path, and can route native CuTe scales through the stale legacy `_scale_from_checkpoint` field when combined with the newer backend work. Recommendation: replace only any residual, non-overlapping behavior after #3630/#3638 rather than retrigger this PR as-is.

nemo_rl/modelopt/models/generation/vllm_quant_backend.py
```diff
@@ -528,10 +528,15 @@ def _synchronize_before_ipc_data_ack(self) -> None:
             return
         super()._synchronize_before_ipc_data_ack()
 
-    def prepare_refit_info(self, state_dict_info: dict[str, Any]) -> None:
-        super().prepare_refit_info(state_dict_info)
+    def prepare_refit_info(
```
seonjinn, line 534: **[Critical] Signature drift: first refit raises `TypeError` for every ModelOpt-backed vLLM engine.**

The base class gained a second positional parameter (`vllm_backend.py:460-464`, `prepare_refit_info(self, state_dict_info, serialized_fp8_config)`), and both drivers now RPC two args unconditionally:

- `vllm_worker.py:1157-1160`: `collective_rpc("prepare_refit_info", args=(state_dict_info, fp8.serialize_fp8_config()))`
- `vllm_worker_async.py:1412-1415`: same

This override still declares `prepare_refit_info(self, state_dict_info)`, so the RPC fails with `TypeError: takes 2 positional arguments but 3 were given` before the body runs - the `_is_real_quant_model()==False` fallback is unreachable, so fake-quant configs break too. No test exercises the 2-arg call against this extension.

Fix: accept and forward the new parameter (`def prepare_refit_info(self, state_dict_info, serialized_fp8_config=None)` and pass it through to `super()`).

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -2932,7 +3040,27 @@ def offload_after_refit(self):
...
+            if self.fp8_cfg and self.fp8_cfg.get("force_clear_fp8_caches", False):
+                self._clear_fp8_caches()
+            if self.cfg["megatron_cfg"].get("clear_memory_caches_before_refit", False):
+                self._clear_rope_and_moe_dispatcher_caches()
+            if (
+                hasattr(self, "optimizer")
+                and self.optimizer is not None
+                and not self.optimizer_cpu_offload
+            ):
+                self.move_optimizer("cpu")
```
seonjinn, line 3668: **[High] Slim path ignores `offload_optimizer_for_refit: False` and moves the optimizer to CPU anyway.**

`offload_before_refit` gates the optimizer move on both knobs (`:2958-2963` includes `self.offload_optimizer_for_refit`), but this slim branch only checks `not self.optimizer_cpu_offload`. A user who sets `offload_optimizer_for_refit: false` to keep the optimizer resident (e.g. CPU-RAM-constrained hosts) gets it silently offloaded whenever `refit_slim_offload_after: true` - the opposite of their config, with no warning.

The new test `test_offload_after_refit_routes_cleanup_by_mode` (test_megatron_worker.py:710-765) never sets `offload_optimizer_for_refit` and asserts `move_optimizer.assert_called_once_with("cpu")` for the slim mode, so it currently enshrines the missing gate.

Fix: add `and self.offload_optimizer_for_refit` to this condition and extend the test with the `False` case.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -443,13 +514,30 @@ def _is_fp8_weight(name, model):
...
             weights_quantized.append((k, v))
             continue
+        if v.dtype == torch.float8_e4m3fn:
+            if global_fp8_config.is_mx and not global_fp8_config.refit_prequantize:
+                raise ValueError(
+                    "MXFP8 E4M3 refit weights require refit_prequantize=true; "
+                    "other FP8 trainer scale layouts are not compatible."
+                )
+            scale_name = k + "_scale_from_checkpoint"
+            if global_fp8_config.is_mx and scale_name not in weight_names:
```
seonjinn: **[High] Per-batch scale check crashes valid refits when a weight/scale pair straddles a transfer-batch boundary.**

`weight_names` covers only the current `load_weights` batch, but every transport batches by size with no pairing constraint, and the trainer yields the E4M3 weight then its scale as two separate entries (`megatron_policy_worker.py:2039-2041`):

- packed collective: the producer breaks after the threshold-crossing tensor (`packed_tensor.py:104-108`). A prequantized weight is ~32x its scale, so the weight is usually the tensor that closes the buffer - its scale opens the next one, and this check raises `Prequantized MXFP8 weight ... is missing ...` on a perfectly healthy stream.
- IPC ping-pong: same straddle at `utils.py:527-535`.
- IPC oversized-param branch (`utils.py:493-524`) ships a larger-than-buffer weight **alone**, so for oversized params (235B-class `embed_tokens`) the crash is deterministic on every refit.

Small functional-test models fit one buffer, so CI passes while the PR's actual targets hit boundaries every refit. Note the load itself is batch-order-independent - a scale arriving in a later batch loads fine (exactly how non-MX `*_scale_inv` already behaves); only this validation is wrong.

Fix options: validate against the full `state_dict_info` manifest instead of the per-batch names, defer the completeness check to refit end (the IPC manifest already tracks it), or make the sender pack weight+scale atomically.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
```diff
@@ -13,6 +13,87 @@
...
+            ) from exc
+        else:
+            x_q, x_scales = flashinfer_mxfp8_quantize(
+                x, is_sf_swizzled_layout=False, alignment=32
+            )
+            if x_scales.ndim == 1 and x.ndim == 2:
+                x_scales = x_scales.view(x.size(0), -1)
+    if x_q is None or x_scales is None:
+        x_q, x_scales = _mxfp8_e4m3_quantize_torch(x)
+    x_scales = torch.squeeze(x_scales, dim=-1)
```
seonjinn: **[Minor, latent] Unconditional `squeeze(dim=-1)` collapses the block dimension when the reduction dim is exactly 32.**

For a weight with `K == 32`, flashinfer's scale output has a trailing dim of size 1 that is the real `K/32` block dimension, not padding; squeezing it yields a 1D scale that later fails vLLM's 2-index slicing in `process_weights_after_loading_mxfp8_linear`. No realistic checkpoint has `K == 32`, so latent - but a `if x_scales.ndim > 2` guard (or reshape to `[M, K // 32]` explicitly, mirroring `_mxfp8_e4m3_quantize_torch`) removes the trap.

Related hardening from the same review pass: this sm100+ fast path has no 3D branch, unlike the torch fallback. The adversarial pass verified 3D expert tensors cannot currently reach it (prequant eligibility requires `.endswith("weight")` on Linear/RoutedExperts HF exports, and `nccl_reshard`+prequantize is config-rejected), so it is unreachable today - a defensive `assert weight.dim() == 2` here would keep it that way when eligibility rules evolve.

seonjinn (commented):
## Self-review at head `88721ced8`

Second team self-review (the first covered `7705d22a3`; the only delta since is the 2026-08-17 `main` merge). Six domain reviewers over the unique diff vs merge-base `3d873ebc2`, then a devil's-advocate verification pass that re-read every citation: 12 findings confirmed, 2 refuted, 3 downgraded, 2 new. Blocking items are posted inline; the rest is summarized here.

### Blocking (inline)

1. **[Critical]** `VllmQuantInternalWorkerExtension.prepare_refit_info` still takes one data argument while both driver workers now RPC two - `TypeError` on the first refit for every ModelOpt-backed vLLM engine.
2. **[High]** The `refit_slim_offload_after` branch drops the `offload_optimizer_for_refit` gate and moves the optimizer to CPU against explicit user config; the new unit test encodes the missing gate as expected behavior.
3. **[High]** The per-batch MXFP8 `*_scale_from_checkpoint` presence check false-positives whenever a weight/scale pair straddles a transfer-batch boundary - deterministic crash for oversized params on the IPC path, probabilistic on packed collective. Small-CI models fit one buffer, so CI cannot catch it; 235B-class targets hit it every refit.

### Rebase-blocking collisions with `main` (not fixable on this branch as-is)

The branch merged `main@3d873ebc2` but `main` has moved 28 commits; a trial merge conflicts in 14 files. Two are semantic, not mechanical:

- **#3477 (`36b5999a7`)**: `main`'s NCCL-reshard receiver imports `quantize_mxfp8_weight` from `fp8.py`; this branch has no such function and instead inlines the equivalent quantize + zero-scale clamp. Taking "ours" on those hunks breaks `main`'s import. The two implementations must be unified (the shared helper is the natural home for this PR's trainer-side prequantizer too).
- **#3391 (`85e02cca3`)**: `main` passes a `CollectiveSenderSpec` (`buffer_size_bytes`, `num_buffers`, `nccl_peer`) into `init_collective`; this branch passes a scalar `buffer_size_bytes` at the same call sites. Two incompatible designs on one interface - needs a design decision, not a merge resolution.

This confirms the 2026-08-18 assessment that the branch must not be mechanically refreshed. Verdicts on that comment's specific claims: the #3477-family overlap is **real and now concrete** (above); the "#3630/#3638" numbers do not appear in `main` history (likely misattributed - #3477/#2744 are the actual drivers); the "CuTe scales" claim was unverifiable (no such reference in either branch), though `*_scale_from_checkpoint` is genuinely contested ground between this branch and `main`'s receiver-side quantize path.

The PR body's "#3477 remains separate; NCCL Reshard changes are not included here" is stale - #3477 merged 2026-08-19, and the scope/interaction statement needs rewriting after rebase.

### Non-blocking notes

- `checkpoint_engine.py` refit now unconditionally runs the full `_weight_update_lifecycle` finalize where base only ran the FP8-KV-cache conditional post-process. Looks like intentional harmonization; confirming intent since it adds per-refit work for non-FP8-KV deployments.
- The prequant handshake mutates trainer state (`_refit_prequant_names`) before the second `generation.prepare_refit_info` push; all 10 call sites verified fail-loud today, so this is latent - worth a comment or commit-after-push ordering.
- `interfaces.py`'s rejection message says `vllm_cfg.refit_prequantize`; the config path users set is `policy.generation.vllm_cfg.refit_prequantize`.
- Four reference configs carry `refit_cache_loader_routes` with no counterpart in their real source configs (`grpo_math_1B.yaml` lacks the key).
- Loader-route cache resets silently when vLLM post-load recreates Parameters; one log line would make a quietly-inert perf knob observable.

### Verified clean (spot-checked by the adversarial pass)

E8M0 zero-scale clamp parity trainer/receiver; `_mxfp8_e4m3_quantize_torch` math incl. 2D/3D reshape; persistent IPC buffer reuse and oversized-solo fallback; loader-route cache record/replay/fail-closed reset under `nccl_reshard`'s in-place receive (an earlier MAJOR claim here was refuted - no `.data` reassignment exists in that path); `buffer_size_bytes` threading; YAML recipe/test matrix; GB200 wire-parity test is wired into L1 CI on target hardware.

seonjinn:
Validation update on `db15416d` (based on `main` at `4601ba2c`):

**Exact PR head**

- MXFP8 quantization unit file: **56 passed, 1 deselected**.
- Separately selected Ray V1 initialization case: **1 passed**. The main unit run covers Ray V2.
- DCO, copyright, secrets, semantic-title, triage, and submodule checks pass.

**Combined mixed-refit compatibility run**

This run includes PR #3294 plus the dependent mixed BF16/MXFP8 refit changes; it is not a PR #3294-only performance result.

- Qwen3.5 35B-A3B, 16x GB200, colocated sync CUDA IPC, CUDA Graph, first 3 / last 5 layers BF16, middle-layer QKVO and routed experts MXFP8: **20/20 steps completed**.
- Steps 2-19: 334.55 E2E tokens/s/GPU, 2,386.89 generation tokens/s/GPU, 243.85 s E2E step time, 9.72 s total refit.
- `gen_kl_error`: mean 0.001732, range 0.001550-0.001860. No refit, shape, manifest, or OOM error occurred.
- [W&B run](https://wandb.ai/nvidia/nemo-rl-pr3294-mixed-sync/runs/66ue66te)

This also exercises the grouped MXFP8 scale-sidecar routing added in this update. The old path failed on the first Qwen refit; this run completed all repeated refits.

seonjinn:
/ok to test [cfbe773](https://github.com/seonjinn/RL/commit/cfbe773c70a61f55835e6ed15339c61e3ae844ed)

nemo_rl/modelopt/models/generation/vllm_quant_backend.py
seonjinn: Fixed. The override now accepts `serialized_fp8_config` and forwards it to the base implementation on the non-real-quant path. The two-argument worker call is covered.

nemo_rl/models/policy/workers/megatron_policy_worker.py
seonjinn: Fixed. Slim cleanup now also checks `offload_optimizer_for_refit`, and the worker test covers both enabled and disabled values.

nemo_rl/models/generation/vllm/quantization/fp8.py
seonjinn: Fixed. Prequantized MXFP8 completeness is validated against the full negotiated refit manifest, so value and scale entries may cross transfer-batch boundaries.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
seonjinn: Fixed. The FlashInfer scale output is reshaped explicitly to `(*weight.shape[:-1], K // 32)`, preserving the block dimension for `K == 32` and supporting the guarded 2D/3D forms.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
```diff
@@ -13,6 +13,88 @@
...
+
+    if x.ndim == 2:
+        scales_uint8 = scales_uint8.view(x.shape[0], -1)
+    elif x.ndim == 3:
+        scales_uint8 = scales_uint8.view(x.shape[0], x.shape[1], -1)
+
+    return x_fp8, scales_uint8
+
+
+def mxfp8_e4m3_quantize_for_refit(
```
sharonyu-115, line 61: Could we discuss the tradeoff in **where quantization happens**, beyond whether the two methods produce equivalent outputs? This PR already preserves receiver-side quantization as the default and adds an optional trainer-side path.

**The tradeoff seems to be tighter alignment with the inference runtime versus lower transfer cost.**

- **[Receiver-side](https://github.com/NVIDIA-NeMo/RL/blob/012215177b65ca7683d436401d8c0e28b96ea297/nemo_rl/models/generation/vllm/quantization/fp8.py#L609):** BF16 weights are transferred to rollout and quantized through vLLM's wrapper in the rollout runtime's own virtual environment. This helps keep quantization aligned with the installed vLLM version, its dependencies, and the inference kernels that consume the result. The cost is transferring the larger BF16 payload and doing quantization on the receiver.
- **[Trainer-side](https://github.com/NVIDIA-NeMo/RL/blob/012215177b65ca7683d436401d8c0e28b96ea297/nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py#L61):** a separate helper runs in the [Megatron worker environment](https://github.com/NVIDIA-NeMo/RL/blob/012215177b65ca7683d436401d8c0e28b96ea297/nemo_rl/distributed/actor_environments.py#L65), calls FlashInfer directly, maintains a local PyTorch fallback, and reproduces scale reshaping and zero-scale handling. Sending quantized values and scales reduces transfer volume and receiver quantization work, but adds trainer work and another implementation/dependency path that must remain compatible with what vLLM consumes.

@seonjinn How do you weigh this tradeoff as vLLM evolves or more formats are supported? If we maintain both paths, could we share conversion logic or define the receiver's accepted format explicitly, with parity tests across the actual environments? Parity testing would help validate compatibility, while the broader question is the ongoing cost of maintaining that runtime alignment across environments.

The [reported 235B results](https://github.com/NVIDIA-NeMo/RL/pull/3294#issuecomment-5029291677) demonstrate a substantial gain for the combined optimizations. Could we isolate sender-side quantization's contribution, holding buffer reuse, slim offload, loader caching, and pinned reference swaps constant, and compare total refit time, full-step time, and peak/retained memory? That would help us judge the incremental benefit against the additional compatibility and maintenance burden.

sharonyu-115: @terrykong @guyueh1 what's your point of view on this?

seonjinn: Thanks for raising this tradeoff. I agree that receiver-side quantization should
remain the default: it follows the rollout environment's vLLM and FlashInfer
versions. Trainer-side prequantization is an opt-in performance path with an
explicit wire contract: E4M3 values preserve the logical checkpoint shape, and
E8M0 provides one scale byte per 32 values along the final dimension. The
rollout workers report which parameter names accept that contract, and the
trainer prequantizes only those names.

We also added a GB200 parity test that compares the sender helper with the
receiver helper byte-for-byte for values and scales in the target environment.
This validates the current contract; it does not assume that arbitrary future
vLLM versions will keep the same contract. Receiver-side quantization remains
the fallback if the contract changes.

The isolated historical A/B is Qwen3-30B-A3B, 16 GB200 GPUs, synchronous
colocated GRPO, GBS 2048, vLLM 0.25.1, and means over steps 3-20. Both rows use
a fixed 4 GiB staging size; reuse, batched shuffle, and loader caching are off.
This predates the current PR head, so I am treating it as supporting evidence,
not the final current-head result.

| Metric | Receiver-side quantization | Trainer-side prequantization | Change |
|---|---:|---:|---:|
| Transfer/update | 11.43 s | 9.46 s | -17.3% |
| Total refit | 15.47 s | 13.54 s | -12.5% |
| E2E step | 191.87 s | 189.96 s | -1.0% |
| E2E throughput | 2,159.84 tok/s/GPU | 2,181.90 tok/s/GPU | +1.02% |

The payload and explicit transient memory can be calculated from the tensor
shapes. MXFP8 sends 32 E4M3 bytes plus one E8M0 byte for every 32 values,
instead of 64 BF16 bytes, so each quantized tensor is
`1 - 33/64 = 48.4375%` smaller.

For this model, the routed-expert scope contains
`48 layers * 128 experts * 3 matrices * 2,048 * 768 = 28,991,029,248`
values. For one logical model copy, the routed-expert payload is therefore
`54.000 GiB BF16 -> 27.844 GiB MXFP8`, a 26.156 GiB reduction. Including the
non-expert tensors that stay BF16, the logical full-model payload is
`56.871 GiB -> 30.714 GiB`, or 45.99% smaller. These are payload numbers, not
claims of an equal peak-memory reduction.

The implementation streams one exported HF tensor at a time. Bridge exports
the largest Qwen expert tensor as grouped `gate_up_proj` with shape
`[128, 1536, 2048]`. Its BF16 source is 768 MiB and already exists in both
paths. Trainer-side quantization returns 384 MiB of E4M3 values plus 12 MiB of
scales. The out-of-place zero-scale clamp can briefly hold another 12 MiB, so
the calculated explicit sender-side transient is about 408 MiB before
FlashInfer workspace. Receiver-side quantization creates the equivalent output
in the rollout process instead. In this colocated run, the optimization mainly
moves that transient allocation between processes on the same GPU; it does not
retain a second model copy.

The prequantization state retained between refits is parameter-name metadata
only. The two persistent IPC buffers total 4 GiB per sending policy worker, but
that allocation is fixed and identical in the OFF and ON arms, so it is not an
incremental prequantization cost. Loader caching and slim offload also retain no
copied weight tensor.

The remaining measurement is a strict current-head OFF/ON pair with all other
optimizations held constant, plus a separately instrumented pair for worst-rank
GPU peak/retained memory and worst-node host memory. I will update this thread
with those measured numbers; allocator fragmentation and FlashInfer workspace cannot be inferred
reliably from tensor shapes alone.

PR #3295, opened 2026-07-20
feat(grpo): reuse generation logprobs as IS prev-logprobs

seonjinn:
## What does this PR do?

Adds `grpo.use_generation_logprobs_as_prev` (default `false`).

Under importance sampling, the trainer recomputes prev-logprobs with a full forward pass every step - recomputing values the inference engine already produced during sampling. Reusing them removes that forward entirely and is exact by construction: the engine **is** the behavior policy, so its processed sampling logprobs are the true denominator of the IS ratio.

- Delivers the logprob-phase savings of `force_on_policy_ratio: true` (cf. the recent perf-recipe PR) while **keeping** importance-sampling correction.

## Benchmark setup

- 20 training steps per run; reported values: arithmetic mean of steps 3-20
- Qwen3-30B-A3B, MXFP8 rollout, B200, 2 nodes x 8 GPU

## Logprob impact

| Metric | Value | Delta |
|---|---|---|
| Logprob time (s) | 87.5 -> 42.8 | -51.1% |
| Logprob tok/s/GPU | 4,740 -> 9,692 | +104% |

## E2E impact

| Metric | Value | Delta |
|---|---|---|
| Step time (s) | 257.2 -> 205.5 | -20.1% (measured with the refit-optimization flags also enabled) |

Isolated-branch reproduction runs are in flight; numbers will be posted as a follow-up comment.

## Convergence / correctness

Seed-matched pairs at 2 seeds, up to 200 steps: train-reward window means within 0.007 in 9 of 9 windows; held-out validation accuracy equal-or-better at every checkpoint; per-step reward curves overlap.

## Trade-off

`gen_kl_error` becomes trivially 0 under this flag (it would compare the reused values with themselves) - monitor reference KL / `token_mult_prob_error` instead. Documented in the config comment.


seonjinn:

Setup: Qwen3-30B-A3B GRPO (OpenMathInstruct-2), 2 nodes x 8 B200, `grpo-qwen3-30ba3b-4n4g-mxfp8-rollout.yaml` (real-IS: `use_importance_sampling_correction=true`, `force_on_policy_ratio=false`), 20 steps, means over steps 3-20. Both arms run this branch; the only difference is `grpo.use_generation_logprobs_as_prev` off vs on.

| steps 3-20 mean | flag off (main behavior) | flag on | delta |
|---|---|---|---|
| policy + reference logprobs (s) | 84.6 | **43.5** | **-41.1 s** |
| total step (s) | 256.3 | **216.2** | **-15.6%** |
| tokens/s/GPU | 1622 | 1923 | +18.6% |
| generation / training (s) | 49.8 / 91.0 | 50.3 / 92.1 | unchanged |
| train reward | 0.528 | 0.529 | = |

Clean isolation: the total-step saving (-40.1 s) equals the logprob-phase saving (-41.1 s) within noise - the removed block is exactly the per-step behavior-policy recompute. The remaining 43.5 s is the reference-policy logprob pass, which the KL penalty still needs and this PR does not touch.

seonjinn:
/ok to test [1358dcc](https://github.com/NVIDIA-NeMo/RL/pull/3295/commits/1358dcc5ec8a44b62f2337074d11254f7c7ff33f)

seonjinn:
/ok to test [734d5a8](https://github.com/NVIDIA-NeMo/RL/pull/3295/commits/734d5a8c5eaf7af57eb44146f8c977b9415e88c1)

seonjinn:
/ok to test [a2c85ba](https://github.com/NVIDIA-NeMo/RL/pull/3295/commits/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55)

tests/unit/algorithms/test_grpo.py
```diff
@@ -2414,6 +2415,123 @@ def test_grpo_train_skips_prev_logprobs_when_force_on_policy_ratio(
...
+    assert torch.equal(
+        trained_data["prev_logprobs"], trained_data["generation_logprobs"]
+    )
+
+
+def test_validate_generation_logprobs_as_prev_rejects_importance_sampling():
+    """PR #3295 guard: prev==generation collapses the IS weight, so the flag is
+    incompatible with use_importance_sampling_correction (and thus TIS)."""
+    with pytest.raises(AssertionError, match="use_importance_sampling_correction"):
+        _validate_generation_logprobs_as_prev(
```
terrykong: [`tests/unit/algorithms/test_grpo.py:2497`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/tests/unit/algorithms/test_grpo.py#L2497)

**Must-fix:** `_validate_generation_logprobs_as_prev` is called here (and at L2507/L2518/L2529) but never imported — the import block at [L29–L46](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/tests/unit/algorithms/test_grpo.py#L29) doesn't include it, so all four new guard tests fail at collection with `NameError` (ruff flags F821 ×4).

**Action:** add it to the `nemo_rl.algorithms.grpo` import list:
```python
    _should_use_async_rollouts,
    _validate_generation_logprobs_as_prev,
    aggregate_rollout_metrics,
```

nemo_rl/algorithms/grpo.py
```diff
@@ -339,6 +349,45 @@ class MasterConfig(BaseModel, extra="allow"):
...
+        "same term). Use it only in the on-policy regime with importance "
+        "sampling correction disabled."
+    )
+    assert not loss_config.force_on_policy_ratio, (
+        "grpo.use_generation_logprobs_as_prev=True is mutually exclusive with "
+        "loss_fn.force_on_policy_ratio=True: force_on_policy_ratio overrides "
+        "prev_logprobs with curr_logprobs.detach() inside the loss, so the reused "
+        "generation logprobs would be silently ignored. Pick one on-policy strategy."
+    )
+    print("  ✓ use_generation_logprobs_as_prev enabled (on-policy, IS correction off)")
```
terrykong: [`nemo_rl/algorithms/grpo.py:388`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L388)

With `prev == generation`, the seq-level logprob-error masking ([`compute_and_apply_seq_logprob_error_masking` keys off `|generation_logprobs - prev_logprobs|`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L2387)) is identically zero — so a user who sets `grpo.seq_logprob_error_threshold` gets a safety filter that silently never triggers (the analogous `force_on_policy_ratio` path [warns](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L2959); the new unit test itself sets the threshold to `None`). Since this guard already receives `grpo_config`, it can fail loudly — one way to do it (a warning is also defensible):

```suggestion
    assert grpo_config.get("seq_logprob_error_threshold") is None, (
        "grpo.use_generation_logprobs_as_prev=True makes prev_logprobs == "
        "generation_logprobs, so seq_logprob_error masking (which keys off "
        "|generation_logprobs - prev_logprobs|) is identically zero and "
        "grpo.seq_logprob_error_threshold would silently never trigger. "
        "Unset seq_logprob_error_threshold (null) to use this flag."
    )
    print("  ✓ use_generation_logprobs_as_prev enabled (on-policy, IS correction off)")
```

nemo_rl/algorithms/grpo.py
```diff
@@ -339,6 +349,45 @@ class MasterConfig(BaseModel, extra="allow"):
 # ===============================================================================
 
 
+def _validate_generation_logprobs_as_prev(
+    grpo_config: GRPOConfig, loss_config: ClippedPGLossConfig
+) -> None:
+    """Guard grpo.use_generation_logprobs_as_prev against incompatible loss settings.
```
terrykong: [`nemo_rl/algorithms/grpo.py:355`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L355)

The reuse branch exists only in sync [`grpo_train`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L2998); [`async_grpo_train` always recomputes `prev_logprobs`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L4360) and never reads the flag. Today flag+async always errors before training, but only *incidentally*: this guard requires IS-correction off while async [asserts it on](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L3798) — and in the IS-off case the resulting async error never names this flag (while the "✓ enabled" line has already printed). If on-policy async ever relaxes that assertion, flag+async becomes a silent no-op.

**Action:** make the sync-only contract explicit here, e.g. `assert not _should_use_async_rollouts(master_config)` (the guard is called from `setup()`, which has `master_config`) with a message naming `grpo.use_generation_logprobs_as_prev` as sync-only.

tests/unit/algorithms/test_grpo.py
```diff
@@ -2414,6 +2415,123 @@ def test_grpo_train_skips_prev_logprobs_when_force_on_policy_ratio(
...
+        )
+
+    assert not policy.get_logprobs.called, (
+        "policy.get_logprobs was called even though "
+        "use_generation_logprobs_as_prev=True. The engine's generation logprobs "
+        "should be reused as prev_logprobs. This is a regression of PR #3295."
+    )
+    policy.train.assert_called_once()
+    trained_data = policy.train.call_args[0][0]
+    assert torch.equal(
```
terrykong: [`tests/unit/algorithms/test_grpo.py:2488`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/tests/unit/algorithms/test_grpo.py#L2488)

Minor test-quality note: this `torch.equal` is vacuous as written — `_patched_logprob_phase` sets both `generation_logprobs` and the `get_logprobs` return value to all-zeros, so the equality holds regardless of which source populated `prev_logprobs`. The `assert not policy.get_logprobs.called` above is the real regression guard (and it's a good one), so this is optional: giving `generation_logprobs` a distinctive non-zero sentinel (which means parameterizing the shared `_patched_logprob_phase` helper) would make the value check independently meaningful.

nemo_rl/algorithms/grpo.py
```diff
@@ -2943,7 +2995,15 @@ def grpo_train(
                         logprob_data, flat_messages, master_config.policy
                     )
 
-                    if not skip_prev_logprobs:
+                    if master_config.grpo.get("use_generation_logprobs_as_prev"):
+                        # The importance-sampling behavior policy is the vLLM
+                        # engine itself, so its returned processed logprobs are
```
terrykong: [`nemo_rl/algorithms/grpo.py:2999`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L2999)

Doc nit: `generation_logprobs` is emitted by all generation backends (sglang and megatron workers too), so "the vLLM engine" reads narrower than the feature is. Same wording appears in the `examples/configs/grpo_math_1B.yaml` comment (L19) and the `GRPOConfig` comment (L274).

```suggestion
                        # The importance-sampling behavior policy is the generation
                        # engine itself, so its returned processed logprobs are
```

nemo_rl/algorithms/grpo.py
```diff
@@ -2943,7 +2995,15 @@ def grpo_train(
                         logprob_data, flat_messages, master_config.policy
                     )
 
-                    if not skip_prev_logprobs:
+                    if master_config.grpo.get("use_generation_logprobs_as_prev"):
```
terrykong: [`nemo_rl/algorithms/grpo.py:2998`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L2998)

FYI only — nothing to do in this PR: this optimization currently lands in the legacy sync `grpo_train` only. The Single Controller sync dataplane still always pays the prev-logprobs forward ([`grpo_sync.py` fills it via `policy.get_logprobs_from_meta`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo_sync.py#L787-L830)). Porting it there is now tracked in #3336. (The async path is intentionally out of scope — the feature is sync-only by nature, since async requires the IS correction this flag forbids.)

terrykong (commented):
Thanks for a well-evidenced PR — benchmarks, seed-matched convergence runs, and the isolated-branch repro make this easy to trust, and extracting [`_validate_generation_logprobs_as_prev`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/grpo.py#L352) as a pure function (vs. the inline sibling guards) is a nice touch that enabled direct unit tests.

**Mathematical correctness (verified):** the reuse is sound in the regime the guard allows. The PPO ratio [`exp(curr - prev)`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/loss/loss_functions.py#L523) never degenerates — `curr` is a fresh trainer forward on every minibatch update; only `prev` is frozen. The off-policy IS weight [`exp(prev - generation)`](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/loss/loss_functions.py#L575) is the term that collapses to 1, which is exactly why the mutual exclusion with `use_importance_sampling_correction` (and transitively TIS, via the [assert that TIS requires it](https://github.com/NVIDIA-NeMo/RL/blob/a2c85bae8f66f7e7b4fbc93e8e0bc43f8fd1ba55/nemo_rl/algorithms/loss/loss_functions.py#L288)) is right and necessary. This matches verl's `bypass_mode` verbatim ([core_algos.py — "old_log_prob = rollout_log_prob … we do NOT apply additional IS weights (would be double-counting)"](https://github.com/volcengine/verl/blob/2b47a68b66d6fa21884990a5d4445ceba59c47e5/verl/trainer/ppo/core_algos.py#L2374)) and slime's `use_rollout_logprobs` ([loss.py#L912](https://github.com/THUDM/slime/blob/aaf5c2092b01219fa0d5c2d323741d409086ca32/slime/backends/megatron_utils/loss.py#L912)).

Non-blocking observations (no action required):
- slime keeps a separate trainer forward ([`train_log_probs_for_tis`](https://github.com/THUDM/slime/blob/aaf5c2092b01219fa0d5c2d323741d409086ca32/slime/backends/megatron_utils/loss.py#L930)) so TIS can coexist with rollout-logprob reuse — the mutual exclusion here is a justified simplification (TIS would need the very forward this flag eliminates), just noting it's a design choice rather than a mathematical necessity.
- With the flag on, the engine/trainer numerical mismatch is absorbed into the PPO clip (as the guard docstring says); under MXFP8 rollouts that can push first-update ratios toward the clip boundary — might be worth one extra line in the docstring.
- Tiny nit: the docstring of `test_validate_generation_logprobs_as_prev_allows_on_policy_no_is` ("with IS correction and force_on_policy off") reads ambiguously; "with IS correction disabled and force_on_policy_ratio off" would be clearer.

Inline: 1 must-fix (missing test import → NameError/ruff F821) and 3 small hardening/doc suggestions.

_Generated by Claude Code_

terrykong:
@yfw can you review as well. want to make sure we aren't missing some edge case where we shouldn't use this

nemo_rl/algorithms/grpo.py
terrykong: cc @ZhiyuLi-Nvidia

ZhiyuLi-Nvidia:
Hi, is there a similar flag force_on_policy_ratio which will bypass prev-logprobs?

PR #3297, opened 2026-07-20
fix(vllm): record logprob for the sampled token

HollowMan6:
# What does this PR do ?

Record logprob for the sampled token

# Issues

https://github.com/NVIDIA-NeMo/RL/pull/2608#discussion_r3612313374

Fix vLLM generation output processing to record the log probability of the token actually sampled, rather than relying on the first entry in the per-position logprob dictionary.

vLLM returns each generated position as a mapping from token ID to logprob metadata. When that mapping contains multiple tokens, the first entry is not guaranteed to be the sampled token.

For example, if token 880 was sampled:

```bash
{
    512: Logprob(logprob=-0.10, rank=1),
    880: Logprob(logprob=-2.30, rank=2),
}
```

The previous implementation recorded -0.10 by taking the first dictionary entry. The correct value for the generated token is -2.30.

So Pair each generated token ID with its corresponding logprob dictionary and look up the logprob by that token ID.

The behavior is unchanged when vLLM returns only the sampled token.

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/models/generation/vllm/vllm_worker.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker.py
@@ -814,12 +814,15 @@ class VllmGenerationWorkerImpl(BaseVllmGenerationWorker):
             full_logprobs = torch.zeros(total_length, dtype=torch.float32)
             if hasattr(generation, "logprobs") and generation.logprobs:
                 try:
-                    for idx, logprob_dict in enumerate(generation.logprobs):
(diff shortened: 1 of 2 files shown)


HollowMan6:
/ok to test https://github.com/NVIDIA-NeMo/RL/commit/bdb414e52247807970025273b2d1f0a28794a7d0

terrykong (approved):

PR #3300, opened 2026-07-21
perf(grpo): make non-colocated refit offload optional

wdykas:
# What does this PR do?

  Makes policy offloading optional before non-colocated Megatron weight refits.

  Non-colocated refits currently move training gradients and optimizer state to CPU before every weight transfer to provide additional GPU memory for staging. This is necessary for memory-constrained configurations, but introduces avoidable CPU/GPU transfer overhead when sufficient GPU memory is already available.

  This PR adds:

  - `policy.generation.mcore_generation_config.offload_policy_before_refit`
  - `true` preserves the existing memory-safe behavior
  - `false` skips policy offloading, reducing refit latency when GPU memory permits
  - Phase-level timing for suspend, offload, weight preparation, transfer, KV-cache preparation, and resume
  - Unit coverage for both enabled and disabled configurations
  - The new setting across shipped Megatron example and reference configurations

  The option only affects non-colocated Megatron refits. Colocated behavior remains unchanged.

  # Issues

  N/A

  # Usage

  ```yaml
  policy:
    generation:
      mcore_generation_config:
        # Keep enabled when refit staging requires additional GPU headroom.
        # Disable when sufficient GPU memory is available to avoid offload overhead.
        offload_policy_before_refit: false

  # Validation

  - [x] Added unit coverage for enabled and disabled policy offloading
  - [x] 59 weight-synchronizer unit tests passed in an isolated/mock-based run
  - [x] Ruff checks passed for the touched Python files
  - [x] Configuration parity checks passed
  - [x] Verified the setting is present in all 18 resolved Megatron configurations
  - [ ] Full GPU functional coverage will run in CI:megatron


shanmugamr1992 (dismissed):

cspades (approved):
I think this gets rid of some very nasty slowdowns reallocating gradient buffers when you are squeezing memory and use `expandable_segments:True`. Might try this PR soon.

tdene (approved):

wdykas (commented):
LGTM.

Verified locally at 41deac5a:
- All 18 resolved Megatron-backend configs under `examples/` (including every `defaults:` chain through `grpo_math_1B.yaml`, `ppo_math_1B.yaml`, `grpo_nanov3.yaml`, `grpo_qwen3_30ba3b_instruct.yaml`) carry the new required `offload_policy_before_refit` key, so the direct `cfg[...]` read in `MegatronWeightSynchronizer.sync_weights` cannot `KeyError` on shipped configs.
- Every other `offload_before_refit()` caller is colocated-only (`grpo.py` inline vLLM/IPC path, the defer-wake-for-save branch, `MegatronGeneration.__init__` HTTP path) or the distillation teacher path, so nothing else needs to honor the flag.
- No consumer aggregates `prepare_for_generation/*` sub-keys (only `prepare_for_generation/total` is read in `algorithms/utils.py`), so the new phase timers can't double-count.
- `Timer.time` is a `@contextmanager`, so `timed_phase`'s deferred entry is behavior-identical to the inline form.

Two non-blocking notes:
- The second commit flips the effective default to offload-off for all non-colocated Megatron runs. The description is accurate that `true` restores the old behavior, but worth calling out in release notes as a behavior change (memory-tight non-colocated runs may now need to set it back to `true`).
- `codecov/patch` is red (73.91% vs 80%). None of the synchronizer unit tests pass a `Timer` into `sync_weights`, so the timer branch of `timed_phase` is never exercised; a single test with a real `Timer()` asserting the phase keys would cover it and pin the metric names.

terrykong (approved):

PR #3308, opened 2026-07-21
perf: avoid CPU round trip during NVFP4 refit

seonjinn:
## What does this PR do?

Real NVFP4 refit quantizes and packs policy weights on GPU, but the existing export path copies every packed tensor to CPU before colocated CUDA IPC copies it back to GPU for vLLM reload.

This PR adds opt-in `generation.real_quant_export_cpu_offload: false` for colocated CUDA-IPC refit. The CPU path remains the default, and GPU-resident export is rejected for non-colocated or explicit refit transports.

## Benchmark setup

- Nemotron Nano 30B-A3B, DAPOMath17K, seed 42
- 2 B200 nodes x 8 GPUs
- 100 training steps per run; means over steps 3-100
- Matched CPU-offload and GPU-resident export runs for W4A16 and W4A4 rollout

## Impact

| Metric | W4A16 | Delta | W4A4 | Delta |
|---|---:|---:|---:|---:|
| Transfer/update (s) | 21.85 -> 4.89 | -77.6% | 22.46 -> 5.78 | -74.3% |
| Refit total (s) | 27.62 -> 11.72 | -57.6% | 34.00 -> 18.22 | -46.4% |
| E2E step time (s) | 59.69 -> 45.10 | -24.5% | 71.57 -> 55.60 | -22.3% |
| E2E tok/s/GPU | 32.18 -> 41.01 | +27.4% | 26.03 -> 33.51 | +28.8% |

## Accuracy impact

| Metric | W4A16 CPU -> GPU | W4A4 CPU -> GPU |
|---|---:|---:|
| Mean rollout reward | 0.1773 -> 0.2864 | 0.2723 -> 0.2487 |

These are single-seed sampled training rewards, not held-out accuracy. The W4A4 paired step-level reward difference has a 95% CI of `[-0.0711, +0.0239]`, which includes zero. Training trajectories diverge after the first update, so this benchmark does not claim convergence parity.

## Correctness

- CPU offload remains the default and fallback for lower GPU-memory headroom.
- GPU-resident export is limited to colocated CUDA IPC with no explicit refit transport.
- W4A16 and W4A4 ModelOpt exports match CPU offload in tensor names, shapes, dtypes, and raw bytes.
- 11 focused CUDA tests passed on B200, including CUDA IPC staging parity.
- All four matched 100-step benchmark arms completed without OOM, NaN, assertion, refit, or Ray actor failures.

## Trade-off

GPU-resident export retains packed tensors on the source GPU during refit, trading the host round trip for additional transient GPU memory. Users can keep the default CPU path when memory headroom is limited.


terrykong (commented):
@parthmannan can you also review?

tests/unit/models/policy/test_megatron_quant_worker.py
```diff
@@ -273,6 +275,68 @@ def test_iter_real_quant_refit_params_uses_megatron_bridge_export():
...
+    list(worker._iter_real_quant_refit_params())
+
+    _, kwargs = worker.megatron_bridge.calls[0]
+    assert kwargs["cpu"] is False
+
+
+@pytest.mark.parametrize("quant_mode", ["w4a16_nvfp4", "nvfp4"])
+@requires_quant
+@requires_weight_folding
+def test_modelopt_real_quant_cpu_and_gpu_exports_are_byte_identical(
```
terrykong, line 299: [`tests/unit/models/policy/test_megatron_quant_worker.py:292`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/tests/unit/models/policy/test_megatron_quant_worker.py#L292)

This byte-identical parity test is the PR's central correctness proof, but this file currently runs in **zero CI lanes**: it has no `pytestmark = pytest.mark.mcore`, so the Mcore_Policy shards (the only lanes with megatron/modelopt deps and a GPU) [filter it out via `--mcore-only`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/tests/unit/conftest.py#L114), and every other lane skips it via `@requires_weight_folding` since megatron isn't installed there. The sibling megatron worker test files all carry the mark, e.g. [`test_megatron_worker.py:45`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/tests/unit/models/policy/test_megatron_worker.py#L45). This gap is pre-existing for the whole file, but it would be great to close it here so this test (and the other new ones) actually run in CI.

**Action:** add `pytestmark = pytest.mark.mcore` near the top of `test_megatron_quant_worker.py` (after the imports). The existing `@requires_quant`/`@requires_weight_folding` skipif guards still protect non-GPU mcore environments.

nemo_rl/models/generation/vllm/config.py
```diff
@@ -148,6 +148,9 @@ class VllmConfig(GenerationConfig):
     # NVFP4 kernels and stream packed quantized weights instead of fake-quant
     # modules. This is intended for ModelOpt NVFP4 rollout experiments.
     real_quant: NotRequired[bool]
+    # CPU offload remains the default. Disabling it is supported only for
+    # colocated CUDA-IPC refit, where packed export tensors can stay on GPU.
+    real_quant_export_cpu_offload: NotRequired[bool]
```
terrykong, line 153: [`nemo_rl/models/generation/vllm/config.py:153`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/nemo_rl/models/generation/vllm/config.py#L153)

The `policy.generation` parameter table in [`docs/guides/quantization-aware-rl.md`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/docs/guides/quantization-aware-rl.md#L327) documents the sibling keys `quant_cfg`, `real_quant`, and `real_quant_ignore`, so it would be helpful to add a row for the new `real_quant_export_cpu_offload` key there too (anchoring this comment here since the docs file isn't in the diff).

**Action:** add a row to the table after `real_quant_ignore`, e.g.:

> | `real_quant_export_cpu_offload` | Optional bool (default `true`). When `true`, packed NVFP4 export tensors are copied to CPU before refit (safe on every transport). Set `false` only for colocated CUDA-IPC refit to keep export tensors GPU-resident and avoid a GPU→CPU→GPU round trip, at the cost of transient GPU memory. Rejected at setup time for non-colocated or explicit `refit_transport`. |

nemo_rl/models/generation/__init__.py
```diff
@@ -45,6 +45,21 @@ def configure_generation_config(
...
     if config["backend"] == "vllm":
         config = cast(VllmConfig, config)
+        if config.get("real_quant"):
+            export_cpu_offload = config.get("real_quant_export_cpu_offload")
+            if not isinstance(export_cpu_offload, bool):
+                raise ValueError(
+                    "generation.real_quant_export_cpu_offload must be a boolean"
+                )
+            if not export_cpu_offload and (
+                not config["colocated"]["enabled"]
```
terrykong: [`nemo_rl/models/generation/__init__.py:55`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/nemo_rl/models/generation/__init__.py#L55)

Minor: `colocated` is [`NotRequired` in `GenerationConfig`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/nemo_rl/models/generation/interfaces.py#L202), so on this opt-in `real_quant_export_cpu_offload=false` path a config that omits `colocated` raises a raw `KeyError` instead of the intended clean `ValueError`. Unreachable for shipped configs (they all inherit `colocated`), so feel free to take or leave this. One way to do it — treating a missing `colocated` block as not-colocated (which correctly rejects GPU export):

```suggestion
                not config.get("colocated", {}).get("enabled", False)
```

terrykong (commented):
Team review of the NVFP4 GPU-resident export path. Nice work — the benchmarks in the description are thorough, and the safety story checks out under scrutiny: all in-repo `real_quant` recipes inherit the new key via their `defaults:` chains rooted at [`grpo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/examples/configs/grpo_math_1B.yaml#L347); [`configure_generation_config`](https://github.com/NVIDIA-NeMo/RL/blob/9643550c4d665802cbca835c150e9c40077a5bda/nemo_rl/models/generation/__init__.py#L33) runs on every entrypoint before workers are built; `refit_transport: null` is the only route to CUDA IPC so the guard has no false positives; and the `cpu=False` path yields a fresh packed tensor with stream-synchronized IPC staging, so there is no aliasing hazard. 3 inline suggestions below — the test-lane one is the only item worth addressing before merge.

_Generated by Claude Code_

nemo_rl/models/generation/__init__.py
seonjinn: Fixed in `30b413920`. Missing optional `colocated` configuration now follows the same topology validation path and raises the intended `ValueError`; a regression test covers the omitted block.

nemo_rl/models/generation/vllm/config.py
seonjinn: Fixed in `30b413920`. Documented `real_quant_export_cpu_offload`, its default, supported colocated CUDA-IPC topology, CPU fallback, and transient GPU-memory tradeoff in the quantization guide.

tests/unit/models/policy/test_megatron_quant_worker.py
seonjinn: Fixed in `30b413920`. Added the module-level `pytest.mark.mcore`; the existing quantization and weight-folding guards remain in place.

terrykong: fixed in 30b41392001402b0f09df63763773216542efbc8

nemo_rl/models/generation/vllm/config.py
terrykong: fixed in 30b41392001402b0f09df63763773216542efbc8

nemo_rl/models/generation/__init__.py
terrykong: fixed in 30b41392001402b0f09df63763773216542efbc8

terrykong (approved):

Issue #2953, opened 2026-06-26
PPO DTensor value-head initialization is nondeterministic across Ray worker launches


yuki-97:
## Description

The original report observed a much higher initial critic loss for DTensor PPO with value dynamic batching, while reward remained aligned: [original PR discussion](https://github.com/NVIDIA-NeMo/RL/pull/2837#discussion_r3480775913).

<img width="1972" height="626" alt="Original DB-OFF/DB-ON comparison" src="https://github.com/user-attachments/assets/8ea9a1c4-7d1b-47b6-8baa-664b81c1c428" />

Follow-up experiments show that the mismatch was not caused by dynamic batching. It came from nondeterministic initialization of the DTensor value head across Ray worker launches.

## Root cause

The PPO controller seeded its own process, but each DTensor value worker runs in a separate Ray process. The base causal-LM checkpoint does not contain a regression `score` head, so the head is initialized when the worker constructs the value model. Before the fix, the controller seed did not reach this initialization, and separate DB-OFF/DB-ON launches received different value heads.

MCore did not show the same cross-launch symptom because its worker setup already seeded model initialization. A shared numeric seed does not make fresh MCore and Automodel heads identical, because the two backends use different head implementations and initialization schemes.

## Resolution

PR #3321 makes fresh DTensor value-model initialization reproducible and enables value dynamic batching:

- add `seed` to the value configuration and default it to `${ppo.seed}`;
- seed each DTensor value worker after distributed setup and before checkpoint-manager/model construction;
- add fresh-initialization and checkpoint-pinned TP2+SP equivalence coverage for DB-OFF versus DB-ON.

## Validation

- Two independent Automodel launches with `seed=42` saved the value model before the first optimizer step. All 584 checkpoint tensors matched exactly, including `score.weight` and `score.bias`:
  - [repeat 1](https://wandb.ai/nvidia/nemo-rl-ppo-port-v1/runs/pr3321-automodel-seed42-r1)
  - [repeat 2](https://wandb.ai/nvidia/nemo-rl-ppo-port-v1/runs/pr3321-automodel-seed42-r2)
- Fresh TP2+SP 100-step DB-OFF/DB-ON runs with `seed=42` closely align without sharing a checkpoint:
  - [DB OFF](https://wandb.ai/nvidia/nemo-rl-ppo-port-v1/runs/seed42-db-off-100)
[...]

PR #2837, opened 2026-06-16
feat: Support for dtensor ppo

fujial-code:
# What does this PR do ?

Adds DTensor/FSDP2 support for the PPO value model path and aligns the DTensor PPO 1.5B GSM8K recipe with the existing Megatron-Core PPO baseline.
The main correctness fix is a value temporal alignment issue: DTensor value inference/training was not using the same `V(s_t)` convention as the Megatron PPO path. After right-shifting token values to match Megatron's temporal semantics, the DTensor reward curve aligns with the Megatron baseline on the Qwen2.5-1.5B GSM8K 1-node recipe.

# Issues
close [[#2047]](https://github.com/NVIDIA-NeMo/RL/issues/2047)

# Summary of Changes
## DTensor PPO Value Model
- Adds `DTensorValueWorkerV2` for PPO value model training and inference on the DTensor/FSDP2 backend.
- Wires DTensor value worker selection through `LMValue`.
- Supports PPO value inference, value training, and checkpoint save/load paths for the DTensor backend.
- Adds lifecycle compatibility with the PPO loop, including inference/training finish hooks where needed.
## Value Temporal Alignment
- Fixes the DTensor value convention to match the Megatron PPO path.
- Applies right-shifted values so token-level value predictions correspond to `V(s_t)`.
- This was the key fix for the DTensor vs Megatron reward mismatch observed during GSM8K validation.
## Recipe and Config Alignment
- Adds the DTensor 1-node 8-GPU Qwen2.5-1.5B GSM8K PPO recipe:
  - `examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-dtensor.yaml`
- Aligns the DTensor recipe with the Megatron recipe for A/B comparison:
  - `ppo_epochs: 1`
  - `math_verify_impl: hf_math_verify`
  - matching value loss config structure
- Removes the unused GSM8K verifier-specific path and uses the existing HF math verifier for the GSM8K recipe.
## Runtime Compatibility Fixes
- Updates DTensor worker call sites to match current Automodel API signatures.
- Removes stale kwargs that no longer exist in the current forward/checkpoint APIs.
- Passes checkpointing config when saving the PPO value model checkpoint.
- Adds missing imports and no-op lifecycle methods needed by the PPO loop.
## Tests

- **Unit tests**: Added DTensor PPO value temporal alignment tests for `_right_shift_values` and `_RightShiftLossWrapper`.
[...]

PR #3320, opened 2026-07-23
feat: support sequence packing and context parallelism for DTensor value models

### LINKED ISSUE #2951: PPO Dtensor value model support sequence packing + CP
Tracking missing features in https://github.com/NVIDIA-NeMo/RL/pull/2837.
- [ ] sequence packing
- [ ] CP

fujial-code:
# What does this PR do?

Adds sequence packing and context parallelism support to the DTensor/FSDP2 PPO value-model path.

This follows the DTensor PPO value-model support introduced in #2837 and closes the two missing features tracked in #2951.

The main correctness work is specific to scalar token-level value predictions:

- Sequence packing must unpack packed scalar scores back to the original batch and sequence layout.
- Context parallelism must restore the global sequence order before applying the PPO value right shift and computing the value loss.
- PPO rollout batches must be padded according to the value model's CP topology because the policy and value models may use different parallel configurations.

Sequence packing and context parallelism are supported as independent modes. Their existing mutual-exclusion validation remains in place.

# Issues

Closes [#2951](https://github.com/NVIDIA-NeMo/RL/issues/2951)

# Summary of Changes

## DTensor Value Sequence Packing

- Enables sequence packing for token-level regression reward/value models.
- Keeps sequence packing unsupported for Bradley-Terry and other sequence-level reward models.
- Passes FlashAttention arguments through the DTensor value training and inference paths when sequence packing is enabled.
- Extends `ScorePostProcessor` to reconstruct padded per-sample value tensors from packed outputs using:
  - `cu_seqlens`
  - original input lengths
  - original batch size and sequence length
- Requires `float16` or `bfloat16` precision because FlashAttention does not support `float32` inputs.
- Removes the PPO setup guard that previously rejected DTensor value sequence packing.

## DTensor Value Context Parallelism

- Adds a value-specific loss postprocessor for context-parallel scalar logits.
- Gathers CP-sharded value logits into the original global token order before:
  - applying the temporal right shift
  - computing PPO value loss and clipping
  - calculating value statistics
- Updates value inference to gather CP shards before applying the same right-shift convention introduced in #2837.
- Corrects backward scaling for the replicated full-sequence value loss.
- Avoids double-counting metrics replicated across TP/CP ranks.
- Pads sequence-aligned PPO tensors to a multiple of `2 * context_parallel_size`, as required by load-balanced causal CP attention.
- Restores the original sequence length before returning value predictions.
- Forces the ring FlashAttention backend for CP to support Qwen2 GQA layouts where:
  - `num_attention_heads != num_key_value_heads`
- Requires `float16` or `bfloat16` precision for CP.
- Removes the PPO setup guard that previously rejected DTensor value CP.

## Configuration and Compatibility

- Sequence packing and CP remain mutually exclusive.
- Dynamic batching remains guarded separately by the existing #2953 validation.
- The implementation supports value-model topologies that differ from the policy topology.
- No changes are made to the Megatron value-model path.

## Tests

Added or extended coverage for:

- Sequence packing configuration validation.
- Regression value models accepting FlashAttention arguments.
- Packed scalar score reconstruction.
- CP gather-before-right-shift behavior.
- CP sequence-length validation and automatic padding.
- DTensor value initialization and inference with:
  - TP=2 sequence packing
  - CP=2
- DTensor value training with sequence packing and CP.
- Pinned-weight inference equivalence against the feature-OFF reference.
- Pinned-weight deterministic train-forward/loss equivalence, including:
  - value loss
  - value mean
  - return mean
  - residual squared mean

# Validation

Validated with the Qwen2.5-1.5B-Instruct GSM8K PPO setup on one 8-GPU node.

Both comparisons resume from the same value-model checkpoint so differences cannot be attributed to random value-head initialization.

## Sequence Packing OFF vs ON

Ran a 100-step PPO comparison with matched settings:

- Value topology: TP=2
- Same initial checkpoint
- Same dataset and training configuration
- Sequence packing disabled vs enabled

Results:
image
image

- Both runs completed 100 PPO steps successfully.
- `train/reward` follows the same learning trajectory.
- `train/critic/loss` closely overlaps across training.
- Pinned inference and train-forward equivalence tests pass.

## Context Parallelism OFF vs ON

Ran a 100-step PPO comparison with matched settings:

- Same initial checkpoint
- Same dataset and training configuration
- CP disabled vs CP=2

Results at step 100:
image
image

- `train/reward`
  - CP OFF: `0.8955`
  - CP ON: `0.9023`
- `train/critic/loss`
  - CP OFF: `0.03147`
  - CP ON: `0.03260`

Both runs complete successfully, and the reward and critic-loss behavior remain aligned.

`train/loss` is the policy loss and is not expected to remain pointwise identical after the stochastic rollout trajectories diverge. The correctness checks for this PR therefore use pinned deterministic value inference/train-loss equivalence together with reward and critic-loss behavior.

# Local Checks

- Ruff formatting passes.
- Ruff lint checks pass.
- Focused CPU unit tests for the new setup and post-processing logic pass.
- Focused GPU worker tests pass for sequence packing and CP inference, training, and equivalence.
- 100-step pinned PPO comparisons complete successfully for both features.


nemo_rl/algorithms/ppo.py
```diff
@@ -283,19 +283,11 @@ def setup(
                 "value.sequence_packing.enabled=true."
             )
     else:
-        # DTensor PPO value model currently doesn't support sequence packing and CP.
+        # DTensor PPO value model does not support sequence packing.
```
yuki-97: this is actually what the PR is supporting, we can remove the comment here.
```suggestion
```

nemo_rl/models/automodel/setup.py
```diff
@@ -707,12 +723,23 @@ def setup_model_and_optimizer(
     from torch.nn.attention import SDPBackend
 
     if cp_size > 1:
-        # Match Automodel's `get_train_context` in `cp_utils.py` where only
-        # flash and efficient backends are supported
```
yuki-97, line 711: nit: can we keep the comments?

nemo_rl/models/automodel/train.py
```diff
@@ -538,7 +540,17 @@ def automodel_forward_backward(
 
                     # when FSDP reduces the gradients over the DP dim, they're automatically averaged
                     # but we want to sum them so we cancel out the average here
```
yuki-97: nit: can we move it to the up of `loss = result * dp_size * backward_cp_scale` to distinguish the comments for `backward_cp_scale`?

nemo_rl/models/automodel/train.py
```diff
@@ -136,11 +137,6 @@ def model_forward(
                 processed_inputs.input_ids
             )
 
-    # Reward models don't support flash_attn_kwargs
-    if is_reward_model:
-        if "flash_attn_kwargs" in model_args:
-            del model_args["flash_attn_kwargs"]
```
yuki-97, line 142: I remember we added it when supporting `examples/run_rm.py` or `rm_env`, could you confirm whether they work now without this?

nemo_rl/models/automodel/train.py
```diff
@@ -538,7 +540,17 @@ def automodel_forward_backward(
...
                     # but we want to sum them so we cancel out the average here
-                    loss = result * dp_size * cp_size
+                    # CP-sharded policy losses need the CP factor to cancel the
+                    # FSDP DP*CP gradient average. ValueLossPostProcessor first
+                    # all-gathers the full sequence and computes the same loss
+                    # on every CP rank; AllGatherCPTensor.backward already sums
+                    # those replicated output gradients across CP.
+                    backward_cp_scale = (
+                        1
+                        if getattr(post_processing_fn, "cp_loss_is_replicated", False)
```
yuki-97: can we use this instead? feels it's more directly.

also need:
1. update the comment to tell why `ValueLossPostProcessor` don't need cp scale.
2. remove `cp_loss_is_replicated` in `ValueLossPostProcessor`.
```suggestion
                        if isinstance(post_processing_fn, ValueLossPostProcessor)
```

nemo_rl/models/automodel/train.py
```diff
@@ -647,6 +659,64 @@ def __call__(
...
+            return super().__call__(
+                logits=logits,
+                data_dict=data_dict,
+                processed_inputs=processed_inputs,
+                global_valid_seqs=global_valid_seqs,
+                global_valid_toks=global_valid_toks,
+                sequence_dim=sequence_dim,
+            )
+
+        assert not self.enable_seq_packing, (
```
yuki-97: can we move the assert to ppo setup near https://github.com/NVIDIA-NeMo/RL/pull/3320/changes#r3712867680? so that we can get the assert at the beginning.

nemo_rl/models/value/lm_value.py
```diff
@@ -392,7 +392,21 @@ def train(
...
         all_mb_metrics = defaultdict(list)
+        dtensor_model_parallel_size = None
+        if self.cfg.get("dtensor_cfg", {}).get("enabled", False):
+            # DTensor value workers have PP=1, so each contiguous CP*TP rank
+            # block corresponds to one DP shard.
+            dtensor_model_parallel_size = self.sharding_annotations.get_axis_size(
+                "context_parallel"
+            ) * self.sharding_annotations.get_axis_size("tensor_parallel")
         for r in results:
+            # DTensor CP/TP outputs are replicated. Aggregate one representative
```
yuki-97: I'm feeling the changes in this file is no-op, since we have `output_is_replicated‎` so we should only get the results from rank 0. or am I missing something here?

could you help confirm if we can remove the changes in this file?

nemo_rl/models/value/workers/dtensor_value_worker_v2.py
```diff
@@ -73,6 +74,65 @@ def right_shift_values(values: torch.Tensor) -> torch.Tensor:
     return torch.cat([torch.zeros_like(values[:, :1]), values[:, :-1]], dim=1)
 
 
+def gather_and_right_shift_values(
```
yuki-97: `gather_and_right_shift_values` + `right_shift_values` duplicate the CP allgather → right-shift pattern already open-coded in the megatron value worker at two sites: `_value_loss_prepare_fn` (`megatron_value_worker.py:207-209`, train) and `collection_fn` (`megatron_value_worker.py:721-726`, `get_values` non-packed path).

**Suggest fix:** move both helpers to a shared module (e.g. `nemo_rl/models/value/utils.py`) and import from both backends. Megatron's extra `values = values[:, : data["returns"].shape[1]]` truncate can stay at the call site.

yuki-97 (commented):
thanks @fujial-code for supporting this, nice work!

left some comments. besides, could you also help to add two nightly tests for sequence packing and CP?

yuki-97:
@bg51717 could you help to take a review as well?

nemo_rl/models/automodel/setup.py
```diff
@@ -294,6 +294,11 @@ def validate_and_prepare_config(
     max_grad_norm = config["max_grad_norm"]
     enable_seq_packing = config["sequence_packing"]["enabled"]
     model_name = config["model_name"]
+    reward_model_cfg = config.get("reward_model_cfg", {})
```
bg51717, line 302: `validate_and_prepare_config()` is also called by `DTensorValueWorkerV2` with a `ValueConfig`, but its annotation and docstring only mention `PolicyConfig`. Could we update them to cover both config types? 
It would also be helpful to clarify that `reward_model_cfg` is used here to select the existing token-level regression model initialization path  for the DTensor value model, rather than indicating that the caller is a standalone reward model.

nemo_rl/models/automodel/setup.py
```diff
@@ -707,12 +723,23 @@ def setup_model_and_optimizer(
...
+            # its output and logsumexp head dimensions diverge. Force ring-flash
+            # for DTensor value models, which use regression reward model setup.
+            sdpa_method = [SDPBackend.FLASH_ATTENTION]
+        else:
+            # Preserve policy CP support for float32 through ring-efficient
+            # attention while preferring ring-flash for lower precision inputs.
+            sdpa_method = [
+                SDPBackend.FLASH_ATTENTION,
+                SDPBackend.EFFICIENT_ATTENTION,
+            ]
```
bg51717, line 752: Since the policy and value models have the same backbone architecture, could you clarify why they need different attention kernel selections for context parallelism here?

bg51717 (commented):
LGTM. I've left some comments.

nemo_rl/algorithms/ppo.py
fujial-code: Done — I removed the stale comment in b9582e99. DTensor value sequence packing is supported now, and the CP + sequence-packing incompatibility is validated explicitly during PPO setup.

nemo_rl/models/automodel/setup.py
fujial-code: Done — I restored the comments and expanded them in 93644b68 to explain the shared policy/value-model setup and the token-level regression model path.

nemo_rl/models/automodel/train.py
fujial-code: Done in b9582e99 — I moved the FSDP DP/CP scaling explanation directly above `loss = result * dp_size * backward_cp_scale` and kept the ValueLossPostProcessor-specific explanation separate.

nemo_rl/models/automodel/train.py
fujial-code: Confirmed that we should keep `flash_attn_kwargs` for reward models. The non-packed `run_rm.py`/`rm_env` path receives no FlashAttention arguments, while packed token-level regression reward/value models require them. I preserved the arguments and added `test_forward_reward_model_allows_flash_attn` to cover this behavior.

nemo_rl/models/automodel/train.py
fujial-code: Done in b9582e99 — this now uses `isinstance(post_processing_fn, ValueLossPostProcessor)`, and `cp_loss_is_replicated` has been removed. I also updated the comment to explain that ValueLossPostProcessor computes a replicated full-sequence loss and that AllGatherCPTensor.backward already performs the CP gradient sum.

A focused regression test verifies the scaling: with DP=3 and CP=2, the policy loss gets a scale of 6 while the value loss gets a scale of 3.

nemo_rl/models/automodel/train.py
fujial-code: Done in b9582e99 — I moved the DTensor CP + sequence-packing incompatibility check into PPO setup so the configuration fails before worker/model initialization.

nemo_rl/models/value/lm_value.py
fujial-code: You're right — I confirmed that `output_is_replicated` already causes the worker-group layer to retain only the representative rank output. The additional filtering in `lm_value.py` was therefore a no-op, so I removed those changes in b9582e99.

nemo_rl/models/value/workers/dtensor_value_worker_v2.py
fujial-code: Done in a4064268 — I moved `right_shift_values` and `gather_and_right_shift_values` into `nemo_rl/models/value/utils.py` and reused them from both the DTensor and Megatron value workers. Megatron's returns-length truncation remains at its call site.

nemo_rl/models/automodel/setup.py
fujial-code: Clarified in 93644b68. The difference is not caused by the backbone architecture; it comes from the token-level regression path used by the value model.

fujial-code:
@yuki-97 Added the two requested nightly tests in 482380f1:

- DTensor PPO value sequence packing with TP2+SP
- DTensor PPO value context parallelism with TP2+SP+CP2

Both recipes, launcher scripts, and nightly suite entries are included. I validated the YAML and shell dry runs and also ran the focused 2-GPU worker suite: all 6 CP and sequence-packing training/equivalence tests passed .

PR #3321, opened 2026-07-23
fix(ppo): enable DTensor value dynamic batching with equivalence coverage

### LINKED ISSUE #2953: PPO DTensor value-head initialization is nondeterministic across Ray worker launches

fujial-code:
## What does this PR do?
Enables dynamic batching for the DTensor/FSDP2 PPO value-model path and makes fresh Hugging Face value-head initialization deterministic across Ray value-worker launches.

Dynamic batching was temporarily disabled after #2953 reported a substantially higher critic loss at the beginning of training. The investigation found two separate concerns:

1. The dynamic-batching OFF and ON comparisons initially used different randomly initialized regression heads. With pinned weights, the value training signal is equivalent.
2. PPO seeds the controller process, but DTensor value models are constructed in separate Ray processes. A missing HF regression head was therefore initialized nondeterministically unless the seed was passed to and applied inside each value worker before model construction.

## Issues

Closes [#2953](https://github.com/NVIDIA-NeMo/RL/issues/2953)

## Summary of Changes

- Remove the PPO setup assertion that rejected `value.dynamic_batching.enabled=true` with the DTensor value backend.
- Add `value.seed`, inherited from `${ppo.seed}` in the PPO exemplar config.
- Seed each DTensor value worker after distributed setup and before model construction, so a missing HF regression head is initialized reproducibly.
- Add pinned-weight train-forward equivalence coverage for:
  - TP=1 dynamic batching.
  - TP=2 sequence parallelism plus dynamic batching.
  - Variable-length samples that force dynamic reordering and repacking.
- Add a fresh-initialization regression test that compares dynamic batching OFF and ON with TP=2 plus sequence parallelism without sharing a checkpoint.
- Rename the Automodel PPO recipe and test driver to include `-dynbatch`, and update the PPO guide and nightly manifest.

The Megatron value-model path is unchanged.

## What the equivalence tests cover

The pinned-weight test compares the deterministic value training signal across dynamic batching OFF and ON:

- dynamic microbatch construction
- variable-length sample reordering
- value-model forward pass
- temporal value right shift
- PPO MSE value loss
- output-order restoration
- global valid-token normalization
- metric aggregation

It compares aggregated `loss`, `values_mean`, `returns_mean`, and `residual_sq_mean`. Metrics are reduced to microbatch-order-independent scalars because dynamic batching can change the number and ordering of microbatches. The test uses `eval_mode=True` so dropout masks do not confound the execution-path comparison.

The separate fresh-initialization test deliberately does not share weights. Matching valid-token values confirm that the PPO seed reaches both Ray value workers before the regression head is constructed.

# Validation

Ran a 100-step pinned PPO comparison with:

- Model: Qwen2.5-1.5B-Instruct
- Dataset: GSM8K
- Value topology: TP=2 with sequence parallelism
- Dynamic batching: OFF vs ON
- Both runs resumed from the same checkpoint
- Matched training configuration

Results:

- `train/critic/loss` follows the same decay and convergence behavior.
- `train/reward` follows the same learning trajectory.
- Both runs reach approximately `0.90` reward by step 100.
- No persistent dynamic-batching-specific critic-loss divergence is observed.

## Critic Loss

image

## Reward

image

These results support removing the temporary DTensor value dynamic-batching guard.


nemo_rl/algorithms/ppo.py
```diff
@@ -296,10 +296,6 @@ def setup(
             "Context parallelism (CP>1) is currently not supported for the DTensor PPO value model. "
             "See https://github.com/NVIDIA-NeMo/RL/issues/2951."
         )
-        assert value_config["dynamic_batching"]["enabled"] is False, (
```
yuki-97, line 334: let's update `ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp` to `ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch` like the megatron one. also just a reminder that don't forget to update things like `checkpoint_dir` in yaml as well.

yuki-97 (commented):
@fujial-code thanks for verifying and adding the tests.

besides one inline comment, according to our discussion offline, dynamic batching for dtensor PPO is fine (curve in the PR description), but the critic/loss in #2953 may be related to the random init weights of value model.

1. could you update the PR description adding how you tested it to get the correct result? (I remember you saved a ckpt for test)
2. file a new issue (or change the title & content of #2953) to track this, since the mismatch is not due to dynamic batching.
3. for changing the threshold in nightly test, let's do it after we figure out the init weights things.

then I think there's no other block for this PR.

nemo_rl/algorithms/ppo.py
fujial-code: Done — I renamed the nightly recipe to
`ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch`
and updated the YAML `checkpoint_dir`, `log_dir`, W&B run name, nightly entry, and PPO documentation in 1a9f496.

tests/test_suites/nightly.txt
```diff
@@ -276,8 +276,8 @@ tests/test_suites/llm/distillation-xtoken-off-policy-multiteacher-qwen3-4b-llama
 # PPO #
 #######
 
-# DTensor PPO with value sequence parallelism (TP2+SP) (Qwen2.5-1.5B, GSM8K)
-tests/test_suites/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp.sh
+# DTensor PPO with value sequence parallelism (TP2+SP) and dynamic batching (Qwen2.5-1.5B, GSM8K)
+tests/test_suites/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch.sh
```
yuki-97: This now advertises dynamic batching, but the recipe never enables it: `examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch.yaml:33-38` has no `dynamic_batching` key, so it inherits `enabled: false` from `ppo_math_1B.yaml:407` and the nightly still runs the pre-rename DB-off config.

That leaves the path this PR unblocks with no e2e coverage in any tier -- `tests/functional/ppo_automodel.sh` doesn't set it either. Both Megatron `-dynbatch` siblings do (`ppo-qwen2.5-1.5b-gsm8k-1n8g-megatron-valuetp2sp-dynbatch.yaml:56-57`).

**Suggest fix:** add `dynamic_batching: {enabled: true}` under `value:` in the recipe.

tests/unit/models/value/test_dtensor_value_worker.py
```diff
@@ -500,6 +502,194 @@ def test_value_worker_parallelism_equivalence(
...
+
+    ``get_values`` coverage alone cannot detect a bug in the value-loss path.
+    This test pins one value-head initialization, reloads it into feature-OFF
+    and feature-ON workers, and compares the deterministic train forward,
+    right-shift, loss, reorder, and normalization chain on variable-length
+    samples. ``eval_mode=True`` disables dropout so different microbatch orders
+    do not confound the equivalence assertion.
+    """
+    feature_id = "-".join(key for key, enabled in feature_updates.items() if enabled)
+    equiv_gpus = max(tp, 1)
```
yuki-97: `max(tp, 1)` is just `tp` for both parametrizations, so the cluster is 1 GPU at tp=1 and 2 at tp=2 -- `dp_size == 1` in both cases (`lm_value.py:139-145` infers DP as the `-1` axis). Dynamic batching's cross-DP striding (`batched_data_dict.py:571-574`) and cross-shard max-seqlen agreement (`:786-796`) are both inert at `shards == 1`.

The sibling `test_value_worker_parallelism_equivalence` hardcodes `bundle_ct_per_node_list=[2]` (`:441`), so its tp=1 dynamic-batching case does run at dp=2 -- this new train-path test ends up strictly narrower than the `get_values` test it mirrors.

**Suggest fix:** use a fixed 2-GPU budget like the sibling, which gives dp=2 for the tp=1 case.

nemo_rl/models/value/config.py
```diff
@@ -41,6 +41,7 @@ class ValueConfig(TypedDict):
 
     model_name: str
     tokenizer: TokenizerConfig
+    seed: int
```
yuki-97: `seed` is required on the shared `ValueConfig`, but only the DTensor worker reads it (`dtensor_value_worker_v2.py:211`). `MegatronValueWorker` builds a fresh `LinearForLastLayer` head seeded from Megatron-Bridge's default `RNGConfig.seed = 1234` -- `make_policy_like_config` (`megatron/setup.py:2055-2107`) drops `seed`, and `_create_megatron_config` never passes `rng=`. So `value.seed: ${ppo.seed}`, which this PR also adds to the Megatron reference config, is inert on that backend.

Worth either honouring it on the Megatron path or documenting the key as DTensor-only. One caveat if you wire it through: `megatron_cfg.rng.seed = config["seed"]` would newly reject `ppo.seed: 0`, since `initialize.py:849` asserts `seed_ > 0` while the DTensor path accepts 0.

tests/unit/models/value/test_dtensor_value_worker.py
```diff
@@ -500,6 +502,194 @@ def test_value_worker_parallelism_equivalence(
...
+    weights_path = os.path.join(str(tmp_path), "value", "weights")
+    loss_fn = MseValueLossFn(MseValueLossConfig(scale=1.0, cliprange=0.5))
+
+    def _base_config() -> ValueConfig:
+        return _create_value_test_config(model_name=tiny_qwen2_model_path, tp=tp)
+
+    torch.manual_seed(42)
+    batch, max_seq_len = 8, 64
```
yuki-97: With `max_seq_len = 64` and `sequence_length_round: 64`, every length in [32, 64] rounds to 64 and then `min(padded, unpadded) = 64` (`batched_data_dict.py:807-813`), so `truncate_tensors` is a full-width no-op on every microbatch and the re-pad branch at `dtensor_value_worker_v2.py:517-522` never fires.

Only the grouping half of dynamic batching is exercised (2 MBs of 4 vs 4 MBs of 2); sequence shapes never vary, so the "repacking" the PR description claims isn't actually covered.

**Suggest fix:** `batch, max_seq_len = 8, 128` with lengths drawn from `max_seq_len // 4` gives lengths rounding to {64, 128}, still inside `train_mb_tokens=256`. (`sequence_length_round` itself is pre-existing, so this keeps the change on the new test.)

tests/unit/models/value/test_dtensor_value_worker.py
```diff
@@ -500,6 +502,194 @@ def test_value_worker_parallelism_equivalence(
...
+)
+def test_value_worker_dynamic_batching_train_forward_equivalence(
+    tiny_qwen2_model_path, tmp_path, tp, feature_updates
+):
+    """Dynamic batching must preserve the pinned train-forward value signal.
+
+    ``get_values`` coverage alone cannot detect a bug in the value-loss path.
+    This test pins one value-head initialization, reloads it into feature-OFF
+    and feature-ON workers, and compares the deterministic train forward,
+    right-shift, loss, reorder, and normalization chain on variable-length
```
yuki-97: The train path never restores output order: `lm_value.py:339` discards the sort indices with `sharded_data, _ = ...`, and `reorder_data` runs only on the `get_values` path (`lm_value.py:302`). All four metrics compared here are `global_valid_toks`-normalized sums, so an ordering bug is structurally unobservable by this test.

The word "reorder" is defensible for the pre-shard sort, but the PR description's "output-order restoration" bullet under what the pinned-weight test compares doesn't hold -- worth moving that claim to the fresh-init test, where `reorder_data` actually runs and per-token values are compared elementwise.

tests/test_suites/nightly.txt
yuki-97: This now advertises dynamic batching, but the recipe never enables it: `examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch.yaml:33-38` has no `dynamic_batching` key, so it inherits `enabled: false` from `ppo_math_1B.yaml:407` and the nightly still runs the pre-rename DB-off config.

That leaves the path this PR unblocks with no e2e coverage in any tier -- `tests/functional/ppo_automodel.sh` doesn't set it either. Both Megatron `-dynbatch` siblings do (`ppo-qwen2.5-1.5b-gsm8k-1n8g-megatron-valuetp2sp-dynbatch.yaml:56-57`).

**Suggest fix:** add `dynamic_batching: {enabled: true}` under `value:` in the recipe.

nemo_rl/models/value/config.py
yuki-97: `seed` is required on the shared `ValueConfig`, but only the DTensor worker reads it (`dtensor_value_worker_v2.py:211`). `MegatronValueWorker` builds a fresh `LinearForLastLayer` head seeded from Megatron-Bridge's default `RNGConfig.seed = 1234` -- `make_policy_like_config` (`megatron/setup.py:2055-2107`) drops `seed`, and `_create_megatron_config` never passes `rng=`. So `value.seed: ${ppo.seed}`, which this PR also adds to the Megatron reference config, is inert on that backend.

Worth either honouring it on the Megatron path or documenting the key as DTensor-only. One caveat if you wire it through: `megatron_cfg.rng.seed = config["seed"]` would newly reject `ppo.seed: 0`, since `initialize.py:849` asserts `seed_ > 0` while the DTensor path accepts 0.

tests/unit/models/value/test_dtensor_value_worker.py
yuki-97: With `max_seq_len = 64` and `sequence_length_round: 64`, every length in [32, 64] rounds to 64 and then `min(padded, unpadded) = 64` (`batched_data_dict.py:807-813`), so `truncate_tensors` is a full-width no-op on every microbatch and the re-pad branch at `dtensor_value_worker_v2.py:517-522` never fires.

Only the grouping half of dynamic batching is exercised (2 MBs of 4 vs 4 MBs of 2); sequence shapes never vary, so the "repacking" the PR description claims isn't actually covered.

**Suggest fix:** `batch, max_seq_len = 8, 128` with lengths drawn from `max_seq_len // 4` gives lengths rounding to {64, 128}, still inside `train_mb_tokens=256`. (`sequence_length_round` itself is pre-existing, so this keeps the change on the new test.)

tests/unit/models/value/test_dtensor_value_worker.py
yuki-97: The train path never restores output order: `lm_value.py:339` discards the sort indices with `sharded_data, _ = ...`, and `reorder_data` runs only on the `get_values` path (`lm_value.py:302`). All four metrics compared here are `global_valid_toks`-normalized sums, so an ordering bug is structurally unobservable by this test.

The word "reorder" is defensible for the pre-shard sort, but the PR description's "output-order restoration" bullet under what the pinned-weight test compares doesn't hold -- worth moving that claim to the fresh-init test, where `reorder_data` actually runs and per-token values are compared elementwise.

tests/test_suites/nightly.txt
yuki-97, line 280: This now advertises dynamic batching, but the recipe never enables it: [`ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch.yaml:33-38`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch.yaml#L33-L38) has no `dynamic_batching` key, so it inherits `enabled: false` from [`ppo_math_1B.yaml:407`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/examples/configs/ppo_math_1B.yaml#L407) and the nightly still runs the pre-rename DB-off config.

That leaves the path this PR unblocks with no e2e coverage in any tier — [`tests/functional/ppo_automodel.sh`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/tests/functional/ppo_automodel.sh) doesn't set it either. Both Megatron `-dynbatch` siblings do ([`ppo-qwen2.5-1.5b-gsm8k-1n8g-megatron-valuetp2sp-dynbatch.yaml:56-57`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-megatron-valuetp2sp-dynbatch.yaml#L56-L57)).

**Suggest fix:** add `dynamic_batching: {enabled: true}` under `value:` in the recipe.

tests/unit/models/value/test_dtensor_value_worker.py
yuki-97, line 611: `max(tp, 1)` is just `tp` for both parametrizations, so the cluster is 1 GPU at tp=1 and 2 at tp=2 — `dp_size == 1` in both cases ([`lm_value.py:139-145`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/lm_value.py#L139-L145) infers DP as the `-1` axis). Dynamic batching's cross-DP striding ([`batched_data_dict.py:571-574`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/distributed/batched_data_dict.py#L571-L574)) and cross-shard max-seqlen agreement ([`:786-796`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/distributed/batched_data_dict.py#L786-L796)) are both inert at `shards == 1`.

The sibling `test_value_worker_parallelism_equivalence` hardcodes `bundle_ct_per_node_list=[2]` ([`:441`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/tests/unit/models/value/test_dtensor_value_worker.py#L441)), so its tp=1 dynamic-batching case does run at dp=2 — this new train-path test ends up strictly narrower than the `get_values` test it mirrors.

**Suggest fix:** use a fixed 2-GPU budget like the sibling. `equiv_gpus` feeds both `bundle_ct_per_node_list` and `num_gpus_per_node`, so one line covers it. This widens the tp=1 case to dp=2; tp=2 stays at dp=1 since dp=2 there would need 4 GPUs, which is the same trade the sibling makes.

```suggestion
    equiv_gpus = 2
```

nemo_rl/models/value/config.py
yuki-97, line 44: `seed` is required on the shared `ValueConfig`, but only the DTensor worker reads it ([`dtensor_value_worker_v2.py:211`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/workers/dtensor_value_worker_v2.py#L211)). `MegatronValueWorker` builds a fresh `LinearForLastLayer` head seeded from Megatron-Bridge's default [`RNGConfig.seed = 1234`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/training/config/common_config.py#L10) — [`make_policy_like_config`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/megatron/setup.py#L2055-L2107) drops `seed`, and [`_create_megatron_config`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/megatron/setup.py#L1348) never passes `rng=`. So `value.seed: ${ppo.seed}`, which this PR also adds to the Megatron reference config, is inert on that backend.

**Suggest fix:** don't make it a config key at all — pass the seed in from the driver. There is exactly one source of truth (`ppo.seed`); `value.seed: ${ppo.seed}` clones it into a second place that can drift, and on the Megatron backend the clone does nothing. `Value.__init__` already takes this class of runtime parameter as a plain kwarg — `init_optimizer`, `weights_path`, `optimizer_path` ([`lm_value.py:44-53`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/lm_value.py#L44-L53)). Concretely:

1. Drop `seed: int` from `ValueConfig`, and `value.seed: ${ppo.seed}` from both [`examples/configs/ppo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/examples/configs/ppo_math_1B.yaml) and [`tests/unit/reference_configs/ppo_math_1B_megatron.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/tests/unit/reference_configs/ppo_math_1B_megatron.yaml).
2. Add `seed: int` (no default) to `Value.__init__` and forward it through the existing [`RayWorkerBuilder`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/lm_value.py#L157-L165) kwargs, next to `init_optimizer` / `weights_path`.
3. Add `seed: int` (no default) to `DTensorValueWorkerV2Impl.__init__` — it already has `**kwargs` ([`:136-143`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/workers/dtensor_value_worker_v2.py#L136-L143)) — and change `set_seed(config["seed"])` to `set_seed(seed)`. `MegatronValueWorker` keeps swallowing it via `**kwargs`.
4. Pass it at the one construction site, [`ppo.py:712`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/algorithms/ppo.py#L712): `seed=ppo_config["seed"]`, the same value already given to `set_seed` at [`ppo.py:336`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/algorithms/ppo.py#L336).
5. In the two value unit tests, replace the new `"seed": 42` config entry with `Value(..., seed=42)`.

One rebase note: on main the single-controller critic (`_build_value` -> `TQValue`, `nemo_rl/algorithms/single_controller_utils/setup.py`) gets the seed for free today because it forwards `master_config.value` wholesale. With the kwarg it needs `seed=algo_cfg.seed` passed explicitly — the same value `set_seed` already takes in that file. `TQValue.__init__` forwards `**kwargs` to `Value`, so nothing else there changes.

With no default anywhere, a missed call site is a `TypeError` rather than a silently wrong seed — and Megatron ignoring the kwarg becomes an ordinary Python fact instead of a YAML key that looks configurable but isn't.

tests/unit/models/value/test_dtensor_value_worker.py
yuki-97, line 619: With `max_seq_len = 64` and `sequence_length_round: 64` ([`_apply_config_updates`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/tests/unit/models/value/test_dtensor_value_worker.py#L237)), every length in [32, 64] rounds to 64 and `min(padded_seqlen, unpadded_seqlen)` hands 64 back ([`batched_data_dict.py:804-814`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/distributed/batched_data_dict.py#L804-L814)), so the `truncate_tensors` call on the dynamic-shapes train iterator ([`batched_data_dict.py:979`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/distributed/batched_data_dict.py#L979)) is a full-width no-op on every microbatch.

Only the grouping half of dynamic batching is exercised (2 MBs of 4 vs 4 MBs of 2); microbatch widths never differ, so the "repacking" the PR description claims isn't actually covered.

**Suggest fix:** raise `max_seq_len` to `max_total_sequence_length` (128). Lengths then land in [64, 128] and round to {64, 128}, so widths do vary between microbatches; `train_mb_tokens` is 128 * 2 = 256, so both a 2x128 and a 4x64 microbatch still fit. (`sequence_length_round` itself is pre-existing, so this keeps the change on the new test.)

```suggestion
    batch, max_seq_len = 8, 128
```

tests/unit/models/value/test_dtensor_value_worker.py
yuki-97, line 606: The train path never restores output order: [`lm_value.py:339`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/lm_value.py#L339) discards the sort indices with `sharded_data, _ = ...`, and `reorder_data` runs only on the `get_values` path ([`lm_value.py:302`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/lm_value.py#L302)). All four metrics compared here are `global_valid_toks`-normalized sums, so an ordering bug is structurally unobservable by this test.

The word "reorder" is defensible for the pre-shard sort, but nothing here restores an order, so both this docstring line and the PR description's "output-order restoration" bullet overstate what is covered — that claim belongs on the fresh-init test, where `reorder_data` actually runs and per-token values are compared elementwise.

```suggestion
    right-shift, loss, and normalization chain on variable-length
```

docs/guides/ppo.md
```diff
@@ -90,7 +90,7 @@ value:
       data_parallel_sharding_strategy: "optim_grads_params"
 ```
 
-For a DTensor PPO recipe, see [ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp.yaml](../../examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp.yaml).
+For a DTensor PPO recipe, see [ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch.yaml](../../examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-automodel-valuetp2sp-dynbatch.yaml).
```
yuki-97, line 93: for these three lines, could you help to revert back and check if the tests could pass?
1. `train/critic/loss` to `< 6.0`
2. `train/critic/grad_norm` to `< 350`

---

1. https://github.com/NVIDIA-NeMo/RL/blob/c0165c8c2753a99bbc6969fee0ea2c1a183cbe50/tests/functional/ppo_automodel.sh#L60
2. https://github.com/NVIDIA-NeMo/RL/blob/c0165c8c2753a99bbc6969fee0ea2c1a183cbe50/tests/functional/ppo_automodel.sh#L63
3. https://github.com/NVIDIA-NeMo/RL/blob/c0165c8c2753a99bbc6969fee0ea2c1a183cbe50/tests/functional/ppo_non_colocated.sh#L60

yuki-97 (commented):
hi @fujial-code , thanks so much for investigating and found it's due to the seed! sorry for so late reply, left some comments.

**Scope:** the DTensor value dynamic-batching enablement — the removed PPO setup assertion, the new `value.seed` plumbing, the two equivalence tests, the recipe rename + nightly manifest, and the raised functional-test thresholds.

**Checked:**

- Recipe rename vs. what the recipe actually sets — traced the `defaults` chain for `value.dynamic_batching`.
- The new train-forward equivalence test vs. what its docstring and the PR description claim it covers — walked the dynamic-batching path (cross-DP striding, `sequence_length_round` rounding, `reorder_data`) at the cluster size and shapes it sets up.
- `value.seed` on both backends — followed the key through `make_policy_like_config` / `_create_megatron_config` into Megatron-Bridge's `rng`.
- The raised `train/critic/*` thresholds in the functional scripts, now that the init-weights cause is understood.
- Convergence evidence for closing #2953 — the DB-on/off comparison in the description is what was needed, thanks for pasting it.

**Remaining comments** cover one substantive item — the nightly still runs a DB-off config despite the `-dynbatch` rename, leaving this PR's path with no e2e coverage in any tier — plus test-coverage, config-shape and threshold follow-ups. The three test comments ship one-line ```suggestion blocks.

### On the seeding change

Recording the why here, since it took some digging and isn't obvious from the diff.

The base causal-LM checkpoint has no regression `score` head, and the value worker explicitly skips it when loading the base ckpt ([`dtensor_value_worker_v2.py:225`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/workers/dtensor_value_worker_v2.py#L225), `skip_task_head_prefixes_for_base_model: ["score."]`), so each worker initializes that head locally. PPO seeds only the controller process ([`ppo.py:336`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/algorithms/ppo.py#L336)), and value workers are separate Ray actors — so before this PR, two separate launches got two different `score` heads. That, not dynamic batching, is what #2953 was actually observing.

MCore never showed the symptom because `initialize_megatron` already seeds inside the worker before model construction ([`initialize.py:206-211`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/training/initialize.py#L206-L211) -> `_set_random_seed`). But it seeds from `ConfigContainer.rng`, which NeMo-RL never populates, so it is always the upstream default [`RNGConfig.seed = 1234`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/training/config/common_config.py#L10). Reproducible, but pinned — changing `ppo.seed` has no effect on the MCore value head. That is what the [`config.py:44`](https://github.com/NVIDIA-NeMo/RL/blob/10b73a0227bb8fa2a6e10c45681686cda5ed4939/nemo_rl/models/value/config.py#L44) comment is about: the new required `seed` key is inert on the Megatron backend, even though this PR also adds `value.seed` to the Megatron reference config.

(Cross-backend the two heads differ regardless — Automodel goes through HF `_init_weights`, MCore through `LinearForLastLayer`'s kaiming-uniform. A shared seed only buys within-backend reproducibility across launches.)

PR #3280, opened 2026-07-18
build: bump vllm 0.20.0 → 0.25.1

terrykong:
# What does this PR do ?

Bumps vLLM from 0.20.0 to 0.25.1 (supersedes the 0.24.0 draft #3137). Rebased onto current main (bc382dc4); all validation below is against that base.

## Changes

_Bullets are prefixed with the commit that made the change, as of head 69a3e5aa8._

**Dependencies (`pyproject.toml` / `uv.lock`)**
- a34945a3b — vLLM wheel URLs → `v0.25.1` (`manylinux_2_28`, changed from `manylinux_2_35` in 0.24+)
- a34945a3b — `flashinfer-python/-cubin/-jit-cache` → `0.6.13` in the `vllm` extra (pinned by vLLM 0.25.1's `requirements/cuda.txt`)
- a34945a3b — xgrammar override relaxed `==0.1.33` → `>=0.2.1,<1.0.0` (vLLM 0.25 imports `xgrammar.normalize_tool_choice`, added in 0.2.x; resolver picks 0.2.3)
- a34945a3b — openai stays at 2.6.1 (nemo-gym pins `openai<=2.7.2` and its child server venvs must exactly match the parent's openai version, so a global override is not viable). vLLM 0.25.1's import of `openai.types.responses.NamespaceTool` (added in openai 2.25.0) is handled by a new source patch instead — see below.

**Async OpenAI-compatible server (`vllm_worker_async.py`)**
- a34945a3b — vLLM 0.25 moved chat preprocessing from `OpenAIServingRender` (module deleted) to `OnlineRenderer` (`vllm/renderers/`); both `/v1/chat/completions` and `/tokenize` still route through `preprocess_chat`, so the NeMo-RL prefix-token override moves to an `OnlineRenderer` subclass
- a34945a3b — `OpenAIServingTokenization` → `ServingTokenization` (takes `online_renderer`, no `engine_client`)
- a34945a3b — `preprocess_chat`'s `tool_parser`/`reasoning_parser` args were folded into a single `parser` arg upstream
- a34945a3b — Max-context-length log filter now targets `vllm.entrypoints.openai.chat_completion.serving` (module logger name)

**fp8 generation (`quantization/fp8.py`)**
- a34945a3b — `FusedMoE` is a factory function in vLLM 0.25 (returns `MoERunner` delegating to a weight-owning `RoutedExperts` submodule) — fp8 weight classification now detects `MoERunner`/`RoutedExperts`
[...]

PR #3323, opened 2026-07-23
fix(eagle3): isolate draft grad clipping

yuekaizhang:
Two fixes for eagle3 draft co-training.

### 1. Spec decode crashes at the max_model_len boundary
The drafter looks `num_speculative_tokens` ahead, so when a request reaches
`max_model_len` it reads/writes KV slots past the engine buffers → CUDA illegal
memory access. Fix: clamp per-request `max_tokens` to leave
`num_speculative_tokens + 1` headroom, so generation stops before the drafter
crosses the boundary. No-op when speculative decoding is off.

### 2. Draft gradients shrink the policy update
With draft co-training, the freshly-initialized draft head dominates the shared
gradient norm (2–5× the policy-only norm), so the global `max_grad_norm` clip
rescales the policy gradient much smaller than the no-draft baseline — making
the curves incomparable. Fix: clip the draft head in its own grad-norm group so
policy and draft are clipped independently. No-draft runs are unchanged;
`train/grad_norm` is now policy-only and `train/draft_grad_norm` is logged.

Green Curve: baseline without eagle3
Red Curve: ealge3 + RL with the PR's fix
Yello Curve: eagle3 + RL without the PR's fix
image


yuekaizhang:
@isomap @yuki-97  Could you help review the PR? Thanks!

nemo_rl/algorithms/loss/loss_functions.py
```diff
@@ -92,7 +92,11 @@ def __call__(
                 False,
             )
         else:
-            teacher_probs = torch.nn.functional.softmax(teacher_logits, dim=-1)
+            # Match DistributedCrossEntropy semantics: backward propagates only
+            # through student logits, never into the (policy) teacher.
+            teacher_probs = torch.nn.functional.softmax(
+                teacher_logits.detach(), dim=-1
```
isomap: Nit: `teacher_logits` is already derived from `logits.detach()` in `nemo_rl/algorithms/loss/utils.py`  before reaching this loss path, so the additional `.detach()` here is redundant. It is harmless, but could we remove it to keep the ownership of this gradient boundary in one place?'

https://github.com/NVIDIA-NeMo/RL/blob/2d3bb79623e9486172b441b6257103005fcd61f2/nemo_rl/algorithms/loss/utils.py#L173-L178

yuekaizhang: Good catch! You're right — `teacher_logits` is already derived from `logits.detach()` in `utils.py`, so the extra `.detach()` here was redundant. Removed it and updated the comment to point to the upstream ownership. Fixed in 5b82d1f.

yuekaizhang:
/ok to test [b9dfa51](https://github.com/NVIDIA-NeMo/RL/pull/3323/commits/b9dfa51c0cedacd236faef85cd8b0cc228640d48)

yuekaizhang:
/ok to test [a8317ca](https://github.com/NVIDIA-NeMo/RL/pull/3323/commits/a8317ca4682df648a19b3116da4f94ebd71ad17c)

nemo_rl/models/megatron/draft/utils.py
```diff
@@ -1203,6 +1203,28 @@ def copy_policy_lm_head_to_draft(
         )
 
 
+DRAFT_GRAD_NORM_GROUP = "draft"
+
+
+def register_draft_grad_norm_group() -> None:
```
yuki-97, line 1209: **[TEST]** No unit guard for the `SEPARATE_GRAD_NORM_GROUPS` monkey-patch.

This helper relies on an upstream mcore module-level tuple that could get renamed / re-typed; the eagle3 nightly won't catch that until a night later. Suggest adding a tiny unit test that verifies idempotency + preservation of the pre-registered `'mtp'` entry.

```python
# tests/unit/models/megatron/test_draft_grad_norm_group.py
import pytest


@pytest.mark.mcore
def test_register_draft_grad_norm_group_is_idempotent_and_preserves_existing():
    from megatron.core.optimizer import optimizer as mcore_opt
    from nemo_rl.models.megatron.draft.utils import (
        DRAFT_GRAD_NORM_GROUP,
        register_draft_grad_norm_group,
    )

    original = mcore_opt.SEPARATE_GRAD_NORM_GROUPS
    try:
        register_draft_grad_norm_group()
        after_first = mcore_opt.SEPARATE_GRAD_NORM_GROUPS
        assert DRAFT_GRAD_NORM_GROUP in after_first
        assert "mtp" in after_first  # not overwritten
        register_draft_grad_norm_group()
        assert mcore_opt.SEPARATE_GRAD_NORM_GROUPS == after_first  # no-op
    finally:
        mcore_opt.SEPARATE_GRAD_NORM_GROUPS = original
```

nemo_rl/models/generation/vllm/vllm_worker.py
```diff
@@ -807,6 +807,30 @@ def generate(
...
+            base_max_tokens = sampling_params.max_tokens
+            sampling_params = [
+                self._build_sampling_params(
+                    greedy=greedy,
+                    stop_strings=stop_strings,
+                    max_new_tokens=max(
+                        1,
+                        min(
+                            base_max_tokens,
+                            max_model_len - int(input_len) - (spec_lookahead + 1),
```
yuki-97: **[TEST]** The spec-decode clamp math is inlined into the list comprehension, so the `(spec_lookahead + 1)` reservation and the `max(1, ...)` floor have no unit guard — a future refactor could drop the `+1` or change the floor to `0` and only the eagle3 nightly would notice.

Suggest lifting the clamp into a small static method so the arithmetic is testable without spinning up a live vLLM:

```python
@staticmethod
def _spec_decode_max_tokens(base_max_tokens, input_len, max_model_len, spec_lookahead):
    return max(1, min(base_max_tokens, max_model_len - input_len - (spec_lookahead + 1)))
```

Then in `generate`:

```python
max_new_tokens=self._spec_decode_max_tokens(
    base_max_tokens, int(input_len), max_model_len, spec_lookahead,
),
```

And a parametrized guard (`tests/unit/models/generation/test_vllm_spec_decode_clamp.py`):

```python
import pytest
from nemo_rl.models.generation.vllm.vllm_worker import VllmGenerationWorkerImpl


@pytest.mark.parametrize(
    ("base_max_tokens", "input_len", "max_model_len", "spec_lookahead", "expected"),
    [
        (256,  100, 1024, 5, 256),   # clamp inactive: base wins
        (256,  900, 1024, 5, 118),   # clamp active: 1024 - 900 - 6
        (256, 1018, 1024, 5,   1),   # at boundary: floor at 1
        (256, 1050, 1024, 5,   1),   # past boundary: floor at 1
        (  8,  100, 1024, 5,   8),   # base < headroom: base wins
    ],
)
def test_spec_decode_max_tokens_clamp(
    base_max_tokens, input_len, max_model_len, spec_lookahead, expected,
):
    assert VllmGenerationWorkerImpl._spec_decode_max_tokens(
        base_max_tokens, input_len, max_model_len, spec_lookahead,
    ) == expected
```

nemo_rl/models/generation/vllm/vllm_worker.py
```diff
@@ -807,6 +807,30 @@ def generate(
             stop_strings=stop_strings,
         )
 
+        # vLLM 0.20 eagle3 spec decode hits a CUDA illegal memory access when a
+        # request's total length reaches max_model_len (the drafter looks ahead
+        # past the boundary). Clamp per-request max_tokens so speculative
+        # requests stop short of the boundary by the drafter lookahead.
+        spec_cfg = self.cfg.get("vllm_kwargs", {}).get("speculative_config") or {}
+        spec_lookahead = int(spec_cfg.get("num_speculative_tokens", 0))
+        if spec_lookahead > 0:
```
yuki-97, line 832: cc @terrykong , maybe need to check later whether we still need this after bump to 0.25 in  https://github.com/NVIDIA-NeMo/RL/pull/3280?

not a blocker for this PR.

yuki-97 (commented):
thanks for the fix, LGTM and let's add some tests to guard.

nemo_rl/models/megatron/draft/utils.py
yuekaizhang: Good suggestion! Added `test_draft_grad_norm_group.py` with a `@pytest.mark.mcore` test that verifies idempotency and that the pre-existing `'mtp'` entry is preserved. Fixed in bbdb04c.

nemo_rl/models/generation/vllm/vllm_worker.py
yuekaizhang: Good point! Extracted the clamp logic into `BaseVllmGenerationWorker._spec_decode_max_tokens` static method and added `test_vllm_spec_decode_clamp.py` with 5 parametrized cases (clamp inactive, clamp active, at-boundary, past-boundary, base wins). Fixed in bbdb04c.

yuekaizhang:
/ok to test [bbdb04c](https://github.com/NVIDIA-NeMo/RL/pull/3323/commits/bbdb04cc82f9de142bcc85c38da5d3ee5c4b12de)

yuki-97 (approved):

PR #2962, opened 2026-06-26
fix: sanitize non-finite logprobs in vllm async worker

yuchenwang3:
vLLM can emit non-finite logprobs. We hit a NaN consistently around 49K context, and vLLM also uses `-inf` intentionally when sampling filters remove tokens. Starlette's `JSONResponse` serializes with `allow_nan=False`, so any remaining NaN or infinity raises `ValueError` and turns the chat-completion request into a 500.

This PR sanitizes the dumped response immediately before JSON serialization:

- recursively replaces NaN, `+inf`, and `-inf` floats with `0.0`;
- emits at most one warning per response with the number of replaced values, so numerical instability is visible rather than silently masked;
- wraps the current `model_dump_chat_response_with_dynamic_message_fields` result on `main`.

`0.0` is a conservative sentinel for the rollout importance ratio because the train-time logprob is recomputed. The warning/count is intended to preserve the signal needed to investigate genuine model or vLLM numerical failures.

The unit test covers nested dict/list values, all three non-finite forms, finite and non-float passthrough, replacement count, one-warning behavior, and the no-warning finite path. Ruff lint/format, syntax compilation, and DCO pass on the rebased head.

Recent bounded retry and exhausted-rollout handling reduce the blast radius of a 500, but they do not make this response serializable; a deterministic NaN can still fail every retry. This keeps the fix at the source boundary.

PR #2918, opened 2026-06-24
fix: improve error message when NeMo Gym returns no generation data

yfw:
# What does this PR do ?

When NeMo Gym returns a rollout with no generation data, `_postprocess_nemo_gym_to_nemo_rl_result` raises a `ValueError` whose message includes the prompt length, computed via `tokenizer.apply_chat_template`. That call could itself throw, so instead of the intended "no generation data" error we surfaced an opaque `apply_chat_template` exception — hiding the real problem.

This change:

- Wraps the `apply_chat_template` call in `try/except` so a tokenizer failure no longer masks the original error; the prompt length is reported as `<unknown — apply_chat_template failed: ...>` instead.
- Adds a second likely cause to the message: all `response.output` items were reasoning/tool-call items with no assistant generation.
- Logs the `response.output` item types (and count) to help distinguish that case from the prompt-exceeds-`max_model_len` case.

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #2742, opened 2026-06-09
feat: upstreaming refit prefetching optimization

youngeunkwon0405:
# What does this PR do ?

Upstreaming prefetch optimization for the non-colocated refit ([ref](https://gitlab-master.nvidia.com/terryk/nemo-rl-internal/-/merge_requests/266/diffs#40342d3add605a36398701f980c6d4d7b99ecbc5)) from the ultra-v3 branch.

| Test case | Baseline | Patch | Δ vs baseline |
|-----------|-----------|--------|---------------|
| dsv3-64n4g | 6.28s (3225704, n=5) | 6.07s (3228643, n=7: 6.10, 6.02, 6.09, 6.05, 6.05, 6.11, 6.04) | −0.21s (−3.3%) |
| qwen3-235b-32n4g | 3.10s (3225718, n=3) | 3.07s (3228641, n=6: 3.10, 2.89, 3.45, 3.07, 2.94, 2.96) | −0.03s (−1%) |
| qwen3-30ba3b-4n4g | 3.68s (3225719, n=3) | 3.01s (3228642, n=5: 2.79, 2.88, 2.82, 3.44, 3.13) | −0.67s (−18%) |

Wandb log: https://wandb.ai/nvidia/nemorl-refit-benchmark?nw=nwuseryoungeunk&panelDisplayName=timing%2Ftrain%2Fweight_sync&panelSectionName=timing

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #3324, opened 2026-07-23
fix: harden multimodal Gym training lifecycle

JeffPengCoder:
# What does this PR do?

Hardens the multimodal NeMo Gym training path discovered while running the
`rohit/gymv-mm-integration` branch across separate generation and policy GPU
classes.

The changes:

- make Docker-prefetched Gym virtual environments self-contained instead of
  leaving links into uv's build cache;
- resolve a configured vLLM chat-template path before constructing the serving
  objects, matching the stock vLLM API server;
- initialize the existing Nemotron Omni RADIO LayerScale hook from the sync
  and async post-init paths, awaiting the async collective explicitly;
- fail loudly on non-finite vLLM response fields and report their exact JSON
  paths instead of silently changing logprob values;
- reject incomplete automatic HF-to-Megatron conversion directories as cache
  hits; and
- allow a common packed-refit buffer size across heterogeneous GPUs and wait
  for all private CUDA streams before returning.

Focused regression tests cover each behavior. The branch also updates one
integration-branch test call for the existing `nemo_gym_row` argument.

# Issues

No linked issue.

# Usage

For a refit collective whose ranks have different GPU memory capacities, set
the same explicit buffer size in every process:

```bash
export NRL_REFIT_BUFFER_SIZE_BYTES=536870912
```

Homogeneous deployments retain the existing
`NRL_REFIT_BUFFER_MEMORY_RATIO` fallback.

# Provenance and related work

This PR is based on
`rohit/gymv-mm-integration@71717873240c99fd1ede2db17480818205766848`.
The Gym/MM/Nemotron Omni implementation already present in that branch,
including the LayerScale hook introduced through #3290, is baseline work and
is not claimed here. This PR only changes when and how that existing hook is
invoked.

The following related changes are intentionally distinguished:

- #2918 already fixed empty-generation attribution on `main`; the duplicate
  local implementation was removed from this PR.
- #2962 proposes replacing non-finite values with `0.0`. This PR instead
  preserves response semantics by rejecting the response and identifying the
  exact offending fields.
- #2771 introduced Gym venv prefetching with symlink mode. This PR is a
  follow-up that makes the baked venv relocatable/self-contained.
- #2742 redesigns refit prefetching. This PR addresses correctness in the
  current packed-transfer implementation and does not include that redesign.

No OSWorld provider code, Gym submodule changes, experiment recipes, internal
registry references, host configuration, scheduler jobs, proxy configuration,
or deployment assets are included.

**Pre checks**:

- [x] Added focused regression tests.
- [x] Ran Ruff over all changed Python files.
- [x] Ran Python compilation and `git diff --check`.
- [x] Previously validated the candidate behaviors with focused tests in the
  project Linux container; GitHub CI is expected to run the submission branch.
- [x] No user-facing documentation change is required; the new environment
  variable is documented in code and in this PR.

# Additional Information

The commits are separated by concern and carry `Signed-off-by` trailers so the
follow-up work can be reviewed independently from the integration branch's
existing commits.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -295,7 +295,7 @@ if [[ -n "${NEMO_GYM_PREFETCH_CONFIGS:-}" ]]; then
         "${NEMO_GYM_VLLM_VERSION}" "${NEMO_GYM_VLLM_VERSION}" "${WHEEL_ARCH}" > "${GYM_VLLM_OVERRIDE}"
     UV_TORCH_BACKEND="${NEMO_GYM_CUDA}" \
     UV_OVERRIDE="${GYM_VLLM_OVERRIDE}" \
-    UV_LINK_MODE=symlink uv run python examples/nemo_gym/prefetch_venvs.py $NEMO_GYM_PREFETCH_CONFIGS
+    UV_LINK_MODE=copy uv run python examples/nemo_gym/prefetch_venvs.py $NEMO_GYM_PREFETCH_CONFIGS
     rm -f "${GYM_VLLM_OVERRIDE}"
 fi
 EOF
--- a/nemo_rl/models/generation/vllm/utils.py
+++ b/nemo_rl/models/generation/vllm/utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from collections import defaultdict
+import math
 from typing import Any, Optional
 
 import torch
@@ -272,6 +273,23 @@ def attach_routed_experts_to_chat_response_choices(
     return response
 
 
+def _find_non_finite_float_paths(value: Any, path: str = "$") -> list[str]:
+    """Return JSON-style paths for non-finite float values without mutating them."""
+    if isinstance(value, float):
+        return [] if math.isfinite(value) else [f"{path}={value!r}"]
+    if isinstance(value, dict):
+        paths = []
+        for key, item in value.items():
+            paths.extend(_find_non_finite_float_paths(item, f"{path}.{key}"))
+        return paths
+    if isinstance(value, (list, tuple)):
+        paths = []
+        for index, item in enumerate(value):
+            paths.extend(_find_non_finite_float_paths(item, f"{path}[{index}]"))
+        return paths
+    return []
+
+
 def model_dump_chat_response_with_routed_experts(response: Any) -> dict[str, Any]:
     """Dump a vLLM OpenAI chat response while preserving dynamic R3 fields."""
     response_dict = response.model_dump()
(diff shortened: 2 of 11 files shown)


nemo_rl/models/megatron/setup.py
```diff
@@ -495,9 +495,22 @@ def validate_model_paths(config: PolicyConfig) -> tuple[str, str, bool]:
         overrides_hash = _get_hf_config_overrides_hash(hf_config_overrides)
         hf_model_subdir = f"{hf_model_subdir}__hfovr_{overrides_hash}"
     pretrained_path = os.path.join(get_megatron_checkpoint_dir(), hf_model_subdir)
-    pt_checkpoint_exists = os.path.exists(pretrained_path) and os.path.exists(
-        os.path.join(pretrained_path, "iter_0000000")
+    # A previous conversion can leave the iteration directory and large distcp
```
rohitrango, line 498: is this due to a partial checkpoint save triggering a load error?

nemo_rl/models/generation/vllm/vllm_worker.py
```diff
@@ -646,6 +641,13 @@ def _create_engine(self, llm_kwargs: dict[str, Any]) -> None:
 
     def post_init(self):
         if self.llm is not None:
+            # Nemotron Omni checkpoints fold RADIO LayerScale into adjacent
```
rohitrango, line 644: why is this code moved to `post_init`

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -283,6 +301,15 @@ def model_dump_chat_response_with_routed_experts(response: Any) -> dict[str, Any
         )
         if routed_experts is not None:
             choice_dict.setdefault("message", {})["routed_experts"] = routed_experts
+
+    non_finite_paths = _find_non_finite_float_paths(response_dict)
```
rohitrango, line 305: this looks like an issue with the chat template. how is this code triggered?

docker/Dockerfile
```diff
@@ -295,7 +295,7 @@ if [[ -n "${NEMO_GYM_PREFETCH_CONFIGS:-}" ]]; then
         "${NEMO_GYM_VLLM_VERSION}" "${NEMO_GYM_VLLM_VERSION}" "${WHEEL_ARCH}" > "${GYM_VLLM_OVERRIDE}"
     UV_TORCH_BACKEND="${NEMO_GYM_CUDA}" \
     UV_OVERRIDE="${GYM_VLLM_OVERRIDE}" \
-    UV_LINK_MODE=symlink uv run python examples/nemo_gym/prefetch_venvs.py $NEMO_GYM_PREFETCH_CONFIGS
+    UV_LINK_MODE=copy uv run python examples/nemo_gym/prefetch_venvs.py $NEMO_GYM_PREFETCH_CONFIGS
```
rohitrango, line 298: revert to symlink to keep the layer small

rohitrango (changes requested):

PR #3127, opened 2026-07-08
feat(sglang): enable NeMo-Gym async-GRPO rollouts on the SGLang backend

Kh4L:
## What

Enables NeMo-Gym rollouts in async GRPO with the SGLang generation backend, on top of the merged SGLang Megatron-refit support (#3612) and the required rollout fault-tolerance layer (#3613).

The PR:

- makes `SGLangGeneration` safe to pass to the trajectory-collector actor;
- exposes the SGLang engine context and HTTP endpoints to shared rollout code;
- starts NeMo Gym for SGLang and normalizes engine roots for Gym's token-ID `/generate` transport;
- routes async training through the SGLang weight synchronizer added by #3612;
- makes SGLang memory-saver tag transitions idempotent, which is required when disaggregated async rollout keeps engines resident between refits; and
- reconciles replacement engines to the tracked memory state before refit.

## Why

Async GRPO previously rejected SGLang even though the backend now supports Megatron-to-SGLang weight refit. NeMo Gym also needs a serializable generation handle and reachable rollout endpoints inside its actor.

The pinned SGLang router does not preserve SGLang's prompt-token extension fields on `/v1/chat/completions`, so this integration uses Gym's supported `/generate` token-ID path. Tool calls remain parsed by Gym on that path.

## Dependency

This PR depends on #3613 because it extends the health-monitor lifecycle and replacement-engine recovery path. It does not depend on the MXFP8/NVFP4 work in #3614/#3615. Review the two commits after #3613.

## Validation

- Ruff and syntax checks pass for the changed files.
- HSG job `6595733` completed successfully (`0:0`) on 2 nodes / 8 GB200 GPUs: Qwen2.5-Math-1.5B, 4-GPU Megatron training + four TP1 SGLang engines, and real stateful multi-turn NeMo-Gym async GRPO over `/generate`.
- The validated integration source was #3063 head `60140cd23496d80c3ae04cd3a5fc276c38d44a60`, containing #3127 head `0ce9be27fb8528e54b9a968530503cba7336653e`, on the image built from #3615 head `736472843ba46eca4c16f127009452bdb148f090`.
- Current #3127 head `17075a1cd16b3d1c4b27445610b314e1ecc5ba7f` replays the same two patch-equivalent commits directly onto #3613; #3614/#3615 are no longer in its ancestry. The HSG run used BF16 Qwen with quantization disabled, so those quantization layers were not exercised by this validation.
[...]

PR #3330, opened 2026-07-23
feat(sglang): add end-to-end training and refit integration

Kh4L:
## Summary

This integration branch makes SGLang a reproducible generation backend for
NeMo RL training and in-memory weight refit:

- synchronous and asynchronous GRPO rollouts, including NeMo-Gym;
- Megatron-to-SGLang non-colocated NCCL broadcast refit;
- colocated CUDA-IPC refit for the supported policy backends;
- fail-closed engine startup, refit deadlines, cleanup, and quarantine;
- immutable-source, immutable-image, public two-node smoke and 16-node async
  integration reproducers with machine-checked evidence.

There is no disk-refit fallback.

## Dependency stack

This is an integration/reproducer PR. The component PRs remain authoritative
for their respective lanes:

- #3187: SGLang engine fault tolerance;
- #3190: Megatron-to-SGLang weight refit;
- #3188: MXFP8 rollout/refit support;
- #3189: side-by-side communicator split fix;
- #3127: async-GRPO and NeMo-Gym enablement, reduced here to the residual work
  not supplied by the refit stack;
- #3063: configurable handling of non-contiguous Gym episodes;
- NVIDIA-NeMo/Gym#1787: the native SGLang Responses adapter pinned by the Gym
  submodule.

## Refit dispatch fix

Running the two-node reproducer surfaced a defect that made every SGLang refit
fail on its first attempt:

```
NRL_SGLANG_REFIT_FAILURE transfer=broadcast refit=1 \
    phase=prepare_or_transfer error_type=ModuleNotFoundError
```

`_refit_sglang_dispatch` runs in the driver but imported a policy worker module
to reach `refit_sglang_{colocated,distributed}`. Those modules import
`megatron.bridge` and `nemo_automodel` at module scope, and the driver
environment can hold neither: `sglang` is declared mutually exclusive with
`mcore`, `fsdp`, `automodel` and `vllm` in `[tool.uv] conflicts`. Both transfer
modes were affected, and `SGLangColocatedWeightSynchronizer._refit_colocated`
carried an independent copy of the same import. The DTensor path had the
identical latent failure on `nemo_automodel`, which is why both backends are
touched.

Those four helpers are driver-side orchestration: they drive both actor groups
through the pause/transfer/quarantine protocol and call only `policy` and
`policy_generation` facades, so they were simply filed in the wrong module.
They move verbatim to `nemo_rl/weight_sync/`, beside the synchronizer that
calls them, and the worker modules re-export the names. A new test parses the
source rather than importing it, so it fails on the offending commit even in an
environment where every backend happens to be installed.

The fix is isolated in its own commit so it can be cherry-picked onto #3190.

## Safety semantics

- SGLang refit is a synchronization barrier; replay collection may be
  pipelined, but generation does not overlap a SGLang weight update.
- Initial engine launch, health checks, cache flush, router registration, and
  driver readiness share one startup deadline.
- Non-colocated broadcast refits have one end-to-end deadline. A failed or
  uncertain partial update aborts the communicator and quarantines the engines
  instead of resuming generation.
- Colocated partial-update failures also quarantine engines. Trainer-side
  gather and per-bucket IPC waits on that path are not yet deadline-bounded and
  are documented as requiring job termination if they stall.

## Validation

CPU-only:

- Ruff check and format check across all changed Python files;
- Python byte-compilation across all changed Python files;
- `git diff --check`;
- 11/11 evidence-validator tests;
- 3/3 driver-import guard tests;
- recipe inheritance and test-suite invariant checks;
- shell syntax and dry-run checks for both reproducer drivers;
- targeted startup/config/router/refit cleanup tests;
- 36/36 focused Gym SGLang-adapter tests.

GPU, two-node smoke (2 nodes, 4 GPUs per node, public Qwen2.5-Math-1.5B-Instruct
at the pinned revision), run against a driver environment built without the
`mcore` extra so the dispatch fix is exercised rather than worked around:

```json
{
  "markers": {"world_size": 5, "engines": 4,
              "group_ready_count": 1, "refit_success_count": 3},
  "metrics": {"expected_max_step": 3, "max_recorded_step": 3},
  "status": "passed"
}
```

The five-rank communicator (trainer rank 0 plus four engine ranks) formed,
streamed weights, and generation resumed on the refitted policy three times
across three GRPO steps with no failure marker. Refit wall time was 5.9 s for
the first transfer, which builds the communicator, then 0.24 s.

This smoke was re-run after rebasing the branch onto current `main`, so the
evidence above describes the branch as it now stands. Both the pre-rebase and
post-rebase runs produced identical markers.

Caveats on that evidence, stated plainly:

- it was produced from a source checkout mounted into an existing prebuilt
  container, not from a digest-qualified image built from this commit, so
  `image_ref` in the bundle is a placeholder and the dependency set is this
  branch's lockfile over that container's base;
- the 16-node asynchronous NeMo-Gym integration run has not been performed.

## Reproduce

See `docs/guides/sglang-refit.md`.

The manual public evidence runs are:

1. A two-node, four-GPU-per-node synchronous smoke test. It requires the exact
   `world_size=5 engines=4` refit topology, at least two successful refits, no
   fatal refit marker, and `train/loss` through step 3.
2. A 16-node, eight-GPU-per-node asynchronous NeMo-Gym integration test. It
   requires the exact `world_size=65 engines=32` topology, at least two
   successful refits, no fatal refit marker, `train/loss` through step 3, and a
   non-empty Gym rollout artifact.

Both drivers require an immutable model revision, a digest-qualified image, a
publicly fetchable pinned Gym commit, and a fresh shared initial
Hugging-Face-to-Megatron conversion cache.

## Remaining publication gates

- [x] Publish the cryptographically signed Gym commit stack and pin its final
      public SHA (NVIDIA-NeMo/Gym#1787; all commits signature-verified, pinned
      by the Gym submodule).
- [x] Demonstrate the two-node refit path end to end on GPUs.
- [x] Rebase onto current `main` and re-run the smoke.
- [x] Prove a clean recursive clone from public HTTPS remotes end to end. A
      `git clone --recursive` of the branch, run with no credentials and with
      global and system git config disabled so no `url.insteadOf` rewrite can
      apply, resolves all four submodules from their public remotes, including
      the pinned Gym commit fetched from `NVIDIA-NeMo/Gym`. `uv lock --check`
      then passes against the committed lockfile in that fresh clone.
- [ ] Build and publish the exact digest-qualified image and re-run the smoke
      against it.
- [ ] Run and attach a sanitized 16-node async integration evidence summary.


tianyi-zhang-02:
@Kh4L — small heads-up on `docs/guides/refit.md`, since this draft and my #3519 both rewrite the same paragraph and the base has moved under this one.

This PR's diff removes:

> Non-colocated SGLang generation is not supported.

but `main` no longer says that. It now reads:

> Non-colocated SGLang generation is supported with a Megatron policy and `refit_transport: null`; it creates its own NCCL weight-update group.

so the `-` side of the hunk won't apply as written, and the replacement it offers ("uses a GRPO-specific side communicator and requires the Megatron policy backend") is close to what's already there rather than a change to it.

#3519 rewrites the same paragraph to add the second path — checkpoint-engine refit via `refit_transport: nixl`, which works with either policy backend. Also updates the transport table row that this PR's table hunk sits next to.

Not asking for anything; the two are compatible in substance and it's a one-paragraph resolution whichever lands first. Flagging it mainly because this draft has been idle since 2026-08-11 and #3612 landed the Megatron→SGLang refit on 2026-08-26, which is what made the old "not supported" line stale. :)

PR #3331, opened 2026-07-23
feat: standardize mcore generation defaults and expose CUDA graph sizing

shanmugamr1992:
## Summary

- expose Megatron-Core's `cuda_graph_sizing_distribution` setting and standardize CUDA-graph/refit defaults across exemplar and reference configs
- surface cumulative prefix-cache counters without overriding `NRL_LOG_LEVEL`
- cover prefix caching and MXFP8 training through the existing Nano nightly suites

## Testing

- `bash -n` on both modified nightly drivers
- `ruff check` on the modified Megatron generation Python files
- recipe/test-suite registration checks (11 passed)
- direct comparison of modified exemplar/reference MCore config blocks

The GPU nightly runs are delegated to CI.


shanmugamr1992:
/ok to test [7ddf92c](https://github.com/NVIDIA-NeMo/RL/pull/3331/commits/7ddf92cca313a5d9ce0f4341533adede55fcdbaa)

tdene (dismissed):

shanmugamr1992:
/ok to test [943e39f](https://github.com/NVIDIA-NeMo/RL/pull/3331/commits/943e39f99ec36eead6b9d75d668cd9ca1288ba14)

shanmugamr1992:
/ok to test [8df209a](https://github.com/NVIDIA-NeMo/RL/pull/3331/commits/8df209a0930212ee5d4adb02450c457e4d13e1a2)

shanmugamr1992:
/ok to test [0470e26](https://github.com/NVIDIA-NeMo/RL/pull/3331/commits/0470e26142811605c47de9f4a580073769f85c13)

shanmugamr1992:
/ok to test [cae34dc](https://github.com/NVIDIA-NeMo/RL/pull/3331/commits/cae34dc622ed90e37f918458c53c8e578f1a4579)

tests/functional/L2_Functional_Tests_Megatron_4.sh
```diff
@@ -0,0 +1,54 @@
...
+        time "$@"
+    elif [[ "${FAST:-0}" == "1" ]]; then
+        echo "FAST: Skipping: $*"
+    else
+        time "$@"
+    fi
+}
+
+
+run_test      bash ./tests/functional/grpo_megatron_generation_async_prefix_caching.sh
```
terrykong: [`tests/functional/L2_Functional_Tests_Megatron_4.sh:47`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/L2_Functional_Tests_Megatron_4.sh#L47)

**2 action items.**

**TL;DR — please delete this file. There is no L2 script lane, and these two tests ask for 2 nodes x 4 GPUs, which no functional runner has. Fold the coverage into an existing nightly instead.**

PR-introduced.

### AI-1 — delete `L2_Functional_Tests_Megatron_4.sh`

Two separate reasons it cannot work here:

**There is no `L2_` script lane.** CI discovers functional suites by filename glob, and the glob is `L1_` only — [`plan-tests.sh#L51`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/.github/scripts/plan-tests.sh#L51). `CI:L2` is a test *level*, not a filename prefix, and today it selects the same jobs as `CI:L1`: every gate in `cicd-main.yml` that names L2 also names L1, so no job runs only at L2. The planner-coverage guard also globs `L1_` only ([`cicd-main.yml#L842`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/.github/workflows/cicd-main.yml#L842)), so nothing warns you the file is unreachable.

**`tests/functional/` is single-node.** The functional runner is one node with a couple of GPUs, and every existing test matches that. These two scripts are the only files in the whole directory that ask for more than one node:

```
$ grep -l 'cluster.num_nodes=[2-9]' tests/functional/*.sh
tests/functional/grpo_megatron_generation_async_mxfp8.sh
tests/functional/grpo_megatron_generation_async_prefix_caching.sh
```

For contrast, 10 of the existing `grpo_megatron_generation*.sh` tests use `cluster.gpus_per_node=2` and set no `num_nodes` at all. So even renaming these to `L1_` would not help — the allocation does not exist.

### AI-2 — fold the coverage into an existing nightly

Multi-node runs live in `tests/test_suites/`, and the house habit is to add args to an existing recipe rather than add a new driver. Two concrete homes:

**Prefix caching** → [`tests/test_suites/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation.sh`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation.sh) (listed in [`nightly.txt#L120`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/nightly.txt#L120)). This already drives the exact recipe your script passes to `--config`, and it is already non-colocated (`colocated.enabled: false`). Adding `policy.generation.mcore_generation_config.enable_prefix_caching=true` plus the prompt-file override there gets the coverage with no new driver. If you would rather not slow the 30B run, [`grpo-llama3.2-1b-instruct-2n8g-megatron_generation-noncolocated.sh`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/llm/grpo-llama3.2-1b-instruct-2n8g-megatron_generation-noncolocated.sh) ([`nightly.txt#L104`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/nightly.txt#L104)) is the cheap version of the same shape.

**MXFP8** → needs Blackwell, so it has to be a GB200 suite. Existing MXFP8 coverage is vLLM-only ([`grpo_vllm_mxfp8_rollout_gb200.sh`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh)), so megatron-generation + MXFP8 is a genuine gap. The closest GB200 nightlies are [`grpo-nanov3-30BA3B-4n4g-megatron_async_colocated.sh`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_async_colocated.sh) and [`grpo-nanov3-30BA3B-4n4g-megatron_colocated_reshard.sh`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_colocated_reshard.sh) ([`nightly_gb200.txt`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/nightly_gb200.txt)), both colocated. If colocated coverage is enough, add the `policy.megatron_cfg.fp8_cfg.*` args to one of those. If non-colocated is the point of the test, this is the one case where a new `tests/test_suites/llm/` driver plus a `nightly_gb200.txt` entry is justified — note the naming rules in [`tests/test_suites/README.md`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/README.md), including that the driver filename must mirror a recipe YAML under `examples/configs/recipes/llm/`, since [`common.env`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/test_suites/llm/common.env#L49) derives `CONFIG_PATH` from it and exits if that file is missing.

One knock-on: my other comments on these two scripts (the `block_size_tokens` header note, the prefix-cache hit assertion, the prompt file location, the cluster name) apply to wherever that logic ends up, not to these paths specifically.

Last thing — all three new scripts point at `.claude/skills/run-nano35-megatron-inference-cog/SKILL.md` for "the wiring", and that file is not in the repo (`ls .claude/skills/` shows 15, none matching; `git ls-files | grep -i nano35` is empty). If any of this survives as a manual script, please inline the setup instead of pointing at something nobody else has.

Not a `suggestion` block: this is a file deletion plus edits in other files.

tests/functional/grpo_megatron_generation_async_mxfp8.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+
+# Async non-colocated Megatron generation on nano-3.5 (30B-A3B mamba-hybrid MoE)
+# with MXFP8 training. Runs generation on a dedicated inference node while training
+# overlaps on the training node (1-off async, in-flight weight updates), quantizing
+# the training GEMMs to MXFP8.
+#
+# This is a CLUSTER test (GB200 / oci-hsg). It expects to run from /opt/nemo-rl on
```
terrykong: [`tests/functional/grpo_megatron_generation_async_mxfp8.sh:8`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/grpo_megatron_generation_async_mxfp8.sh#L8)

**1 action item.** PR-introduced, small.

This is a public repo, and `oci-hsg` is an internal cluster name. Referring to the hardware by GPU SKU alone says everything a reader needs.

Three lines, all new in this PR — the other two are [`grpo_megatron_generation_async_prefix_caching.sh#L16`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/grpo_megatron_generation_async_prefix_caching.sh#L16) and [`L2_Functional_Tests_Megatron_4.sh#L27`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/L2_Functional_Tests_Megatron_4.sh#L27). I checked: there is no pre-existing mention anywhere else in the tree, so this would be the first.

**Action:** drop the cluster name, keep `GB200`. Same edit in the other two files.

```suggestion
# This is a CLUSTER test (GB200). It expects to run from /opt/nemo-rl on
```

examples/configs/ppo_math_1B.yaml
```diff
@@ -266,12 +266,16 @@ policy:
     mcore_generation_config:
       buffer_size_gb: 20
       buffer_guaranteed_fraction: 0.1
-      num_cuda_graphs: 16
+      cuda_graph_impl: "local"  # CUDA-graph implementation. Options: "none", "local", "transformer_engine", "full_iteration".
+      cuda_graph_sizing_distribution: "linear"  # How captured CUDA-graph sizes are spaced. Options: "exponential" (log-spaced), "linear" (denser at the top end).
+      inference_cuda_graph_scope: "block"  # Inference CUDA-graph scope. Options: "none", "layer", "block".
```
terrykong: [`examples/configs/ppo_math_1B.yaml:269-271`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/configs/ppo_math_1B.yaml#L269-L271)

**1 action item.**

**TL;DR — this re-adds the `cuda_graph_impl` comment that the commit right before this PR deliberately deleted, so after merge the repo documents the same key two contradictory ways.**

PR-introduced.

[`eaf02ef6b`](https://github.com/NVIDIA-NeMo/RL/commit/eaf02ef6b) ("fix: Megatron Inference colocated CUDA graphs", #3718) removed exactly this line from `grpo_math_1B.yaml` and replaced it with a pairings block:

```diff
-      cuda_graph_impl: "local"  # CUDA-graph implementation. Options: "none", "local", "transformer_engine", "full_iteration".
+      # Generation CUDA graphs. MCore accepts only these pairings:
+      #   cuda_graph_impl "none"  -> omit inference_cuda_graph_scope
+      #   cuda_graph_impl "local" -> inference_cuda_graph_scope "layer" or "block"
+      cuda_graph_impl: "local"
```

That block is still there, in a file this PR also edits ([`grpo_math_1B.yaml#L402-L406`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/configs/grpo_math_1B.yaml#L402-L406)). This PR puts the old wording back into `ppo_math_1B.yaml`, [`research/template_project/configs/grpo_math_1B.yaml#L299-L301`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/research/template_project/configs/grpo_math_1B.yaml#L299-L301), and [`tests/unit/reference_configs/ppo_math_1B_megatron.yaml#L245-L247`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/unit/reference_configs/ppo_math_1B_megatron.yaml#L245-L247).

The `inference_cuda_graph_scope` line has the same problem: it offers `"none"`, but `"none"` is not allowed next to `cuda_graph_impl: "local"` — Megatron-Core's [`ALLOWED_INFERENCE_SCOPES`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/transformer/cuda_graph_config.py#L17) maps `local` to `{layer, block}` only, asserted in [`transformer_config.py`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/transformer/transformer_config.py#L2521).

**Action:** copy the pairings block from `grpo_math_1B.yaml` into all three files instead. Same edit in the other two.

```suggestion
      # Generation CUDA graphs. MCore accepts only these pairings:
      #   cuda_graph_impl "none"  -> omit inference_cuda_graph_scope
      #   cuda_graph_impl "local" -> inference_cuda_graph_scope "layer" or "block"
      cuda_graph_impl: "local"
      cuda_graph_sizing_distribution: "linear"  # How captured CUDA-graph sizes are spaced. Options: "exponential" (log-spaced), "linear" (denser at the top end).
      inference_cuda_graph_scope: "block"
```

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -70,6 +72,11 @@
 )
 from nemo_rl.utils.nsys import wrap_with_nvtx_name
 
+logger = logging.getLogger(__name__)
+# NeMo-RL configures the root logger at WARNING (see nemo_rl/__init__.py), so opt
+# this module in to INFO explicitly; records still propagate to the root handler.
+logger.setLevel(logging.INFO)
```
terrykong: [`nemo_rl/models/generation/megatron/megatron_worker.py:76-78`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/models/generation/megatron/megatron_worker.py#L76-L78)

**1 action item.**

**TL;DR — this line does nothing on the default path, and when someone sets `NRL_LOG_LEVEL=WARNING` it silently forces this one module back to INFO.**

PR-introduced.

The comment says the root logger is at WARNING so the module has to opt in. The root logger part is true, but it is not the reason this module's records get through. [`nemo_rl/__init__.py#L40`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/__init__.py#L40) already does:

```python
logging.getLogger("nemo_rl").setLevel(os.environ.get("NRL_LOG_LEVEL", "INFO").upper())
```

`nemo_rl.models.generation.megatron.megatron_worker` is a child of `nemo_rl`, so it inherits INFO. The comment block just above that line ([L23-L39](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/__init__.py#L23-L39)) explains the choice: set the level on `nemo_rl` rather than root, on purpose, so the env var controls it.

So the new line is a no-op when things work, and breaks the env var when they do not:

```
$ NRL_LOG_LEVEL=WARNING ...   # expected: this module goes quiet
                              # actual:   setLevel(INFO) puts it straight back
```

It also cannot be turned up — `NRL_LOG_LEVEL=DEBUG` no longer reaches this module either. It is the only module-level `setLevel` on an `nemo_rl.*` logger in the package.

**Action:** delete the two comment lines and the `setLevel` call. The INFO line still prints — `basicConfig` at [`__init__.py#L20`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/__init__.py#L20) installs a root handler with no level, and a handler is what filters a propagated record, not the root logger's level.

```suggestion
logger = logging.getLogger(__name__)
```

tests/functional/grpo_megatron_generation_async_prefix_caching.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+
+# Async non-colocated Megatron generation on nano-3.5 (30B-A3B mamba-hybrid MoE)
+# with PREFIX CACHING enabled, in a scenario engineered to force real cache hits:
+#   - num_prompts_per_step=1, num_generations_per_prompt=16 -> every rollout in a
+#     step is the SAME prompt, so the whole prompt is a shared prefix.
+#   - a long fixed instruction preamble (examples/prompts/prefix_caching_cot_prompt.txt,
+#     ~430 tokens) is prepended to every prompt. mcore only caches/matches WHOLE
+#     KV blocks and the paged-KV block size must be divisible by 256, so the prefix
```
terrykong: [`tests/functional/grpo_megatron_generation_async_prefix_caching.sh:9`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/grpo_megatron_generation_async_prefix_caching.sh#L9)

"the paged-KV block size must be divisible by 256" is not a Megatron-Core rule. At the pinned commit there are exactly two constraints on `block_size_tokens`, and neither mentions 256:

- Flash MLA requires it to be exactly **64** — [`dynamic_context.py#L363-L366`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/contexts/dynamic_context.py#L363-L366)
- `num_speculative_tokens < block_size_tokens` — [`dynamic_context.py#L386`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/contexts/dynamic_context.py#L386)

The rest of the sentence is right: matching really is whole-block ([`dynamic_context.py#L3003`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/contexts/dynamic_context.py#L3003) gates reuse on `finished % block_size_tokens == 0`). The real reason the preamble has to be long is simply that this recipe sets [`block_size_tokens: 256`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/configs/grpo_math_1B.yaml#L409) — a value you can change.

Worth correcting because the next person tuning this test would read it as a hard rule and conclude 64 is illegal.

```suggestion
#     KV blocks, and this recipe sets block_size_tokens=256, so the prefix
```

examples/configs/grpo_math_1B.yaml
```diff
@@ -398,13 +398,14 @@ policy:
         shard_expert_weights: false
     mcore_generation_config:
       max_model_len: ${policy.max_total_sequence_length}  # Engine-side max sequence length.
+      cuda_graph_sizing_distribution: "linear"  # How captured CUDA-graph sizes are spaced. Options: "exponential" (log-spaced), "linear" (denser at the top end).
```
terrykong: [`examples/configs/grpo_math_1B.yaml:401`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/configs/grpo_math_1B.yaml#L401)

**1 action item — a question, not a change request.**

**TL;DR — `linear` + `num_cuda_graphs: -1` captures roughly 51 graph sizes where the old `4` captured 4, and Megatron-Core's own comment describes `linear` as the legacy mode kept for vLLM-comparison benchmarking.**

PR-introduced. This is the one config choice in the PR I could not talk myself into from the diff alone.

Counting the sizes at the pinned Megatron-Core:

1. `linear` + `-1` returns `[1, 2, 4] + range(8, 256, 8) + range(256, cuda_graph_max_tokens + 1, 16)` — [`batch_dimensions_utils.py#L340-L350`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/batch_dimensions_utils.py#L340-L350).
2. NeMo-RL never sets `cuda_graph_max_tokens`, so it takes mcore's default of **512** ([`config.py#L264`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/config.py#L264)), clamped by [`dynamic_context.py#L764-L769`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/contexts/dynamic_context.py#L764-L769) to `min(max(512, max_requests), max_tokens)` — and `max_tokens` here is 16384, so a large `max_requests` pushes it above 512, not below.
3. At 512 that is **51 sizes**. For contrast, `exponential` + `-1` recurses with `auto_n = max(4, log2(512) + 2) = 11` → about **10 sizes** ([`batch_dimensions_utils.py#L258-L273`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/batch_dimensions_utils.py#L258-L273)).
4. Before this PR these configs had `num_cuda_graphs: 4` and no distribution key, so mcore's `exponential` default gave ~4.

There is a second difference the key's comment does not mention. `linear` collapses the mixed prefill+decode graphs to a single prefill-request count, while `exponential` fans out over a doubling grid — [`batch_dimensions_utils.py#L531-L546`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/batch_dimensions_utils.py#L531-L546), which says:

> Under EXPONENTIAL distribution (default): generate mixed CGs across a geometric P-grid {1, 2, 4, ..., max_requests}. This bounds the relative overhead per real batch (~2x P slack worst case) and **is the structural fix** that makes mixed CGs usable for real batches with P != fixed_P.
>
> Under LINEAR distribution: use **the legacy fixed P value** ... same single-P behavior main has today, **for apples-to-apples benchmarking against vLLM-style configurations**.

**Action:** could you say in the PR description what made `linear` the right default, and roughly what 51 graphs costs in capture time and graph memory versus the old 4? You have presumably measured this. It matters because these are the small tutorial configs people run on one GPU, and because it also lands on the two 30B-A3B MoE recipes ([`grpo_nanov3.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/nemo_gym/grpo_nanov3.yaml), [`grpo_qwen3_30ba3b_instruct.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml)). If the win is only on the big recipes, a fixed count on the 1B exemplars would keep tutorial startup fast.

Not a `suggestion` block — the ask is numbers, not an edit.

examples/configs/grpo_math_1B.yaml
```diff
@@ -413,7 +414,7 @@ policy:
       kv_cache_management_mode: "persist"  # KV cache lifecycle across suspend/resume. Options: "persist", "offload". To select "recompute", set grpo.async_grpo.recompute_kv_cache_after_weight_updates=true.
       materialize_only_last_token_logits: true
       num_speculative_tokens: 0
-      refit_backend: "nccl"  # Copy-service backend for non-colocated megatron weight refit. Options: "gloo" or "nccl"
+      refit_backend: "nccl"  # Copy-service backend for non-colocated megatron weight refit. Options: "gloo", "nccl", "nvshmem".
```
terrykong: [`examples/configs/grpo_math_1B.yaml:417`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/configs/grpo_math_1B.yaml#L417)

The comment changed from `Options: "gloo" or "nccl"` to `Options: "gloo", "nccl", "nvshmem".` — but `nvshmem` does not work today. The worker warns as soon as you pick it:

```python
if refit_backend == "nvshmem":
    warnings.warn(
        'refit_backend="nvshmem" is currently broken; prefer "nccl" or '
        '"gloo". See https://github.com/NVIDIA-NeMo/RL/issues/3646',
        stacklevel=2,
    )
```
([`megatron_worker.py#L894-L899`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/models/generation/megatron/megatron_worker.py#L894-L899)) — and NVIDIA-NeMo/RL#3646 ("MCore refit backend nvshmem does not function") is still open.

It is accurate that `nvshmem` is an accepted value, but the old comment left it out, which was less likely to send someone down a broken path.

**Action:** keep the caveat. Same edit in the five other files that got this change: [`ppo_math_1B.yaml#L278`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/configs/ppo_math_1B.yaml#L278), [`grpo_nanov3.yaml#L249`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/nemo_gym/grpo_nanov3.yaml#L249), [`research/template_project/configs/grpo_math_1B.yaml#L307`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/research/template_project/configs/grpo_math_1B.yaml#L307), [`tests/unit/reference_configs/grpo_math_1B.yaml#L402`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/unit/reference_configs/grpo_math_1B.yaml#L402), [`tests/unit/reference_configs/ppo_math_1B_megatron.yaml#L254`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/unit/reference_configs/ppo_math_1B_megatron.yaml#L254).

```suggestion
      refit_backend: "nccl"  # Copy-service backend for non-colocated megatron weight refit. Options: "gloo", "nccl" ("nvshmem" is accepted but currently broken, see #3646).
```

examples/configs/grpo_math_1B.yaml
```diff
@@ -398,13 +398,14 @@ policy:
...
       max_model_len: ${policy.max_total_sequence_length}  # Engine-side max sequence length.
+      cuda_graph_sizing_distribution: "linear"  # How captured CUDA-graph sizes are spaced. Options: "exponential" (log-spaced), "linear" (denser at the top end).
       # Generation CUDA graphs. MCore accepts only these pairings:
       #   cuda_graph_impl "none"  -> omit inference_cuda_graph_scope
       #   cuda_graph_impl "local" -> inference_cuda_graph_scope "layer" or "block"
       cuda_graph_impl: "local"
       inference_cuda_graph_scope: "block"
       buffer_size_gb: 10
-      num_cuda_graphs: 4  # Number of CUDA graphs to pre-compile for different batch sizes
+      num_cuda_graphs: -1  # Number of CUDA graphs to pre-compile for different batch sizes (-1 = auto).
```
terrykong: [`examples/configs/grpo_math_1B.yaml:408`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/configs/grpo_math_1B.yaml#L408)

**1 action item.** PR-introduced staleness in a file outside the diff.

[`docs/design-docs/generation.md`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/docs/design-docs/generation.md) is the only page that documents `mcore_generation_config`, and it still shows the old value in both of its example blocks — [L134](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/docs/design-docs/generation.md#L134) and [L209](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/docs/design-docs/generation.md#L209) both say `num_cuda_graphs: 16`. `cuda_graph_sizing_distribution` appears nowhere in `docs/`.

So after this PR every config says `-1` while the doc says `16`, and the key the PR exists to expose is undocumented.

**Action:** change both snippets to `-1` and add a bullet for the new key next to the existing `num_cuda_graphs` bullet at [L143](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/docs/design-docs/generation.md#L143). Suggested text:

> - **num_cuda_graphs**: Number of CUDA graphs to pre-allocate for different batch sizes. More graphs avoid runtime graph capture but use more memory and make engine startup slower. Use `-1` to let Megatron-Core choose the set.
> - **cuda_graph_sizing_distribution**: How the captured sizes are spaced. `"exponential"` (Megatron-Core's default) halves from the maximum down, giving about `log2(max_tokens)` graphs. `"linear"` captures many more, packed more densely near the top.

Not a `suggestion` block: the file is not in this diff, so there is no line to attach one to.

tests/functional/grpo_megatron_generation_async_prefix_caching.sh
```diff
@@ -0,0 +1,125 @@
...
+# Prefix-cache HIT assertion: with num_generations_per_prompt=16 the same prompt
+# prefix is served many times, so the mcore prefix cache MUST report hits. mcore's
+# own DynamicInferenceEngine counters are only surfaced via its (wandb) metrics
+# writer / a logging.info line that NeMo-RL does not configure to emit, so the
+# MegatronPolicyWorker mirrors the cumulative counters to captured stdout when
+# prefix caching is on (see megatron_worker.py `_sleep`):
+#     [Rank <r>] mcore prefix cache (cumul): <N> hits, <M> blocks matched
+# Fail loudly if no hit is observed, so an accidentally-disabled cache (or a
+# regression that stops matching prefixes) cannot pass silently.
+PREFIX_CACHE_HIT_PATTERN='mcore prefix cache \(cumul\): [1-9][0-9]* hits'
```
terrykong: [`tests/functional/grpo_megatron_generation_async_prefix_caching.sh:113-115`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/grpo_megatron_generation_async_prefix_caching.sh#L113-L115)

**1 action item, 1 optional follow-up.**

**TL;DR — the counter only ever goes up, so once step 1 records a single hit this grep passes for the rest of the run; the comment right above it promises a guarantee the check does not give.**

PR-introduced.

How the counter works:

1. Every step, mcore adds the per-step hits from the context into an engine-level total and zeroes the per-step ones — [`dynamic_engine.py#L2352-L2361`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/engines/dynamic_engine.py#L2352-L2361).
2. The engine-level `_prefix_cache_hits` is set to 0 exactly once, in `__init__` — [`dynamic_engine.py#L393`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/engines/dynamic_engine.py#L393). Nothing in the pause/suspend path resets it.
3. So the number the worker logs is a running total across all 5 steps.

That means a run where prefix matching works on step 1 and then stops completely still prints `... 1 hits` on every later step and passes `[1-9][0-9]* hits`. The comment at [L113-L114](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/grpo_megatron_generation_async_prefix_caching.sh#L113-L114) says "a regression that stops matching prefixes cannot pass silently" — with this check, it can.

**Action:** check the final count against a floor instead of against zero. With `num_prompts_per_step=1` and `num_generations_per_prompt=16` over 5 steps there is a lot of headroom, so even a low floor catches "it hit once and stopped":

```bash
PREFIX_CACHE_HIT_PATTERN='mcore prefix cache \(cumul\): ([0-9]+) hits'
hits=$(grep -aoE 'mcore prefix cache \(cumul\): [0-9]+ hits' "$RUN_LOG" | grep -oE '[0-9]+' | tail -1)
if [[ -n "$hits" && "$hits" -ge 10 ]]; then
    echo "PASS: prefix-cache hits (cumulative) = $hits"
else
    echo "FAIL: expected a cumulative prefix-cache hit count >= 10, got '${hits:-none}'."
    exit 1
fi
```

If you would rather keep it as a smoke check, that is fine — then please soften the comment above so it does not claim more than it checks.

### Follow-up (optional, no tracking issue yet)

Two smaller things in the same area, both about [`_sleep`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/models/generation/megatron/megatron_worker.py#L375-L376):

- `getattr(engine, "_prefix_cache_hits", 0)` silently yields 0 if Megatron-Core ever renames that private attribute, and this test would then fail with "no prefix-cache hit signal found" — pointing the reader at the cache instead of at a missing attribute. Dropping the `, 0` makes it raise instead. (Reading the private attribute is the right call, to be clear — mcore exposes no public accessor; its metrics path needs a wandb writer and its console line goes to the root logger.)
- There is already a [`get_inference_cuda_graph_capture_count()`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/models/generation/megatron/megatron_worker.py#L195) worker method next door that unit tests call ([`test_megatron_generation.py#L494`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/unit/models/generation/test_megatron_generation.py#L494)). A `get_prefix_cache_stats()` beside it would let this assert on a number instead of grepping a log, and would give the unit test below something to call.

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -295,6 +302,13 @@ def _initialize_inference_engine(self, mcore_generation_config: dict) -> None:
             max_requests=max_requests,
         )
 
+        if "cuda_graph_sizing_distribution" in mcore_generation_config:
+            inference_config.cuda_graph_sizing_distribution = (
+                CudaGraphSizingDistribution(
+                    mcore_generation_config["cuda_graph_sizing_distribution"]
+                )
+            )
```
terrykong: [`nemo_rl/models/generation/megatron/megatron_worker.py:305-310`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/models/generation/megatron/megatron_worker.py#L305-L310)

Nothing in `tests/unit/` reaches this branch — grepping `cuda_graph_sizing_distribution` across `tests/` hits only the two reference YAMLs, which are compared as data and never loaded into a worker. So the feature this PR exists to add has no test, and the two new shell tests that would exercise it run in no CI lane (see my comment on `L2_Functional_Tests_Megatron_4.sh`).

`CudaGraphSizingDistribution(...)` is a bare enum call, so a bad value raises `ValueError` at engine-init time — minutes into a multi-node run rather than at config load.

**Action:** the cheapest useful test needs no GPU and no mocks — assert the `Literal` in the config still matches mcore's enum, so a Megatron-Core bump that renames a member fails in CI instead of on a cluster. [`tests/unit/models/generation/test_megatron_generation_parse.py`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/unit/models/generation/test_megatron_generation_parse.py) is the file for CPU-only slices of this worker:

```python
def test_cuda_graph_sizing_distribution_literal_matches_mcore():
    from typing import get_args
    from megatron.core.inference.config import CudaGraphSizingDistribution
    from nemo_rl.models.generation.megatron.config import MCoreGenerationSpecificArgs

    declared = get_args(
        get_args(
            MCoreGenerationSpecificArgs.__annotations__["cuda_graph_sizing_distribution"]
        )[0]
    )
    assert set(declared) == {m.value for m in CudaGraphSizingDistribution}
```

I could not run it here (no GPU, and `uv run` needs the linux lock), so please check it passes before landing.

Worth knowing if you add it: that file carries no `mcore` marker today, and no L0 lane collects it — the `Models_*` lanes pass `--ignore=unit/models/generation/`, and `L0_Unit_Tests_Megatron.sh` keeps only items marked `mcore` ([`conftest.py#L123-L133`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/unit/conftest.py#L123-L133)). A `pytestmark = pytest.mark.mcore` at the top would make the file actually run. It already imports `megatron.core` at module scope, so the marker only states what is already true.

Not a `suggestion` block: the change belongs in a different file.

tests/test_suites/fixtures/prefix_caching_cot_prompt.txt
```diff
@@ -0,0 +1,10 @@
+You are an extremely careful, methodical mathematical reasoning assistant. Your single most important objective is to arrive at a correct final answer by reasoning slowly, explicitly, and without skipping steps. Before you write anything, read the entire problem statement at least twice and make sure you understand exactly what quantity is being asked for, what information is given, and what constraints must hold. Restate the problem in your own words so that any hidden assumptions become visible. Identify every quantity, name it clearly, and write down its units if it has any. If the problem involves several parts, list the parts explicitly and address them one at a time in order.
```
terrykong, line 1: [`examples/prompts/prefix_caching_cot_prompt.txt`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/examples/prompts/prefix_caching_cot_prompt.txt)

This is a test fixture sitting in a user-facing directory. Please move it under `tests/`.

It is referenced from exactly one place — [`grpo_megatron_generation_async_prefix_caching.sh#L83`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/grpo_megatron_generation_async_prefix_caching.sh#L83) — and from no config or doc, unlike `cot.txt`, `math.txt`, `gsm8k.txt`, `geo3k.txt` and the rest of that directory, which are all pulled in by example configs.

It is also not a prompt anyone should copy: it is three paragraphs written to be long enough to fill one 256-token KV block. Someone browsing `examples/prompts/` will read it as a recommended chain-of-thought prompt.

Nothing ties it to `examples/`. `prompt_file` is opened as a plain path with no base directory ([`nemo_rl/data/interfaces.py#L70-L80`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/nemo_rl/data/interfaces.py#L70-L80)), and the script already passes an absolute path.

**Action:** move it to `tests/functional/fixtures/prefix_caching_cot_prompt.txt` and update line 83:

```bash
    data.default.prompt_file=$SCRIPT_DIR/fixtures/prefix_caching_cot_prompt.txt \
```

Not a `suggestion` block: the fix is a file move plus an edit in another file.

examples/configs/grpo_math_1B.yaml
terrykong: does nvshmem work now? @shanmugamr1992

terrykong (commented):
Reviewed with a team of specialized agents (RL codebase, Megatron-Core API, bug scan, tests, design, plus an adversarial pass). 11 inline comments below; the three biggest are the two new cluster tests running in no CI lane, a config comment this PR reintroduces after the previous commit removed it, and a `logger.setLevel` that overrides `NRL_LOG_LEVEL`.

Thanks for pulling these defaults together — having one `cuda_graph_sizing_distribution` knob wired end to end, and a functional test that asserts the prefix cache is actually *hit* rather than merely enabled, are both good additions.

## Before this can merge

- **The PR is `CONFLICTING`.** It needs a rebase or merge on `main`.
- **The title is truncated mid-word**: it ends `...expose cuda_gr` followed by a literal `…`, and the tail (`aph_sizing_distribution`) spilled into the first line of the description. Some tool cut it. Worth fixing — this becomes the squash-merge commit message, and the semantic-PR check only validates the `feat:` prefix so it will not catch this.
- **No test job has run on this head.** All ten green checks on `05ff25bd7` are metadata-only (DCO, copyright, secrets, semantic PR, submodule, triage) — there is no `cicd-main` run at all, because the last `/ok to test` was for an older commit. For a PR that edits six configs plus both reference configs, the green tick does not tell us much yet. Please `/ok to test` after the rebase.
- **The description says it disables the `grpo_megatron_generation_topp_topk` functional test, but that is not in the diff.** [`L1_Functional_Tests_Megatron_4.sh#L59`](https://github.com/NVIDIA-NeMo/RL/blob/05ff25bd7b5b0d40406b4ab1668a94e6499e3cd1/tests/functional/L1_Functional_Tests_Megatron_4.sh#L59) still runs it, and that file is not in the PR's file list. Probably a leftover line — worth removing so reviewers are not looking for a change that is not there.
- The template sections below your summary are still blank. More useful than filling them in mechanically: the description does not mention the two new cluster test scripts, the new prompt file, or the prefix-cache logging added to `_sleep()`, which is the change most likely to affect a running job.
- The only review this PR ever had was auto-dismissed when new commits landed, so it currently has no approvals.

## Checked and clean, so you do not have to re-check

- `num_cuda_graphs: -1` is genuinely supported — explicitly handled in three places at the pinned Megatron-Core, with the `>= 1` asserts only on the other branch. `examples/nemo_gym/grpo_nanov3.yaml` and the omni VLM recipe already used it before this PR.
- `CudaGraphSizingDistribution("linear")` (by value) next to `InferenceCudaGraphScope["block"]` (by name) looks inconsistent but both are correct: the first is a `str` Enum with string values, the second a plain Enum with int values. Each line uses the only form that works.
- Assigning `cuda_graph_sizing_distribution` after `InferenceConfig(...)` is built is safe — `__post_init__` does not read it, and the field is not consumed until `DynamicInferenceContext` is created further down.
- `_prefix_cache_hits` / `_prefix_cache_blocks_matched` exist under exactly those names, accumulate on every rank with no rank guard, and are cumulative — so logging from all ranks and the `(cumul)` label are both right.
- The exemplar and reference configs are in sync, so `test_config_v2.py::test_reference_configs_up_to_date` stays green. The four megatron configs that do not set the new key all inherit it through `defaults:`, including the VLM recipe.
- `bash -n` passes on all three new scripts; `=-1` parses as an integer through the config CLI; the grep pattern cannot collide with mcore's own similar log line; `ruff` (the repo's pinned 0.9.9) is clean on both changed Python files.
- The mxfp8 test's `< 1.5` bound is fine — it comes with a measured justification, and the only other FP8 test in the repo uses `< 2.0`.

_Generated by Claude Code_

examples/configs/ppo_math_1B.yaml
terrykong: cc @tdene since he landed the previous change

tests/functional/grpo_megatron_generation_async_prefix_caching.sh
shanmugamr1992: Done in ea0db4f96. The registered nightly extracts the final cumulative count and requires at least 10 hits. I also removed the silent getattr fallbacks so private-counter API drift fails at the source.

tests/functional/L2_Functional_Tests_Megatron_4.sh
shanmugamr1992: Done in ea0db4f96. Deleted the unreachable L2/functional drivers, moved prefix-cache coverage into the existing 2n8g Megatron-generation nightly, and added MXFP8 training to the existing GB200 async-colocated nightly.

tests/functional/grpo_megatron_generation_async_prefix_caching.sh
shanmugamr1992: Done in ea0db4f96. The manual script was removed; the nightly assertion now states that this recipe uses a 256-token block and does not imply a Megatron-Core divisibility rule.

examples/configs/grpo_math_1B.yaml
shanmugamr1992: Done in ea0db4f96. All six config comments now keep nvshmem documented as accepted but currently broken, with #3646 referenced.

nemo_rl/models/generation/megatron/megatron_worker.py
shanmugamr1992: Done in ea0db4f96. Removed the module-level setLevel call so NRL_LOG_LEVEL controls this logger through the nemo_rl hierarchy.

examples/configs/ppo_math_1B.yaml
shanmugamr1992: Done in ea0db4f96. Copied the valid CUDA-graph pairing block into the PPO exemplar, template config, and PPO reference config.

tests/test_suites/fixtures/prefix_caching_cot_prompt.txt
shanmugamr1992: Done in ea0db4f96. Moved the prompt to tests/test_suites/fixtures because its consumer is now the registered nightly driver, and updated that driver to use the test-only path.

tests/functional/grpo_megatron_generation_async_mxfp8.sh
shanmugamr1992: Done in ea0db4f96. The three manual scripts were removed, so the internal cluster name is no longer present in the PR.

tests/test_suites/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation-prefix-caching.sh
```diff
@@ -0,0 +1,62 @@
...
+    $@ \
+    2>&1 | tee $RUN_LOG
+
+# Convert tensorboard logs to json
+uv run tests/json_dump_tb_logs.py $LOG_DIR --output_path $JSON_METRICS
+
+# Only run metrics if the target step is reached
+if [[ $(jq 'to_entries | .[] | select(.key == "train/loss") | .value | keys | map(tonumber) | max' $JSON_METRICS) -ge $MAX_STEPS ]]; then
+    uv run tests/check_metrics.py $JSON_METRICS \
+        'max(data["train/reward"]) > 0.0'
```
terrykong, line 43: [`grpo-nanov3-30BA3B-2n8g-megatron_generation-prefix-caching.sh:42-43`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation-prefix-caching.sh#L42-L43)

**1 action item.**

**TL;DR — this driver checks that the prefix cache was hit, but no longer checks that it did not corrupt logprobs, which is the failure this test exists to catch.**

PR-introduced, in the rewrite from the manual script.

The script this replaced asserted gen/train logprob agreement, with a comment naming the risk:

```bash
# Correctness: gen<->train logprob agreement must stay tight (prefix caching must
# not corrupt logprobs, cf. the vLLM prefix-caching NaN bug).
    'max(data["train/token_mult_prob_error"]) < 1.1'
```

What survived is `max(data["train/reward"]) > 0.0` plus the hit count. A cache that serves wrong KV blocks still produces a nonzero reward and still reports hits, so both checks pass while the thing the test was written for goes undetected.

The four sibling Megatron-generation tests all keep this bound: [`grpo_megatron_generation.sh#L45`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/functional/grpo_megatron_generation.sh#L45), [`_non_colocated.sh#L48`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/functional/grpo_megatron_generation_non_colocated.sh#L48), [`_colocated_reshard.sh#L52`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/functional/grpo_megatron_generation_colocated_reshard.sh#L52), [`_topology.sh#L71`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/functional/grpo_megatron_generation_topology.sh#L71) — all at `< 1.05`.

### AI-1

Add the bound back. `1.1` is your own measured number from the deleted script; use `1.05` if this run matches the siblings.

```suggestion
    uv run tests/check_metrics.py $JSON_METRICS \
        'max(data["train/reward"]) > 0.0' \
        'max(data["train/token_mult_prob_error"]) < 1.1'
```

tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_async_colocated-mxfp8.sh
```diff
@@ -0,0 +1,43 @@
...
+    checkpointing.enabled=False \
+    $@ \
+    2>&1 | tee $RUN_LOG
+
+# Convert tensorboard logs to json
+uv run tests/json_dump_tb_logs.py $LOG_DIR --output_path $JSON_METRICS
+
+uv run tests/check_metrics.py $JSON_METRICS \
+    'max(data["train/reward"]) > 0.0' \
+    'median(data["train/gen_kl_error"]) < 1.3'
```
terrykong, line 43: [`grpo-nanov3-30BA3B-4n4g-megatron_async_colocated-mxfp8.sh:41-43`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_async_colocated-mxfp8.sh#L41-L43)

**1 action item.**

**TL;DR — both assertions here are copied unchanged from the bf16 driver, so turning MXFP8 on is not checked for numerical damage.**

PR-introduced, in the rewrite from the manual script.

These are the same two lines as [`grpo-nanov3-30BA3B-4n4g-megatron_async_colocated.sh#L35-L37`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_async_colocated.sh#L35-L37), which runs in bf16. So the only difference between the two nightlies is the four `fp8_cfg` arguments, and nothing measures their effect.

The script this replaced did measure it, and explained the number:

```bash
# MXFP8 quantizes the TRAINING GEMMs, so the training-side recomputed logprobs
# diverge more from the bf16 generation logprobs than on a pure-bf16 run
# (token_mult_prob_error ~1.3 vs ~1.02 measured on nano-3.5).
    'max(data["train/token_mult_prob_error"]) < 1.5'
```

That is the bound that separates "FP8 is noisier, as expected" from "FP8 broke the run".

### AI-1

Add it back alongside the inherited checks.

```suggestion
uv run tests/check_metrics.py $JSON_METRICS \
    'max(data["train/reward"]) > 0.0' \
    'median(data["train/gen_kl_error"]) < 1.3' \
    'max(data["train/token_mult_prob_error"]) < 1.5'
```

tests/test_suites/disabled.txt
```diff
@@ -1,4 +1,10 @@
 # Tests that are temporarily disabled
+# Generic Nano Megatron drivers are superseded in recurring suites by the
+# explicitly named prefix-caching and MXFP8 variants. Keep these drivers for
+# manual use and for other tests that consume their generic recipe YAMLs.
+tests/test_suites/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation.sh
+tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_async_colocated.sh
```
terrykong: [`tests/test_suites/disabled.txt:2-6`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/disabled.txt#L2-L6)

**1 action item.**

**TL;DR — nano-v3 30B loses its bf16 baselines: after this PR the only nightly runs of non-colocated Megatron generation and of async colocated on that model have prefix caching and MXFP8 turned on, so a failure cannot be told apart from a baseline regression.**

PR-introduced.

The two variants did not get added next to the originals — they replaced them:

```
nightly.txt:126        grpo-nanov3-30BA3B-2n8g-megatron_generation.sh
                    →  grpo-nanov3-30BA3B-2n8g-megatron_generation-prefix-caching.sh
nightly_gb200.txt:98   grpo-nanov3-30BA3B-4n4g-megatron_async_colocated.sh
                    →  grpo-nanov3-30BA3B-4n4g-megatron_async_colocated-mxfp8.sh
disabled.txt        +  both originals
```

Smaller-model coverage is unaffected ([`nightly.txt#L24`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/nightly.txt#L24), [`#L108`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/nightly.txt#L108)), but 30B MoE is where refit, expert parallel and the mamba cache interact, and that is the path that now only runs with the new features on.

The new recipe YAMLs also carry none of the config they are named for — both features are enabled by CLI arguments in the drivers, so each recipe is a copy of the one it replaces with three strings changed:

```
$ diff grpo-nanov3-30BA3B-2n8g-megatron_generation.yaml \
       grpo-nanov3-30BA3B-2n8g-megatron_generation-prefix-caching.yaml
6c6    <   checkpoint_dir: results/grpo-nanov3-30BA3B-2n8g-megatron_generation
       >   checkpoint_dir: results/grpo-nanov3-30BA3B-2n8g-megatron_generation-prefix-caching
47c47  <     name: grpo-nanov3-30BA3B-2n8g-megatron_generation
       >     name: grpo-nanov3-30BA3B-2n8g-megatron_generation-prefix-caching
```

(The MXFP8 pair differs by one comment and three name strings.)

### AI-1

Add the feature arguments to the two existing drivers instead of cloning them, then delete both new `.sh` files, both new recipe YAMLs, and these [`disabled.txt`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/disabled.txt#L5-L6) entries, and restore the original entries in [`nightly.txt`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/nightly.txt#L124) and [`nightly_gb200.txt`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/nightly_gb200.txt#L98). Since both features are pure CLI overrides, that is the two lines already in your prefix-caching driver and the four `fp8_cfg` lines in your MXFP8 driver, appended to the existing ones.

If you would rather keep the variants as separate drivers, then keep the originals in the suites too and drop only these [`disabled.txt`](https://github.com/NVIDIA-NeMo/RL/blob/579539729c96742f49d742b9e1dd372f09866004/tests/test_suites/disabled.txt#L5-L6) lines — but that leaves two duplicate recipe YAMLs that will drift from the ones they were copied from.

Not a `suggestion` block: the change spans four files plus two deletions.

terrykong (dismissed):

shanmugamr1992:
/ok to test [c39f696](https://github.com/NVIDIA-NeMo/RL/pull/3331/commits/c39f6963c650cfd6c2680819c7645db403099d1e)

PR #3339, opened 2026-07-24
build: exclude opencv-python-headless from shipped container

### LINKED ISSUE #3254: Make install_audio_deps.sh robust to torch version upgrades
## Context

`tools/install_audio_deps.sh` currently hardcodes `torchaudio==2.11.0`. This was flagged in [PR #3107](https://github.com/NVIDIA-NeMo/RL/pull/3107#discussion_r2261xxx) by @terrykong — if torch is upgraded in `pyproject.toml`, the script needs a manual matching update.

## Problem

The torchaudio version does not always exactly match the torch version (e.g. torchaudio 2.0.0 requires torch 2.0.1). Deriving the version dynamically from the installed torch at runtime is therefore unreliable.

## Options to explore

1. **Add `torchaudio`/`torchcodec` as an `audio` optional extra in `pyproject.toml`** so `uv lock` resolves and pins the version, and the install script reads it from `pyproject.toml` via `tomllib`. The Dockerfile only syncs named extras (vllm, sglang, mcore, etc.) so the `audio` extra would not be baked into the container.
2. **Hardcode with a prominent comment** cross-referencing the torch pin in `pyproject.toml` and the [torchaudio compatibility table](https://docs.pytorch.org/audio/stable/installation.html).

Option 1 gives a single source of truth; option 2 is simpler with a clear maintenance signal.

## Acceptance criteria

- Bumping `torch` in `pyproject.toml` either automatically installs the correct `torchaudio`, or there is an obvious, hard-to-miss signal that the install script also needs updating.

kajalj22:
## Summary

- `opencv-python-headless` bundles FFmpeg codec libs (`libavcodec`, `libavformat`, etc.) in `opencv_python_headless.libs/`. Its included patent-encumbered codecs (H.264, H.265, AAC) may incur royalties when distributed in a shipped container.
- Mark it `sys_platform == 'never'` in `[tool.uv.override-dependencies]`, keeping the `>=5.0.0` CVE version pin, so it is never baked into the image.
- Add it to `tools/install_audio_deps.sh` so it is installed at test/runtime alongside `torchaudio` and `torchcodec`.

**vllm compatibility:** vllm defaults to opencv for video IO but wraps all three backends (cv2, av, torchcodec) in a `try/except PlaceholderModule` pattern — if cv2 is absent it falls back gracefully to torchcodec, which is already installed by this script. NeMo-RL's data loading explicitly uses `backend="torchcodec"` in `multimodal_utils.py` and `processors.py`.

Closes #3254

## Test plan

- [ ] Nightly suite in progress - Pipeline: 59459207


thomasdhc (dismissed):

terrykong (dismissed):

tools/install_audio_deps.sh
```diff
@@ -26,4 +26,10 @@ uv pip install \
     --reinstall-package torchaudio \
     "torchaudio==2.11.0" \
     "torchcodec>=0.3.0"
+
+# opencv-python-headless is excluded from the shipped container (bundled FFmpeg codec
+# libs incur royalties). vllm defaults to opencv for video IO; install >=5.0.0 which
+# also patches CVE-2025-9951, CVE-2025-1594, and CVE-2024-31582.
+echo "[audio-deps] Installing opencv-python-headless..."
+uv pip install "opencv-python-headless>=5.0.0"
```
yfw: [`tools/install_audio_deps.sh:34`](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/tools/install_audio_deps.sh#L34)

This install is a silent no-op. `uv pip install` walks up from the working directory, finds the repo `pyproject.toml`, and applies the `[tool.uv] override-dependencies` you added one line away — so `sys_platform == 'never'` suppresses the very install meant to compensate for it. It exits 0, so `set -euo pipefail` does not catch it and the script prints `[audio-deps] Done.` having installed nothing.

Reproduced from the repo root at the containers' pinned uv 0.11.18, and again inside a real release image at its `WORKDIR=/opt/nemo-rl`:
```
uv pip install "opencv-python-headless>=5.0.0"  ->  Checked 1 package / Would make no changes   (rc=0)
uv pip install --no-config "..."                ->  Installed opencv-python-headless==5.0.0.93
```
It applies in all four contexts — dev checkout, CI jobs, release container, and `Dockerfile.ngc_pytorch` — since `pyproject.toml` ships at [`/opt/nemo-rl`](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/docker/Dockerfile#L182) and the `--exclude` at [line 366](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/docker/Dockerfile#L365-L366) preserves rather than removes it.

The repo already has the fix and the rationale in-tree, in the sibling script for the `av` exclusion — [`Megatron-Bridge/scripts/install_diffusion_deps.sh`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/a056408d29ad1070fb926512991075998c9e023f/scripts/install_diffusion_deps.sh#L27):
> `--no-config` ignores the project's `[tool.uv]` config, so the `sys_platform == 'never'` override does not neutralize the install.

**Knock-on:** because cv2 never lands, the new guard at [line 12](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/tools/install_audio_deps.sh#L12) can never pass, so every call re-runs `apt-get update && apt-get install ffmpeg` plus the torchaudio reinstall — breaking the "Safe to call multiple times" contract at [line 8](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/tools/install_audio_deps.sh#L8) and hard-failing for non-root users on re-run. The suggestion below resolves that too. (`--no-config` is preferred over `pip install`, which [`build-and-dependency`](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/.claude/skills/build-and-dependency/SKILL.md) prohibits.)

```suggestion
uv pip install --no-config "opencv-python-headless>=5.0.0"
python -c "import cv2" || { echo "[audio-deps] ERROR: cv2 not importable after install" >&2; exit 1; }
```

pyproject.toml
```diff
@@ -347,17 +347,18 @@ default-groups = ["dev", "build"]
...
+# via tools/install_audio_deps.sh. vllm defaults to opencv for video IO but falls back to torchcodec.
 # The timm override is needed because sglang requires timm==1.0.16.
 override-dependencies = [
   "transformer-engine[pytorch,core_cu13] @ git+https://github.com/NVIDIA/TransformerEngine.git@release_v2.15",
   "nvidia-cublas==13.5.1.27; sys_platform != 'darwin'",
   "nvidia-cudnn-cu13==9.20.0.48; sys_platform != 'darwin'",
   "nvidia-cudnn-frontend==1.23.0",
   "nvidia-nvshmem-cu13>=3.6.5; sys_platform != 'darwin'",
-  "opencv-python-headless>=5.0.0",
+  "opencv-python-headless>=5.0.0; sys_platform == 'never'",
```
yfw: [`pyproject.toml:361`](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/pyproject.toml#L361)

`uv.lock` needs regenerating — `uv lock --check` fails on this commit:
```
The lockfile at `uv.lock` needs to be updated, but `--check` was provided.
```
The clearest way to see it: this list already has two never-excluded packages, and the lock records both with markers, while opencv is the only one out of sync —
```
uv.lock:77   { name = "av",                      marker = "sys_platform == 'never'" }
uv.lock:78   { name = "decord2",                 marker = "sys_platform == 'never'" }
uv.lock:92   { name = "opencv-python-headless",  specifier = ">=5.0.0" }        <- no marker
```
A controlled A/B (identical tree and lock, only `pyproject.toml` varying) gives 516 packages / exit 0 on base versus 515 / exit 1 here, so this change is the sole cause.

Impact is wider than the image build: [`docker/Dockerfile:260`](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/docker/Dockerfile#L260) `uv lock --check` and `uv sync --locked` at [:288](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/docker/Dockerfile#L288)/[:307](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/docker/Dockerfile#L307) fail, **and** every Ray worker fails at startup, since all eight executables in [`nemo_rl/distributed/virtual_cluster.py:59-78`](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/nemo_rl/distributed/virtual_cluster.py#L59-L78) use `uv run --locked` — that one reproduces from a plain source checkout with no container involved.

**Separately, on this same line:** once the marker is always-false the `>=5.0.0` never participates in resolution, so the CVE floor is silently dropped rather than kept. The `av` entry a few lines below already splits these two concerns — exclusion in `override-dependencies` with no specifier ([:393](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/pyproject.toml#L393)), CVE floor in `constraint-dependencies` ([`"av>=17.1.0"`, :419](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/pyproject.toml#L419)). Matching that pattern would keep the floor live for anyone who does install it. Suggestion below covers this line; the floor would move to `constraint-dependencies` alongside `av`.

```suggestion
  "opencv-python-headless; sys_platform == 'never'",
```

tools/install_audio_deps.sh
```diff
@@ -26,4 +26,10 @@ uv pip install \
     --reinstall-package torchaudio \
     "torchaudio==2.11.0" \
     "torchcodec>=0.3.0"
+
+# opencv-python-headless is excluded from the shipped container (bundled FFmpeg codec
```
yfw: [`tools/install_audio_deps.sh:30`](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/tools/install_audio_deps.sh#L30)

Worth a look before finalizing: even once `--no-config` is added, this line installs cv2 into the base venv (`/opt/nemo_rl_venv`) — a venv where nothing imports it. The in-container consumers, vllm and `mistral_common`, run in Ray worker venvs, which are created without `--system-site-packages` at [`nemo_rl/utils/venvs.py:80`](https://github.com/NVIDIA-NeMo/RL/blob/a5f8335be1c8407e5fe51008f9bdc0748a56ed93/nemo_rl/utils/venvs.py#L80) and so cannot see the base venv. Verified on two worker venvs built by different uv versions: `include-system-site-packages = false`, no `.pth` bridging back.

The torchaudio/torchcodec lines above are correctly placed, since NeMo-RL's own data loading runs in the driver — that asymmetry is what makes this one stand out.

This has no in-repo trigger today (nothing in `nemo_rl/` imports cv2, and the worker venvs get opencv correctly stripped by the override via `uv sync`), so it may be that the cleanest resolution is dropping the opencv line here rather than adding it. If it is genuinely wanted for vllm, it would need to target the worker venv through the `UV_PROJECT_ENVIRONMENT` mechanism instead. Either way, worth a deliberate decision rather than leaving a line that cannot have an effect.

pyproject.toml
kajalj22: updated the lock file

tools/install_audio_deps.sh
kajalj22: removed open cv as it isn't used, added the --no-config flag to other deps that are installed in a similar manner

yfw (approved):

chtruong814 (approved):

PR #3340, opened 2026-07-24
fix: reserve MOPD teacher clusters before Gym startup

yfw:
# What does this PR do ?

## Summary

Reserve topology-aware MOPD teacher placement groups before NeMo Gym starts allocating GPU-backed services. Teacher workers and model loading remain deferred until policy initialization completes, avoiding checkpoint-conversion races.

## Problem

Previously, teacher topology was planned before Gym startup, but the corresponding placement groups were not reserved until afterward. Gym could consume GPUs in those selected NVLink domains, leaving a teacher placement group permanently unschedulable despite sufficient cluster-wide capacity. This caused setup to time out after some teachers initialized successfully.

## Fix

Split teacher setup into two phases:

1. Reserve all topology-aware teacher placement groups before Gym startup.
2. Initialize teacher workers and load their models after policy initialization.

This preserves topology-aware teacher placement while preventing Gym from taking the reserved resources.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


nemo_rl/algorithms/grpo.py
```diff
@@ -940,6 +940,21 @@ def _spinup_nemo_gym(base_urls, model_name):
...
 
+    # Reserve topology-aware teacher placement groups before NeMo Gym starts
+    # opportunistically placing its GPU-backed services. Worker creation and
+    # model loading remain deferred until the policy is ready to avoid racing
+    # Megatron-Bridge checkpoint conversion.
+    teacher_clusters: dict[str, RayVirtualCluster] = {}
+    teacher_reservation_time = 0.0
+    if enable_opd_teachers:
+        t0 = time.perf_counter()
+        teacher_clusters = opd_module.reserve_teacher_clusters(
```
terrykong, line 918: [`nemo_rl/algorithms/grpo.py:951`](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/algorithms/grpo.py#L951)

Adjacent footgun rather than something this PR broke — but claiming the teachers first means the *policy* now absorbs the error instead of the teacher.

[`TeacherResourceConfig.gpus_per_node`](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/algorithms/opd.py#L51) defaults to `8` and is never checked against `cluster.gpus_per_node`, which is `4` in 42 of the 138 `gpus_per_node` entries under `examples/configs/recipes/llm/`. A teacher block that just omits the field:

```yaml
cluster: {gpus_per_node: 4, num_nodes: 3}    # 12 GPUs total
non_colocated_teachers:
  default_teacher_cfg: {num_nodes: 1}        # gpus_per_node -> 8
```

- [grpo.py:697](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/algorithms/grpo.py#L697) deducts 1 whole node → `policy_nodes = 2`
- the only check is cluster-**total** GPUs ([virtual_cluster.py:852-855](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/distributed/virtual_cluster.py#L852-L855), reading `ray.cluster_resources()`): 8 ≤ 12 passes. There is no per-node check.
- `PACK` is soft — Ray defines it as "packs Bundles into as few nodes as possible", vs `STRICT_PACK`'s "not allowed to span multiple nodes" — so the 8-bundle group at [virtual_cluster.py:893-897](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/distributed/virtual_cluster.py#L893-L897) spreads over 2 nodes and **succeeds**, printing `✓ Reserved teacher 'default_teacher' cluster: 1 node(s), 8 GPUs/node`
- 4 GPUs remain, policy needs 8 → 180s stall → [`TimeoutError`](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/distributed/virtual_cluster.py#L912-L915) that names no cluster

`gpus_per_node` exceeding the node's GPU count is never a valid configuration, so an assert blocks nothing legitimate.

**Action:** add this next to the existing teacher-node accounting at [grpo.py:695-701](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/algorithms/grpo.py#L695-L701), mirroring the precedent at [grpo.py:792-799](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/algorithms/grpo.py#L792-L799):

```python
for tcfg in teacher_configs:
    assert tcfg.gpus_per_node <= cluster_config["gpus_per_node"], (
        f"OPD teacher '{tcfg.alias}' requests gpus_per_node={tcfg.gpus_per_node} > "
        f"cluster.gpus_per_node={cluster_config['gpus_per_node']}; its placement group "
        "cannot fit on one node and will split across nodes reserved for the policy."
    )
    opd_teacher_nodes += tcfg.num_nodes
```

(Defaulting `gpus_per_node` to `cluster.gpus_per_node` instead of `8` would kill the whole class, but that's a bigger change and clearly outside this PR.)

terrykong (commented):
Reviewed at `f39d43b`. The two-phase split looks right to me — claiming the teacher placement groups before Gym at [grpo.py:943-956](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/algorithms/grpo.py#L943-L956) while deferring worker creation to [grpo.py:1447](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/algorithms/grpo.py#L1447) is what the described failure needs, and the ordering claim checks out (Gym spins up at grpo.py:1153 and :1240, both after the new reservation). Making `teacher_clusters` required in f39d43b was a good call — the type signature now carries the reserve-first contract on its own. One polish item below, no blockers.

A few things I checked that came back clean, so they don't get re-litigated:
- Hoisting `RayVirtualCluster` / `prepare_segment_topology` to module scope is cycle-free (`virtual_cluster.py` imports only stdlib + `ray`), and it's what makes the reservation logic unit-testable — keeping the `teacher_worker_group` imports function-local with a why-comment is the right split.
- The eager claim can't change the placement-group *shape*: pre-claiming would permanently block a later unified PG, but teachers only ever go through `RayWorkerGroup`, which requests per-node groups. Safe — just a non-obvious invariant.
- No lifetime bug: `RayWorkerGroup` holds the cluster reference, so the reserved clusters survive `setup()`'s local going out of scope.
- Re-deriving `teacher_configs` for the alias cross-check is worth it rather than duplication — the configs are needed anyway, so the guard is a free set-compare.
- The new tests hold up: the exact-`events`-list assertion is load-bearing rather than brittle (deferring the claims to a second pass after the loop makes it fail), the fake's signature matches the real [`prepare_segment_topology`](https://github.com/NVIDIA-NeMo/RL/blob/f39d43b99be7dc84150ceebd757e5de84312aee4/nemo_rl/distributed/virtual_cluster.py#L541), and they're correctly placed in the Algorithms L0 shard.
- Both `Raises:` blocks in `opd.py` are accurate as written, and `pre-commit run --all-files` passes.

One question out of curiosity: which multi-node MOPD + Gym configuration reproduced the hang? Asking mainly because the `segment_size`-unset path also changes behaviour here (the claim became unconditional) and isn't mentioned in the description — it'd be good to know whether both the pinned and unpinned paths were exercised.

_Generated by Claude Code_

nemo_rl/algorithms/grpo.py
yfw: Addressed in 38e837b794c8e99f5f5eda90291585af6c03df4b

terrykong (approved):

yfw:
> One question out of curiosity: which multi-node MOPD + Gym configuration reproduced the hang? Asking mainly because the `segment_size`-unset path also changes behaviour here (the claim became unconditional) and isn't mentioned in the description — it'd be good to know whether both the pinned and unpinned paths were exercised.

This was a 48-node GB200 ultra-mopd-min run with segment=16. Gym consumed resources before teacher reservation, two teachers initialized, and then the third placement group timed out.

We originally didn't test the `segment_size`-unset path e2e, but the resource-ordering race exists independently of topology pinning. Added a unit test for that path in https://github.com/NVIDIA-NeMo/RL/pull/3340/changes/38e837b794c8e99f5f5eda90291585af6c03df4b#diff-b599a3cdd4218e821fbeb980a4b4efdaae8fb37097029631552b211873551529R477

PR #3345, opened 2026-07-25
feat(mpo): add Nemotron Omni Nano and Super training

jinglinglingling:
## Summary
- add MPO with driver-owned reward-shift state and MMPR multimodal preprocessing
- preserve chosen/rejected pair atomicity in NeMo-RL's Megatron packing pipeline
- add Nano qualification recipes plus a Super no-pack parity recipe
- add resumable Slurm/W&B chains that restore model, optimizer, scheduler, dataloader, and MPO state
- merge current `main` (`3a0a625db`; MBridge `d352aceda`)

## Supported scope
The production recipe targets single-image MMPR training with Nemotron Omni Nano. Nano supports packed and unpacked MPO training.

Super MPO is included only with sequence packing disabled. Super packed configurations are intentionally excluded because their training curves diverged from the matched no-pack baseline; that path is being investigated separately and does not block this no-pack qualification.

Multi-image, video/audio preference data, Super packing, and MTP training require separate qualification.

## Validation
- [x] conflict-free with current `main`
- [x] DCO sign-off
- [x] Ruff, formatting, recipe-minimization, launcher syntax, and diff checks on the previous head
- [x] focused MPO/DPO, MMPR/processor/collator, packing, and resume tests on the previous head
- [x] Super no-pack 2K 100-step parity run: https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpo4a75dd
- [x] Super no-pack legacy baseline: https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/fpieeug0
- [ ] refreshed CI on the current merged head


jinglinglingling:
/ok to test [efe1b3b](https://github.com/NVIDIA-NeMo/RL/pull/3345/commits/efe1b3b7bdee7cd9f89733b79bc9d0146968c3a3)

jinglinglingling:
Rebase verification for the latest #3290 refresh:

- PR base branch head: `07e943233438fd8ff14dbd8ff87cbf7d1db24683`
- first rewritten MPO commit: `81dbbb8df4b9b3bbe2e9f4752bfdf45b71e21ed1`
- parent of that first MPO commit: `07e943233438fd8ff14dbd8ff87cbf7d1db24683`
- current MPO head: `b4b247fd8acc09e0e787ad3dea4e6649a3d24b68`
- MBridge gitlink inherited from #3290: `2f0f8c80e832ad5b4acf9b2b16883aab55923ca9`

A rebase rewrites the existing MPO commit stack rather than adding a separate rebase/merge commit, so the #3290 base commits do not appear in this PR's commit list.

jinglinglingling:
Live parity update (2026-07-31):

- [Nano no-pack](https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpo20761fc0): running step 22/100; latest completed loss `0.8276`
- [Nano packed](https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpo8fe55f1b): running step 26/100; latest completed loss `0.8348`
- [Super packed CP2](https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpoe2d24b8e): segment 1 running step 7/100; latest completed loss `0.8361`

All three runs remain healthy with online W&B logging and no training/runtime errors. Super segments 2 and 3 are queued with `afterok` dependencies for automatic resume in the same W&B run.

jinglinglingling:
Parity progress update:

- ✅ [Nano packed](https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpo8fe55f1b) completed 100/100 (Slurm `14769703`, exit 0). Final train loss/accuracy: `0.7583` / `0.6602`; final validation loss/accuracy: `0.7668` / `0.6465`; step-100 optimizer checkpoint saved successfully.
- ⏳ [Nano no-pack](https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpo20761fc0) is running step 90/100. Latest completed loss/accuracy: `0.7652` / `0.6523`.
- ⏳ [Super packed CP2](https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpoe2d24b8e) segment 1 is running step 33/100. Latest completed loss/accuracy: `0.7988` / `0.7188`; automatic resume segments remain queued.

No NaN, OOM, or runtime errors observed.

jinglinglingling:
/ok to test [efe1b3b](https://github.com/NVIDIA-NeMo/RL/pull/3345/commits/efe1b3b7bdee7cd9f89733b79bc9d0146968c3a3)

jinglinglingling:
/ok to test [c38b23e](https://github.com/NVIDIA-NeMo/RL/pull/3345/commits/c38b23e9db90abb8fc362e8c5cf6bb9b9f4ec953)

jinglinglingling:
/ok to test [51e0641](https://github.com/NVIDIA-NeMo/RL/pull/3345/commits/51e06419161135b93ff1b1b9bb7afc7e2a63de20)

jinglinglingling:
/ok to test [816340c](https://github.com/NVIDIA-NeMo/RL/pull/3345/commits/816340c88)

jinglinglingling:
/ok to test [816340c](https://github.com/NVIDIA-NeMo/RL/pull/3345/commits/816340c88fbe6c2030f0fb2d8d6a9a6c75ff58c9)

jinglinglingling:
/ok to test [1f4fd4e](https://github.com/NVIDIA-NeMo/RL/pull/3345/commits/1f4fd4ed07b0226ef69bd0df5e022f6c2d04dfae)

jinglinglingling:
Rebased the MPO stack onto the latest foundation head `822d7b131` (MBridge pin `573e088c`). New MPO head: `345a8c1c1`. The obsolete optional-Hydra logprobs override was dropped because the foundation now restores `processed_logprobs`; all quality hooks and 24 focused final-head tests pass.

jinglinglingling:
Latest post-rebase validation update:

- Rebased head: `345a8c1c1` on foundation `822d7b131`
- Super packed 16K CP2 smoke: Slurm `15105481`, completed successfully (1 train step + validation + checkpoint), W&B [`svnkqjsn`](https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/svnkqjsn)
- Fresh 100-step parity chain: Slurm `15106168` (steps 1–50) → `15106169` (steps 51–100), one shared W&B run [`mpo71ce51`](https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpo71ce51)

The chain is using the latest rebased NeMo-RL/MBridge dependency state and the rebuilt vLLM 0.25.1 environments. Segment 2 is gated with `afterok` and restores the same checkpoint/W&B identity.

jinglinglingling:
Done — I removed draft mode and rebased the MPO stack directly onto the latest `main` (`ae07eafe8`). The conflicts are resolved and the PR is now mergeable; the new head is `c71ee5c29`. I also corrected the chained-resume scheduler horizon mismatch found during the latest Super run. Fresh CI is in progress.

jinglinglingling:
Rebased once more onto latest `main` (`af451b1a1`) and force-updated the PR head to `46356ccba`. Conflicts with the new multimodal dedup path and MBridge/config changes are resolved. Local Ruff/config checks and 72 focused MPO/data/packing tests passed; fresh CI is running. The PR is mergeable and ready for review again.

jinglinglingling:
`main` advanced again overnight, so I rebased once more onto `2c5ee18f4` and updated the PR head to `0fdd14775`. The PR is conflict-free and mergeable. All checks passed except a transient GitHub HTTP/2 failure while fetching the flash-attn wheel in `uv lock check`; that failed job has been rerun and is in progress.

jinglinglingling (commented):

Ran the upstream multi-agent self-review workflow against head `1ca5ab86191e12a2ab14cc7d092809242fae776e` after rebasing onto the latest `main`.

### Review process

- Parallel reviews covered RL/MPO correctness, multimodal Megatron integration, general bug scanning, tests, design, documentation, and existing PR discussions.
- An independent devil's-advocate pass challenged and filtered every candidate finding.
- The initial pass found high-confidence issues in MMPR cache invalidation and response normalization, image-placeholder handling, optimizer-state detection, chained-run defaults, qualification documentation, and functional assertions.
- These were fixed in `1ca5ab86191e12a2ab14cc7d092809242fae776e` with regression coverage.
- A second independent review found one production MMPR compatibility regression in the first fix; that was corrected and re-reviewed.

### Final result

**LGTM from self-review:** no remaining findings at or above the workflow's 80% confidence threshold.

### Validation

- Full repository pre-commit suite: passed
- Ruff and formatting: passed
- Pyrefly: passed
- Recipe minimization: passed
- Shell syntax and `git diff --check`: passed
- GitHub checks currently triggered for this head: passed

Local pytest execution was not available on this host because the Python runtime exited with SIGBUS; GPU/runtime qualification was therefore not rerun as part of this self-review. The PR documentation now distinguishes implemented support from final-head production qualification.

_Generated using the upstream `review-pr-team` workflow._

nemo_rl/utils/checkpoint.py
```diff
@@ -61,6 +61,26 @@ def _load_megatron_common_state_dict(iteration_dir: Path) -> dict[str, Any]:
...
+    except ImportError as error:
+        raise RuntimeError(
+            "Megatron-Core is required to inspect optimizer state in the distributed "
+            f"checkpoint at {iteration_dir}. Install NeMo-RL with the `mcore` extra."
+        ) from error
+
+    return {str(key) for key in load_sharded_metadata(str(iteration_dir))}
+
+
+def _is_megatron_optimizer_key(key: str) -> bool:
```
aroshanghias-nvd, line 79: **1 action item.**

**TL;DR — `_is_megatron_optimizer_key` never matches a real MoE checkpoint key, so every Nemotron Omni resume silently reinitializes the optimizer.** PR-introduced in `1ca5ab86`; a regression against this PR's own previous head `0fdd1477`.

[`nemo_rl/utils/checkpoint.py:79`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/utils/checkpoint.py#L79-L81)

First — the instinct here is right, and worth saying: replacing the `run_config.yaml` / `save_optim` heuristic (which only recorded *configuration intent*) with a probe of what the checkpoint actually contains is the better design. The problem is only the predicate.

**Background.** MCore's `ChainedOptimizer.sharded_state_dict` prefixes **every** sharded key with `chained_<idx>.` whenever the model has more than one sub-optimizer. Any MoE model has at least two (dense + expert), and both recipes this PR ships are MoE — Nano 30B-A3B and Super 120B-A12B.

**What breaks**, on a checkpoint produced by this stack:

```
$ # keys in step_2/policy/weights/iter_0000000/.metadata
chained_N.optimizer keys: 64     bare optimizer.* keys: 0
chained_1.optimizer.distributed.dp_group_idx_5.gbuf_idx_0.dtype_(torch.bfloat16, torch.bfloat16)...
```

1. `any(_is_megatron_optimizer_key(key) for key in sharded_keys)` → `False` (nothing starts with `optimizer`)
2. control falls through to the `warnings.warn` and returns `optimizer_path = None`
3. [`nemo_rl/models/megatron/setup.py:1346`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/models/megatron/setup.py#L1346) turns that into `load_optim=False`
4. **Symptom:** the resume succeeds, but Adam moments and the LR schedule are silently reinitialized — no error, just a training-curve discontinuity at every segment boundary.

That is the exact failure this commit set out to remove, and it lands on the chained-resume path. Note it is also a regression versus `0fdd1477`: those checkpoints carry `save_optim: true` in `run_config.yaml`, so the old heuristic returned the optimizer path correctly.

**Action:** strip the chained prefix before testing. `re` is already imported at `checkpoint.py:22`.

```suggestion
def _is_megatron_optimizer_key(key: str) -> bool:
    """Return whether a flattened Megatron checkpoint key belongs to the optimizer."""
    # ChainedOptimizer prepends ``chained_<idx>.`` to every sharded key when the
    # model has more than one sub-optimizer, which is always the case for MoE
    # models (dense + expert). Both Nemotron Omni recipes take that path.
    key = re.sub(r"^chained_\d+\.", "", key)
    return key == "optimizer" or key.startswith(("optimizer.", "optimizer/"))
```

Please also build the regression test from a real key string rather than the synthetic `"optimizer.state.exp_avg"` the new tests mock in — that shape does not occur in any checkpoint this stack writes, which is why the added coverage passes while the production path fails.

nemo_rl/utils/checkpoint.py
```diff
@@ -228,6 +248,14 @@ def get_resume_paths(
                     # state from the weights_path.
                     return weights_path, optimizer_path
 
+                # Modern torch_dist checkpoints shard optimizer tensors separately
+                # from the common state. Inspect stored keys rather than trusting
+                # run_config.save_optim, which only records configuration intent.
+                if (iteration_dir / "metadata.json").exists():
+                    sharded_keys = _load_megatron_sharded_metadata_keys(iteration_dir)
```
aroshanghias-nvd, line 298: **2 action items.**

**TL;DR — this call makes a pre-existing `main` test fail, and because `addopts` contains `-x` that failure aborts the file, so the two tests this commit adds never execute in CI.** PR-introduced in `1ca5ab86`.

[`nemo_rl/utils/checkpoint.py:254`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/utils/checkpoint.py#L254-L256)

`test_get_resume_paths_warns_when_megatron_optimizer_missing` is **not** added by this PR — it exists at the merge base (`git show 85e02cca:tests/unit/utils/test_checkpoint.py` → line 489). It creates `iter_0000000/metadata.json` and mocks `_load_megatron_common_state_dict`, then expects a `UserWarning`. The new block calls the *un*-mocked `_load_megatron_sharded_metadata_keys` on that fixture.

A/B on the same file, same command, same container:

```
NEW head 1ca5ab86:  1 failed, 71 passed
OLD head 0fdd1477:  71 passed
```
```
nemo_rl/utils/checkpoint.py:255: in get_resume_paths
    sharded_keys = _load_megatron_sharded_metadata_keys(iteration_dir)
E   RuntimeError: Megatron-Core is required to inspect optimizer state ...
E   Failed: DID NOT WARN.
```

This is not environment-specific: with mcore installed, `load_sharded_metadata` → `check_is_distributed_checkpoint` on a directory holding a 0-byte `metadata.json` and no `.metadata` raises `CheckpointingException` instead, so the expected warning is still never emitted.

**Second-order:** `pyproject.toml:650` sets `addopts = "--durations=100 -s -rA -x"`. With `-x`, this failure aborts the file at line 489 — so `test_get_resume_paths_torch_dist_megatron_optimizer` and `test_get_resume_paths_ignores_save_optim_without_optimizer_shards` **never run**. Combined with the sibling comment on line 79, the net coverage of the new predicate is zero.

**AI-1:** add the missing mock to the pre-existing test in `tests/unit/utils/test_checkpoint.py` (around line 489):

```python
    monkeypatch.setattr(
        checkpoint_module,
        "_load_megatron_sharded_metadata_keys",
        MagicMock(return_value=set()),
    )
```

**AI-2:** this is worth a `/ok to test 1ca5ab86191e12a2ab14cc7d092809242fae776e` before merge — no unit or functional job has run at this head (the checks currently attached are all lightweight gating jobs), so this failure is invisible to the PR right now.

tests/unit/algorithms/test_mpo_loss.py
```diff
@@ -0,0 +1,138 @@
...
+        data=data,
+        global_valid_seqs=data["sample_mask"].sum(),
+        global_valid_toks=data["token_mask"][:, 1:].sum(),
+    )
+
+
+def test_mpo_loss_is_invariant_to_pair_row_order():
+    logprobs, data = _mpo_batch()
+    loss_fn = MPOLossFn(_loss_config())
+    reordered_loss, reordered_metrics = _call_loss(loss_fn, logprobs, data)
```
aroshanghias-nvd, line 76: **1 action item.**

**TL;DR — none of the MPO loss arithmetic is pinned: 6 of 8 semantic mutations to `MPOLossFn` leave all 6 tests in this file green, including flipping the BCO sign and deleting the reward-shift term.** PR-introduced.

[`tests/unit/algorithms/test_mpo_loss.py:73`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/tests/unit/algorithms/test_mpo_loss.py#L73-L99)

Credit first, because two of these tests genuinely have teeth: `test_mpo_loss_is_invariant_to_pair_row_order` and `test_mpo_rejects_malformed_pair_metadata` both fail when `split_output_tensor` is made to ignore pair metadata, and `test_mpo_reward_shift_updates_once_from_global_statistics` fails when the EMA is replaced by a raw batch mean. The pair plumbing and the reward-shift lifecycle are covered. The gap is specifically the *arithmetic*.

The file never compares `loss`, `preference_loss`, `bco_loss`, `sft_loss` or `accuracy` to an independently derived value. The order-invariance test compares the loss **to itself** under a permutation, and the only pinned scalars — `bco_reward_sum == -10.0`, `bco_reward_count == 4.0` — are the sufficient statistics fed to the EMA, computed before the terms are weighted and summed, so they are independent of every term in `total_loss`.

Mutation results against these 6 tests:

| mutation to `MPOLossFn` | result |
|---|---|
| BCO term sign flipped in `total_loss` | **all 6 pass** |
| reward-shift term deleted (`shift → 0`) | **all 6 pass** |
| SFT term deleted (`0.0 * sft_loss`) | **all 6 pass** |
| `rewards_delta = rejected - chosen` | **all 6 pass** |
| β dropped from the quality reward | **all 6 pass** |
| β dropped from the preference term | **all 6 pass** |
| `2.0 * total_loss` | **all 6 pass** |
| SFT computed on the rejected row | **all 6 pass** |

A cheaper way to state the same thing: injecting an unconditional `raise` into `MPOLossFn.__call__` fails only **3 of the 6** tests — half the file never invokes the loss forward at all.

For contrast, the sibling `test_dpo_loss` at `tests/unit/algorithms/test_loss_functions.py:196` pins `-logsigmoid(0)` and a closed-form SFT term. `MPOLossFn.input_type = LossInputType.LOGPROB`, so a closed-form test needs no GPU.

**Action:** add a closed-form test. Verified passing at this head, and it fails under all six surviving mutants. β is set to 0.5 deliberately — at β = 1.0 the two β-dropping mutants are undetectable. It reuses this file's existing `_mpo_batch` / `_loss_config` / `_call_loss` helpers.

```python
def test_mpo_loss_matches_closed_form_components():
    """Pin every MPO term to a hand-derived expectation, not to the implementation.

    Rows are (rejected p1, chosen p0, rejected p0, chosen p1) with two unmasked
    tokens each and a zero reference, so per-sequence logratios are -4, -1, -3, -2.
    Pair-aligned that is chosen=(-1, -2), rejected=(-3, -4).
    """
    logsigmoid = torch.nn.functional.logsigmoid
    logprobs, data = _mpo_batch()
    beta = 0.5
    loss_fn = MPOLossFn(_loss_config(reference_policy_kl_penalty=beta))

    loss, metrics = _call_loss(loss_fn, logprobs, data)

    chosen = torch.tensor([-1.0, -2.0])
    rejected = torch.tensor([-3.0, -4.0])
    pairs, shift = 2.0, 1.0

    expected_preference = (-logsigmoid(beta * (chosen - rejected))).sum() / pairs
    expected_sft = torch.tensor([1.0, 2.0]).sum() / pairs
    expected_bco = (-logsigmoid(beta * chosen - shift)).sum() / pairs + (
        -logsigmoid(-(beta * rejected - shift))
    ).sum() / pairs
    expected_total = 1.0 * expected_preference + 0.25 * expected_sft + 0.5 * expected_bco

    torch.testing.assert_close(loss, expected_total)
    assert metrics["preference_loss"] == pytest.approx(expected_preference.item())
    assert metrics["sft_loss"] == pytest.approx(expected_sft.item())
    assert metrics["bco_loss"] == pytest.approx(expected_bco.item())
    assert metrics["accuracy"] == pytest.approx(1.0)
    assert metrics["rewards_chosen_mean"] == pytest.approx(-1.5)
    assert metrics["rewards_rejected_mean"] == pytest.approx(-3.5)
```

To be clear about scope: the implementation itself checks out — an independently written reference reproduces the loss exactly (`0.74280560` vs `0.74280560`) with every component metric matching. This is a coverage finding, not a correctness one.

nemo_rl/algorithms/mpo.py
```diff
@@ -0,0 +1,227 @@
...
+    if configured_grouping not in (None, "pair_index"):
+        raise ValueError(
+            "MPO requires policy.sequence_packing.pair_grouping_key='pair_index'; "
+            f"got {configured_grouping!r}."
+        )
+    packing_config["pair_grouping_key"] = "pair_index"
+    # This grouping is applied before NeMo-RL's Megatron data path packs each
+    # microbatch into full THD input. Keeping one atomic pair per microbatch
+    # both preserves chosen/rejected alignment and bounds vision-encoder memory.
+    packing_config.setdefault("max_sequences_per_bin", 1)
```
aroshanghias-nvd: **1 action item, 2 follow-ups.**

**TL;DR — with `max_sequences_per_bin=1`, any `train_global_batch_size` not divisible by the DP degree crashes at step 1 with an error naming two things that are not config keys.** PR-introduced.

[`nemo_rl/algorithms/mpo.py:108`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/mpo.py#L108-L111)

**Walk the failure** — 2 nodes × 8 GPUs, `TP=8, PP=1, CP=1` → `dp_size=2`, packing enabled as every shipped MPO recipe does:

```yaml
policy:
  train_global_batch_size: 3
  sequence_packing: {enabled: true, pair_grouping_key: pair_index, max_sequences_per_bin: 1}
```

1. `dpo_train` calls `policy.train(..., gbs=train_global_batch_size * 2)` = 6 rows — [`nemo_rl/algorithms/dpo.py:640`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/dpo.py#L640)
2. `shard_by_batch_size` **succeeds when it shouldn't**: its only divisibility guard is on rows, `6 % 2 == 0` — [`nemo_rl/distributed/batched_data_dict.py:543`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/distributed/batched_data_dict.py#L543)
3. pair grouping collapses 6 rows to 3 atomic groups; the cap of 1 forces exactly 3 bins — [`nemo_rl/distributed/batched_data_dict.py:606`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/distributed/batched_data_dict.py#L606)
4. `_adjust_bin_count` must round 3 up to a multiple of `shards=2` → 4, but every bin already holds its cap — [`nemo_rl/data/packing/algorithms.py:132`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/data/packing/algorithms.py#L132)
5. **Symptom:** `ValueError: Cannot create 4 bins with only 3 sequences. Each bin must contain at least one sequence. Either reduce min_bin_count/bin_count_multiple or provide more sequences.` — neither `min_bin_count` nor `bin_count_multiple` is a user-facing key; both are derived from `dp_size`.

Executed across a `(pairs, dp)` grid — `(3,2)`, `(5,2)`, `(6,4)`, `(12,8)` all raise; with the cap removed every one passes, which also shows pair grouping alone already guarantees atomicity. The shipped recipes happen to satisfy `gbs % dp == 0` (256 pairs at dp=1, 256 at dp=2), but `scripts/vlm_mpo.sh` exposes `MPO_TRAIN_GLOBAL_BATCH_SIZE` and the guide tells users to change it. Validation self-heals because `maybe_pad_last_batch` pads to a multiple of `val_mbs * dp`; training never pads.

**Action:** add a setup-time check that `policy.train_global_batch_size` is divisible by the DP degree when the cap is 1, naming both the config key and the cluster shape. A `gbs` indivisible by DP is never valid once each pair is one bin, so this rejects nothing legitimate.

Two follow-ups, same root cause, no action needed in this PR if you'd rather track them:

- **`train_mb_tokens` is superseded.** `vlm_mpo-nemotron-super-omni-packed-2k-parity.yaml` asks for `train_mb_tokens: 65536` (~16 pairs/microbatch); the cap forces one pair, a 16× inflation the config does not reveal. A one-line `warnings.warn` naming both values would make the effective microbatch size visible.
- **The requested packer is silently substituted.** All six MPO YAMLs ask for `modified_first_fit_decreasing`, and the cap routes every one of them to `FirstFitDecreasingPacker` instead — [`nemo_rl/data/packing/algorithms.py:557`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/data/packing/algorithms.py#L557-L562). The reasoning in the comment there is sound; the silence is the problem, since any future packing-efficiency comparison starts from a false premise.

nemo_rl/models/megatron/data.py
```diff
@@ -272,6 +272,19 @@ def process_microbatch(
     ctx = straggler_timer(bdata=True) if straggler_timer is not None else nullcontext()
     with ctx:
         input_ids = data_dict["input_ids"]
+        if (
+            not pack_sequences
+            and model_slices_context_parallel_inputs
+            and "pixel_values" in data_dict
+        ):
+            raise NotImplementedError(
```
aroshanghias-nvd: **2 action items.**

**TL;DR — this guard's stated cause cannot be reproduced against the pinned Megatron-Bridge, and it cites a PR that was closed unmerged a year's worth of pins ago.** PR-introduced.

[`nemo_rl/models/megatron/data.py:275`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/models/megatron/data.py#L275-L287)

**The cited fix will never land.** MBridge **#5181** is `state=closed, merged=false` (2026-08-09), superseded by **#5304** — which is an ancestor of this pin *and* of the previous one. It is referenced in three places plus the tracking issue: the message above, [`docs/guides/nemotron-omni-mpo.md:98`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/docs/guides/nemotron-omni-mpo.md#L98), and `tests/unit/models/megatron/test_megatron_data.py:191` (`match=r"Megatron-Bridge#5181"`).

**The stated failure mode does not occur.** At pin `8c46dc42`, `forward` only adopts `attention_mask` as a validity mask when the ranks match:

```python
if media_token_validity_mask is None:
    if padding_mask is not None:
        media_token_validity_mask = ~padding_mask
    elif attention_mask is not None and attention_mask.dim() == input_ids.dim():
        media_token_validity_mask = attention_mask
```
[`modeling_nemotron_omni.py#L706-L710`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L706-L710)

NeMo-RL's dense path supplies a 4-D mask and no `padding_mask`, so this resolves to `None`, and `_merge_projected_media`'s shape check is behind `if attention_mask is not None` ([#L297-L306](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L297-L306)). I extracted both regions from the pinned checkout and executed them on CPU tensors:

```
mask = None  (all-image batch)      -> OK, merge succeeds
4-D MCore causal mask (explicit)    -> ValueError  <- the failure this message describes
explicit [B,S] bool mask            -> OK
```

Only the middle case raises, and NeMo-RL never constructs it. Upstream asserts the same contract in `test_image_forward_does_not_use_mcore_causal_mask_as_token_validity` — and `git show 0c565c9a:` confirms **that test was already present at the previous pin**, so this was never a pin-version problem.

One clarification in case it comes up: RL#3631's `media_token_validity_mask` is *not* what unblocks this. `build_media_token_validity_mask` returns `None` unless some row carries zero media, so an ordinary all-image unpacked batch attaches nothing and takes the dim-check path above.

**AI-1 (this PR):** whichever route you take below, drop the `#5181` reference from all three sites and from #3525's completion criteria, and reword the message to name the condition (`models that slice context-parallel inputs`) rather than "Nemotron Omni" — it keys on a model-advertised flag, so it will fire for the next model that sets it.

**AI-2 (this PR):** pick one —
- **(a)** drop the guard and run `examples/configs/experiments/vlm_mpo-nemotron-omni-nopack-legacy-parity.yaml` — which this PR ships purely as a dead control ("Retain this control for requalification after NVIDIA-NeMo/RL#3525") — as the smoke that closes #3525; or
- **(b)** keep the gate for qualification reasons and state that as the reason:

```suggestion
            raise NotImplementedError(
                "Nemotron Omni image batches without sequence packing are not "
                "qualified: the pinned Megatron-Bridge merges media against the "
                "dense 4D decoder mask correctly, but NeMo-RL has no parity run "
                "for the unpacked path. Enable policy.sequence_packing.enabled=true. "
                "See NVIDIA-NeMo/RL#3525."
            )
```

Stating my confidence plainly: **high** that the cited cause is gone (executed against the pinned source, corroborated by an upstream test). **Not asserted** that the whole unpacked path is green end-to-end — nobody has run it, which is exactly what (a) would settle.

nemo_rl/algorithms/mpo.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def _configure_pair_safe_packing(policy_config: PolicyConfig) -> None:
+    if policy_config["dynamic_batching"]["enabled"]:
+        raise ValueError("Dynamic batching is not supported with MPO.")
+    if not policy_config["sequence_packing"]["enabled"]:
+        return
+
+    packing_config = cast(dict[str, Any], policy_config["sequence_packing"])
+    packing_config["fuse_loss"] = True
```
aroshanghias-nvd: **1 action item.**

[`nemo_rl/algorithms/mpo.py:93`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/mpo.py#L93-L111)

`_configure_pair_safe_packing` applies three different conflict policies to three adjacent keys, and the middle one is silent. PR-introduced.

| key | behaviour when the user set it |
|---|---|
| `pair_grouping_key` | **raises** with a clear message (line 103) |
| `fuse_loss` | **silently force-writes `True`** (line 100) |
| `max_sequences_per_bin` | `setdefault(..., 1)` — user-overridable (line 111) |

A user who writes `policy.sequence_packing.fuse_loss: false` gets `true` with no message, and that flag selects a materially different loss path ([`nemo_rl/models/megatron/train.py:451`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/models/megatron/train.py#L451-L470)).

**There is also no backend guard.** `fuse_loss` is read *only* by the Megatron worker, so with `policy.dtensor_cfg.enabled=true` it is dropped silently and the non-fused `SequencePackingLossWrapper` calls the loss once per sequence ([`nemo_rl/algorithms/loss/wrapper.py:91`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/loss/wrapper.py#L91)) — which for a preference batch fails at the first microbatch with `ValueError: Preference batch must contain exactly one chosen and one rejected row for every pair_index.`, pointing nowhere near the backend.

All six shipped MPO YAMLs already set these keys explicitly, so the mutation buys nothing today beyond a divergence between the config the user wrote, the config that gets logged, and the config that runs.

**Action:** validate uniformly instead of mutating, and add the backend check. MPO's packed path needs `prepare_packed_loss_input`, which itself asserts a Megatron TP group ([`nemo_rl/algorithms/loss/utils.py:307`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/loss/utils.py#L307)), so rejecting DTensor+packing blocks nothing legitimate. One way to do it, of several:

```python
def _validate_pair_safe_packing(policy_config: PolicyConfig) -> None:
    if policy_config["dynamic_batching"]["enabled"]:
        raise ValueError("Dynamic batching is not supported with MPO.")
    if not policy_config["sequence_packing"]["enabled"]:
        return
    if not policy_config["megatron_cfg"]["enabled"]:
        raise ValueError(
            "MPO with sequence_packing requires the Megatron backend; the fused "
            "packed-loss path is not implemented for DTensor. Set "
            "policy.megatron_cfg.enabled=true or disable sequence packing."
        )
    packing = policy_config["sequence_packing"]
    for key, required in (("pair_grouping_key", "pair_index"), ("fuse_loss", True)):
        if packing.get(key) != required:
            raise ValueError(
                f"MPO requires policy.sequence_packing.{key}={required!r} to keep each "
                f"chosen/rejected pair atomic; got {packing.get(key)!r}."
            )
```

nemo_rl/algorithms/dpo.py
```diff
@@ -192,7 +199,18 @@ def setup(
     checkpointer = CheckpointManager(checkpointing_config)
     last_checkpoint_path = checkpointer.get_latest_checkpoint_path()
     loaded_state = checkpointer.load_training_info(last_checkpoint_path)
-    dpo_save_state = _get_dpo_save_state(loaded_state)
+    if loaded_state is not None:
+        # Filter to only known save-state fields; checkpoints may carry
+        # extra keys (e.g. validation metrics from previous runs).
+        known_fields = {f.name for f in fields(save_state_cls)}
```
aroshanghias-nvd: **1 action item.**

[`nemo_rl/algorithms/dpo.py:202`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/dpo.py#L202-L213)

This PR replaced the call to `_get_dpo_save_state` with inline reconstruction, but left the helper in the file. Grep confirms it now has **zero production callers** — only `tests/unit/algorithms/test_dpo.py` (lines 25, 35, 46). So `test_get_dpo_save_state_handles_legacy_checkpoint_and_filters_metrics` tests dead code, while the restore path that actually runs has no unit coverage. PR-introduced.

The two are also not equivalent. The helper seeded from `vars(_initial_dpo_save_state())` and *overlaid* known keys, explicitly "so partial/legacy checkpoints remain loadable" ([`nemo_rl/algorithms/dpo.py:67`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/dpo.py#L67-L69)). The inline version does `save_state_cls(**filtered)` with a one-off `setdefault` for `total_valid_tokens`, and `DPOSaveState` has no field defaults — so a `training_info.json` missing any other field now raises `TypeError: DPOSaveState.__init__() missing 1 required positional argument` instead of defaulting. The PR's own test asserts `_get_dpo_save_state({}) == _initial_dpo_save_state()`, i.e. an empty dict is a supported input the production path no longer handles.

This also matters for MPO specifically: `MPOSaveState.reward_shift` is what reseeds the loss on resume, and the save→restore→`_make_loss_fn` handoff has no test at all. Break any link and the BCO shift silently resets to the config default at every chained-segment boundary — a discontinuity that looks exactly like ordinary EMA drift on the W&B curve.

**Action:** generalize the existing helper and call it, so the tested path is the live path and the dead-code/dead-test pair goes away:

```python
def _get_dpo_save_state(
    loaded_state: Optional[dict[str, Any]],
    save_state_cls: type = DPOSaveState,
    initial_save_state_fn: Callable[[], Any] = _initial_dpo_save_state,
) -> Any:
    if loaded_state is None:
        return initial_save_state_fn()
    known_fields = {field.name for field in fields(save_state_cls)}
    state_values = vars(initial_save_state_fn()).copy()
    state_values.update(
        {key: value for key, value in loaded_state.items() if key in known_fields}
    )
    return save_state_cls(**state_values)
```

then `dpo_save_state = _get_dpo_save_state(loaded_state, save_state_cls, initial_save_state_fn)` here, and extend the existing test with an MPO case asserting `reward_shift` survives a round trip.

nemo_rl/algorithms/dpo.py
```diff
@@ -553,6 +580,7 @@ def dpo_train(
     val_at_start = dpo_config.val_at_start
     val_at_end = dpo_config.val_at_end
     max_num_epochs = dpo_config.max_num_epochs
+    stop_after_step = getattr(dpo_config, "stop_after_step", None)
```
aroshanghias-nvd: **1 action item.**

[`nemo_rl/algorithms/dpo.py:583`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/dpo.py#L583)

`stop_after_step` is declared only on `MPOConfig` ([`nemo_rl/algorithms/mpo.py:50`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/mpo.py#L50)), and `dpo_train` sniffs for it with `getattr`. `AGENTS.md` names this pattern directly: *"side-channel/out-of-band state (e.g. setting an attribute the caller reads via `hasattr`)"*. PR-introduced.

The consequence is proved by this PR's own new test: `tests/unit/algorithms/test_dpo.py:274` does `DPOConfig.model_validate({..., "stop_after_step": 3})`, which works because `DPOConfig` is `extra="allow"`. So **`dpo.stop_after_step` is a live, undeclared, undocumented DPO feature today** — not in `DPOConfig`, not in `examples/configs/dpo.yaml`, and unvalidated, because `_validate_stop_after_step` ([`nemo_rl/algorithms/mpo.py:114`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/mpo.py#L114-L121)) is only called from `mpo.setup`.

Worst case for a plain DPO user: `dpo.stop_after_step: 0` is picked up by the `getattr`, `0 >= 0` is true at [`nemo_rl/algorithms/dpo.py:606`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/dpo.py#L606), and the run prints `Configured stop-after step 0 was already reached; no additional training is needed` and **exits 0 without training**.

**Action:** declare the field where it is already reachable and delete the sniff.

```suggestion
    stop_after_step = dpo_config.stop_after_step
```

with `stop_after_step: int | None = None` added to `DPOConfig` (then drop the override in `MPOConfig`), `_validate_stop_after_step` moved into `dpo.setup()` so both algorithms get the bounds check, and the key documented in `examples/configs/dpo.yaml`.

Untested alongside it: the resume branch at [`nemo_rl/algorithms/dpo.py:606`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/dpo.py#L606) — the one every re-queued segment takes. `test_configured_stop_forces_checkpoint_without_shortening_run` covers the forward direction only, so a regression there would silently retrain steps a predecessor already committed.

examples/configs/recipes/vlm/vlm_mpo-nemotron-omni-30ba3b-mmpr-1n8g-megatron-tp8.v1.yaml
```diff
@@ -0,0 +1,89 @@
+defaults: ../../dpo.yaml
```
aroshanghias-nvd: **1 action item.**

**TL;DR — there is no MPO exemplar config, so this recipe inherits `dpo.yaml`, and a derived recipe that forgets one line will train on MMPR images while validating on HelpSteer3 text, silently.** PR-introduced.

[`examples/configs/recipes/vlm/vlm_mpo-nemotron-omni-30ba3b-mmpr-1n8g-megatron-tp8.v1.yaml:1`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/examples/configs/recipes/vlm/vlm_mpo-nemotron-omni-30ba3b-mmpr-1n8g-megatron-tp8.v1.yaml#L1)

Three facts that only bite in combination:

1. This recipe declares `defaults: ../../dpo.yaml`, and `tests/unit/test_recipes_and_test_suites.py:45` registers `"vlm_mpo": "examples/configs/dpo.yaml"`.
2. `examples/configs/dpo.yaml` puts every algorithm default under **`dpo:`**; MPO reads **`mpo:`**. It also sets `validation: {dataset_name: HelpSteer3}`, which this recipe has to cancel with an explicit `validation: null`.
3. `mpo.MasterConfig` is `extra="ignore"` ([`nemo_rl/algorithms/mpo.py:54`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/mpo.py#L54)), diverging from every other `MasterConfig` in the repo — all nine others are `extra="allow"`.

So the inherited `dpo:` block lands in the merged config and is silently dropped, and any top-level typo in any MPO config is silently ignored rather than rejected. **The concrete failure:** the next MPO recipe (Super, per this PR's stated scope) omits `validation: null` assuming the exemplar handles it, and gets HelpSteer3 validation against MMPR training with no error. The adjacent overwrite at [`nemo_rl/data/utils.py:400`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/data/utils.py#L400) — where an explicit `validation` block replaces `val_dataset["default"]` with no warning — makes that path quiet in both directions.

Every other algorithm family in the repo has an exemplar (`sft.yaml`, `dpo.yaml`, `rm.yaml`, `grpo_math_1B.yaml`, `vlm_grpo_3B.yaml`). MPO is the only one without.

**Action:** add `examples/configs/vlm_mpo.yaml` — a commented exemplar carrying the full `mpo:` block and the three new `policy.sequence_packing.*` keys, mirroring `dpo.yaml`'s layout. Then point this recipe's `defaults:` at it, re-run `uv run ./tools/config_cli.py minimize`, and update `ALGO_MAPPING_TO_BASE_YAML`.

Sequence matters: switch `mpo.MasterConfig` to `extra="allow"` **after** the exemplar exists, not before — with no exemplar the orphan `dpo:` block would land in `__pydantic_extra__` and show up in `model_dump()`, which `examples/run_vlm_mpo.py:61` pprints into the logger and run config.

While you're there: `examples/run_vlm_mpo.py:46-53` defaults `--config` to this recipe rather than to an exemplar, unlike `run_dpo.py`. The bare command therefore tries to load a 30B checkpoint with `data_path: REPLACE_WITH_MMPR_METADATA_PATH` — it is the first thing a new user hits.

examples/configs/experiments/vlm_mpo-nemotron-super-omni-nopack-2k-parity.yaml
```diff
@@ -0,0 +1,27 @@
+# Matched Super MPO control based on legacy W&B run fpieeug0:
+#   nvidia/vlm-mpo-dev-mmpr/super-NOPACK-2k-100step
+#
+# The 2K context is intentional: it is the longest legacy Super no-packing
+# control that completed on 4 nodes. Packing changes only in the sibling config.
+defaults: vlm_mpo-nemotron-omni-nopack-legacy-parity.yaml
+
+policy:
+  model_name: /lustre/fsw/portfolios/llmservice/users/smohsenitahe/checkpoints/super_sft_16k
```
aroshanghias-nvd: **1 action item.**

[`examples/configs/experiments/vlm_mpo-nemotron-super-omni-nopack-2k-parity.yaml:9`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/examples/configs/experiments/vlm_mpo-nemotron-super-omni-nopack-2k-parity.yaml#L9)

`examples/configs/experiments/` is a new config tier this PR creates, and `tests/unit/test_recipes_and_test_suites.py` scans `examples/configs/recipes/` only — so these five files escape the `defaults`-resolution check, the schema check, and the minimization check every recipe must pass. They will rot silently. PR-introduced.

Three concrete problems in them:

- This line pins `model_name: /lustre/fsw/portfolios/llmservice/users/smohsenitahe/checkpoints/super_sft_16k` — a second contributor's scratch path, inherited by both Super configs.
- `vlm_mpo-nemotron-omni-nopack-legacy-parity.yaml:115` pins another user's Lustre `data_path`, and `:140` pins a fixed W&B `entity`.
- `tests/unit/algorithms/test_mpo.py:157` then *asserts* on those values (`entity == "joc"`, `model_name.endswith("/super_sft_16k")`), so anyone who replaces them with placeholders to make the configs usable breaks a unit test.

**Action:** pick one —
- **Preferred:** drop these from the PR and record the parity matrix (settings + W&B run ids) in `docs/guides/nemotron-omni-mpo.md`, which already carries the run history.
- **If they must ship:** replace both Lustre paths with `REPLACE_WITH_*` placeholders as the recipe already does, drop the `entity`/run-name pins, relax the `test_mpo.py` assertions to shape rather than identity, and extend `test_recipes_and_test_suites.py` to glob `examples/configs/experiments/` so they get the same checks as recipes.

Also worth noting for the Super no-pack file specifically: it sets `sequence_packing.enabled: false` on a multimodal dataset without the "does not run at this pin" header its Nano sibling carries. Whichever way the guard question resolves, the two files should say the same thing.

nemo_rl/data/datasets/preference_datasets/mmpr.py
```diff
@@ -0,0 +1,231 @@
...
+                            if "rejected_response" in record
+                            else "rejected"
+                        )
+                        # Preserve the legacy MPO data contract used for the
+                        # parity baseline with the reasoning checkpoint.
+                        for response_key in (chosen_key, rejected_key):
+                            response = str(record[response_key])
+                            if "<think>" not in response:
+                                record[response_key] = "<think></think>\n\n" + response
+                        record.setdefault("system", "")
```
aroshanghias-nvd: **1 action item, 1 follow-up.**

[`nemo_rl/data/datasets/preference_datasets/mmpr.py:179`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/data/datasets/preference_datasets/mmpr.py#L179-L190)

Credit first: the per-response `<think>` normalization immediately below this line is a real fix — `rejected` previously inherited `chosen`'s verdict — and it is well covered by the new `test_mmpr_preference_dataset_normalizes_responses_independently`. Worth calling out in the PR body.

The `system` change in the same hunk is the one to look at. `1ca5ab86` moved `record["system"] = ""` (which ran **only** when `chosen` lacked `<think>`) to an unconditional `record.setdefault("system", "")`. Two reachable deltas:

1. A row whose `chosen` **already contains `<think>`** and has no `system` field previously ended up with `system = None` after `Dataset.from_list` schema unification, and [`nemo_rl/data/datasets/preference_datasets/mmpr.py:75`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/data/datasets/preference_datasets/mmpr.py#L75-L76) — `if example.get("system") is not None:` — emitted **no** system turn. It now gets `""`, which is not `None`, so an **empty system turn is now prepended** to every already-reasoning row. That changes the rendered prompt and the token count against the parity baseline.
2. A row with a **non-empty** `system` and no `<think>` used to be blanked; it is now preserved.

The commit message says "preserve preference data contracts" and the in-place comment says "Preserve the legacy MPO data contract used for the parity baseline" — but the legacy contract blanked `system`. The new mixed-fixture test asserts only on `chosen`/`rejected`; the only `system` assertion in the file is a pre-existing one at line 95.

**Action:** either restore the conditional, or keep the change and state it as intentional with a test covering the already-reasoning row:

```suggestion
                        # Preserve the legacy MPO data contract used for the
                        # parity baseline with the reasoning checkpoint.
                        if "<think>" not in str(record[chosen_key]):
                            record["system"] = ""
```

**Follow-up (no action needed in this PR):** the new cache fingerprint at [`nemo_rl/data/datasets/preference_datasets/mmpr.py:132`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/data/datasets/preference_datasets/mmpr.py#L132-L141) stats every annotation file and image-root directory *before* the prepared cache is consulted at line 151. That makes a prepared cache unusable when the raw sources are not mounted — which is the workflow the guide advertises at [`docs/guides/nemotron-omni-mpo.md:61`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/docs/guides/nemotron-omni-mpo.md#L61-L64) ("prepare once, then run") — and turns a retired subset into a bare `FileNotFoundError`, where the record loop at lines 167-171 deliberately skips such rows. Wrapping `_path_signature`'s `path.stat()` in `try/except OSError` and returning a `|missing` sentinel would keep the invalidation behaviour while matching the loop's own tolerance.

scripts/vlm_mpo_chain.sh
```diff
@@ -0,0 +1,88 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+NEMORL="${NEMORL:-$(cd "${SCRIPT_DIR}/.." && pwd)}"
+CHAIN_SEGMENTS="${CHAIN_SEGMENTS:-2}"
+CHAIN_STEP_TARGETS="${CHAIN_STEP_TARGETS:-}"
+CHAIN_ID="${CHAIN_ID:-$(date +%Y%m%d-%H%M%S)}"
+CHAIN_NAME="${CHAIN_NAME:-nemotron-omni-mpo-chain-${CHAIN_ID}}"
+RESULTS_NAME="${RESULTS_NAME:-${CHAIN_NAME}}"
+WANDB_NAME="${WANDB_NAME:-${CHAIN_NAME}}"
+WANDB_RUN_ID="${WANDB_RUN_ID:-$(python -c 'import secrets; print("mpo" + secrets.token_hex(3))')}"
```
aroshanghias-nvd: **1 action item.**

[`scripts/vlm_mpo_chain.sh:14`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/scripts/vlm_mpo_chain.sh#L14)

Flagging explicitly because `1ca5ab86` added four lines to this file: **line 14 is byte-identical to the previous head**, so the new `CHAIN_STEP_TARGETS` guard at lines 22-25 is a different concern and this one is still open. PR-introduced.

```bash
WANDB_RUN_ID="${WANDB_RUN_ID:-$(python -c 'import secrets; print("mpo" + secrets.token_hex(3))')}"
```

The CI container has **no `python` and no `python3`** on the default `PATH` — only `/opt/nemo_rl_venv/bin/python`. Under `set -euo pipefail` (line 3) a failed command substitution inside an assignment aborts the script, so the whole chain fails to submit with a bare `python: command not found`. The `build-and-dependency` and `linting-and-formatting` skills both mandate `uv run` for this reason.

**Action:**

```suggestion
WANDB_RUN_ID="${WANDB_RUN_ID:-mpo$(head -c3 /dev/urandom | od -An -tx1 | tr -d ' \n')}"
```

(or `${RUNNER:-uv run --no-sync} python -c ...` if you'd rather keep Python). Note the new `test_vlm_mpo_chain.py` passes only because pytest runs with the venv already on `PATH`, so it gives no coverage of this path.

Minor, same file: `CHAIN_SEGMENTS` defaults to `2`, so a bare `scripts/vlm_mpo_chain.sh` now exits 2 where it previously submitted — and `CHAIN_STEP_TARGETS` appears nowhere in `docs/guides/nemotron-omni-mpo.md`. The guard itself is correct (each segment must exit 0 deliberately for the `afterok` chain to advance); it just needs documenting.

docs/guides/nemotron-omni-mpo.md
```diff
@@ -0,0 +1,120 @@
+# Nemotron Omni MPO
+
+This path ports offline image MPO to the canonical `NemotronOmniModel`
+from NeMo-RL PR #3290 (which supersedes #3227) and Megatron-Bridge PR #4885.
+Development is based on the rewritten
+`aroshanghias/nemotron-omni-main-migration` head. It deliberately does not
```
aroshanghias-nvd: **1 action item, plus a batch of minor items.**

[`docs/guides/nemotron-omni-mpo.md:5`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/docs/guides/nemotron-omni-mpo.md#L5-L6)

The new "Support and qualification status" section added in `1ca5ab86` is a genuine improvement — saying plainly that Nano parity was measured "on an earlier revision" and that Super is unqualified is exactly the right kind of honesty. The rest of the guide still reads as a migration memo rather than user documentation:

- line 6 — "Development is based on the rewritten `aroshanghias/nemotron-omni-main-migration` head", a personal branch that will not exist after merge
- [`docs/guides/nemotron-omni-mpo.md:24`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/docs/guides/nemotron-omni-mpo.md#L24) — `--tag nemorl-omni-mpo:pr3290`, a PR-numbered image tag
- [`docs/guides/nemotron-omni-mpo.md:80`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/docs/guides/nemotron-omni-mpo.md#L80) — W&B defaults `joc` / `nemotron-omni-main-migration`
- [`docs/guides/nemotron-omni-mpo.md:109`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/docs/guides/nemotron-omni-mpo.md#L109) — a "Qualification order" section that is a test plan, not user guidance

It also documents **zero** config keys. None of `mpo.bco_loss_weight`, `mpo.reward_shift_momentum`, `mpo.preference_loss_weight`, `mpo.sft_loss_weight`, `mpo.stop_after_step`, `policy.sequence_packing.pair_grouping_key`, `.max_sequences_per_bin`, `.fuse_loss`, `data.train.split_validation_size`, `.legacy_validation_split`, `.max_samples`, or `.cache_dir` appears anywhere. Compare `docs/guides/dpo.md`. The no-pack limitation section is good and should stay.

**Action:** drop the branch/PR-tag/internal-W&B references and the qualification checklist, and add an "MPO configuration" section documenting each `mpo.*` and `policy.sequence_packing.*` key the code actually reads, with its default. Also consider filing it under `docs/guides/models/nemotron/` — it sits between `dpo.md` and `dapo.md` today, among algorithm guides, while every other Nemotron document lives under the model directory.

---

**Context — no action required.** Batching the small items so they don't each need a thread:

- [`nemo_rl/algorithms/loss/loss_functions.py:1235`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/loss/loss_functions.py#L1235) — `MPOLossFn.set_reward_shift` has zero callers anywhere and bypasses `MPOSaveState`, so calling it would desync the checkpointed shift from the loss. Two-line deletion; `_update_reward_shift` is already the single writer.
- [`nemo_rl/models/policy/__init__.py:222`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/models/policy/__init__.py#L222) — `pair_grouping_key: NotRequired[str]` could be `Literal["pair_index"]` to match its sibling `microbatch_order` two lines up, since `mpo.py:103` raises for anything else. All three new keys are also undocumented, unlike their neighbours.
- `tests/unit/models/megatron/test_megatron_setup.py:3336` — `torch.testing.assert_close(model_param, torch.zeros(2))` compares a local to itself; the real assertion is the `assert_not_called()` above it.
- Test coverage gaps worth a follow-up: the odd-`min_padding` raise at [`nemo_rl/algorithms/utils.py:494`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/algorithms/utils.py#L494) is unreachable from the suite; the padding assertion in `test_utils.py` checks labels `maybe_pad_last_batch` rewrites unconditionally, so duplicating the last *row* instead of the last *pair* still passes; pair atomicity is only exercised at `max_sequences_per_bin=1`.
- Unrelated churn worth dropping to shrink the diff: the single blank-line deletion in [`nemo_rl/models/generation/vllm/vllm_worker.py:901`](https://github.com/NVIDIA-NeMo/RL/blob/1ca5ab86191e12a2ab14cc7d092809242fae776e/nemo_rl/models/generation/vllm/vllm_worker.py#L901) (this PR starts no vLLM worker); the three `from wandb import X` → `from wandb.data_types import X` rewrites, which are a semantic no-op at `wandb==0.28.1` and redundant with the `known-third-party = ["wandb"]` isort rule this PR adds; and `tests/functional/grpo_topp_topk.sh:33`, which sets `logprobs_mode=processed_logprobs` when `examples/configs/grpo_math_1B.yaml:399` (the default config that script loads) already sets it.

aroshanghias-nvd (commented):
Thanks for the sustained work here, and for running the self-review workflow on it — the three-way Nano/Super parity comparison with linked W&B runs, and the qualification-status section added in `1ca5ab86`, are considerably more evidence than a change this size usually arrives with. Two of the fixes in that commit are clearly right: the per-response `<think>` normalization (`rejected` previously inherited `chosen`'s verdict), and replacing the `run_config.yaml` / `save_optim` heuristic with a probe of what the checkpoint actually contains.

This is an independent pass over the same head, `1ca5ab86191e12a2ab14cc7d092809242fae776e`, by the same workflow. It reaches a different conclusion in two places, and I think the reason is the one your own summary names: *"Local pytest execution was not available on this host because the Python runtime exited with SIGBUS."* Both blocking items below are test-detectable, and both were found by running the suite.

**Two blocking items:**

1. **`_is_megatron_optimizer_key` never matches a real MoE checkpoint key.** `ChainedOptimizer` prefixes every sharded key with `chained_<idx>.` when a model has more than one sub-optimizer — always true for MoE, and both shipped recipes are MoE. Measured on a checkpoint from this stack: **64 `chained_N.optimizer` keys, 0 bare**. The probe therefore returns `optimizer_path = None` and every Nemotron Omni resume silently reinitializes the Adam moments and the LR schedule. This is a regression against this PR's own previous head, where `save_optim: true` made the old heuristic work.

2. **A pre-existing `main` test now fails.** `test_get_resume_paths_warns_when_megatron_optimizer_missing` exists at the merge base; A/B on the same file and command gives `1 failed, 71 passed` at `1ca5ab86191e12a2ab14cc7d092809242fae776e` versus `71 passed` at `0fdd1477`. Because `pyproject.toml:650` sets `-x`, that failure aborts the file — so the two tests `1ca5ab86` adds never execute, which is why the new predicate has no effective coverage.

**Merge readiness.** The `disabled.txt` conflict is resolved and the PR is mergeable — that part is done. The gap now is CI: the checks attached to this head are the lightweight gating jobs only (DCO, triage, copyright, secrets, semantic-PR, submodule, `uv lock check`). The most recent `/ok to test` targets `11dd654b0` from 2026-08-04, and the head has been updated several times since, so no unit or functional job has run against this code. A fresh `/ok to test 1ca5ab86191e12a2ab14cc7d092809242fae776e` is the unblocking action — and with item 2 above it will come back red, which is the useful outcome.

Since the parity runs predate both the latest rebase and the Megatron-Bridge pin move (`0c565c9a` → `8c46dc42`), it would also help to state in the description which numbers were measured at which head. Related: a few description claims are true of the foundation rather than this PR — `pad_between_seqs` and its regression test are already on `main` (they do not appear in this diff), and the vLLM 0.25.1 / `processed_logprobs` adaptations are not demonstrable from the changes here.

**Verification behind this review.** Unit tests run in the CI container at this head: `test_mpo_loss` 6, `test_mpo` 10, `test_dpo` 7, `test_batched_data_dict` 42, `test_mmpr_preference_dataset` 7, `test_collate_fn` 2, `test_vlm_preference_processor` 1, `test_vlm_mpo_chain` 1 — all passing, and the 23 / 42 counts in your checklist reproduce exactly. Linters at their pinned versions (ruff check/format/isort, config minimize-check, markdown naming) all pass. Megatron-Bridge claims were verified against the pinned submodule at `8c46dc42`, not against the container, which ships `573e088c`. Findings were filtered by an adversarial pass; several candidates were dropped for failing verification, including a `fix_mistral_regex` version claim refuted by reading the pinned `transformers`, and an `add_special_tokens` change with no demonstrable effect.

_Generated by Claude Code_

examples/configs/experiments/vlm_mpo-nemotron-super-omni-nopack-2k-parity.yaml
```diff
@@ -0,0 +1,27 @@
...
+  tokenizer:
+    name: ${policy.model_name}
+  max_total_sequence_length: 2048
+  megatron_cfg:
+    tensor_model_parallel_size: 8
+    expert_model_parallel_size: 32
+    pipeline_model_parallel_size: 1
+    context_parallel_size: 1
+    # Initial Super parity excludes the checkpoint's auxiliary MTP head.
+    mtp_num_layers: 0
```
DanialTaheri: Could you quickly test the MPO with MTP head training activated to find out if the loss stays the same for super model.
megatron_cfg:
mtp_num_layers: 1
mtp_use_repeated_layer: false
mtp_detach_heads: true
mtp_loss_scaling_factor: 0.1

jinglinglingling:
Super no-pack qualification completed on the current main-based branch.

- New 100-step run: https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/mpo4a75dd
- Legacy baseline: https://wandb.ai/nvidia/vlm-mpo-dev-mmpr/runs/fpieeug0
- Matched settings: 4 nodes / 32 GPUs, TP8, EP32, CP1, 2K, GBS 256, MTP disabled, sequence packing disabled
- Loss parity over 99 aligned steps: mean delta `+0.00047`, Pearson `r=0.956`; LR is identical
- The 50→51 resume boundary is continuous: loss `0.8151 → 0.8111`, with a smooth reward-shift update and no reset/drop

Both chained Slurm segments completed successfully and the single W&B run contains all 100 training steps.

nemo_rl/utils/checkpoint.py
jinglinglingling: Addressed in 8485c3ac6. `_is_megatron_optimizer_key` now strips the `chained_<idx>.` prefix before checking optimizer keys, and the regression test uses a realistic `chained_1.optimizer...` metadata key.

nemo_rl/utils/checkpoint.py
jinglinglingling: Addressed in 8485c3ac6. The pre-existing missing-optimizer test now mocks `_load_megatron_sharded_metadata_keys`, so it reaches and verifies the warning path; the torch-dist/MoE test covers chained optimizer metadata.

tests/unit/algorithms/test_mpo_loss.py
jinglinglingling: Addressed in 8485c3ac6. Added `test_mpo_loss_matches_closed_form_components`, which independently pins preference, SFT, BCO, total loss, accuracy, and reward metrics with beta=0.5.

nemo_rl/algorithms/mpo.py
jinglinglingling: Addressed in 8485c3ac6. MPO setup now validates that `train_global_batch_size` is divisible by the data-parallel degree when `max_sequences_per_bin=1`, with a regression test for the invalid case. The two warning/packer observations remain non-blocking follow-ups as noted.

nemo_rl/models/megatron/data.py
jinglinglingling: Addressed in 8485c3ac6. Removed the guard and stale Megatron-Bridge #5181 references. The unpacked path passed smoke and a 100-step Super no-pack parity run; the results are linked in the PR and #3525 is closed.

nemo_rl/algorithms/mpo.py
jinglinglingling: Addressed in 8485c3ac6. `_validate_pair_safe_packing` now validates rather than mutates the config: Megatron backend, `pair_grouping_key="pair_index"`, and `fuse_loss=true` are required. Regression tests cover DTensor, non-fused loss, and invalid grouping keys.

nemo_rl/algorithms/dpo.py
jinglinglingling: Addressed in 8485c3ac6. `_get_dpo_save_state` is now the live restore path and is generalized for the save-state class and initializer. Added MPO coverage proving `reward_shift` survives restore from a partial checkpoint.

nemo_rl/algorithms/dpo.py
jinglinglingling: Addressed in 8485c3ac6. `stop_after_step` is declared on `DPOConfig`, documented in `examples/configs/dpo.yaml`, validated from shared DPO setup, and read directly rather than via `getattr`. Added coverage for an already-reached resume boundary.

examples/configs/recipes/vlm/vlm_mpo-nemotron-omni-30ba3b-mmpr-1n8g-megatron-tp8.v1.yaml
jinglinglingling: Addressed in 8485c3ac6. Added `examples/configs/vlm_mpo.yaml`, pointed the MPO recipe and `ALGO_MAPPING_TO_BASE_YAML` to it, changed the bare runner default to the exemplar, enabled `MasterConfig` extras consistently, minimized the recipe, and added validation coverage.

examples/configs/experiments/vlm_mpo-nemotron-super-omni-nopack-2k-parity.yaml
jinglinglingling: Addressed in 8485c3ac6. Experiment configs now use portable `REPLACE_WITH_*` placeholders and no pinned W&B entity. The recipe/config test suite now resolves every experiment config and rejects internal path prefixes and pinned W&B entities.

nemo_rl/data/datasets/preference_datasets/mmpr.py
jinglinglingling: Addressed in 8485c3ac6. Restored the conditional legacy behavior: already-reasoning rows do not render an empty system turn, while non-reasoning rows receive the legacy blank system value. The mixed-row tests now pin both cases.

scripts/vlm_mpo_chain.sh
jinglinglingling: Addressed in 8485c3ac6. Run IDs are generated with shell-native `/dev/urandom` plus `od`, so the launcher no longer depends on `python` being on PATH. Chained-job requirements and `CHAIN_STEP_TARGETS` are documented and launcher-tested.

docs/guides/nemotron-omni-mpo.md
jinglinglingling: Addressed in 8485c3ac6. Reworked the guide as user documentation, removed internal branch/PR-image/W&B defaults, and documented MPO and pair-safe packing keys. Also removed the unused reward-shift setter, tightened `pair_grouping_key` to a Literal, cleaned the redundant test assertion, and reverted unrelated W&B/vLLM/GRPO diff churn.

examples/configs/recipes/vlm/vlm_mpo-nemotron-omni-30ba3b-mmpr-1n8g-megatron-tp8.v1.yaml
aroshanghias-nvd: partially fixed in 8485c3ac6e5b29c4c7eed0b4706330c648b47ff6 — the MPO exemplar and recipe mapping are now in place, but current head b9a6dbd952d37cf1d45045abcc518f0498679a58 still fails L0_Unit_Tests_Other: `test_all_config_files_have_required_keys[.../examples/configs/vlm_mpo.yaml]` raises `Could not determine algorithm type` because `tests/unit/test_config_validation.py` has no MPO `MasterConfig` import or `elif "mpo"` branch. Please add that branch and rerun the job.

jinglinglingling: Fixed in 44852bf3f07fe776dbea3cadab9cb9754dc2253e. Added the MPO MasterConfig import and the `elif "mpo"` dispatch in `tests/unit/test_config_validation.py`, and retriggered CI for this head. I will leave this thread open until L0_Unit_Tests_Other passes.

yfw (approved):

PR #3348, opened 2026-07-25
fix(async_utils): only close event loop after worker thread stops

andrewwhitecdw:
## Problem

`AsyncLoopThread.close()` calls `loop.close()` unconditionally after joining the worker thread, even if `join()` times out and the worker thread is still running. Closing a loop that is still executing in another thread can raise `RuntimeError: Cannot close a running event loop` and leaves the worker thread in an undefined state.

## Fix

Only call `loop.close()` when the worker thread has actually stopped (`not self._thread.is_alive()`). If the thread is still alive after `join(timeout)`, we leave the loop open so the thread can finish stopping it safely.

## Verification

The existing regression tests in `tests/test_async_utils.py` already assert this exact behavior:
- `test_close_does_not_close_loop_while_worker_thread_is_alive`
- `test_close_closes_loop_when_worker_thread_is_dead`
- `test_close_is_idempotent_when_loop_already_closed`

This change makes the implementation match those regression expectations.


andrewwhitecdw:
Closing this sweep-generated PR: PR has 3 commits; sweep requires exactly one commit per PR. It does not meet the sweep requirements (single signed-off commit).

andrewwhitecdw:
Squashed the branch down to a single commit and rebased on the latest `main`. This PR is now ready for review.

andrewwhitecdw:
Hi team, this PR is squashed to a single commit and ready for review. The change makes AsyncLoopThread.close() only call loop.close() when the worker thread has actually stopped, avoiding a RuntimeError when join() times out. Please let me know if anything else is needed.

PR #3235, opened 2026-07-16
build: install megatron-bridge from submodule; auto-flow its deps on bump

terrykong:
# What does this PR do ?

Installs `megatron-bridge` straight from the submodule's own `pyproject.toml` so its dependencies (and transitively Megatron-LM's) auto-flow into `uv.lock` on every submodule bump.

**Why this simplifies things:** megatron-bridge's dependency list used to be maintained in **three places** that had to be kept in sync by hand on every bump — the submodule's `pyproject.toml` (upstream truth), `setup.py::CACHED_DEPENDENCIES` in the proxy (checked against upstream at build time), and the `[[tool.uv.dependency-metadata]]` mirror in nemo-rl's `pyproject.toml` (what uv actually resolved), with `tools/check_mbridge_deps.py` + the `mbridge-deps-sync.yml` CI workflow guarding the copies against drift. This PR deletes copies 2 and 3 and the entire sync-check apparatus; the single source of truth is the submodule, and drift detection moves into `uv lock` itself: a new always-on `lockfile-check.yml` workflow runs `uv lock --check` on every PR touching dependency paths (verified locally to catch a submodule bump without relock), and the container build's `uv sync --locked` backstops it in the gated pipeline. The bump procedure shrinks from six steps to three.

| Deleted | What it was |
|---|---|
| `3rdparty/Megatron-Bridge-workspace/setup.py` (+ proxy `pyproject.toml`, `is_megatron_bridge_installed.py`) | Proxy package with `CACHED_DEPENDENCIES`, a hand-synced copy of the submodule's dep list |
| `[[tool.uv.dependency-metadata]]` for megatron-bridge | Third copy of the same list, fed to the resolver |
| `tools/check_mbridge_deps.py` + `.github/workflows/mbridge-deps-sync.yml` | CI guard that the copies hadn't drifted — nothing left to guard |

(Replacement guard added: `.github/workflows/lockfile-check.yml` — ungated `uv lock --check` on dependency-path PRs.)

# Issues

N/A

# Usage

The Megatron-Bridge (and transitively Megatron-LM) bump procedure is now:

```bash
# 1. Check out the new Megatron-Bridge commit
cd 3rdparty/Megatron-Bridge-workspace/Megatron-Bridge
git fetch origin && git checkout <new-commit>
git submodule update --init --recursive
cd -

# 2. Relock — deps flow from the submodule pyproject automatically
uv lock

# 3. Review `git diff uv.lock`, commit submodule pointer + uv.lock together
```

[...]

PR #3352, opened 2026-07-26
chore: update NCCL to 2.30.4 via uv override

yfw:
# What does this PR do ?

NCCL 2.30.4 fixes a context-parallel hang on GB200 when GPUs are non-contiguous.

torch 2.11.0+cu130 pins nvidia-nccl-cu13==2.28.9, so a plain version bump is unresolvable ("No solution found"). Use [tool.uv] override-dependencies, the same mechanism already used for nvidia-cublas, nvidia-cudnn-cu13 and nvidia-nvshmem-cu13.

Also raise the ceiling in the vendored TensorRT-LLM workspace from <=2.29.7 to <=2.30.4 so the declared constraint matches what is installed rather than being silently overridden. Upstream's ceiling tracks what NGC rel-26-04 ships and is not a tested incompatibility; nothing in the TRT-LLM build path references NCCL.

Unlike excluding the wheel in the Dockerfile, this covers non-container `uv sync` users and worker venvs built at runtime, and pins the version reproducibly in uv.lock instead of inheriting whatever the base image ships.

Thanks to @cspades @ZhiyuLi-Nvidia @macandro96 for identifying the root cause and testing.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -354,6 +354,9 @@ override-dependencies = [
   "nvidia-cublas==13.5.1.27; sys_platform != 'darwin'",
   "nvidia-cudnn-cu13==9.20.0.48; sys_platform != 'darwin'",
   "nvidia-cudnn-frontend==1.23.0",
+  # NCCL 2.30.4 fixes a context-parallel hang on GB200 when GPUs are non-contiguous.
+  # An override is required because torch 2.11.0+cu130 pins nvidia-nccl-cu13==2.28.9.
+  "nvidia-nccl-cu13==2.30.4",
   "nvidia-nvshmem-cu13>=3.6.5; sys_platform != 'darwin'",
   "opencv-python-headless; sys_platform == 'never'",
   "timm<=1.0.22",
--- a/3rdparty/TensorRT-LLM-workspace/pyproject.toml
+++ b/3rdparty/TensorRT-LLM-workspace/pyproject.toml
@@ -31,7 +31,9 @@ dependencies = [
   "torch>=2.11.0",
   "torchvision",
   "nvidia-modelopt>=0.37.0",
-  "nvidia-nccl-cu13>=2.28.9,<=2.29.7",
+  # Ceiling raised from upstream's <=2.29.7 (which tracks what NGC rel-26-04 ships,
+  # not a tested incompatibility) so it matches the nvidia-nccl-cu13==2.30.4 override.
+  "nvidia-nccl-cu13>=2.28.9,<=2.30.4",
   "nvidia-cuda-nvrtc",
   "nvidia-cutlass-dsl[cu13]==4.5.0",
   "nvidia-ml-py>=13",
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -23,7 +23,7 @@
 #   --build-arg SKIP_SGLANG_BUILD=1  # Skip SGLang dependencies
 #   --build-arg SKIP_TRTLLM_BUILD=1  # Skip TRT-LLM dependencies
 
-ARG BASE_IMAGE=nvcr.io/nvidia/cuda-dl-base:26.03-cuda13.2-devel-ubuntu24.04
+ARG BASE_IMAGE=nvcr.io/nvidia/cuda-dl-base:26.05-cuda13.2-devel-ubuntu24.04
 FROM scratch AS nemo-rl
 ARG NRL_GIT_REF=main
 ADD --keep-git-dir=true https://github.com/NVIDIA-NeMo/RL.git#${NRL_GIT_REF} /
--- a/examples/nemo_gym/nemotron-3-ultra/ifbench_teacher.yaml
+++ b/examples/nemo_gym/nemotron-3-ultra/ifbench_teacher.yaml
@@ -56,6 +56,7 @@ grpo:
   advantage_clip_low: -20
   advantage_clip_high: 20
   val_period: 10000
+  val_start_at: -1
   val_at_start: false
   val_at_end: false
   overlong_filtering: false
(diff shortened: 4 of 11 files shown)


saumishr (dismissed):

pyproject.toml
```diff
@@ -358,6 +358,9 @@ override-dependencies = [
   "nvidia-cublas==13.5.1.27; sys_platform != 'darwin'",
   "nvidia-cudnn-cu13==9.20.0.48; sys_platform != 'darwin'",
   "nvidia-cudnn-frontend==1.23.0",
+  # NCCL 2.30.4 fixes a context-parallel hang on GB200 when GPUs are non-contiguous.
+  # An override is required because torch 2.11.0+cu130 pins nvidia-nccl-cu13==2.28.9.
+  "nvidia-nccl-cu13==2.30.4",
```
terrykong, line 359: [`pyproject.toml:363`](https://github.com/NVIDIA-NeMo/RL/blob/755b5abe689d064c05e56df311b771b8a37c0d43/pyproject.toml#L363)

This branch is 11 commits behind `main` and `uv.lock` conflicts (`git merge-tree --write-tree HEAD origin/main` → `CONFLICT (content): Merge conflict in uv.lock`). Worth calling out because the rebase is not routine:

1. `main` restructured the `mcore` extra and `[tool.uv.sources]` in [#3235](https://github.com/NVIDIA-NeMo/RL/pull/3235) — `megatron-core`/`transformer-engine`/`transformers`/`nvidia-modelopt`/`onnxscript` replaced by `megatron-bridge[te,ssm]`, the `megatron-bridge` source path moved to `3rdparty/Megatron-Bridge-workspace/Megatron-Bridge`, and `emerging-optimizers`/`fast-hadamard-transform` moved into `constraint-dependencies`. The lock in this PR predates all of that.
2. `main` also added [`.github/workflows/lockfile-check.yml`](https://github.com/NVIDIA-NeMo/RL/blob/main/.github/workflows/lockfile-check.yml), which does not exist on this branch. It runs `uv lock --check` on every PR touching `pyproject.toml`, `uv.lock` or `3rdparty/**` — so it will gate this PR once rebased.

**Action:** rebase on `main`, then regenerate with `uv lock` rather than hand-resolving the `uv.lock` conflict — a hand-merge carrying the pre-#3235 resolution will fail the new `uv lock --check` job.

terrykong (commented):
Reviewed the NCCL override, the vendored TRT-LLM ceiling raise, and the base-image bump. The mechanism is sound and the commit messages are unusually good — the load-order argument in commit 2 (system NCCL vs. the wheel via `DT_RPATH`) is the right reason to couple the image bump to the override.

**This PR currently conflicts with `main`** (`mergeable: CONFLICTING`). Details in the inline comment on `pyproject.toml` — the short version is that the rebase needs `uv lock` rather than a hand-merge, because `main` has since restructured the `mcore` extra and added a `uv lock --check` gate.

Three things I checked that turned out **fine**, noted so they don't get re-raised:

- **NCCL net-plugin compatibility on AWS EFA.** `docker/Dockerfile` consumes the host's `/opt/amazon/ofi-nccl/lib`, and only `aws-ofi-nccl` >= v1.20.0 is [tested against NCCL v2.30.4-1](https://github.com/aws/aws-ofi-nccl/releases), so I expected older installers to silently fall back to socket transport. They don't: NCCL 2.30.4 still loads plugin interfaces v12 down to v6 ([`src/plugin/net.cc`](https://github.com/NVIDIA/nccl/blob/v2.30.4-1/src/plugin/net.cc#L39-L41), `int ncclNetVersion[7] = {12, 11, 10, 9, 8, 7, 6}`). No action needed.
- **The missing `; sys_platform != 'darwin'` marker** on the new override, where its siblings all have one. Inert — [`[tool.uv] environments`](https://github.com/NVIDIA-NeMo/RL/blob/755b5abe689d064c05e56df311b771b8a37c0d43/pyproject.toml#L323-L326) restricts resolution to linux x86_64/aarch64.
- **The cuDNN skew** the commit message flags (system 9.22.0.52 vs. the pinned wheel 9.20.0.48). The [`CUDNN_HOME`/`LD_LIBRARY_PATH` guard](https://github.com/NVIDIA-NeMo/RL/blob/755b5abe689d064c05e56df311b771b8a37c0d43/docker/Dockerfile#L170-L179) makes the pip copy win, the documented failure mode is a loud crash rather than a silent wrong result, and `_build_container.yml` resolves `BASE_IMAGE` from the Dockerfile `ARG` — so the CI:L2 run genuinely exercises 26.05 and would surface it. Worth eyeballing `te.get_cudnn_version()` on the first image as you noted, but it doesn't need to gate review.

_Generated by Claude Code_

kajalj22 (dismissed):

kajalj22:
LGTM modulo Terry's comment on PR being behind main and the incoming vllm bump

pyproject.toml
yfw: rebased on top of recent `main` and reran `uv lock`

terrykong (dismissed):

terrykong (dismissed):

pyproject.toml
```diff
@@ -354,6 +354,9 @@ override-dependencies = [
...
```
youngeunkwon0405, line 359: Hi @yfw, if I want to change the nccl version later, is this line the only thing I need to change?
E.g., if I need nccl v2.30.7, then just changing the version number here is the only requirement?

yfw: We also had to update the base image https://github.com/NVIDIA-NeMo/RL/pull/3352/changes#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcL26

terrykong (approved):

pyproject.toml
youngeunkwon0405: Thanks!

PR #2789, opened 2026-06-11
fix(vllm): fix FP8 MoE kernel initialization for CUDA graph mode

achartier:
# What does this PR do ?

Fixes two bugs in `fp8.py` that prevented FP8 blockwise vLLM generation from working with CUDA graphs (`enforce_eager: false`):

1. **`hasattr` guard always True** — `FusedMoEMethodBase.__init__` sets `self.moe_kernel = None`, so the guard never triggered and `make_fp8_moe_kernel` was never called. Fix: gate on `self.moe_kernel is None`.
2. **Kernel re-created on every weight refit** — `process_weights_after_loading` runs outside a `set_current_vllm_config` context, crashing on `get_current_vllm_config()`. Fix: skip re-creation if `moe_kernel` is already set.

Validated on GLM-5.1 64n×8g — ~3.7× faster generation vs bf16 eager after this fix.

# Issues

N/A

# Usage

No API change. FP8 + CUDA graphs now works without requiring `enforce_eager: true`.

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [x] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* N/A

PR #3353, opened 2026-07-26
fix: merge grouped-expert LoRA weights during DTensor refit

rahul-sarvam:
# What does this PR do ?

Merges LoRA adapters on Automodel's grouped MoE experts into the base expert weights during the
DTensor refit, so GRPO with expert adapters no longer breaks weight streaming into the
inference engine.

# Issues

None filed — happy to open one first if you prefer that order.

# Details

`policy.dtensor_cfg.lora_cfg.target_modules` can match Automodel's grouped expert modules,
because `ModuleMatcher` matching is name-based rather than type-based, and
`apply_lora_to_linear_modules` then wraps them in `GroupedExpertsLoRA` /
`GroupedExpertsDeepEPLoRA`. Those keep their adapters as **bare parameters on the expert
module** — `lora_gate_and_up_A/B`, `lora_down_A/B` — next to fused base weights
`gate_and_up_projs` / `down_projs`, rather than as `LinearLoRA` submodules with
`.lora_A.weight`.

Three places in `dtensor_policy_worker_v2.py` assume the `LinearLoRA` layout:

| Location | Assumption | Consequence |
|---|---|---|
| `_maybe_merge_lora_weight` | `fqn.endswith(".weight")` and `isinstance(module, LinearLoRA)` | grouped-expert base weights refit **unmerged** |
| `dtensor_params_generator` skip filter | name ends `.lora_A.weight` / `.lora_B.weight` | adapter tensors are **streamed** to the engine |
| `prepare_refit_info` skip filter | same | adapter keys are **advertised** to the engine |

On a `google/gemma-4-26B-A4B-it` GRPO run (1 node × 8 B200, EP=8, `experts: gmm`,
`dispatcher: deepep`, `target_modules: ['*_proj', '*experts']`) the first refit fails:

```
KeyError: 'layers.0.moe.experts.lora_gate_and_up_A'
  (vLLM RayWorkerWrapper, during DTensorPolicyWorkerV2.stream_weights_via_ipc_zmq)
```

An engine that tolerated unknown keys would instead generate from unadapted experts while
training adapts them — a divergence that grows from zero as `lora_B` leaves its zero
initialisation, which is harder to notice.

**The merge.** The expert forward computes `x @ W + (x @ A @ B) * scale`, so the delta is a
per-expert batched matmul already in the base weight's orientation — no transpose. Under expert
parallelism `A`/`B` are sharded over the expert dimension exactly as `W` is, so `full_tensor()`
gathers them in matching expert order. Matching is by attribute presence rather than by
importing Automodel's classes, so it survives refactors on either side.

# Usage

```yaml
policy:
  dtensor_cfg:
    _v2: true
    lora_cfg:
      enabled: true
      match_all_linear: false
      # '*experts' reaches the grouped expert module; '*_proj' covers attention and the
      # shared expert while leaving the router (`moe.gate.proj`) and lm_head frozen.
      target_modules: ['*_proj', '*experts']
      dim: 8
      alpha: 32
```

**Pre checks**:

`tests/unit/models/dtensor/test_lora_expert_refit.py` adds 17 CPU-only cases covering adapter
detection for both layouts, adapter resolution, the per-expert delta, out-of-place merging
(the generator may hand over a parameter's own storage), the zero-`lora_B` no-op, pass-through
for un-adapted modules, and dtype preservation.

Run against NeMo-RL `bbc58087` + Automodel `24b47e85`:

```
tests/unit/models/dtensor/test_lora_expert_refit.py ..  17 passed
tests/unit/models/dtensor/test_lora.py ..............   11 passed   (unchanged)
```

End-to-end on the run above: generation KL error 0.00068 / 0.00056 / 0.00052 over three GRPO
steps, matching a full fine-tune of the same model, i.e. the refit reproduces the training-side
computation. Adapter checkpoint 894 MB / 247 M parameters; steps 38–41 s versus 36 s with
attention-only LoRA.

I did not touch `docs/guides/lora.md`; happy to add a note there that reaching the experts
requires `target_modules` (since `match_all_linear` only matches `nn.Linear` and so silently
skips grouped experts) if you'd like that in the same PR.

# Additional Information

* **A submodule bump is also needed to exercise this path.** Reaching the refit requires the
  expert-LoRA dtype fix from Automodel PR #2789 (`_to_grouped_mm_operand`). `main` pins
  Automodel `24b47e856263d313b942f0ed666c63fff83306b4`, which predates it, so grouped-expert
  LoRA currently aborts earlier with
  `RuntimeError: Expected b.scalar_type() == torch::kBFloat16 to be true, but got false`.
  I left that out of this PR since bumping the submodule is your call; I tested with #2789
  backported locally.
* **Scope.** This affects every model on Automodel's grouped-expert path — 21 model packages at
  the pinned revision, including `qwen3_moe`, `qwen3_next`, `deepseek_v3`, `gpt_oss`,
  `glm4_moe`, `minimax_m2` and `nemotron_v3` — not just Gemma 4.
* This work was done with AI assistance; the code, tests and measurements were reviewed by me.

### DIFF
--- a/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
+++ b/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
@@ -102,7 +102,7 @@ def dtensor_params_generator(
     """
     module_map = dict(model.named_modules())
     for name, tensor in model.state_dict().items():
-        if name.endswith(".lora_A.weight") or name.endswith(".lora_B.weight"):
+        if _is_lora_adapter_tensor(name):
             continue
         full_tensor = tensor.full_tensor() if isinstance(tensor, DTensor) else tensor
         merged_tensor = _maybe_merge_lora_weight(module_map, name, full_tensor)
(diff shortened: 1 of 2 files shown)


yuki-97 (commented):
thanks for the contribution! @RayenTian could you review?

RayenTian:
Thanks for adding this @rahul-sarvam ! Overall LGTM. One thing before merge: this PR keeps Automodel pinned at 24b47e85, which predates #2789, so on main this path crashes earlier (Expected ... kBFloat16) and never hits the new code — the E2E numbers were from a local backport of #2789.
 So this PR likely needs to be blocked until Automodel is bumped to include #2789; alternatively, file a tracking issue noting the fix is inert until then so it doesn't land as unreachable code. 
WDYT @yuki-97

rahul-sarvam:
Happy either way:

- **Merge + tracking issue** — I'll file one that records the dependency (#2789 → submodule bump → this becomes reachable), links back here, and includes the repro so it's easy to verify once the pin moves.
- **Hold until the bump** — also fine; let me know when the submodule advances and I'll re-run the E2E check against the real pin and post the numbers.

Two optional extras I'm happy to add:

1. A note in `docs/guides/lora.md` that `match_all_linear` cannot reach grouped MoE experts (it short-circuits on `_is_linear_module`), so expert adapters need `target_modules` naming the experts module. Today the docs describe an empty `target_modules` / `match_all_linear` as covering "all linear layers".
2. A clearer failure at setup time when expert LoRA is requested against an Automodel without #2789, so users get an actionable message instead of a grouped-GEMM dtype assert.

RayenTian:
Thanks @rahul-sarvam ! After discussing with team, this PR may need block until automodel bumped. Once Nemorl bumps the `automodel` version, we will review this PR again.

PR #3356, opened 2026-07-27
fix(refit): stream a parameter larger than the IPC staging buffer

terrykong:
## The bug

`stream_weights_via_ipc_zmq_impl` sizes its two ping-pong staging buffers from **free memory** (`NRL_REFIT_BUFFER_MEMORY_RATIO`, default `0.3`, halved again for ping-pong) with **no floor at the largest parameter**. A parameter that does not fit a single buffer cannot be packed at all, and the refit aborted on an assertion.

DeepSeek-V3 hits it. `model.embed_tokens.weight` is 129280 x 7168 bf16 = 1.73 GiB against a 1.65 GiB buffer — only 4.6% over:

```
AssertionError: Parameter model.embed_tokens.weight too large for buffer: 1853358080 > 1771261132
```

raised from `MegatronPolicyWorker.stream_weights_via_ipc_zmq`. It kills every colocated DSv3 run during the **first refit**, i.e. long after a 32-node allocation has been paid for. This is pre-existing on `main` and unrelated to any vLLM version.

## The fix

Ship such a parameter on its own in a buffer sized to fit: flush whatever group is pending, send the oversized parameter alone, and consume its ACK before freeing the one-off buffer — unlike the ping-pong pair, it is not kept alive across the next send.

This mirrors the **HTTP streaming path**, which already gives an oversized parameter a bucket of its own (`if bucket and bucket_size + param_size >= buffer_size_bytes`) instead of raising. The two sibling implementations now agree.

Deliberately **not** fixed by raising the buffer size — neither by flooring the auto-computed value nor by setting `refit_buffer_size_gb` per recipe. The 70% of free memory that the buffer does not take is reserved for parameter all-gather across the expert-parallel dimension, so growing the buffer trades a clean assertion for an OOM, and a per-recipe constant has to be re-tuned per model and cluster.

## Blast radius: none for working refits

The new branch is reachable only where the old code raised `AssertionError` — a guaranteed failure. Any refit that works today takes byte-for-byte the same path. `test_parameters_that_fit_are_still_batched` pins that.

## Validation

`tests/unit/models/policy/test_refit_oversized_param.py` — 5 tests, ~1.5s, **no GPU and no Ray**: the packing and hand-off logic is plain Python around a byte buffer, so CPU tensors and a stub socket exercise it faithfully.

Against the **unpatched** code the suite splits exactly as intended, so it is not vacuous:

| Test | Unpatched | Patched |
|---|---|---|
| `test_oversized_parameter_is_streamed_instead_of_aborting` | ❌ `AssertionError: Parameter model.embed_tokens.weight too large for buffer: 3072 > 2048` | ✅ |
| `test_oversized_parameter_alone` | ❌ | ✅ |
| `test_consecutive_oversized_parameters` | ❌ | ✅ |
| `test_parameters_that_fit_are_still_batched` | ✅ | ✅ |
| `test_alignment_is_what_decides_oversized` | ✅ | ✅ |

The failure signature matches production exactly, at reduced scale.

Writing the harness surfaced one protocol detail worth recording: the receiver ACKs the end-of-stream `COMPLETE` marker as well as each data group, so a fake socket that only ACKs data groups deadlocks the streamer.

## Not covered here

**Not yet validated on a real refit.** That needs a cluster allocation. DeepSeek-V3 is the natural end-to-end check, but it currently also needs the vLLM 0.25 node-spanning startup fixes from #3350 and #3355, so a green DSv3 run is not the success criterion for this PR on its own.

### DIFF
--- a/nemo_rl/models/policy/utils.py
+++ b/nemo_rl/models/policy/utils.py
@@ -443,9 +443,48 @@ def stream_weights_via_ipc_zmq_impl(
                 current_buffer = buffer_a
 
             aligned_size = calculate_aligned_size(tensor.nbytes)
-            assert aligned_size <= buffer_size_bytes, (
-                f"Parameter {name} too large for buffer: {aligned_size} > {buffer_size_bytes}"
-            )
+
+            # A parameter larger than a single staging buffer cannot be packed
+            # at all. Ship it on its own in a buffer sized to fit rather than
+            # failing the refit. The staging buffers are sized from *free
+            # memory* (NRL_REFIT_BUFFER_MEMORY_RATIO, default 0.3, halved again
+            # for ping-pong) with no floor at the largest parameter, so a big
+            # embedding can exceed one: DeepSeek-V3's model.embed_tokens.weight
+            # is 1.73 GiB against a 1.65 GiB buffer. This mirrors the HTTP
+            # streaming path, which already gives an oversized parameter a
+            # bucket of its own instead of raising.
+            if aligned_size > buffer_size_bytes:
+                if param_names:
+                    await_recv = send_buffer_group_overlap(
+                        current_buffer, param_names, used_bytes, await_recv
+                    )
+                    count_of_groups += 1
+                    current_buffer = (
+                        buffer_b if current_buffer is buffer_a else buffer_a
+                    )
+                    used_bytes, param_names = 0, []
+
+                oversized_buffer = torch.empty(
+                    aligned_size,
+                    device=current_buffer.device,
+                    dtype=torch.uint8,
+                    requires_grad=False,
+                )
+                try:
+                    packed_bytes = pack_tensor(oversized_buffer, tensor, 0)
+                    send_buffer_group_overlap(
+                        oversized_buffer, [name], packed_bytes, await_recv
+                    )
+                    count_of_groups += 1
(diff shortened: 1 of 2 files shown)


terrykong:
Closing as superseded — folded into #3280 (commit `dde85d331` on `terryk/bump-vllm-0.25.1`, now rebased onto `main`).

Rationale: validating DeepSeek-V3 requires the vLLM 0.25 node-spanning startup fixes (#3350, #3355) **and** this refit fix together. Keeping them on separate branches would have meant running the DSv3 perf suite twice, on two SKUs, for no added signal — so both live on one branch and get one validation cycle.

The change itself is unmodified from what was green here (90 pass / 0 fail): `stream_weights_via_ipc_zmq_impl` now ships a parameter larger than a single staging buffer on its own instead of asserting, plus `tests/unit/models/policy/test_refit_oversized_param.py` (5 tests, no GPU, 3 of which fail against the unpatched code with the production assertion at reduced scale).

Worth restating since it now rides a much larger PR: the new branch is reachable **only** where the old code raised `AssertionError`, so every refit that works today takes byte-for-byte the same path. The bug is pre-existing on `main` and unrelated to vLLM — if #3280 slips, this is the piece worth re-splitting out and landing on its own.

PR #3360, opened 2026-07-27
chore: unify setup nemogym

yuki-97:
1. cherry-pick 6afeaae to let `grpo.py` also use `spinup_nemo_gym_actor` in `nemo_rl.environments.nemo_gym`, which is missing in #3267.
2. add `routed_experts_dtype` and `use_fastokens` to include new changes in main branch.

### DIFF
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -613,6 +613,7 @@ def spinup_nemo_gym_actor(
     env_configs: dict[str, Any],
     base_urls: list[Optional[str]],
     model_name: str,
+    *,
     enable_router_replay: bool,
     routed_experts_dtype: str,
     use_fastokens: bool,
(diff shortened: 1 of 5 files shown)


nemo_rl/environments/nemo_gym.py
```diff
@@ -614,25 +614,29 @@ def spinup_nemo_gym_actor(
     base_urls: list[Optional[str]],
     model_name: str,
     enable_router_replay: bool,
+    routed_experts_dtype: str,
+    use_fastokens: bool,
 ) -> Any:
```
RayenTian, line 620: Not  a blocker, what about adding a bare  * after `model_name` to force pass parameters as keyword?
Note that the last four parameters are str, bool, str, bool in an alternating order, they can be passed either by position or by keyword.

Risk: If someone calls it positionally like this in the future:

`spinup_nemo_gym_actor(env_configs, base_urls, model_name, "int16", True, ...)`

it's easy to reverse the order of two bools (enable_router_replay / use_fastokens) or two strs (model_name / routed_experts_dtype).

RayenTian (dismissed):

RayenTian:
LGTM! Approved!

nemo_rl/environments/nemo_gym.py
yuki-97: fixed in 9bacae73744da398d14ff078ea7cebe3a851ba92

RayenTian (approved):

Issue #3365, opened 2026-07-27
NeMo-RL vs megatron-core `main`: current functional-test breakage


shanmugamr1992:
## L1 functional tests against megatron-core `main`

Last pass: 2026-08-06 00:30 UTC · 6/7 green.

This issue is maintained automatically and edited in place. It tracks breakage that is **not** caused by any particular Megatron-LM pull request, so that labeled PRs do not each have to rediscover it.

### Stack under test

- megatron-core: `main` @ `19c50e86`
- Megatron-Bridge: `2940a635`
- NeMo-RL: `main` plus the pending fixes below @ `86774472`

### Current status

| Test | Status | Assessment |
| --- | --- | --- |
| `grpo_megatron_generation` | ✅ pass | — |
| `grpo_megatron_generation_non_colocated` | ✅ pass | — |
| `grpo_megatron_generation_async` | ✅ pass | — |
| `grpo_megatron_generation_colocated_async` | ✅ pass | — |
| `grpo_megatron_generation_async_gym` | ✅ pass | — |
| `grpo_megatron_generation_topp_topk` | ❌ fail | max(data["train/token_mult_prob_error"]) < 1.06 (measured 1.078047513961792). Known issue `topp-topk-token-mult-prob-error-over-threshold` (see below). |
| `grpo_megatron_generation_colocated_gym` | ✅ pass | — |

### Diagnoses

**`topp-topk-token-mult-prob-error-over-threshold`** — 1 test(s): `grpo_megatron_generation_topp_topk`

Pre-existing on megatron-core main, not caused by any labeled PR. grpo_megatron_generation_topp_topk runs to completion but max(train/token_mult_prob_error) lands at 1.078047513961792 against a < 1.06 threshold. The value is bit-identical on Megatron-LM#5382 head eb01b689 and on main 3aee84c3 (both with the Bridge FSDP fix), so it is a deterministic train/generation logprob mismatch in the top-p/top-k sampling path rather than run-to-run noise. Marginal at ~1.7% over. No fix raised; needs an owner to decide whether the sampling path regressed or the threshold is stale. **No fix raised yet.**

PR #3366, opened 2026-07-27
fix: pin megatron non-colocated functional tests to the nccl refit backend

shanmugamr1992:
## Failure

Three L1 megatron functional tests fail against megatron-core `main`:

| Test | Symptom |
| --- | --- |
| `grpo_megatron_generation_non_colocated` | `RuntimeError: NVSHMEM_MAX_CTAS must be set to '2' ...` at startup |
| `grpo_megatron_generation_async` | same |
| `grpo_megatron_generation_async_gym` | survives the failed refit and fails `median(train/gen_kl_error) < 1.3` instead |

## Root cause

megatron-core `main` changed `NVSHMEM_MAX_CTAS` from a logged recommendation
into a hard error in
[`nvshmem_copy_service/core/gpu_resource_manager.py`](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/resharding/nvshmem_copy_service/core/gpu_resource_manager.py):

```python
max_ctas = os.environ.get("NVSHMEM_MAX_CTAS")
if max_ctas != "2":
    raise RuntimeError(...)
```

These tests inherit `mcore_generation_config.refit_backend: nvshmem` from
`grpo_math_1B.yaml`, and nothing exports that variable, so every non-colocated
megatron refit now fails.

The `async_gym` case is worth noting: it does not die, it continues with weights
that were never refit and then fails a divergence metric. The symptom points at
numerics, the cause is the refit.

## Fix

Name the backend explicitly in the three tests. `nccl` is fully supported —
`refit_backend: Literal["gloo", "nccl", "nvshmem"]` with an implemented branch in
`megatron_worker.py`; only the YAML comment is stale about it.

A functional test that inherits its refit backend from a shared default silently
changes what it covers whenever that default moves, which is how this went
unnoticed until now.

## What this deliberately does not fix

`grpo_math_1B.yaml` still ships `refit_backend: nvshmem`, so non-colocated
megatron refit remains broken against megatron-core `main` for anyone using the
default. That needs either `NVSHMEM_MAX_CTAS=2` in the worker environment or a
change of default, and that call belongs to the refit owners rather than to a
test-only change. Note the nvshmem refit path loses functional coverage until it
is made to work.

## Validation

Found by the standing NeMo-RL-vs-megatron-core-`main` watch (#3365). The three
tests are being re-run with this branch applied; result will be posted here.


shanmugamr1992:
## Validated on GB200 against megatron-core `main` (`076f61fc`)

All three tests pass with this branch applied:

| Test | Before | After |
| --- | --- | --- |
| `grpo_megatron_generation_non_colocated` | ❌ `NVSHMEM_MAX_CTAS` | ✅ pass (149s) |
| `grpo_megatron_generation_async` | ❌ `NVSHMEM_MAX_CTAS` | ✅ pass (106s) |
| `grpo_megatron_generation_async_gym` | ❌ `median(train/gen_kl_error) < 1.3` | ✅ pass (262s) |

`async_gym` recovering confirms the diagnosis: its metric failure was the
downstream effect of generating with weights that were never refit, not a
numerics regression of its own.

Run alongside #3363, which these tests also need — on this branch alone they hit
the separate `prompt_tokens=None` failure that #3363 fixes. Neither PR is
sufficient by itself.

PR #3367, opened 2026-07-27
refactor(nemo-gym): centralize actor construction

ananthsub:
## Summary

Centralizes construction of the `NemoGym` Ray actor in `nemo_rl/environments/nemo_gym.py`.

`grpo.py`, `distillation.py`, and `single_controller_utils/setup.py` previously assembled the actor configuration independently. Those paths could diverge in how they handled NeMo RL integration keys, uv directories, router replay, routed-expert dtype, and `use_fastokens`.

The shared implementation now separates those responsibilities:

- `build_nemo_gym_config()` removes NeMo RL integration settings from the configuration forwarded to Gym and applies model-dependent rollout settings.
- `spinup_nemo_gym_actor()` creates the actor runtime environment, preserves existing placement behavior, starts Gym, and installs the tokenizer.
- SingleController accepts deferred model-server URLs while generation starts concurrently.
- GRPO and SingleController now set `VIRTUAL_ENV` and `UV_PROJECT_ENVIRONMENT` to the virtual-environment root through `make_actor_runtime_env()`, matching the directory values already used by distillation.
- Distillation now promotes `tokenizer_config` and `pad_dynamic_image_shapes` into the top-level `NemoGymConfig` fields consumed by the actor. This enables the VLM distillation path, although no current in-tree distillation recipe sets `is_vlm: true`.

## Stack

This is the first PR in the NeMo-Gym actor-sharding stack. Actor lifetime fixes follow in #3368.

### DIFF
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -31,7 +31,6 @@ from nemo_rl.data.multimodal_utils import (
     media_sources_equal,
     uses_image_placeholder,
 )
-from nemo_rl.distributed.ray_actor_environment_registry import get_actor_python_env
 from nemo_rl.distributed.virtual_cluster import (
     DEFAULT_GYM_PORT_RANGE_HIGH,
     DEFAULT_GYM_PORT_RANGE_LOW,
@@ -54,7 +53,9 @@ from nemo_rl.models.generation.interfaces import should_use_async_rollouts
 from nemo_rl.models.policy import PolicyConfig, TokenizerConfig
 from nemo_rl.utils.routed_experts_codec import decode_routed_experts
 from nemo_rl.utils.timer import Timer
-from nemo_rl.utils.venvs import create_local_venv_on_each_node
+from nemo_rl.utils.venvs import make_actor_runtime_env
+
+NEMO_GYM_ACTOR_FQN = "nemo_rl.environments.nemo_gym.NemoGym"
 
 # Kept local so the Gym actor does not depend on model-config dtype resolution.
 # Must cover every name resolve_routed_experts_dtype can produce.
@@ -961,21 +962,19 @@ def setup_nemo_gym_config(config, tokenizer) -> None:
         env_cfg.setdefault("tokenizer_config", dict(config.policy["tokenizer"]))
 
 
-def spinup_nemo_gym_actor(
+def build_nemo_gym_config(
     env_configs: dict[str, Any],
-    base_urls: list[str],
-    model_name: str,
     *,
-    tokenizer: PreTrainedTokenizerBase,
-    enable_router_replay: bool,
-    routed_experts_dtype: str,
-    use_fastokens: bool,
-) -> Any:
-    """Spin up the NeMo-Gym actor against the given generation server URLs.
+    base_urls: list[Optional[str]],
+    model_name: str,
+    enable_router_replay: bool = False,
+    use_fastokens: bool = False,
+) -> NemoGymConfig:
+    """Build the ``NemoGymConfig`` for a NeMo-Gym actor.
 
-    When env_configs["nemo_gym"]["num_gpu_nodes"] > 0, the actor is scheduled
-    with soft NodeAffinity to the current Ray node so its colocated GPU
-    resources land where the caller expects.
+    Splits ``env_configs["nemo_gym"]`` into the NeMo-RL-side fields the actor
+    reads directly and the remainder, which is forwarded verbatim as NeMo-Gym's
+    initial global config.
 
     Args:
         env_configs: The master_config.env mapping; env_configs["nemo_gym"] supplies
@@ -983,17 +982,10 @@ def spinup_nemo_gym_actor(
             thinking_tags, num_gpu_nodes).
         base_urls: Per-DP-rank OpenAI-compatible server base URLs from the generation backend.
         model_name: Served model name the Gym rollouts should target.
-        tokenizer: Installed on the actor once, here, rather than passed per
-            rollout call. See NemoGym.set_tokenizer for why that distinction is
-            the difference between a working run and a stalled one.
-        enable_router_replay: Sets require_routed_experts on the NemoGymConfig.
-        routed_experts_dtype: Dtype name for R3 routed_experts tensors ("int8"/"int16"/"int32"),
-            resolved by the caller from the model's expert count.
-        use_fastokens: Forwarded from policy.tokenizer.use_fastokens so the rollout actor
-            patches its tokenizer consistently with the driver.
-
-    Returns:
-        The spun-up NemoGym Ray actor handle (_spinup already awaited).
+        enable_router_replay: Sets ``require_routed_experts`` and selects the
+            routed-experts carry dtype ("int8"/"int16"/"int32") for the model.
+        use_fastokens: Forwarded from ``policy.tokenizer.use_fastokens`` so the
+            actor patches its tokenizer the same way the driver does.
     """
     nemo_gym_dict = dict(env_configs["nemo_gym"])
 
@@ -1020,7 +1012,16 @@ def spinup_nemo_gym_actor(
     if uv_venv_dir is not None:
         nemo_gym_dict.setdefault("uv_venv_dir", uv_venv_dir)
 
-    nemo_gym_cfg = NemoGymConfig(
+    routed_experts_dtype = "int16"
+    if enable_router_replay:
+        # Deferred so the actor module stays free of generation-package imports.
+        from nemo_rl.models.generation.interfaces import (
+            resolve_routed_experts_dtype_name_for_model,
+        )
+
+        routed_experts_dtype = resolve_routed_experts_dtype_name_for_model(model_name)
+
+    return NemoGymConfig(
         model_name=model_name,
         base_urls=base_urls,
         invalid_tool_call_patterns=invalid_tool_call_patterns,
@@ -1033,26 +1034,47 @@ def spinup_nemo_gym_actor(
         **multimodal_flags,
     )
 
-    nemo_gym_py_exec = get_actor_python_env("nemo_rl.environments.nemo_gym.NemoGym")
-    if nemo_gym_py_exec.startswith("uv"):
-        nemo_gym_py_exec = create_local_venv_on_each_node(
-            nemo_gym_py_exec, "nemo_rl.environments.nemo_gym.NemoGym"
-        )
 
-    nemo_gym_opts: dict[str, Any] = {}
-    if nemo_gym_dict.get("num_gpu_nodes", 0):
+def spinup_nemo_gym_actor(
+    env_configs: dict[str, Any],
+    *,
+    base_urls: list[Optional[str]],
+    model_name: str,
+    tokenizer: PreTrainedTokenizerBase,
+    enable_router_replay: bool = False,
+    use_fastokens: bool = False,
+) -> Any:
+    """Spin up the NeMo-Gym actor against the given generation server URLs.
+
+    When ``env_configs["nemo_gym"]["num_gpu_nodes"] > 0``, the actor is
+    scheduled with soft NodeAffinity to the caller's Ray node so its colocated
+    GPU resources land where the caller expects.
+
+    Args:
+        tokenizer: Installed on the actor once, here, rather than passed per
+            rollout call. See ``NemoGym.set_tokenizer`` for why that
+            distinction is the difference between a working run and a stalled
+            one.
+
+    Returns:
+        The spun-up ``NemoGym`` Ray actor handle (``_spinup`` already awaited).
+    """
+    nemo_gym_cfg = build_nemo_gym_config(
+        env_configs,
+        base_urls=base_urls,
+        model_name=model_name,
+        enable_router_replay=enable_router_replay,
+        use_fastokens=use_fastokens,
+    )
+
+    nemo_gym_opts: dict[str, Any] = {
+        "runtime_env": make_actor_runtime_env(NEMO_GYM_ACTOR_FQN)
+    }
+    if env_configs["nemo_gym"].get("num_gpu_nodes", 0):
         nemo_gym_opts["scheduling_strategy"] = NodeAffinitySchedulingStrategy(
             node_id=ray.get_runtime_context().get_node_id(),
             soft=True,
         )
-    nemo_gym_opts["runtime_env"] = {
-        "py_executable": nemo_gym_py_exec,
-        "env_vars": {
-            **os.environ,
-            "VIRTUAL_ENV": nemo_gym_py_exec,
-            "UV_PROJECT_ENVIRONMENT": nemo_gym_py_exec,
-        },
-    }
 
     actor = NemoGym.options(**nemo_gym_opts).remote(nemo_gym_cfg)
     ray.get(actor._spinup.remote())
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -49,7 +49,10 @@ from nemo_rl.experience.failures import (
     RolloutDataFailure,
     http_status_is_infra,
 )
-from nemo_rl.models.generation.interfaces import should_use_async_rollouts
+from nemo_rl.models.generation.interfaces import (
+    resolve_routed_experts_dtype_name_for_model,
+    should_use_async_rollouts,
+)
 from nemo_rl.models.policy import PolicyConfig, TokenizerConfig
 from nemo_rl.utils.routed_experts_codec import decode_routed_experts
 from nemo_rl.utils.timer import Timer
@@ -965,10 +968,10 @@ def setup_nemo_gym_config(config, tokenizer) -> None:
 def build_nemo_gym_config(
     env_configs: dict[str, Any],
     *,
-    base_urls: list[Optional[str]],
+    base_urls: list[str],
     model_name: str,
-    enable_router_replay: bool = False,
-    use_fastokens: bool = False,
+    enable_router_replay: bool,
+    use_fastokens: bool,
 ) -> NemoGymConfig:
     """Build the ``NemoGymConfig`` for a NeMo-Gym actor.
 
@@ -986,6 +989,11 @@ def build_nemo_gym_config(
             routed-experts carry dtype ("int8"/"int16"/"int32") for the model.
         use_fastokens: Forwarded from ``policy.tokenizer.use_fastokens`` so the
             actor patches its tokenizer the same way the driver does.
+
+    Returns:
+        A ``NemoGymConfig`` with NeMo-RL fields at the top level and the
+        remaining ``env_configs["nemo_gym"]`` keys under
+        ``initial_global_config_dict``. The caller's ``env_configs`` is not mutated.
     """
     nemo_gym_dict = dict(env_configs["nemo_gym"])
 
@@ -1012,14 +1020,11 @@ def build_nemo_gym_config(
     if uv_venv_dir is not None:
         nemo_gym_dict.setdefault("uv_venv_dir", uv_venv_dir)
 
-    routed_experts_dtype = "int16"
-    if enable_router_replay:
-        # Deferred so the actor module stays free of generation-package imports.
-        from nemo_rl.models.generation.interfaces import (
-            resolve_routed_experts_dtype_name_for_model,
-        )
-
-        routed_experts_dtype = resolve_routed_experts_dtype_name_for_model(model_name)
+    routed_experts_dtype = (
+        resolve_routed_experts_dtype_name_for_model(model_name)
+        if enable_router_replay
+        else "int16"
+    )
 
     return NemoGymConfig(
         model_name=model_name,
@@ -1038,11 +1043,11 @@ def build_nemo_gym_config(
 def spinup_nemo_gym_actor(
     env_configs: dict[str, Any],
     *,
-    base_urls: list[Optional[str]],
+    base_urls: list[str],
     model_name: str,
     tokenizer: PreTrainedTokenizerBase,
-    enable_router_replay: bool = False,
-    use_fastokens: bool = False,
+    enable_router_replay: bool,
+    use_fastokens: bool,
 ) -> Any:
     """Spin up the NeMo-Gym actor against the given generation server URLs.
 
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -25,7 +25,6 @@ import torch
 from ray.util.scheduling_strategies import NodeAffinitySchedulingStrategy
 from transformers import PreTrainedTokenizerBase
 
-from nemo_rl.distributed.ray_actor_environment_registry import get_actor_python_env
 from nemo_rl.distributed.virtual_cluster import (
     DEFAULT_GYM_PORT_RANGE_HIGH,
     DEFAULT_GYM_PORT_RANGE_LOW,
@@ -34,7 +33,9 @@ from nemo_rl.distributed.virtual_cluster import (
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
 from nemo_rl.utils.timer import Timer
-from nemo_rl.utils.venvs import create_local_venv_on_each_node
+from nemo_rl.utils.venvs import make_actor_runtime_env
+
+NEMO_GYM_ACTOR_FQN = "nemo_rl.environments.nemo_gym.NemoGym"
 
 # Kept local (not imported from models.generation) so the gym actor stays free of
 # generation-module imports. Must cover every name resolve_routed_experts_dtype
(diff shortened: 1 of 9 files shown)


nemo_rl/environments/nemo_gym.py
```diff
@@ -1187,7 +1179,16 @@ def spinup_nemo_gym_actor(
...
 
-    nemo_gym_cfg = NemoGymConfig(
+    routed_experts_dtype = "int16"
+    if enable_router_replay:
+        # Deferred so the actor module stays free of generation-package imports.
+        from nemo_rl.models.generation.interfaces import (
+            resolve_routed_experts_dtype_name_for_model,
+        )
+
+        routed_experts_dtype = resolve_routed_experts_dtype_name_for_model(model_name)
```
terrykong: **1 action item.** PR-introduced.

**TL;DR — the deferred import buys nothing (this module already imports from `nemo_rl.models.generation.interfaces` at the top), and hoisting it silently breaks an existing test unless the patch target moves in the same commit.**

The comment says the import is deferred so the actor module stays free of generation-package imports, but [line 49](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L49) already does `from nemo_rl.models.generation.interfaces import should_use_async_rollouts` at module scope. There is no import cycle to dodge — line 49 working today proves it. The repo's [linting-and-formatting skill](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/.agents/contributor-skills/linting-and-formatting/SKILL.md#L118-L122) says it directly: "An in-function `import` with no such reason should move to the top. In particular, deferring stdlib (`concurrent.futures`, `collections`, …) or a module that is *already* imported at module top buys nothing — hoist it."

The test edit is not optional. [`test_build_nemo_gym_config_router_replay_resolves_dtype`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/environments/test_nemo_gym_utils.py#L188) patches the name on the *source* module, which only works while the import is deferred. Once hoisted, `nemo_gym` holds its own reference, the patch stops taking effect, and `patch()` raises nothing — see the fold for what the test then does instead.

### AI-1

Hoist the import and drop the comment; move the test's patch target in the same commit.

Anchored region (lines 1182-1189):

```suggestion
    routed_experts_dtype = (
        resolve_routed_experts_dtype_name_for_model(model_name)
        if enable_router_replay
        else "int16"
    )
```

The other two edits are outside this PR's diff or in another file, so GitHub will not take a suggestion block for them.

At [`nemo_gym.py` line 49](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L49):

```python
from nemo_rl.models.generation.interfaces import (
    resolve_routed_experts_dtype_name_for_model,
    should_use_async_rollouts,
)
```

At [`test_nemo_gym_utils.py` lines 188-191](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/environments/test_nemo_gym_utils.py#L188-L191):

```python
    with patch.object(
        nemo_gym_mod,
        "resolve_routed_experts_dtype_name_for_model",
        return_value="int8",
    ) as mock_resolve:
```

<details><summary>Why the patch target has to change with the hoist</summary>

Four combinations, run against two toy modules:

| import style | patch form | result |
|---|---|---|
| deferred | `patch("pkg.mod.name")` | patched (today's behavior) |
| deferred | `patch.object(mod, "name")` | `AttributeError` |
| hoisted | `patch("pkg.mod.name")` | **not** patched, and `patch()` raises nothing |
| hoisted | `patch.object(mod, "name")` | patched |

That is what makes the two edits strictly paired: `patch.object` raises `AttributeError` today, before the hoist, and `patch("nemo_rl.models.generation.interfaces.resolve_routed_experts_dtype_name_for_model")` stops intercepting after it — with no error of any kind. The real resolver then runs and calls `AutoConfig.from_pretrained("test-model", trust_remote_code=True)` ([interfaces.py line 96](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/models/generation/interfaces.py#L96)) — a network call from a unit test. Its own `except (OSError, ValueError)` swallows the failure and returns the fallback `"int16"`, so the test fails on `cfg["routed_experts_dtype"] == "int8"`, which names nothing about the real cause.

</details>

tests/unit/environments/test_nemo_gym_utils.py
```diff
@@ -102,3 +108,145 @@ def test_get_nemo_gym_uv_cache_dir_uses_uv_inside_container(monkeypatch):
...
     assert get_nemo_gym_uv_cache_dir() == "/root/.cache/uv"
+
+
+def _env_configs(**overrides):
+    nemo_gym = {
+        "num_gpu_nodes": 1,
+        "invalid_tool_call_patterns": ["bad_call"],
+        "thinking_tags": ["<think>"],
+        "config_paths": ["gym.yaml"],
+    }
```
terrykong, line 121: **1 action item.** PR-introduced gap in the new tests.

**TL;DR — nothing asserts that `tokenizer_config` and `pad_dynamic_image_shapes` get promoted to top-level `NemoGymConfig` fields, so the exact behavior this PR fixes for distillation can be re-broken with the suite still green.**

`build_nemo_gym_config` pops [`tokenizer_config`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1163) and [`pad_dynamic_image_shapes`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1167-L1172) out of Gym's global config and sets them as top-level fields. That promotion matters: the actor reads `tokenizer_config` as a [top-level key](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L499), and [`_spinup`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L564) strips `pad_dynamic_image_shapes` from Gym's global config but not `tokenizer_config`. The new [`_env_configs()` fixture](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/environments/test_nemo_gym_utils.py#L113) contains neither key, so the promotion is never exercised.

Setting `tokenizer_config=None` and dropping `**multimodal_flags` from the return — re-breaking exactly what this PR fixes — still gives "21 passed". With the assertions below it gives "1 failed, 20 passed". [`test_nemo_gym_image_placeholders.py`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/environments/test_nemo_gym_image_placeholders.py#L79) does not cover it either: it builds `NemoGymConfig` directly with the field already set, so it never goes through `build_nemo_gym_config`.

### AI-1

The two edits split the work: the fixture change alone already covers the "are they stripped from Gym's global config" half, because the exact-equality assertion on [`initial_global_config_dict`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/environments/test_nemo_gym_utils.py#L150) starts failing the moment someone stops popping the keys; the assertion change adds the "do they arrive as top-level fields" half. Apply the fixture first, since the assertions on their own would fail against the current fixture.

Fixture (lines 114-119):

```suggestion
    nemo_gym = {
        "num_gpu_nodes": 1,
        "invalid_tool_call_patterns": ["bad_call"],
        "thinking_tags": ["<think>"],
        "tokenizer_config": {"name": "test-tokenizer"},
        "pad_dynamic_image_shapes": True,
        "config_paths": ["gym.yaml"],
    }
```

Then, at [line 149](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/environments/test_nemo_gym_utils.py#L149) — a different region, so not a suggestion block:

```python
    assert cfg["thinking_tags"] == ["<think>"]
    # VLM/tokenizer knobs are promoted too -- the actor reads them as top-level
    # fields, so leaving them in Gym's global config would silently drop them.
    assert cfg["tokenizer_config"] == {"name": "test-tokenizer"}
    assert cfg["pad_dynamic_image_shapes"] is True
```

nemo_rl/environments/nemo_gym.py
```diff
@@ -1128,39 +1129,30 @@ def setup_nemo_gym_config(config, tokenizer) -> None:
...
-    base_urls: list[str],
-    model_name: str,
     *,
-    tokenizer: PreTrainedTokenizerBase,
-    enable_router_replay: bool,
-    routed_experts_dtype: str,
-    use_fastokens: bool,
-) -> Any:
-    """Spin up the NeMo-Gym actor against the given generation server URLs.
+    base_urls: list[Optional[str]],
```
terrykong: **1 action item.** PR-introduced, low severity.

**TL;DR — the widened `base_urls` accepts a value the body cannot handle: a `None` would go straight into a `List[str]` field and on to Gym, which rejects it — and on this path a `None` cannot occur in the first place.**

The reason given for widening does not hold. The commit message says the list "can contain None until the servers finish loading", but the deferred path fills it from reserved ports *before* any model loads, and the class docstring says so outright:

> `dp_openai_server_base_urls` is populated immediately from reserved ports

[vllm_generation.py:141](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/models/generation/vllm/vllm_generation.py#L141). Ports are bound in [`__init__`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/models/generation/vllm/vllm_worker_async.py#L122), and [`get_reserved_url`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/models/generation/vllm/vllm_worker_async.py#L343) is documented as "available before model loading". The only `[None]` return in [`_collect_reserved_urls`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/models/generation/vllm/vllm_generation.py#L548) is the sync-engine branch, which [`should_use_nemo_gym` asserts against](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L91).

So `Optional` here advertises a state that cannot happen, and if it did, Gym's `base_url` field would kill the server at startup rather than degrade.

### AI-1

Narrow back to `list[str]` rather than widening [`NemoGymConfig.base_urls`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L190) to match — widening the field would record the opposite, that NeMo-Gym accepts entries with no URL.

```suggestion
    base_urls: list[str],
```

Same change on [`spinup_nemo_gym_actor`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1208) and [`_spinup_gym`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/single_controller_utils/setup.py#L644) — both are different regions, so they cannot be suggestion blocks here.

<details><summary>Where the wider type is really coming from</summary>

Megatron already reports this the narrow way: it drops the `None`s and returns a possibly-empty list, which lets the line above it use a plain truth test.

```python
if not self.dp_openai_server_base_urls and ...:      # L368
    ...
    self.dp_openai_server_base_urls = [
        url for url in ray.get(url_futures) if url is not None    # L374-376
    ]
```
[megatron_generation.py:368](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/models/generation/megatron/megatron_generation.py#L368), [:374-376](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/models/generation/megatron/megatron_generation.py#L374-L376)

vLLM, TRT-LLM and Dynamo instead keep a placeholder `None` entry, which is what makes the producer's declared type `Optional` even though no mixed list is ever produced. That is the root cause, and it is filed separately as [#3977](https://github.com/NVIDIA-NeMo/RL/issues/3977) since those files are outside this PR. Once it lands, the `list[Optional[str]]` on [`gym_base_urls`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/single_controller_utils/setup.py#L1197) can narrow too and the mismatch disappears at the source.

</details>

nemo_rl/environments/nemo_gym.py
```diff
@@ -1128,39 +1129,30 @@ def setup_nemo_gym_config(config, tokenizer) -> None:
...
-        use_fastokens: Forwarded from policy.tokenizer.use_fastokens so the rollout actor
-            patches its tokenizer consistently with the driver.
-
-    Returns:
-        The spun-up NemoGym Ray actor handle (_spinup already awaited).
+        enable_router_replay: Sets ``require_routed_experts`` and selects the
+            routed-experts carry dtype ("int8"/"int16"/"int32") for the model.
+        use_fastokens: Forwarded from ``policy.tokenizer.use_fastokens`` so the
+            actor patches its tokenizer the same way the driver does.
     """
```
terrykong, line 997: **1 action item.** Nit.

**TL;DR — [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1132)'s docstring has an `Args:` block but no `Returns:`, so what it hands back (and that it does not mutate the caller's config) is undocumented.**

See [lines 1146-1156](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1146-L1156). Ruff will not flag it — `D417` is in the ignore list at [pyproject.toml line 761](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/pyproject.toml#L761) — so this is convention, not a broken check.

### AI-1

```suggestion
        use_fastokens: Forwarded from ``policy.tokenizer.use_fastokens`` so the
            actor patches its tokenizer the same way the driver does.

    Returns:
        A ``NemoGymConfig`` with the NeMo-RL-side fields set at top level and the
        remaining ``env_configs["nemo_gym"]`` keys under
        ``initial_global_config_dict``. The caller's ``env_configs`` is not mutated.
    """
```

**Context — no action.** [`spinup_nemo_gym_actor`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1205)'s docstring documenting only `tokenizer` and leaving the config args to [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1132) two lines below is fine as is. Copying that `Args` list into the wrapper would recreate the duplication this PR removes.

nemo_rl/environments/nemo_gym.py
```diff
@@ -1200,26 +1201,47 @@ def spinup_nemo_gym_actor(
...
-    nemo_gym_opts: dict[str, Any] = {}
-    if nemo_gym_dict.get("num_gpu_nodes", 0):
+def spinup_nemo_gym_actor(
+    env_configs: dict[str, Any],
+    *,
+    base_urls: list[Optional[str]],
+    model_name: str,
+    tokenizer: PreTrainedTokenizerBase,
+    enable_router_replay: bool = False,
+    use_fastokens: bool = False,
```
terrykong: **1 action item.** PR-introduced.

**TL;DR — neither default is ever used: all three call sites pass `use_fastokens`, and only distillation omits `enable_router_replay` — so making them required again costs one line and closes the way a future call site silently loses router replay.**

Both were required keyword arguments before this PR ([base](https://github.com/NVIDIA-NeMo/RL/blob/e9a855ff4bc0a41967e9491c41de093527e123b6/nemo_rl/environments/nemo_gym.py#L1131-L1140)). Who relies on the defaults today:

| default | relied on by |
|---|---|
| `use_fastokens` | nobody — [grpo](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/grpo.py#L825), [distillation](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/distillation.py#L528) and [SC](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/single_controller_utils/setup.py#L668) all pass it |
| `enable_router_replay` | only [distillation](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/distillation.py#L523-L529) |
| both, on [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1137-L1138) | nobody — its [only caller](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1229-L1235) passes both |

A call site that forgets `enable_router_replay` gets an actor with `require_routed_experts=False`, so rollouts come back with no routed experts and [the check that would catch it](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L878) never runs. Nothing raises.

### AI-1

Drop the defaults on both functions and have distillation pass `enable_router_replay=False` explicitly — which also puts its existing comment ("Distillation does not configure vLLM for router replay") next to the argument it explains rather than next to an absence.

```suggestion
    enable_router_replay: bool,
    use_fastokens: bool,
```

The matching pair on [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1137-L1138) and the one-line addition at the [distillation call](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/distillation.py#L523-L529) are different regions, so they cannot be suggestion blocks here.

terrykong (commented):
Three call sites collapse into one factory, and the refactor holds up — this is real dedup, not code moved around. Distillation drops ~40 hand-rolled lines, and the routed-experts dtype resolution that was copy-pasted in [`grpo.py`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/grpo.py#L819) and [`setup.py`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/single_controller_utils/setup.py#L662) now lives in one place. Nothing here blocks the merge.

The new tests in [`test_nemo_gym_utils.py`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/environments/test_nemo_gym_utils.py#L135) are a genuine improvement. The coverage migration is complete, and it adds three cases that existed nowhere before: the `num_gpu_nodes == 0` branch, dtype resolution with router replay on, and the default to `int16` with it off. The [`ray.get` ordering assertion](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/environments/test_nemo_gym_utils.py#L252) is stricter than the one it replaces. [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1157) also copies `env_configs["nemo_gym"]` rather than mutating the caller's config, and there is a test pinning that.

Two behavior changes are not mentioned in the PR description. Both look like fixes; no code change needed, but each deserves a line in the description:

1. `VIRTUAL_ENV` / `UV_PROJECT_ENVIRONMENT` for the grpo and SingleController Gym actors now point at the venv root instead of the python binary, because those paths moved onto [`make_actor_runtime_env`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/utils/venvs.py#L217). Distillation already worked this way, and every consumer in the tree treats these as directories, so the new value is the right one.
2. Distillation now gets `tokenizer_config` and `pad_dynamic_image_shapes` promoted to top-level `NemoGymConfig` fields for the first time, which fixes VLM distillation. To be exact: this was not broken in production — no in-tree config exercises it. All three `is_vlm: true` configs under `examples/configs/recipes/vlm/` are GRPO recipes, and no distillation config sets it. So it fixes a path nothing currently runs.

Environment caveat, stated once: this review is read from source at the pinned SHAs, with no GPU. The exception is the unit tests, which were actually run in a hand-built CPU-only venv — 21 passed in `test_nemo_gym_utils.py`, 22 in `test_distillation.py`, 16 selected in `test_grpo.py`, and 68 in `test_setup.py` with one pre-existing failure unrelated to this PR ([`test_single_controller_mopd_recipe_resolves_to_runtime_contract`](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/tests/unit/single_controller/test_setup.py#L1107), which dies on an unset `HF_HOME`). Ruff is clean: the pinned 0.9.9 hooks (`check`, `check --select I`, `format --check`) all pass. CI is 60/60 green.

_Generated by Claude Code_

nemo_rl/environments/nemo_gym.py
ananthsub: Addressed. The resolver import is now at module scope, and the unit test patches the reference owned by `nemo_gym`.

tests/unit/environments/test_nemo_gym_utils.py
ananthsub: Addressed. The fixture now includes both fields, and the test verifies that they are removed from Gym global config and promoted to top-level `NemoGymConfig` fields.

nemo_rl/environments/nemo_gym.py
ananthsub: Addressed. The actor factory and SingleController helper now accept `list[str]`; the existing producer-side optional type is narrowed only after the path has established that the reserved URLs are present.

nemo_rl/environments/nemo_gym.py
ananthsub: Addressed. The docstring now documents the returned config shape and that the input mapping is not mutated.

nemo_rl/environments/nemo_gym.py
ananthsub: Addressed. Both behavior flags are required throughout the factory path, and distillation now passes `enable_router_replay=False` explicitly.

nemo_rl/environments/nemo_gym.py
terrykong: fixed in 8aa1cb973d8037af0639902f3b138c4008748506

tests/unit/environments/test_nemo_gym_utils.py
terrykong: fixed in 8aa1cb973d8037af0639902f3b138c4008748506

nemo_rl/environments/nemo_gym.py
terrykong: fixed in 8aa1cb973d8037af0639902f3b138c4008748506

nemo_rl/environments/nemo_gym.py
terrykong: fixed in 8aa1cb973d8037af0639902f3b138c4008748506

nemo_rl/environments/nemo_gym.py
terrykong: fixed in 8aa1cb973d8037af0639902f3b138c4008748506

terrykong (approved):

PR #3368, opened 2026-07-27
fix(nemo-gym): define actor failure and teardown ownership

ananthsub:
## Summary

Defines clear ownership and failure behavior for `NemoGym` before one job starts multiple actors.

- Configures actors with `max_restarts=0` and `max_task_retries=0`. Ray reruns `__init__` after an actor restart but does not rerun the deferred `_spinup()` method that creates the Gym servers, so a replacement actor cannot serve rollouts.
- Clears the `RunHelper` handle before invoking its non-idempotent shutdown method. Repeated cleanup and cleanup before spinup remain safe.
- Adds `shutdown_environments()`. It deduplicates aliased train and validation handles, waits for graceful shutdown under a bounded timeout, and kills actors that do not stop.
- Runs environment shutdown from `finally` blocks in synchronous GRPO and distillation.
- Keeps one shutdown owner on each GRPO path. The training function owns legacy synchronous and asynchronous training cleanup. The runner retains cleanup for TransferQueue synchronous training and trajectory-only collection.

These guarantees apply to the existing single-actor path and prevent sharding from multiplying leaked actors or leaving restarted actors without Gym server processes.

## Stack

This PR depends on #3367. Resolved-config introspection follows in #3633.

### DIFF
--- a/examples/run_grpo.py
+++ b/examples/run_grpo.py
@@ -21,8 +21,8 @@ from omegaconf import OmegaConf
 
 from nemo_rl.algorithms.grpo import (
     MasterConfig,
+    grpo_train,
     setup,
-    shutdown_environments,
 )
 from nemo_rl.algorithms.utils import get_tokenizer
 from nemo_rl.data.utils import setup_response_data
@@ -32,6 +32,7 @@ from nemo_rl.data_plane.factory import (
     select_sync_trainer,
 )
 from nemo_rl.distributed.virtual_cluster import init_ray
+from nemo_rl.environments.utils import shutdown_environments
 from nemo_rl.models.generation import configure_generation_config
 from nemo_rl.telemetry.setup import init_telemetry_driver, shutdown_telemetry
 from nemo_rl.utils.config import (
@@ -162,6 +163,7 @@ def main() -> None:
                 print(f"  {label}: {value:.1f}s")
         print("=" * 60 + "\n", flush=True)
 
+        trainer_owns_environment_shutdown = False
         try:
             # Check if async mode is enabled
             if config.grpo.async_grpo.enabled:
@@ -211,6 +213,7 @@ def main() -> None:
                 # Two parallel synchronous trainers (verl-style — main_ppo.py vs
                 # main_ppo_sync.py). data_plane.enabled selects which one runs.
                 trainer = select_sync_trainer(master_config)
+                trainer_owns_environment_shutdown = trainer is grpo_train
                 # grpo_train_sync defers checkpoint finalization to the checkpointer's
                 # background threads; the context manager guarantees they are flushed on
                 # exit. (grpo_train also flushes internally; shutdown() is idempotent.)
@@ -230,7 +233,8 @@ def main() -> None:
                         master_config,
                     )
         finally:
-            shutdown_environments(task_to_env, val_task_to_env)
+            if not trainer_owns_environment_shutdown:
+                shutdown_environments(task_to_env, val_task_to_env)
             try:
                 policy_generation.shutdown()
             except Exception as error:
--- a/nemo_rl/algorithms/distillation.py
+++ b/nemo_rl/algorithms/distillation.py
@@ -52,6 +52,7 @@ from nemo_rl.environments.nemo_gym import (
     should_use_nemo_gym,
     spinup_nemo_gym_actor,
 )
+from nemo_rl.environments.utils import shutdown_environments
 from nemo_rl.experience.rollouts import (
     run_async_multi_turn_rollout,
     run_multi_turn_rollout,
@@ -638,8 +639,7 @@ def setup(
 # ===============================================================================
 
 
-@trace_fn(RLSpanGroup.JOB, "rl.distillation.job")
-def distillation_train(
+def _distillation_train_impl(
     student_policy: ColocatablePolicyInterface,
     teacher_policy: ColocatablePolicyInterface,
     student_generation: Optional[GenerationInterface],
@@ -1181,6 +1181,43 @@ def distillation_train(
     # without this the daemon finalization thread could be killed before the
     # final tmp_step_N is renamed.
     checkpointer.shutdown()
+
+
+@trace_fn(RLSpanGroup.JOB, "rl.distillation.job")
+def distillation_train(
+    student_policy: ColocatablePolicyInterface,
+    teacher_policy: ColocatablePolicyInterface,
+    student_generation: Optional[GenerationInterface],
+    dataloader: StatefulDataLoader,
+    val_dataloader: Optional[StatefulDataLoader],
+    tokenizer: TokenizerType,
+    loss_fn: DistillationLossFn,
+    task_to_env: dict[str, EnvironmentInterface],
+    val_task_to_env: Optional[dict[str, EnvironmentInterface]],
+    logger: Logger,
+    checkpointer: CheckpointManager,
+    distillation_save_state: DistillationSaveState,
+    master_config: MasterConfig,
+) -> None:
+    """Run distillation training and always tear down its environments."""
+    try:
+        _distillation_train_impl(
+            student_policy=student_policy,
+            teacher_policy=teacher_policy,
+            student_generation=student_generation,
+            dataloader=dataloader,
+            val_dataloader=val_dataloader,
+            tokenizer=tokenizer,
+            loss_fn=loss_fn,
+            task_to_env=task_to_env,
+            val_task_to_env=val_task_to_env,
+            logger=logger,
+            checkpointer=checkpointer,
+            distillation_save_state=distillation_save_state,
+            master_config=master_config,
+        )
+    finally:
+        shutdown_environments(task_to_env, val_task_to_env)
 
 
 def validate(
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -80,6 +80,7 @@ from nemo_rl.distributed.virtual_cluster import (
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
 from nemo_rl.environments.nemo_gym import spinup_nemo_gym_actor
+from nemo_rl.environments.utils import shutdown_environments
 from nemo_rl.experience.interfaces import (
     NEMO_GYM_TASK_INDEX_KEY,
     NEXT_NEMO_GYM_TASK_INDEX_KEY,
@@ -3465,11 +3466,13 @@ def grpo_train(
             total_steps += 1
             if should_save_by_timeout:
                 checkpointer.shutdown()
+                shutdown_environments(task_to_env, val_task_to_env)
                 memory_tracker.snapshot_start_of_stage("", dir())
                 print("Timeout has been reached, stopping training early", flush=True)
                 return
             if total_steps >= max_num_steps:
                 checkpointer.shutdown()
+                shutdown_environments(task_to_env, val_task_to_env)
                 memory_tracker.snapshot_start_of_stage("", dir())
                 print(
                     "Max number of steps has been reached, stopping training early",
@@ -3486,6 +3489,10 @@ def grpo_train(
     # so without this the daemon finalization thread would be killed before the
     # final tmp_step_N is renamed.
     checkpointer.shutdown()
+    # Environments own long-lived server subprocesses (NeMo-Gym runs ~60 per
+    # actor), so they have to be stopped on every exit from the synchronous
+    # trainer too, not just the async one.
+    shutdown_environments(task_to_env, val_task_to_env)
 
 
 def validate(
@@ -4894,21 +4901,7 @@ def async_grpo_train(
         except Exception as e:
             print(f"Error stopping replay buffer: {e}")
 
-        # Environments must be shut down before generation workers because
-        # they may have in-flight HTTP requests to vLLM HTTP endpoints.
-        # Killing generation first leaves environments retrying dead connections.
-        for env_dict in (task_to_env, val_task_to_env):
-            if env_dict is None:
-                continue
-            for task_name, env in env_dict.items():
-                print(f"🛑 Shutting down environment {task_name}...")
-                try:
-                    ray.get(env.shutdown.remote(), timeout=10)
-                except Exception:
-                    try:
-                        ray.kill(env)
-                    except Exception as e:
-                        print(f"Error shutting down environment {task_name}: {e}")
+        shutdown_environments(task_to_env, val_task_to_env)
 
         print("🛑 Shutting down generation workers...")
         try:
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -2868,8 +2868,7 @@ def _validation_early_stop_message(
     )
 
 
-@trace_fn(RLSpanGroup.JOB, "rl.grpo.job")
-def grpo_train(
+def _grpo_train_impl(
     policy: ColocatablePolicyInterface,
     policy_generation: Optional[GenerationInterface],
     wrapped_dataloader: StatefulDataLoader | MultipleDataloaderWrapper,
@@ -4040,13 +4039,11 @@ def grpo_train(
                 return
             if should_save_by_timeout:
                 checkpointer.shutdown()
-                shutdown_environments(task_to_env, val_task_to_env)
                 memory_tracker.snapshot_start_of_stage("", dir())
                 print("Timeout has been reached, stopping training early", flush=True)
                 return
             if total_steps >= max_num_steps:
                 checkpointer.shutdown()
-                shutdown_environments(task_to_env, val_task_to_env)
                 memory_tracker.snapshot_start_of_stage("", dir())
                 print(
                     "Max number of steps has been reached, stopping training early",
@@ -4063,10 +4060,43 @@ def grpo_train(
     # so without this the daemon finalization thread would be killed before the
     # final tmp_step_N is renamed.
     checkpointer.shutdown()
-    # Environments own long-lived server subprocesses (NeMo-Gym runs ~60 per
-    # actor), so they have to be stopped on every exit from the synchronous
-    # trainer too, not just the async one.
-    shutdown_environments(task_to_env, val_task_to_env)
+
+
+@trace_fn(RLSpanGroup.JOB, "rl.grpo.job")
+def grpo_train(
+    policy: ColocatablePolicyInterface,
+    policy_generation: Optional[GenerationInterface],
+    wrapped_dataloader: StatefulDataLoader | MultipleDataloaderWrapper,
+    val_dataloader: Optional[StatefulDataLoader],
+    tokenizer: TokenizerType,
+    loss_fn: LossFunction,
+    task_to_env: dict[str, EnvironmentInterface],
+    val_task_to_env: Optional[dict[str, EnvironmentInterface]],
+    logger: Logger,
+    checkpointer: CheckpointManager,
+    grpo_save_state: GRPOSaveState,
+    master_config: MasterConfig,
+    processor: Optional[AutoProcessor] = None,
+) -> None:
+    """Run GRPO training and always tear down its environments."""
+    try:
+        _grpo_train_impl(
+            policy=policy,
+            policy_generation=policy_generation,
+            wrapped_dataloader=wrapped_dataloader,
+            val_dataloader=val_dataloader,
+            tokenizer=tokenizer,
+            loss_fn=loss_fn,
+            task_to_env=task_to_env,
+            val_task_to_env=val_task_to_env,
+            logger=logger,
+            checkpointer=checkpointer,
+            grpo_save_state=grpo_save_state,
+            master_config=master_config,
+            processor=processor,
+        )
+    finally:
+        shutdown_environments(task_to_env, val_task_to_env)
 
 
 def validate(
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -86,6 +86,7 @@ from nemo_rl.distributed.virtual_cluster import (
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
 from nemo_rl.environments.nemo_gym import should_use_nemo_gym, spinup_nemo_gym_actor
+from nemo_rl.environments.utils import shutdown_environments
 from nemo_rl.experience.interfaces import (
     FRONTIER_ORDINAL_KEY,
     NEMO_GYM_TASK_INDEX_KEY,
@@ -490,35 +491,6 @@ def _needs_hf_refit_handshake(
     if generation_backend == "megatron":
         return False
     return not (nccl_reshard_refit_enabled and not colocated_inference)
-
-
-def shutdown_environments(
-    task_to_env: dict[str, EnvironmentInterface] | None,
-    val_task_to_env: dict[str, EnvironmentInterface] | None,
-) -> None:
-    """Shut down each unique environment actor before generation stops."""
-    seen_environment_handles: set[int] = set()
-    for environment_map in (task_to_env, val_task_to_env):
-        if environment_map is None:
-            continue
-        for task_name, environment in environment_map.items():
-            handle_id = id(environment)
-            if handle_id in seen_environment_handles:
-                continue
-            seen_environment_handles.add(handle_id)
-
-            print(f"🛑 Shutting down environment {task_name}...")
-            try:
-                ray.get(environment.shutdown.remote(), timeout=10)
-            except Exception as shutdown_error:
-                print(
-                    f"Environment {task_name} graceful shutdown failed: "
-                    f"{shutdown_error}"
-                )
-                try:
-                    ray.kill(environment)
-                except Exception as kill_error:
-                    print(f"Error stopping environment {task_name}: {kill_error}")
 
 
 def setup(
@@ -4068,11 +4040,13 @@ def grpo_train(
                 return
             if should_save_by_timeout:
                 checkpointer.shutdown()
+                shutdown_environments(task_to_env, val_task_to_env)
                 memory_tracker.snapshot_start_of_stage("", dir())
                 print("Timeout has been reached, stopping training early", flush=True)
                 return
             if total_steps >= max_num_steps:
                 checkpointer.shutdown()
+                shutdown_environments(task_to_env, val_task_to_env)
                 memory_tracker.snapshot_start_of_stage("", dir())
                 print(
                     "Max number of steps has been reached, stopping training early",
@@ -4089,6 +4063,10 @@ def grpo_train(
     # so without this the daemon finalization thread would be killed before the
     # final tmp_step_N is renamed.
     checkpointer.shutdown()
+    # Environments own long-lived server subprocesses (NeMo-Gym runs ~60 per
+    # actor), so they have to be stopped on every exit from the synchronous
+    # trainer too, not just the async one.
+    shutdown_environments(task_to_env, val_task_to_env)
 
 
 def validate(
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -174,12 +174,17 @@ def _detect_invalid_tool_call_and_malformed_thinking(
     return is_invalid_tool_call, has_malformed_thinking
 
 
-@ray.remote(max_restarts=-1, max_task_retries=-1)  # pragma: no cover
+# Fail fast rather than restart. The servers this actor owns are started in
+# _spinup, which Ray does not re-run after a restart, so a restarted actor is
+# permanently broken: every later call raises AttributeError on self.rch and
+# the caller never sees the RayActorError it is waiting for.
+@ray.remote(max_restarts=0, max_task_retries=0)  # pragma: no cover
 class NemoGym(EnvironmentInterface):
     """This environment class isn't really used for training. It's really meant as an integration wrapper around NeMo-Gym that hooks into the existing NeMo RL resource management via ray. So there is still one source of truth for resource management in NeMo RL."""
 
     def __init__(self, cfg: NemoGymConfig):
         self.cfg = cfg
+        self.rh = None
 
     def _spinup(self) -> None:
         """Start the NeMo-Gym head server and rollout collection helper.
@@ -512,7 +517,16 @@ Output prompt token IDs: {output_item_dict["prompt_token_ids"]}
         }
 
     def shutdown(self) -> None:
-        self.rh.shutdown()
+        """Stop the Gym servers. Safe to call more than once, and before spinup.
+
+        Callers routinely hold the same actor under both the train and the
+        validation environment map, so this is reached twice on a normal exit.
+        The underlying RunHelper.shutdown() raises on a second call, which used
+        to escalate an otherwise graceful teardown into a ray.kill().
+        """
+        rh, self.rh = self.rh, None
+        if rh is not None:
+            rh.shutdown()
 
     def step(self, message_log_batch, metadata):
         # This is not used since NeMo-Gym will handle the rollouts entirely.
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -326,7 +326,7 @@ def get_pad_dynamic_image_shapes(env_config: Mapping[str, Any]) -> bool:
 
 # Fail fast rather than restart. The servers this actor owns are started in
 # _spinup, which Ray does not re-run after a restart, so a restarted actor is
-# permanently broken: every later call raises AttributeError on self.rch and
+# permanently broken: _require_spinup() rejects every later rollout call, and
 # the caller never sees the RayActorError it is waiting for.
 @ray.remote(max_restarts=0, max_task_retries=0)  # pragma: no cover
 class NemoGym(EnvironmentInterface):
@@ -1151,13 +1151,10 @@ output prompt token ids till seen: {output_item_dict["prompt_token_ids"][: len(s
     def shutdown(self) -> None:
         """Stop the Gym servers. Safe to call more than once, and before spinup.
 
-        Callers routinely hold the same actor under both the train and the
-        validation environment map, so this is reached twice on a normal exit.
-        The underlying RunHelper.shutdown() raises on a second call, which used
-        to escalate an otherwise graceful teardown into a ray.kill(). Teardown
-        also runs in a finally block, so it must not turn a real training error
-        into a confusing AttributeError from a never-spun-up (e.g. restarted)
-        actor.
+        Teardown runs in a finally block and may be requested more than once.
+        RunHelper.shutdown() is not idempotent, so the handle is cleared before
+        it is used. A failure therefore cannot leave a live handle that a later
+        cleanup attempt invokes again.
         """
         rh, self.rh = self.rh, None
         if rh is not None:
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -324,7 +324,11 @@ def get_pad_dynamic_image_shapes(env_config: Mapping[str, Any]) -> bool:
     return bool(nemo_gym_config.get("pad_dynamic_image_shapes"))
 
 
-@ray.remote(max_restarts=-1, max_task_retries=-1)  # pragma: no cover
+# Fail fast rather than restart. The servers this actor owns are started in
+# _spinup, which Ray does not re-run after a restart, so a restarted actor is
+# permanently broken: every later call raises AttributeError on self.rch and
+# the caller never sees the RayActorError it is waiting for.
+@ray.remote(max_restarts=0, max_task_retries=0)  # pragma: no cover
 class NemoGym(EnvironmentInterface):
     """This environment class isn't really used for training. It's really meant as an integration wrapper around NeMo-Gym that hooks into the existing NeMo RL resource management via ray. So there is still one source of truth for resource management in NeMo RL."""
 
@@ -1145,13 +1149,19 @@ output prompt token ids till seen: {output_item_dict["prompt_token_ids"][: len(s
         return result
 
     def shutdown(self) -> None:
-        # Teardown runs in a finally block, so it must not turn a real training error
-        # into a confusing AttributeError from a never-spun-up (e.g. restarted) actor.
-        if self.rh is None:
-            return
-        run_helper = self.rh
-        self.rh = None
-        run_helper.shutdown()
+        """Stop the Gym servers. Safe to call more than once, and before spinup.
+
+        Callers routinely hold the same actor under both the train and the
+        validation environment map, so this is reached twice on a normal exit.
+        The underlying RunHelper.shutdown() raises on a second call, which used
+        to escalate an otherwise graceful teardown into a ray.kill(). Teardown
+        also runs in a finally block, so it must not turn a real training error
+        into a confusing AttributeError from a never-spun-up (e.g. restarted)
+        actor.
+        """
+        rh, self.rh = self.rh, None
+        if rh is not None:
+            rh.shutdown()
 
     def step(self, message_log_batch, metadata):
         # This is not used since NeMo-Gym will handle the rollouts entirely.
--- a/nemo_rl/environments/utils.py
+++ b/nemo_rl/environments/utils.py
@@ -149,16 +149,16 @@ def shutdown_environments(
     """Gracefully shut down every distinct environment actor in the given maps.
 
     Runners commonly bind the same actor — often the very same mapping — to
-    both training and validation, so handles are deduped before teardown.
-    Without that, the second shutdown of an already-stopped actor fails and
-    escalates a graceful exit into a ``ray.kill``.
+    both training and validation, so each handle is asked to stop once no
+    matter how many maps contain it.
 
     Environments must be torn down before generation workers: they may have
     in-flight HTTP requests to the vLLM endpoints, and killing generation first
     leaves them retrying dead connections.
 
     Args:
-        env_maps: Task-name to environment mappings. ``None`` entries are skipped.
+        env_maps: Task-name to environment mappings. ``None`` and empty
+            mappings are skipped.
         timeout: Seconds to wait for each actor's ``shutdown()`` before killing it.
     """
     seen: set[int] = set()
--- a/nemo_rl/environments/utils.py
+++ b/nemo_rl/environments/utils.py
@@ -12,13 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import os
-from typing import Any, Dict, NotRequired, TypedDict
+from typing import Any, Dict, NotRequired, Optional, TypedDict
 
+import ray
 from hydra.utils import get_object
 
 from nemo_rl.distributed.ray_actor_environment_registry import get_actor_python_env
 from nemo_rl.environments.interfaces import EnvironmentInterface
 from nemo_rl.utils.venvs import create_local_venv_on_each_node
+
+DEFAULT_ENV_SHUTDOWN_TIMEOUT_SECONDS = 10.0
 
 
 # Environment registry entry schema.
@@ -137,3 +140,43 @@ def register_env(env_name: str, actor_class_fqn: str) -> None:
         raise ValueError(f"Env name {env_name} already registered")
 
     ENV_REGISTRY[env_name] = {"actor_class_fqn": actor_class_fqn}
+
+
+def shutdown_environments(
+    *env_maps: Optional[Dict[str, EnvironmentInterface]],
+    timeout: float = DEFAULT_ENV_SHUTDOWN_TIMEOUT_SECONDS,
+) -> None:
+    """Gracefully shut down every distinct environment actor in the given maps.
+
+    Runners commonly bind the same actor — often the very same mapping — to
+    both training and validation, so handles are deduped before teardown.
+    Without that, the second shutdown of an already-stopped actor fails and
+    escalates a graceful exit into a ``ray.kill``.
+
+    Environments must be torn down before generation workers: they may have
+    in-flight HTTP requests to the vLLM endpoints, and killing generation first
+    leaves them retrying dead connections.
+
+    Args:
+        env_maps: Task-name to environment mappings. ``None`` entries are skipped.
+        timeout: Seconds to wait for each actor's ``shutdown()`` before killing it.
+    """
+    seen: set[int] = set()
+    for env_map in env_maps:
+        if not env_map:
+            continue
+        for task_name, env in env_map.items():
+            if id(env) in seen:
+                continue
+            seen.add(id(env))
+            print(f"🛑 Shutting down environment {task_name}...")
+            try:
+                ray.get(env.shutdown.remote(), timeout=timeout)
+            except Exception as e:
+                print(f"Graceful shutdown of environment {task_name} failed: {e}")
+                try:
+                    ray.kill(env)
+                except Exception as kill_error:
+                    print(
+                        f"Error killing environment {task_name}: {kill_error}",
+                    )
--- a/tests/unit/environments/test_environment_utils.py
+++ b/tests/unit/environments/test_environment_utils.py
@@ -11,9 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from unittest.mock import MagicMock, patch
+
 import pytest
 
-from nemo_rl.environments.utils import ENV_REGISTRY, register_env
+from nemo_rl.environments.utils import (
+    ENV_REGISTRY,
+    register_env,
+    shutdown_environments,
+)
 
 
 def test_register_new_env_success():
@@ -50,3 +56,61 @@ def test_register_env_duplicate_raises_error():
         # Restore original registry state
         ENV_REGISTRY.clear()
         ENV_REGISTRY.update(original_registry)
+
+
+def _fake_env(name):
+    env = MagicMock(name=name)
+    env.shutdown.remote.return_value = f"{name}-ref"
+    return env
+
+
+def test_shutdown_environments_dedupes_shared_handles():
+    """Runners bind the same actor to train and val; it must be stopped once.
+
+    Shutting an already-stopped NeMo-Gym actor down a second time raises and
+    would escalate a graceful teardown into a ray.kill.
+    """
+    env = _fake_env("gym")
+    task_to_env = {"nemo_gym": env}
+    val_task_to_env = task_to_env  # the aliasing real runners use
+
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        shutdown_environments(task_to_env, val_task_to_env)
+
+    env.shutdown.remote.assert_called_once_with()
+    mock_ray.get.assert_called_once_with("gym-ref", timeout=10.0)
+    mock_ray.kill.assert_not_called()
+
+
+def test_shutdown_environments_stops_each_distinct_actor():
+    train_env, val_env = _fake_env("train"), _fake_env("val")
+
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        shutdown_environments({"a": train_env}, {"a": val_env}, None)
+
+    train_env.shutdown.remote.assert_called_once_with()
+    val_env.shutdown.remote.assert_called_once_with()
+    assert mock_ray.get.call_count == 2
+    mock_ray.kill.assert_not_called()
+
+
+def test_shutdown_environments_kills_only_on_failure():
+    healthy, wedged = _fake_env("healthy"), _fake_env("wedged")
+
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        mock_ray.get.side_effect = (
+            lambda ref, timeout: None
+            if ref == "healthy-ref"
+            else (_ for _ in ()).throw(TimeoutError("no response"))
+        )
+        shutdown_environments({"healthy": healthy, "wedged": wedged})
+
+    mock_ray.kill.assert_called_once_with(wedged)
+
+
+def test_shutdown_environments_tolerates_empty_input():
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        shutdown_environments(None, {})
+
+    mock_ray.get.assert_not_called()
+    mock_ray.kill.assert_not_called()
--- a/tests/unit/environments/test_environment_utils.py
+++ b/tests/unit/environments/test_environment_utils.py
@@ -65,11 +65,7 @@ def _fake_env(name):
 
 
 def test_shutdown_environments_dedupes_shared_handles():
-    """Runners bind the same actor to train and val; it must be stopped once.
-
-    Shutting an already-stopped NeMo-Gym actor down a second time raises and
-    would escalate a graceful teardown into a ray.kill.
-    """
+    """Runners can bind one actor to train and val; it must be stopped once."""
     env = _fake_env("gym")
     task_to_env = {"nemo_gym": env}
     val_task_to_env = task_to_env  # the aliasing real runners use
@@ -94,8 +90,8 @@ def test_shutdown_environments_stops_each_distinct_actor():
     mock_ray.kill.assert_not_called()
 
 
-def test_shutdown_environments_kills_only_on_failure():
-    healthy, wedged = _fake_env("healthy"), _fake_env("wedged")
+def test_shutdown_environments_continues_after_a_failure():
+    wedged, healthy = _fake_env("wedged"), _fake_env("healthy")
 
     with patch("nemo_rl.environments.utils.ray") as mock_ray:
         mock_ray.get.side_effect = (
@@ -103,9 +99,26 @@ def test_shutdown_environments_kills_only_on_failure():
             if ref == "healthy-ref"
             else (_ for _ in ()).throw(TimeoutError("no response"))
         )
-        shutdown_environments({"healthy": healthy, "wedged": wedged})
+        shutdown_environments({"wedged": wedged, "healthy": healthy})
 
     mock_ray.kill.assert_called_once_with(wedged)
+    healthy.shutdown.remote.assert_called_once_with()
+
+
+def test_shutdown_environments_continues_after_a_failed_kill():
+    wedged, healthy = _fake_env("wedged"), _fake_env("healthy")
+
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        mock_ray.get.side_effect = (
+            lambda ref, timeout: None
+            if ref == "healthy-ref"
+            else (_ for _ in ()).throw(TimeoutError("no response"))
+        )
+        mock_ray.kill.side_effect = RuntimeError("already gone")
+        shutdown_environments({"wedged": wedged, "healthy": healthy})
+
+    mock_ray.kill.assert_called_once_with(wedged)
+    healthy.shutdown.remote.assert_called_once_with()
 
 
 def test_shutdown_environments_tolerates_empty_input():
--- a/examples/nemo_gym/run_grpo_nemo_gym.py
+++ b/examples/nemo_gym/run_grpo_nemo_gym.py
@@ -36,7 +36,6 @@ from nemo_rl.algorithms.grpo import (
     grpo_train,
     refit_policy_generation,
     setup,
-    shutdown_environments,
 )
 from nemo_rl.algorithms.utils import get_tokenizer
 from nemo_rl.data.utils import setup_response_data
@@ -46,6 +45,7 @@ from nemo_rl.environments.nemo_gym import (
     setup_nemo_gym_config,
     should_use_nemo_gym,
 )
+from nemo_rl.environments.utils import shutdown_environments
 from nemo_rl.experience.rollouts import run_nemo_gym_rollout_sync
 from nemo_rl.models.generation import configure_generation_config
 from nemo_rl.models.generation.vllm.config import materialize_vllm_video_config
@@ -274,6 +274,7 @@ The validation set you pass in will directly be used for validation with no addi
     task_to_env = {"nemo_gym": nemo_gym}
     val_task_to_env = task_to_env
 
+    trainer_owns_environment_shutdown = False
     try:
         if is_trajectory_collection:
             collect_trajectories(
@@ -334,6 +335,7 @@ The validation set you pass in will directly be used for validation with no addi
             print("🚀 Running synchronous GRPO training")
 
             # Run standard GRPO training
+            trainer_owns_environment_shutdown = True
             grpo_train(
                 policy,
                 policy_generation,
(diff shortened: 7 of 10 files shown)


nemo_rl/environments/nemo_gym.py
```diff
@@ -307,7 +323,162 @@ def get_pad_dynamic_image_shapes(env_config: Mapping[str, Any]) -> bool:
...
+        images=images,
+        processor=processor,
+        pad_dynamic_image_shapes=pad_dynamic_image_shapes,
+    )
+
+
+# Fail fast rather than restart. The servers this actor owns are started in
+# _spinup, which Ray does not re-run after a restart, so a restarted actor is
+# permanently broken: every later call raises AttributeError on self.rch and
+# the caller never sees the RayActorError it is waiting for.
```
terrykong: **1 action item, in two places.**
**TL;DR — The new comment blames an `AttributeError` that cannot happen, so it tells the next reader the wrong reason for `max_restarts=0`.**

Both halves of the reason are wrong. `__init__` sets `self.rch = None` ([nemo_gym.py L491](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L491)), so reading it never raises `AttributeError`. And [`run_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L667) calls `self._require_spinup()` as its very first statement, about 26 lines before it touches `self.rch` ([L693](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L693)); [`_require_spinup`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L519-L527) raises a clear `RuntimeError`, and [`health_check`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L537) does the same.

It also contradicts the correct comment 8 lines below it ([L487-L489](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L487-L489)), which says these attributes are pre-declared exactly so you do *not* get an `AttributeError`. `_require_spinup` came from e98357616 (#3589) and is already on main, so this is a new comment describing code that has already changed.

The `max_restarts=0, max_task_retries=0` change itself is right and should stay. Only the reason needs fixing.

### AI-1
```suggestion
# Fail fast rather than restart. The servers this actor owns are started in
# _spinup, which Ray does not re-run after a restart, so a restarted actor is
# permanently broken: _require_spinup() rejects every later rollout call, and
# the caller never sees the RayActorError it is waiting for.
```

The same wording is repeated in the docstring of [`test_nemo_gym_fails_fast_instead_of_restarting`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/environments/test_nemo_gym_utils.py#L255-L261) and needs the same correction there.

nemo_rl/environments/nemo_gym.py
```diff
@@ -847,13 +1022,19 @@ def _postprocess_nemo_gym_to_nemo_rl_result(
...
+        """Stop the Gym servers. Safe to call more than once, and before spinup.
+
+        Callers routinely hold the same actor under both the train and the
+        validation environment map, so this is reached twice on a normal exit.
+        The underlying RunHelper.shutdown() raises on a second call, which used
+        to escalate an otherwise graceful teardown into a ray.kill(). Teardown
+        also runs in a finally block, so it must not turn a real training error
+        into a confusing AttributeError from a never-spun-up (e.g. restarted)
+        actor.
+        """
```
**TL;DR — The docstring says the guard stops teardown escalating to `ray.kill()`, but that escalation cannot happen on this base, so the guard now has a false reason attached to it.**

The first half is right: [`RunHelper.shutdown()`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L625) sets `self._head_server = None`, so a second call reaches [`self._head_server.should_exit = True`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L622) with `None` and raises `AttributeError`.

The second half is stale. The `if self.rh is None: return` guard that keeps `RunHelper` from being reached twice already exists at the base — see the [base version of these lines](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1020-L1027) (linked at the base, since the PR replaces them). It landed on main in #3589 (e98357616), not in this stack. So the escalation to `ray.kill()` never happened, and the new tuple-swap rewrite changes no behavior. This matters because a docstring that gives a false reason for a guard is how the guard gets deleted later.

### AI-1
```suggestion
        """Stop the Gym servers. Safe to call more than once, and before spinup.

        Callers routinely hold the same actor under both the train and the
        validation environment map, and teardown runs from a finally block, so
        this is reached more than once on both the success and the failure
        path. RunHelper.shutdown() raises on a second call, so the handle is
        cleared before it is used.
        """
```

nemo_rl/environments/utils.py
```diff
@@ -137,3 +140,43 @@ def register_env(env_name: str, actor_class_fqn: str) -> None:
...
+    escalates a graceful exit into a ``ray.kill``.
+
+    Environments must be torn down before generation workers: they may have
+    in-flight HTTP requests to the vLLM endpoints, and killing generation first
+    leaves them retrying dead connections.
+
+    Args:
+        env_maps: Task-name to environment mappings. ``None`` entries are skipped.
+        timeout: Seconds to wait for each actor's ``shutdown()`` before killing it.
+    """
```
terrykong: **1 action item.**
**TL;DR — The docstring says dedup is needed because a second shutdown of a stopped actor escalates a graceful exit into a `ray.kill`; no environment in the tree does that, so the stated reason is false.**

All six were checked. [`NemoGym.shutdown`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L1024) returns when `rh is None`. [`MathEnvironment.shutdown`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/math_environment.py#L396), [`CodeEnvironment.shutdown`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/code_environment.py#L258), [`VLMEnvironment.shutdown`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/vlm_environment.py#L141) and [`code_jaccard_environment.py L163`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/code_jaccard_environment.py#L163) are all `for worker in self.workers: ray.kill(worker)`, and `ray.kill` on a dead actor is a no-op — [ray `_private/worker.py` L3242-L3280](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/worker.py#L3242-L3280) raises only for a non-`ActorHandle` or a handle from another session. [`reward_model_environment.py L332`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/reward_model_environment.py#L332) is `hasattr`-guarded. Keep the dedup — it still saves a duplicate remote call and a duplicate log line — just fix the reason.

The Args line also says "``None`` entries are skipped", but [`if not env_map`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L166) skips `{}` too, which [`test_shutdown_environments_tolerates_empty_input`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/environments/test_environment_utils.py#L111) relies on.

### AI-1
```suggestion
    """Gracefully shut down every distinct environment actor in the given maps.

    Runners commonly bind the same actor — often the very same mapping — to
    both training and validation, so handles are deduped: each actor is asked
    to stop once, no matter how many maps it appears in.

    Environments must be torn down before generation workers: they may have
    in-flight HTTP requests to the vLLM endpoints, and killing generation first
    leaves them retrying dead connections.

    Args:
        env_maps: Task-name to environment mappings. ``None`` and empty
            mappings are skipped.
        timeout: Seconds to wait for each actor's ``shutdown()`` before killing it.
    """
```

**Context — no action.**
- Convention: [`*env_maps: Optional[Dict[str, EnvironmentInterface]]`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L146) moved away from the house form. The code being moved used `dict[str, EnvironmentInterface] | None` ([base grpo.py L483-L484](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/grpo.py#L483-L484)). Under `nemo_rl/`, `Optional[dict[...]]` appears 131 times against 3 for `Optional[Dict[...]]`. Going back also frees the new [`Optional` import](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L15). Ruff is clean either way, so this is convention only.
- Optional: putting the handles in a `set` would dedup by actor id instead of `id()`. Ray's `ActorHandle` defines [`__hash__` over `_ray_actor_id`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L2518) and [`__eq__`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L2529), and [`create_env`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L129-L135) returns a real `ActorHandle`. `id()` is safe here only because the dicts hold strong references for the whole call.

nemo_rl/environments/utils.py
```diff
@@ -12,14 +12,17 @@
...
+from typing import Any, Dict, NotRequired, Optional, TypedDict
 
+import ray
 from hydra.utils import get_object
 
 from nemo_rl.distributed.ray_actor_environment_registry import get_actor_python_env
 from nemo_rl.environments.interfaces import EnvironmentInterface
 from nemo_rl.utils.venvs import create_local_venv_on_each_node
 
+DEFAULT_ENV_SHUTDOWN_TIMEOUT_SECONDS = 10.0
```
terrykong: **1 action item.**
**TL;DR — The new 10s budget times a call that has no time bound of its own, and when the budget runs out, `ray.kill` cuts the actor down before it can force-kill the servers still holding ports.**

`NemoGym.shutdown()` calls [`RunHelper.shutdown()`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L586), which ends with [`self._head_server_thread.join()`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L623) — no timeout.

For the server processes, `RunHelper.shutdown()` [sends SIGINT to all of them first](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L588-L589), so healthy servers stop at the same time and the cost is roughly the slowest one, not the sum. But every process that ignores SIGINT costs up to 1s on [its own wait](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L596) ([`_GRACEFUL_SHUTDOWN_TIMEOUT_SEC = 1`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L91)) plus up to 2s to [reap it after SIGKILL](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L602) ([`_FORCE_KILL_REAP_TIMEOUT_SEC = 2`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L93)). So about four SIGINT-ignoring servers blow the 10s budget. For scale, the largest in-tree Gym config, [examples/nemo_gym/nemotron-3-super/stage1_rlvr.yaml](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/nemotron-3-super/stage1_rlvr.yaml), declares at least 13 server blocks.

When the budget runs out, `shutdown_environments` calls [`ray.kill(env)`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L178). That kills the actor before `RunHelper` reaches [its own SIGKILL escalation](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L598) — so the servers that ignore SIGINT, which are exactly the ones holding ports, are the ones left behind.

### AI-1
Raise `DEFAULT_ENV_SHUTDOWN_TIMEOUT_SECONDS`, or pass a larger `timeout=` for the NeMo-Gym actor. No suggestion block because the right number is yours to pick.

nemo_rl/algorithms/grpo.py
```diff
@@ -4057,6 +4029,43 @@ def grpo_train(
...
+            tokenizer,
+            loss_fn,
+            task_to_env,
+            val_task_to_env,
+            logger,
+            checkpointer,
+            grpo_save_state,
+            master_config,
+            processor,
+        )
```
terrykong: **1 action item.**
**TL;DR — The wrapper forwards 13 arguments positionally, so reordering two parameters in the impl and forgetting the wrapper sends the wrong values through with no error.**

`grpo_train` re-declares the parameters of [`_grpo_train_impl`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo.py#L2838) and passes them by position. The adjacent pairs `wrapped_dataloader`/`val_dataloader` and `task_to_env`/`val_task_to_env` have matching types, so a swap runs fine and trains on the wrong data. Nothing catches it: there is no mypy or pyright anywhere in the repo (nothing in [pyproject.toml](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/pyproject.toml) or [.pre-commit-config.yaml](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/.pre-commit-config.yaml)), and the new test at [tests/unit/algorithms/test_grpo.py L6027](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/algorithms/test_grpo.py#L6027) patches `_grpo_train_impl` out entirely, so the forwarding has no coverage at all. Keyword forwarding makes the whole class of mistake impossible.

This is not a request to change the wrapper shape — the split is the right call, since [`trace_fn`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/telemetry/instrumentation.py#L343-L354) does not require it and a `try`/`finally` in place would re-indent about 1200 lines.

### AI-1
```suggestion
        _grpo_train_impl(
            policy=policy,
            policy_generation=policy_generation,
            wrapped_dataloader=wrapped_dataloader,
            val_dataloader=val_dataloader,
            tokenizer=tokenizer,
            loss_fn=loss_fn,
            task_to_env=task_to_env,
            val_task_to_env=val_task_to_env,
            logger=logger,
            checkpointer=checkpointer,
            grpo_save_state=grpo_save_state,
            master_config=master_config,
            processor=processor,
        )
```

<details><summary>Why a later change to the finally block would not turn any test red</summary>

Both wrapper tests raise from the impl ([test_grpo.py L6027](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/algorithms/test_grpo.py#L6027) and [test_distillation.py L920](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/algorithms/test_distillation.py#L920)), so swapping `finally:` for `except Exception: shutdown(...); raise` would keep them green while dropping teardown on every clean run. This test covers the success path (run and passing):

```python
def test_grpo_train_shuts_down_environments_on_success():
    task_to_env = {"nemo_gym": MagicMock()}

    with (
        patch("nemo_rl.algorithms.grpo._grpo_train_impl"),
        patch("nemo_rl.algorithms.grpo.shutdown_environments") as shutdown,
    ):
        grpo_train(
            policy=MagicMock(),
            policy_generation=MagicMock(),
            wrapped_dataloader=MagicMock(),
            val_dataloader=None,
            tokenizer=MagicMock(),
            loss_fn=MagicMock(),
            task_to_env=task_to_env,
            val_task_to_env=task_to_env,
            logger=MagicMock(),
            checkpointer=MagicMock(),
            grpo_save_state=MagicMock(),
            master_config=MagicMock(),
        )

    shutdown.assert_called_once_with(task_to_env, task_to_env)
```
</details>

nemo_rl/algorithms/distillation.py
```diff
@@ -1183,6 +1183,43 @@ def distillation_train(
...
+            val_dataloader,
+            tokenizer,
+            loss_fn,
+            task_to_env,
+            val_task_to_env,
+            logger,
+            checkpointer,
+            distillation_save_state,
+            master_config,
+        )
```
terrykong: **1 action item.**
**TL;DR — The wrapper forwards 13 arguments positionally to [`_distillation_train_impl`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/distillation.py#L642), so reordering two impl parameters without updating the wrapper sends the wrong values through with no error.**

The adjacent pairs `dataloader`/`val_dataloader` and `task_to_env`/`val_task_to_env` have matching types, so a swap runs fine and trains on the wrong data, and the repo has no static type checker to catch it. Keyword forwarding removes the risk.

### AI-1
```suggestion
        _distillation_train_impl(
            student_policy=student_policy,
            teacher_policy=teacher_policy,
            student_generation=student_generation,
            dataloader=dataloader,
            val_dataloader=val_dataloader,
            tokenizer=tokenizer,
            loss_fn=loss_fn,
            task_to_env=task_to_env,
            val_task_to_env=val_task_to_env,
            logger=logger,
            checkpointer=checkpointer,
            distillation_save_state=distillation_save_state,
            master_config=master_config,
        )
```

tests/unit/environments/test_environment_utils.py
```diff
@@ -50,3 +56,61 @@ def test_register_env_duplicate_raises_error():
...
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        shutdown_environments({"a": train_env}, {"a": val_env}, None)
+
+    train_env.shutdown.remote.assert_called_once_with()
+    val_env.shutdown.remote.assert_called_once_with()
+    assert mock_ray.get.call_count == 2
+    mock_ray.kill.assert_not_called()
+
+
+def test_shutdown_environments_kills_only_on_failure():
```
terrykong: **1 action item.**
**TL;DR — This test passes even if the `try`/`except` is moved outside the loop, so it does not cover the thing it looks like it covers: that one wedged actor must not strand the environments behind it.**

The dict is `{"healthy": healthy, "wedged": wedged}`. Python keeps insertion order, so the failing environment is last and the loop ends right after it — move the `try`/`except` out of the loop and this test still goes green while every environment after the first failure is silently skipped. The nested `except` around `ray.kill` ([utils.py L177-L182](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L177-L182)) is untested too.

### AI-1
Add these two tests. No suggestion block — the fix is two new tests, not an edit to this line.

```python
def test_shutdown_environments_continues_after_a_failure():
    """One wedged actor must not strand the environments behind it."""
    wedged, healthy = _fake_env("wedged"), _fake_env("healthy")

    with patch("nemo_rl.environments.utils.ray") as mock_ray:
        mock_ray.get.side_effect = (
            lambda ref, timeout: (_ for _ in ()).throw(TimeoutError("no response"))
            if ref == "wedged-ref"
            else None
        )
        shutdown_environments({"wedged": wedged, "healthy": healthy})

    healthy.shutdown.remote.assert_called_once_with()
    mock_ray.kill.assert_called_once_with(wedged)


def test_shutdown_environments_survives_a_failed_kill():
    """A kill that itself raises must not abort teardown."""
    wedged = _fake_env("wedged")

    with patch("nemo_rl.environments.utils.ray") as mock_ray:
        mock_ray.get.side_effect = TimeoutError("no response")
        mock_ray.kill.side_effect = RuntimeError("already gone")
        shutdown_environments({"wedged": wedged})

    mock_ray.kill.assert_called_once_with(wedged)
```

nemo_rl/algorithms/grpo.py
```diff
@@ -4057,6 +4029,43 @@ def grpo_train(
...
+    finally:
+        shutdown_environments(task_to_env, val_task_to_env)
```
terrykong: **1 action item.**
**TL;DR — Environment teardown now has four different shapes across the training entry points, and on the sync GRPO path it runs twice, so a timed-out actor makes `ActorDiedError` the job's last log line.**

The four shapes:
- `grpo_train` — wrapper `finally`, new (this line)
- [`distillation_train`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/distillation.py#L1220) — wrapper `finally`, new
- [`async_grpo_train`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo.py#L5975) — its own inline block, in a different order
- [`grpo_train_sync`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo_sync.py#L380) — nothing at all

On the sync GRPO path `shutdown_environments` now runs twice: once here, then again in the runner's own `finally` ([examples/run_grpo.py L260](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_grpo.py#L260) and [examples/nemo_gym/run_grpo_nemo_gym.py L353](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_grpo_nemo_gym.py#L353)). If the first pass times out and calls `ray.kill`, the second pass calls `shutdown.remote()` on a dead actor and the job's last log line becomes `Graceful shutdown of environment X failed: ActorDiedError ...`, which points whoever is on call at the wrong problem.

Both runner-level calls are still required today, so neither can just be deleted:
- [examples/run_grpo.py L260](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_grpo.py#L260) is the only teardown for the `grpo_train_sync` branch, which [`_select_trainer`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_grpo.py#L43-L56) returns when `data_plane.enabled`.
- [examples/nemo_gym/run_grpo_nemo_gym.py L353](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_grpo_nemo_gym.py#L353) is the only teardown for that runner's [`collect_trajectories` branch](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_grpo_nemo_gym.py#L279).

### AI-1
Preferred: give [`grpo_train_sync`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo_sync.py#L380) the same wrapper, and confirm the [`collect_trajectories`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_grpo_nemo_gym.py#L279) branch, and **only then** drop the runner-level calls, so teardown has one owner. If that is more than you want in this PR: add a one-line comment on `grpo_train_sync` saying its caller owns environment teardown. No suggestion block — the change spans several files.

nemo_rl/environments/utils.py
```diff
@@ -137,3 +140,43 @@ def register_env(env_name: str, actor_class_fqn: str) -> None:
         raise ValueError(f"Env name {env_name} already registered")
 
     ENV_REGISTRY[env_name] = {"actor_class_fqn": actor_class_fqn}
+
+
+def shutdown_environments(
```
terrykong: **No action needed** — non-blocking follow-up, a separate PR is fine.
**TL;DR — PPO still carries a third hand-rolled copy of this loop, and unlike this helper it does not dedup, so an aliased train/val map asks the same actor to stop twice.**

[nemo_rl/algorithms/ppo.py L2271-L2281](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/ppo.py#L2271-L2281) repeats the same logic inline:

```python
for env_dict in (task_to_env, val_task_to_env):
    if env_dict is None: continue
    for task_name, env in env_dict.items():
        print(f"🛑 Shutting down environment {task_name}...")
        try: ray.get(env.shutdown.remote(), timeout=10)
        except Exception:
            try: ray.kill(env)
```

Separately, the synchronous [`ppo_train`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/ppo.py#L1237) has no environment teardown at all, and the only `finally` in [examples/run_ppo.py L214](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_ppo.py#L214) is `shutdown_telemetry()`. [`nemo_rl/algorithms/ppo.py`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/ppo.py) mentions nemo_gym in 10 places, so this path is live.

### Follow-up
Point PPO at this helper and give [`ppo_train`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/ppo.py#L1237) the same wrapper — that finishes what this PR started, one shared helper for teardown. No suggestion block — different file than this anchor.

nemo_rl/environments/nemo_gym.py
terrykong, line 313: **1 action item, in two places.**
**TL;DR — The new comment blames an `AttributeError` that cannot happen, so it tells the next reader the wrong reason for `max_restarts=0`.**

Both halves of the reason are wrong. `__init__` sets `self.rch = None` ([nemo_gym.py L491](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L491)), so reading it never raises `AttributeError`. And [`run_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L667) calls `self._require_spinup()` as its very first statement, about 26 lines before it touches `self.rch` ([L693](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L693)); [`_require_spinup`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L519-L527) raises a clear `RuntimeError`, and [`health_check`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L537) does the same.

It also contradicts the correct comment 8 lines below it ([L487-L489](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L487-L489)), which says these attributes are pre-declared exactly so you do *not* get an `AttributeError`. `_require_spinup` came from e98357616 (#3589) and is already on main, so this is a new comment describing code that has already changed.

The `max_restarts=0, max_task_retries=0` change itself is right and should stay. Only the reason needs fixing.

### AI-1
```suggestion
# Fail fast rather than restart. The servers this actor owns are started in
# _spinup, which Ray does not re-run after a restart, so a restarted actor is
# permanently broken: _require_spinup() rejects every later rollout call, and
# the caller never sees the RayActorError it is waiting for.
```

The same wording is repeated in the docstring of [`test_nemo_gym_fails_fast_instead_of_restarting`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/environments/test_nemo_gym_utils.py#L255-L261) and needs the same correction there.

nemo_rl/environments/nemo_gym.py
terrykong, line 860: **1 action item.**
**TL;DR — The docstring says the guard stops teardown escalating to `ray.kill()`, but that escalation cannot happen on this base, so the guard now has a false reason attached to it.**

The first half is right: [`RunHelper.shutdown()`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L625) sets `self._head_server = None`, so a second call reaches [`self._head_server.should_exit = True`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L622) with `None` and raises `AttributeError`.

The second half is stale. The `if self.rh is None: return` guard that keeps `RunHelper` from being reached twice already exists at the base — see the [base version of these lines](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1020-L1027) (linked at the base, since the PR replaces them). It landed on main in #3589 (e98357616), not in this stack. So the escalation to `ray.kill()` never happened, and the new tuple-swap rewrite changes no behavior. This matters because a docstring that gives a false reason for a guard is how the guard gets deleted later.

### AI-1
```suggestion
        """Stop the Gym servers. Safe to call more than once, and before spinup.

        Callers routinely hold the same actor under both the train and the
        validation environment map, and teardown runs from a finally block, so
        this is reached more than once on both the success and the failure
        path. RunHelper.shutdown() raises on a second call, so the handle is
        cleared before it is used.
        """
```

nemo_rl/environments/utils.py
terrykong, line 163: **1 action item.**
**TL;DR — The docstring says dedup is needed because a second shutdown of a stopped actor escalates a graceful exit into a `ray.kill`; no environment in the tree does that, so the stated reason is false.**

All six were checked. [`NemoGym.shutdown`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/nemo_gym.py#L1024) returns when `rh is None`. [`MathEnvironment.shutdown`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/math_environment.py#L396), [`CodeEnvironment.shutdown`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/code_environment.py#L258), [`VLMEnvironment.shutdown`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/vlm_environment.py#L141) and [`code_jaccard_environment.py L163`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/code_jaccard_environment.py#L163) are all `for worker in self.workers: ray.kill(worker)`, and `ray.kill` on a dead actor is a no-op — [ray `_private/worker.py` L3242-L3280](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/worker.py#L3242-L3280) raises only for a non-`ActorHandle` or a handle from another session. [`reward_model_environment.py L332`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/reward_model_environment.py#L332) is `hasattr`-guarded. Keep the dedup — it still saves a duplicate remote call and a duplicate log line — just fix the reason.

The Args line also says "``None`` entries are skipped", but [`if not env_map`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L166) skips `{}` too, which [`test_shutdown_environments_tolerates_empty_input`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/environments/test_environment_utils.py#L111) relies on.

### AI-1
```suggestion
    """Gracefully shut down every distinct environment actor in the given maps.

    Runners commonly bind the same actor — often the very same mapping — to
    both training and validation, so handles are deduped: each actor is asked
    to stop once, no matter how many maps it appears in.

    Environments must be torn down before generation workers: they may have
    in-flight HTTP requests to the vLLM endpoints, and killing generation first
    leaves them retrying dead connections.

    Args:
        env_maps: Task-name to environment mappings. ``None`` and empty
            mappings are skipped.
        timeout: Seconds to wait for each actor's ``shutdown()`` before killing it.
    """
```

**Context — no action.**
- Convention: [`*env_maps: Optional[Dict[str, EnvironmentInterface]]`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L146) moved away from the house form. The code being moved used `dict[str, EnvironmentInterface] | None` ([base grpo.py L483-L484](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/algorithms/grpo.py#L483-L484)). Under `nemo_rl/`, `Optional[dict[...]]` appears 131 times against 3 for `Optional[Dict[...]]`. Going back also frees the new [`Optional` import](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L15). Ruff is clean either way, so this is convention only.
- Optional: putting the handles in a `set` would dedup by actor id instead of `id()`. Ray's `ActorHandle` defines [`__hash__` over `_ray_actor_id`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L2518) and [`__eq__`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L2529), and [`create_env`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L129-L135) returns a real `ActorHandle`. `id()` is safe here only because the dicts hold strong references for the whole call.

nemo_rl/algorithms/grpo.py
terrykong, line 4063: **1 action item.**
**TL;DR — The wrapper forwards 13 arguments positionally, so reordering two parameters in the impl and forgetting the wrapper sends the wrong values through with no error.**

`grpo_train` re-declares the parameters of [`_grpo_train_impl`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo.py#L2838) and passes them by position. The adjacent pairs `wrapped_dataloader`/`val_dataloader` and `task_to_env`/`val_task_to_env` have matching types, so a swap runs fine and trains on the wrong data. Nothing catches it: there is no mypy or pyright anywhere in the repo (nothing in [pyproject.toml](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/pyproject.toml) or [.pre-commit-config.yaml](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/.pre-commit-config.yaml)), and the new test at [tests/unit/algorithms/test_grpo.py L6027](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/algorithms/test_grpo.py#L6027) patches `_grpo_train_impl` out entirely, so the forwarding has no coverage at all. Keyword forwarding makes the whole class of mistake impossible.

This is not a request to change the wrapper shape — the split is the right call, since [`trace_fn`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/telemetry/instrumentation.py#L343-L354) does not require it and a `try`/`finally` in place would re-indent about 1200 lines.

### AI-1
```suggestion
        _grpo_train_impl(
            policy=policy,
            policy_generation=policy_generation,
            wrapped_dataloader=wrapped_dataloader,
            val_dataloader=val_dataloader,
            tokenizer=tokenizer,
            loss_fn=loss_fn,
            task_to_env=task_to_env,
            val_task_to_env=val_task_to_env,
            logger=logger,
            checkpointer=checkpointer,
            grpo_save_state=grpo_save_state,
            master_config=master_config,
            processor=processor,
        )
```

<details><summary>Why a later change to the finally block would not turn any test red</summary>

Both wrapper tests raise from the impl ([test_grpo.py L6027](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/algorithms/test_grpo.py#L6027) and [test_distillation.py L920](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/algorithms/test_distillation.py#L920)), so swapping `finally:` for `except Exception: shutdown(...); raise` would keep them green while dropping teardown on every clean run. This test covers the success path (run and passing):

```python
def test_grpo_train_shuts_down_environments_on_success():
    task_to_env = {"nemo_gym": MagicMock()}

    with (
        patch("nemo_rl.algorithms.grpo._grpo_train_impl"),
        patch("nemo_rl.algorithms.grpo.shutdown_environments") as shutdown,
    ):
        grpo_train(
            policy=MagicMock(),
            policy_generation=MagicMock(),
            wrapped_dataloader=MagicMock(),
            val_dataloader=None,
            tokenizer=MagicMock(),
            loss_fn=MagicMock(),
            task_to_env=task_to_env,
            val_task_to_env=task_to_env,
            logger=MagicMock(),
            checkpointer=MagicMock(),
            grpo_save_state=MagicMock(),
            master_config=MagicMock(),
        )

    shutdown.assert_called_once_with(task_to_env, task_to_env)
```
</details>

nemo_rl/algorithms/distillation.py
terrykong, line 1218: **1 action item.**
**TL;DR — The wrapper forwards 13 arguments positionally to [`_distillation_train_impl`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/distillation.py#L642), so reordering two impl parameters without updating the wrapper sends the wrong values through with no error.**

The adjacent pairs `dataloader`/`val_dataloader` and `task_to_env`/`val_task_to_env` have matching types, so a swap runs fine and trains on the wrong data, and the repo has no static type checker to catch it. Keyword forwarding removes the risk.

### AI-1
```suggestion
        _distillation_train_impl(
            student_policy=student_policy,
            teacher_policy=teacher_policy,
            student_generation=student_generation,
            dataloader=dataloader,
            val_dataloader=val_dataloader,
            tokenizer=tokenizer,
            loss_fn=loss_fn,
            task_to_env=task_to_env,
            val_task_to_env=val_task_to_env,
            logger=logger,
            checkpointer=checkpointer,
            distillation_save_state=distillation_save_state,
            master_config=master_config,
        )
```

tests/unit/environments/test_environment_utils.py
terrykong: **1 action item.**
**TL;DR — This test passes even if the `try`/`except` is moved outside the loop, so it does not cover the thing it looks like it covers: that one wedged actor must not strand the environments behind it.**

The dict is `{"healthy": healthy, "wedged": wedged}`. Python keeps insertion order, so the failing environment is last and the loop ends right after it — move the `try`/`except` out of the loop and this test still goes green while every environment after the first failure is silently skipped. The nested `except` around `ray.kill` ([utils.py L177-L182](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L177-L182)) is untested too.

### AI-1
Add these two tests. No suggestion block — the fix is two new tests, not an edit to this line.

```python
def test_shutdown_environments_continues_after_a_failure():
    """One wedged actor must not strand the environments behind it."""
    wedged, healthy = _fake_env("wedged"), _fake_env("healthy")

    with patch("nemo_rl.environments.utils.ray") as mock_ray:
        mock_ray.get.side_effect = (
            lambda ref, timeout: (_ for _ in ()).throw(TimeoutError("no response"))
            if ref == "wedged-ref"
            else None
        )
        shutdown_environments({"wedged": wedged, "healthy": healthy})

    healthy.shutdown.remote.assert_called_once_with()
    mock_ray.kill.assert_called_once_with(wedged)


def test_shutdown_environments_survives_a_failed_kill():
    """A kill that itself raises must not abort teardown."""
    wedged = _fake_env("wedged")

    with patch("nemo_rl.environments.utils.ray") as mock_ray:
        mock_ray.get.side_effect = TimeoutError("no response")
        mock_ray.kill.side_effect = RuntimeError("already gone")
        shutdown_environments({"wedged": wedged})

    mock_ray.kill.assert_called_once_with(wedged)
```

nemo_rl/algorithms/grpo.py
terrykong, line 4065: **1 action item.**
**TL;DR — Environment teardown now has four different shapes across the training entry points, and on the sync GRPO path it runs twice, so a timed-out actor makes `ActorDiedError` the job's last log line.**

The four shapes:
- `grpo_train` — wrapper `finally`, new (this line)
- [`distillation_train`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/distillation.py#L1220) — wrapper `finally`, new
- [`async_grpo_train`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo.py#L5975) — its own inline block, in a different order
- [`grpo_train_sync`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo_sync.py#L380) — nothing at all

On the sync GRPO path `shutdown_environments` now runs twice: once here, then again in the runner's own `finally` ([examples/run_grpo.py L260](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_grpo.py#L260) and [examples/nemo_gym/run_grpo_nemo_gym.py L353](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_grpo_nemo_gym.py#L353)). If the first pass times out and calls `ray.kill`, the second pass calls `shutdown.remote()` on a dead actor and the job's last log line becomes `Graceful shutdown of environment X failed: ActorDiedError ...`, which points whoever is on call at the wrong problem.

Both runner-level calls are still required today, so neither can just be deleted:
- [examples/run_grpo.py L260](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_grpo.py#L260) is the only teardown for the `grpo_train_sync` branch, which [`_select_trainer`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_grpo.py#L43-L56) returns when `data_plane.enabled`.
- [examples/nemo_gym/run_grpo_nemo_gym.py L353](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_grpo_nemo_gym.py#L353) is the only teardown for that runner's [`collect_trajectories` branch](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_grpo_nemo_gym.py#L279).

### AI-1
Preferred: give [`grpo_train_sync`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo_sync.py#L380) the same wrapper, and confirm the [`collect_trajectories`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_grpo_nemo_gym.py#L279) branch, and **only then** drop the runner-level calls, so teardown has one owner. If that is more than you want in this PR: add a one-line comment on `grpo_train_sync` saying its caller owns environment teardown. No suggestion block — the change spans several files.

nemo_rl/environments/utils.py
terrykong, line 145: **No action needed** — non-blocking follow-up, a separate PR is fine.
**TL;DR — PPO still carries a third hand-rolled copy of this loop, and unlike this helper it does not dedup, so an aliased train/val map asks the same actor to stop twice.**

[nemo_rl/algorithms/ppo.py L2271-L2281](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/ppo.py#L2271-L2281) repeats the same logic inline:

```python
for env_dict in (task_to_env, val_task_to_env):
    if env_dict is None: continue
    for task_name, env in env_dict.items():
        print(f"🛑 Shutting down environment {task_name}...")
        try: ray.get(env.shutdown.remote(), timeout=10)
        except Exception:
            try: ray.kill(env)
```

Separately, the synchronous [`ppo_train`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/ppo.py#L1237) has no environment teardown at all, and the only `finally` in [examples/run_ppo.py L214](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_ppo.py#L214) is `shutdown_telemetry()`. [`nemo_rl/algorithms/ppo.py`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/ppo.py) mentions nemo_gym in 10 places, so this path is live.

### Follow-up
Point PPO at this helper and give [`ppo_train`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/ppo.py#L1237) the same wrapper — that finishes what this PR started, one shared helper for teardown. No suggestion block — different file than this anchor.

terrykong (commented):
Nice cleanup, and it closes a real gap. Four runners had no environment teardown at all before and now get it through the wrappers: [examples/run_vlm_grpo.py L174](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_vlm_grpo.py#L174), [examples/run_grpo_sliding_puzzle.py L289](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_grpo_sliding_puzzle.py#L289), [examples/run_distillation.py L113](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/run_distillation.py#L113) and [examples/nemo_gym/run_distillation_nemo_gym.py L168](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/examples/nemo_gym/run_distillation_nemo_gym.py#L168). Moving `shutdown_environments` into [nemo_rl/environments/utils.py](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L145) is the right home: it sits with `create_env` and `ENV_REGISTRY`, only touches `EnvironmentInterface` maps, creates no import cycle, and stops the examples importing a non-GRPO helper out of [nemo_rl/algorithms/grpo.py](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/algorithms/grpo.py#L4033). Taking `max_restarts` from -1 to 0 is a real behavior change and the right one.

Two housekeeping asks alongside the inline comments.

**The description and one commit message have drifted past the base.** The description says "Makes `NemoGym.shutdown()` idempotent, including shutdown before startup completes", and commit 746cea4e6's bug #2 says the second call "raised and fell through to ray.kill -- making ungraceful kill the *normal* exit path". Neither is true of the current base — the `if self.rh is None: return` guard is already there ([base nemo_gym.py L1020-L1027](https://github.com/NVIDIA-NeMo/RL/blob/a77c04ea30a206b6b6e64fc28065dd0668fd35d7/nemo_rl/environments/nemo_gym.py#L1020-L1027)). That guard landed on main separately in #3589 (e98357616) after these commits were written, so the code is fine; the description and commit message just need a refresh.

**One stale duplicate test to delete.** [`test_shutdown_environments_drains_unique_actors_before_kill`](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/algorithms/test_grpo.py#L1961) still patches `nemo_rl.algorithms.grpo.ray.get`/`.kill` for a function that now lives in [nemo_rl/environments/utils.py](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/nemo_rl/environments/utils.py#L145), and passes only because both modules bind the same `ray` module object; its `assert timeout == 10` survives only because `10.0 == 10`. The new tests at [test_environment_utils.py L67](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/environments/test_environment_utils.py#L67) and [L97](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/environments/test_environment_utils.py#L97) cover it fully. Also drop the then-unused import at [test_grpo.py L80](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/algorithms/test_grpo.py#L80).

How this was checked: the repo's pinned ruff hooks (`check`, `check --select I`, `format --check` at v0.9.9) are clean on all 10 changed files — `pre-commit` itself cannot build a venv on macOS, so the hooks were run directly at the pinned version. The 7 new tests were run and pass (30 passed across [test_environment_utils.py](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/environments/test_environment_utils.py) and [test_nemo_gym_utils.py](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/environments/test_nemo_gym_utils.py), plus the two wrapper tests), and all of them are collected by a real L0 lane — `tests/unit/environments/` by [L0_Unit_Tests_Environments.sh L21](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/L0_Unit_Tests_Environments.sh#L21) and `tests/unit/algorithms/` by [L0_Unit_Tests_Algorithms.sh L22](https://github.com/NVIDIA-NeMo/RL/blob/d9c789f8c4116bbe9b04554ecc56c4b9e0bae7a9/tests/unit/L0_Unit_Tests_Algorithms.sh#L22).

_Generated by Claude Code_

nemo_rl/environments/nemo_gym.py
ananthsub: Fixed in 45d8f836. The actor comment and test now describe `_require_spinup()` rejecting calls on a restarted actor; they no longer claim `self.rch` raises `AttributeError`.

nemo_rl/environments/nemo_gym.py
ananthsub: Fixed in 45d8f836. The shutdown docstring no longer attributes idempotence or kill escalation to this stack. It now explains why the `RunHelper` handle is cleared before invoking its non-idempotent shutdown.

nemo_rl/environments/utils.py
ananthsub: Fixed in 45d8f836. The helper now describes deduplication as avoiding duplicate stop requests and logs, and documents that both `None` and empty mappings are skipped.

nemo_rl/algorithms/grpo.py
ananthsub: Fixed in 45d8f836. `grpo_train()` now forwards every implementation argument by keyword. Success and failure tests both verify that environment shutdown runs from the `finally` block.

nemo_rl/algorithms/distillation.py
ananthsub: Fixed in 45d8f836. `distillation_train()` now forwards every implementation argument by keyword, with success and failure coverage for its shutdown path.

tests/unit/environments/test_environment_utils.py
ananthsub: Fixed in 45d8f836. The tests now put the wedged actor before a healthy actor and verify cleanup continues. A separate case makes `ray.kill()` fail and verifies later actors are still shut down.

nemo_rl/algorithms/grpo.py
ananthsub: Fixed in 45d8f836 with explicit ownership at the runners. Legacy synchronous and asynchronous GRPO own their cleanup. `run_grpo.py` retains runner cleanup only for the TransferQueue synchronous trainer, and `run_grpo_nemo_gym.py` retains it only for trajectory collection. The duplicate legacy synchronous cleanup is removed.

nemo_rl/environments/utils.py
ananthsub: Acknowledged. PPO cleanup remains outside this stack as the comment requested; no PPO behavior was changed here.

examples/run_grpo.py
```diff
@@ -217,6 +218,7 @@ def _make_policy(**kwargs):
                 print("🚀 Running async GRPO training")
 
                 # Run async GRPO training
+                trainer_owns_environment_shutdown = True
```
terrykong, line 221: **1 action item.**
**TL;DR — `async_grpo_train` does not start its `try` until 495 lines into the function, so handing it ownership before the call means a config error in that prologue leaks the environments the runner used to clean up.**

The new ownership rule is right, and it holds for [`grpo_train`](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L4032) and `distillation_train`: their `try` is the first statement, so the `finally` covers everything the caller could have hit.

Async is different. [`async_grpo_train`](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L4427) starts at line 4427, but its [`try`](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L4922) is at 4922 and the [`finally`](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L5935) that calls [`shutdown_environments`](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L5955) hangs off it. Everything between is unprotected, and it validates config — for example [`assert backend in (...)`](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L4469), [`assert should_use_async_rollouts(...)`](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L4473) and [`raise NotImplementedError`](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L4488).

Since this line runs before the call, [the runner then skips its own teardown](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/examples/run_grpo.py#L263) and nothing stops the Gym servers. Before this PR the runner's `finally` was unconditional and caught it.

How it shows up: run async GRPO with a Gym environment and `policy.generation.vllm_cfg.async_engine=false`. `setup()` has already spun the actor up, the assert at [grpo.py L4473](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/nemo_rl/algorithms/grpo.py#L4473) fires, and the servers stay up — the leak this PR is meant to close.

Same line in the other runner: [run_grpo_nemo_gym.py L316](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/examples/nemo_gym/run_grpo_nemo_gym.py#L316), with its [skip](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7/examples/nemo_gym/run_grpo_nemo_gym.py#L356).

### AI-1
Give `async_grpo_train` the same thin wrapper you gave the other two — rename the body to `_async_grpo_train_impl` and have `async_grpo_train` be a `try`/`finally` around it — so ownership is true from its first statement. Then this flag needs no change. No suggestion block: the edit is in `nemo_rl/algorithms/grpo.py`, a different file from this anchor.

If you would rather not reshape async in this PR, the smaller move is to set this flag `False` for the async branch and let the runner keep cleaning up. The second call is safe now, so the only cost is a repeated log line.

terrykong (commented):
One follow-up on the teardown-ownership change. Everything else from the last round checks out at [790b8692a](https://github.com/NVIDIA-NeMo/RL/blob/790b8692ac5aa227568f0a7f8de93a7e4667b1a7) — I re-read each thread against the code rather than the replies: the actor comment and both docstrings are accurate now, both wrappers forward by keyword, the new environment tests put the wedged actor first (better than what I suggested), the stale duplicate test is gone, and the PR description and commit messages no longer claim the shutdown guard came from this stack.

Two small things, neither needing a reply: the "Fixed in 45d8f83" notes point at a commit the rebase removed, so they no longer resolve; and GitHub now reports the branch as conflicting with the base, so it will need a rebase before merge.

_Generated by Claude Code_

terrykong:
@ananthsub bump on the review + CI (you might need to merge in main to get the CI to pass - there were a few breakages on main)

terrykong (approved):
LGTM. The async gap from the last round is closed: dropping `trainer_owns_environment_shutdown = True` from the async branch in both runners means the runner's `finally` covers `async_grpo_train`'s ~495-line prologue again, so a config error there no longer leaks the Gym servers. Ownership now reads cleanly — `grpo_train` owns its own teardown, everything else stays with the runner.

Checked at 937a1852464ec98c85d067d38512e880fc4d64f1: ruff (check, check --select I, format --check at the pinned v0.9.9) clean on all 10 changed files, the environment and wrapper unit tests pass locally (38 + 2), and CI is 73/73 green. The two commits still touch only the 10 files they did before — the `data_plane.factory` and dedup-validation changes in the range diff came from main via the rebase, not from this stack.

The PPO cleanup stays a follow-up as discussed.

_Generated by Claude Code_

PR #3369, opened 2026-07-27
feat(nemo-gym): define shard configuration

ananthsub:
## Summary

Defines the configuration used to split one NeMo-Gym stack across several Ray actors. This PR validates and composes shard configurations but does not create multiple actors yet.

`nemo_rl/environments/nemo_gym_shards.py` adds:

- `ShardSpec`, which describes one named configuration slice, its inherited entries, explicit overrides, replicas, CPU allocation, and optional port range.
- `ShardPlan`, which describes shared entries, shared overrides, duplicate-entry allowances, and placement settings for the complete stack.
- `parse_shard_plan()`, which validates shard ownership and preserves the unchanged unsharded fallback.
- `apply_shard_overlay()`, which composes the Gym configuration for one shard.
- `apply_shard_log_dir()`, which gives each shard instance a separate log directory.

A one-shard plan automatically owns unclaimed entries. A multi-shard plan must assign each Gym entry exactly once through `inherited_overlays` or `common_inherited_overlays`. Missing, duplicate, and unknown claims fail with the resolved entry inventory. NeMo RL integration settings such as `effort_levels` and `tokenizer_config` are excluded from that inventory.

Configuration precedence is deterministic: inherited values, then `common_overrides`, then shard-specific `overrides`. A non-null top-level `config_paths` value is rejected because each path must belong to a specific shard. Explicit port ranges use Gym's inclusive bounds and must stay within ports 1 through 65535.

## Example

```yaml
env:
  nemo_gym:
    common_inherited_overlays:
      - policy_model
    allowed_duplicate_entries:
      - policy_model
    shards:
      - name: tools
        inherited_overlays:
          - code_gen
        config_paths:
          - resources_servers/workplace_assistant/configs/workplace_assistant.yaml
      - name: judged
        inherited_overlays:
          - math_with_judge
        config_paths:
          - resources_servers/math_with_judge/configs/math_with_judge.yaml
```

## Stack

This PR depends on #3633. Multi-actor creation follows in #3371.

### DIFF
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -32,6 +32,11 @@ from nemo_rl.distributed.virtual_cluster import (
     _get_node_ip_local,
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
+from nemo_rl.environments.nemo_gym_shards import (
+    SHARDING_CONFIG_KEYS,
+    ShardConfigError,
+    parse_shard_plan,
+)
 from nemo_rl.utils.timer import Timer
 from nemo_rl.utils.venvs import make_actor_runtime_env
 
@@ -651,6 +656,24 @@ def build_nemo_gym_config(
             actor patches its tokenizer the same way the driver does.
     """
     nemo_gym_dict = dict(env_configs["nemo_gym"])
+
+    # Validate the shards block even though only single-actor creation is wired
+    # up so far, so a malformed or premature sharded config fails at setup with
+    # a precise message instead of silently running unsharded.
+    shard_plan = parse_shard_plan(nemo_gym_dict)
+    if shard_plan is not None:
+        raise ShardConfigError(
+            f"env.nemo_gym.shards defines {len(shard_plan.shards)} shards "
+            f"({', '.join(s.name for s in shard_plan.shards)}), but multi-actor "
+            f"creation is not wired up yet. Remove 'shards' and use "
+            f"'config_paths' to run this job on a single actor."
+        )
+
+    # NeMo-RL-only keys are consumed here and must never reach Gym: the merged
+    # config is serialized into every Gym child process, and unrecognized
+    # dict-shaped top-level keys are parsed as server instance configs.
+    for key in SHARDING_CONFIG_KEYS:
+        nemo_gym_dict.pop(key, None)
 
     # NeMo-RL-side detection knobs are top-level NemoGymConfig fields
     # (where the detector reads them), not part of Gym's global config.
--- a/nemo_rl/environments/nemo_gym_shards.py
+++ b/nemo_rl/environments/nemo_gym_shards.py
@@ -1,0 +1,260 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Schema for splitting one NeMo-Gym stack across several actors.
+
+A single ``NemoGym`` actor runs every environment in the job on one node, which
+caps a run at what that node can host. Sharding partitions the Gym config into
+groups that each get their own actor on their own node.
+
+NeMo RL composes shard configs; it never interprets them. Everything forwarded
+into a shard's merge is opaque Gym config, and validation compares entry
+*names*, never meanings.
+"""
+
+from dataclasses import dataclass, field
+from typing import Any, Mapping
+
+from omegaconf import OmegaConf
+
+# Keys under ``env.nemo_gym`` that NeMo RL consumes itself. They must never
+# reach NeMo Gym: the merged config is serialized into NEMO_GYM_CONFIG_DICT for
+# every one of the ~60 child processes, and Gym treats unrecognized dict-shaped
+# top-level keys as server instance configs, which can trip its
+# error_on_almost_servers path.
+SHARDING_CONFIG_KEYS = frozenset(
+    {"shards", "common_overrides", "allowed_duplicate_entries"}
+)
+
+# Per-shard keys consumed by NeMo RL rather than forwarded to Gym.
+SHARD_SPEC_KEYS = frozenset(
+    {
+        "name",
+        "config_paths",
+        "overrides",
+        "replicas",
+        "actor_cpus",
+        "port_range_low",
+        "port_range_high",
+    }
+)
+
+
+class ShardConfigError(ValueError):
+    """Raised for a malformed ``env.nemo_gym.shards`` block."""
+
+
+@dataclass(frozen=True)
+class ShardSpec:
+    """One shard: a slice of the Gym config that gets its own actor.
+
+    ``replicas`` instances are created from a single merge, so replicas are
+    identical apart from the node they land on.
+    """
+
+    name: str
+    config_paths: list[str]
+    overrides: dict[str, Any] = field(default_factory=dict)
+    replicas: int = 1
+    actor_cpus: float | None = None
+    port_range_low: int | None = None
+    port_range_high: int | None = None
+
+
+@dataclass(frozen=True)
+class ShardPlan:
+    """The parsed ``shards`` block plus the settings that apply to all shards."""
+
+    shards: list[ShardSpec]
+    common_overrides: dict[str, Any] = field(default_factory=dict)
+    allowed_duplicate_entries: frozenset[str] = frozenset()
+
+    @property
+    def total_instances(self) -> int:
+        """Actors to create, counting replicas. One node is needed per instance."""
+        return sum(shard.replicas for shard in self.shards)
+
+
+def _as_plain_dict(value: Any, *, context: str) -> dict[str, Any]:
+    """Coerce a mapping (possibly OmegaConf-backed) to a plain dict."""
+    if value is None:
+        return {}
+    if OmegaConf.is_config(value):
+        value = OmegaConf.to_container(value, resolve=True)
+    if not isinstance(value, Mapping):
+        raise ShardConfigError(
+            f"{context} must be a mapping, got {type(value).__name__}"
+        )
+    return dict(value)
+
+
+def _parse_shard(raw: Any, index: int) -> ShardSpec:
+    entry = _as_plain_dict(raw, context=f"shards[{index}]")
+
+    unknown = set(entry) - SHARD_SPEC_KEYS
+    if unknown:
+        raise ShardConfigError(
+            f"shards[{index}] has unrecognized keys {sorted(unknown)}. "
+            f"Gym config overlays belong under this shard's 'overrides'."
+        )
+
+    name = entry.get("name")
+    if not isinstance(name, str) or not name.strip():
+        raise ShardConfigError(f"shards[{index}] needs a non-empty string 'name'")
+
+    config_paths = entry.get("config_paths")
+    if OmegaConf.is_config(config_paths):
+        config_paths = OmegaConf.to_container(config_paths, resolve=True)
+    if not isinstance(config_paths, list) or not config_paths:
+        raise ShardConfigError(
+            f"shard '{name}' needs at least one entry in 'config_paths'"
+        )
+
+    replicas = entry.get("replicas", 1)
+    if not isinstance(replicas, int) or isinstance(replicas, bool) or replicas < 1:
+        raise ShardConfigError(
+            f"shard '{name}' has replicas={replicas!r}; must be an integer >= 1"
+        )
+
+    port_low = entry.get("port_range_low")
+    port_high = entry.get("port_range_high")
+    if (port_low is None) != (port_high is None):
+        raise ShardConfigError(
+            f"shard '{name}' must set both port_range_low and port_range_high, or neither"
+        )
+    if port_low is not None and port_low >= port_high:
+        raise ShardConfigError(
+            f"shard '{name}' has an empty port range [{port_low}, {port_high})"
+        )
+
+    return ShardSpec(
+        name=name,
+        config_paths=[str(path) for path in config_paths],
+        overrides=_as_plain_dict(
+            entry.get("overrides"), context=f"shard '{name}' overrides"
+        ),
+        replicas=replicas,
+        actor_cpus=entry.get("actor_cpus"),
+        port_range_low=port_low,
+        port_range_high=port_high,
+    )
+
+
+def find_gym_config_entries(nemo_gym_config: Mapping[str, Any]) -> list[str]:
+    """Return top-level keys that Gym would read as server instance configs.
+
+    Mirrors Gym's own rule (``filter_for_server_instance_configs``): any
+    dict-shaped top-level key that is not a known setting is an entry overlay.
+    Under sharding these are ambiguous — they have no shard to belong to — so
+    the caller rejects them and asks the author to place them explicitly.
+    """
+    return sorted(
+        key
+        for key, value in nemo_gym_config.items()
+        if key not in SHARDING_CONFIG_KEYS
+        and (isinstance(value, Mapping) or OmegaConf.is_dict(value))
+    )
+
+
+def parse_shard_plan(nemo_gym_config: Mapping[str, Any]) -> ShardPlan | None:
+    """Parse ``env.nemo_gym.shards``, or return None when the job is unsharded.
+
+    Returning None is the backward-compatible path: a config with
+    ``config_paths`` and no ``shards`` behaves exactly as it did before
+    sharding existed.
+
+    Raises:
+        ShardConfigError: the shards block is malformed, or it is combined with
+            config that has no unambiguous shard to belong to.
+    """
+    if "shards" not in nemo_gym_config:
+        return None
+
+    raw_shards = nemo_gym_config["shards"]
+    if OmegaConf.is_config(raw_shards):
+        raw_shards = OmegaConf.to_container(raw_shards, resolve=True)
+    if not isinstance(raw_shards, list) or not raw_shards:
+        raise ShardConfigError("env.nemo_gym.shards must be a non-empty list")
+
+    # Anything whose scope would be implicit has to be placed by hand: a
+    # top-level config_paths or entry overlay could plausibly belong to one
+    # shard or to all of them, and guessing wrong silently changes what each
+    # actor hosts.
+    if "config_paths" in nemo_gym_config:
+        raise ShardConfigError(
+            "env.nemo_gym.config_paths cannot be combined with 'shards'. "
+            "Move each path into the shard that should host it."
+        )
+    stray_entries = find_gym_config_entries(nemo_gym_config)
+    if stray_entries:
+        raise ShardConfigError(
+            f"Gym config overlays {stray_entries} sit at the top level alongside "
+            f"'shards', so their scope is ambiguous. Move each one into "
+            f"'common_overrides' (to apply to every shard) or into the "
+            f"'overrides' of the shard that hosts it."
+        )
+
+    shards = [_parse_shard(raw, index) for index, raw in enumerate(raw_shards)]
+
+    duplicate_names = sorted(
+        {
+            shard.name
+            for shard in shards
+            if [s.name for s in shards].count(shard.name) > 1
+        }
+    )
+    if duplicate_names:
+        raise ShardConfigError(f"Duplicate shard names: {duplicate_names}")
+
+    raw_allowed = nemo_gym_config.get("allowed_duplicate_entries") or []
+    if OmegaConf.is_config(raw_allowed):
+        raw_allowed = OmegaConf.to_container(raw_allowed, resolve=True)
+    if not isinstance(raw_allowed, list) or not all(
+        isinstance(entry, str) for entry in raw_allowed
+    ):
+        raise ShardConfigError(
+            "env.nemo_gym.allowed_duplicate_entries must be a list of strings"
+        )
+
+    return ShardPlan(
+        shards=shards,
+        common_overrides=_as_plain_dict(
+            nemo_gym_config.get("common_overrides"), context="common_overrides"
+        ),
+        allowed_duplicate_entries=frozenset(raw_allowed),
+    )
+
+
+def apply_shard_overlay(
+    base_config: dict[str, Any], plan: ShardPlan, shard: ShardSpec
+) -> dict[str, Any]:
+    """Build one shard's Gym config: base, then common overrides, then the shard's.
+
+    The ``config_paths`` merge itself still happens inside Gym, exactly as for
+    an unsharded job; this only settles which paths and overlays that shard
+    starts from. Per-shard overrides win over ``common_overrides`` on conflict.
+    """
+    merged = OmegaConf.to_container(
+        OmegaConf.merge(
+            OmegaConf.create(base_config),
+            OmegaConf.create(plan.common_overrides),
+            OmegaConf.create(shard.overrides),
+        ),
+        resolve=True,
+    )
+    assert isinstance(merged, dict)
+    merged["config_paths"] = list(shard.config_paths)
+    if shard.port_range_low is not None:
+        merged["port_range_low"] = shard.port_range_low
+        merged["port_range_high"] = shard.port_range_high
+    return merged
--- a/tests/unit/environments/test_nemo_gym_shards.py
+++ b/tests/unit/environments/test_nemo_gym_shards.py
@@ -1,0 +1,208 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Unit tests for the NeMo-Gym shard schema (pure Python, no Gym import)."""
+
+import pytest
+from omegaconf import OmegaConf
+
+from nemo_rl.environments.nemo_gym_shards import (
+    ShardConfigError,
+    ShardPlan,
+    ShardSpec,
+    apply_shard_overlay,
+    find_gym_config_entries,
+    parse_shard_plan,
+)
+
+
+def _sharded_config(**overrides):
+    config = {
+        "uv_venv_dir": "/opt/gym_venvs",
+        "num_gpu_nodes": 2,
+        "shards": [
+            {"name": "judged", "config_paths": ["judge.yaml"]},
+            {"name": "tools", "config_paths": ["proxy.yaml", "tools.yaml"]},
+        ],
+    }
+    config.update(overrides)
+    return config
+
+
+def test_unsharded_config_returns_no_plan():
+    """config_paths without shards is the pre-sharding path and must stay working."""
+    assert parse_shard_plan({"config_paths": ["gym.yaml"], "num_gpu_nodes": 1}) is None
+
+
+def test_parse_shard_plan_reads_shards_and_common_settings():
+    plan = parse_shard_plan(
+        _sharded_config(
+            common_overrides={"policy_model": {"num_workers": 4}},
+            allowed_duplicate_entries=["policy_model"],
+        )
+    )
+
+    assert [shard.name for shard in plan.shards] == ["judged", "tools"]
+    assert plan.shards[1].config_paths == ["proxy.yaml", "tools.yaml"]
+    assert plan.common_overrides == {"policy_model": {"num_workers": 4}}
+    assert plan.allowed_duplicate_entries == frozenset({"policy_model"})
+
+
+def test_parse_shard_plan_accepts_omegaconf_input():
+    """Job configs arrive as OmegaConf nodes, not plain dicts."""
+    plan = parse_shard_plan(OmegaConf.create(_sharded_config()))
+
+    assert [shard.name for shard in plan.shards] == ["judged", "tools"]
+    assert isinstance(plan.shards[0].config_paths, list)
+
+
+def test_replicas_count_toward_total_instances():
+    plan = parse_shard_plan(
+        _sharded_config(
+            shards=[
+                {"name": "judged", "config_paths": ["judge.yaml"]},
+                {"name": "tools", "config_paths": ["tools.yaml"], "replicas": 3},
+            ]
+        )
+    )
+
+    assert plan.total_instances == 4
+
+
+def test_top_level_config_paths_conflicts_with_shards():
+    with pytest.raises(ShardConfigError, match="cannot be combined with 'shards'"):
+        parse_shard_plan(_sharded_config(config_paths=["gym.yaml"]))
+
+
+def test_stray_entry_overlay_is_rejected():
+    """A top-level Gym overlay has no unambiguous shard, so it must be placed."""
+    with pytest.raises(ShardConfigError, match="nl2bash_judge_model"):
+        parse_shard_plan(
+            _sharded_config(
+                nl2bash_judge_model={"responses_api_models": {"local_vllm_model": {}}}
+            )
+        )
+
+
+@pytest.mark.parametrize(
+    ("shards", "expected"),
+    [
+        ([], "non-empty list"),
+        ([{"config_paths": ["a.yaml"]}], "non-empty string 'name'"),
+        ([{"name": "  ", "config_paths": ["a.yaml"]}], "non-empty string 'name'"),
+        ([{"name": "a", "config_paths": []}], "at least one entry in 'config_paths'"),
+        ([{"name": "a"}], "at least one entry in 'config_paths'"),
+        (
+            [{"name": "a", "config_paths": ["a.yaml"], "replicas": 0}],
+            "must be an integer >= 1",
+        ),
+        (
+            [{"name": "a", "config_paths": ["a.yaml"], "port_range_low": 5000}],
+            "both port_range_low and port_range_high",
+        ),
+        (
+            [
+                {
+                    "name": "a",
+                    "config_paths": ["a.yaml"],
+                    "port_range_low": 5200,
+                    "port_range_high": 5200,
+                }
+            ],
+            "empty port range",
+        ),
+        (
+            [{"name": "a", "config_paths": ["a.yaml"], "judge_model": {}}],
+            "unrecognized keys",
+        ),
+        (
+            [
+                {"name": "dup", "config_paths": ["a.yaml"]},
+                {"name": "dup", "config_paths": ["b.yaml"]},
+            ],
+            "Duplicate shard names",
+        ),
+    ],
+)
+def test_malformed_shards_are_rejected(shards, expected):
+    with pytest.raises(ShardConfigError, match=expected):
+        parse_shard_plan({"shards": shards})
+
+
+def test_allowed_duplicate_entries_must_be_strings():
+    with pytest.raises(ShardConfigError, match="list of strings"):
+        parse_shard_plan(_sharded_config(allowed_duplicate_entries=[{"nope": 1}]))
+
+
+def test_apply_shard_overlay_layers_shard_over_common():
+    plan = ShardPlan(
+        shards=[],
+        common_overrides={
+            "policy_model": {"num_workers": 4, "timeout": 30},
+            "shared_judge": {"replicas": 1},
+        },
+    )
+    shard = ShardSpec(
+        name="judged",
+        config_paths=["judge.yaml"],
+        overrides={"policy_model": {"num_workers": 16}},
+    )
+
+    merged = apply_shard_overlay({"default_host": "10.0.0.1"}, plan, shard)
+
+    assert merged["config_paths"] == ["judge.yaml"]
+    assert merged["default_host"] == "10.0.0.1"
+    # Shard wins on conflict, and sibling keys from common survive the merge.
+    assert merged["policy_model"] == {"num_workers": 16, "timeout": 30}
+    assert merged["shared_judge"] == {"replicas": 1}
+
+
+def test_apply_shard_overlay_applies_per_shard_port_range():
+    shard = ShardSpec(
+        name="ci",
+        config_paths=["a.yaml"],
+        port_range_low=6000,
+        port_range_high=6099,
+    )
+
+    merged = apply_shard_overlay(
+        {"port_range_low": 5000, "port_range_high": 5999}, ShardPlan(shards=[]), shard
+    )
+
+    assert (merged["port_range_low"], merged["port_range_high"]) == (6000, 6099)
+
+
+def test_apply_shard_overlay_keeps_global_ports_by_default():
+    merged = apply_shard_overlay(
+        {"port_range_low": 5000, "port_range_high": 5999},
+        ShardPlan(shards=[]),
+        ShardSpec(name="prod", config_paths=["a.yaml"]),
+    )
+
+    assert (merged["port_range_low"], merged["port_range_high"]) == (5000, 5999)
+
+
+def test_find_gym_config_entries_ignores_scalars_and_known_keys():
+    entries = find_gym_config_entries(
+        {
+            "uv_venv_dir": "/opt/venvs",
+            "num_gpu_nodes": 2,
+            "shards": [{"name": "a"}],
+            "common_overrides": {"x": 1},
+            "allowed_duplicate_entries": ["policy_model"],
+            "genrm_model": {"responses_api_models": {}},
+            "safety_judge_model": {"responses_api_models": {}},
+        }
+    )
+
+    assert entries == ["genrm_model", "safety_judge_model"]


nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+            OmegaConf.create(shared_base),
+            OmegaConf.create(selected_inherited),
+            OmegaConf.create(plan.common_overrides),
+            OmegaConf.create(shard.overrides),
+        ),
+        resolve=True,
+    )
+    assert isinstance(merged, dict)
+    merged["config_paths"] = list(shard.config_paths)
+    if shard.port_range_low is not None:
```
terrykong: **1 action item.**
**TL;DR — the per-shard port range is validated, merged and written into the shard config, then overwritten before Gym ever sees it.**

`apply_shard_overlay` writes `port_range_low`/`port_range_high` into the merged dict here. That dict becomes `initial_global_config_dict`, and [`NemoGym._spinup` overwrites both keys unconditionally](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L605-L606):

```python
initial_global_config_dict["port_range_low"] = _gym_port_low
initial_global_config_dict["port_range_high"] = _gym_port_high
```

Those values come from `self.cfg`, the [`NemoGymConfig`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L209), whose `port_range_low` is a `NotRequired` field that nothing ever populates — [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L1266-L1277) does not set it, and neither does `_build_gym_actor_config` in #3371. So every job gets the 5000-5999 default.

This already breaks a shipped recipe: [`swe_teacher.yaml` asks for 15001-20000](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/examples/nemo_gym/nemotron-3-ultra/swe_teacher.yaml#L413-L414) under `env.nemo_gym` and silently runs on 5000-5999. It is live — launched by [ultra_launch.sh](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh), documented in [nemotron-3-ultra.md](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/docs/guides/nemotron-3-ultra.md), and listed in [test_config.py](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/tests/unit/utils/test_config.py#L29). Every other recipe sets exactly 5000/5999, which is why nobody noticed.

The global knob being dead is pre-existing. What this PR adds is a second, per-shard copy of the same dead knob.

### AI-1
Not a suggestion block — the fix lands in a different file than this anchor. In [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L1266-L1277), pop `port_range_low`/`port_range_high` off `nemo_gym_dict` and pass them onto the returned `NemoGymConfig`, so the one place `_spinup` reads is populated. The per-shard values then reach Gym too, because `apply_shard_overlay` already puts them in the shard's config.

One more thing while you are here: the [bound at line 186](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L186) should be `port_high > 65_535`. NeMo RL says so in writing — [the port map states](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/distributed/virtual_cluster.py#L88) *"Python port-range bounds below are half-open: [low, high)"* — and this module [rejects `low >= high`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L191) to match. (The Gym row of that same map is written [`5000-5999`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/distributed/virtual_cluster.py#L95), inclusive-style, under that half-open heading, so the ambiguity is already in-tree.), but Gym includes both ends. The two messages here disagree too — [line 189](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L189) says `[1, 65536]`, [line 193](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L193) says `[low, high)`. This is unreachable until the plumbing above is fixed, so do that first.

<details><summary>Why 65536 kills the actor instead of being retried</summary>

Gym picks a port with [`randint(port_range_low, port_range_high)`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L1067), inclusive on both ends, so 65536 can be drawn. `socket.bind(("", 65536))` raises `OverflowError: bind(): port must be 0-65535`. That is not an `OSError`, and Gym's [retry loop](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L1069-L1073) catches only `OSError`, so it escapes and kills the actor.

The mismatch is harmless at every other value — Gym drawing 20000 from `swe_teacher`'s range is fine. Only 65536 is an invalid port.
</details>

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+    Returning None is the backward-compatible path: a config with
+    ``config_paths`` and no ``shards`` behaves exactly as it did before
+    sharding existed.
+
+    Raises:
+        ShardConfigError: the shards block is malformed, or it is combined with
+            config that has no unambiguous shard to belong to.
+    """
+    if "shards" not in nemo_gym_config:
+        return None
```
terrykong: **1 action item.**
**TL;DR — a config with the sharding settings but no `shards:` block runs unsharded and drops every one of those settings without a word.**

PR-introduced. `parse_shard_plan` returns `None` as soon as `shards` is absent, so `common_overrides`, `common_inherited_overlays` and `allowed_duplicate_entries` are never read. [`build_nemo_gym_config` then pops all four keys](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L1231-L1232). That pop is only reachable in this case, because the sharded path [raises first](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L1220) — so the pop is what makes the drop silent.

Same exit code, no warning, different training. The trigger is ordinary: commenting out a `shards:` block while iterating, or a rebase dropping it. No recipe in `examples/` carries any of these four keys today, so raising here cannot break an existing config.

### AI-1
```suggestion
    if "shards" not in nemo_gym_config:
        stray = sorted(
            key
            for key in SHARDING_CONFIG_KEYS
            if key != "shards" and key in nemo_gym_config
        )
        if stray:
            raise ShardConfigError(
                f"env.nemo_gym has sharding keys {stray} but no 'shards' block. "
                f"Add a 'shards' block, or remove these keys."
            )
        return None
```

**Context — no action.** Keep the [pop](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L1231-L1232) after you add this. It becomes unreachable from `build_nemo_gym_config` alone once both branches raise, but `_build_gym_actor_config` in #3371 runs it for every actor, so it is not dead.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+DEFAULT_REPLICAS = 1
+SHARD_NAME_PATTERN = re.compile(r"^[A-Za-z0-9][A-Za-z0-9_.-]*$")
+
+# Gym's log-directory key (``nemo_gym.global_config.NEMO_GYM_LOG_DIR_KEY_NAME``).
+# Restated rather than imported: this module is imported on the driver, and
+# nemo_gym is installed only in the actor's venv.
+GYM_LOG_DIR_KEY = "nemo_gym_log_dir"
+
+# Per-shard keys consumed by NeMo RL rather than forwarded to Gym.
+SHARD_SPEC_KEYS = frozenset(
```
terrykong: **1 action item.**
**TL;DR — `SHARD_SPEC_KEYS` copies `ShardSpec`'s field names by hand, so a future per-shard knob gets rejected with advice that points the wrong way.**

PR-introduced. This list restates the eight fields of [`ShardSpec`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L79). They agree today. If a later PR adds a knob to `ShardSpec` and [`_parse_shard`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L139) but misses this frozenset, the [unknown-key check](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L142) rejects the user's own new key and tells them to move it under `overrides`.

This kind of drift already happened inside this PR: the tip commit `b6bdd3303` "fix(nemo-gym): exclude integration settings from overlay inventory" fixes exactly that in the sibling list [`NEMO_RL_DICT_CONFIG_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L49), whose entries must stay in step with pops in another file ([`tokenizer_config`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L1238), [`effort_levels`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L581)). The schema is still growing — #3373 adds `placement_strategy` to `ShardPlan`.

### AI-1
Not a suggestion block — the fix is a move plus an import change. Derive the set from the dataclass:

```python
from dataclasses import dataclass, field, fields, replace

SHARD_SPEC_KEYS = frozenset(f.name for f in fields(ShardSpec))
```

The assignment must move **below** the `ShardSpec` dataclass. It sits at line 60 and the dataclass starts at [line 79](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L79), so pasting it in place raises `NameError` at import. Anywhere after line 93 works; [first use](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L142) is line 142.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+
+    raw_allowed = nemo_gym_config.get("allowed_duplicate_entries") or []
+    if OmegaConf.is_config(raw_allowed):
+        raw_allowed = OmegaConf.to_container(raw_allowed, resolve=True)
+    if not isinstance(raw_allowed, list) or not all(
+        isinstance(entry, str) for entry in raw_allowed
+    ):
+        raise ShardConfigError(
+            "env.nemo_gym.allowed_duplicate_entries must be a list of strings"
+        )
```
terrykong: **1 action item.**
**TL;DR — `allowed_duplicate_entries` hand-rolls its checks, so it accepts input every sibling list rejects.**

PR-introduced, and small. Every other string list goes through [`_as_string_set`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L118). This one does not, so `["", "dup", "dup"]` is accepted — empty string and duplicate both pass — while [`inherited_overlays`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L258) rejects both. The trailing `or []` also swallows `{}` without complaint, even though `{"a": 1}` correctly raises.

### AI-1
```suggestion
    raw_allowed = _as_string_set(
        nemo_gym_config.get("allowed_duplicate_entries"),
        context="env.nemo_gym.allowed_duplicate_entries",
    )
```
The variable name is kept so [`allowed_duplicate_entries=frozenset(raw_allowed)`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L340) below still works. That `frozenset(...)` call is now redundant and can be dropped.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+        OmegaConf.merge(
+            OmegaConf.create(shared_base),
+            OmegaConf.create(selected_inherited),
+            OmegaConf.create(plan.common_overrides),
+            OmegaConf.create(shard.overrides),
+        ),
+        resolve=True,
+    )
+    assert isinstance(merged, dict)
+    merged["config_paths"] = list(shard.config_paths)
```
terrykong: **1 action item.**
**TL;DR — the two override layers can set entry keys no shard claimed, and they treat `config_paths` and the port keys differently — none of it written down.**

PR-introduced, and a contract question rather than a bug. `parse_shard_plan` guarantees [each inherited entry is claimed exactly once](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L286-L322), then [`apply_shard_overlay`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L360-L372) merges `common_overrides` and `shard.overrides` on top with no check. Two results a user can hit:

- `common_overrides: {code_gen: {...}}` gives every shard a complete, startable `code_gen` resources server, including shards that never claimed it. #3371 only notices after all actors have started, so the cost is a wasted spin-up.
- `config_paths` inside `overrides` is silently clobbered by this line, while `port_range_low`/`port_range_high` inside `overrides` [survive](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L375-L377). Same layer, two behaviours.

The first is arguable — `common_overrides` means "apply to every shard", so injecting into every shard may be the point, and `allowed_duplicate_entries` exists so a user can say they meant it. The gap is that nothing states which reading is intended.

### AI-1
Not a suggestion block — this is a choice between two edits in different places. Pick either.

Preferred: at parse time, reject entry-shaped keys in the override layers when they belong to a different shard, so the failure comes before any actor starts.

If that is more than you want in this PR: write the rule into [`apply_shard_overlay`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L344)'s docstring — which keys the override layers may set, and that `config_paths` there is ignored.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+    )
+
+
+def find_gym_config_entries(nemo_gym_config: Mapping[str, Any]) -> list[str]:
+    """Return top-level keys that Gym would read as server instance configs.
+
+    Mirrors Gym's own rule (``filter_for_server_instance_configs``): any
+    dict-shaped top-level key that is not a known setting is an entry overlay.
+    Under sharding these values are inherited from the resolved parent config.
+    A shard claims them by key so their YAML bodies do not need to be copied.
```
terrykong: **1 action item.**
**TL;DR — the docstring says this mirrors Gym's rule; it is deliberately broader, and should say so.**

PR-introduced, docstring only. Gym's [`filter_for_server_instance_configs`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L386-L401) does two things: drop the 31 [reserved top-level keys](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L101-L133), then require the value to pass pydantic validation as a `ServerInstanceConfig` via [`maybe_get_server_instance_config`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L689-L704). This function excludes six keys and accepts any Mapping.

**Do not change the code.** The module deliberately cannot import Gym — it is installed only in the actor's venv, per the [comment above](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L54-L56) — so the pydantic step is not available here, and being broader is the safe direction. The one dict-shaped reserved key, `head_server`, is [overwritten unconditionally by `NemoGym._spinup`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L628-L631) anyway, and no in-tree recipe sets it. Only the wording needs fixing.

### AI-1
```suggestion
    Deliberately broader than Gym's own rule
    (``filter_for_server_instance_configs``), which also requires the value to
    validate as a ``ServerInstanceConfig``: this module cannot import Gym, so
    any dict-shaped top-level key that is not a known setting is treated as an
    entry overlay. Under sharding these values are inherited from the resolved
    parent config. A shard claims them by key so their YAML bodies do not need
    to be copied.
```

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+from pathlib import PurePosixPath
+from typing import Any, Mapping
+
+from omegaconf import OmegaConf
+
+# Keys under ``env.nemo_gym`` that NeMo RL consumes itself. They must never
+# reach NeMo Gym: the merged config is serialized into NEMO_GYM_CONFIG_DICT for
+# every one of the ~60 child processes, and Gym treats unrecognized dict-shaped
+# top-level keys as server instance configs, which can trip its
+# error_on_almost_servers path.
```
terrykong: **1 action item.**
**TL;DR — this comment gives a reason that cannot happen; the real reason is the one just before it.**

PR-introduced. Stripping these keys is right, but not because of `error_on_almost_servers`. [`is_almost_server`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L708-L728) returns False unless the dict has one of `responses_api_models` / `resources_servers` / `responses_api_agents` **and** an inner config carrying an `entrypoint`. `shards` is a list, so [`maybe_get_server_instance_config`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L689-L704) bails at its `isinstance(..., DictConfig)` guard and `is_almost_server` is False as well. Gym just ignores the key; the [raise](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L745-L754) never fires.

The real reason — these keys get serialized into `NEMO_GYM_CONFIG_DICT` for every child process — already stands on its own.

### AI-1
```suggestion
# Keys under ``env.nemo_gym`` that NeMo RL consumes itself. They must never
# reach NeMo Gym: the merged config is serialized into NEMO_GYM_CONFIG_DICT for
# every one of the ~60 child processes, so anything left here is shipped to all
# of them.
```

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+        or isinstance(port_low, bool)
+        or not isinstance(port_high, int)
+        or isinstance(port_high, bool)
+        or port_low < 1
+        or port_high > 65_536
+    ):
+        raise ShardConfigError(
+            f"shard '{name}' port range must use integers within [1, 65536]"
+        )
+    if port_low is not None and port_low >= port_high:
```
terrykong: **1 action item.**
**TL;DR — this new module is not type-checked, and it does not pass pyrefly today, so adding it to the list as-is turns CI red.**

Adjacent rather than broken by this PR: the file is missing from [`project-includes`](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/pyrefly.toml#L25), which already lists 11 files from `nemo_rl/environments/`. The repo's pinned pyrefly 0.24.2 reports three errors, one on this anchored line, where it cannot see through the [both-or-neither guard](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L176-L181) and still thinks `port_high` may be `None`.

Note [nemo_gym.py](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L1) is not in that list either, so this is an opportunity on a new module that imports nothing beyond the standard library and omegaconf, not a rule the PR breaks.

### AI-1
Not a suggestion block — this is a two-file change, the source plus [pyrefly.toml](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/pyrefly.toml#L25). All three errors are one-liners: narrow `port_high` explicitly before this comparison, and cast or re-key the merged dict at [line 378](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L378). Then add the module's path to `project-includes` so it stays checked.

<details><summary>The three pyrefly errors</summary>

pyrefly 0.24.2, the version pinned in [pyproject.toml](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/pyproject.toml#L280), run on the file:

```
:121  return dict(value)
      No matching overload found for function dict.__init__

:191  if port_low is not None and port_low >= port_high:
      `>=` is not supported between `int` and `None`

:378  return merged
      Returned type `dict[Enum | bool | bytes | float | int | str, Any]`
      is not assignable to declared return type `dict[str, Any]`
```

The third comes from `OmegaConf.to_container` widening the key type.
</details>

nemo_rl/environments/nemo_gym.py
```diff
@@ -1050,14 +1210,27 @@ def build_nemo_gym_config(
...
-        ``initial_global_config_dict``. The caller's ``env_configs`` is not mutated.
     """
     nemo_gym_dict = dict(env_configs["nemo_gym"])
 
+    # Validate the shards block even though only single-actor creation is wired
+    # up so far, so a malformed or premature sharded config fails at setup with
+    # a precise message instead of silently running unsharded.
+    shard_plan = parse_shard_plan(nemo_gym_dict)
+    if shard_plan is not None:
+        raise ShardConfigError(
```
**TL;DR — the obvious test for this raise would keep passing after #3371 while testing a different function, because the phrase it matches moves.**

PR-introduced. This raise and the [pop below it](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym.py#L1231-L1232) have no test — the diff does not touch [test_nemo_gym_utils.py](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/tests/unit/environments/test_nemo_gym_utils.py#L1), where `build_nemo_gym_config` is tested. Delete both and the suite stays green.

#3371 does not cover it either, and it sets a trap. Its `test_spinup_nemo_gym_actor_rejects_a_sharded_config` matches `"not wired up yet"` — but against a different raise inside `spinup_nemo_gym_actor` ending "rollout routing is not wired up yet". #3371 also rewrites *this* message, dropping that phrase. So `"not wired up yet"` moves between two raises in two functions, and a test here matching it would still pass afterwards while pinning the wrong one. Match the prefix `"shards defines"`, which survives the rewrite unchanged.

### AI-1
Not a suggestion block — the test file is not part of this diff. Add at the end of [test_nemo_gym_utils.py](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/tests/unit/environments/test_nemo_gym_utils.py#L1):

```python
def test_build_nemo_gym_config_rejects_a_sharded_config(detected_uv_dirs):
    """Multi-actor creation is not wired up yet, so 'shards' must fail loudly."""
    from nemo_rl.environments.nemo_gym_shards import ShardConfigError

    env_configs = _env_configs(
        shards=[
            {"name": "judged", "config_paths": ["judge.yaml"]},
            {"name": "tools", "config_paths": ["tools.yaml"]},
        ],
    )
    # 'shards' replaces the top-level config_paths; keeping both is its own error.
    del env_configs["nemo_gym"]["config_paths"]

    with pytest.raises(ShardConfigError, match="shards defines"):
        build_nemo_gym_config(env_configs, base_urls=[], model_name="test-model")
```

Once the [stray-key raise](https://github.com/NVIDIA-NeMo/RL/blob/b6bdd3303070669942e218e7596892c6fe64ef16/nemo_rl/environments/nemo_gym_shards.py#L253) lands, extend this with a case passing the sharding keys without `shards`.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,408 @@
...
+@dataclass(frozen=True)
+class ShardPlan:
+    """The parsed ``shards`` block plus the settings that apply to all shards."""
+
+    shards: list[ShardSpec]
+    common_inherited_overlays: frozenset[str] = frozenset()
+    common_overrides: dict[str, Any] = field(default_factory=dict)
+    allowed_duplicate_entries: frozenset[str] = frozenset()
+
+    @property
```
terrykong: **1 action item.**
**TL;DR — `ShardPlan.total_instances` has no production caller anywhere in the stack.**

PR-introduced, and a nit. Across #3371-#3374 there is no production use: #3371 and #3372 have no hits, #3373's only hit is unchanged context in a hunk adding `placement_strategy`, and #3374's is a test assertion. #3371 computes the same number itself as `len(_shard_instances(plan))`.

### AI-1
Not a suggestion block — the choice is wire-it-up or delete. Either have #3371's placement sizing call `plan.total_instances` so there is one definition of the count, or drop the property until something needs it. If it is deliberately staged for #3373's placement work, say so and this can be closed.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+            OmegaConf.create(shared_base),
+            OmegaConf.create(selected_inherited),
+            OmegaConf.create(plan.common_overrides),
+            OmegaConf.create(shard.overrides),
+        ),
+        resolve=True,
+    )
+    assert isinstance(merged, dict)
+    merged["config_paths"] = list(shard.config_paths)
+    if shard.port_range_low is not None:
```
terrykong, line 407: **1 action item.**
**TL;DR — the per-shard port range is validated, merged and written into the shard config, then overwritten before Gym ever sees it.**

`apply_shard_overlay` writes `port_range_low`/`port_range_high` into the merged dict here. That dict becomes `initial_global_config_dict`, and [`NemoGym._spinup` overwrites both keys unconditionally](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L438-L439):

```python
initial_global_config_dict["port_range_low"] = _gym_port_low
initial_global_config_dict["port_range_high"] = _gym_port_high
```

The crux is that `NemoGymConfig` has **two** separate places a port range can sit, both spelled `port_range_low`: a [top-level field](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L215), and a key nested inside [`initial_global_config_dict`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L211).

The recipe's value arrives in the **nested** one, because that dict is the raw `env.nemo_gym` block, so anything the user typed is already in it. `_spinup` reads the **top-level** field, which nothing ever populates — [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1113-L1157) does not set it, and neither does `_build_gym_actor_config` in #3371 — so it falls back to 5000-5999 and writes that over the nested value that was already right.

The nested copy is not a spare, which is why the fix below is a move and not an extra assignment: it is the dict handed to Gym, and Gym reads its port range straight out of it. That last line is the deliberate handoff — publish what NeMo RL decided into what Gym gets. It misfires only because the field it reads from is empty.

The global knob being dead is pre-existing; what this PR adds is a second, per-shard copy of it. No recipe in the tree sets a non-default range today, so nothing is visibly broken — the knob just does not work. Worth fixing while the per-shard copy is still new.

### AI-1
Not a suggestion block — the fix lands in a different file than this anchor. In [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1113-L1157), pop `port_range_low`/`port_range_high` off `nemo_gym_dict` and pass them onto the returned `NemoGymConfig`, so the one place `_spinup` reads is populated. The per-shard values then reach Gym too, because `apply_shard_overlay` already puts them in the shard's config.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+    Returning None is the backward-compatible path: a config with
+    ``config_paths`` and no ``shards`` behaves exactly as it did before
+    sharding existed.
+
+    Raises:
+        ShardConfigError: the shards block is malformed, or it is combined with
+            config that has no unambiguous shard to belong to.
+    """
+    if "shards" not in nemo_gym_config:
+        return None
```
terrykong, line 253: **1 action item.**
**TL;DR — `common_overrides`, `common_inherited_overlays` and `allowed_duplicate_entries` do nothing without `shards`, but a config carrying them without it is accepted and they are dropped rather than rejected.**

PR-introduced. `parse_shard_plan` returns `None` as soon as `shards` is absent, so the other three are never read, and [`build_nemo_gym_config` pops all four](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1081-L1082). That pop is reachable only in this case — the sharded path [raises first](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1070) — so the pop is what makes the drop silent.

What the drop costs depends on whether a top-level `config_paths` is still present, and it is bad either way:

- **`config_paths` present** — you commented out `shards:` to run unsharded for a moment, or you are part-way through converting a recipe and have not moved the paths into shards yet. The job runs, unsharded, with none of your `common_overrides` applied, and nothing in the log mentions it. You get a complete run at settings you did not ask for.
- **`config_paths` absent** — the normal shape of a sharded config, since [combining the two is rejected](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L265-L269). Gym then fails with *"No server instances are configured, so there is nothing to run"* ([global_config.py:413](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L403-L416)), which points at a missing `config_paths` and never mentions the sharding block that is the actual mistake.

So the raise earns its keep in both branches: it replaces a silent wrong run in the first, and an error that blames the wrong thing in the second.

To be straight about the timing: this cannot bite today, because [any `shards` block raises](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1070) until #3371 lands. The reason to add it now is that it is free — no file under `examples/` uses any of the four keys, so nothing can start failing — and `parse_shard_plan` is the function this PR is writing.

### AI-1
The check belongs here in `parse_shard_plan`, not in [`build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1069): #3371's shard path calls `parse_shard_plan` as well, so one raise covers both entry points for good.

```suggestion
    if "shards" not in nemo_gym_config:
        stray = sorted(
            key
            for key in SHARDING_CONFIG_KEYS
            if key != "shards" and key in nemo_gym_config
        )
        if stray:
            raise ShardConfigError(
                f"env.nemo_gym has sharding keys {stray} but no 'shards' block. "
                f"Add a 'shards' block, or remove these keys."
            )
        return None
```

**Context — no action.** After this lands, the [pop in `build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1081-L1082) looks unreachable from this PR alone, because both branches now raise. Please keep it — #3371's `_build_gym_actor_config` runs that same pop for every actor it builds.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+DEFAULT_REPLICAS = 1
+SHARD_NAME_PATTERN = re.compile(r"^[A-Za-z0-9][A-Za-z0-9_.-]*$")
+
+# Gym's log-directory key (``nemo_gym.global_config.NEMO_GYM_LOG_DIR_KEY_NAME``).
+# Restated rather than imported: this module is imported on the driver, and
+# nemo_gym is installed only in the actor's venv.
+GYM_LOG_DIR_KEY = "nemo_gym_log_dir"
+
+# Per-shard keys consumed by NeMo RL rather than forwarded to Gym.
+SHARD_SPEC_KEYS = frozenset(
```
terrykong: **1 action item.**
**TL;DR — both hand-written key sets in this module list exactly the fields of a dataclass a few lines below them, so every new field means remembering to edit two places.**

PR-introduced. [`SHARD_SPEC_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L60) lists the eight fields of [`ShardSpec`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L79), and [`SHARDING_CONFIG_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L38) lists the four fields of [`ShardPlan`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L96-L103). Both match today. What a miss costs differs by set:

- miss `SHARD_SPEC_KEYS` → the [unknown-key check](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L142) rejects the user's own new key and tells them to move it under `overrides`, which sends them the wrong way.
- miss `SHARDING_CONFIG_KEYS` → the key is never stripped, so a NeMo-RL-only setting rides into the Gym config handed to every child process.

The paired edit is already a real tax rather than a hypothetical one. #3373 adds `placement_strategy` to `ShardPlan` and correctly remembers to add it to `SHARDING_CONFIG_KEYS` in the same diff — that is the work this removes. And this PR's tip commit `b6bdd3303` "fix(nemo-gym): exclude integration settings from overlay inventory" is a fix for the case where the pairing *was* missed, on the sibling list [`NEMO_RL_DICT_CONFIG_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L49).

### AI-1
Read both sets off their dataclasses:

```python
from dataclasses import dataclass, field, fields, replace

SHARD_SPEC_KEYS = frozenset(f.name for f in fields(ShardSpec))
SHARDING_CONFIG_KEYS = frozenset(f.name for f in fields(ShardPlan))
```

Not a suggestion block: each assignment has to move below the dataclass it reads — `SHARD_SPEC_KEYS` past [line 79](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L79), `SHARDING_CONFIG_KEYS` past [line 103](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L103) — so it is two moves plus an import, not an edit to the anchored line. Nothing between the old and new positions uses either set at module level, and first use is [line 142](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L142). `fields()` returns only real fields, so the [`total_instances`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L105) property does not leak into the set; both derived sets come back identical to the literals they replace, on this head and after #3373's new field.

**Context — no action.** [`NEMO_RL_DICT_CONFIG_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L49) cannot be derived the same way — it means "the dict-shaped keys NeMo RL pops elsewhere", which has no single declaration to read ([`tokenizer_config`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1088) and [`effort_levels`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L414) are popped in two different functions). A test asserting it agrees with what those two strip is the closest equivalent.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+
+    raw_allowed = nemo_gym_config.get("allowed_duplicate_entries") or []
+    if OmegaConf.is_config(raw_allowed):
+        raw_allowed = OmegaConf.to_container(raw_allowed, resolve=True)
+    if not isinstance(raw_allowed, list) or not all(
+        isinstance(entry, str) for entry in raw_allowed
+    ):
+        raise ShardConfigError(
+            "env.nemo_gym.allowed_duplicate_entries must be a list of strings"
+        )
```
terrykong: **1 action item.**
**TL;DR — `allowed_duplicate_entries` hand-rolls its checks, so it accepts input every sibling list rejects.**

PR-introduced, and small. Every other string list goes through [`_as_string_set`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L124). This one does not, so `["", "dup", "dup"]` is accepted — empty string and duplicate both pass — while [`inherited_overlays`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L212-L214) rejects both. The trailing `or []` also swallows `{}` without complaint, even though `{"a": 1}` correctly raises.

### AI-1
```suggestion
    raw_allowed = _as_string_set(
        nemo_gym_config.get("allowed_duplicate_entries"),
        context="env.nemo_gym.allowed_duplicate_entries",
    )
```
The variable name is kept so [`allowed_duplicate_entries=frozenset(raw_allowed)`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L341) below still works. That `frozenset(...)` call is now redundant and can be dropped.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+        OmegaConf.merge(
+            OmegaConf.create(shared_base),
+            OmegaConf.create(selected_inherited),
+            OmegaConf.create(plan.common_overrides),
+            OmegaConf.create(shard.overrides),
+        ),
+        resolve=True,
+    )
+    assert isinstance(merged, dict)
+    merged["config_paths"] = list(shard.config_paths)
```
terrykong, line 406: **1 action item.**
**TL;DR — the parser enforces "each Gym entry is claimed by exactly one shard", then the two override layers are merged on top without being checked against it.**

PR-introduced, low severity. [`apply_shard_overlay` merges `common_overrides` and the shard's `overrides` with no claim check](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L361-L373). Two things slip through: naming a whole entry in `common_overrides` puts a startable server on shards that never claimed it, which #3371 catches only after every actor is up; and a `config_paths` written into either layer is [silently overwritten](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L375) by the shard's own list, while `port_range_low`/`port_range_high` in that same layer [do take effect](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L376-L378).

Neither is likely, and a partial override such as `code_gen: {num_workers: 64}` is inert — Gym ignores a top-level key that is not shaped like a server. Raising is still worth it: the parser already holds everything both checks need.

### AI-1
In [`parse_shard_plan`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L287-L323), beside the claim check, reject an override-layer key that is in the entry inventory but claimed by a **different** shard, and reject `config_paths` in either layer. Not a suggestion block: the fix belongs in `parse_shard_plan`, while this anchor is only where the discard shows up.

Worked example, if useful: [how an override puts a Gym server on a shard that never asked for it](https://terrykong.github.io/gh-pages-poc/terryk/pr-3369-override-layers.html).

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+    )
+
+
+def find_gym_config_entries(nemo_gym_config: Mapping[str, Any]) -> list[str]:
+    """Return top-level keys that Gym would read as server instance configs.
+
+    Mirrors Gym's own rule (``filter_for_server_instance_configs``): any
+    dict-shaped top-level key that is not a known setting is an entry overlay.
+    Under sharding these values are inherited from the resolved parent config.
+    A shard claims them by key so their YAML bodies do not need to be copied.
```
terrykong: **1 action item.**
**TL;DR — the docstring says this mirrors Gym's rule; it is deliberately broader, and should say so.**

PR-introduced, docstring only. Gym's [`filter_for_server_instance_configs`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L386-L401) does two things: drop the 31 [reserved top-level keys](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L101-L133), then require the value to pass pydantic validation as a `ServerInstanceConfig` via [`maybe_get_server_instance_config`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L689-L704). This function excludes six keys and accepts any Mapping.

**Do not change the code.** The module deliberately cannot import Gym — it is installed only in the actor's venv, per the [comment above](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L54-L56) — so the pydantic step is not available here, and being broader is the safe direction. The one dict-shaped reserved key, `head_server`, is [overwritten unconditionally by `NemoGym._spinup`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L461-L464) anyway, and no in-tree recipe sets it. Only the wording needs fixing.

### AI-1
```suggestion
    Deliberately broader than Gym's own rule
    (``filter_for_server_instance_configs``), which also requires the value to
    validate as a ``ServerInstanceConfig``: this module cannot import Gym, so
    any dict-shaped top-level key that is not a known setting is treated as an
    entry overlay. Under sharding these values are inherited from the resolved
    parent config. A shard claims them by key so their YAML bodies do not need
    to be copied.
```

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+from pathlib import PurePosixPath
+from typing import Any, Mapping
+
+from omegaconf import OmegaConf
+
+# Keys under ``env.nemo_gym`` that NeMo RL consumes itself. They must never
+# reach NeMo Gym: the merged config is serialized into NEMO_GYM_CONFIG_DICT for
+# every one of the ~60 child processes, and Gym treats unrecognized dict-shaped
+# top-level keys as server instance configs, which can trip its
+# error_on_almost_servers path.
```
terrykong: **1 action item.**
**TL;DR — this comment gives a reason that cannot happen; the real reason is the one just before it.**

PR-introduced. Stripping these keys is right, but not because of `error_on_almost_servers`. [`is_almost_server`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L708-L728) returns False unless the dict has one of `responses_api_models` / `resources_servers` / `responses_api_agents` **and** an inner config carrying an `entrypoint`. `shards` is a list, so [`maybe_get_server_instance_config`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L689-L704) bails at its `isinstance(..., DictConfig)` guard and `is_almost_server` is False as well. Gym just ignores the key; the [raise](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L745-L754) never fires.

The real reason — these keys get serialized into `NEMO_GYM_CONFIG_DICT` for every child process — already stands on its own.

### AI-1
```suggestion
# Keys under ``env.nemo_gym`` that NeMo RL consumes itself. They must never
# reach NeMo Gym: the merged config is serialized into NEMO_GYM_CONFIG_DICT for
# every one of the ~60 child processes, so anything left here is shipped to all
# of them.
```

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+        or isinstance(port_low, bool)
+        or not isinstance(port_high, int)
+        or isinstance(port_high, bool)
+        or port_low < 1
+        or port_high > 65_535
+    ):
+        raise ShardConfigError(
+            f"shard '{name}' port range must use integers within [1, 65535]"
+        )
+    if port_low is not None and port_low >= port_high:
```
terrykong: **1 action item.**
**TL;DR — this new module is not type-checked, and it does not pass pyrefly today, so adding it to the list as-is turns CI red.**

Adjacent rather than broken by this PR: the file is missing from [`project-includes`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/pyrefly.toml#L25), which already lists 11 files from `nemo_rl/environments/`. The repo's pinned pyrefly 0.24.2 reports three errors, one on this anchored line, where it cannot see through the [both-or-neither guard](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L176-L181) and still thinks `port_high` may be `None`.

Note [nemo_gym.py](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1) is not in that list either, so this is an opportunity on a new module that imports nothing beyond the standard library and omegaconf, not a rule the PR breaks.

### AI-1
Not a suggestion block — this is a two-file change, the source plus [pyrefly.toml](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/pyrefly.toml#L25). All three errors are one-liners: narrow `port_high` explicitly before this comparison, and cast or re-key the merged dict at [line 378](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L379). Then add the module's path to `project-includes` so it stays checked.

<details><summary>The three pyrefly errors</summary>

pyrefly 0.24.2, the version pinned in [pyproject.toml](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/pyproject.toml#L277), run on the file:

```
:121  return dict(value)
      No matching overload found for function dict.__init__

:191  if port_low is not None and port_low >= port_high:
      `>=` is not supported between `int` and `None`

:378  return merged
      Returned type `dict[Enum | bool | bytes | float | int | str, Any]`
      is not assignable to declared return type `dict[str, Any]`
```

The third comes from `OmegaConf.to_container` widening the key type.
</details>

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -0,0 +1,409 @@
...
+@dataclass(frozen=True)
+class ShardPlan:
+    """The parsed ``shards`` block plus the settings that apply to all shards."""
+
+    shards: list[ShardSpec]
+    common_inherited_overlays: frozenset[str] = frozenset()
+    common_overrides: dict[str, Any] = field(default_factory=dict)
+    allowed_duplicate_entries: frozenset[str] = frozenset()
+
+    @property
```
terrykong: **1 action item.**
**TL;DR — `ShardPlan.total_instances` has no production caller anywhere in the stack.**

PR-introduced, and a nit. Across #3371-#3374 there is no production use: #3371 and #3372 have no hits, #3373's only hit is unchanged context in a hunk adding `placement_strategy`, and #3374's is a test assertion. #3371 computes the same number itself as `len(_shard_instances(plan))`.

### AI-1
Not a suggestion block — the choice is wire-it-up or delete. Either have #3371's placement sizing call `plan.total_instances` so there is one definition of the count, or drop the property until something needs it. If it is deliberately staged for #3373's placement work, say so and this can be closed.

terrykong (commented):
This PR adds [nemo_gym_shards.py](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L1): a schema and parser for splitting one NeMo-Gym stack across several actors, plus a [raise in `build_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym.py#L1070) that rejects a `shards:` block until multi-actor creation exists.

The module is the right shape and worth keeping that way. It is pure functions plus frozen dataclasses with no `nemo_rl` imports, so all 40 of its tests run in 0.26s with no Ray, no GPU and no Gym. The error messages are unusually good — they name the resolved entry inventory, so a user who mistypes an overlay key can see what was actually available. And the schema holds up against its real next consumer: #3371 uses it without reaching around it once.

Worth calling out, because it answers "is this worth landing on its own": on `main` today a `shards:` block under `env.nemo_gym` is a list, so Gym's [`maybe_get_server_instance_config`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L689-L704) returns None for it and `is_almost_server` is False — Gym ignores it and the job runs unsharded with no message at all. The raise added here turns that silent misconfiguration into a precise error. The gated-off schema is a user-visible fix today, not only scaffolding.

The one finding worth fixing before merge is the port range, on [nemo_gym_shards.py:375](https://github.com/NVIDIA-NeMo/RL/blob/1a56c04939d5889a4cf51413a2b1711b03af72b5/nemo_rl/environments/nemo_gym_shards.py#L376): the per-shard values are validated, merged and written, then overwritten by `NemoGym._spinup`. No recipe sets a non-default range today, so nothing is visibly broken — the knob just does not work, and this PR adds a second, per-shard copy of it. The rest are smaller: a silent-drop path when the sharding keys appear without `shards:`, two hand-copied key lists that each mirror a dataclass, an unvalidated override layer, two comment claims that are not accurate, and type-checking coverage.

Two asks that belong on other PRs, noted so they are not lost:

1. On #3371, `DEFAULT_SHARD_CPUS = 8` is a magic constant in the consumer while its sibling `replicas` defaults on the schema. Move the default onto `ShardSpec.actor_cpus` so both per-shard defaults live in one place. Its `test_spinup_nemo_gym_actor_rejects_a_sharded_config` also matches on "not wired up yet", a phrase this PR's `build_nemo_gym_config` raise carries today and #3371 moves onto a different raise — worth tightening to something unique to `spinup_nemo_gym_actor`, such as "rollout routing".
2. This stack adds 12 new user-facing YAML keys (4 top-level, 8 per-shard) with no docs and no exemplar YAML. That should land with #3371 — the PR that makes them usable — in `docs/design-docs/nemo-gym-integration.md`.

CI is green on this head (80 passing, 14 skipped). The PR is blocked only on a missing approval from one of the three requested reviewer teams, not on CI.

Reviewed at head `1a56c0493`, re-anchored after the rebase. The `65_536` port bound and the two error messages that disagreed about inclusive-vs-exclusive were both fixed in the meantime, so that ask is gone. Findings were verified by running the module directly, since the repo's `uv run` path does not work on macOS. ruff's three pinned hooks are clean on all three files, and pyrefly 0.24.2 was run by hand.

_Generated by Claude Code_

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. `build_nemo_gym_config` now removes `port_range_low` and `port_range_high` from the nested Gym config and passes them as top-level `NemoGymConfig` fields, which is where `_spinup` reads them. `test_build_nemo_gym_config_moves_port_range_to_actor_fields` covers the handoff.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. `parse_shard_plan` now rejects any sharding-only keys when there is no `shards` block, and `test_unsharded_config_rejects_stray_sharding_keys` covers each key.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. `SHARD_SPEC_KEYS` and `SHARDING_CONFIG_KEYS` are now derived with `dataclasses.fields` immediately after their corresponding dataclasses.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. `allowed_duplicate_entries` now uses `_as_string_set`, so empty values and duplicates are rejected consistently with the other string-list fields. The parameterized validation test covers both cases.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. The parser rejects `config_paths` in both override layers, rejects common overrides of entries not claimed as common, and rejects shard overrides of entries owned elsewhere. Focused tests cover these ownership and precedence cases.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. The docstring now explicitly says this inventory is deliberately broader than Gym’s validation rule and explains why this driver-side module cannot perform `ServerInstanceConfig` validation.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. The comment now states only the actual reason: these NeMo-RL-only keys would otherwise be serialized and shipped to every Gym child process.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. The conversion helpers now narrow mapping keys and port bounds explicitly, and `nemo_rl/environments/nemo_gym_shards.py` is included in `pyrefly.toml`. Pyrefly and Ruff pass on the rebased branch.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Fixed in the current head. The unused `ShardPlan.total_instances` property and its test were removed.

terrykong (dismissed):
All nine comments are addressed. I checked each one against the current head (`56dbaad`) rather than going by the replies:

- The per-shard port range now reaches the actor — `build_nemo_gym_config` pops the range off the Gym dict and passes it as a `NemoGymConfig` field, so `_spinup` reads a populated field instead of falling back to 5000-5999.
- `parse_shard_plan` raises when sharding keys appear with no `shards` block, naming the stray keys.
- `SHARD_SPEC_KEYS` and `SHARDING_CONFIG_KEYS` both derive from their dataclasses, so neither can drift again.
- `allowed_duplicate_entries` goes through `_as_string_set` like its sibling lists.
- Both override layers reject `config_paths` and reject entries another shard claims.
- The `find_gym_config_entries` docstring and the stripped-keys comment now describe what the code actually does.
- The module is in `pyrefly.toml` and passes — pyrefly 0.24.2 reports 0 errors on it.
- `total_instances` is gone.

I also ran it: 50 unit tests pass, up from 40. Since the new guards exist because I asked for them, I checked they do not reject anything legitimate — tuning a commonly-inherited entry through `common_overrides`, a shard tuning the entry it claims, defining a brand-new shared entry, a plain setting, the single-shard auto-claim, and the unsharded path are all still accepted. The four mistake shapes are rejected with messages that name the fix.

Two notes, neither blocking. CI was triggered at this head a few minutes ago and has not finished, so the approval is on the code, not on a green run. And this is stacked on #3633, so it cannot reach `main` until the PRs below it land.

_Generated by Claude Code_

terrykong (approved):

PR #3371, opened 2026-07-27
feat(nemo-gym): create actors for configured shards

ananthsub:
## Summary

Adds the factory that creates one `NemoGym` actor for every configured shard instance.

For a sharded configuration, `build_nemo_gym_actors()`:

- Expands each shard's replica count into physical actor instances.
- Materializes the actor runtime environment before reserving shard CPUs. The runtime-environment builder may temporarily reserve one CPU on every node while it creates virtual environments.
- Creates one placement-group bundle per instance and uses `STRICT_SPREAD` by default.
- Composes each actor's Gym configuration and assigns a separate log directory.
- Starts all actors under one shared deadline, then installs the tokenizer on each actor.
- Reads resolved entries through `NemoGym.list_entries()` and builds the agent-to-shard map.
- Rejects ambiguous duplicate agents and entries.
- Waits for in-progress startup calls to leave `RunHelper.start()` before cleanup. It then shuts down every actor and removes the placement group without killing an actor while it still owns Gym subprocesses.

The unsharded path uses the same cleanup guarantee if startup or tokenizer installation fails. Errors identify the shard that failed instead of exposing only the underlying Ray or Gym exception.

This PR provides actor creation and discovery. Rollout routing is added in the next layer.

## Dependency

This PR targets `main`. [PR #3372](https://github.com/NVIDIA-NeMo/RL/pull/3372) builds on this branch.

### DIFF
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -17,12 +17,17 @@ import subprocess
 import sys
 from collections import Counter
 from collections.abc import AsyncGenerator
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Dict, List, NotRequired, Optional, TypedDict
 
 import ray
 import torch
-from ray.util.scheduling_strategies import NodeAffinitySchedulingStrategy
+from ray.util.placement_group import placement_group, remove_placement_group
+from ray.util.scheduling_strategies import (
+    NodeAffinitySchedulingStrategy,
+    PlacementGroupSchedulingStrategy,
+)
 from transformers import PreTrainedTokenizerBase
 
 from nemo_rl.distributed.virtual_cluster import (
@@ -35,8 +40,15 @@ from nemo_rl.environments.interfaces import EnvironmentInterface
 from nemo_rl.environments.nemo_gym_shards import (
     SHARDING_CONFIG_KEYS,
     ShardConfigError,
+    ShardPlan,
+    ShardSetupError,
+    ShardSpec,
+    apply_shard_log_dir,
+    apply_shard_overlay,
+    build_agent_shard_map,
     parse_shard_plan,
 )
+from nemo_rl.environments.utils import shutdown_environments
 from nemo_rl.utils.timer import Timer
 from nemo_rl.utils.venvs import make_actor_runtime_env
 
@@ -50,6 +62,26 @@ GYM_SERVER_TYPE_KEYS = (
     "responses_api_models",
     "resources_servers",
 )
+
+# Shard name used when the job is unsharded, so a single actor and a sharded
+# set have the same shape and callers need only one code path.
+DEFAULT_SHARD_NAME = "nemo_gym"
+
+# Logical CPUs reserved per shard bundle. This is a scheduling reservation, not
+# a limit: it decides whether a node can host a shard and steers Ray away from
+# stacking other CPU work there. Gym's subprocesses are not metered against it,
+# so a shard can use more than it reserves. Override per shard for known-heavy
+# stacks (e.g. code_gen and its sandbox pool).
+DEFAULT_SHARD_CPUS = 8
+
+# Waiting for STRICT_SPREAD bundles. Failing here means the allocation has
+# fewer usable nodes than the plan needs, which is worth reporting quickly.
+DEFAULT_SHARD_PG_READY_TIMEOUT_SECONDS = 180.0
+
+# Waiting for a shard to finish _spinup. Generous because a node that has never
+# run Gym builds every server's venv first; with venvs baked into the image
+# this is far shorter.
+DEFAULT_SHARD_SPINUP_TIMEOUT_SECONDS = 1800.0
 
 # Kept local (not imported from models.generation) so the gym actor stays free of
 # generation-module imports. Must cover every name resolve_routed_experts_dtype
@@ -679,7 +711,7 @@ def build_nemo_gym_config(
     enable_router_replay: bool = False,
     use_fastokens: bool = False,
 ) -> NemoGymConfig:
-    """Build the ``NemoGymConfig`` for a NeMo-Gym actor.
+    """Build the ``NemoGymConfig`` for a single, unsharded NeMo-Gym actor.
 
     Splits ``env_configs["nemo_gym"]`` into the NeMo-RL-side fields the actor
     reads directly and the remainder, which is forwarded verbatim as NeMo-Gym's
@@ -696,20 +728,44 @@ def build_nemo_gym_config(
             routed-experts carry dtype for the model.
         use_fastokens: Forwarded from ``policy.tokenizer.use_fastokens`` so the
             actor patches its tokenizer the same way the driver does.
+
+    Raises:
+        ShardConfigError: The config is sharded. One config cannot describe
+            several shards; use :func:`build_nemo_gym_actors`.
     """
     nemo_gym_dict = dict(env_configs["nemo_gym"])
 
-    # Validate the shards block even though only single-actor creation is wired
-    # up so far, so a malformed or premature sharded config fails at setup with
-    # a precise message instead of silently running unsharded.
     shard_plan = parse_shard_plan(nemo_gym_dict)
     if shard_plan is not None:
         raise ShardConfigError(
             f"env.nemo_gym.shards defines {len(shard_plan.shards)} shards "
-            f"({', '.join(s.name for s in shard_plan.shards)}), but multi-actor "
-            f"creation is not wired up yet. Remove 'shards' and use "
-            f"'config_paths' to run this job on a single actor."
+            f"({', '.join(s.name for s in shard_plan.shards)}), so it does not "
+            f"describe a single actor. Use build_nemo_gym_actors() instead."
         )
+
+    return _build_gym_actor_config(
+        nemo_gym_dict,
+        base_urls=base_urls,
+        model_name=model_name,
+        enable_router_replay=enable_router_replay,
+        use_fastokens=use_fastokens,
+    )
+
+
+def _build_gym_actor_config(
+    nemo_gym_dict: dict[str, Any],
+    *,
+    base_urls: list[Optional[str]],
+    model_name: str,
+    enable_router_replay: bool,
+    use_fastokens: bool,
+) -> NemoGymConfig:
+    """Turn one already-resolved Gym config mapping into a ``NemoGymConfig``.
+
+    Shared by the unsharded path and by each shard, so every actor gets the
+    same treatment of NeMo-RL-side keys regardless of how it was composed.
+    """
+    nemo_gym_dict = dict(nemo_gym_dict)
 
     # NeMo-RL-only keys are consumed here and must never reach Gym: the merged
     # config is serialized into every Gym child process, and unrecognized
@@ -752,6 +808,289 @@ def build_nemo_gym_config(
     )
 
 
+@dataclass
+class NemoGymShardSet:
+    """The live actors behind one NeMo-Gym stack, sharded or not.
+
+    An unsharded job is the one-shard, one-replica case, so callers do not need
+    a separate code path for it.
+
+    Attributes:
+        handles: Shard name to its replica handles, in replica order.
+        agent_to_shard: Agent entry name to the shard hosting it. Empty when
+            unsharded, where every row goes to the only actor anyway.
+        placement_group: The STRICT_SPREAD group pinning shards to distinct
+            nodes, or None when unsharded.
+    """
+
+    handles: Dict[str, List[Any]]
+    agent_to_shard: Dict[str, str] = field(default_factory=dict)
+    placement_group: Any = None
+
+    @property
+    def is_sharded(self) -> bool:
+        return self.placement_group is not None
+
+    @property
+    def all_handles(self) -> List[Any]:
+        return [handle for replicas in self.handles.values() for handle in replicas]
+
+    def sole_handle(self) -> Any:
+        """The only actor, for callers that predate routing.
+
+        Raises:
+            ShardSetupError: There is more than one actor, so picking one would
+                silently drop the rest.
+        """
+        handles = self.all_handles
+        if len(handles) != 1:
+            raise ShardSetupError(
+                f"Expected a single NeMo-Gym actor but this set has "
+                f"{len(handles)} across shards {sorted(self.handles)}. The "
+                f"caller needs the shard-aware rollout router."
+            )
+        return handles[0]
+
+    def shutdown(self) -> None:
+        """Stop every actor, then release the bundles they were pinned to."""
+        shutdown_environments(
+            {
+                f"nemo_gym[{shard}][{replica}]": handle
+                for shard, replicas in self.handles.items()
+                for replica, handle in enumerate(replicas)
+            }
+        )
+        if self.placement_group is not None:
+            try:
+                remove_placement_group(self.placement_group)
+            except Exception as error:
+                print(f"Failed to release the NeMo-Gym placement group: {error}")
+            self.placement_group = None
+
+
+def _shard_instances(plan: ShardPlan) -> List[tuple[ShardSpec, int]]:
+    """Expand shards into one (shard, replica_index) entry per actor."""
+    return [
+        (shard, replica) for shard in plan.shards for replica in range(shard.replicas)
+    ]
+
+
+def build_nemo_gym_actors(
+    env_configs: dict[str, Any],
+    *,
+    base_urls: list[Optional[str]],
+    model_name: str,
+    enable_router_replay: bool = False,
+    use_fastokens: bool = False,
+    pg_ready_timeout: float = DEFAULT_SHARD_PG_READY_TIMEOUT_SECONDS,
+    spinup_timeout: float = DEFAULT_SHARD_SPINUP_TIMEOUT_SECONDS,
+) -> NemoGymShardSet:
+    """Create and spin up every NeMo-Gym actor this job needs.
+
+    Without ``shards`` this makes exactly one actor, scheduled as before. With
+    ``shards`` it makes one actor per replica, each pinned by a STRICT_SPREAD
+    placement group to a distinct node, each holding its own complete Gym
+    stack. Actors are spun up concurrently, so the wall-clock cost is roughly
+    the slowest shard rather than the sum.
+
+    Returns:
+        A :class:`NemoGymShardSet` whose actors are all running and validated.
+
+    Raises:
+        ShardSetupError: The bundles could not be placed, a shard failed to
+            start, or the shards' entries did not pass the startup checks. Any
+            actors already created are torn down first.
+    """
+    nemo_gym_dict = dict(env_configs["nemo_gym"])
+    plan = parse_shard_plan(nemo_gym_dict)
+
+    if plan is None:
+        return _build_single_gym_actor(
+            nemo_gym_dict,
+            base_urls=base_urls,
+            model_name=model_name,
+            enable_router_replay=enable_router_replay,
+            use_fastokens=use_fastokens,
+        )
+
+    return _build_sharded_gym_actors(
+        nemo_gym_dict,
+        plan,
+        base_urls=base_urls,
+        model_name=model_name,
+        enable_router_replay=enable_router_replay,
+        use_fastokens=use_fastokens,
+        pg_ready_timeout=pg_ready_timeout,
+        spinup_timeout=spinup_timeout,
+    )
+
+
+def _build_single_gym_actor(
+    nemo_gym_dict: dict[str, Any],
+    *,
+    base_urls: list[Optional[str]],
+    model_name: str,
+    enable_router_replay: bool,
+    use_fastokens: bool,
+) -> NemoGymShardSet:
+    """The pre-sharding path: one actor, no placement group, no discovery.
+
+    Discovery is skipped rather than merely unused. Its checks compare entry
+    names *between* shards, so with one shard there is nothing they could find.
+    """
+    actor_config = _build_gym_actor_config(
+        nemo_gym_dict,
+        base_urls=base_urls,
+        model_name=model_name,
+        enable_router_replay=enable_router_replay,
+        use_fastokens=use_fastokens,
+    )
+
+    actor_options: dict[str, Any] = {
+        "runtime_env": make_actor_runtime_env(NEMO_GYM_ACTOR_FQN)
+    }
+    if nemo_gym_dict.get("num_gpu_nodes", 0):
+        actor_options["scheduling_strategy"] = NodeAffinitySchedulingStrategy(
+            node_id=ray.get_runtime_context().get_node_id(),
+            soft=True,
+        )
+
+    actor = NemoGym.options(**actor_options).remote(actor_config)
+    ray.get(actor._spinup.remote())
+    return NemoGymShardSet(handles={DEFAULT_SHARD_NAME: [actor]})
+
+
+def _build_sharded_gym_actors(
+    nemo_gym_dict: dict[str, Any],
+    plan: ShardPlan,
+    *,
+    base_urls: list[Optional[str]],
+    model_name: str,
+    enable_router_replay: bool,
+    use_fastokens: bool,
+    pg_ready_timeout: float,
+    spinup_timeout: float,
+) -> NemoGymShardSet:
+    instances = _shard_instances(plan)
+
+    # num_gpu_nodes normally pins the single actor to the driver node. Under
+    # sharding that is the opposite of what we want, so the placement group
+    # wins; num_gpu_nodes keeps its other job of sizing the allocation.
+    if nemo_gym_dict.get("num_gpu_nodes", 0):
+        print(
+            "env.nemo_gym.shards is set, so the num_gpu_nodes affinity hint is "
+            "superseded by STRICT_SPREAD placement across "
+            f"{len(instances)} nodes."
+        )
+
+    base_gym_dict = {
+        key: value
+        for key, value in nemo_gym_dict.items()
+        if key not in SHARDING_CONFIG_KEYS
+    }
+    # One merge per shard; replicas are identical stamps of it apart from the
+    # log directory, so they must not re-run the merge.
+    merged_by_shard = {
+        shard.name: apply_shard_overlay(base_gym_dict, plan, shard)
+        for shard in plan.shards
+    }
+
+    pg = placement_group(
+        bundles=[
+            {"CPU": float(shard.actor_cpus or DEFAULT_SHARD_CPUS)}
+            for shard, _ in instances
+        ],
+        strategy="STRICT_SPREAD",
+    )
+    try:
+        ray.get(pg.ready(), timeout=pg_ready_timeout)
+    except BaseException as error:
+        remove_placement_group(pg)
+        raise ShardSetupError(
+            f"Could not place {len(instances)} NeMo-Gym shard instances on "
+            f"distinct nodes within {pg_ready_timeout}s. STRICT_SPREAD needs "
+            f"one node per instance with the requested CPUs free; the "
+            f"allocation may be too small or its nodes too busy."
+        ) from error
+
+    shard_set = NemoGymShardSet(handles={}, placement_group=pg)
+    try:
+        for bundle_index, (shard, replica) in enumerate(instances):
+            instance_gym_dict = apply_shard_log_dir(
+                merged_by_shard[shard.name],
+                shard.name,
+                replica_index=replica if shard.replicas > 1 else None,
+            )
+            actor = NemoGym.options(
+                runtime_env=make_actor_runtime_env(NEMO_GYM_ACTOR_FQN),
+                scheduling_strategy=PlacementGroupSchedulingStrategy(
+                    placement_group=pg,
+                    placement_group_bundle_index=bundle_index,
+                ),
+            ).remote(
+                _build_gym_actor_config(
+                    instance_gym_dict,
+                    base_urls=base_urls,
+                    model_name=model_name,
+                    enable_router_replay=enable_router_replay,
+                    use_fastokens=use_fastokens,
+                )
+            )
+            shard_set.handles.setdefault(shard.name, []).append(actor)
+
+        _spinup_shards_concurrently(shard_set, spinup_timeout)
+        shard_set.agent_to_shard = _discover_agent_shard_map(shard_set, plan)
+    except BaseException:
+        # A ray.get timeout does not cancel the actor-side work, so a
+        # half-started stack would keep running with nothing left to stop it.
+        shard_set.shutdown()
+        raise
+
+    print(f"NeMo-Gym shard map (agent -> shard): {shard_set.agent_to_shard}")
+    return shard_set
+
+
+def _spinup_shards_concurrently(
+    shard_set: NemoGymShardSet, spinup_timeout: float
+) -> None:
+    """Start every shard at once, naming the shard behind any failure.
+
+    Config faults surface inside ``_spinup`` -- a server ref pointing at an
+    entry that is missing from *this* shard's slice raises Gym's
+    ``ServerRefNotFoundError`` there. Gym names the entry and field but has no
+    concept of a shard, so the shard name is added here.
+    """
+    pending = {
+        (shard_name, replica): handle._spinup.remote()
+        for shard_name, replicas in shard_set.handles.items()
+        for replica, handle in enumerate(replicas)
+    }
+    for (shard_name, replica), reference in pending.items():
+        try:
+            ray.get(reference, timeout=spinup_timeout)
+        except BaseException as error:
+            raise ShardSetupError(
+                f"NeMo-Gym shard '{shard_name}' (replica {replica}) failed to "
+                f"start. A reference to an entry that is not in this shard's "
+                f"config_paths is the usual cause: {error}"
+            ) from error
+
+
+def _discover_agent_shard_map(
+    shard_set: NemoGymShardSet, plan: ShardPlan
+) -> Dict[str, str]:
+    """Ask one replica per shard what it spawned, then validate across shards.
+
+    Replicas of a shard are stamped from one merge, so they host identical
+    entries and only the first needs to be asked.
+    """
+    entries_by_shard = {
+        shard.name: ray.get(shard_set.handles[shard.name][0].list_entries.remote())
+        for shard in plan.shards
+    }
+    return build_agent_shard_map(entries_by_shard, plan.allowed_duplicate_entries)
+
+
 def spinup_nemo_gym_actor(
     env_configs: dict[str, Any],
     *,
@@ -760,7 +1099,7 @@ def spinup_nemo_gym_actor(
     enable_router_replay: bool = False,
     use_fastokens: bool = False,
 ) -> Any:
-    """Spin up the NeMo-Gym actor against the given generation server URLs.
+    """Spin up a single NeMo-Gym actor against the given generation server URLs.
 
     When ``env_configs["nemo_gym"]["num_gpu_nodes"] > 0``, the actor is
     scheduled with soft NodeAffinity to the caller's Ray node so its colocated
@@ -768,24 +1107,24 @@ def spinup_nemo_gym_actor(
 
     Returns:
         The spun-up ``NemoGym`` Ray actor handle (``_spinup`` already awaited).
+
+    Raises:
+        ShardConfigError: The config is sharded. Callers that dispatch to a
+            single handle cannot serve several shards; they need the router.
     """
-    nemo_gym_cfg = build_nemo_gym_config(
+    plan = parse_shard_plan(dict(env_configs["nemo_gym"]))
+    if plan is not None:
+        raise ShardConfigError(
+            f"env.nemo_gym.shards defines {len(plan.shards)} shards "
+            f"({', '.join(shard.name for shard in plan.shards)}), but this "
+            f"entrypoint dispatches rollouts to one actor handle. Shard-aware "
+            f"rollout routing is not wired up yet."
+        )
+
+    return build_nemo_gym_actors(
         env_configs,
         base_urls=base_urls,
         model_name=model_name,
         enable_router_replay=enable_router_replay,
         use_fastokens=use_fastokens,
-    )
-
-    nemo_gym_opts: dict[str, Any] = {
-        "runtime_env": make_actor_runtime_env(NEMO_GYM_ACTOR_FQN)
-    }
-    if env_configs["nemo_gym"].get("num_gpu_nodes", 0):
-        nemo_gym_opts["scheduling_strategy"] = NodeAffinitySchedulingStrategy(
-            node_id=ray.get_runtime_context().get_node_id(),
-            soft=True,
-        )
-
-    actor = NemoGym.options(**nemo_gym_opts).remote(nemo_gym_cfg)
-    ray.get(actor._spinup.remote())
-    return actor
+    ).sole_handle()
--- a/nemo_rl/environments/nemo_gym_shards.py
+++ b/nemo_rl/environments/nemo_gym_shards.py
@@ -60,6 +60,10 @@ class ShardConfigError(ValueError):
     """Raised for a malformed ``env.nemo_gym.shards`` block."""
 
 
+class ShardSetupError(RuntimeError):
+    """Raised when a sharded stack fails to start or fails its startup checks."""
+
+
 @dataclass(frozen=True)
 class ShardSpec:
     """One shard: a slice of the Gym config that gets its own actor.
@@ -264,6 +268,56 @@ def apply_shard_overlay(
         merged["port_range_low"] = shard.port_range_low
         merged["port_range_high"] = shard.port_range_high
     return merged
+
+
+def build_agent_shard_map(
+    entries_by_shard: Mapping[str, Mapping[str, list[str]]],
+    allowed_duplicate_entries: frozenset[str] | set[str] = frozenset(),
+) -> dict[str, str]:
+    """Map each agent entry to its shard, rejecting duplicates across shards.
+
+    Takes what each shard reported from ``NemoGym.list_entries()`` and returns
+    ``{agent_entry_name: shard_name}``, the lookup the router dispatches on.
+
+    Two failures are caught here rather than at first dispatch. An agent hosted
+    by two shards is always an error: rows naming it could go to either, so
+    routing would be silently nondeterministic. Any other entry in two shards
+    has to be allowlisted, because duplication is usually accidental — a shared
+    YAML dropped into two shards' path lists quietly brings its judge along and
+    doubles that judge's GPU claim.
+
+    Only names are compared. What an entry means is Gym's business.
+    """
+    agent_to_shard: dict[str, str] = {}
+    for shard_name, entries in entries_by_shard.items():
+        for entry, types in entries.items():
+            if "responses_api_agents" not in types:
+                continue
+            if entry in agent_to_shard:
+                raise ShardSetupError(
+                    f"Agent '{entry}' is hosted by both shard "
+                    f"'{agent_to_shard[entry]}' and shard '{shard_name}'. An "
+                    f"agent must live in exactly one shard so rows naming it "
+                    f"have one destination."
+                )
+            agent_to_shard[entry] = shard_name
+
+    hosting_shard: dict[str, str] = {}
+    for shard_name, entries in entries_by_shard.items():
+        for entry in entries:
+            if entry in agent_to_shard:
+                continue
+            if entry in hosting_shard and entry not in allowed_duplicate_entries:
+                raise ShardSetupError(
+                    f"Config entry '{entry}' appears in shard "
+                    f"'{hosting_shard[entry]}' and shard '{shard_name}' but is "
+                    f"not listed in allowed_duplicate_entries. If this entry "
+                    f"starts a model engine, duplicating it doubles its GPU "
+                    f"claim; if the duplication is intended, allowlist it."
+                )
+            hosting_shard.setdefault(entry, shard_name)
+
+    return agent_to_shard
 
 
 def apply_shard_log_dir(
--- a/tests/unit/environments/test_nemo_gym_shards.py
+++ b/tests/unit/environments/test_nemo_gym_shards.py
@@ -20,9 +20,11 @@ from nemo_rl.environments.nemo_gym_shards import (
     GYM_LOG_DIR_KEY,
     ShardConfigError,
     ShardPlan,
+    ShardSetupError,
     ShardSpec,
     apply_shard_log_dir,
     apply_shard_overlay,
+    build_agent_shard_map,
     find_gym_config_entries,
     parse_shard_plan,
 )
@@ -192,6 +194,75 @@ def test_apply_shard_overlay_keeps_global_ports_by_default():
     )
 
     assert (merged["port_range_low"], merged["port_range_high"]) == (5000, 5999)
+
+
+def test_build_agent_shard_map_routes_each_agent_to_its_shard():
+    agent_to_shard = build_agent_shard_map(
+        {
+            "judged": {
+                "math_agent": ["responses_api_agents"],
+                "math_env": ["resources_servers"],
+            },
+            "tools": {
+                "bash_agent": ["responses_api_agents"],
+                "bash_tools": ["resources_servers"],
+            },
+        }
+    )
+
+    assert agent_to_shard == {"math_agent": "judged", "bash_agent": "tools"}
+
+
+def test_build_agent_shard_map_rejects_an_agent_in_two_shards():
+    """Rows naming the agent could go to either shard, so routing is undefined."""
+    with pytest.raises(ShardSetupError, match="hosted by both shard"):
+        build_agent_shard_map(
+            {
+                "judged": {"math_agent": ["responses_api_agents"]},
+                "tools": {"math_agent": ["responses_api_agents"]},
+            }
+        )
+
+
+def test_agents_are_never_allowlisted_for_duplication():
+    """The allowlist covers shared support entries, not routing ambiguity."""
+    with pytest.raises(ShardSetupError, match="hosted by both shard"):
+        build_agent_shard_map(
+            {
+                "a": {"math_agent": ["responses_api_agents"]},
+                "b": {"math_agent": ["responses_api_agents"]},
+            },
+            allowed_duplicate_entries={"math_agent"},
+        )
+
+
+def test_build_agent_shard_map_rejects_an_unlisted_duplicate_entry():
+    with pytest.raises(ShardSetupError, match="allowed_duplicate_entries"):
+        build_agent_shard_map(
+            {
+                "judged": {"shared_judge": ["responses_api_models"]},
+                "tools": {"shared_judge": ["responses_api_models"]},
+            }
+        )
+
+
+def test_build_agent_shard_map_allows_a_listed_duplicate_entry():
+    """Policy proxies are copied into every shard on purpose."""
+    agent_to_shard = build_agent_shard_map(
+        {
+            "judged": {
+                "math_agent": ["responses_api_agents"],
+                "policy_model": ["responses_api_models"],
+            },
+            "tools": {
+                "bash_agent": ["responses_api_agents"],
+                "policy_model": ["responses_api_models"],
+            },
+        },
+        allowed_duplicate_entries={"policy_model"},
+    )
+
+    assert agent_to_shard == {"math_agent": "judged", "bash_agent": "tools"}
 
 
 def test_apply_shard_log_dir_gives_each_shard_its_own_directory():
--- a/tests/unit/environments/test_nemo_gym_utils.py
+++ b/tests/unit/environments/test_nemo_gym_utils.py
@@ -349,3 +349,312 @@ def test_list_entries_before_spinup_raises():
 
     with pytest.raises(RuntimeError, match="call _spinup"):
         actor.list_entries()
+
+
+class _FakeGymCluster:
+    """Stands in for Ray while the factory creates, starts and queries actors.
+
+    Each actor returns tagged sentinels from ``.remote()`` so ``ray.get`` can
+    tell spinups from entry queries and fail whichever the test asks it to.
+    """
+
+    def __init__(
+        self, *, entries_by_index=None, spinup_failures=None, pg_ready_error=None
+    ):
+        self.entries_by_index = entries_by_index or {}
+        self.spinup_failures = spinup_failures or {}
+        self.pg_ready_error = pg_ready_error
+        self.actors = []
+        self.actor_options = []
+        self.actor_configs = []
+        self.removed_placement_groups = []
+        self.placement_group_calls = []
+        self.placement_group = MagicMock(name="placement_group")
+        self.placement_group.ready.return_value = "pg-ready"
+
+    def make_placement_group(self, **kwargs):
+        self.placement_group_calls.append(kwargs)
+        return self.placement_group
+
+    def remove_placement_group(self, pg):
+        self.removed_placement_groups.append(pg)
+
+    def make_actor_class(self):
+        actor_class = MagicMock(name="NemoGym")
+
+        def options(**option_kwargs):
+            holder = MagicMock()
+
+            def remote(config):
+                index = len(self.actors)
+                actor = MagicMock(name=f"gym-actor-{index}")
+                actor._spinup.remote.return_value = ("spinup", index)
+                actor.list_entries.remote.return_value = ("entries", index)
+                self.actors.append(actor)
+                self.actor_options.append(option_kwargs)
+                self.actor_configs.append(config)
+                return actor
+
+            holder.remote = remote
+            return holder
+
+        actor_class.options = options
+        return actor_class
+
+    def ray_get(self, reference, timeout=None):
+        if reference == "pg-ready":
+            if self.pg_ready_error is not None:
+                raise self.pg_ready_error
+            return None
+        kind, index = reference
+        if kind == "spinup":
+            failure = self.spinup_failures.get(index)
+            if failure is not None:
+                raise failure
+            return None
+        return self.entries_by_index.get(index, {})
+
+
+@contextmanager
+def _patched_cluster(cluster):
+    with (
+        patch.object(nemo_gym_mod, "NemoGym", cluster.make_actor_class()),
+        patch.object(
+            nemo_gym_mod, "make_actor_runtime_env", return_value={"py_executable": "p"}
+        ),
+        patch.object(
+            nemo_gym_mod, "placement_group", side_effect=cluster.make_placement_group
+        ),
+        patch.object(
+            nemo_gym_mod,
+            "remove_placement_group",
+            side_effect=cluster.remove_placement_group,
+        ),
+        patch.object(nemo_gym_mod, "shutdown_environments") as shutdown,
+        patch.object(nemo_gym_mod, "ray") as mock_ray,
+    ):
+        mock_ray.get.side_effect = cluster.ray_get
+        mock_ray.get_runtime_context.return_value.get_node_id.return_value = "a" * 56
+        cluster.shutdown_environments = shutdown
+        yield cluster
+
+
+def _shard_env_configs(**overrides):
+    nemo_gym = {
+        "num_gpu_nodes": 1,
+        "shards": [
+            {"name": "judged", "config_paths": ["judge.yaml"]},
+            {"name": "tools", "config_paths": ["tools.yaml"], "replicas": 2},
+        ],
+        "allowed_duplicate_entries": ["policy_model"],
+        "nemo_gym_log_dir": "/logs/gym",
+    }
+    nemo_gym.update(overrides)
+    return {"nemo_gym": nemo_gym}
+
+
+def test_build_nemo_gym_actors_unsharded_makes_exactly_one_actor(detected_uv_dirs):
+    """The pre-sharding path must stay identical: one actor, no placement group."""
+    cluster = _FakeGymCluster()
+
+    with _patched_cluster(cluster):
+        shard_set = nemo_gym_mod.build_nemo_gym_actors(
+            _env_configs(),
+            base_urls=["http://vllm-0"],
+            model_name="test-model",
+        )
+
+    assert not shard_set.is_sharded
+    assert shard_set.sole_handle() is cluster.actors[0]
+    assert cluster.placement_group_calls == []
+    # Node affinity still applies when the actor has colocated GPUs.
+    assert isinstance(
+        cluster.actor_options[0]["scheduling_strategy"],
+        nemo_gym_mod.NodeAffinitySchedulingStrategy,
+    )
+
+
+def test_build_nemo_gym_actors_spreads_every_replica_onto_its_own_node(
+    detected_uv_dirs,
+):
+    cluster = _FakeGymCluster(
+        entries_by_index={
+            0: {"math_agent": ["responses_api_agents"]},
+            1: {"bash_agent": ["responses_api_agents"]},
+        }
+    )
+
+    with _patched_cluster(cluster):
+        shard_set = nemo_gym_mod.build_nemo_gym_actors(
+            _shard_env_configs(),
+            base_urls=["http://vllm-0"],
+            model_name="test-model",
+        )
+
+    # Two shards, one with replicas: 2, so three actors on three nodes.
+    assert len(cluster.actors) == 3
+    assert [len(replicas) for replicas in shard_set.handles.values()] == [1, 2]
+
+    (pg_kwargs,) = cluster.placement_group_calls
+    assert pg_kwargs["strategy"] == "STRICT_SPREAD"
+    assert pg_kwargs["bundles"] == [
+        {"CPU": float(nemo_gym_mod.DEFAULT_SHARD_CPUS)} for _ in range(3)
+    ]
+
+    # Each actor is pinned to its own bundle.
+    bundle_indices = [
+        options["scheduling_strategy"].placement_group_bundle_index
+        for options in cluster.actor_options
+    ]
+    assert bundle_indices == [0, 1, 2]
+    assert shard_set.agent_to_shard == {"math_agent": "judged", "bash_agent": "tools"}
+
+
+def test_shards_get_their_own_config_paths_and_log_directories(detected_uv_dirs):
+    cluster = _FakeGymCluster()
+
+    with _patched_cluster(cluster):
+        nemo_gym_mod.build_nemo_gym_actors(
+            _shard_env_configs(),
+            base_urls=["http://vllm-0"],
+            model_name="test-model",
+        )
+
+    gym_configs = [
+        config["initial_global_config_dict"] for config in cluster.actor_configs
+    ]
+    assert [config["config_paths"] for config in gym_configs] == [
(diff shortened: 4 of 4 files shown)


nemo_rl/environments/nemo_gym.py
```diff
@@ -1419,11 +1177,353 @@ def build_nemo_gym_config(
...
+
+    def shutdown(self) -> None:
+        """Stop every actor, then release the bundles they were pinned to."""
+        shutdown_environments(
+            {
+                f"nemo_gym[{shard}][{replica}]": handle
+                for shard, replicas in self.handles.items()
+                for replica, handle in enumerate(replicas)
+            },
+            timeout=None,
```
terrykong: **2 action items. Each fixes a different hang; doing one leaves the other live.**

**TL;DR — there are two unbounded waits here, not one: the startup drain blocks before teardown can ever begin, and the shutdown itself is untimed, which makes the `ray.kill` fallback unreachable. Neither fix covers the other's case.**

Not a regression — the pre-PR path was also an untimed `ray.get`. The narrower claim: the new 1800s `spinup_timeout` does not bound the failure path it was added for.

**Hang 1 — `_spinup` never returns.** Gym's `RunHelper.start()` has two `while True: … sleep(3)` loops with no deadline: the [head-server wait](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L460-L470) and [`wait_for_spinup`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L555-L582), whose only exit is `poll()`, which fires only when a subprocess *exits*. (This is not "all of Gym is unbounded" — [`wait_for_model_endpoints`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L648-L662) does take a bound.) When the 1800s budget runs out, the drain at [L1487-L1491](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym.py#L1487-L1491) re-gets every pending ref with **no timeout** and blocks there, so the code never reaches the `raise` or `shard_set.shutdown()`. A finite shutdown timeout does nothing for this case. **Only AI-2 helps.**

**Hang 2 — `RunHelper.shutdown()` wedges.** The drain is not involved. Gym builds the head server with [`uvicorn.Config(app, host, port)`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/server_utils.py#L793-L797), leaving `timeout_graceful_shutdown` unset, so uvicorn's shutdown waits with `asyncio.wait_for(..., timeout=None)`. The `force_exit` escape from that wait is set in exactly one place — inside `handle_exit`, which only `capture_signals` installs, and `capture_signals` returns early off the main thread. Gym runs the server as [`Thread(target=uvicorn_server.run, daemon=True)`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/server_utils.py#L800-L801), so `force_exit` stays `False` and one open connection keeps [`_head_server_thread.join()`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L622-L623) spinning. Then [`shutdown_environments(..., timeout=None)`](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/utils.py#L174-L183) means the `ray.get` never raises, so the `except Exception → ray.kill` arm is dead code. **Only AI-1 helps.**

No hang was observed here — what is proven from source is that the wait has no bound and `force_exit` can never be set off the main thread. A trigger for it is plausible but was not run.

`ray.kill` is the only thing that can stop an actor wedged this way, because `NemoGym` is an asyncio actor: it has [`async def run_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym.py#L585), and Ray sets `is_asyncio = has_async_methods(cls)` ([ray 2.56.1 actor.py:1806](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L1806)). Its default `max_concurrency` is 1000, not 1 — but that is 1000 coroutines on ONE event loop, and Ray wraps a sync method with [`sync_to_async`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/async_compat.py#L44-L54), a bare coroutine wrapper with no `run_in_executor`. `_spinup`, `shutdown`, `set_tokenizer` and `list_entries` are all plain `def`, so a blocking body never yields and owns the loop. A queued `shutdown()` cannot run while `_spinup` is stuck. While either hang holds, `remove_placement_group` is never reached, so the bundles stay reserved too.

Two tests pin the current behaviour and have to change with whichever fix lands. [test_nemo_gym_utils.py:827-829](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/tests/unit/environments/test_nemo_gym_utils.py#L827-L829) asserts `invocation.kwargs == {"timeout": None}`, pinning `timeout=None` directly; [test_nemo_gym_utils.py:746](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/tests/unit/environments/test_nemo_gym_utils.py#L746) asserts `spinup_zero_calls[1] is None`, where the second entry is the drain's re-get, pinning the drain as untimed. No suggestion block for either: the right assertion depends on which shape you pick, so a committable block now would only be correct after a change this same comment is asking for.

### AI-1

Make the shutdown timeout finite, so `ray.kill` becomes reachable when `RunHelper.shutdown()` wedges. Pick either — both are finite, which is the whole ask.

Preferred: name it beside the two existing `DEFAULT_SHARD_*_TIMEOUT_SECONDS` constants and let callers override.

```python
# Waiting for a shard actor to stop. Sized above Gym's serial reaping so it can
# only fire on a real wedge; it exists to make ray.kill reachable, not to be a
# sufficient budget.
DEFAULT_SHARD_SHUTDOWN_TIMEOUT_SECONDS = 300.0

    def shutdown(self, timeout: float | None = DEFAULT_SHARD_SHUTDOWN_TIMEOUT_SECONDS) -> None:
        ...
        shutdown_environments({...}, timeout=timeout)
```

Keep that default finite. A parameter defaulting to `None` leaves every no-argument caller unbounded, which is exactly the case that hangs.

If you would rather not add a constant in this PR, the one-click version:

```suggestion
            timeout=300.0,
```

Why 300 and not the repo's [10s default](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/utils.py#L24): `RunHelper.shutdown()` [signals every process, then waits and reaps them one at a time](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L586-L620) — the SIGINTs go out in their own loop first, then each process gets [1s graceful + 2s reap](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L91-L93) in the worst case, so roughly 3s × N. 10s is below that for a large stack and really could cut a reap short; avoiding it was right, only picking unbounded overshot. 300s clears legitimate reaping and can only fire on a real wedge.

That reap order also answers the obvious objection to `ray.kill`: the subprocesses are reaped **before** the wedging `join()`, so when a finite timeout fires there they are already dead and `ray.kill` orphans nothing.

The point is finiteness, not sufficiency — the untimed `join()` means no finite number is provably enough. A finite value is what makes `ray.kill` reachable at all.

### AI-2

Bound the startup drain at [L1487-L1491](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym.py#L1487-L1491) with one shared grace deadline, so a `_spinup` that never returns cannot stop teardown from starting. Not a one-click block — that region is a different part of the file than this comment's anchor.

```python
    drain_deadline = monotonic() + DEFAULT_SHARD_DRAIN_GRACE_SECONDS
    for _, _, pending_reference in pending:
        try:
            ray.get(pending_reference, timeout=max(0.0, drain_deadline - monotonic()))
        except Exception:
            pass
```

One shared deadline, not one per ref — per-ref multiplies the wait by the number of shards.

Narrow the swallow to `except Exception:` in the same edit. That `except BaseException: pass` is the only bare `BaseException` swallow in `nemo_rl/`; the other 11 sites all clean up and then `raise`. It eats `KeyboardInterrupt`, so an operator has to press Ctrl-C once per pending shard.

This fix lands nowhere else in the stack — #3372 touches `utils.py` but keeps `ray.get(..., timeout=timeout)` and changes neither the drain nor the deadline.

<details><summary>Why the existing drain test does not catch this</summary>

The fake in [test_nemo_gym_utils.py](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/tests/unit/environments/test_nemo_gym_utils.py#L725-L747) is faithful for a ref that is slow but does complete — it returns on the re-get. It **cannot model a ref that never resolves**, which is the case here. Its guard includes `timeout is not None`, so it is coded such that an untimed get always succeeds: it encodes the bug's own assumption as its model of Ray.

`shutdown_environments` is also mocked in that test, so `timeout=None` is never exercised on the real path.
</details>

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -379,6 +383,56 @@ def apply_shard_overlay(
...
+                    f"agent must live in exactly one shard so rows naming it "
+                    f"have one destination."
+                )
+            agent_to_shard[entry] = shard_name
+
+    hosting_shard: dict[str, str] = {}
+    for shard_name, entries in entries_by_shard.items():
+        for entry in entries:
+            if entry in agent_to_shard:
+                continue
```
terrykong: **2 action items.**

**TL;DR — this `continue` skips any name that is an agent in one shard and a plain entry in another, so that collision reaches neither check and two shards silently start the same GPU-holding model engine.**

Loop 1 registers a name only when *that* shard reports it under `responses_api_agents`. Loop 2 then skips anything already in `agent_to_shard`. So a name that is an agent in shard A and a non-agent in shard B is registered by loop 1 (from A) and skipped by loop 2 (for B) — it never reaches the `allowed_duplicate_entries` check.

The function's own docstring says that is exactly what it prevents ([shards.py L395-L397](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym_shards.py#L395-L397)):

> a shared YAML dropped into two shards' path lists quietly brings its judge along and doubles that judge's GPU claim

Repro:

```python
build_agent_shard_map({
  'judged': {'policy_model': ['responses_api_agents','responses_api_models']},
  'tools':  {'policy_model': ['responses_api_models']},
})
# -> {'policy_model': 'judged'}   no raise, no allowlist entry required
```

In `tools`, `policy_model` is a model engine, so the second copy really does claim GPUs.

### AI-1

Delete the skip. It cannot false-positive: an agent duplicated across shards already raises in loop 1, so loop 2 never sees one. All 46 existing tests in [test_nemo_gym_shards.py](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/tests/unit/environments/test_nemo_gym_shards.py) still pass with the skip removed — the allowlist behaviour is unchanged.

The suggestion block below is empty on purpose — it deletes the two lines.

```suggestion
```

### AI-2

Add a regression test for the collision to [tests/unit/environments/test_nemo_gym_shards.py](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/tests/unit/environments/test_nemo_gym_shards.py). Today it fails with `DID NOT RAISE ShardSetupError`; it passes once the two lines above are deleted, so commit it together with that change rather than on its own. Not a one-click block — it targets a different file than this comment's anchor; paste it at the end of that file.

```python
    assert agent_to_shard == {"math_agent": "judged", "bash_agent": "tools"}


def test_a_name_that_is_an_agent_in_one_shard_is_still_a_duplicate_elsewhere():
    """Reusing an agent's name for a model engine doubles that engine's GPU claim."""
    with pytest.raises(ShardSetupError, match="allowed_duplicate_entries"):
        build_agent_shard_map(
            {
                "judged": {"math": ["responses_api_agents"]},
                "tools": {
                    "math": ["responses_api_models"],
                    "bash": ["responses_api_agents"],
                },
            }
        )
```

nemo_rl/environments/nemo_gym.py
```diff
@@ -1419,11 +1177,353 @@ def build_nemo_gym_config(
...
+                f"config_paths is the usual cause: {error}"
+            ) from error
+
+    installs = [
+        (shard_name, replica, handle.set_tokenizer.remote(tokenizer))
+        for shard_name, replica, handle in instance_handles
+    ]
+    for shard_name, replica, reference in installs:
+        try:
+            ray.get(reference, timeout=max(0.0, deadline - monotonic()))
```
terrykong: **2 action items.**

**TL;DR — the spinup loop and this install loop share one deadline, so when startup legitimately uses the whole budget every tokenizer install gets `timeout=0.0` and fails instantly with "did not accept the tokenizer" — an error about the tokenizer when nothing is wrong with the tokenizer.**

[`deadline`](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym.py#L1474) is set once and feeds both the [spinup loop](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym.py#L1482) and this one. Ray treats `timeout=0.0` as an immediate `GetTimeoutError`, not as "no timeout" — [worker.py L999-L1001](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/worker.py#L999-L1001) and the [`ray.get` docstring](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/worker.py#L2895-L2899).

Cold node, spinup takes ~1790s of the 1800s budget → `deadline - monotonic()` is at or below 0 for the install pass → every install that has not already finished raises → the operator reads `NeMo-Gym shard 'X' (replica N) did not accept the tokenizer`. Your own comment at [L105-L108](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym.py#L105-L108) says a cold node nearly exhausting 1800s is the expected case.

Installs that already completed still return at `timeout=0`, so which replica trips varies run to run. Replica 0 is checked first and so has had the least time to finish — it is the likeliest victim.

### AI-1

Give the install pass its own budget. One option — of several; a smaller dedicated constant would be fine too. This one reuses a value already in scope so it applies cleanly on its own.

```suggestion
    # The install pass gets its own budget. Spinup legitimately using all of
    # ``spinup_timeout`` must not leave every install with timeout=0.0.
    install_deadline = monotonic() + spinup_timeout
    for shard_name, replica, reference in installs:
        try:
            ray.get(reference, timeout=max(0.0, install_deadline - monotonic()))
```

### AI-2

Add a test for the tokenizer-install failure branch to [tests/unit/environments/test_nemo_gym_utils.py](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/tests/unit/environments/test_nemo_gym_utils.py). That branch has zero coverage today: `accept the tokenizer` appears nowhere under `tests/`, and `_FakeGymCluster.ray_get` unconditionally returns `None` for tokenizer refs ([test_nemo_gym_utils.py:484-485](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/tests/unit/environments/test_nemo_gym_utils.py#L484-L485)), so neither the failure branch nor its teardown has ever run. The test below passes against today's unmodified code, so you can commit it on its own, in either order relative to AI-1. Not a one-click block — it targets a different file than this comment's anchor; paste it at the end of that file.

It also stays correct after the separate-deadline change, because it triggers a real install failure — it raises `RuntimeError("tokenizer is not serializable")` on a tokenizer ref and asserts the shard-naming and teardown. It does not test a spent deadline. A genuine tokenizer failure should keep reporting "did not accept the tokenizer"; only the exhausted-budget case should stop using that message.

```python
    # Releasing a group twice raises; the second shutdown must not try.
    remove.assert_called_once_with(pg)


def test_a_shard_that_rejects_the_tokenizer_names_itself_and_tears_everything_down(
    detected_uv_dirs,
):
    """A replica without the tokenizer fails on its first rollout, not at setup.

    The install pass is the last thing that can catch that, so its failure must
    name the shard and take the whole set down.
    """
    cluster = _FakeGymCluster()
    fake_ray_get = cluster.ray_get

    def ray_get(reference, timeout=None):
        if reference == ("tokenizer", 1):
            raise RuntimeError("tokenizer is not serializable")
        return fake_ray_get(reference, timeout=timeout)

    cluster.ray_get = ray_get

    with _patched_cluster(cluster):
        with pytest.raises(
            nemo_gym_mod.ShardSetupError,
            match="shard 'tools' \\(replica 0\\) did not accept the tokenizer",
        ):
            nemo_gym_mod.build_nemo_gym_actors(
                _shard_env_configs(),
                base_urls=["http://vllm-0"],
                model_name="test-model",
                tokenizer=_TOKENIZER,
                enable_router_replay=False,
                use_fastokens=False,
            )

    cluster.shutdown_environments.assert_called_once()
    assert len(cluster.shutdown_environments.call_args.args[0]) == 3
    assert cluster.removed_placement_groups == [cluster.placement_group]
```

nemo_rl/environments/nemo_gym.py
```diff
@@ -1419,11 +1177,353 @@ def build_nemo_gym_config(
...
+        handles: Shard name to its replica handles, in replica order.
+        agent_to_shard: Agent entry name to the shard hosting it. Empty when
+            unsharded, where every row goes to the only actor anyway.
+        placement_group: The STRICT_SPREAD group pinning shards to distinct
+            nodes, or None when unsharded.
+    """
+
+    handles: Dict[str, List[Any]]
+    agent_to_shard: Dict[str, str] = field(default_factory=dict)
+    placement_group: Any = None
```
terrykong: **1 action item.** Low.

**TL;DR — `placement_group: Any` hides that `is_sharded` is just a `None` check on a placement group, and the repo already has a concrete spelling for both of these fields.**

The established spellings are `list[ray.actor.ActorHandle]`, in [worker_groups.py L371](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/distributed/worker_groups.py#L371), and `Optional[list[PlacementGroup]]`, in [virtual_cluster.py L818](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/distributed/virtual_cluster.py#L818).

### AI-1

Not a one-click block: it also needs `PlacementGroup` added to the [ray import at L27](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym.py#L27), which is a different region than this anchor.

```python
    handles: Dict[str, List[ray.actor.ActorHandle]]
    agent_to_shard: Dict[str, str] = field(default_factory=dict)
    placement_group: Optional[PlacementGroup] = None
```

**Context — no action.** This cannot turn CI red either way: `nemo_gym.py` is not in [`pyrefly.toml` `project-includes`](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/pyrefly.toml#L25).

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -379,6 +383,56 @@ def apply_shard_overlay(
...
+    has to be allowlisted, because duplication is usually accidental — a shared
+    YAML dropped into two shards' path lists quietly brings its judge along and
+    doubles that judge's GPU claim.
+
+    Only names are compared. What an entry means is Gym's business.
+    """
+    agent_to_shard: dict[str, str] = {}
+    for shard_name, entries in entries_by_shard.items():
+        for entry, types in entries.items():
+            if "responses_api_agents" not in types:
```
terrykong: **1 action item.** Low.

**TL;DR — `"responses_api_agents"` is written here as a bare string, while this module's stated convention is to give Gym's keys a named module constant.**

The string itself is right and complete. It is the spelling that differs from the module's own rule ([shards.py L54-L57](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym_shards.py#L54-L57)):

```python
# Gym's log-directory key (``nemo_gym.global_config.NEMO_GYM_LOG_DIR_KEY_NAME``).
# Restated rather than imported: this module is imported on the driver, and
# nemo_gym is installed only in the actor's venv.
GYM_LOG_DIR_KEY = "nemo_gym_log_dir"
```

### AI-1

Add a local constant next to `GYM_LOG_DIR_KEY` and use it here. Not a one-click block — the declaration goes at the top of the module, which is a different region than this anchor.

```python
# Gym's server-type key marking an entry as an agent.
GYM_AGENTS_KEY = "responses_api_agents"
```

then here:

```python
            if GYM_AGENTS_KEY not in types:
```

**Context — no action.** Do not import `GYM_SERVER_TYPE_KEYS` from `nemo_gym.py` instead — [`nemo_gym.py` already imports from this module](https://github.com/NVIDIA-NeMo/RL/blob/917ffc3ec18b9ec85ca25158628b2b9cf0b38bfc/nemo_rl/environments/nemo_gym.py#L53), so that would be an import cycle. Keep the constant local.

nemo_rl/environments/nemo_gym.py
```diff
@@ -1429,6 +1488,356 @@ def build_nemo_gym_config(
...
+
+    def shutdown(self) -> None:
+        """Stop every actor, then release the bundles they were pinned to."""
+        shutdown_environments(
+            {
+                f"nemo_gym[{shard}][{replica}]": handle
+                for shard, replicas in self.handles.items()
+                for replica, handle in enumerate(replicas)
+            },
+            timeout=None,
```
terrykong: **2 action items. Each fixes a different hang; doing one leaves the other live.**

**TL;DR — there are two unbounded waits here, not one: the startup drain blocks before teardown can begin, and the shutdown itself is untimed, which makes the `ray.kill` fallback unreachable. Neither fix covers the other's case.**

Not a regression: the pre-PR path was also untimed. The claim is narrower — the new 1800s `spinup_timeout` does not bound the failure path it was added for.

**Hang 1 — `_spinup` never returns.** Gym's `RunHelper.start()` has two `while True: … sleep(3)` loops with no deadline: the [head-server wait](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L460-L470) and [`wait_for_spinup`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L555-L582). When the budget runs out, the drain at [L1801-L1805](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L1801-L1805) re-gets every pending ref with **no timeout** and blocks there, so the code never reaches the `raise` or `shard_set.shutdown()`. **Only AI-2 helps.**

**Hang 2 — `RunHelper.shutdown()` wedges** on an untimed [`_head_server_thread.join()`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L622-L623). Then [`shutdown_environments(..., timeout=None)`](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/utils.py#L175) means the `ray.get` never raises, so the `except Exception → ray.kill` arm is dead code. **Only AI-1 helps.**

`ray.kill` is the only way out of either one — a queued `shutdown()` cannot even start while `_spinup` holds the actor's event loop. Both also leave the bundles reserved.

### AI-1

Make the shutdown timeout finite, so `ray.kill` becomes reachable. The point is finiteness, not sufficiency — the untimed `join()` means no number is provably enough.

```suggestion
            timeout=300.0,
```

Better if you want it named: add `DEFAULT_SHARD_SHUTDOWN_TIMEOUT_SECONDS = 300.0` beside the two existing `DEFAULT_SHARD_*_TIMEOUT_SECONDS` constants and give `shutdown()` a `timeout` parameter defaulting to it. Keep that default finite — defaulting to `None` leaves every no-argument caller unbounded, which is the case that hangs.

### AI-2

Bound the startup drain at [L1801-L1805](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L1801-L1805) with one shared grace deadline, so a `_spinup` that never returns cannot stop teardown. Not a one-click block — that region is elsewhere in the file.

```python
    drain_deadline = monotonic() + DEFAULT_SHARD_DRAIN_GRACE_SECONDS
    for _, _, pending_reference in pending:
        try:
            ray.get(pending_reference, timeout=max(0.0, drain_deadline - monotonic()))
        except Exception:
            pass
```

One shared deadline, not one per ref — per-ref multiplies the wait by the shard count. Narrow the swallow to `except Exception:` in the same edit: that bare `except BaseException: pass` is the only one in `nemo_rl/`, and it eats `KeyboardInterrupt`, so Ctrl-C has to be pressed once per pending shard.

<details><summary>Why <code>RunHelper.shutdown()</code> can wedge forever</summary>

Gym builds the head server with [`uvicorn.Config(app, host, port)`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/server_utils.py#L793-L797), leaving `timeout_graceful_shutdown` unset, so uvicorn's shutdown waits with `asyncio.wait_for(..., timeout=None)`. The `force_exit` escape from that wait is set in exactly one place — inside `handle_exit`, which only `capture_signals` installs, and `capture_signals` returns early off the main thread. Gym runs the server as [`Thread(target=uvicorn_server.run, daemon=True)`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/server_utils.py#L800-L801), so `force_exit` stays `False` and one open connection keeps the join spinning.

No hang was observed here. What is proven from source is that the wait has no bound and `force_exit` cannot be set off the main thread; a trigger is plausible but was not run.

This is not "all of Gym is unbounded" — [`wait_for_model_endpoints`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L648-L662) does take a bound.

Why `ray.kill` is the only way out, in full: `NemoGym` has [`async def run_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L684), so Ray sets `is_asyncio = has_async_methods(cls)` ([actor.py:1806](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L1806)) and gives it a default `max_concurrency` of 1000 — but that is 1000 coroutines on ONE event loop, and Ray wraps a sync method with [`sync_to_async`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/async_compat.py#L44-L54), a bare coroutine wrapper with no `run_in_executor`. `_spinup`, `shutdown`, `set_tokenizer` and `list_entries` are all plain `def`, so a blocking body never yields and a queued `shutdown()` never runs. And while either hang holds, `remove_placement_group` is never reached, so the bundles stay reserved.
</details>

<details><summary>Where 300s comes from, and why killing there orphans nothing</summary>

`RunHelper.shutdown()` [signals every process, then waits and reaps them one at a time](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L586-L620): the SIGINTs go out in their own loop first, then each process gets [1s graceful + 2s reap](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L91-L93) in the worst case — roughly 3s × N. The repo's [10s default](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/utils.py#L24) is below that for a large stack and really could cut a reap short, so avoiding it was right; only picking unbounded overshot.

That reap order also answers the obvious objection to `ray.kill`: the subprocesses are reaped **before** the wedging join, so when a finite timeout fires there they are already dead and `ray.kill` orphans nothing.
</details>

<details><summary>Two existing tests pin the current behaviour and must change with the fix</summary>

[test_nemo_gym_utils.py:859](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/tests/unit/environments/test_nemo_gym_utils.py#L859) asserts `invocation.kwargs == {"timeout": None}`, pinning `timeout=None` directly. [test_nemo_gym_utils.py:777](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/tests/unit/environments/test_nemo_gym_utils.py#L777) asserts `spinup_zero_calls[1] is None`, where the second entry is the drain's re-get, pinning the drain as untimed. No suggestion block for either: the right assertion depends on which shape you pick.

The fake in [that test](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/tests/unit/environments/test_nemo_gym_utils.py#L756-L778) is faithful for a ref that is slow but does complete — it returns on the re-get. It **cannot model a ref that never resolves**, which is the case here: its guard includes `timeout is not None`, so it is coded such that an untimed get always succeeds. `shutdown_environments` is also mocked there, so `timeout=None` is never exercised on the real path.

This fix lands nowhere else in the stack — #3372 touches `utils.py` but keeps `ray.get(..., timeout=timeout)` and changes neither the drain nor the deadline.
</details>

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -410,6 +414,59 @@ def apply_shard_overlay(
...
+                    f"agent or resources server must live in exactly one shard "
+                    f"so rows naming it have one destination."
+                )
+            route_to_shard[entry] = shard_name
+
+    hosting_shard: dict[str, str] = {}
+    for shard_name, entries in entries_by_shard.items():
+        for entry in entries:
+            if entry in route_to_shard:
+                continue
```
terrykong: **2 action items.**

**TL;DR — when a name is routable in one shard and a model engine in another, this `continue` skips it in *both*, so the second shard's model engine is never duplicate-checked and quietly claims GPUs a second time.**

The vocabulary first, because the bug lives in it. An entry is a named config block that declares which kinds of server it provides, and it can declare more than one. This is what a shard reports from [`list_entries()`](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L639):

```python
{"math_agent":   ["responses_api_agents"],                          # rollouts route here
 "math_env":     ["resources_servers"],                             # also routable, via task_source
 "judge_model":  ["responses_api_models"],                          # a model engine, holds GPUs
 "policy_model": ["responses_api_agents", "responses_api_models"]} # both at once
```

"Routable" is not a kind of thing — it is whether *this name, in this shard*, declares `responses_api_agents` or `resources_servers` ([L441](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym_shards.py#L441)). Shards load different `config_paths`, so the same name can declare different kinds in different shards, and nothing checks that it does not.

Loop 1 records every name that is routable in *some* shard. Loop 2 then skips anything in that record, for *every* shard. So a name that is routable in `judged` and only a model engine in `tools` gets skipped in both passes, and the `tools` copy never reaches the `allowed_duplicate_entries` check:

```python
build_route_shard_map({
  'judged': {'policy_model': ['responses_api_agents','responses_api_models']},
  'tools':  {'policy_model': ['responses_api_models']},
})
# -> {'policy_model': 'judged'}   no raise, no allowlist entry required
# 'tools' is now running a second policy_model model engine. Nobody is told.
```

That is the harm the function's own docstring says it prevents ([L429-L433](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym_shards.py#L429-L433)):

> a shared YAML dropped into two shards' path lists quietly brings its judge along and doubles that judge's GPU claim

Both cases traced line by line, with the loop variables at each step: [the duplicate check](https://terrykong.github.io/gh-pages-poc/terryk/pr-3371-duplicate-check.html).

### AI-1

Delete the skip. It cannot over-fire: a name routable in two shards already raised in loop 1, so loop 2 never sees one; a lone routable entry is simply recorded once and passes. All existing tests in [test_nemo_gym_shards.py](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/tests/unit/environments/test_nemo_gym_shards.py) still pass with it removed.

The suggestion block below is empty on purpose — it deletes the two lines.

```suggestion
```

### AI-2

Add a regression test for the collision to [tests/unit/environments/test_nemo_gym_shards.py](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/tests/unit/environments/test_nemo_gym_shards.py). Today it fails with `DID NOT RAISE ShardSetupError`; it passes once the two lines above are deleted, so commit it together with that change rather than on its own. Not a one-click block — it targets a different file than this comment's anchor; paste it at the end of that file.

```python
def test_a_routable_name_is_still_a_duplicate_when_it_is_a_model_elsewhere():
    """Reusing a routable name for a model engine doubles that engine's GPU claim."""
    with pytest.raises(ShardSetupError, match="allowed_duplicate_entries"):
        build_route_shard_map(
            {
                "judged": {"math": ["responses_api_agents"]},
                "tools": {
                    "math": ["responses_api_models"],
                    "bash": ["responses_api_agents"],
                },
            }
        )
```

nemo_rl/environments/nemo_gym.py
```diff
@@ -1429,6 +1488,356 @@ def build_nemo_gym_config(
...
+                f"config_paths is the usual cause: {error}"
+            ) from error
+
+    installs = [
+        (shard_name, replica, handle.set_tokenizer.remote(tokenizer))
+        for shard_name, replica, handle in instance_handles
+    ]
+    for shard_name, replica, reference in installs:
+        try:
+            ray.get(reference, timeout=max(0.0, deadline - monotonic()))
```
terrykong: **2 action items.**

**TL;DR — when the startup budget runs out, this loop reports `shard 'X' (replica N) did not accept the tokenizer` — naming a shard and the tokenizer when neither is at fault and the real cause is that startup used the whole budget.**

Sharing one [`deadline`](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L1788) across the [spinup loop](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L1796) and this one is right: `spinup_timeout` should bound the whole job of bringing shards up, and installing the tokenizer is part of that. Two independent budgets would make the worst case twice `spinup_timeout` from a single knob. **Not asking to change that.**

What is wrong is only the message. On a cold node spinup finishes just under the budget — your own comment at [L106-L109](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L106-L109) says that is the expected case — so the remaining time is at or near zero, Ray's `timeout=0.0` raises at once, and the operator reads an error about the tokenizer. Nothing about the tokenizer is wrong; the budget simply ran out.

Background, with the numbers: [the shared deadline](https://terrykong.github.io/gh-pages-poc/terryk/pr-3371-shared-deadline.html).

### AI-1

Say the budget ran out when it ran out. Not a one-click block — the change spans the loop body and its `raise` at [L1816-L1823](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L1816-L1823), which is wider than this comment's anchor.

```python
    for shard_name, replica, reference in installs:
        remaining = max(0.0, deadline - monotonic())
        try:
            ray.get(reference, timeout=remaining)
        except BaseException as error:
            if remaining <= 0.0:
                raise ShardSetupError(
                    f"NeMo-Gym shards used the whole {spinup_timeout}s startup "
                    f"budget before the tokenizer reached every replica. Raise "
                    f"env.nemo_gym.spinup_timeout, or bake the Gym venvs into "
                    f"the image so a cold node starts faster."
                ) from error
            raise ShardSetupError(
                f"NeMo-Gym shard '{shard_name}' (replica {replica}) did not "
                f"accept the tokenizer: {error}"
            ) from error
```

It also fixes a smaller annoyance: finished installs still return at `timeout=0`, so the shard named today changes from run to run. A budget message names none.

### AI-2

Add a test for the tokenizer-install failure branch to [tests/unit/environments/test_nemo_gym_utils.py](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/tests/unit/environments/test_nemo_gym_utils.py). That branch has zero coverage today: `accept the tokenizer` appears nowhere under `tests/`, and `_FakeGymCluster.ray_get` unconditionally returns `None` for tokenizer refs ([L515](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/tests/unit/environments/test_nemo_gym_utils.py#L515)), so neither the failure branch nor its teardown has ever run. The test below passes against today's unmodified code, so you can commit it on its own, in either order relative to AI-1. Not a one-click block — it targets a different file than this comment's anchor; paste it at the end of that file.

It triggers a real install failure rather than an exhausted budget, so it stays correct after AI-1 — that is the branch that should keep saying "did not accept the tokenizer".

```python
def test_a_shard_that_rejects_the_tokenizer_names_itself_and_tears_everything_down(
    detected_uv_dirs,
):
    """A replica without the tokenizer fails on its first rollout, not at setup."""
    cluster = _FakeGymCluster()
    fake_ray_get = cluster.ray_get

    def ray_get(reference, timeout=None):
        if reference == ("tokenizer", 1):
            raise RuntimeError("tokenizer is not serializable")
        return fake_ray_get(reference, timeout=timeout)

    cluster.ray_get = ray_get

    with _patched_cluster(cluster):
        with pytest.raises(
            nemo_gym_mod.ShardSetupError,
            match="shard 'tools' \\(replica 0\\) did not accept the tokenizer",
        ):
            nemo_gym_mod.build_nemo_gym_actors(
                _shard_env_configs(),
                base_urls=["http://vllm-0"],
                model_name="test-model",
                tokenizer=_TOKENIZER,
                enable_router_replay=False,
                use_fastokens=False,
            )

    cluster.shutdown_environments.assert_called_once()
    assert len(cluster.shutdown_environments.call_args.args[0]) == 3
    assert cluster.removed_placement_groups == [cluster.placement_group]
```

<details><summary>Why an exhausted budget raises instantly rather than waiting</summary>

Ray maps `timeout=0.0` to an immediate `GetTimeoutError` rather than to "no timeout" — [worker.py L999-L1001](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/worker.py#L999-L1001) and the [`ray.get` docstring](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/worker.py#L2895-L2899). So once the shared deadline is spent, every remaining install fails on the spot rather than getting a short grace period. That is the correct behaviour for a spent budget; it is only the wording of the resulting error that misleads.
</details>

nemo_rl/environments/nemo_gym.py
```diff
@@ -1429,6 +1488,356 @@ def build_nemo_gym_config(
...
+        handles: Shard name to its replica handles, in replica order.
+        route_to_shard: Agent or task-source entry name to the shard hosting
+            it. Empty when unsharded, where every row goes to the only actor.
+        placement_group: The STRICT_SPREAD group pinning shards to distinct
+            nodes, or None when unsharded.
+    """
+
+    handles: Dict[str, List[Any]]
+    route_to_shard: Dict[str, str] = field(default_factory=dict)
+    placement_group: Any = None
```
terrykong: **1 action item.** Low.

**TL;DR — `placement_group: Any` hides that `is_sharded` is just a `None` check on a placement group, and the repo already has a concrete spelling for both of these fields.**

The established spellings are `list[ray.actor.ActorHandle]`, in [worker_groups.py L371](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/distributed/worker_groups.py#L371), and `Optional[list[PlacementGroup]]`, in [virtual_cluster.py L860](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/distributed/virtual_cluster.py#L860).

### AI-1

Not a one-click block: it also needs `PlacementGroup` added to the [ray import at L28](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L28), which is a different region than this anchor.

```python
    handles: Dict[str, List[ray.actor.ActorHandle]]
    agent_to_shard: Dict[str, str] = field(default_factory=dict)
    placement_group: Optional[PlacementGroup] = None
```

**Context — no action.** This cannot turn CI red either way: `nemo_gym.py` is not in [`pyrefly.toml` `project-includes`](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/pyrefly.toml#L26).

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -410,6 +414,59 @@ def apply_shard_overlay(
...
+    two shards has to be allowlisted, because duplication is usually accidental
+    — a shared YAML dropped into two shards' path lists quietly brings its judge
+    along and doubles that judge's GPU claim.
+
+    Only names are compared. What an entry means is Gym's business.
+    """
+    route_to_shard: dict[str, str] = {}
+    for shard_name, entries in entries_by_shard.items():
+        for entry, types in entries.items():
+            if not {"responses_api_agents", "resources_servers"}.intersection(types):
```
terrykong: **1 action item.** Low.

**TL;DR — the two Gym server-type keys are written here as bare strings, while this module's stated convention is to give Gym's keys a named module constant.**

The strings themselves are right. It is the spelling that differs from the module's own rule ([L48-L51](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym_shards.py#L48-L51)):

```python
# Gym's log-directory key (``nemo_gym.global_config.NEMO_GYM_LOG_DIR_KEY_NAME``).
# Restated rather than imported: this module is imported on the driver, and
# nemo_gym is installed only in the actor's venv.
GYM_LOG_DIR_KEY = "nemo_gym_log_dir"
```

### AI-1

Name the routable set once at the top of the module and use it here. Not a one-click block — the declaration goes at the top, which is a different region than this anchor.

```python
# Gym's server-type keys that make an entry a rollout destination.
GYM_ROUTABLE_KEYS = frozenset({"responses_api_agents", "resources_servers"})
```

then here:

```python
            if not GYM_ROUTABLE_KEYS.intersection(types):
```

**Context — no action.** Do not import `GYM_SERVER_TYPE_KEYS` from `nemo_gym.py` instead — [`nemo_gym.py` already imports from this module](https://github.com/NVIDIA-NeMo/RL/blob/5368ae619aff52975c8dd21ab690c6cc5853937d/nemo_rl/environments/nemo_gym.py#L54), so that would be an import cycle. Keep the constant local.

nemo_rl/environments/nemo_gym.py
```diff
@@ -1439,65 +1581,329 @@ def spinup_nemo_gym_actor(
...
+                )
             )
+            shard_set.handles.setdefault(shard.name, []).append(actor)
+
+        _spinup_shards_concurrently(shard_set, spinup_timeout, tokenizer=tokenizer)
+        shard_set.route_to_shard = _discover_route_shard_map(shard_set, plan)
+    except BaseException:
+        # A ray.get timeout does not cancel the actor-side work, so a
+        # half-started stack would keep running with nothing left to stop it.
+        shard_set.shutdown(force_kill=True)
```
terrykong: **2 action items.**

**TL;DR — this call leaves `timeout` at its `None` default, so `shutdown_environments` blocks forever on a wedged actor and the `ray.kill` underneath it never runs; the drain above never lets us get here in the first place.**

The unsharded path already does this correctly — [L1674-L1677](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L1674-L1677) passes both `timeout=NEMO_GYM_GRACEFUL_SHUTDOWN_TIMEOUT_S` and `force_kill=True`. This sharded call site passes only `force_kill`.

That gap matters because of the order inside [`shutdown()`](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L1535-L1559): it calls `shutdown_environments(..., timeout=timeout)` first and only reaches [`if force_kill:`](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L1554) once that returns. With `timeout=None`, the [`ray.get`](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/utils.py#L175) inside it has no bound, so a wedged actor parks there and the kill loop is never entered.

**And a second, independent wait sits in front of it.** When `spinup_timeout` fires, the drain at [L1820-L1824](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L1820-L1824) re-gets every pending ref with **no timeout** and blocks, so the code never reaches the `raise` — or this teardown — at all. Gym's `RunHelper.start()` is not guaranteed to return: two `while True: … sleep(3)` loops with no deadline, the [head-server wait](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L460-L470) and [`wait_for_spinup`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L555-L582). Fixing either one alone still leaves a hang.

Not a regression: the pre-PR path was also untimed. The claim is narrower — the 1800s `spinup_timeout` does not bound the failure path it was added for.

### AI-1

Pass the timeout here too, matching the unsharded call site.

```suggestion
        shard_set.shutdown(
            timeout=NEMO_GYM_GRACEFUL_SHUTDOWN_TIMEOUT_S,
            force_kill=True,
        )
```

### AI-2

Bound the startup drain at [L1820-L1824](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L1820-L1824) with one shared grace deadline, so a `_spinup` that never returns cannot stop teardown from starting. Not a one-click block — that region is elsewhere in the file.

```python
    drain_deadline = monotonic() + DEFAULT_SHARD_DRAIN_GRACE_SECONDS
    for _, _, pending_reference in pending:
        try:
            ray.get(pending_reference, timeout=max(0.0, drain_deadline - monotonic()))
        except Exception:
            pass
```

One shared deadline, not one per ref — per-ref multiplies the wait by the shard count. Narrow the swallow to `except Exception:` in the same edit: that bare `except BaseException: pass` is the only one in `nemo_rl/`, and it eats `KeyboardInterrupt`, so Ctrl-C has to be pressed once per pending shard.

<details><summary>Why a wedged actor is reachable at all</summary>

`RunHelper.shutdown()` can wedge on its final, untimed [`_head_server_thread.join()`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L622-L623). Gym builds the head server with [`uvicorn.Config(app, host, port)`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/server_utils.py#L793-L797), leaving `timeout_graceful_shutdown` unset, so uvicorn waits with `asyncio.wait_for(..., timeout=None)`. Its `force_exit` escape is set in exactly one place — inside `handle_exit`, which only `capture_signals` installs, and `capture_signals` returns early off the main thread. Gym runs the server as [`Thread(target=uvicorn_server.run, daemon=True)`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/server_utils.py#L800-L801), so `force_exit` stays `False` and one open connection keeps the join spinning.

No hang was observed here — what is proven from source is that the wait has no bound and `force_exit` cannot be set off the main thread. A trigger is plausible but was not run.

`ray.kill` is the only thing that stops an actor wedged this way: `NemoGym` has [`async def run_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L685), so Ray makes it an asyncio actor ([actor.py:1806](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L1806)) and wraps sync methods with [`sync_to_async`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/async_compat.py#L44-L54), a bare coroutine wrapper with no `run_in_executor`. A blocking `_spinup` owns the loop, so a queued `shutdown()` never starts.

The reap order is also why killing there is safe: `RunHelper.shutdown()` [reaps every subprocess first](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L586-L620) — [1s graceful + 2s reap](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L91-L93) each — and only then hits the wedging join, so by the time a finite timeout fires they are already dead.
</details>

<details><summary>Two existing tests pin the current behaviour and must change with the fix</summary>

[test_nemo_gym_utils.py:942](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/tests/unit/environments/test_nemo_gym_utils.py#L942) asserts `invocation.kwargs == {"timeout": None}`. [test_nemo_gym_utils.py:860](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/tests/unit/environments/test_nemo_gym_utils.py#L860) asserts `spinup_zero_calls[1] is None`, pinning the drain as untimed. No suggestion block for either: the right assertion depends on which shape you pick.

The fake in [that test](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/tests/unit/environments/test_nemo_gym_utils.py#L839-L861) is faithful for a ref that is slow but does complete — it returns on the re-get. It **cannot model a ref that never resolves**, which is the case here: its guard includes `timeout is not None`, so an untimed get always succeeds.
</details>

nemo_rl/environments/nemo_gym_shards.py
terrykong: **2 action items.**

**TL;DR — when a name is routable in one shard and a model engine in another, this `continue` skips it in *both*, so the second shard's model engine is never duplicate-checked and quietly claims GPUs a second time.**

The vocabulary first, because the bug lives in it. An entry is a named config block that declares which kinds of server it provides, and it can declare more than one. This is what a shard reports from [`list_entries()`](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L640):

```python
{"math_agent":   ["responses_api_agents"],                          # rollouts route here
 "math_env":     ["resources_servers"],                             # also routable, via task_source
 "judge_model":  ["responses_api_models"],                          # a model engine, holds GPUs
 "policy_model": ["responses_api_agents", "responses_api_models"]} # both at once
```

"Routable" is not a kind of thing — it is whether *this name, in this shard*, declares `responses_api_agents` or `resources_servers` ([L441](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym_shards.py#L441)). Shards load different `config_paths`, so the same name can declare different kinds in different shards, and nothing checks that it does not.

Loop 1 records every name that is routable in *some* shard. Loop 2 then skips anything in that record, for *every* shard. So a name that is routable in `judged` and only a model engine in `tools` gets skipped in both passes, and the `tools` copy never reaches the `allowed_duplicate_entries` check:

```python
build_route_shard_map({
  'judged': {'policy_model': ['responses_api_agents','responses_api_models']},
  'tools':  {'policy_model': ['responses_api_models']},
})
# -> {'policy_model': 'judged'}   no raise, no allowlist entry required
# 'tools' is now running a second policy_model model engine. Nobody is told.
```

That is the harm the function's own docstring says it prevents ([L429-L433](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym_shards.py#L429-L433)):

> a shared YAML dropped into two shards' path lists quietly brings its judge along and doubles that judge's GPU claim

Both cases traced line by line, with the loop variables at each step: [the duplicate check](https://terrykong.github.io/gh-pages-poc/terryk/pr-3371-duplicate-check.html).

### AI-1

Delete the skip. It cannot over-fire: a name routable in two shards already raised in loop 1, so loop 2 never sees one; a lone routable entry is simply recorded once and passes.

The suggestion block below is empty on purpose — it deletes the two lines.

```suggestion
```

### AI-2

Add a regression test for the collision to [test_nemo_gym_shards.py](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/tests/unit/environments/test_nemo_gym_shards.py). Today it fails with `DID NOT RAISE ShardSetupError`; it passes once the two lines above are deleted, so commit it together with that change rather than on its own. Not a one-click block — it targets a different file than this comment's anchor; paste it at the end of that file.

```python
def test_a_routable_name_is_still_a_duplicate_when_it_is_a_model_elsewhere():
    """Reusing a routable name for a model engine doubles that engine's GPU claim."""
    with pytest.raises(ShardSetupError, match="allowed_duplicate_entries"):
        build_route_shard_map(
            {
                "judged": {"math": ["responses_api_agents"]},
                "tools": {
                    "math": ["responses_api_models"],
                    "bash": ["responses_api_agents"],
                },
            }
        )
```

nemo_rl/environments/nemo_gym.py
```diff
@@ -1439,65 +1581,329 @@ def spinup_nemo_gym_actor(
...
+                f"config_paths is the usual cause: {error}"
+            ) from error
+
+    installs = [
+        (shard_name, replica, handle.set_tokenizer.remote(tokenizer))
+        for shard_name, replica, handle in instance_handles
+    ]
+    for shard_name, replica, reference in installs:
+        try:
+            ray.get(reference, timeout=max(0.0, deadline - monotonic()))
```

**TL;DR — when the startup budget runs out, this loop reports `shard 'X' (replica N) did not accept the tokenizer` — naming a shard and the tokenizer when neither is at fault and the real cause is that startup used the whole budget.**

Sharing one [`deadline`](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L1807) across the [spinup loop](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L1815) and this one is right: `spinup_timeout` should bound the whole job of bringing shards up, and installing the tokenizer is part of that. Two independent budgets would make the worst case twice `spinup_timeout` from a single knob. **Not asking to change that.**

What is wrong is only the message. On a cold node spinup finishes just under the budget — your own comment at [L107-L110](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L107-L110) says that is the expected case — so the remaining time is at or near zero, Ray's `timeout=0.0` raises at once, and the operator reads an error about the tokenizer. Nothing about the tokenizer is wrong; the budget simply ran out.

Background, with the numbers: [the shared deadline](https://terrykong.github.io/gh-pages-poc/terryk/pr-3371-shared-deadline.html).

### AI-1

Say the budget ran out when it ran out. Not a one-click block — the change spans the loop body and its `raise` at [L1835-L1842](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L1835-L1842), which is wider than this comment's anchor.

```python
    for shard_name, replica, reference in installs:
        remaining = max(0.0, deadline - monotonic())
        try:
            ray.get(reference, timeout=remaining)
        except BaseException as error:
            if remaining <= 0.0:
                raise ShardSetupError(
                    f"NeMo-Gym shards used the whole {spinup_timeout}s startup "
                    f"budget before the tokenizer reached every replica. Raise "
                    f"env.nemo_gym.spinup_timeout, or bake the Gym venvs into "
                    f"the image so a cold node starts faster."
                ) from error
            raise ShardSetupError(
                f"NeMo-Gym shard '{shard_name}' (replica {replica}) did not "
                f"accept the tokenizer: {error}"
            ) from error
```

It also fixes a smaller annoyance: finished installs still return at `timeout=0`, so the shard named today changes from run to run. A budget message names none.

<details><summary>Why an exhausted budget raises instantly rather than waiting</summary>

Ray maps `timeout=0.0` to an immediate `GetTimeoutError` rather than to "no timeout" — [worker.py L999-L1001](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/worker.py#L999-L1001) and the [`ray.get` docstring](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/_private/worker.py#L2895-L2899). So once the shared deadline is spent, every remaining install fails on the spot. That is correct for a spent budget; only the wording of the error misleads.

This branch also has no test coverage today: `accept the tokenizer` appears nowhere under `tests/`, and `_FakeGymCluster.ray_get` unconditionally returns `None` for tokenizer refs ([L598](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/tests/unit/environments/test_nemo_gym_utils.py#L598)), so neither it nor its teardown has ever run.
</details>

nemo_rl/environments/nemo_gym.py
```diff
@@ -1430,7 +1489,90 @@ def build_nemo_gym_config(
...
+        handles: Shard name to its replica handles, in replica order.
+        route_to_shard: Agent or task-source entry name to the shard hosting
+            it. Empty when unsharded, where every row goes to the only actor.
+        placement_group: The STRICT_SPREAD group pinning shards to distinct
+            nodes, or None when unsharded.
+    """
+
+    handles: Dict[str, List[Any]]
+    route_to_shard: Dict[str, str] = field(default_factory=dict)
+    placement_group: Any = None
```
terrykong: **1 action item.** Low.

**TL;DR — `placement_group: Any` hides that `is_sharded` is just a `None` check on a placement group, and the repo already has a concrete spelling for both of these fields.**

The established spellings are `list[ray.actor.ActorHandle]`, in [worker_groups.py L371](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/distributed/worker_groups.py#L371), and `Optional[list[PlacementGroup]]`, in [virtual_cluster.py L860](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/distributed/virtual_cluster.py#L860).

### AI-1

Not a one-click block: it also needs `PlacementGroup` added to the [ray import at L29](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L29), which is a different region than this anchor.

```python
    handles: Dict[str, List[ray.actor.ActorHandle]]
    route_to_shard: Dict[str, str] = field(default_factory=dict)
    placement_group: Optional[PlacementGroup] = None
```

**Context — no action.** This cannot turn CI red either way: `nemo_gym.py` is not in [`pyrefly.toml` `project-includes`](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/pyrefly.toml#L26).

nemo_rl/environments/nemo_gym_shards.py
terrykong: **1 action item.** Low.

**TL;DR — the two Gym server-type keys are written here as bare strings, while this module's stated convention is to give Gym's keys a named module constant.**

The strings themselves are right. It is the spelling that differs from the module's own rule ([L48-L51](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym_shards.py#L48-L51)):

```python
# Gym's log-directory key (``nemo_gym.global_config.NEMO_GYM_LOG_DIR_KEY_NAME``).
# Restated rather than imported: this module is imported on the driver, and
# nemo_gym is installed only in the actor's venv.
GYM_LOG_DIR_KEY = "nemo_gym_log_dir"
```

### AI-1

Name the routable set once at the top of the module and use it here. Not a one-click block — the declaration goes at the top, which is a different region than this anchor.

```python
# Gym's server-type keys that make an entry a rollout destination.
GYM_ROUTABLE_KEYS = frozenset({"responses_api_agents", "resources_servers"})
```

then here:

```python
            if not GYM_ROUTABLE_KEYS.intersection(types):
```

**Context — no action.** Do not import `GYM_SERVER_TYPE_KEYS` from `nemo_gym.py` instead — [`nemo_gym.py` already imports from this module](https://github.com/NVIDIA-NeMo/RL/blob/e123efadb31228121a98d60039fdc45d3ab0f239/nemo_rl/environments/nemo_gym.py#L55), so that would be an import cycle. Keep the constant local.

terrykong (commented):
This PR adds the actor-creation half of the shard feature: it expands a `ShardPlan` into one actor per shard instance, pins them to distinct nodes with a STRICT_SPREAD placement group, starts them all at once, installs the tokenizer, and then asks each shard what it spawned to build the route → shard map. It is the middle of a four-PR stack (#3369 schema → this → #3372 routing → #3373 wiring).

Several things are done well. The concurrent startup is right: every `_spinup.remote()` is dispatched before the first `ray.get`, so wall-clock is the slowest shard rather than the sum. The `build_route_shard_map` / `_discover_route_shard_map` split is a good seam — the pure half is unit-tested with no Ray at all. Treating one shard as just the one-shard case keeps callers on a single code path. And the `timeout` / `force_kill` split on `NemoGymShardSet.shutdown()` is the right shape for teardown; the comment below is only about the one call site that does not use it yet.

Reviewed at `e123efa`. An earlier pass was done at `917ffc3`, and every finding here was re-checked against the current head after the rebases and the `build_agent_shard_map` → `build_route_shard_map` rewrite. The duplicate-check finding was re-confirmed by running the rewritten function.

The sharded path is not reachable from any production entrypoint yet — all three call `spinup_nemo_gym_actor`, which raises on a `shards:` block — so this is reviewed as a building block rather than as shipping behaviour.

Two things deliberately not raised inline, so you know they were not missed:

1. `env.nemo_gym.shards` is documented nowhere in the stack — `docs/design-docs/nemo-gym-integration.md` enumerates the `env.nemo_gym` keys and lists none of the shard keys. That belongs on #3373, where the feature becomes reachable and the example config lands.
2. Findings about the shard schema itself, the per-shard port range, and `pyrefly` coverage of `nemo_gym_shards.py` belong to #3369.

_Generated by Claude Code_

nemo_rl/environments/nemo_gym.py
ananthsub: Legitimate; fixed both paths. Failed shard construction now calls `shutdown(timeout=NEMO_GYM_GRACEFUL_SHUTDOWN_TIMEOUT_S, force_kill=True)`. The startup drain uses one shared finite grace deadline and catches only `Exception`, so a permanently wedged `_spinup` cannot prevent teardown and `KeyboardInterrupt` is not swallowed. Tests cover a recoverable timeout and a startup that remains wedged through the drain.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Legitimate; fixed. The second pass no longer skips a name merely because it is routable somewhere. It now duplicate-checks that name in every shard, so a routable entry in one shard colliding with a model-only entry in another requires an explicit allowlist entry. Added the proposed regression case.

nemo_rl/environments/nemo_gym.py
ananthsub: Legitimate; fixed. The tokenizer loop records the remaining shared startup budget before `ray.get`. If that budget is already exhausted and the get times out, the error now reports that the overall startup budget was consumed and suggests increasing `env.nemo_gym.spinup_timeout` or baking venvs into the image. A focused test pins this distinction.

nemo_rl/environments/nemo_gym.py
ananthsub: Legitimate; fixed. `handles` is now typed as `Dict[str, List[ray.actor.ActorHandle]]`, `placement_group` as `Optional[PlacementGroup]`, and `PlacementGroup` is imported from Ray’s placement-group module.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Legitimate; fixed. Added the local `GYM_ROUTABLE_KEYS` frozenset and use it for route classification, avoiding an import cycle with `nemo_gym.py`.

terrykong (approved):

PR #2530, opened 2026-05-19
feat: PPO with MCore

bg51717:
# What does this PR do ?

Adds full Proximal Policy Optimization (PPO) support to NeMo-RL with an actor-critic architecture, using the **Megatron-Core (mcore) backend** for both the policy and value models. The policy (actor) and value function (critic) are jointly trained using Generalized Advantage Estimation (GAE). Both models run on Megatron-Core with GPU/CPU offloading for colocated execution on the same set of GPUs as vLLM generation.

> **Backend support:** This PR implements PPO on the **Megatron-Core backend only**. The DTensor/FSDP2 backend is **not yet supported** for the value model. All recipes and tests use `megatron_cfg.enabled: true`.
# Issues
close #2048 

# Summary of Changes

## PPO Training Algorithm
- Complete PPO training loop with critic-before-actor update order
- Multiple training steps per rollout (`steps_per_epoch`)
- Configurable critic warmup (`policy_training_start_step`) — trains value model alone before policy updates begin
- Dynamic sampling support
- Colocated architecture with GPU memory management via model offloading

## Generalized Advantage Estimation (GAE)
- Token-level GAE with carry-forward masking for correct multi-turn/padding handling
- Token-level KL penalty in rewards (configurable coefficient and KL type: k1/k3)
- VAPO decoupled GAE: separate lambda for value returns vs. policy advantages
- Length-adaptive lambda: `lambda_policy = 1 - 1/(alpha * response_length)`
- Reward whitening

## Value Model (Megatron-Core backend)
- LM backbone + scalar value head, reusing Megatron-Core policy infrastructure
- Supports TP/PP/DP parallelism, distributed optimizer, sequence packing
- GPU/CPU offloading for colocated execution with policy and vLLM
- Full checkpoint save/load including value head weights
- Clipped MSE value loss with configurable loss scale and clip range
- VAPO NLL auxiliary loss on correct samples

## Shared Algorithm Improvements
- Refactored clipped PG loss to support both GRPO and PPO
- Added Reinforce++ and raw-reward advantage estimators
- GSM8K answer extraction and verification environment

## Configuration and Recipes
- Base config: `examples/configs/ppo_math_1B_megatron.yaml` (DAPO-style PPO: no KL penalty, asymmetric clipping, dual-clip, reward scaling)
- `ppo-dsr1-7b-math-8n8g-megatron` — DeepSeek-R1-7B on DAPOMath-17K, 8 nodes, KL penalty + importance sampling
[...]

PR #3372, opened 2026-07-27
feat(nemo-gym): route rollouts across shard actors

ananthsub:
## Summary

Routes NeMo-Gym rollout work across the actors in a `NemoGymShardSet` and merges their streamed results into the existing accumulator.

- Assigns batch-global row indices before dispatch so results can be restored to input order.
- Routes complete prompt groups by `agent_ref` and rejects groups that mix agents.
- Keeps synchronous accumulation size separate from the prompt-group size used for routing.
- Selects replicas round-robin without splitting one prompt group across actors.
- Keeps `NemoGymShardSet` serializable across Ray boundaries by excluding its process-local replica lock and recreating the lock after deserialization.
- Merges multiple Ray streams while preserving row identity and attributing stream failures to the selected shard.
- Reports shard timing metrics and uses the maximum duration for concurrent wall-clock rollups.
- Retries only prompt groups that have not reached the replay buffer. A `RayActorError` in the exception chain stops retry immediately.

A bare actor handle remains the one-shard, one-replica compatibility case, so existing callers still produce one bucket and one stream.

## Dependency

This PR targets the branch from [PR #3371](https://github.com/NVIDIA-NeMo/RL/pull/3371). [PR #3373](https://github.com/NVIDIA-NeMo/RL/pull/3373) builds on this branch.

### DIFF
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -826,6 +826,7 @@ class NemoGymShardSet:
     handles: Dict[str, List[Any]]
     agent_to_shard: Dict[str, str] = field(default_factory=dict)
     placement_group: Any = None
+    _next_replica: Dict[str, int] = field(default_factory=dict, repr=False)
 
     @property
     def is_sharded(self) -> bool:
@@ -834,6 +835,58 @@ class NemoGymShardSet:
     @property
     def all_handles(self) -> List[Any]:
         return [handle for replicas in self.handles.values() for handle in replicas]
+
+    @property
+    def hosted_agents(self) -> frozenset[str]:
+        """Agent entry names this set can route to."""
+        return frozenset(self.agent_to_shard)
+
+    def shard_for_agent(self, agent_name: str) -> str:
+        """Name the shard hosting an agent.
+
+        Unsharded jobs have one actor and no map, so every agent resolves to
+        it; nothing was discovered because nothing could have conflicted.
+
+        Raises:
+            ShardSetupError: No shard hosts the agent, so its rows have nowhere
+                to go. Startup validates the datasets against this same map, so
+                reaching it here means the row named an agent the datasets did
+                not declare.
+        """
+        if not self.agent_to_shard:
+            return next(iter(self.handles))
+        try:
+            return self.agent_to_shard[agent_name]
+        except KeyError:
+            raise ShardSetupError(
+                f"No NeMo-Gym shard hosts agent '{agent_name}'. Hosted agents: "
+                f"{sorted(self.agent_to_shard)}."
+            ) from None
+
+    def pick_handle(self, agent_name: str) -> Any:
+        """Choose the actor instance to serve an agent's next prompt group.
+
+        The shard is fixed by the data; the replica rotates round-robin. Round
+        robin is deterministic and easy to reason about, which matters more
+        than adaptivity here: within a synchronous step there is no completion
+        feedback to adapt on, so an even split is the best available policy.
+        Least-in-flight would pay off on the async path, where dispatch is
+        continuous, and can replace this without touching callers.
+        """
+        shard_name = self.shard_for_agent(agent_name)
+        replicas = self.handles[shard_name]
+        if len(replicas) == 1:
+            return replicas[0]
+        index = self._next_replica.get(shard_name, 0)
+        self._next_replica[shard_name] = (index + 1) % len(replicas)
+        return replicas[index]
+
+    def shard_name_of(self, handle: Any) -> str:
+        """Reverse-look up a handle's shard, for error messages and metric tags."""
+        for shard_name, replicas in self.handles.items():
+            if any(replica is handle for replica in replicas):
+                return shard_name
+        raise ShardSetupError("Handle does not belong to this NeMo-Gym shard set")
 
     def sole_handle(self) -> Any:
         """The only actor, for callers that predate routing.
@@ -873,6 +926,19 @@ def _shard_instances(plan: ShardPlan) -> List[tuple[ShardSpec, int]]:
     return [
         (shard, replica) for shard in plan.shards for replica in range(shard.replicas)
     ]
+
+
+def as_nemo_gym_shard_set(environment: Any) -> NemoGymShardSet:
+    """Read the NeMo-Gym entry of ``task_to_env`` as a shard set either way.
+
+    Call sites that predate sharding put a bare actor handle there. Rather than
+    make every one of them build a set first, treat a lone handle as the
+    one-shard, one-replica case it already is, so the routing path is identical
+    whether or not the job is sharded.
+    """
+    if isinstance(environment, NemoGymShardSet):
+        return environment
+    return NemoGymShardSet(handles={DEFAULT_SHARD_NAME: [environment]})
 
 
 def build_nemo_gym_actors(
--- a/nemo_rl/environments/utils.py
+++ b/nemo_rl/environments/utils.py
@@ -157,6 +157,10 @@ def shutdown_environments(
     in-flight HTTP requests to the vLLM endpoints, and killing generation first
     leaves them retrying dead connections.
 
+    An entry is either a Ray actor handle or a local object that owns actors of
+    its own (a NeMo-Gym shard set); the latter is asked to shut itself down
+    rather than killed, since only it knows what it holds.
+
     Args:
         env_maps: Task-name to environment mappings. ``None`` entries are skipped.
         timeout: Seconds to wait for each actor's ``shutdown()`` before killing it.
@@ -170,10 +174,20 @@ def shutdown_environments(
                 continue
             seen.add(id(env))
             print(f"🛑 Shutting down environment {task_name}...")
+            # An actor handle offers shutdown as a remote call; an owner object
+            # offers it as a plain method. Nothing else distinguishes them here.
+            remote_shutdown = getattr(getattr(env, "shutdown", None), "remote", None)
             try:
-                ray.get(env.shutdown.remote(), timeout=timeout)
+                if remote_shutdown is not None:
+                    ray.get(remote_shutdown(), timeout=timeout)
+                else:
+                    env.shutdown()
             except Exception as e:
                 print(f"Graceful shutdown of environment {task_name} failed: {e}")
+                if remote_shutdown is None:
+                    # Nothing to kill: the failure came from an owner object
+                    # that already tried, and reported, its own teardown.
+                    continue
                 try:
                     ray.kill(env)
                 except Exception as kill_error:
--- a/nemo_rl/experience/rollouts.py
+++ b/nemo_rl/experience/rollouts.py
@@ -46,7 +46,11 @@ from nemo_rl.environments.interfaces import (
     EnvironmentInterface,
     EnvironmentReturn,
 )
-from nemo_rl.environments.nemo_gym import DEFAULT_THINKING_TAGS
+from nemo_rl.environments.nemo_gym import (
+    DEFAULT_THINKING_TAGS,
+    NemoGymShardSet,
+    as_nemo_gym_shard_set,
+)
 from nemo_rl.experience.interfaces import NEMO_GYM_TASK_INDEX_KEY
 from nemo_rl.experience.metric_utils import calculate_single_metric, pct
 from nemo_rl.models.generation.interfaces import (
@@ -1969,6 +1973,123 @@ def _prepare_nemo_gym_rows(
         row["_rowidx"] = row_index
 
 
+def _bucket_nemo_gym_rows_by_shard(
+    rows: list[dict],
+    shard_set: NemoGymShardSet,
+    num_generations: int,
+) -> list[tuple[str, Any, list[dict]]]:
+    """Split already-stamped rows into one bucket per actor, by prompt group.
+
+    Bucketing is by group and never by row. A group split across shards would
+    deadlock any verifier that scores rollouts against their peers: those
+    buffer a whole group in one process and block until it is complete, so each
+    half would wait forever for rollouts sitting in the other.
+
+    Rows keep the global ``_rowidx`` stamped over the whole batch, so the
+    accumulator downstream never learns that shards exist.
+
+    Returns:
+        ``(shard_name, handle, rows)`` per actor that has work, in first-use
+        order. Actors with no rows are omitted rather than sent an empty batch.
+    """
+    buckets: dict[int, tuple[str, Any, list[dict]]] = {}
+    for start in range(0, len(rows), num_generations):
+        group = rows[start : start + num_generations]
+        agent_names = {row["agent_ref"]["name"] for row in group}
+        shard_names = {shard_set.shard_for_agent(name) for name in agent_names}
+        if len(shard_names) != 1:
+            raise ValueError(
+                f"NeMo-Gym prompt group at row {start} spans shards "
+                f"{sorted(shard_names)} via agents {sorted(agent_names)}. A "
+                f"group must be served by one shard."
+            )
+        handle = shard_set.pick_handle(next(iter(agent_names)))
+        if id(handle) not in buckets:
+            buckets[id(handle)] = (shard_names.pop(), handle, [])
+        buckets[id(handle)][2].extend(group)
+    return list(buckets.values())
+
+
+async def _merge_nemo_gym_shard_streams(
+    buckets: list[tuple[str, Any, list[dict]]],
+    tokenizer: TokenizerType,
+    timer_prefix: str,
+) -> AsyncGenerator[tuple[int, dict, dict | None, str], None]:
+    """Interleave K actor streams into one, yielding rows as they complete.
+
+    ``run_rollouts`` is a streaming generator, so there is nothing to gather:
+    this replaces one stream with a merge of K and hands the merged rows to the
+    accumulator that already exists. With a single bucket it degenerates to
+    that one stream, which is what keeps the unsharded path unchanged.
+
+    Yields:
+        ``(row_index, result, timing_metrics, shard_name)``. ``timing_metrics``
+        is non-None only on the last row of each shard's bucket.
+    """
+    iterators = {}
+    for shard_name, handle, rows in buckets:
+        stream = handle.run_rollouts.options(num_returns="streaming").remote(
+            rows, tokenizer, timer_prefix
+        )
+        iterators[stream.__aiter__()] = shard_name
+
+    pending = {
+        asyncio.ensure_future(anext(iterator)): iterator for iterator in iterators
+    }
+    try:
+        while pending:
+            done, _ = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
+            for task in done:
+                iterator = pending.pop(task)
+                shard_name = iterators[iterator]
+                try:
+                    future = task.result()
+                except StopAsyncIteration:
+                    continue
+                except Exception as error:
+                    raise RuntimeError(
+                        f"NeMo-Gym shard '{shard_name}' failed during rollout "
+                        f"collection: {error}"
+                    ) from error
+                row_index, result, timing_metrics = await future
+                yield row_index, result, timing_metrics, shard_name
+                pending[asyncio.ensure_future(anext(iterator))] = iterator
+    finally:
+        # Leaving the merge early (an error downstream, or the consumer closing
+        # the generator) must not strand the anext tasks still in flight.
+        for task in pending:
+            task.cancel()
+
+
+def _merge_nemo_gym_timing_metrics(
+    timing_by_shard: dict[str, dict[str, Any]], timer_prefix: str
+) -> dict[str, Any]:
+    """Combine per-shard actor timings into per-shard plus rolled-up metrics.
+
+    Each shard reports the same key names, so the roll-up takes the maximum
+    rather than the sum: shards run concurrently and the step is gated by the
+    slowest one, not by their total. That also means the rolled-up
+    ``postprocess_results_pct`` is the worst shard's share rather than a
+    batch-wide average, which is the honest reading -- averaging percentages
+    over shards with different row counts would not mean anything.
+
+    With one shard this returns exactly what that shard reported, so the
+    unsharded path is unchanged.
+    """
+    if len(timing_by_shard) <= 1:
+        return dict(next(iter(timing_by_shard.values()), {}))
+
+    merged: dict[str, Any] = {}
+    for shard_name, metrics in timing_by_shard.items():
+        for key, value in metrics.items():
+            suffix = (
+                key[len(timer_prefix) + 1 :] if key.startswith(timer_prefix) else key
+            )
+            merged[f"{timer_prefix}/shard/{shard_name}/{suffix}"] = value
+            merged[key] = max(merged[key], value) if key in merged else value
+    return merged
+
+
 def _tensorize_nemo_gym_result(result: dict) -> None:
     """Convert token fields returned by the Gym actor back to tensors."""
     _tensorize_by_key(result["input_message_log"], "token_ids")
@@ -2106,13 +2227,15 @@ async def run_async_nemo_gym_rollout(
             allow_mixed_agents=returns_entire_batch,
         )
         final_rollout_result: NemoGymRolloutResult | None = None
-        actor_timing_metrics: dict[str, Any] = {}
-        nemo_gym_environment = task_to_env["nemo_gym"]
+        actor_timing_by_shard: dict[str, dict[str, Any]] = {}
+        shard_set = as_nemo_gym_shard_set(task_to_env["nemo_gym"])
         with timer.time(run_rollouts_timer_label):
-            rollout_gen = nemo_gym_environment.run_rollouts.options(
-                num_returns="streaming"
-            ).remote(nemo_gym_rows, tokenizer, timer_prefix)
-        rollout_iterator = rollout_gen.__aiter__()
+            buckets = _bucket_nemo_gym_rows_by_shard(
+                nemo_gym_rows, shard_set, num_generations
+            )
+            rollout_iterator = _merge_nemo_gym_shard_streams(
+                buckets, tokenizer, timer_prefix
+            )
 
     while True:
         stream_finished = False
@@ -2120,15 +2243,15 @@ async def run_async_nemo_gym_rollout(
         with timer.time(total_timer_label):
             with timer.time(run_rollouts_timer_label):
                 try:
-                    future = await anext(rollout_iterator)
+                    rowidx, result, timing_metrics, shard_name = await anext(
+                        rollout_iterator
+                    )
                 except StopAsyncIteration:
                     stream_finished = True
-                else:
-                    rowidx, result, timing_metrics = await future
 
             if not stream_finished:
                 if timing_metrics is not None:
-                    actor_timing_metrics = timing_metrics
+                    actor_timing_by_shard[shard_name] = timing_metrics
 
                 _tensorize_nemo_gym_result(result)
                 completed_group = accumulator.add(rowidx, result)
@@ -2166,7 +2289,9 @@ async def run_async_nemo_gym_rollout(
                 "NeMo-Gym completed without producing a final prompt group"
             )
 
-    final_rollout_result.rollout_metrics.update(actor_timing_metrics)
+    final_rollout_result.rollout_metrics.update(
+        _merge_nemo_gym_timing_metrics(actor_timing_by_shard, timer_prefix)
+    )
     final_rollout_result.rollout_metrics.update(timer.get_timing_metrics("sum"))
     yield final_rollout_result
 
--- a/tests/unit/environments/test_environment_utils.py
+++ b/tests/unit/environments/test_environment_utils.py
@@ -108,6 +108,46 @@ def test_shutdown_environments_kills_only_on_failure():
     mock_ray.kill.assert_called_once_with(wedged)
 
 
+class _FakeOwner:
+    """An environment that owns actors rather than being one.
+
+    Written out rather than mocked because the distinction under test is
+    precisely that ``shutdown`` is a plain method with no ``.remote``, and a
+    MagicMock grows a ``.remote`` on demand.
+    """
+
+    def __init__(self, error=None):
+        self.error = error
+        self.shutdown_calls = 0
+
+    def shutdown(self):
+        self.shutdown_calls += 1
+        if self.error is not None:
+            raise self.error
+
+
+def test_an_owner_object_shuts_itself_down_instead_of_being_killed():
+    """A shard set holds its own actors, so only it can tear them down."""
+    shard_set = _FakeOwner()
+
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        shutdown_environments({"nemo_gym": shard_set})
+
+    assert shard_set.shutdown_calls == 1
+    mock_ray.get.assert_not_called()
+    mock_ray.kill.assert_not_called()
+
+
+def test_a_failed_owner_teardown_is_not_escalated_to_a_kill():
+    """There is no single handle to kill, and the owner already reported why."""
+    shard_set = _FakeOwner(error=RuntimeError("a shard would not stop"))
+
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        shutdown_environments({"nemo_gym": shard_set})
+
+    mock_ray.kill.assert_not_called()
+
+
 def test_shutdown_environments_tolerates_empty_input():
     with patch("nemo_rl.environments.utils.ray") as mock_ray:
         shutdown_environments(None, {})
--- a/tests/unit/experience/test_nemo_gym_router.py
+++ b/tests/unit/experience/test_nemo_gym_router.py
@@ -1,0 +1,225 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Routing of NeMo-Gym rollout rows across sharded actors.
+
+These cover the pure dispatch layer -- bucketing, the K-stream merge, and the
+metrics roll-up -- with fake actors, so they need neither Ray nor a GPU.
+"""
+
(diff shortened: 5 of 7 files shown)


nemo_rl/algorithms/ppo.py
```diff
@@ -1467,6 +1467,9 @@ def ppo_train(
                             ),
                             max_rollout_turns=None,
                             greedy=False,
+                            num_generations_per_prompt=master_config.ppo[
+                                "num_generations_per_prompt"
+                            ],
```
terrykong: **1 action item.**

**TL;DR — `master_config.ppo["num_generations_per_prompt"]` subscripts a pydantic model, so every sync PPO + NeMo-Gym run raises `TypeError` on its first rollout.**

PR-introduced regression. [`PPOConfig`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/algorithms/ppo.py#L168) is a pydantic `BaseModel`, and [`MasterConfig.ppo`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/algorithms/ppo.py#L302) is typed as one, so the object built at [`run_ppo.py:111`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/examples/run_ppo.py#L111) has no instance `__getitem__` — pydantic v2 defines only `__class_getitem__`. The line sits inside [`if should_use_nemo_gym(...)`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/algorithms/ppo.py#L1455), and NeMo-Gym is rejected only for *async* PPO at [`run_ppo.py:82`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/examples/run_ppo.py#L82), so sync PPO reaches it unguarded and dies with `TypeError: 'PPOConfig' object is not subscriptable` before any rollout is dispatched.

It is a regression, not a new path: this file's diff is +3/−0, and the PPO NeMo-Gym branch has been in the tree since #2530. Neither #3373 nor #3374 touches this file.

### AI-1

```suggestion
                            num_generations_per_prompt=(
                                master_config.ppo.num_generations_per_prompt
                            ),
```

<details><summary>Why CI stays green on this</summary>

The sync PPO test mocks the flag off at [`test_ppo.py:919`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/tests/unit/algorithms/test_ppo.py#L919), and every shipped `examples/nemo_gym/*.yaml` recipe is GRPO or distillation — none is PPO — so this line has never executed. Every other read of `master_config.ppo` in this file uses attribute access, and the sibling call sites #3373 adds use `master_config.grpo.num_generations_per_prompt`.
</details>

nemo_rl/environments/utils.py
```diff
@@ -171,10 +175,20 @@ def shutdown_environments(
...
             print(f"🛑 Shutting down environment {task_name}...")
+            # An actor handle offers shutdown as a remote call; an owner object
+            # offers it as a plain method. Nothing else distinguishes them here.
+            remote_shutdown = getattr(getattr(env, "shutdown", None), "remote", None)
             try:
-                ray.get(env.shutdown.remote(), timeout=timeout)
+                if remote_shutdown is not None:
+                    ray.get(remote_shutdown(), timeout=timeout)
+                else:
+                    env.shutdown()
```
terrykong: **1 action item.**

**TL;DR — the new owner-object branch ignores `timeout`, so the teardown bound this function documents reaches nothing once shards are wired up.**

The branch is new here; the `timeout=None` it collides with is pre-existing (#3371). The docstring promises `timeout` is [`seconds to wait before killing`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/environments/utils.py#L166), and the actor branch one line above honours it with [`ray.get(..., timeout=timeout)`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/environments/utils.py#L183). This branch calls a blocking method with no bound at all.

Downstream the bound disappears completely: the owner object on the sharded path is `NemoGymShardSet`, whose `shutdown` re-enters this function with [`timeout=None`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/environments/nemo_gym.py#L1610), so the default passed at [`grpo.py:4079`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/algorithms/grpo.py#L4079) never applies. Teardown runs in a `finally`, so a wedged actor would replace a real training error with a hang, and the new [`continue`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/environments/utils.py#L191) skips the `ray.kill` that would otherwise break out.

Scope: this only bites once #3373 puts a `NemoGymShardSet` into `task_to_env` — in this PR alone the branch is not reached, and nothing upstack fixes it. The hang is read off `ray.get(timeout=None)` rather than observed.

### AI-1

Thread `timeout` through to the owner call, or drop `timeout=None` at [`nemo_gym.py:1610`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/environments/nemo_gym.py#L1610) so the shard set inherits the bounded default. Not a suggestion block: the one-token fix is on a line in another file that this PR does not touch, so it cannot be committed from here.

nemo_rl/experience/rollouts.py
```diff
@@ -2702,6 +2861,7 @@ def run_nemo_gym_rollout_sync(
     effort_config: Optional[EffortLevelsConfig] = None,
     reward_penalty_config: dict[str, Any] | BaseModel | None = None,
     thinking_tags: list[str] | tuple[str, ...] | None = None,
+    num_generations_per_prompt: int = 1,
```
terrykong: **1 action item.**

**TL;DR — this parameter defaults to `1`, and `1` is the only value the validation can never reject even though it scatters one prompt's generations across different actors.**

PR-introduced, and no live bug today. The value feeds `routing_group_size`, which is checked only for divisibility at [`rollouts.py:2743`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2743) — and 1 divides everything. At 1, [`_bucket_nemo_gym_rows_by_shard`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2430) treats every row as its own group and hands them out round-robin, which is the split its own docstring says [`would deadlock a peer-scoring verifier`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2438). Concretely, with two replicas of one shard and a 4-generation prompt, rows 0 and 2 go to one replica and rows 1 and 3 to the other.

Nothing is broken right now: this PR wires two call sites, #3373 wires four more, and the one caller left on the default — [`run_grpo_nemo_gym.py:94`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/examples/nemo_gym/run_grpo_nemo_gym.py#L94) — collects one row per prompt, where 1 is correct. That is exactly one call site relying on the default, so removing it costs one line.

### AI-1

Make `num_generations_per_prompt` keyword-only and required, and pass `num_generations_per_prompt=1` explicitly at the example call site. Not a suggestion block: the change spans two files.

tests/unit/experience/test_nemo_gym_router.py
```diff
@@ -0,0 +1,290 @@
...
+
+    buckets = _bucket_nemo_gym_rows_by_shard(rows, shard_set, num_generations=2)
+
+    assert len(buckets) == 2
+    dispatched = {id(handle): rows for _, handle, rows in buckets}
+    assert [row["_rowidx"] for row in dispatched[id(first)]] == [0, 1, 4, 5]
+    assert [row["_rowidx"] for row in dispatched[id(second)]] == [2, 3, 6, 7]
+
+
+def test_a_group_that_mixes_agents_across_shards_is_rejected():
```
terrykong: **1 action item.**

**TL;DR — this test's name says it covers a group spanning two shards, but it stops at the same-route guard, so the cross-shard branch it is named for never runs.**

PR-introduced. This test and [`its sibling at L195`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/tests/unit/experience/test_nemo_gym_router.py#L195) both put two different route names in one group and both assert `match="mixes routes"`, so both stop at [`the first guard`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2455). The [`spans shards raise`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2461) cannot be reached from here anyway: past the first guard a group carries one route name, so it maps to one shard. The two-shard topology this test sets up is never consulted.

### AI-1

```suggestion
def test_a_group_naming_two_routes_is_rejected():
```

With the name corrected, [`the sibling at L195`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/tests/unit/experience/test_nemo_gym_router.py#L195) is a duplicate and can go.

tests/unit/environments/test_environment_utils.py
```diff
@@ -121,6 +121,46 @@ def test_shutdown_environments_continues_after_a_failed_kill():
...
+
+
+def test_a_failed_owner_teardown_is_not_escalated_to_a_kill():
+    """There is no single handle to kill, and the owner already reported why."""
+    shard_set = _FakeOwner(error=RuntimeError("a shard would not stop"))
+
+    with patch("nemo_rl.environments.utils.ray") as mock_ray:
+        shutdown_environments({"nemo_gym": shard_set})
+
+    mock_ray.kill.assert_not_called()
```
terrykong: **1 action item.**

**TL;DR — this test asserts only that nothing was killed, so it would stay green even if `shutdown_environments` stopped calling owner objects entirely.**

PR-introduced. The failing-owner case never checks that the owner's own `shutdown` ran, so the behaviour the test is named for is not pinned. Its sibling gets this right at [`L149`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/tests/unit/environments/test_environment_utils.py#L149).

### AI-1

```suggestion
    assert shard_set.shutdown_calls == 1
    mock_ray.kill.assert_not_called()
```

nemo_rl/algorithms/ppo.py
```diff
@@ -1467,6 +1467,9 @@ def ppo_train(
                             ),
                             max_rollout_turns=None,
                             greedy=False,
+                            num_generations_per_prompt=master_config.ppo[
```
macandro96: **1 action item.** PR-introduced.

`master_config.ppo` is a pydantic `BaseModel`, not a TypedDict, so this subscript raises at argument-evaluation time — before `run_nemo_gym_rollout_sync` is even entered.

Reproduced against the installed pydantic 2.13.4:
```
>>> PPOConfig()["num_generations_per_prompt"]
TypeError: 'PPOConfig' object is not subscriptable
```

[`PPOConfig`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/algorithms/ppo.py#L168) is `class PPOConfig(BaseModel, extra="allow")`. Ninety-one lines up, [`ppo.py:1380`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/algorithms/ppo.py#L1380) reads the same value correctly as `master_config.ppo.num_generations_per_prompt` — this is the only subscript among the file's `master_config.ppo` reads.

No shipped recipe enables Gym for PPO ([`test_ppo.py:919`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/tests/unit/algorithms/test_ppo.py#L919) forces `should_use_nemo_gym` to `False`, and there is no `ppo-*-gym` suite), so CI stays green — but this is live code and crashes the first time the branch is exercised.

### AI-1
```suggestion
                            num_generations_per_prompt=master_config.ppo.num_generations_per_prompt,
```

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -122,6 +122,18 @@ def _unanimous_task_index(rows: list[Any]) -> Optional[int]:
     return int(ordinal) if ordinal is not None else None
 
 
+def _caused_by_ray_actor_death(error: BaseException) -> bool:
+    """Return whether an exception chain contains a terminal actor failure."""
+    current: BaseException | None = error
+    seen: set[int] = set()
+    while current is not None and id(current) not in seen:
+        if isinstance(current, ray.exceptions.RayActorError):
```
macandro96: **1 action item.** PR-introduced.

**TL;DR — `ActorUnavailableError` subclasses `RayActorError`, so a *transient* actor blip breaks out of the Gym retry loop that exists to absorb exactly that.**

`_caused_by_ray_actor_death` is documented as detecting a "terminal actor failure", but it matches the parent class. Verified against the installed Ray 2.56.1:

```
ActorUnavailableError.__mro__ -> (ActorUnavailableError, RayActorError, RayError, Exception, ...)
```

Ray documents `ActorUnavailableError` as "the actor is temporarily unavailable but may be available later."

What breaks:
1. A node hiccups; Ray raises `ActorUnavailableError` out of the Gym stream.
2. It reaches [`:1930`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1930) — `if _caused_by_ray_actor_death(last_error): break`.
3. The loop exits after attempt 1, discarding the remaining `_MAX_NEMO_GYM_STREAM_RETRIES` and their exponential backoff — the mechanism built for transient unavailability.

This also contradicts the repo's own taxonomy: [`failures.py:148-149`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/failures.py#L148) lists **both** `RayActorError` and `ActorUnavailableError` in `_INFRA_TYPES`, i.e. retryable infra. The same class means "retry me" there and "give up" here.

### AI-1
```suggestion
        if isinstance(current, ray.exceptions.ActorDiedError):
```

nemo_rl/experience/rollouts.py
macandro96: **1 action item.** PR-introduced.

**TL;DR — the `= 1` default is forwarded as `routing_group_size`, so any caller that forgets the argument silently gets per-row routing and splits prompt groups across actors.**

This value is passed straight through as [`routing_group_size`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2924). At the default of 1, routing is per row rather than per group.

Breaking case: a 4-generation prompt group with two instances routes rows `[0, 2]` to one and `[1, 3]` to the other. That is precisely the hazard [this module's own docstring](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2437) warns about — a peer-scoring verifier buffers a whole cohort in one process and blocks until it is complete, so each half waits forever on rollouts sitting in the other.

Two guidelines bear on it: config-conventions forbids "a function parameter defaulting a config value", and error-handling's fail-loud rule means a forgotten argument should raise rather than silently re-route.

Latent today — it only bites with more than one instance, which `spinup_nemo_gym_actor` still refuses — which is exactly why it is cheap now and hard to spot after #3373 lands.

Note the signature has no `*` marker, so the parameter cannot simply be made required in place (non-default after default is a `SyntaxError`). Defaulting to `None` and raising in the body keeps it fail-loud without reordering:

### AI-1
```suggestion
    num_generations_per_prompt: Optional[int] = None,
```
…with a guard at the top of the body raising if it is `None`. Callers that genuinely want per-row routing then say `num_generations_per_prompt=1` explicitly.

nemo_rl/environments/nemo_gym.py
```diff
@@ -1516,6 +1533,56 @@ def is_sharded(self) -> bool:
...
+        than adaptivity here: within a synchronous step there is no completion
+        feedback to adapt on, so an even split is the best available policy.
+        Least-in-flight would pay off on the async path, where dispatch is
+        continuous, and can replace this without touching callers.
+        """
+        shard_name = self.shard_for_route(route_name)
+        replicas = self.handles[shard_name]
+        if len(replicas) == 1:
+            return replicas[0]
+        index = self._next_replica.get(shard_name, 0)
```
macandro96: **1 action item.** PR-introduced.

The round-robin cursor is a read-modify-write on shared mutable state with no lock:

```python
index = self._next_replica.get(shard_name, 0)
self._next_replica[shard_name] = (index + 1) % len(replicas)
```

The async path reaches this from multiple OS threads: `AsyncTrajectoryCollector` starts one [`threading.Thread` per batch](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/algorithms/async_utils/trajectory_collector.py#L978) (`rollout-batch-target-{n}`), each running its own `asyncio.run` loop, and each reaches [`rollout_manager.py:1125`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollout_manager.py#L1125) → `pick_handle` on the same `NemoGymShardSet`. There is no `await` between the two statements, but separate OS threads can be preempted between bytecodes regardless.

Two threads then read the same `index`, return the same replica, and one increment is lost. The impact is scheduling skew rather than misrouting — but the docstring's "deterministic" round robin then isn't.

The same collector already guards shared state with about ten `threading.Lock`s (`_pending_lock`, `_outstanding_lock`, `_threads_lock`, …); this cursor is the outlier.

### AI-1
Take a small lock around the cursor update. Latent until `replicas > 1`, and there is no concurrent-selection test — worth adding one alongside the fix.

nemo_rl/experience/rollouts.py
```diff
@@ -2424,6 +2427,164 @@ def _prepare_nemo_gym_rows(
...
+    finally:
+        # Leaving the merge early (an error downstream, or the consumer closing
+        # the generator) must not strand the anext tasks still in flight.
+        for task in pending:
+            task.cancel()
+        if pending:
+            await asyncio.gather(*pending, return_exceptions=True)
+        for iterator in active_iterators:
+            stream = streams_by_iterator[iterator]
+            if isinstance(stream, ray.ObjectRefGenerator):
```
macandro96, line 2555: **1 action item, non-blocking.** PR-introduced.

**TL;DR — the first stream failure cancels every healthy stream, so one bad shard discards in-flight work from all the others; and because `ray.cancel` never reaches Gym's HTTP tasks, that work keeps running server-side anyway.**

The merge raises on the first failed stream and this `finally` then cancels every remaining generator. Correctness holds — completed groups are buffered and missing ones retried — but failure isolation is weaker than the resource isolation sharding implies: a failure in `tools/1` discards in-flight rollouts from `math/0`.

The cancel is also only partly effective. `NemoGym.run_rollouts` iterates a generator built by Gym's `run_examples` → `_run_examples_with_metadata`, which ends in `tqdm.as_completed(map(_post_subroutine, examples))`. `asyncio.as_completed` consumes its argument into a set and `ensure_future`s every coroutine, so all the HTTP POSTs to Gym (and thence vLLM) are already scheduled. `ray.cancel` cancels the outer `run_rollouts` coroutine, not those tasks or their sockets — so the discarded work still occupies the generation pool the trainer believes is idle, and the retry pays for it a second time.

### AI-1
Consider recording the first failure, letting the healthy streams drain, and raising once they finish. That preserves the error while cutting the duplicated rollout work.

nemo_rl/environments/utils.py
```diff
@@ -171,10 +175,20 @@ def shutdown_environments(
                 continue
             seen.add(id(env))
             print(f"🛑 Shutting down environment {task_name}...")
+            # An actor handle offers shutdown as a remote call; an owner object
```
macandro96: **2 action items.** AI-1 is PR-introduced; AI-2 is pre-existing.

**TL;DR — inferring "actor handle vs owner object" from whether `.shutdown.remote` exists also silently drops the `ray.kill` fallback for an actor handle that has no `shutdown`.**

The comment concedes the design — "Nothing else distinguishes them here." Behaviour is inferred from the attributes an object happens to expose, which CLAUDE.md calls out as a smell worth flagging.

It is not purely stylistic. Previously an env whose class lacked `shutdown` raised `AttributeError` inside the `try` and escalated to `ray.kill`. Now `remote_shutdown` is `None`, the `else` branch calls `env.shutdown()` which also raises, and the new `continue` suppresses the kill — so a genuine Ray actor handle would leak instead of being killed.

Testing the type directly addresses both:

### AI-1
```suggestion
            remote_shutdown = (
                env.shutdown.remote if isinstance(env, ray.actor.ActorHandle) else None
            )
```
Every env class today defines `shutdown`, so there is no shipped impact — this is about keeping the safety net.

### AI-2 — pre-existing (from #3371), not introduced here
[`NemoGymShardSet.shutdown()`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/environments/nemo_gym.py#L1610) hardcodes `timeout=None`, so a caller's `shutdown_environments(..., timeout=30)` becomes unbounded for a shard set. Worth threading the caller's timeout through.

tests/unit/experience/test_nemo_gym_router.py
```diff
@@ -0,0 +1,290 @@
...
+    assert sum(timing is not None for _, timing, _ in drained) == 2
+
+
+def test_a_failing_shard_names_itself_rather_than_surfacing_a_bare_ray_error():
+    healthy, broken = _FakeActor("healthy"), _FakeActor("broken", fail=True)
+    buckets = [
+        ("healthy", healthy, _rows(["alpha"], num_generations=2)),
+        ("broken", broken, _rows(["beta"], num_generations=2)),
+    ]
+
```
macandro96, line 289: **1 action item.** Nit. PR-introduced.

This bucket is labelled `"busy/1"`, but [`_bucket_nemo_gym_rows_by_shard`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2467) cannot produce that form at this commit — it labels with a bare `shard_names.pop()`. The test passes only because it hand-builds the bucket and bypasses the producer, so it asserts against an input production cannot generate.

The `shard/index` form does show up later in the stack (#3374, `0fde7d12b`, which adds `instance_label`), which is presumably where it came from.

Related, and the reason this isn't caught elsewhere: [`test_replicas_of_one_shard_take_turns_by_group`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/tests/unit/experience/test_nemo_gym_router.py#L176) destructures `for _, handle, rows in buckets`, discarding the label. It is the only multi-replica test, so nothing asserts that two replicas of one shard get distinguishable labels — today they both get the bare shard name, and [`actor_timing_by_shard[shard_name]`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2797) drops one replica's timings.

### AI-1
Either use a bare label here to match what the producer emits, or pull the labelling fix forward — and assert the labels in the replica test.

nemo_rl/experience/rollouts.py
```diff
@@ -2424,6 +2427,164 @@ def _prepare_nemo_gym_rows(
...
+    for start in range(0, len(rows), num_generations):
+        group = rows[start : start + num_generations]
+        route_names = {get_nemo_gym_route_name(row) for row in group}
+        if len(route_names) != 1:
+            raise ValueError(
+                f"NeMo-Gym prompt group at row {start} mixes routes "
+                f"{sorted(route_names)}. A group must reference exactly one route."
+            )
+        shard_names = {shard_set.shard_for_route(name) for name in route_names}
+        if len(shard_names) != 1:
```
macandro96: **1 action item.** Nit. PR-introduced.

This branch is unreachable. [`:2453`](https://github.com/NVIDIA-NeMo/RL/blob/33d77fc8a8ed2f1f57b2f6ea5880f597cd9c08b5/nemo_rl/experience/rollouts.py#L2453) already raises unless `len(route_names) == 1`, and `shard_names` is a set comprehension over that single route name, so it always holds exactly one element.

### AI-1
Drop it, or keep it as an explicit `assert` if it is meant as a guard against a future multi-route group.

macandro96 (commented):
Team review of the shard router. Eight inline comments: one must-fix, two should-fix, three low-severity, two nits.

The only finding that fires today is the `master_config.ppo[...]` subscript in `ppo.py` — a hard `TypeError` on the PPO + NeMo-Gym path. Everything else is latent behind `spinup_nemo_gym_actor`'s `ShardConfigError` guard and goes live the moment #3373 removes it, which is the argument for fixing them here rather than after.

Verified along the way: the PR's own suite passes (13/13 in `test_nemo_gym_router.py`), `test_async_utils.py` passes deterministically (the one failure was a `pytest-randomly` ordering artifact), and `pre-commit` is clean for every file in this diff. Reviewed against base `5368ae619` (#3371's head), not `main`.

nemo_rl/algorithms/ppo.py
ananthsub: Fixed. PPO now reads the Pydantic field with `master_config.ppo.num_generations_per_prompt`, so argument evaluation no longer raises `TypeError`.

nemo_rl/algorithms/async_utils/trajectory_collector.py
ananthsub: Fixed. `_caused_by_ray_actor_death` now treats only `ActorDiedError` as terminal. `ActorUnavailableError` remains eligible for the existing retry/backoff path, and focused tests cover both classifications.

nemo_rl/experience/rollouts.py
ananthsub: Fixed. The compatibility wrapper now defaults this argument to `None` only to preserve the callable signature and raises immediately when it is omitted. Every in-tree caller now passes the prompt-group size explicitly; callers that intentionally route per row pass `1`.

nemo_rl/environments/nemo_gym.py
ananthsub: Fixed. Replica cursor reads and updates are protected by a `threading.Lock`. A concurrent-selection test blocks the first cursor read and verifies the second thread cannot enter the critical section, then confirms the two calls select successive replicas.

nemo_rl/experience/rollouts.py
ananthsub: Fixed. The merge now records the first failed instance, removes only that iterator, drains all healthy streams, and raises the attributed failure after the remaining streams finish. The failure test verifies that all healthy rows are yielded before the error is raised.

nemo_rl/environments/utils.py
ananthsub: Fixed both items. Teardown now distinguishes actor handles with `isinstance(env, ray.actor.ActorHandle)`, so a handle without `shutdown` reaches the `ray.kill` fallback. Owner objects receive the caller-provided timeout through `shutdown(timeout=timeout)`. Focused tests cover the missing-shutdown fallback and timeout propagation.

tests/unit/experience/test_nemo_gym_router.py
ananthsub: Fixed by pulling per-instance labeling into this layer. Replicated actors now produce `shard/index` labels while single-instance shards retain the bare shard name. The replica routing test asserts both distinct labels, preventing timing metrics from overwriting one another.

nemo_rl/experience/rollouts.py
ananthsub: Fixed. The redundant `shard_names` set and unreachable cardinality check were removed. After validating that a prompt group has one route, routing proceeds directly through that route to the selected actor instance.

nemo_rl/algorithms/grpo.py
```diff
@@ -4172,6 +4173,7 @@ def validate(
                     ),
                     max_rollout_turns=None,
                     greedy=False,
+                    num_generations_per_prompt=master_config.grpo.num_generations_per_prompt,
```
macandro96, line 4196: **1 action item.** PR-introduced.

**TL;DR — `validate()` repeats `val_batch` by `val_num_generations_per_prompt` but routes it by the training `num_generations_per_prompt`, which is the likely cause of the `divisible by routing_group_size` failures on `fast_L1_Functional_Tests_Gym` / `fast_L1_Functional_Tests_Megatron_4`.**

[`validate()`](https://github.com/NVIDIA-NeMo/RL/blob/296d4a8a4c40243d977ee7b201e94440e6b45618/nemo_rl/algorithms/grpo.py#L4083) computes the validation-specific group size and uses it correctly to shape the batch, then passes a different value to the routing call two lines later:

```python
val_num_generations_per_prompt = master_config.grpo.val_num_generations_per_prompt
...
if val_num_generations_per_prompt > 1:
    val_batch = val_batch.repeat_interleave(val_num_generations_per_prompt)
...
num_generations_per_prompt=master_config.grpo.num_generations_per_prompt,  # line 4176 — wrong variable
```

`val_num_generations_per_prompt` defaults independently of training's `num_generations_per_prompt` (16), so whenever they differ, `val_batch` is shaped for one group size while [`run_nemo_gym_rollout_sync`](https://github.com/NVIDIA-NeMo/RL/blob/296d4a8a4c40243d977ee7b201e94440e6b45618/nemo_rl/experience/rollouts.py#L2864) chunks rows using the other — exactly the `ValueError: NeMo-Gym rollout batch size must be divisible by routing_group_size` the functional tests are hitting, and CI's own log reads `❌ Initial validation failed: ...divisible by routing_group_size`, i.e. this exact call site.

The parallel training call site ([`grpo.py:3155`](https://github.com/NVIDIA-NeMo/RL/blob/296d4a8a4c40243d977ee7b201e94440e6b45618/nemo_rl/algorithms/grpo.py#L3155)) and both `distillation.py` call sites are correct — distillation has no separate val group-size field, so this is the one call site with the mismatch.

### AI-1
```suggestion
                    num_generations_per_prompt=val_num_generations_per_prompt,
```

nemo_rl/environments/nemo_gym.py
```diff
@@ -1507,6 +1525,12 @@ class NemoGymShardSet:
     handles: Dict[str, List[Any]]
     route_to_shard: Dict[str, str] = field(default_factory=dict)
     placement_group: Any = None
+    _next_replica: Dict[str, int] = field(default_factory=dict, repr=False)
+    _replica_lock: threading.Lock = field(
+        default_factory=threading.Lock,
+        repr=False,
+        compare=False,
+    )
```
macandro96, line 1538: **1 action item.** PR-introduced. Latent.

**TL;DR — the new `_replica_lock` field makes `NemoGymShardSet` unpicklable; harmless today, but #3373 (stacked on this PR) starts constructing and passing around real shard sets, and the first one that crosses a Ray boundary will crash.**

Reproduced directly:
```
pickle.dumps(shard_set)      -> TypeError: cannot pickle '_thread.lock' object
cloudpickle.dumps(shard_set) -> TypeError: cannot pickle '_thread.lock' object
ray.put(shard_set)           -> TypeError: Could not serialize the put value NemoGymShardSet(...)
                                 !!! FAIL serialization: cannot pickle '_thread.lock' object
```

Not reachable in this PR alone — [`grpo.py`](https://github.com/NVIDIA-NeMo/RL/blob/296d4a8a4c40243d977ee7b201e94440e6b45618/nemo_rl/algorithms/grpo.py#L804) still calls `spinup_nemo_gym_actor`, which never constructs a real `NemoGymShardSet` for the driver to hold. #3373 replaces that with `build_nemo_gym_actors`, which does, and if that object (or `task_to_env` containing it) ever crosses a `ray.put`, an actor constructor argument, or a checkpoint, it fails immediately.

Cheap to fix now, before it's load-bearing:

### AI-1
```suggestion
    _replica_lock: threading.Lock = field(
        default_factory=threading.Lock,
        repr=False,
        compare=False,
    )

    def __getstate__(self) -> dict[str, Any]:
        state = self.__dict__.copy()
        state["_replica_lock"] = None
        return state

    def __setstate__(self, state: dict[str, Any]) -> None:
        state["_replica_lock"] = threading.Lock()
        self.__dict__.update(state)
```

macandro96: I didn't catch the lock-serialization thingy previously. Okay with fixing here or in the next PR.

ananthsub: Resolved in the replication layer, PR #3374. `NemoGymShardSet.__getstate__` removes `_replica_lock` before Ray/cloudpickle serialization, and `__setstate__` recreates it after deserialization. A cloudpickle round-trip test verifies that replica rotation retains its cursor and uses the recreated lock.

nemo_rl/algorithms/grpo.py
ananthsub: Resolved in PR #3372. Validation now passes `val_num_generations_per_prompt`, the same value used to repeat `val_batch`, as the routing group size. This keeps validation row shaping and shard routing aligned when training and validation generation counts differ.

macandro96 (dismissed):

nemo_rl/environments/nemo_gym.py
ananthsub: Follow-up: moved this fix down from #3374 into this PR, where `_replica_lock` is introduced. The serialization hooks and cloudpickle round-trip regression test are now present directly in #3372.

terrykong (approved):

PR #3373, opened 2026-07-27
feat(nemo-gym): wire sharded stacks into training

ananthsub:
## Summary

Connects the shard factory and rollout router to GRPO, distillation, and SingleController so `env.nemo_gym.shards` can run end to end.

Each entrypoint stores a `NemoGymShardSet`, including the unsharded one-actor case. Setup verifies that every `agent_ref` in the training and validation datasets is hosted by a shard. If generation, trainer, or coverage setup fails after the Gym future succeeds, the completed shard set is recovered from that future and shut down.

Dataset coverage uses filesystem identities recorded by NeMo-Gym dataset loaders. The identity includes the canonical path, device, inode, change time, modification time, and size. Sharded jobs scan each unchanged source once and cache its agent names. Replaced or changed files fall back to scanning the rows already loaded into the dataset. Unsharded jobs skip this work.

This layer also handles requirements that arise when a sharded configuration inherits from an existing Hydra recipe:

- Null inherited keys are treated as removed and are not forwarded to Gym. A child recipe can therefore remove inherited `config_paths` or an entry that no shard hosts.
- `placement_strategy` can relax `STRICT_SPREAD` for local testing. Colocated shards must declare port ranges with no shared endpoint because Gym treats both bounds as inclusive. Startup reports that node-level capacity isolation no longer applies.

`examples/nemo_gym/grpo_sharded_gym_smoke.yaml` provides a two-shard local configuration. It claims the parent recipe's `policy_model` entry by key rather than copying its value. SingleController preserves deferred model loading while Gym starts and validates the resulting shard set.

## Dependency

This PR targets the branch from [PR #3372](https://github.com/NVIDIA-NeMo/RL/pull/3372). [PR #3374](https://github.com/NVIDIA-NeMo/RL/pull/3374) builds on this branch.

### DIFF
--- a/examples/nemo_gym/grpo_sharded_gym_smoke.yaml
+++ b/examples/nemo_gym/grpo_sharded_gym_smoke.yaml
@@ -1,0 +1,79 @@
+# Two-shard NeMo-Gym smoke config, for exercising sharding by hand.
+#
+# This is not a training recipe. It is the smallest config that puts two Gym
+# stacks behind one job so the sharding machinery can be watched end to end:
+# per-shard spinup, the agent->shard map, routing whole prompt groups to the
+# shard hosting their agent, per-shard log directories, and teardown of both
+# actors plus their placement group.
+#
+# The split is by environment, since an agent may live in exactly one shard:
+# workplace_assistant on one, math_with_judge on the other. Both were already
+# a supported multi-environment combination -- see the commented config_paths
+# in the base recipe -- so this only changes where they run, not what they do.
+#
+# Run `ng_prepare_data` for BOTH environments and point data.train/validation
+# at the combined jsonl before running. A dataset covering only one of them
+# still starts, but every row lands on one shard and the other sits idle, so
+# routing is never exercised. A dataset naming a third agent fails at setup,
+# which is itself worth seeing once.
+#
+# Single node:
+#   the default STRICT_SPREAD wants one node per shard, so override
+#   env.nemo_gym.placement_strategy=PACK to colocate them. That gives up the
+#   capacity isolation sharding exists for, which is fine when the point is to
+#   test the mechanism rather than to relieve a node.
+#
+# Two or more nodes:
+#   leave placement_strategy alone and confirm the shards land on different
+#   nodes.
+
+defaults: "grpo_workplace_assistant_nemotron_nano_v2_9b.yaml"
+
+env:
+  nemo_gym:
+    # Retract the inherited single-stack settings. A Hydra override can blank
+    # an inherited key but not delete it, so null is how these are given up;
+    # the sharded schema reads null as absent.
+    config_paths: null
+    policy_model: null
+    # An overlay for an environment neither shard hosts. Left in place it would
+    # be rejected as having no shard to belong to, which is the check working.
+    code_gen: null
+
+    # Each shard gets its own port window so two stacks on one node cannot
+    # collide. The base recipe's 5000-5999 is split in half.
+    shards:
+      - name: workplace
+        port_range_low: 5000
+        port_range_high: 5499
+        config_paths:
+          - responses_api_models/vllm_model/configs/vllm_model_for_training.yaml
+          - resources_servers/workplace_assistant/configs/workplace_assistant.yaml
+
+      - name: math
+        port_range_low: 5500
+        port_range_high: 5999
+        config_paths:
+          - responses_api_models/vllm_model/configs/vllm_model_for_training.yaml
+          - resources_servers/math_with_judge/configs/math_with_judge.yaml
+
+    # Both shards need policy_model, which is a client of NeMo RL's vLLM
+    # servers rather than an engine of its own, so hosting it twice claims no
+    # extra GPU. That is what the allowlist below is for; without it the
+    # duplicate-entry check would stop the job, which is the right default
+    # since a duplicated engine-bearing entry doubles its GPU claim silently.
+    common_overrides:
+      policy_model:
+        responses_api_models:
+          vllm_model:
+            uses_reasoning_parser: false
+            extra_body:
+              chat_template_kwargs:
+                enable_thinking: false
+    allowed_duplicate_entries:
+      - policy_model
+
+logger:
+  log_dir: "logs/grpo-sharded-gym-smoke"
+  wandb_enabled: false
+  monitor_gpus: false
--- a/nemo_rl/algorithms/distillation.py
+++ b/nemo_rl/algorithms/distillation.py
@@ -52,7 +52,7 @@ from nemo_rl.distributed.virtual_cluster import (
     prepare_segment_topology,
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
-from nemo_rl.environments.nemo_gym import spinup_nemo_gym_actor
+from nemo_rl.environments.nemo_gym import build_nemo_gym_actors
 from nemo_rl.experience.rollouts import (
     run_async_multi_turn_rollout,
     run_multi_turn_rollout,
@@ -505,7 +505,7 @@ def setup(
             def init_nemo_gym():
                 # Distillation does not configure vLLM for router replay, so the
                 # actor must not require routed experts on its output items.
-                return spinup_nemo_gym_actor(
+                return build_nemo_gym_actors(
                     env_configs,
                     base_urls=deferred_vllm.dp_openai_server_base_urls,
                     model_name=generation_config["model_name"],
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -79,7 +79,10 @@ from nemo_rl.distributed.virtual_cluster import (
     prepare_segment_topology,
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
-from nemo_rl.environments.nemo_gym import spinup_nemo_gym_actor
+from nemo_rl.environments.nemo_gym import (
+    build_nemo_gym_actors,
+    validate_dataset_agent_coverage,
+)
 from nemo_rl.environments.utils import shutdown_environments
 from nemo_rl.experience.interfaces import (
     NEMO_GYM_TASK_INDEX_KEY,
@@ -590,16 +593,28 @@ def setup(
     nemo_gym_actor = None
 
     def _spinup_nemo_gym(base_urls, model_name):
-        """Spin up the NeMo Gym actor against the given generation server URLs."""
+        """Spin up the NeMo Gym stack against the given generation server URLs.
+
+        Returns a shard set, which is the one actor of an unsharded job as much
+        as it is the K of a sharded one.
+        """
         t0 = time.perf_counter()
-        actor = spinup_nemo_gym_actor(
+        shard_set = build_nemo_gym_actors(
             env_configs,
             base_urls=base_urls,
             model_name=model_name,
             enable_router_replay=router_replay_enabled(policy_config),
             use_fastokens=bool(policy_config["tokenizer"].get("use_fastokens")),
         )
-        return actor, time.perf_counter() - t0
+        train_splits = (
+            {f"train[{name}]": split for name, split in dataset.items()}
+            if isinstance(dataset, dict)
+            else {"train": dataset}
+        )
+        validate_dataset_agent_coverage(
+            shard_set, {**train_splits, "validation": val_dataset}
+        )
+        return shard_set, time.perf_counter() - t0
 
     total_nodes = cluster_config["num_nodes"]
     segment_size = cluster_config.get("segment_size")
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -49,7 +49,7 @@ from nemo_rl.data.utils import setup_response_data
 from nemo_rl.data_plane import DataPlaneClient, build_data_plane_client
 from nemo_rl.distributed.virtual_cluster import RayVirtualCluster
 from nemo_rl.environments.interfaces import EnvironmentInterface
-from nemo_rl.environments.nemo_gym import spinup_nemo_gym_actor
+from nemo_rl.environments.nemo_gym import build_nemo_gym_actors
 from nemo_rl.experience.rollout_manager import RolloutManager
 from nemo_rl.models.generation.sglang.config import SGLangConfig
 from nemo_rl.models.generation.sglang.sglang_generation import SGLangGeneration
@@ -384,12 +384,20 @@ def setup_single_controller(
     if use_nemo_gym:
         # TODO(#2625): Mirror GRPO's deferred vLLM load so NeMo-Gym spinup
         # overlaps model loading instead of running serially afterward.
-        env_handles["nemo_gym"] = spinup_nemo_gym_actor(
-            master_config.env,
-            base_urls=generation.dp_openai_server_base_urls,
-            model_name=generation_config["model_name"],
-            enable_router_replay=router_replay_enabled(master_config.policy),
-            use_fastokens=bool(master_config.policy["tokenizer"].get("use_fastokens")),
+        # EnvironmentInterface describes step()-based environments; the gym
+        # entry has never been one. It used to be an actor handle and is now a
+        # shard set, and callers only ever ask it to run rollouts or shut down.
+        env_handles["nemo_gym"] = cast(
+            EnvironmentInterface,
+            build_nemo_gym_actors(
+                master_config.env,
+                base_urls=generation.dp_openai_server_base_urls,
+                model_name=generation_config["model_name"],
+                enable_router_replay=router_replay_enabled(master_config.policy),
+                use_fastokens=bool(
+                    master_config.policy["tokenizer"].get("use_fastokens")
+                ),
+            ),
         )
 
     # ==========================
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -11,12 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import json
 import math
 import os
 import subprocess
 import sys
 from collections import Counter
-from collections.abc import AsyncGenerator
+from collections.abc import AsyncGenerator, Mapping
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Dict, List, NotRequired, Optional, TypedDict
@@ -38,6 +39,7 @@ from nemo_rl.distributed.virtual_cluster import (
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
 from nemo_rl.environments.nemo_gym_shards import (
+    DEFAULT_PLACEMENT_STRATEGY,
     SHARDING_CONFIG_KEYS,
     ShardConfigError,
     ShardPlan,
@@ -888,22 +890,6 @@ class NemoGymShardSet:
                 return shard_name
         raise ShardSetupError("Handle does not belong to this NeMo-Gym shard set")
 
-    def sole_handle(self) -> Any:
-        """The only actor, for callers that predate routing.
-
-        Raises:
-            ShardSetupError: There is more than one actor, so picking one would
-                silently drop the rest.
-        """
-        handles = self.all_handles
-        if len(handles) != 1:
-            raise ShardSetupError(
-                f"Expected a single NeMo-Gym actor but this set has "
-                f"{len(handles)} across shards {sorted(self.handles)}. The "
-                f"caller needs the shard-aware rollout router."
-            )
-        return handles[0]
-
     def shutdown(self) -> None:
         """Stop every actor, then release the bundles they were pinned to."""
         shutdown_environments(
@@ -1045,8 +1031,14 @@ def _build_sharded_gym_actors(
     if nemo_gym_dict.get("num_gpu_nodes", 0):
         print(
             "env.nemo_gym.shards is set, so the num_gpu_nodes affinity hint is "
-            "superseded by STRICT_SPREAD placement across "
-            f"{len(instances)} nodes."
+            f"superseded by {plan.placement_strategy} placement across "
+            f"{len(instances)} bundles."
+        )
+    if plan.placement_strategy != DEFAULT_PLACEMENT_STRATEGY:
+        print(
+            f"env.nemo_gym.placement_strategy is {plan.placement_strategy}, not "
+            f"{DEFAULT_PLACEMENT_STRATEGY}, so shards may share a node and the "
+            f"per-node capacity isolation sharding exists for does not hold."
         )
 
     base_gym_dict = {
@@ -1066,16 +1058,17 @@ def _build_sharded_gym_actors(
             {"CPU": float(shard.actor_cpus or DEFAULT_SHARD_CPUS)}
             for shard, _ in instances
         ],
-        strategy="STRICT_SPREAD",
+        strategy=plan.placement_strategy,
     )
     try:
         ray.get(pg.ready(), timeout=pg_ready_timeout)
     except BaseException as error:
         remove_placement_group(pg)
         raise ShardSetupError(
-            f"Could not place {len(instances)} NeMo-Gym shard instances on "
-            f"distinct nodes within {pg_ready_timeout}s. STRICT_SPREAD needs "
-            f"one node per instance with the requested CPUs free; the "
+            f"Could not place {len(instances)} NeMo-Gym shard instances with "
+            f"strategy {plan.placement_strategy} within {pg_ready_timeout}s. "
+            f"Every instance needs the requested CPUs free, and "
+            f"{DEFAULT_PLACEMENT_STRATEGY} needs them on distinct nodes; the "
             f"allocation may be too small or its nodes too busy."
         ) from error
 
@@ -1157,40 +1150,61 @@ def _discover_agent_shard_map(
     return build_agent_shard_map(entries_by_shard, plan.allowed_duplicate_entries)
 
 
-def spinup_nemo_gym_actor(
-    env_configs: dict[str, Any],
-    *,
-    base_urls: list[Optional[str]],
-    model_name: str,
-    enable_router_replay: bool = False,
-    use_fastokens: bool = False,
-) -> Any:
-    """Spin up a single NeMo-Gym actor against the given generation server URLs.
+def validate_dataset_agent_coverage(
+    shard_set: NemoGymShardSet,
+    datasets: Mapping[str, Any],
+) -> None:
+    """Fail at setup if any row names an agent no shard hosts.
 
-    When ``env_configs["nemo_gym"]["num_gpu_nodes"] > 0``, the actor is
-    scheduled with soft NodeAffinity to the caller's Ray node so its colocated
-    GPU resources land where the caller expects.
+    Without this the mistake still surfaces, but only when a row naming the
+    missing agent is first dispatched. A rare agent can sit unseen for hours of
+    training, so the useful time to catch it is before the first step.
 
-    Returns:
-        The spun-up ``NemoGym`` Ray actor handle (``_spinup`` already awaited).
+    Unsharded jobs are skipped: there is one actor, every agent resolves to it,
+    and there is nothing a scan could discover.
+
+    Args:
+        shard_set: The running actors, carrying the agent map built at setup.
+        datasets: Split name to dataset, for the error message. ``None`` values
+            and datasets without gym rows are skipped.
 
     Raises:
-        ShardConfigError: The config is sharded. Callers that dispatch to a
-            single handle cannot serve several shards; they need the router.
+        ShardSetupError: A split references agents no shard hosts.
     """
-    plan = parse_shard_plan(dict(env_configs["nemo_gym"]))
-    if plan is not None:
-        raise ShardConfigError(
-            f"env.nemo_gym.shards defines {len(plan.shards)} shards "
-            f"({', '.join(shard.name for shard in plan.shards)}), but this "
-            f"entrypoint dispatches rollouts to one actor handle. Shard-aware "
-            f"rollout routing is not wired up yet."
-        )
+    if not shard_set.is_sharded:
+        return
 
-    return build_nemo_gym_actors(
-        env_configs,
-        base_urls=base_urls,
-        model_name=model_name,
-        enable_router_replay=enable_router_replay,
-        use_fastokens=use_fastokens,
-    ).sole_handle()
+    hosted = shard_set.hosted_agents
+    for split, dataset in datasets.items():
+        unhosted = sorted(_iter_dataset_agent_names(dataset) - hosted)
+        if unhosted:
+            raise ShardSetupError(
+                f"The {split} dataset references agents that no shard hosts: "
+                f"{unhosted}. Hosted agents: {sorted(hosted)}."
+            )
+
+
+def _iter_dataset_agent_names(dataset: Any) -> set[str]:
+    """Collect the agent names a dataset's rows reference.
+
+    Rows store ``extra_env_info`` as a JSON string, since a HuggingFace
+    ``Dataset`` does not hold the nested structure well, so reading the agent
+    out means parsing each row. That cost is paid once, at setup, and only by
+    sharded jobs.
+    """
+    if dataset is None:
+        return set()
+    # AllTaskProcessedDataset wraps the raw rows; a plain sequence is also fine.
+    rows = getattr(dataset, "dataset", dataset)
+
+    names: set[str] = set()
+    for row in rows:
+        extra_env_info = row.get("extra_env_info") if hasattr(row, "get") else None
+        if isinstance(extra_env_info, str):
+            extra_env_info = json.loads(extra_env_info)
+        if not isinstance(extra_env_info, dict):
+            continue
+        agent_ref = extra_env_info.get("agent_ref")
+        if isinstance(agent_ref, dict) and "name" in agent_ref:
+            names.add(str(agent_ref["name"]))
+    return names
--- a/nemo_rl/environments/nemo_gym_shards.py
+++ b/nemo_rl/environments/nemo_gym_shards.py
@@ -34,8 +34,20 @@ from omegaconf import OmegaConf
 # top-level keys as server instance configs, which can trip its
 # error_on_almost_servers path.
 SHARDING_CONFIG_KEYS = frozenset(
-    {"shards", "common_overrides", "allowed_duplicate_entries"}
+    {
+        "shards",
+        "common_overrides",
+        "allowed_duplicate_entries",
+        "placement_strategy",
+    }
 )
+
+# Ray placement-group strategies a shard plan may ask for. STRICT_SPREAD is the
+# point of sharding -- one actor per node -- and anything else colocates shards
+# and gives up the capacity isolation. PACK exists so the mechanism can be
+# exercised on a single machine.
+PLACEMENT_STRATEGIES = frozenset({"STRICT_SPREAD", "SPREAD", "STRICT_PACK", "PACK"})
+DEFAULT_PLACEMENT_STRATEGY = "STRICT_SPREAD"
 
 # Gym's log-directory key (``nemo_gym.global_config.NEMO_GYM_LOG_DIR_KEY_NAME``).
 # Restated rather than imported: this module is imported on the driver, and
@@ -88,6 +100,7 @@ class ShardPlan:
     shards: list[ShardSpec]
     common_overrides: dict[str, Any] = field(default_factory=dict)
     allowed_duplicate_entries: frozenset[str] = frozenset()
+    placement_strategy: str = DEFAULT_PLACEMENT_STRATEGY
 
     @property
     def total_instances(self) -> int:
@@ -200,10 +213,15 @@ def parse_shard_plan(nemo_gym_config: Mapping[str, Any]) -> ShardPlan | None:
     # top-level config_paths or entry overlay could plausibly belong to one
     # shard or to all of them, and guessing wrong silently changes what each
     # actor hosts.
-    if "config_paths" in nemo_gym_config:
+    # A null reads as absent throughout this block. A sharded config is
+    # normally written as a Hydra override of an unsharded recipe, and an
+    # override can only blank an inherited key, never delete it -- so
+    # `config_paths: null` is how one says "the shards own the paths now".
+    if nemo_gym_config.get("config_paths") is not None:
         raise ShardConfigError(
             "env.nemo_gym.config_paths cannot be combined with 'shards'. "
-            "Move each path into the shard that should host it."
+            "Move each path into the shard that should host it, and set the "
+            "inherited config_paths to null."
         )
     stray_entries = find_gym_config_entries(nemo_gym_config)
     if stray_entries:
@@ -226,6 +244,15 @@ def parse_shard_plan(nemo_gym_config: Mapping[str, Any]) -> ShardPlan | None:
     if duplicate_names:
         raise ShardConfigError(f"Duplicate shard names: {duplicate_names}")
 
+    placement_strategy = (
+        nemo_gym_config.get("placement_strategy") or DEFAULT_PLACEMENT_STRATEGY
+    )
+    if placement_strategy not in PLACEMENT_STRATEGIES:
+        raise ShardConfigError(
+            f"env.nemo_gym.placement_strategy must be one of "
+            f"{sorted(PLACEMENT_STRATEGIES)}, got {placement_strategy!r}"
+        )
+
     raw_allowed = nemo_gym_config.get("allowed_duplicate_entries") or []
     if OmegaConf.is_config(raw_allowed):
         raw_allowed = OmegaConf.to_container(raw_allowed, resolve=True)
@@ -242,6 +269,7 @@ def parse_shard_plan(nemo_gym_config: Mapping[str, Any]) -> ShardPlan | None:
             nemo_gym_config.get("common_overrides"), context="common_overrides"
         ),
         allowed_duplicate_entries=frozenset(raw_allowed),
+        placement_strategy=placement_strategy,
     )
 
 
@@ -253,6 +281,11 @@ def apply_shard_overlay(
     The ``config_paths`` merge itself still happens inside Gym, exactly as for
     an unsharded job; this only settles which paths and overlays that shard
     starts from. Per-shard overrides win over ``common_overrides`` on conflict.
+
+    Top-level keys left null are dropped rather than forwarded. Blanking a key
+    is the only way a Hydra override can retract one it inherited (see
+    ``parse_shard_plan``), and Gym should not receive a null where it expects
+    an entry.
     """
     merged = OmegaConf.to_container(
         OmegaConf.merge(
@@ -263,6 +296,7 @@ def apply_shard_overlay(
         resolve=True,
     )
     assert isinstance(merged, dict)
+    merged = {key: value for key, value in merged.items() if value is not None}
     merged["config_paths"] = list(shard.config_paths)
     if shard.port_range_low is not None:
         merged["port_range_low"] = shard.port_range_low
--- a/tests/unit/algorithms/test_distillation.py
+++ b/tests/unit/algorithms/test_distillation.py
@@ -984,7 +984,7 @@ def test_distillation_setup_non_colocated_smoke(monkeypatch, refit_transport):
         patch.object(
             distil_mod, "create_weight_synchronizer"
         ) as mock_create_synchronizer,
-        patch.object(distil_mod, "spinup_nemo_gym_actor") as mock_spinup_nemo_gym,
+        patch.object(distil_mod, "build_nemo_gym_actors") as mock_spinup_nemo_gym,
         patch.object(distil_mod, "ray") as mock_ray,
     ):
         mock_ckpt_mgr.return_value.get_latest_checkpoint_path.return_value = None
@@ -1125,7 +1125,7 @@ def test_distillation_setup_nemo_gym_uses_deferred_vllm(monkeypatch):
         patch.object(distil_mod, "Policy", DummyPolicy),
         patch.object(distil_mod, "VllmGeneration", DummyVllmGeneration),
         patch.object(
-            distil_mod, "spinup_nemo_gym_actor", return_value=nemo_gym_actor
+            distil_mod, "build_nemo_gym_actors", return_value=nemo_gym_actor
         ) as mock_spinup_nemo_gym,
         patch.object(distil_mod, "ray") as mock_ray,
     ):
--- a/tests/unit/environments/test_nemo_gym_shards.py
+++ b/tests/unit/environments/test_nemo_gym_shards.py
@@ -88,6 +88,25 @@ def test_top_level_config_paths_conflicts_with_shards():
         parse_shard_plan(_sharded_config(config_paths=["gym.yaml"]))
 
 
+def test_an_inherited_config_paths_can_be_retracted_with_null():
+    """A Hydra override can blank a key it inherited but cannot delete it, so
+    null has to read as absent or no recipe could ever be sharded."""
+    plan = parse_shard_plan(_sharded_config(config_paths=None))
+
+    assert plan is not None
+    assert [shard.name for shard in plan.shards] == ["judged", "tools"]
+
+
+def test_a_nulled_entry_overlay_is_not_forwarded_to_gym():
+    """Gym should get the key gone, not the key set to null."""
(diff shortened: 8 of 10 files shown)


nemo_rl/algorithms/grpo.py
```diff
@@ -3156,6 +3176,9 @@ def _grpo_train_impl(
                             effort_config=_get_effort_config(master_config),
                             reward_penalty_config=master_config.reward_penalties,
                             thinking_tags=get_nemo_gym_thinking_tags(master_config.env),
+                            num_generations_per_prompt=(
```
macandro96, line 3199: [`grpo.py:3175-3181`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/grpo.py#L3175-L3181)

**TL;DR — this call already had a `num_generations_per_prompt=` kwarg at line 3175; adding a second one makes `import nemo_rl.algorithms.grpo` raise `SyntaxError: keyword argument repeated`.**

Reproduced directly: `python -c "compile(open('nemo_rl/algorithms/grpo.py').read(), 'g', 'exec')"` -> `SyntaxError: keyword argument repeated: num_generations_per_prompt (grpo.py, line 3179)`. The base branch (#3372) compiles clean, and this PR's own lint job reports the same error. This breaks collection of every test file that transitively imports this module.

Both values are textually identical (`master_config.grpo.num_generations_per_prompt`), so this is a pure duplicate -- drop the newer one.

Note: #3374 also removes this duplicate, but folding the fix in here keeps this layer of the stack independently valid (right now `main`-bound history would include a commit range where `grpo.py` doesn't import).

### AI-1
```suggestion
                            num_generations_per_prompt=master_config.grpo.num_generations_per_prompt,
                            effort_config=_get_effort_config(master_config),
                            reward_penalty_config=master_config.reward_penalties,
                            thinking_tags=get_nemo_gym_thinking_tags(master_config.env),
```

nemo_rl/algorithms/grpo.py
```diff
@@ -4177,6 +4200,7 @@ def validate(
                     effort_config=_get_effort_config(master_config),
                     reward_penalty_config=master_config.reward_penalties,
                     thinking_tags=get_nemo_gym_thinking_tags(master_config.env),
+                    num_generations_per_prompt=val_num_generations_per_prompt,
```
macandro96, line 4223: [`grpo.py:4199-4203`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/grpo.py#L4199-L4203)

**TL;DR — same duplicate-kwarg `SyntaxError` as the train-site call above, but here the two values differ and the newly-added one is the correct one to keep.**

`val_batch` is repeat-interleaved by `val_num_generations_per_prompt` right before this call ([`grpo.py:4162-4163`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/grpo.py#L4162-L4163)), so the routing group size passed here must match that value, not the train-time `master_config.grpo.num_generations_per_prompt`. Delete the pre-existing line (4199), keep the PR-added one (4203). As with the train site, fold this fix in directly rather than relying on #3374 to remove it.

### AI-1
```suggestion
                    effort_config=_get_effort_config(master_config),
                    reward_penalty_config=master_config.reward_penalties,
                    thinking_tags=get_nemo_gym_thinking_tags(master_config.env),
                    num_generations_per_prompt=val_num_generations_per_prompt,
```

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -1527,8 +1534,23 @@ def _build_trainer_then_megatron_generation() -> tuple[
                 weight_synchronizer.sync_weights()
                 setup_timing_metrics.weight_sync_time_s = time.perf_counter() - t0
             if use_nemo_gym:
-                env_handles["nemo_gym"], gym_time = submitted["nemo_gym"].result()
+                nemo_gym_shards, gym_time = submitted["nemo_gym"].result()
+                validate_dataset_agent_coverage(
+                    nemo_gym_shards,
+                    {"training": dataset, "validation": _val_dataset},
+                )
+                env_handles["nemo_gym"] = cast(EnvironmentInterface, nemo_gym_shards)
```
macandro96, line 1604: [`setup.py:1542`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/single_controller_utils/setup.py#L1542)

**1 action item.** The lock originated in #3372; the serialization failure is introduced by #3373 when it passes the shard set to SingleControllerActor.remote()

**TL;DR — `NemoGymShardSet` carries a `threading.Lock` field with no pickle support, and this line is where it starts flowing into `SingleControllerActor.remote(...)`'s cloudpickled constructor args. This crashes every SingleController + NeMo-Gym run, including unsharded configs -- not just sharded ones.**

[`build_nemo_gym_actors`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/environments/nemo_gym.py#L1668-L1680) returns a `NemoGymShardSet` unconditionally -- the unsharded path (`plan is None`) still wraps its single actor in one, per its own docstring ("Without `shards` this makes exactly one actor... Returns: A `NemoGymShardSet`"). So this is not sharding-specific.

[`NemoGymShardSet`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/environments/nemo_gym.py#L1513-L1536) has a `_replica_lock: threading.Lock` field with no `__getstate__`/`__setstate__`. Reproduced directly:
```
>>> import cloudpickle
>>> cloudpickle.dumps(NemoGymShardSet(handles={"a": [object()]}))
TypeError: cannot pickle '_thread.lock' object
```
Ray's own docs note objects with unpicklable state (locks, sockets, etc.) cannot cross the actor-construction serialization boundary. Reachability chain, starting at this line:
1. `env_handles["nemo_gym"] = cast(EnvironmentInterface, nemo_gym_shards)`.
2. `env_handles` becomes `SingleControllerActorArgs.env_handles` ([`setup.py:1795`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/single_controller_utils/setup.py#L1795), field typed `dict[str, EnvironmentInterface]`).
3. [`examples/run_grpo_single_controller.py:154`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/examples/run_grpo_single_controller.py#L154): `SingleControllerActor.remote(actor_args=actor_args, ...)` -- a `@ray.remote` actor ([`single_controller.py:210-211`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/single_controller.py#L210-L211)), whose constructor args Ray cloudpickles.
4. `cloudpickle.dumps` on a `NemoGymShardSet`-bearing `actor_args` raises the error above.

### AI-1
Fix at the dataclass, in `nemo_gym.py`:
```python
    _replica_lock: threading.Lock = field(
        default_factory=threading.Lock,
        repr=False,
        compare=False,
    )

    def __getstate__(self) -> dict[str, Any]:
        state = self.__dict__.copy()
        # A lock cannot cross a process boundary; recreate it on the far side.
        state.pop("_replica_lock", None)
        return state

    def __setstate__(self, state: dict[str, Any]) -> None:
        self.__dict__.update(state)
        self._replica_lock = threading.Lock()
```

tests/unit/single_controller/test_setup.py
```diff
@@ -1209,7 +1210,7 @@ def test_invalid_config_fails_before_setup_factories(
 
         with (
             patch.object(sc_setup_mod, "should_use_nemo_gym", return_value=use_gym),
-            patch.object(sc_setup_mod, "spinup_nemo_gym_actor") as mock_spinup,
+            patch.object(sc_setup_mod, "build_nemo_gym_actors") as mock_spinup,
```
macandro96, line 1215: [`test_setup.py:1213`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/tests/unit/single_controller/test_setup.py#L1213) (this hunk -- see [`test_setup.py:818`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/tests/unit/single_controller/test_setup.py#L818) for the actual bug, which this PR's diff never touches directly)

**TL;DR — one test in this file, `test_periodic_checkpointing_warns_without_per_step_trainer_anchors` at line 818, still patches `spinup_nemo_gym_actor`, which this PR's own commit renamed to `build_nemo_gym_actors` everywhere else -- including right here -- so it now fails with `AttributeError`.**

`spinup_nemo_gym_actor` is no longer imported into `single_controller_utils.setup`, so `patch.object(sc_setup_mod, "spinup_nemo_gym_actor", ...)` raises `AttributeError: <module '...single_controller_utils.setup'> does not have the attribute 'spinup_nemo_gym_actor'` at test collection/setup time. Verified `a143a5bd3` (the rename commit) is part of this PR's own commit range, not the base branch. Every other call site in this file, including this one, was updated to `build_nemo_gym_actors`; line 818 was missed.

### AI-1
At `test_setup.py:818`:
```python
                sc_setup_mod, "build_nemo_gym_actors", return_value=MagicMock()
```

nemo_rl/data/utils.py
```diff
@@ -107,6 +108,25 @@ def load_dataloader_state(
     dataloader.load_state_dict(saved_state)
 
 
+def _combine_agent_name_sources(
+    datasets: list[Any],
+) -> frozenset[NemoGymSourceIdentity] | None:
+    return _combine_agent_name_source_sets(
```
macandro96, line 114: [`data/utils.py:111-116`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/data/utils.py#L111-L116)

**1 action item.** PR-introduced. Compatibility blocker.

**TL;DR — bare `dataset.agent_name_sources` reads with no `getattr` fallback mean any external/custom dataset class -- which is a supported pattern that does not require subclassing `RawDataset` -- now raises `AttributeError` in `setup_response_data`, for unsharded and non-gym jobs too.**

The documented custom-dataset contract (`docs/guides/grpo.md`) only requires implementing `set_task_spec`/`set_processor`. `load_response_dataset` explicitly supports fully external classes via `resolve_external_dataset_class`. The sibling loop 20 lines below this function correctly guards with `hasattr(data, "val_dataset")`; this one doesn't. This is a compatibility break for existing external-dataset configs, independent of whether gym or sharding is in play at all.

### AI-1
```suggestion
def _combine_agent_name_sources(
    datasets: list[Any],
) -> frozenset[NemoGymSourceIdentity] | None:
    return _combine_agent_name_source_sets(
        [getattr(dataset, "agent_name_sources", None) for dataset in datasets]
    )
```
Apply the same `getattr(..., "agent_name_sources", None)` pattern at the two other new call sites in this file (`data.agent_name_sources` and `val_data.agent_name_sources`).

nemo_rl/algorithms/grpo.py
```diff
@@ -799,17 +803,33 @@ def init_train_dataloader(dataset, suffix: str = ""):
...
+            {f"train[{name}]": split for name, split in dataset.items()}
+            if isinstance(dataset, dict)
+            else {"train": dataset}
+        )
+        try:
+            validate_dataset_agent_coverage(
+                shard_set, {**train_splits, "validation": val_dataset}
+            )
+        except BaseException:
+            shard_set.shutdown()
```
macandro96, line 843: [`grpo.py:830`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/grpo.py#L830)

**1 action item.** PR-introduced, at 4 new exception-cleanup call sites.

**TL;DR — this and 3 sibling call sites invoke `shard_set.shutdown()` with `timeout=None` (unbounded wait); they should use the existing bounded timeout and `force_kill=True`.**

`shutdown()` defaults to `timeout=None`, reaching `ray.get(..., timeout=None)` inside `shutdown_environments` -- an indefinite wait. The other 3 new sites are [`distillation.py:538`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/distillation.py#L538), [`distillation.py:563`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/distillation.py#L563), and [`setup.py:1552`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/single_controller_utils/setup.py#L1552). `NEMO_GYM_GRACEFUL_SHUTDOWN_TIMEOUT_S = 120` already exists and is used correctly at exactly one existing site ([`nemo_gym.py:1767-1769`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/environments/nemo_gym.py#L1767-L1769), with `force_kill=True`).

### AI-1
One option -- of several -- fix the default so a future 5th caller can't repeat the mistake:
```python
    def shutdown(
        self,
        *,
        timeout: float | None = NEMO_GYM_GRACEFUL_SHUTDOWN_TIMEOUT_S,
        force_kill: bool = True,
    ) -> None:
```
(in `nemo_gym.py`, `NemoGymShardSet.shutdown`). Otherwise, pass `timeout=NEMO_GYM_GRACEFUL_SHUTDOWN_TIMEOUT_S, force_kill=True` explicitly at all 4 call sites.

nemo_rl/algorithms/grpo.py
```diff
@@ -799,17 +803,33 @@ def init_train_dataloader(dataset, suffix: str = ""):
...
+            raise
+        return shard_set, time.perf_counter() - t0
```
macandro96, line 845: [`grpo.py:817-835`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/grpo.py#L817-L835) (`_spinup_nemo_gym`, called from the Megatron-path and vLLM-path `ThreadPoolExecutor` blocks further down this file)

**1 action item.** The concurrent-initialization structure predates this PR, but this PR amplifies the failure: `_spinup_nemo_gym` now returns an owner holding multiple actors plus a placement group (rather than a single actor), and this PR added shutdown-on-sibling-failure recovery for distillation.py and setup.py but not for GRPO's two `ThreadPoolExecutor` blocks that call this function.

**TL;DR — if policy/generation startup fails after this function's Gym future succeeds, the completed shard set is never recovered or shut down; its actors, placement group, and Gym subprocesses/ports leak.**

Both GRPO paths that call `_spinup_nemo_gym` submit it alongside policy/generation as concurrent futures and only reach `nemo_gym_future.result()` (or `results["nemo_gym"]`) after the other futures resolve. If a sibling future raises first, that line is never reached, so the built `NemoGymShardSet` from this function's `return shard_set, time.perf_counter() - t0` (line 832) is never assigned anywhere the caller can shut down. This affects both the Megatron-path executor (`grpo.py:1429-1449`) and the vLLM-path executor (`grpo.py:1586-1590`, which has no exception handling at all).

SingleController already recovers this correctly ([`setup.py:1544-1553`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/single_controller_utils/setup.py#L1544-L1553)), and distillation.py gained the same pattern in this PR ([`distillation.py:556-564`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/distillation.py#L556-L564)).

### AI-1
Apply the same recover-and-shutdown pattern to both GRPO `ThreadPoolExecutor` blocks, and add a test covering policy/generation startup failing after Gym succeeds.

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -324,6 +333,45 @@ def parse_shard_plan(nemo_gym_config: Mapping[str, Any]) -> ShardPlan | None:
...
+            assert shard.port_range_low is not None
+            assert shard.port_range_high is not None
+            ranges.append((shard.port_range_low, shard.port_range_high, shard.name))
+        ranges.sort()
+        for (_, previous_high, previous_name), (
+            current_low,
+            _,
+            current_name,
+        ) in zip(ranges, ranges[1:]):
+            if current_low <= previous_high:
```
macandro96: [`nemo_gym_shards.py:347-372`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/environments/nemo_gym_shards.py#L347-L372)

**1 action item.** PR-introduced. Valid in #3373 considered in isolation.

**TL;DR — the disjoint-port-range check compares one range per shard, not per shard x replica instance, so a shard with `replicas > 1` under relaxed placement (`PACK`/`STRICT_PACK`) can put multiple replicas in the same port window on one node.**

`apply_shard_overlay` stamps the same `port_range_low`/`port_range_high` onto every replica of a shard. This check only compares ranges across distinct shard *names*, so it never flags two replicas of the *same* shard sharing a window. Verified: a 3-replica shard under `PACK` with an explicit `port_range_low/high` passes validation with all 3 replicas sharing that exact range.

Correctly fixed downstream by #3374, which requires the default `STRICT_SPREAD` placement whenever `replicas` is configured -- so this is a gap in #3373 considered in isolation, not in the finished stack. Worth fixing here too since #3373 is mergeable on its own.

### AI-1
Expand the disjointness check to `(shard, replica_index)` pairs, or reject `PACK`/`STRICT_PACK` when any shard has `replicas > 1` without per-replica ranges. Suggested regression test (verified passing, `1 passed in 1.07s`), for `tests/unit/environments/test_nemo_gym_shards.py`:
```suggestion
def test_relaxed_placement_does_not_detect_port_collisions_across_a_shards_own_replicas():
    """Documents a gap: parse_shard_plan's disjoint-port check compares one
    range per *shard*, not per shard x replica instance. A shard with
    replicas > 1 under PACK reuses the same port_range_low/high for every
    replica, so replicas of the same shard can still collide with each other
    on one node even though this validation raises nothing.
    """
    plan = parse_shard_plan(
        _sharded_config(
            placement_strategy="PACK",
            shards=[
                {
                    "name": "judged",
                    "config_paths": ["judge.yaml"],
                    "replicas": 3,
                    "port_range_low": 5000,
                    "port_range_high": 5500,
                },
                {
                    "name": "tools",
                    "config_paths": ["tools.yaml"],
                    "port_range_low": 5501,
                    "port_range_high": 6000,
                },
            ],
        )
    )
    judged = next(s for s in plan.shards if s.name == "judged")
    assert judged.replicas == 3
    assert (judged.port_range_low, judged.port_range_high) == (5000, 5500)
```

nemo_rl/environments/nemo_gym_shards.py
```diff
@@ -39,6 +39,12 @@
 # The actor factory or rollout code consumes these separately.
 NEMO_RL_DICT_CONFIG_KEYS = frozenset({"effort_levels", "tokenizer_config"})
 
+# Ray placement-group strategies a shard plan may ask for. STRICT_SPREAD is the
+# point of sharding -- one actor per node -- and anything else colocates shards
+# and gives up the capacity isolation. PACK exists so the mechanism can be
+# exercised on a single machine.
```
macandro96, line 45: [`nemo_gym_shards.py:41-46`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/environments/nemo_gym_shards.py#L41-L46)

**1 action item.** Valid follow-up, not a blocker.

This PR introduces several user-facing configuration fields (`env.nemo_gym.shards`, `env.nemo_gym.placement_strategy`, `common_inherited_overlays`, `allowed_duplicate_entries`) but documents them only in the new smoke config's header comment (`examples/nemo_gym/grpo_sharded_gym_smoke.yaml`), not in `docs/`. Under this repo's docs guidelines, user-facing config surfaces warrant proper documentation.

### AI-1
Add a `docs/guides/` page covering `env.nemo_gym.shards` and `placement_strategy`, even a short one -- the smoke config's header is good raw material to start from.

nemo_rl/algorithms/distillation.py
```diff
@@ -538,9 +548,20 @@ def init_nemo_gym():
...
             )
-            with ThreadPoolExecutor(max_workers=len(init_tasks)) as executor:
-                submitted = {k: executor.submit(fn) for k, fn in init_tasks.items()}
-                results = {k: f.result() for k, f in submitted.items()}
+            submitted = {}
+            try:
+                with ThreadPoolExecutor(max_workers=len(init_tasks)) as executor:
+                    submitted = {k: executor.submit(fn) for k, fn in init_tasks.items()}
+                    results = {k: f.result() for k, f in submitted.items()}
+            except BaseException:
```
macandro96, line 556: [`distillation.py:552-564`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/algorithms/distillation.py#L552-L564)

**No action needed** -- context only, not a confirmed leak.

This recovery block tears down the gym shard set on sibling-task failure but doesn't also shut down `deferred_vllm`/the `"vllm"` future if gym is the one that fails. The asymmetry is untidy, but `VllmGeneration.__del__` ([`vllm_generation.py:1478-1484`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/models/generation/vllm/vllm_generation.py#L1478-L1484)) is an explicit safety net that calls `self.shutdown()` on garbage collection, and this asymmetry predates this PR. Explicit cleanup here would still be cleaner, but this isn't a confirmed permanent leak.

nemo_rl/data/interfaces.py
```diff
@@ -32,6 +32,33 @@
...
+
+    path: str
+    device: int
+    inode: int
+    ctime_ns: int
+    mtime_ns: int
+    size: int
+
+    @classmethod
+    def from_stat(cls, path: str, stat: os.stat_result) -> "NemoGymSourceIdentity":
```
macandro96, line 47: [`interfaces.py:47`](https://github.com/NVIDIA-NeMo/RL/blob/35eccf8294fa3708c6d358fec509c5f83ed0457b/nemo_rl/data/interfaces.py#L47)

**1 action item.** PR-introduced. Nit.

`NemoGymSourceIdentity.from_stat` is a public classmethod imported by 4 files; its sibling `.matches` (line 57) has a docstring, this doesn't.

### AI-1
```suggestion
    def from_stat(cls, path: str, stat: os.stat_result) -> "NemoGymSourceIdentity":
        """Build a source identity snapshot from a resolved path and its `os.stat` result."""
```

macandro96 (commented):
Team review of PR https://github.com/NVIDIA-NeMo/RL/pull/3373 (rl-expert, bug-finder, test-agent, design-reviewer, expert-gym, devil-advocate), refined against author feedback and re-verified directly against source where claims were contested.

4 blockers, all independently reproduced: (1) grpo.py currently fails to import -- duplicate num_generations_per_prompt= kwargs cause a SyntaxError; (2) NemoGymShardSet carries an unpicklable threading.Lock and is now reachable into SingleControllerActor.remote(...)'s cloudpickled constructor args, crashing every SingleController + NeMo-Gym run (sharded or not) at actor construction - can be fixed here or in the previous PR; (3) a test still patches the renamed-away spinup_nemo_gym_actor symbol and fails collection; (4) _combine_agent_name_sources reads a new attribute unconditionally, breaking any external/custom dataset class, gym or not.

4 lower-severity items below: an unbounded shutdown timeout on 4 new call sites, a resource leak in GRPO's (pre-existing but now more consequential) concurrent-init blocks, a type-hygiene note on cast(EnvironmentInterface, ...), and a missing docs page. One item (an asymmetric vLLM-cleanup path) is included as no-action context since VllmGeneration.__del__ already provides a safety net.

examples/nemo_gym/grpo_sharded_gym_smoke.yaml
```diff
@@ -0,0 +1,72 @@
...
+  nemo_gym:
+    # Retract inherited settings that no shard owns.
+    # A Hydra override can blank an inherited key but cannot delete it.
+    # The schema reads null as absent.
+    config_paths: null
+    code_gen: null
+
+    # Each shard gets its own port window so two stacks on one node cannot
+    # collide. The base recipe's 5000-5999 is split in half.
+    shards:
```
terrykong: **1 action item.**
**TL;DR — the port-isolation test now picks up this config but never reads these shard port ranges, so it looks like it covers sharding and does not.**

[`test_config.py:41`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/utils/test_config.py#L41) selects configs by content — `if "port_range" in path.read_text()` — and this file matches four times, so it is in `PORT_ISOLATION_CONFIG_PATHS`. But [`effective_range`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/utils/test_config.py#L306-L312) builds keys as `f"{prefix}.port_range_low"`, and the gym call at [`test_config.py:324`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/utils/test_config.py#L324) passes prefix `"env.nemo_gym"`. This config has no `env.nemo_gym.port_range_low` — its ranges live under `shards[*]` — so the check falls back to `DEFAULT_GYM_PORT_RANGE_LOW/HIGH` (5000/5999) and passes without reading a single shard.

The config itself is correct today: 5000-5499 and 5500-5999 both sit inside the gym window. The problem is the guard, not a live violation — a shard pinned into the generation window (3000-4999) would pass too.

### AI-1
Extend [`effective_range`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/utils/test_config.py#L306) (or its caller) to fold in every `env.nemo_gym.shards[*]` range, so each shard's low/high is checked against the generation and router windows. Not a suggestion block: the change is in [`tests/unit/utils/test_config.py`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/utils/test_config.py#L306), outside this PR's diff.

Separately, nothing loads this YAML through `load_config`, so the null-retraction path is only tested against a hand-built dict — the config was verified end to end, so that is a coverage gap rather than a bug.

examples/nemo_gym/grpo_sharded_gym_smoke.yaml
```diff
@@ -0,0 +1,72 @@
...
+#   nodes.
+
+defaults: "grpo_workplace_assistant_nemotron_nano_v2_9b.yaml"
+
+env:
+  nemo_gym:
+    # Retract inherited settings that no shard owns.
+    # A Hydra override can blank an inherited key but cannot delete it.
+    # The schema reads null as absent.
+    config_paths: null
```
terrykong: **1 action item.**
**TL;DR — the stack adds four user-facing config keys and the null-retraction rule with no docs entry anywhere; `grep -rn "nemo_gym.*shard\|placement_strategy" docs/` returns nothing.**

The new keys are `shards`, `placement_strategy`, `common_inherited_overlays`, and `allowed_duplicate_entries`, plus the rule that a shard's `config_paths` retracts the top-level one — which is exactly what this line does.

### AI-1
The target page already exists: [`docs/design-docs/nemo-gym-integration.md`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/docs/design-docs/nemo-gym-integration.md) already documents `env.nemo_gym.config_paths`, the key this config nulls out. So this is an edit to an existing page — no new file and no [`docs/index.md`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/docs/index.md) change. Much of the text already exists in this YAML's header comment and could be lifted.

#3372 added no docs either, so this is the stack's gap rather than only this PR's, and it is not a blocker — is a docs PR planned later in the stack? Not a suggestion block since the change lands in a different file.

nemo_rl/algorithms/grpo.py
```diff
@@ -812,17 +816,33 @@ def init_train_dataloader(dataset, suffix: str = ""):
...
             use_fastokens=bool(policy_config["tokenizer"].get("use_fastokens")),
         )
-        return actor, time.perf_counter() - t0
+        train_splits = (
+            {f"train[{name}]": split for name, split in dataset.items()}
+            if isinstance(dataset, dict)
+            else {"train": dataset}
+        )
+        try:
+            validate_dataset_agent_coverage(
```
terrykong: **1 action item.**
**TL;DR — of the three entrypoints this PR touches, grpo's Gym cleanup is both the least complete and the least tested: it is missing the sibling-failure recovery that distillation and SingleController got, and the one recovery path it does have is never run by a test.**

**Incomplete.** The "if a sibling init task fails, shut down the Gym shard set that already finished" block is present at [`distillation.py:559`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/algorithms/distillation.py#L559) and [`setup.py:1608`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/algorithms/single_controller_utils/setup.py#L1608), and missing at [`grpo.py:1448`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/algorithms/grpo.py#L1448) (megatron — its `finally` holds only `ray.kill(port_holder)`) and [`grpo.py:1606`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/algorithms/grpo.py#L1606) (vLLM — no `try` at all).

The PR description says "If generation, trainer, or coverage setup fails … the completed shard set is recovered from that future and shut down". True for the coverage half in all three entrypoints; true for the generation/trainer half in only two of three.

This is a consistency gap, not a leak. `setup()` is not wrapped in a `try` at either grpo caller, the driver process exits, and the actors are not `lifetime="detached"`, so Ray reclaims them — there is no case where the placement group outlives the process.

**Untested.** The coverage half grpo does have — the `except BaseException: shard_set.shutdown(); raise` at [`grpo.py:842-844`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/algorithms/grpo.py#L842-L844), directly below this line — is not reached by any test. All three grpo setup tests pass `MagicMock(is_sharded=False)` ([test_grpo.py:2939](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/algorithms/test_grpo.py#L2939), [:3298](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/algorithms/test_grpo.py#L3298), [:3412](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/algorithms/test_grpo.py#L3412)), which sends `validate_dataset_agent_coverage` down its [early return at `nemo_gym.py:2053-2054`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/environments/nemo_gym.py#L2053-L2054), so the handler never runs. Those tests are fine for what they assert — this is a gap in coverage, not a fault in them.

What makes it worth raising is that this PR adds exactly this test for SingleController: [`test_setup.py:1646`, `test_nemo_gym_coverage_failure_shuts_down_shards`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/single_controller/test_setup.py#L1646). The same PR that covered one entrypoint's teardown left the other's uncovered.

### AI-1
Two things:
- Either add the recover-and-shutdown block to grpo's two init sites, or narrow that sentence in the PR description.
- Add the missing coverage-failure test for grpo, mirroring SC's. This is small — parametrizing the existing trtllm setup test is enough; it does not need new scaffolding.

Not a suggestion block: the changes land in [`nemo_rl/algorithms/grpo.py`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/algorithms/grpo.py#L1448) and [`tests/unit/algorithms/test_grpo.py`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/tests/unit/algorithms/test_grpo.py#L3298), neither at this anchor. Two things anyone factoring the init blocks into a shared helper hits first: the megatron site uses named futures with no `submitted` dict, so it is not a copy-paste of the distillation block; and the three futures return different shapes (distillation returns a bare shard set, grpo and setup return `(shard_set, seconds)`).

nemo_rl/environments/nemo_gym.py
```diff
@@ -1825,16 +1835,17 @@ def _build_sharded_gym_actors(
             }
             for shard, _ in instances
         ],
-        strategy="STRICT_SPREAD",
+        strategy=plan.placement_strategy,
```
**TL;DR — making the strategy configurable here leaves the `placement_group` docstring wrong: it still promises STRICT_SPREAD, which is false under `PACK`.**

[`nemo_gym.py:1530-1531`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/environments/nemo_gym.py#L1530-L1531) reads "The STRICT_SPREAD group pinning shards to distinct nodes, or None when unsharded." Under `PACK` it is neither STRICT_SPREAD nor pinning shards to distinct nodes.

### AI-1
Replacement text for those two lines:

```python
        placement_group: The placement group holding the shards, created with
            `plan.placement_strategy`; None when unsharded.
```

Plain fence, not a suggestion block, because those lines are outside this PR's diff hunks.

nemo_rl/environments/nemo_gym.py
```diff
@@ -2016,3 +2027,112 @@ def spinup_nemo_gym_actor(
...
+            return source_agent_names
+
+    # AllTaskProcessedDataset wraps the raw rows; a plain sequence is also fine.
+    rows = getattr(dataset, "dataset", dataset)
+
+    names: set[str] = set()
+    for row in rows:
+        extra_env_info = row.get("extra_env_info") if hasattr(row, "get") else None
+        if isinstance(extra_env_info, str):
+            extra_env_info = json.loads(extra_env_info)
```
**TL;DR — the two paths that read the same file disagree about a malformed line: the cached path catches `json.JSONDecodeError`, this fallback lets it escape, which contradicts the function's `Raises:` block.**

The fast path at [`nemo_gym.py:2124`](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/environments/nemo_gym.py#L2124) does `except (OSError, json.JSONDecodeError): return None` — it expects a bad line. This row-scan parse has no guard, and [`validate_dataset_agent_coverage`'s `Raises:` block](https://github.com/NVIDIA-NeMo/RL/blob/0996e3655ea83e43ae1a5cfaed4a14c085beeb27/nemo_rl/environments/nemo_gym.py#L2050-L2051) lists only `ShardSetupError`.

It is reachable: `NemoGymDataset.__init__` stores raw lines without parsing them, so a bad line survives to this scan, and the fast path returning `None` for any reason routes straight here. The realistic source is a truncated file from an interrupted `gym dataset collate` run, since Gym writes it line by line — that part is thin. The docstring mismatch is the solid half.

### AI-1
Guard this parse the way the fast path does and re-raise as `ShardSetupError` naming the split and row index, or widen the `Raises:` block to admit `json.JSONDecodeError`. Not a suggestion block: the surrounding `for` loop has no split/row-index variable to name in the error yet, so the fix needs more than this one line.

nemo_rl/algorithms/grpo.py
ananthsub: Resolved. The duplicate training `num_generations_per_prompt` argument was removed in the lower router layer so both PR #3372 and this cumulative PR compile independently. The remaining argument uses `master_config.grpo.num_generations_per_prompt`.

nemo_rl/algorithms/grpo.py
ananthsub: Resolved. The duplicate validation argument was removed, and the retained value is `val_num_generations_per_prompt`. That value matches the factor used to repeat the validation batch and therefore the prompt groups used for shard routing.

nemo_rl/algorithms/single_controller_utils/setup.py
ananthsub: Resolved in PR #3374. The shard set now defines pickle state hooks that omit the process-local lock and recreate it after deserialization. This allows both unsharded and sharded `NemoGymShardSet` instances to cross the `SingleControllerActor.remote()` boundary.

tests/unit/single_controller/test_setup.py
ananthsub: Resolved. `test_periodic_checkpointing_warns_without_per_step_trainer_anchors` now patches `build_nemo_gym_actors`, matching the factory imported and called by `single_controller_utils.setup`. The focused test passes with the renamed factory.

nemo_rl/data/utils.py
ananthsub: Resolved. All new `agent_name_sources` reads now use `getattr(..., None)`, including separate training datasets and both validation-source paths. External dataset classes that implement the documented dataset contract without this optional metadata continue to load. A regression test covers a dataset object with no `agent_name_sources` attribute.

nemo_rl/algorithms/grpo.py
ananthsub: Resolved by making `NemoGymShardSet.shutdown()` bounded and forceful by default. Callers now use `NEMO_GYM_GRACEFUL_SHUTDOWN_TIMEOUT_S` unless they explicitly override it, and remaining actor handles are force-killed after the graceful attempt. The teardown test verifies both the propagated timeout and forced-kill behavior.

nemo_rl/algorithms/grpo.py
ananthsub: Resolved for both concurrent GRPO startup paths. If policy or generation startup raises, the error path now retrieves a successfully completed Gym future and shuts down its `NemoGymShardSet`; if Gym startup itself failed, its original exception remains untouched. A regression test verifies teardown of a completed Gym startup future after a sibling failure.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Resolved. Added `docs/guides/nemo-gym-sharding.md` and linked it from `docs/index.md`. The guide documents shard configuration, inherited overlays, allowed duplicates, replicas, port ranges, and the placement-strategy tradeoff between the default `STRICT_SPREAD` and single-node `PACK` testing.

nemo_rl/algorithms/distillation.py
ananthsub: Acknowledged. No change was made for this context-only note because the existing `VllmGeneration.__del__` safety net covers this pre-existing asymmetry, and the comment did not identify a confirmed leak introduced by this stack.

nemo_rl/data/interfaces.py
ananthsub: Resolved. `NemoGymSourceIdentity.from_stat` now has a Google-style docstring describing that it builds the source identity from a resolved path and its stat result.

nemo_rl/environments/nemo_gym_shards.py
ananthsub: Updated the placement of this documentation: the sharding section now lives in the existing `docs/design-docs/nemo-gym-integration.md` design document. The temporary standalone guide and its `docs/index.md` entry were removed.

docs/design-docs/nemo-gym-integration.md
```diff
@@ -49,6 +49,55 @@ Gym result payloads are needed for a short debugging run.
...
+
+For a single-node test, set:
+
+```yaml
+env:
+  nemo_gym:
+    placement_strategy: PACK
+```
+
+`PACK` validates configuration, routing, and teardown on one machine, but it does not provide node-level capacity isolation. The other supported Ray strategies are `SPREAD` and `STRICT_PACK`.
```
macandro96: [`nemo-gym-integration.md:97`](https://github.com/NVIDIA-NeMo/RL/blob/2e9d5d3267f54417c43ada56640349810a7cdedc/docs/design-docs/nemo-gym-integration.md#L97)

**1 action item, low priority.**

PR-introduced. This section doesn't note that `replicas > 1` requires the default `STRICT_SPREAD` -- #3374's `parse_shard_plan` rejects it under `PACK`, `SPREAD`, and `STRICT_PACK`, since replicas share one merged config and port range. Someone following this section's own `PACK` example two paragraphs after `replicas` is introduced could reasonably combine them and hit a runtime `ShardConfigError`.

### AI-1
```suggestion
`PACK` validates configuration, routing, and teardown on one machine, but it does not provide node-level capacity isolation. The other supported Ray strategies are `SPREAD` and `STRICT_PACK`. Shards with `replicas > 1` require the default `STRICT_SPREAD` -- replicas share one merged config and port range, so any other strategy risks colocating them.
```

docs/design-docs/nemo-gym-integration.md
```diff
@@ -49,6 +49,55 @@ Gym result payloads are needed for a short debugging run.
...
+
+See `examples/nemo_gym/grpo_sharded_gym_smoke.yaml` for a complete manual smoke configuration.
```
macandro96, line 99: [`nemo-gym-integration.md:99`](https://github.com/NVIDIA-NeMo/RL/blob/2e9d5d3267f54417c43ada56640349810a7cdedc/docs/design-docs/nemo-gym-integration.md#L99)

**1 action item, follow-up.**

`examples/nemo_gym/grpo_sharded_gym_smoke.yaml` isn't referenced by any `tests/functional/*.sh` or `tests/test_suites/*.txt` -- the NeMo-Gym actor-sharding feature across this stack (#3372/#3373/#3374) has no CI execution at all, only unit tests against mocked actors. (The existing `shard` hits in `tests/functional/grpo_sc_generation_shard_recovery.sh` are the pre-existing, unrelated vLLM generation-fleet sharding concept.)

### Follow-up
A one-node, two-shard `PACK` functional test (real Ray actors, real ports, this smoke config as the starting point) would catch regressions the unit tests can't -- e.g. actual placement-group teardown, port collisions, real Gym server startup.

macandro96 (commented):
Follow-up on my 2026-09-10 team review: verified all 10 threads against the current head (replies to follow after this review is submitted). 9 of 10 hold up; 1 does not -- the threading.Lock pickling fix only exists in #3374, not in this PR's own tree, and this PR is still OPEN/MERGEABLE, so the crash is live if merged standalone. Also adding 2 new items surfaced during verification: a design-doc gap and a CI coverage gap for the whole sharding stack.

_Generated by Claude Code_

nemo_rl/algorithms/grpo.py
macandro96: fixed in 2e9d5d3267f54417c43ada56640349810a7cdedc

nemo_rl/algorithms/grpo.py
macandro96: fixed in 2e9d5d3267f54417c43ada56640349810a7cdedc

tests/unit/single_controller/test_setup.py
macandro96: fixed in 2e9d5d3267f54417c43ada56640349810a7cdedc

nemo_rl/data/utils.py
macandro96: fixed in 2e9d5d3267f54417c43ada56640349810a7cdedc

nemo_rl/algorithms/grpo.py
macandro96: fixed in 2e9d5d3267f54417c43ada56640349810a7cdedc

nemo_rl/algorithms/grpo.py
macandro96: fixed in 2e9d5d3267f54417c43ada56640349810a7cdedc

nemo_rl/environments/nemo_gym_shards.py
macandro96: fixed in 2e9d5d3267f54417c43ada56640349810a7cdedc

nemo_rl/data/interfaces.py
macandro96: fixed in 2e9d5d3267f54417c43ada56640349810a7cdedc

nemo_rl/algorithms/single_controller_utils/setup.py
macandro96: Reopening -- this is still broken in #3373 standalone, verified against 2e9d5d3267f54417c43ada56640349810a7cdedc (this PR's current head, not #3374's):

```
>>> cloudpickle.dumps(NemoGymShardSet(handles={"a": [object()]}))
TypeError: cannot pickle '_thread.lock' object
```

nemo_gym.py has no __getstate__/__setstate__ in this PR's own tree -- that fix only exists in #3374. setup.py:1657 still flows the shard set into SingleControllerActor.remote(...)'s cloudpickled constructor args. #3373 is currently OPEN/MERGEABLE (not merged), so if it merges to main before #3374 lands, every SingleController + NeMo-Gym run crashes at actor construction in the interim.

Could you fold the pickle-state fix into this PR directly, or otherwise make sure #3373 and #3374 land together rather than #3373 merging alone first?

docs/design-docs/nemo-gym-integration.md
ananthsub: Resolved in the dependent replication PR, #3374. The integration design now states that `replicas > 1` requires `STRICT_SPREAD` because replicas share a merged configuration and port range and therefore cannot safely be colocated.

docs/design-docs/nemo-gym-integration.md
ananthsub: Resolved. Added `tests/functional/grpo_sharded_gym_pack.sh` and registered it in the full `L1_Functional_Tests_Gym` suite. It reuses the existing real Gym GRPO test with `grpo_sharded_gym_smoke.yaml` and `placement_strategy=PACK`, so CI exercises two real Gym actors, their port ranges, routing setup, and teardown on one node. It is intentionally not part of `CI:Lfast`.

ananthsub:
Addressed both points from the follow-up review. The lock-pickling fix now lives in the lower dependency, [PR #3372](https://github.com/NVIDIA-NeMo/RL/pull/3372), where `_replica_lock` is introduced. `NemoGymShardSet` removes the lock during serialization and recreates it after deserialization

nemo_rl/algorithms/single_controller_utils/setup.py
ananthsub: Follow-up: moved the lock-pickling fix into the lower dependency, #3372. This PR now inherits serialization-safe `NemoGymShardSet` behavior before it passes the shard set through `SingleControllerActor.remote()`.

macandro96: fixed in 665dac1494ea6963f017339fc24f6d46ac84c250

docs/design-docs/nemo-gym-integration.md
macandro96: fixed in 1d0634ecf739d260fb3f3a11ed284f4a888b15f7

docs/design-docs/nemo-gym-integration.md
macandro96: fixed in 1d0634ecf739d260fb3f3a11ed284f4a888b15f7

ananthsub: Follow-up: the sharded Gym functional test is now marked `fast` in `L1_Functional_Tests_Gym.sh`, so it runs under the PRs’ `CI:Lfast` label as well as the full L1 suite.

macandro96 (dismissed):

terrykong (dismissed):

terrykong (approved):

PR #3374, opened 2026-07-27
fix(nemo-gym): preserve replica identity during routing

ananthsub:
## Summary

Preserves physical actor identity when one logical shard has multiple replicas.

Replica expansion and round-robin dispatch already use multiple actors. Other paths still keyed results by logical shard name, so one replica could overwrite another replica's timing metrics and failures could identify only the shard.

`NemoGymShardSet.instance_label()` now uses the shard name when it has one replica and `shard/index` when it has several. Actor timing metrics, log directories, and stream failures use the same label. Routing metrics zero-fill every actor instance for each prompt group, so normal metric aggregation reports the fraction of groups routed to each replica. Stream validation keeps local input errors as `ValueError`, while transport failures identify the actor instance that failed.

Replica selection remains round-robin. A future least-in-flight policy requires callers to acquire and release a dispatch lease so completed work can decrement the selected actor's load. This PR does not claim that lifecycle or hide it behind the current handle-only API.

The schema rejects `replicas > 1` when `placement_strategy` is not `STRICT_SPREAD`. Replicas share one merged configuration and one port range. Placing two replicas on the same node would let them race for the same ports. Colocated instances must instead be separate shards with separate port ranges.

## Example

```yaml
env:
  nemo_gym:
    shards:
      - name: tools
        replicas: 3
        config_paths:
          - resources_servers/workplace_assistant/configs/workplace_assistant.yaml
```

These instances are reported as `tools/0`, `tools/1`, and `tools/2`.

## Dependency

This PR targets the branch from [PR #3373](https://github.com/NVIDIA-NeMo/RL/pull/3373).

### DIFF
--- a/examples/nemo_gym/grpo_sharded_gym_smoke.yaml
+++ b/examples/nemo_gym/grpo_sharded_gym_smoke.yaml
@@ -57,6 +57,14 @@ env:
           - responses_api_models/vllm_model/configs/vllm_model_for_training.yaml
           - resources_servers/math_with_judge/configs/math_with_judge.yaml
 
+        # `replicas: N` makes N identical instances of a shard, one per node,
+        # and the router hands each prompt group to one of them round-robin.
+        # Reach for it when a single shard is what saturates a node. It needs
+        # N more nodes and requires the default STRICT_SPREAD, so it cannot be
+        # exercised by this config: replicas share one merged config and hence
+        # one port range, which only holds up when each gets its own node.
+        # replicas: 2
+
     # Both shards need policy_model, which is a client of NeMo RL's vLLM
     # servers rather than an engine of its own, so hosting it twice claims no
     # extra GPU. That is what the allowlist below is for; without it the
--- a/tests/unit/environments/test_nemo_gym_shards.py
+++ b/tests/unit/environments/test_nemo_gym_shards.py
@@ -177,6 +177,22 @@ def test_an_unknown_placement_strategy_is_rejected():
         parse_shard_plan(_sharded_config(placement_strategy="SPRED"))
 
 
+def test_replicas_that_could_share_a_node_would_share_a_port_range():
+    """Replicas come from one merge, so they cannot be given separate ranges."""
+    config = _sharded_config(placement_strategy="PACK")
+    config["shards"][1]["replicas"] = 2
+
+    with pytest.raises(ShardConfigError, match=r"\['tools'\] declare replicas"):
+        parse_shard_plan(config)
+
+
+def test_replicas_are_fine_on_the_default_strategy_that_spreads_them():
+    config = _sharded_config()
+    config["shards"][1]["replicas"] = 2
+
+    assert parse_shard_plan(config).total_instances == 3
+
+
 def test_allowed_duplicate_entries_must_be_strings():
     with pytest.raises(ShardConfigError, match="list of strings"):
         parse_shard_plan(_sharded_config(allowed_duplicate_entries=[{"nope": 1}]))
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -883,11 +883,18 @@ class NemoGymShardSet:
         self._next_replica[shard_name] = (index + 1) % len(replicas)
         return replicas[index]
 
-    def shard_name_of(self, handle: Any) -> str:
-        """Reverse-look up a handle's shard, for error messages and metric tags."""
+    def instance_label(self, handle: Any) -> str:
+        """Name one actor, for error messages and metric keys.
+
+        A shard with one replica is named by the shard alone, so the common
+        case reads as it did before replicas existed; a replicated shard adds
+        the replica index. This is the same rule the per-shard log directories
+        follow, so a metric and its logs carry the same name.
+        """
         for shard_name, replicas in self.handles.items():
-            if any(replica is handle for replica in replicas):
-                return shard_name
+            for index, replica in enumerate(replicas):
+                if replica is handle:
+                    return shard_name if len(replicas) == 1 else f"{shard_name}/{index}"
         raise ShardSetupError("Handle does not belong to this NeMo-Gym shard set")
 
     def shutdown(self) -> None:
@@ -1040,7 +1047,6 @@ def _build_sharded_gym_actors(
             f"{DEFAULT_PLACEMENT_STRATEGY}, so shards may share a node and the "
             f"per-node capacity isolation sharding exists for does not hold."
         )
-
     base_gym_dict = {
         key: value
         for key, value in nemo_gym_dict.items()
--- a/nemo_rl/environments/nemo_gym_shards.py
+++ b/nemo_rl/environments/nemo_gym_shards.py
@@ -253,6 +253,23 @@ def parse_shard_plan(nemo_gym_config: Mapping[str, Any]) -> ShardPlan | None:
             f"{sorted(PLACEMENT_STRATEGIES)}, got {placement_strategy!r}"
         )
 
+    # Replicas are stamped from one merge, so they share that shard's port
+    # range and no per-replica override exists to give them separate ones.
+    # Only STRICT_SPREAD guarantees each one its own node, and therefore its
+    # own port space; under any other strategy two replicas can land together
+    # and race for the same ports at spinup.
+    replicated = [shard.name for shard in shards if shard.replicas > 1]
+    if replicated and placement_strategy != DEFAULT_PLACEMENT_STRATEGY:
+        raise ShardConfigError(
+            f"Shards {replicated} declare replicas, which requires "
+            f"placement_strategy {DEFAULT_PLACEMENT_STRATEGY} so that each "
+            f"replica gets its own node and port space; got "
+            f"{placement_strategy!r}. Replicas share one merged config, so "
+            f"they cannot be given separate port ranges. Either drop the "
+            f"replicas or split them into separate shards with their own "
+            f"port_range_low/high."
+        )
+
     raw_allowed = nemo_gym_config.get("allowed_duplicate_entries") or []
     if OmegaConf.is_config(raw_allowed):
         raw_allowed = OmegaConf.to_container(raw_allowed, resolve=True)
--- a/nemo_rl/experience/rollouts.py
+++ b/nemo_rl/experience/rollouts.py
@@ -1973,15 +1973,18 @@ def _prepare_nemo_gym_rows(
         row["_rowidx"] = row_index
 
 
-def _bucket_nemo_gym_rows_by_shard(
+def _bucket_nemo_gym_rows_by_instance(
     rows: list[dict],
     shard_set: NemoGymShardSet,
     num_generations: int,
 ) -> list[tuple[str, Any, list[dict]]]:
     """Split already-stamped rows into one bucket per actor, by prompt group.
 
-    Bucketing is by group and never by row. A group split across shards would
-    deadlock any verifier that scores rollouts against their peers: those
+    Buckets are per actor instance, not per shard: a shard with replicas gets
+    one bucket per replica, which is how replication spreads a step's load.
+
+    Bucketing is by group and never by row. A group split across instances
+    would deadlock any verifier that scores rollouts against their peers: those
     buffer a whole group in one process and block until it is complete, so each
     half would wait forever for rollouts sitting in the other.
 
@@ -1989,8 +1992,9 @@ def _bucket_nemo_gym_rows_by_shard(
     accumulator downstream never learns that shards exist.
 
     Returns:
-        ``(shard_name, handle, rows)`` per actor that has work, in first-use
-        order. Actors with no rows are omitted rather than sent an empty batch.
+        ``(instance_label, handle, rows)`` per actor that has work, in
+        first-use order. Actors with no rows are omitted rather than sent an
+        empty batch.
     """
     buckets: dict[int, tuple[str, Any, list[dict]]] = {}
     for start in range(0, len(rows), num_generations):
@@ -2005,12 +2009,12 @@ def _bucket_nemo_gym_rows_by_shard(
             )
         handle = shard_set.pick_handle(next(iter(agent_names)))
         if id(handle) not in buckets:
-            buckets[id(handle)] = (shard_names.pop(), handle, [])
+            buckets[id(handle)] = (shard_set.instance_label(handle), handle, [])
         buckets[id(handle)][2].extend(group)
     return list(buckets.values())
 
 
-async def _merge_nemo_gym_shard_streams(
+async def _merge_nemo_gym_instance_streams(
     buckets: list[tuple[str, Any, list[dict]]],
     tokenizer: TokenizerType,
     timer_prefix: str,
@@ -2023,15 +2027,15 @@ async def _merge_nemo_gym_shard_streams(
     that one stream, which is what keeps the unsharded path unchanged.
 
     Yields:
-        ``(row_index, result, timing_metrics, shard_name)``. ``timing_metrics``
-        is non-None only on the last row of each shard's bucket.
+        ``(row_index, result, timing_metrics, instance_label)``.
+        ``timing_metrics`` is non-None only on the last row of each bucket.
     """
     iterators = {}
-    for shard_name, handle, rows in buckets:
+    for instance_label, handle, rows in buckets:
         stream = handle.run_rollouts.options(num_returns="streaming").remote(
             rows, tokenizer, timer_prefix
         )
-        iterators[stream.__aiter__()] = shard_name
+        iterators[stream.__aiter__()] = instance_label
 
     pending = {
         asyncio.ensure_future(anext(iterator)): iterator for iterator in iterators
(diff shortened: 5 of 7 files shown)


nemo_rl/experience/rollout_manager.py
```diff
@@ -944,6 +944,8 @@ async def _run_rollouts(
         # They share an agent and must stay on one instance.
         shard_set = as_nemo_gym_shard_set(self._task_to_env["nemo_gym"])
         nemo_gym_env = shard_set.pick_handle(inputs[0]["agent_ref"]["name"])
+        instance_label = shard_set.instance_label(nemo_gym_env)
+        instance_timer_prefix = f"{timer_prefix}/shard/{instance_label}"
```
terrykong: **1 action item.**

**TL;DR — the Gym actor builds its own timing key names from the prefix it is handed, so adding `/shard/<label>` renames three existing series instead of adding copies of them. On the default unsharded config `train/timing/rollout/await_results` and its two siblings go permanently flat.** PR-introduced; observability, not training correctness.

Line 948 builds `instance_timer_prefix`, and [line 1001](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollout_manager.py#L1001) forwards it to `_stream_rows`, which [hands it to the actor](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollout_manager.py#L904). The actor names its keys from that string ([await_results](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/environments/nemo_gym.py#L624), [postprocess_results_pct](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/environments/nemo_gym.py#L658)), so the renaming happens inside the actor. [Line 1060](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollout_manager.py#L1060) copies those keys in as they are, and nothing rebuilds the un-prefixed name.

With no `shards:` block at all, [`as_nemo_gym_shard_set` wraps the bare handle](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/environments/nemo_gym.py#L1311) as one shard named `nemo_gym`, so the label is `nemo_gym` and three series move:

```
timing/rollout/await_results            -> timing/rollout/shard/nemo_gym/await_results
timing/rollout/postprocess_results      -> timing/rollout/shard/nemo_gym/postprocess_results
timing/rollout/postprocess_results_pct  -> timing/rollout/shard/nemo_gym/postprocess_results_pct
```

They reach W&B and TensorBoard through [`log_metrics(step_metrics, ..., prefix="train")`](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/algorithms/single_controller.py#L2200), so `train/timing/rollout/await_results` and its two siblings stop updating. There is no error and no failing test: [the stub at `test_rollouts.py:2203`](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/tests/unit/experience/test_rollouts.py#L2203) throws away `timer_prefix`, so nothing observes what reaches the actor. Changing it to `del inputs` plus `assert timer_prefix == "timing/test"` pins this in one line.

The sibling [`_merge_nemo_gym_timing_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollouts.py#L2470-L2481) emits both the per-instance key and the un-prefixed roll-up, and its docstring in this PR says "With one instance this returns exactly what it reported, so the unsharded path is unchanged." The Single Controller path [did the same at the base SHA](https://github.com/NVIDIA-NeMo/RL/blob/f93ba52988756807f0ede4336ca3e75e463cd18f/nemo_rl/experience/rollout_manager.py#L995) and now does not.

### AI-1 — send the plain prefix to the actor, then add the per-instance copies afterwards

Not a `suggestion` block: this touches two regions, and applying either half alone would be wrong (reverting the forward on its own drops the per-instance keys the PR wants).

```python
# rollout_manager.py:994-1001 -- send the plain prefix again
timing_metrics = await self._stream_rows(
    nemo_gym_env, pending, results, total_rows, timer_prefix
)

# rollout_manager.py:1060 -- then add the per-instance copies here
rollout_metrics.update(env_timing_metrics)
for key, value in env_timing_metrics.items():
    suffix = key[len(timer_prefix) + 1 :] if key.startswith(timer_prefix) else key
    rollout_metrics[f"{timer_prefix}/shard/{instance_label}/{suffix}"] = value
```

`instance_timer_prefix` on line 948 then has no other use and can be removed. Do not reuse `_merge_nemo_gym_timing_metrics` as a shared helper here — it short-circuits when there is one instance, and this path only ever has one instance per call, so it would never emit a per-instance key.

nemo_rl/experience/rollout_manager.py
```diff
@@ -1048,6 +1058,7 @@ async def _run_rollouts(
             )
 
         rollout_metrics.update(env_timing_metrics)
+        rollout_metrics[f"{timer_prefix}/routing/groups/{instance_label}"] = 1
```
terrykong: **1 action item.**

**TL;DR — this writes a constant `1`, and each `_run_rollouts` call handles one prompt group, so after averaging every instance reports exactly 1.0 on every step no matter how lopsided the split. A metric named `groups` reads as a count but is not one.** PR-introduced; observability, not training correctness.

Groups that went to a different instance leave the key out entirely, so the [per-name list](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/algorithms/single_controller.py#L2032-L2036) only ever holds `1`s, and `aggregate_rollout_metrics` takes [the mean](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/algorithms/grpo.py#L4362).

Worked example — shard `busy` with `replicas: 2`, 100 prompt groups, 90 to replica 0 and 10 to replica 1. Logged:

```
timing/rollout/routing/groups/busy/0 = 1.0
timing/rollout/routing/groups/busy/1 = 1.0
```

### AI-1 — record a 1.0/0.0 share per instance so the mean becomes the real split

One way to do it — several shapes work, since the key name is a choice.

```suggestion
        for handle in shard_set.all_handles:
            label = shard_set.instance_label(handle)
            rollout_metrics[f"{timer_prefix}/shard/{label}/routing_group_share"] = float(
                handle is nemo_gym_env
            )
```

Each group then contributes one 1.0 and the rest 0.0, so the mean is each instance's share of the step's groups and the shares sum to 1.0. Uses [`all_handles`](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/environments/nemo_gym.py#L1213). Putting the label mid-key also matches [the sibling layout](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollouts.py#L2480) and keeps a shard name out of the last key segment, where the aggregator's suffix rules live.

nemo_rl/experience/rollout_manager.py
```diff
@@ -1002,6 +1008,9 @@ async def _run_rollouts(
                             classify_rollout_failure(error) is not FailureClass.INFRA
                             or attempt == self._max_gym_row_attempts
                         ):
+                            error.add_note(
```
terrykong: **1 action item.**

**TL;DR — `error.add_note(...)` only runs inside `except Exception`, so a timed-out prompt group never gets the instance name: `asyncio.timeout` cancels with `CancelledError`, which is a `BaseException` and is not caught.** Low severity. Adjacent rather than something this PR broke — the PR builds the attribution and leaves this one exit uncovered.

When [the deadline](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollout_manager.py#L979) expires, [`_Deadline.__aexit__`](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollout_manager.py#L365-L374) turns the cancellation into `RolloutTimeout`. The operator sees `RolloutTimeout: NeMo-Gym prompt group exceeded 600.0s` with no instance name — the case where knowing which replica hung matters most. The other two exits from `_run_rollouts` are covered: this re-raise, and [the missing-rows `GymTransportError`](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollout_manager.py#L1021).

### AI-1 — put the instance label in the deadline description

`_Deadline` interpolates `description` straight into its message, so this is a one-line change at [line 979](https://github.com/NVIDIA-NeMo/RL/blob/020a5d3ce73081f4558df728a98c79333e1962d8/nemo_rl/experience/rollout_manager.py#L979). Not a `suggestion` block, because that line is not in this PR's diff.

```python
async with _Deadline(
    self._timeouts.rollout_s, f"NeMo-Gym prompt group on instance '{instance_label}'"
):
```

nemo_rl/experience/rollout_manager.py
```diff
@@ -1364,6 +1370,7 @@ async def _run_rollouts(
             )
 
         rollout_metrics.update(env_timing_metrics)
+        rollout_metrics[f"{timer_prefix}/routing/groups/{instance_label}"] = 1
```
macandro96: [`rollout_manager.py:1373`](https://github.com/NVIDIA-NeMo/RL/blob/563d477aaf6e540b5e79d36a62979c1737faeec4/nemo_rl/experience/rollout_manager.py#L1373)

**1 action item.**

PR-introduced. `rollout_metrics[f"{timer_prefix}/routing/groups/{instance_label}"] = 1` always logs `1`. `aggregate_rollout_metrics` averages any key that isn't `/min`, `/max`, or `total_turns`; since only the selected instance's key appears per group, every list built for this metric is all-`1`s, so `routing/groups/<label>` is always `1.0` regardless of how the step's groups actually split across replicas.

### AI-1
Pick one semantic and implement it:
- **Share** (recommended): zero-fill every instance via `shard_set.all_handles` so the average reflects each instance's true fraction of groups. Consider renaming to `routing/group_share/<instance>` so the name matches what it reports.
- **Count**: keep it a raw count by aggregating with `sum()` instead of the default average -- this needs a new convention in `aggregate_rollout_metrics` (e.g. a `/count` suffix), since only exact-name and `/min`/`/max` are special-cased today.

Either way, add a test with 3-4 groups across two replicas asserting a non-1.0 result -- that's the case the current code silently gets wrong.

```suggestion
        for handle in shard_set.all_handles:
            rollout_metrics[
                f"{timer_prefix}/routing/groups/{shard_set.instance_label(handle)}"
            ] = 0
        rollout_metrics[f"{timer_prefix}/routing/groups/{instance_label}"] = 1
```
_one option -- of several (the share semantic; the alternative is a `sum`-based count, see above)_

tests/unit/experience/test_rollouts.py
```diff
@@ -2466,6 +2475,59 @@ def test_prepare_nemo_gym_rows_stamps_distinct_legacy_prompt_groups():
     assert [row["_rowidx"] for row in rows] == [0, 1, 2, 3]
 
 
+def test_rollout_manager_attributes_awaited_stream_failure_to_instance():
```
macandro96, line 2542: [`test_rollouts.py:2478`](https://github.com/NVIDIA-NeMo/RL/blob/563d477aaf6e540b5e79d36a62979c1737faeec4/tests/unit/experience/test_rollouts.py#L2478)

**1 action item.**

No test threads a real multi-replica `NemoGymShardSet` through `AsyncNemoGymRolloutImpl._run_rollouts` end-to-end -- existing manager tests in this file (including this one, and `test_rollout_manager_consumes_stream_and_restores_input_order` at line 2239) stub `_task_to_env["nemo_gym"]` as a single bare actor, so round-robin dispatch and per-instance metric labels are only exercised at the bucketing-helper level (`test_nemo_gym_router.py`), never through the manager itself.

### AI-1
Add a test that builds the manager with a real two-replica `NemoGymShardSet`, calls `_run_rollouts` twice (one per group), and asserts the two calls pick different replicas and emit distinct instance-labeled metrics.

tests/unit/environments/test_nemo_gym_shards.py
```diff
@@ -450,6 +450,22 @@ def test_an_unknown_placement_strategy_is_rejected():
         parse_shard_plan(_sharded_config(placement_strategy="SPRED"))
 
 
+def test_replicas_that_could_share_a_node_would_share_a_port_range():
```
macandro96: [`test_nemo_gym_shards.py:453`](https://github.com/NVIDIA-NeMo/RL/blob/563d477aaf6e540b5e79d36a62979c1737faeec4/tests/unit/environments/test_nemo_gym_shards.py#L453)

**1 action item, low priority.**

`test_replicas_that_could_share_a_node_would_share_a_port_range` only exercises `PACK`; `PLACEMENT_STRATEGIES` also has `SPREAD` and `STRICT_PACK`. The guard is a single `placement_strategy != DEFAULT_PLACEMENT_STRATEGY` check with no per-strategy branching, so this mainly guards against a future per-strategy special case rather than a live gap today.

### AI-1
Optional: parametrize this test over `["PACK", "SPREAD", "STRICT_PACK"]`.

examples/nemo_gym/grpo_sharded_gym_smoke.yaml
```diff
@@ -54,6 +54,14 @@ env:
           - responses_api_models/vllm_model/configs/vllm_model_for_training.yaml
           - resources_servers/math_with_judge/configs/math_with_judge.yaml
 
+        # `replicas: N` makes N identical instances of a shard.
+        # Each instance runs on one node.
+        # The router assigns prompt groups round-robin.
+        # Replication needs N more nodes and the default STRICT_SPREAD.
```
macandro96: [`grpo_sharded_gym_smoke.yaml:60`](https://github.com/NVIDIA-NeMo/RL/blob/563d477aaf6e540b5e79d36a62979c1737faeec4/examples/nemo_gym/grpo_sharded_gym_smoke.yaml#L60)

**No action needed** -- technically correct relative to the "Single node" section just above it, but the baseline it's "more than" is easy to miss on a skim.

```suggestion
        # With STRICT_SPREAD, this full configuration needs N+1 nodes total --
        # N more than the one-node PACK setup described above.
```

macandro96 (commented):
Delta review on top of the earlier pass (previous critical test bug and the __getstate__/__setstate__ pickling fix are both confirmed fixed and verified locally -- 238 unit tests pass, linters + pyrefly pass). Remaining items below are a routing-metric semantics gap, two test coverage gaps, and one optional doc clarity suggestion.

Generated by Claude Code

nemo_rl/experience/rollout_manager.py
ananthsub: Resolved with the share semantic. `_run_rollouts` now emits `routing/group_share/<instance>` for every actor instance, zero-filling unselected instances and setting the selected one to 1. The existing average aggregation therefore reports each instance’s actual fraction of prompt groups. A three-group/two-replica test verifies shares of 2/3 and 1/3.

tests/unit/experience/test_rollouts.py
ananthsub: Resolved. Added an `AsyncNemoGymRolloutImpl._run_rollouts` test backed by a real two-replica `NemoGymShardSet`. Three prompt groups rotate across both handles, emit distinct `tools/0` and `tools/1` metric keys, and aggregate to the expected per-instance shares.

tests/unit/environments/test_nemo_gym_shards.py
ananthsub: Resolved. The replica-placement rejection test is now parameterized over `PACK`, `SPREAD`, and `STRICT_PACK`, covering every supported strategy other than the required `STRICT_SPREAD`.

examples/nemo_gym/grpo_sharded_gym_smoke.yaml
ananthsub: Updated for clarity even though no code change was required. The smoke config now says that the complete two-shard configuration with `replicas: N` needs `N+1` nodes under `STRICT_SPREAD`, which is `N` more than the one-node `PACK` setup.

nemo_rl/experience/rollout_manager.py
macandro96: fixed in 42d82197b42d3acc6f46ffba51b97e63608b790b

tests/unit/experience/test_rollouts.py
macandro96: fixed in 42d82197b42d3acc6f46ffba51b97e63608b790b

tests/unit/environments/test_nemo_gym_shards.py
macandro96: fixed in 42d82197b42d3acc6f46ffba51b97e63608b790b

macandro96 (dismissed):

terrykong (dismissed):

terrykong (approved):

PR #3378, opened 2026-07-27
feat(sc): support router replay with TQ

### LINKED ISSUE #3327: [Single Controller] Router replay (R3) not supported/validated on the async + TransferQueue path
### Summary

Router replay (R3) is **partially wired but not supported end-to-end or validated** on the Single Controller (SC) async + TransferQueue (TQ) path, and there is no fail-loud guard — so enabling it runs silently with incorrect routing.

The SC path is exactly the "async + TQ data plane" combination that the legacy path explicitly blocks:

https://github.com/NVIDIA-NeMo/RL/blob/850b58377eb6ed54716419a90b75d749907ad063/nemo_rl/algorithms/grpo.py#L3539-L3546

```
policy.router_replay.enabled=true with async GRPO is currently supported only
when data_plane.enabled=false. Async + TQ support has not been merged yet.
```

### What is wired (via #3267 and the SC stack)

- Generation side: `configure_vllm_for_router_replay(master_config.policy)` in SC `_build_generation` sets `enable_return_routed_experts` (`single_controller_utils/setup.py:157`).
- NeMo-Gym actor: `require_routed_experts=router_replay_enabled(...)` (`setup.py:364-369`).
- Train side field selection: `TQPolicy` reads `fields_with_optional_routed_experts(DP_TRAIN_FIELDS, enabled=self._router_replay_enabled)` (`nemo_rl/models/policy/tq_policy.py:121-183, 279-398`).

### What is missing

1. **No `routed_experts` preservation in the SC rollout → TQ → train path.** `routed_experts` capture lives only in the legacy sync path `nemo_rl/experience/rollouts.py` (`_attach_routed_experts_to_message_log_prefix`, etc.) and in grpo's driver path (`_preserve_router_replay_routed_experts`, grpo.py:1731-1757). The SC rollout/buffer path (`experience/rollout_manager.py`, `algorithms/async_utils/`, DataPlane payload/schema) never writes `routed_experts` into the TQ records, so `TQPolicy` reads them missing at train time.
2. **No SC-side guard.** No `validate_router_replay_config` / NotImplementedError is raised anywhere in `single_controller.py` / `single_controller_utils/` / `tq_policy.py`. Unlike grpo.py:3539, an SC run with `policy.router_replay.enabled=true` does not fail loud.
3. **No SC R3 validation.** The existing R3 async test (`tests/unit/algorithms/test_grpo_router_replay_async.py`) and recipe (`examples/configs/recipes/llm/grpo-qwen3-30ba3b-thinking-swe1-16n8g-megatron-cp2-r3-async-gym.yaml`) target the legacy non-TQ path; there is no SC + R3 recipe or functional test.

### Effect

An SC run (`data_plane.enabled=true`, `impl=transfer_queue`) with `policy.router_replay.enabled=true` produces generations with routed-expert traces that are then dropped before training, so the trainer computes MoE logprobs without the replayed routing — the exact correctness problem R3 exists to prevent — with no error surfaced.

### Ask

Either:
- (a) Wire `routed_experts` through the SC rollout → `TQReplayBuffer` → DataPlane → `TQPolicy` train path end-to-end (mirroring grpo's `_preserve_router_replay_routed_experts`) and add an SC + R3 recipe/functional test; **or**
- (b) Until (a) lands, add a fail-loud guard at SC setup (mirror grpo.py:3539) so `router_replay.enabled=true` on the SC/TQ path raises instead of silently mis-training.

Flagged during review of #3267. Tracked as an SC loose end in #2625.

zyzhou5:
# What does this PR do ?
 This PR adds R3 router replay support to the SingleController native async GRPO + TQ path. It propagates vLLM routed-expert indices through async rollout message logs into the TQ training payload, validates that routes are present when router replay is enabled, and safely handles partially failed rollout groups by using the existing -1 fallback semantics. It also adds diagnostic tracing, unit coverage, an R3 SingleController recipe with a launch command, and a nightly end-to-end correctness test.

It closes #3327 

## Validation
Logs can be found in (Test 5) https://api.wandb.ai/links/nvidia-nemo-fw-public/6j8mkubs

Screenshot 2026-07-28 at 4 03 46 PM

### DIFF
--- a/docs/guides/router-replay.md
+++ b/docs/guides/router-replay.md
@@ -49,13 +49,14 @@ Router Replay validation covers two end-to-end questions:
 2. whether matched R3-on runs reduce train-vs-rollout mismatch relative to
    matched R3-off controls.
 
-### Trace Debugging
+### Validation and Trace Debugging
 
 Router Replay can emit JSONL traces for a small number of training steps. This
 is intended for correctness debugging, not long training runs.
 
 | Environment variable | Default | Meaning |
 | --- | --- | --- |
+| `NRL_ROUTER_REPLAY_VALIDATE` | `0` | Validate replay tensors before Megatron installs them, rejecting partially missing routes, duplicate top-k expert IDs, and out-of-range expert IDs. |
 | `NRL_R3_TRACE` | `0` | Master switch for R3 JSONL trace emission. |
 | `NRL_R3_TRACE_STEPS` | `1` | Number of training steps to trace. |
 | `NRL_R3_TRACE_SAMPLES` | `2` | Number of samples with full tensor previews. |
@@ -116,7 +117,11 @@ all returned vLLM routes are still replayed exactly.
 The fallback is intentionally route-local: it does not disable Router Replay for
 the whole batch or sample.
 
-When fallback is used, NeMo RL logs
-`r3/routed_experts_fallback_token_route_fraction`. This metric should normally
-be zero or near-zero. A nonzero value means some token routes used Megatron's
-normal router instead of replay.
+When fallback is used, the vLLM worker emits a `R3 router replay fallback:` warning
+to the run log naming the affected sample count and missing token-route count.
+Fallback should normally be absent or rare; frequent warnings mean a meaningful
+share of token routes used Megatron's normal router instead of replay.
+
+The generation backend also computes
+`r3/routed_experts_fallback_token_route_fraction`, but no training loop currently
+forwards it to the metric logger, so do not rely on it in dashboards or gates.
--- a/docs/guides/router-replay.md
+++ b/docs/guides/router-replay.md
@@ -34,6 +34,12 @@ An example recipe is available at:
 examples/configs/recipes/llm/grpo-qwen3-30ba3b-8n8g-megatron-cp2-r3.yaml
 ```
 
+The native async TransferQueue path uses the SingleController entrypoint with:
+
+```text
+examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
+```
+
 ## Validation
 
 Router Replay validation covers two end-to-end questions:
--- a/examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
+++ b/examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
@@ -1,4 +1,7 @@
 # SingleController async+TQ variant of the Qwen3-30B-A3B R3 recipe.
+# Launch with:
+#   uv run examples/run_grpo_single_controller.py \
+#     --config examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
 defaults: ./grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async.yaml
 
 # SC does not support validation and checkpointing yet.
--- a/examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
+++ b/examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
@@ -1,0 +1,26 @@
+# SingleController async+TQ variant of the Qwen3-30B-A3B R3 recipe.
+defaults: ./grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async.yaml
+
+# SC does not support validation and checkpointing yet.
+grpo:
+  val_period: 0
+
+checkpointing:
+  checkpoint_dir: results/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller
+
+# TransferQueue is mandatory for the SingleController path.
+data_plane:
+  enabled: true
+
+async_rl:
+  sampler:
+    name: windowed
+    max_staleness_versions: 1
+  min_groups_for_streaming_train: ${grpo.num_prompts_per_step}
+  max_inflight_prompts: ${grpo.num_prompts_per_step}
+  max_buffered_rollouts: 64
+
+logger:
+  log_dir: logs/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller
+  wandb:
+    name: grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -86,6 +86,7 @@ from nemo_rl.experience.interfaces import (
 )
 from nemo_rl.experience.rollouts import (
     EffortLevelsConfig,
+    backfill_missing_routed_experts,
     get_nemo_gym_thinking_tags,
     run_async_multi_turn_rollout,
     run_multi_turn_rollout,
@@ -1740,12 +1741,16 @@ def add_grpo_token_loss_masks_and_generation_logprobs(
     generated assistant messages have generation_logprobs, so use that field as the
     trainable-token marker. This function mutates each message in-place by adding a
     token_loss_mask and, when missing, a zero-valued generation_logprobs tensor.
+    Router-replay routes get the same treatment via
+    :func:`backfill_missing_routed_experts`, so every per-token field is defined
+    for every tokenized message before the batch is flattened.
 
     Args:
         message_logs: Batch of tokenized message logs. Each message must contain a
             ``role`` and ``token_ids`` field. Messages that already contain
             ``generation_logprobs`` are treated as rollout-generated messages.
     """
+    backfill_missing_routed_experts(message_logs)
     for message_log in message_logs:
         for message in message_log:
             role = cast(str, message["role"])
@@ -2948,6 +2953,10 @@ def grpo_train(
                     # Save baseline for logging (before deletion)
                     baseline_for_log = baseline.clone()
 
+                    # Must precede prompt extraction: it reuses the same message
+                    # dicts, so this also protects the prompt flatten below.
+                    backfill_missing_routed_experts(repeated_batch["message_log"])
+
                     # Extract original prompt messages using the length field
                     # This correctly handles multi-turn prompts that contain assistant messages
                     initial_prompt_message_logs = extract_initial_prompt_messages(
@@ -3860,9 +3869,12 @@ def async_grpo_train(
         master_config.data_plane or {}
     ).get("enabled", False):
         raise NotImplementedError(
-            "policy.router_replay.enabled=true with async GRPO is currently "
-            "supported only when data_plane.enabled=false. Async + TQ support "
-            "has not been merged yet."
+            "policy.router_replay.enabled=true with async GRPO on this "
+            "entrypoint is supported only when data_plane.enabled=false. For "
+            "async + TransferQueue, use the SingleController entrypoint: "
+            "examples/run_grpo_single_controller.py with e.g. "
+            "examples/configs/recipes/llm/"
+            "grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml"
         )
 
     if master_config.grpo["async_grpo"]["max_trajectory_age_steps"] > 1:
@@ -4353,6 +4365,10 @@ def async_grpo_train(
 
                 print("▶ Processing rewards...")
                 with timer.time("reward_calculation"):
+                    # Must precede prompt extraction: it reuses the same message
+                    # dicts, so this also protects the prompt flatten below.
+                    backfill_missing_routed_experts(repeated_batch["message_log"])
+
                     # Extract original prompt messages using the length field
                     # This correctly handles multi-turn prompts that contain assistant messages
                     initial_prompt_message_logs = extract_initial_prompt_messages(
--- a/nemo_rl/models/generation/interfaces.py
+++ b/nemo_rl/models/generation/interfaces.py
@@ -28,6 +28,11 @@ from nemo_rl.distributed.batched_data_dict import BatchedDataDict
 # determined, fall back to int16.
 ROUTED_EXPERTS_FALLBACK_DTYPE = torch.int16
 
+# A routed-expert row whose every top-k slot is this value means "no route was
+# captured for this token"; the Megatron replay install falls back to the model's
+# own router for those rows. Partially-negative rows are rejected as corruption.
+ROUTED_EXPERTS_MISSING_ROUTE_SENTINEL = -1
+
 _ROUTED_EXPERTS_DTYPE_NAMES = {
     torch.int8: "int8",
     torch.int16: "int16",
--- a/tests/test_suites/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.sh
+++ b/tests/test_suites/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.sh
@@ -23,6 +23,13 @@ NUM_MINUTES=60
 exit_if_max_steps_reached
 
 cd $PROJECT_ROOT
+
+# check_r3_trace.py cross-checks every record in the trace dir, so a prior
+# attempt's producer records (written before it died) would have no matching
+# fetch and fail this attempt. Trace files are per-host/per-pid and appended,
+# so a retry never overwrites them.
+rm -rf "$NRL_R3_TRACE_DIR"
+
 uv run examples/run_grpo_single_controller.py \
     --config $CONFIG_PATH \
     grpo.max_num_steps=$MAX_STEPS \
--- a/tests/test_suites/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.sh
+++ b/tests/test_suites/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
+source $SCRIPT_DIR/common.env
+
+export NRL_IGNORE_TP_ACCURACY_CHECK=1
+export NRL_ROUTER_REPLAY_VALIDATE=1
+export NRL_R3_TRACE=1
+export NRL_R3_TRACE_VERIFY_FORWARD=1
+export NRL_R3_TRACE_STEPS=1
+export NRL_R3_TRACE_SAMPLES=1
+export NRL_R3_TRACE_MICROBATCHES=1
+export NRL_R3_TRACE_DIR="$LOG_DIR/r3_trace"
+
+# ===== BEGIN CONFIG =====
+NUM_NODES=10
+GPUS_PER_NODE=8
+STEPS_PER_RUN=10
+MAX_STEPS=10
+NUM_RUNS=$(( (MAX_STEPS + STEPS_PER_RUN - 1) / STEPS_PER_RUN ))  # Round up
+NUM_MINUTES=60
+# ===== END CONFIG =====
+
+exit_if_max_steps_reached
+
+cd $PROJECT_ROOT
+uv run examples/run_grpo_single_controller.py \
+    --config $CONFIG_PATH \
+    grpo.max_num_steps=$MAX_STEPS \
+    logger.log_dir=$LOG_DIR \
+    logger.wandb_enabled=True \
+    logger.wandb.project=nemo-rl \
+    logger.wandb.name=$EXP_NAME \
+    logger.monitor_gpus=True \
+    logger.tensorboard_enabled=True \
+    checkpointing.enabled=False \
+    checkpointing.checkpoint_dir=$CKPT_DIR \
+    $@ \
+    2>&1 | tee $RUN_LOG
+
+uv run tests/json_dump_tb_logs.py $LOG_DIR --output_path $JSON_METRICS
+
+if [[ $(jq 'to_entries | .[] | select(.key == "train/loss") | .value | keys | map(tonumber) | max' $JSON_METRICS) -ge $MAX_STEPS ]]; then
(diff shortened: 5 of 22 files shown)


zyzhou5:
/ok to test [695b5fc](https://github.com/NVIDIA-NeMo/RL/commit/695b5fc7dcb5349190b980c085de9abdb05957d2)

docs/guides/router-replay.md
```diff
@@ -110,7 +116,11 @@ all returned vLLM routes are still replayed exactly.
...
-`r3/routed_experts_fallback_token_route_fraction`. This metric should normally
-be zero or near-zero. A nonzero value means some token routes used Megatron's
-normal router instead of replay.
+When fallback is used, the vLLM worker emits a `R3 router replay fallback:` warning
+to the run log naming the affected sample count and missing token-route count.
+Fallback should normally be absent or rare; frequent warnings mean a meaningful
+share of token routes used Megatron's normal router instead of replay.
+
+The generation backend also computes
+`r3/routed_experts_fallback_token_route_fraction`, but no training loop currently
```
ZhiyuLi-Nvidia, line 126: > no training loop currently forwards it to the metric logger

Is that as expected?

`_add_r3_fallback_metrics` writes into the inner `gen_metrics`. We should be able to trace it using `rollout_metrics`.

tests/test_suites/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
+source $SCRIPT_DIR/common.env
+
+export NRL_IGNORE_TP_ACCURACY_CHECK=1
+export NRL_ROUTER_REPLAY_VALIDATE=1
```
ZhiyuLi-Nvidia, line 6: This env variable seems not mentioned in the docs/guides/router-replay.md. Can we add it?

ZhiyuLi-Nvidia:
Awesome @zyzhou5 ! Thank you for the PR and all the experiments.

LGTM.

ZhiyuLi-Nvidia (dismissed):

docs/guides/router-replay.md
zyzhou5: SC does not seem to propagate `rollout_metrics` through TQ.

zyzhou5:
/ok to test [893c9fa](https://github.com/NVIDIA-NeMo/RL/commit/893c9faae26ee6db65a3626f07fa7478b1009233)

docs/guides/router-replay.md
terrykong: @ZhiyuLi-Nvidia @zyzhou5 — I dug into this one and filed it so it doesn't get lost: https://github.com/NVIDIA-NeMo/RL/issues/2625#issuecomment-5127202378

Short version: @zyzhou5's answer is right, and it turns out there are two stacked gaps rather than one.

1. **SC drops `rollout_metrics` wholesale**, not just the R3 keys — `PromptGroupRecord.rollout_metrics` is populated but never read by `record_to_train_batch`, so it dies at the TQ boundary. Both legacy drivers do log it (`grpo.py:2787` sync, `grpo.py:4659` async), so this is an SC-only gap and matches the existing TODO at `single_controller.py:552-556`.
2. **The `r3/*` counters were never in `rollout_metrics` on any path** — `_add_r3_fallback_metrics` writes them into a local `gen_metrics` dict that both legacy call sites drop on the floor. So @ZhiyuLi-Nvidia, tracing via `rollout_metrics` wouldn't have worked even on the legacy path today; and on SC `_add_r3_fallback_metrics` isn't called at all.

Which means fixing (1) alone still wouldn't surface the fallback fraction — it needs both. Nothing to change in this PR; the docs caveat @zyzhou5 added is accurate as written. Filed against #2625 rather than here since (1) is much broader than R3.

nemo_rl/algorithms/grpo.py
```diff
@@ -1730,12 +1731,16 @@ def add_grpo_token_loss_masks_and_generation_logprobs(
...
+    Router-replay routes get the same treatment via
+    :func:`backfill_missing_routed_experts`, so every per-token field is defined
+    for every tokenized message before the batch is flattened.
 
     Args:
         message_logs: Batch of tokenized message logs. Each message must contain a
             ``role`` and ``token_ids`` field. Messages that already contain
             ``generation_logprobs`` are treated as rollout-generated messages.
     """
+    backfill_missing_routed_experts(message_logs)
```
terrykong, line 1753: [`nemo_rl/algorithms/grpo.py:1743`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/algorithms/grpo.py#L1743)

Adding the backfill here fixes the ragged-`routed_experts` case for the *main* flatten on the legacy sync and async drivers too — nice side effect. But both legacy drivers extract **and flatten the prompt** before this function runs, so the prompt flatten is still unprotected. Pre-existing, not introduced here; flagging because this PR is what establishes the fix, and it landed in `payload.py`/`sync_rollout_actor.py` but not in the two `grpo.py` drivers.

Failing case — the legacy async R3 recipe that is **already in nightly** (`grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async.yaml`, i.e. `router_replay.enabled=true`, `async_grpo.enabled=true`, `data_plane.enabled=false`):

1. Sample 0's turn-1 generation succeeds, so its *prompt* messages get routes via [`_attach_routed_experts_to_message_log_prefix`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L490-L493).
2. Sample 1's turn-1 generation raises; the handler [prints and `break`s](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L1076-L1078) and the sample stays in the batch with a prompt-only log and **no** `routed_experts`.
3. [`async_grpo_train` flattens the prompt at grpo.py:4303](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/algorithms/grpo.py#L4298-L4305) — **before** `add_grpo_token_loss_masks_and_generation_logprobs`, so before any backfill.
4. Inside, the missing value becomes [`torch.zeros(0, ...)` — 1-D](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/data/llm_message_utils.py#L386-L393), then [`_pad_tensor` pads it to `[P]`, still 1-D](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/data/llm_message_utils.py#L199-L211).
5. [`torch.stack(padded)`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/data/llm_message_utils.py#L397) → `RuntimeError: stack expects each tensor to be equal size, but got [P, L, K] at entry 0 and [P] at entry 1`.

(If the *failed* sample is index 0 the `isinstance(values[0], Tensor)` check at [L369-371](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/data/llm_message_utils.py#L369-L371) short-circuits and the key silently degrades to a Python list instead — quieter, still wrong. So the crash needs a routed sample first.)

Your own new [`test_record_to_train_batch_backfills_routes_for_failed_completion`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/tests/unit/experience/test_payload.py#L688-L709) is exactly this scenario for the SC path.

**Action:** hoist the backfill above the prompt extraction in `grpo_train` (~L2885) and `async_grpo_train` (~L4298), mirroring the comment you already wrote in `payload.py`. Not a diff line, so as a snippet — `backfill_missing_routed_experts` is already imported at [grpo.py:95](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/algorithms/grpo.py#L95) and no-ops when R3 is off:

```python
# Must precede the prompt extraction: it reuses the same message dicts, so
# backfilling here also covers the prompt flatten below.
backfill_missing_routed_experts(repeated_batch["message_log"])
initial_prompt_message_logs = extract_initial_prompt_messages(
    repeated_batch["message_log"],
    repeated_batch["length"],
)
```

nemo_rl/models/generation/interfaces.py
```diff
@@ -28,6 +28,11 @@
 # determined, fall back to int16.
 ROUTED_EXPERTS_FALLBACK_DTYPE = torch.int16
 
+# A routed-expert row whose every top-k slot is this value means "no route was
+# captured for this token"; the Megatron replay install falls back to the model's
+# own router for those rows. Partially-negative rows are rejected as corruption.
+ROUTED_EXPERTS_MISSING_ROUTE_SENTINEL = -1
```
terrykong, line 34: [`nemo_rl/models/generation/interfaces.py:34`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/models/generation/interfaces.py#L34)

Good change — promoting this to a named constant and repointing `router_replay.py` at it removes one silent-drift point. PR-introduced nit: the consolidation stopped one file short of the **producer**.

[`nemo_rl/models/generation/vllm/utils.py:26`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/models/generation/vllm/utils.py#L26) still declares its own copy:

```python
R3_MISSING_ROUTE_SENTINEL = -1
```

and it is what actually writes the sentinel, at [utils.py:232](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/models/generation/vllm/utils.py#L232) (`full[routes_to_copy:expected_routes] = R3_MISSING_ROUTE_SENTINEL`). So the producer and the two consumers agree on `-1` by coincidence, which is the thing the shared constant exists to prevent. Values are identical today, so no behavioural bug.

**Action:** in `nemo_rl/models/generation/vllm/utils.py`, drop the local literal and import the shared one (`from nemo_rl.models.generation.interfaces import ROUTED_EXPERTS_MISSING_ROUTE_SENTINEL`), keeping `R3_MISSING_ROUTE_SENTINEL` as an alias if the shorter local name is preferred at the call site.

Separately, and while you're here: [`payload.py:93-94`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/payload.py#L93-L94) uses the bare string `"routed_experts"` while `replay_buffer.py` imports `ROUTED_EXPERTS_FIELD` for the same field. Same nit, one line.

examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
```diff
@@ -0,0 +1,29 @@
+# SingleController async+TQ variant of the Qwen3-30B-A3B R3 recipe.
+# Launch with:
+#   uv run examples/run_grpo_single_controller.py \
+#     --config examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
+defaults: ./grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async.yaml
```
terrykong, line 5: [`examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml:5`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml#L5)

Now that this recipe exists, the error message on the legacy entrypoint is factually stale. [`grpo.py:3803-3810`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/algorithms/grpo.py#L3803-L3810) still says:

> "policy.router_replay.enabled=true with async GRPO is currently supported only when data_plane.enabled=false. **Async + TQ support has not been merged yet.**"

To be clear, **the guard itself is correct and should stay** — I traced the entrypoints and SC goes `examples/run_grpo_single_controller.py` → `setup_single_controller` → `single_controller.py`, never reaching `async_grpo_train`. It is only the last sentence that is now wrong, and it is the exact string a user will search after hitting it (issue #3327 quotes it verbatim as the definition of the gap). They would conclude the feature doesn't exist while this PR ships a recipe for it.

**Action:** reword the message in `nemo_rl/algorithms/grpo.py::async_grpo_train` to point at this recipe. Not a diff line, so as a snippet:

```python
        raise NotImplementedError(
            "policy.router_replay.enabled=true with async GRPO on this "
            "entrypoint is supported only when data_plane.enabled=false. For "
            "async + TransferQueue, use the SingleController entrypoint: "
            "examples/run_grpo_single_controller.py with e.g. "
            "examples/configs/recipes/llm/"
            "grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml"
        )
```

docs/guides/router-replay.md
```diff
@@ -110,7 +117,11 @@ all returned vLLM routes are still replayed exactly.
...
+forwards it to the metric logger, so do not rely on it in dashboards or gates.
```
terrykong, line 127: [`docs/guides/router-replay.md:125-127`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/docs/guides/router-replay.md#L125-L127)

@ZhiyuLi-Nvidia — following up on your question here: **@zyzhou5's answer is correct, and if anything the doc line understates it.** I traced it end to end and there's nothing to fix, but three details seem worth recording since this is the one paragraph telling people not to trust the metric:

1. **It's dropped on *every* path, not just SC.** [`_add_r3_fallback_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L64-L87) is called at exactly two sites — [rollouts.py:370](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L370) (legacy sync) and [rollouts.py:511](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L511) (legacy async). Both callers pop only `_response_truncated` ([L768](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L768)) and `gen_leader_worker_idx` ([L1059-1072](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L1059-L1072)) and let the dict fall out of scope; neither the sync [`rollout_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L890) nor the async [`sample_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollouts.py#L1171) ever merges `gen_metrics`. Nothing repo-wide consumes the `r3/*` keys.
2. **On SC it isn't even computed.** `RolloutManager` never goes through `generate_responses*`; [`_generate_response` builds `gen_metrics` from scratch](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/rollout_manager.py#L329-L341) and never calls `_add_r3_fallback_metrics`. So closing this on SC is two changes, not one.
3. **Your `rollout_metrics` hypothesis is right about the second one.** [`TQReplayBuffer.commit` ships only `record_to_train_batch(record, ...)`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/algorithms/async_utils/replay_buffer.py#L724), and that function never reads `record.rollout_metrics` ([payload.py:84-95](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/experience/payload.py#L84-L95)) — matching the standing TODO at [single_controller.py:552-556](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/algorithms/single_controller.py#L552-L556), which already lists `rollout_metrics`.

Filed against the SC cleanup tracking issue so it isn't lost: https://github.com/NVIDIA-NeMo/RL/issues/2625#issuecomment-5127202378 (it turned out to be broader than R3 — SC drops `rollout_metrics` wholesale, not just the `r3/*` keys). Nothing to change in this PR.

**Action (optional, docs-only):** the sentence is accurate; this only makes point 2 explicit for an SC reader.

```suggestion
The legacy (non-Single-Controller) rollout paths also compute
`r3/routed_experts_fallback_token_route_fraction`, but no training loop currently
forwards it to the metric logger, and the Single Controller path does not compute
it at all — so do not rely on it in dashboards or gates.
```

terrykong (commented):
Static review of the SC + R3 wiring, run as a 6-agent pass (RL expert, bug finder, test reviewer, design reviewer, comment reviewer, plus adjudication). **Overall LGTM** — one real bug on an adjacent path, the rest are low-severity nits and process items. Nothing here should block merge.

The change is tightly scoped and mirrors the existing patterns well. Worth calling out explicitly:

- [`tq_policy.py:521`](https://github.com/NVIDIA-NeMo/RL/blob/893c9faae26ee6db65a3626f07fa7478b1009233/nemo_rl/models/policy/tq_policy.py#L521) was the **only** one of five field-selection sites still hardcoding `list(DP_TRAIN_FIELDS)` — the other four already used `fields_with_optional_routed_experts`. That one-liner is the actual fix, and it's the right one.
- `backfill_missing_routed_experts`'s "no-op when the batch carries no routes at all" is subtle and correct: it's what preserves the producer-side guard's ability to distinguish "replay off" from "capture failed". A naive always-fill would have silently defeated the guard this PR adds. The docstring says why, which is the right call.
- The ordering comment at `payload.py:62-64` is load-bearing and accurate — we verified `extract_initial_prompt_messages` really does reuse the same dicts.
- Guard placement at `TQReplayBuffer.commit` is the true choke point (both the native and Gym SC rollouts converge there before the TQ write) and it fires before any side effect. `generate_and_push` removes the reserved group on `BaseException` and re-raises, so it can't strand a slot.
- The new tests are Ray/GPU-free, land in the right L0 lane, and the partially-failed-group test encodes exactly the invariant the backfill exists for.

Checked and found clean, so nobody re-derives it: the recipe's `defaults:` chain resolves `checkpointing.enabled` to **false** (`grpo_math_1B_megatron.yaml:31` overrides the `true` in `grpo_math_1B.yaml`), so the documented launch command works; `routed_experts` is in `TOKEN_ALIGNED_FIELDS` so zero-pad rows never reach `_validate_replay_tensor`; `-1` is representable in all three signed carry dtypes; the R3 recipe sets `enable_prefix_caching: false`; SC + NeMo Gym fails loud rather than silently (`nemo_gym.py:410` + the new commit guard); `check_r3_trace.py` accepts both new flags and hard-fails if `trace_rollout_payload` never fires; 10x8x60min = 80 GPU-h matches the 3480→3560 bump exactly and nothing references the renamed test function; all seven R3 env vars are documented.

Two things we deliberately did **not** raise, in case they come up: an early read that the recipe was missing `checkpointing.enabled: false` was wrong (disproved by the `defaults:` chain above), and a suggestion to tighten the nightly gate from `< 1.02` to `< 1.015` was dropped — it rested on eyeballing values off the screenshot and both sibling R3 nightlies use 1.02.

The convergence evidence holds up: the screenshot is a matched 4-arm study (SC-r3-on / SC-r3-off / legacy-r3-on / legacy-r3-off, 100 steps) where SC-r3-on tracks legacy-r3-on on both `token_mult_prob_error` and `js_divergence_error` while both r3-off arms separate cleanly — exactly the claim, and exactly what the guide's own "Effectiveness Check" asks for.

**Not run:** no tests were executed and no linter ran — `uv` refuses this host (the lockfile is `sys_platform == 'linux'` only), so this is entirely static review.

_Generated by Claude Code_

zyzhou5:
/ok to test [4c4e96b](https://github.com/NVIDIA-NeMo/RL/pull/3378/commits/4c4e96bd4689af87dff9e1dcab46067feb094bdf)

nemo_rl/algorithms/grpo.py
terrykong: fixed in 4c4e96bd4689af87dff9e1dcab46067feb094bdf

nemo_rl/models/generation/interfaces.py
terrykong: fixed in 4c4e96bd4689af87dff9e1dcab46067feb094bdf

examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
terrykong: fixed in 4c4e96bd4689af87dff9e1dcab46067feb094bdf

zyzhou5:
/ok to test [3fd854d](https://github.com/NVIDIA-NeMo/RL/commit/3fd854d1ca40a074c7be8a9498769bf8cfbf9aa3)

zyzhou5:
/ok to test [5b191ab](https://github.com/NVIDIA-NeMo/RL/commit/5b191abdae6006cf470c51dffbf6aa1f9eb7166b)

zyzhou5:
/ok to test [d5be3d9](https://github.com/NVIDIA-NeMo/RL/commit/d5be3d96694fcdb1513117eaa26cc376354b8815)

terrykong (approved):

PR #3379, opened 2026-07-28
feat: warn when UV_CACHE_DIR is set

terrykong:
# What does this PR do ?

**Warns on `import nemo_rl` when `UV_CACHE_DIR` is set, since overriding the default uv cache is almost always unintentional and harmful.**

Two failure modes have been observed in the wild:

1. **Unnecessary reinstalls / very long install times.** Setting `UV_CACHE_DIR` points uv away from the cache that holds NeMo RL's prebuilt wheels (e.g. the one baked into the container), so venv creation re-downloads and re-builds packages. If you see long install times despite using a prebuilt container, this env var is sometimes the culprit.
2. **Cache corruption.** When `UV_CACHE_DIR` points somewhere persistent outside the container, the cache has been observed to get corrupted when multiple jobs access it simultaneously.

The default is the right choice both in a container and on baremetal, so the warning's only recommendation is to unset the variable.

# Issues

N/A

# Usage

```console
$ UV_CACHE_DIR=/lustre/shared/uv python -c "import nemo_rl"
WARNING:root:__init__.py:79: UV_CACHE_DIR is set to '/lustre/shared/uv'. This bypasses the
default uv cache holding NeMo RL's prebuilt wheels (causing slow reinstalls even with a
prebuilt container), and a cache on a persistent/shared filesystem can be corrupted by
concurrent jobs. Recommended: `unset UV_CACHE_DIR` and let uv use its default, whether in a
container or on baremetal.
```

**Pre checks**:

# Additional Information

* Warning only — nothing is unset or mutated on the user's behalf, so existing setups keep working.
* New unit test `tests/unit/test_uv_cache_dir_warning.py` (picked up by the `L0_Unit_Tests_Other` shard) covers unset / empty / set cases.

### DIFF
--- a/nemo_rl/__init__.py
+++ b/nemo_rl/__init__.py
@@ -76,8 +76,7 @@ def _warn_if_uv_cache_dir_set():
         "holding NeMo RL's prebuilt wheels (causing slow reinstalls even with a prebuilt "
         "container), and a cache on a persistent/shared filesystem can be corrupted by "
         "concurrent jobs. Recommended: `unset UV_CACHE_DIR` and let uv use its default, "
-        "whether in a container or on baremetal. With ray.sub, use UV_CACHE_DIR_OVERRIDE "
-        "instead (see docs/cluster.md)."
+        "whether in a container or on baremetal."
     )
 
 
--- a/nemo_rl/__init__.py
+++ b/nemo_rl/__init__.py
@@ -59,6 +59,31 @@ os.environ["RAY_USAGE_STATS_ENABLED"] = "0"
 os.environ["RAY_ENABLE_UV_RUN_RUNTIME_ENV"] = "0"
 
 
+def _warn_if_uv_cache_dir_set():
+    """Warn if UV_CACHE_DIR is set, since the default is almost always the right choice.
+
+    Pointing uv at a non-default cache bypasses the cache that holds NeMo RL's
+    prebuilt wheels (e.g. the one baked into the container), and a cache placed on
+    a persistent/shared filesystem has been observed to corrupt when several jobs
+    use it concurrently.
+    """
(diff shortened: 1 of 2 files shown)


nemo_rl/__init__.py
```diff
@@ -59,6 +59,30 @@
...
+    Pointing uv at a non-default cache bypasses the cache that holds NeMo RL's
+    prebuilt wheels (e.g. the one baked into the container), and a cache placed on
+    a persistent/shared filesystem has been observed to corrupt when several jobs
+    use it concurrently.
+    """
+    uv_cache_dir = os.environ.get("UV_CACHE_DIR")
+    if not uv_cache_dir:
+        return
+
+    logging.warning(
```
yuki-97, line 74: warns on this repo's own recommended `UV_CACHE_DIR` uses — users following documented guidance will see a warning telling them to unset a variable they were told to set:

- `skills/nemo-rl-brev-etiquette/SKILL.md:61` — `export UV_CACHE_DIR=$CACHE_ROOT/uv` for Brev.
- `infra/nrl_k8s/examples/ultra_{48,64}n_pipeclean.gb300.infra.yaml:169` — set `UV_CACHE_DIR=${PERSISTENT_CACHE}/uv`.

two options: unset in those three sites, or add an escape-hatch env var (e.g. `NRL_ACK_UV_CACHE_DIR=1`) and set it there.

yuki-97: besides the above, `docker/Dockerfile.ngc_pytorch` will always have the warning? but seems it's `However, it is still a work in progress and is not yet ready for production use.`.
https://github.com/NVIDIA-NeMo/RL/blob/42e490efbcfc67fd83251efe0089cf01a7c93a9a/docker/Dockerfile.ngc_pytorch#L89

yuki-97 (approved):

PR #3364, opened 2026-07-27
fix: Qwen3.5-397B training stability and async-GRPO robustness

michal2409:
## What does this PR do?

Merges the Qwen3.5-397B reference training changes into `opt/dev`, rebased onto current `main` (branch is 0 commits behind `opt/dev`).

Included commits:
- **fix(megatron):** GB300 Qwen3.5-397B training stability knobs
- **fix(vllm):** Qwen3.5 MoE refit and async-server fixes
- **feat(grpo):** async-GRPO robustness for long-horizon SWE rollouts
- **feat(grpo):** make env-flagged sample masking optional

## Convergence data

**Setup**
64×GB300 (4 GPUs/node): 16 training nodes (Megatron TP4·PP2·EP32, sequence
packing @ 65 536 tokens, bf16) + 48 generation nodes (24 vLLM engines, TP8/EP8, async
engine). GBS 256 = 16 prompts × 16 generations/step, async GRPO lag-1, NeMo-Gym
OpenHands agents on R2E-Gym easy (curriculum-v2 train set), validation = 256 instances
× 4 generations @ temp 0.1 / top-p 0.95, seed 3978.

**Training reward (fraction of 256 rollouts solved, per step):**

| step | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
|---|---|---|---|---|---|---|---|---|---|---|
| reward | .406 | .441 | .406 | .434 | .520 | .387 | .410 | .457 | .629 | .598 |

| step | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | → eval@20 |
|---|---|---|---|---|---|---|---|---|---|---|
| reward | .602 | .461 | .617 | .586 | .637 | .613 | .648 | .574 | .605 | **0.789** |

Screenshot 2026-07-28 at 15 57 50

Screenshot 2026-07-28 at 15 58 01

Screenshot 2026-07-28 at 15 57 40

Screenshot 2026-07-28 at 15 57 33

Screenshot 2026-07-28 at 15 56 55

Screenshot 2026-07-28 at 16 01 32

PR #3382, opened 2026-07-28
fix: Qwen3.5-397B training stability and async-GRPO robustness

michal2409:
## What does this PR do?

Core-algorithm remainder of the Qwen3.5-397B reference training bring-up, targeting `main`. Per review feedback the independent pieces were split into their own PRs and removed from this branch; what stays here are the changes that touch the core async-GRPO path. Same lineage as #3364 (which targets `opt/dev` per the agreed workflow).

Included commits:
- **feat(megatron): make DDP `check_for_nan_in_grad` configurable** — plumbed through `distributed_data_parallel_config` (absent keeps the always-on default).
- **fix(vllm): async-server robustness for multi-turn agentic rollouts** — normalize non-list/str message content before template rendering, and rebuild the isolated prefix render from vLLM's normalized message representation (Qwen3.5 re-renders assistant history with normalized reasoning content, so the raw deepcopy diverged token-wise). Includes `replace_prefix_tokens` unit tests.
- **feat(grpo): async-GRPO robustness for long-horizon SWE rollouts** — start trajectory collection only after the initial refit; drain in-flight rollouts before in-loop validation; close pause races at validation boundaries; optional `env.nemo_gym.backfill_failed_rollouts`; refit-safe prefix caching (invalidate vLLM prefix cache on every refit path); positional advantage grouping for NeMo-Gym rollouts (`_stable_group_ids`); replay buffer consumes trajectories staged for future steps instead of stalling; opt-in `NRL_LP_MASK_DEBUG` diagnostics.

Split out per review (separate PRs, removed from this branch):
- #3400 — `grpo.val_start_at` (delayed periodic validation)
- #3401 — validation-only sampling params + grouped pass@k validation accuracy
- #3402 — configurable env-flagged sample masking (`grpo.mask_env_flagged_samples`)
- #3404 — `grpo.stop_at_validation_accuracy` early stop

Also removed from this branch:
- MLPerf telemetry hooks (`TrainTelemetryHooks` / `mlperf_logger`) — MLPerf compliance logging now lives entirely in the benchmark launch stack as a `Logger` wrapper, so nemo-rl needs no hooks.
- The Qwen3.5 MoE refit expert-split (#3403) — isolation experiments showed vLLM 0.20 loads the stacked fused-expert export natively (byte-identical to cold load) once `moe_backend=triton` is set; the historical refit failure was the FlashInfer TRT-LLM post-load repack, which no export format survives. See #3403 for details.

Diff is now 11 files, +843/−116 (was 17 files, +1542/−204).

## Convergence data

The commits here are a subset of the branch validated end-to-end on the benchmark (results below from that full stack; the removed pieces are config-gated features that were active in the run via their split-PR equivalents).

**Setup**
64×GB300 (4 GPUs/node): 16 training nodes (Megatron TP4·PP2·EP32, sequence
packing @ 65 536 tokens, bf16) + 48 generation nodes (24 vLLM engines, TP8/EP8, async
engine). GBS 256 = 16 prompts × 16 generations/step, async GRPO lag-1, NeMo-Gym
OpenHands agents on R2E-Gym easy (curriculum-v2 train set), validation = 256 instances
× 4 generations @ temp 0.1 / top-p 0.95, seed 3978.

**Training reward (fraction of 256 rollouts solved, per step):**

| step | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
|---|---|---|---|---|---|---|---|---|---|---|
| reward | .406 | .441 | .406 | .434 | .520 | .387 | .410 | .457 | .629 | .598 |

| step | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | → eval@20 |
|---|---|---|---|---|---|---|---|---|---|---|
| reward | .602 | .461 | .617 | .586 | .637 | .613 | .648 | .574 | .605 | **0.789** |

### DIFF
--- a/nemo_rl/algorithms/async_utils/replay_buffer.py
+++ b/nemo_rl/algorithms/async_utils/replay_buffer.py
@@ -186,10 +186,11 @@ class ReplayBufferImpl(ReplayBufferProtocol):
     ) -> Optional[dict[str, Any]]:
         """Sample per-prompt trajectory groups intended for the current training step.
 
-        Only returns trajectories with target_weight_version == current_weight_version.
-        If insufficient trajectories are available, returns None to stall training
-        until the remaining trajectories are generated. This ensures no trajectory
-        loses its last chance to be used for its intended training step.
+        Only returns trajectories with target_weight_version >= current_weight_version,
+        so trajectories staged for a future step may be consumed early instead of
+        stalling. If insufficient trajectories are available, returns None to stall
+        training until the remaining trajectories are generated. This ensures no
+        trajectory loses its last chance to be used for its intended training step.
 
         Returns:
             Dictionary with 'trajectories' and 'avg_trajectory_age' keys, or None if insufficient data
@@ -252,7 +253,7 @@ class ReplayBufferImpl(ReplayBufferProtocol):
             intended_indices = [
                 i
                 for i in valid_indices
-                if self.target_weight_versions[i] == current_weight_version
+                if self.target_weight_versions[i] >= current_weight_version
             ]
 
             print(
@@ -283,8 +284,10 @@ class ReplayBufferImpl(ReplayBufferProtocol):
                 f"✅ Selected counts by generation weight-version: {Counter(sampled_weights)}"
             )
             print(f"📊 Average trajectory age: {avg_trajectory_age:.2f} steps")
+            sampled_targets = Counter(self.target_weight_versions[i] for i in selected)
             print(
-                f"🎯 All selected trajectories target step {current_weight_version} (100% target match)"
+                f"🎯 Selected trajectory targets for step {current_weight_version}: "
+                f"{dict(sampled_targets)}"
             )
 
             # Remove selected items in reverse order to maintain correct indices
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1688,6 +1688,25 @@ def scale_rewards(
     return repeated_batch
 
 
+def _stable_group_ids(prompt_ids_for_adv, num_generations_per_prompt):
+    """Stable per-prompt grouping key for GRPO advantage computation.
+
+    GRPO groups samples by prompt (torch.unique) to compute the leave-one-out baseline. The default
+    key is the rendered prompt token-ids, but for agentic gym rollouts each generation's first-turn
+    prompt tokenizes slightly differently (observed on Qwen3-Instruct + hermes: every generation
+    becomes its own singleton group -> leave-one-out baseline == reward -> advantage == 0 -> zero
+    gradient; Exp 26). The training batch is laid out as contiguous num_gen blocks per prompt
+    (async: BatchedDataDict.from_batches of per-prompt groups; sync: repeat_interleave), so the
+    correct, model-agnostic group id is positional: index // num_gen. Falls back to the original
+    token-id grouping if the batch is not an exact multiple of num_gen (e.g. dynamic sampling).
+    """
+    n = int(prompt_ids_for_adv.shape[0])
+    g = int(num_generations_per_prompt)
+    if g <= 0 or n % g != 0:
+        return prompt_ids_for_adv
+    return (torch.arange(n, device=prompt_ids_for_adv.device) // g).unsqueeze(1)
+
+
 def extract_initial_prompt_messages(
     message_logs: list,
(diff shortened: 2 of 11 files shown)


nemo_rl/algorithms/grpo.py
```diff
@@ -3064,8 +3223,17 @@ def grpo_train(
                     sample_mask = train_data["sample_mask"]
                     mask = token_mask * sample_mask.unsqueeze(-1)
 
+                    # Positional grouping is only needed for agentic gym rollouts,
+                    # where per-generation prompt tokenization is non-deterministic;
+                    # keep main's token-id grouping for every other GRPO user.
+                    advantage_group_ids = prompt_ids_for_adv
+                    if _should_use_nemo_gym(master_config):
+                        advantage_group_ids = _stable_group_ids(
```
terrykong: [`grpo.py:3231`](https://github.com/NVIDIA-NeMo/RL/blob/f82b9bac75546f83061e7ec79cbb7c9d271a5feb/nemo_rl/algorithms/grpo.py#L3231)

`_stable_group_ids` is applied here and at [:4773](https://github.com/NVIDIA-NeMo/RL/blob/f82b9bac75546f83061e7ec79cbb7c9d271a5feb/nemo_rl/algorithms/grpo.py#L4773), but the **dynamic-sampling std is computed from the raw token-id grouping** at [`grpo.py:2975`](https://github.com/NVIDIA-NeMo/RL/blob/f82b9bac75546f83061e7ec79cbb7c9d271a5feb/nemo_rl/algorithms/grpo.py#L2975) and [`:2991`](https://github.com/NVIDIA-NeMo/RL/blob/f82b9bac75546f83061e7ec79cbb7c9d271a5feb/nemo_rl/algorithms/grpo.py#L2991) (`calculate_baseline_and_std_per_prompt(input_ids, ...)`) — the exact key this function's own docstring describes as producing singleton groups for agentic gym rollouts.

If that holds, then with NeMo-Gym + `use_dynamic_sampling: true`, singleton groups give `std = 0` for every sample, `non_zero_std_mask` filters everything, and the loop exhausts `dynamic_sampling_max_gen_batches` and raises `ValueError: Dynamic sampling has reached the maximum allowed number of batches`. The positional fix at this line runs after that point.

Flagging this as **lower confidence than the rest of this review** — it was the one finding our verification pass didn't independently adjudicate, so please treat it as a question rather than a conclusion. If dynamic sampling isn't used with the gym path in practice, ignore it.

Separately and more mundanely: `_stable_group_ids` has no type annotations, unlike its sibling `_calculate_observed_pass_metrics` added in the same PR.

nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -186,10 +186,11 @@ def sample(
...
         """Sample per-prompt trajectory groups intended for the current training step.
 
-        Only returns trajectories with target_weight_version == current_weight_version.
-        If insufficient trajectories are available, returns None to stall training
-        until the remaining trajectories are generated. This ensures no trajectory
-        loses its last chance to be used for its intended training step.
+        Only returns trajectories with target_weight_version >= current_weight_version,
+        so trajectories staged for a future step may be consumed early instead of
+        stalling. If insufficient trajectories are available, returns None to stall
+        training until the remaining trajectories are generated. This ensures no
```
terrykong: [`replay_buffer.py:192`](https://github.com/NVIDIA-NeMo/RL/blob/f82b9bac75546f83061e7ec79cbb7c9d271a5feb/nemo_rl/algorithms/async_utils/replay_buffer.py#L192)

The retained sentence *"This ensures no trajectory loses its last chance to be used for its intended training step"* no longer holds after the `==` → `>=` change at [:256](https://github.com/NVIDIA-NeMo/RL/blob/f82b9bac75546f83061e7ec79cbb7c9d271a5feb/nemo_rl/algorithms/async_utils/replay_buffer.py#L256): `intended_indices[:num_prompt_groups]` is FIFO over a now-*mixed* target set, so a trajectory targeting step `v` can be displaced by one targeting `v+1`, then permanently excluded at step `v+1` and later evicted.

Low likelihood in practice (insertion order roughly tracks target order), and I confirmed the change does **not** cause the deadlock you might expect — consuming future-targeted trajectories reduces occupancy, so the collector pauses *less*, not more. Just worth updating the docstring so it stops asserting an invariant the code no longer maintains.

PR #3386, opened 2026-07-28
feat(slurm): support external vLLM services in one allocation

yfw:
# What does this PR do ?

## Summary

Adds an opt-in framework for running external Gym vLLM service pools alongside NeMo RL in one Slurm heterogeneous job.

This provides gang scheduling while keeping service nodes isolated from the NeMo RL Ray cluster:

- Hetgroup 0 runs NeMo RL and the pool load balancers.
- Hetgroup 1 runs independent Ray/vLLM replicas.
- Training starts only after all required backends and load balancers are healthy.

## Motivation

External judges previously required separate Slurm jobs or had to share the training Ray cluster. Separate jobs are not gang scheduled and may become available at different times, while shared clusters risk resource contention and accidental node reuse.

This change allocates training and serving resources together while preserving separate Ray clusters.

## Implementation

- Adds a generic registration interface for arbitrary named vLLM pools.
- Validates node counts, TP topology, ports, paths, URL placeholders, containers, and required Python packages before startup.
- Starts each replica in an isolated private Ray cluster.
- Supports per-pool environment variables and vLLM arguments.
- Injects resolved load-balancer URLs into the training command.
- Provides health-aware, affinity-aware, least-loaded routing and per-request retry across live backends.
- Monitors required processes and terminates the heterogeneous job if a required service fails.
- Forwards shutdown signals and drains in-flight load-balancer requests.
- Documents the interface and integration requirements.

Model-specific configuration remains in the consuming recipe or launcher; the framework only manages allocation, lifecycle, routing, and endpoint injection.

## Compatibility

The feature is opt-in and does not modify ray.sub. Existing launch workflows are unchanged unless they explicitly use tools/external_gym_vllm/run_in_allocation.sh.

External replicas currently consume whole nodes, so each pool’s tensor-parallel size must be divisible by the configured GPUs per node.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


terrykong:
offline discussion was to generalize w/ hetgroups to re-use gang scheduling (will also help with head node on a separate node cc @macandro96 )

tools/external_gym_vllm/pool_config.sh
```diff
@@ -0,0 +1,298 @@
...
+  local display_name="${pool}"
+  local model=""
+  local container=""
+  local python=""
+  local replicas=""
+  local tensor_parallel_size=""
+  local lb_port=""
+  local url_placeholder=""
+  local group_id=""
+  local served_model_name="model"
```
terrykong, line 125: [`pool_config.sh:125`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/pool_config.sh#L125) — **PR-introduced**, with pre-existing context below.

**1 action item, 1 FYI.**

### AI-1 — the default `served-model-name` will 404 against what Gym actually sends

`POOL_SERVED_MODEL_NAME` defaults to the literal string `model`, which reaches vLLM as [`--served-model-name "${SERVED_MODEL_NAME}"`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L435). But Gym unconditionally overwrites the request's model field with its own config value ([`vllm_model/app.py:229`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/vllm_model/app.py#L229)):

```python
body_dict["model"] = self.config.model
```

So unless the consuming Gym server's config literally says `model: model`, every request 404s. The readiness probe [`curl -sfm 10 "${pool_urls[...]}/models"`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L647) doesn't compare names, so this passes startup and surfaces at the first judge call as `The model \`/lustre/ckpt/...\` does not exist.` — after the allocation is up.

**Action:** document the coupling in the [README row](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/README.md#L77) — "must equal the `model` field of the Gym server that calls this pool; Gym overwrites the request's `model` unconditionally". Optionally assert the served name appears in the `/models` response so this fails at startup instead. (I'd avoid silently defaulting it to `POOL_MODEL` — that breaks a config that legitimately says `model: model`.)

### FYI — not introduced here, but it blocks the documented path

[`ultra_launch.sh:91-93`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L91-L93) is an `if/elif`, so `GENRM_BASE_URL` and `GENRM_MODEL` are mutually exclusive — meaning a user on the external-endpoint path **cannot** set the `model` that would make the names match. Separately, [`VLLMModelConfig.model`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/vllm_model/app.py#L51) is a required non-Optional `str` while the exemplar ships `model: null`, so `GENRM_BASE_URL` alone fails config validation outright. That file isn't in this PR and the pre-PR docs already pointed at `GENRM_BASE_URL`, so this is context rather than something you broke — but it's worth knowing the documented path can't currently be made to work end to end.

tools/external_gym_vllm/run_in_allocation.sh
```diff
@@ -0,0 +1,685 @@
...
+fi
+if (( ${#external_nodes[@]} != total_external_nodes )); then
+  echo "[FATAL] Slurm hetgroup 1 has ${#external_nodes[@]} nodes, expected ${total_external_nodes}" >&2
+  for pool in "${pool_names[@]}"; do
+    echo "[FATAL]   ${display_names[${pool}]}: ${node_counts[${pool}]} nodes" >&2
+  done
+  exit 1
+fi
+
+LOG_DIR="${BASE_LOG_DIR}/${SLURM_JOB_ID}-logs"
```
terrykong: [`run_in_allocation.sh:239`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L239) — **PR-introduced.**

**2 action items.**

### AI-1 — mirror `ray.sub`'s restart-count suffix

`ray.sub` deliberately suffixes `LOG_DIR` with the restart count ([`ray.sub:194-200`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L194-L200)), commented *"so we don't clobber the previous attempt's logs (and signal files) under the same job ID"*. It computes this itself — it does not inherit yours — so on a requeued job the two diverge, and `ENDED` is your only teardown channel into it.

With `--requeue` and `SLURM_RESTART_COUNT=1`: launcher writes `$BASE_LOG_DIR/$JOBID-logs`, `ray.sub` polls `$BASE_LOG_DIR/$JOBID-1-logs` ([:635](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L635), [:866](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L866)) → [`touch "${LOG_DIR}/ENDED"`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L672) is never seen. The direct `kill "${ray_sub_pid}"` still fires, so this is a degraded teardown rather than a total one, but the in-container pollers miss it.

```suggestion
# Must match ray.sub's LOG_DIR: the ENDED file is the teardown channel between them.
if [[ -n "${SLURM_RESTART_COUNT:-}" ]]; then
  LOG_DIR="${BASE_LOG_DIR}/${SLURM_JOB_ID}-${SLURM_RESTART_COUNT}-logs"
else
  LOG_DIR="${BASE_LOG_DIR}/${SLURM_JOB_ID}-logs"
fi
```

### AI-2 — add `head_ip_*` to the startup cleanup

The `rm -f` list at [:248-254](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L248-L254) clears the registry and `*_url` files but not `head_ip_${replica_index}` ([:569](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L569)). Requeued into the same unsuffixed directory, a non-head task can win the race against the head's [`rm -f "${HEAD_IP_FILE}"`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L383): the `[[ -s … ]]` check at [:455](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L455) passes on the *previous* attempt's IP, and the worker burns 120×2s on `ray start --address=<dead head>` before `[FATAL] failed to join private Ray cluster`. Fixing AI-1 mostly sidesteps this, but the explicit cleanup is worth having.

tools/external_gym_vllm/run_in_allocation.sh
```diff
@@ -0,0 +1,685 @@
...
+    if ! kill -0 "${VLLM_PID}" 2>/dev/null; then
+      echo "[${REPLICA_ID}] ERROR: vLLM exited before becoming healthy" >&2
+      exit 1
+    fi
+    sleep 5
+  done
+
+  registry_add "${REPLICA_ID}" "${HEAD_IP}" "${VLLM_HTTP_PORT}"
+  echo "[${REPLICA_ID}] Registered healthy backend ${HEAD_IP}:${VLLM_HTTP_PORT}"
+  wait "${VLLM_PID}"
```
terrykong: [`run_in_allocation.sh:452`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L452) — **PR-introduced.** Hardening — I could not establish that vLLM's server actually exits 0 on its own, so treat this as closing a gap rather than fixing a live failure.

**1 action item.**

### AI-1 — treat *any* vLLM exit as a failure

If the server ever returns 0, `wait` returns 0, `set -e` never trips, and this is the last statement — so the head task exits 0 too. [`--kill-on-bad-exit=1`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L581) only fires on non-zero, and when `nodes_per_replica > 1` (the documented TP=8 / `GPUS_PER_NODE`=4 shape) the sibling tasks sit in [`tail -f /dev/null`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L486) keeping the step alive, so [`check_service_steps`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L304)'s `kill -0` still succeeds. Meanwhile [`cleanup_replica`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L373) has already deregistered the backend — the pool shrinks with nothing logged beyond a lone `Removed backend:` line in `load_balancer.log`, against the README's step-7 promise.

A vLLM server exiting at all is a failure, so make it explicit:

```suggestion
  wait "${VLLM_PID}"
  vllm_status=$?
  echo "[${REPLICA_ID}] vLLM exited with status ${vllm_status}" >&2
  exit $(( vllm_status == 0 ? 1 : vllm_status ))
```

(Worth pairing with replacing the workers' `tail -f /dev/null` with a poll of the head's `/health` that self-terminates, so a dead head doesn't leave siblings parked forever.)

tools/external_gym_vllm/run_in_allocation.sh
```diff
@@ -0,0 +1,685 @@
...
+: "${EXTERNAL_VLLM_POOLS:?EXTERNAL_VLLM_POOLS is required}"
+: "${EXTERNAL_VLLM_TOOLS_DIR_HOST:?EXTERNAL_VLLM_TOOLS_DIR_HOST is required}"
+
+if [[ "${SLURM_HET_SIZE}" != "2" ]]; then
+  echo "[FATAL] Expected exactly two Slurm hetgroups, got ${SLURM_HET_SIZE}" >&2
+  exit 1
+fi
+
+RAY_SUB="${RAY_SUB:-${SLURM_SUBMIT_DIR}/ray.sub}"
+export GPUS_PER_NODE="${GPUS_PER_NODE:-4}"
```
terrykong, line 42: [`run_in_allocation.sh:42`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L42) — **PR-introduced.** Usability / fail-late, not a correctness bug.

**1 action item.**

### AI-1 — document that this export also reconfigures `ray.sub`

This is the one launcher variable that shadows an existing `ray.sub` knob with a *different* default — [`ray.sub:228`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L228) uses `${GPUS_PER_NODE:-8}` — and because it's **exported**, it silently overrides `ray.sub`'s default for the training component too.

I'm not suggesting you change the default (4 is clearly deliberate: it's documented at [README:115](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/README.md#L115) and matches the `--gres=gpu:4` sbatch example and GB200's 4-GPU nodes). The issue is only that a user who omits it on 8-GPU hardware gets no early signal:
- On a **GRES** cluster: topology validation passes, the whole fleet warms up (bounded only by `STARTUP_TIMEOUT=3600`), and only then does `ray.sub` abort at [`:50`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L50) with `Error: GPUS_PER_NODE=4 but GRES detected is gpu:8 …` — loud, but after the allocation is spent.
- On a cluster with **no GRES** configured, that assert never runs ([`ray.sub:46`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L46) gates it), so `4` flows into [`NUM_ACTORS`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L386) and training quietly uses half the GPUs.

**Action:** add a clause to the README's optional-globals row noting that this value is exported into `ray.sub` and must match the real GPUs-per-node for both components. Note [`pool_config.sh`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/pool_config.sh#L68) carries the same `:-4` fallback, and reads it at *registration* time — so a launcher that sets `GPUS_PER_NODE` after `register_external_vllm_pool` sizes hetgroup 1 with the wrong divisor.

tools/external_gym_vllm/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+Registration appends `SAFETY` to `EXTERNAL_VLLM_POOLS` and exports the
+normalized `SAFETY_*` contract inherited by `sbatch`. Adding this pool requires
+no change to `run_in_allocation.sh`.
+
+The generated fields are:
+
+| Variable | Required | Default | Purpose |
+|---|---:|---|---|
+| `POOL_MODEL` | yes | — | `/lustre` checkpoint path or Hugging Face model ID. |
```
terrykong: [`README.md:68`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/README.md#L68) — **PR-introduced.** Three small doc items, all in this file.

**3 action items.**

### AI-1 — `/lustre` is hardcoded in two rows that `EXTERNAL_VLLM_SHARED_ROOT` generalizes

This row says "`/lustre` checkpoint path" and [`POOL_SHARED_PATHS`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/README.md#L80) says "visible through the `/lustre` mount", but both the guard ([`run_in_allocation.sh:215`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L215)) and the mount ([`:497`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L497)) key off `EXTERNAL_VLLM_SHARED_ROOT`, documented four rows down at [:118](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/README.md#L118). A site on `/gpfs` reading these two rows would conclude the tool is Lustre-only.

### AI-2 — `NUM_EXTERNAL_SERVICE_NODES` is used but never documented

The example calls `validate_external_vllm_submission "$COMMAND" "$NUM_EXTERNAL_SERVICE_NODES"` ([:57](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/README.md#L57)), but that name appears in neither globals table. When it's unset the node-count check is **silently skipped** ([`pool_config.sh:229`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/pool_config.sh#L229), [:242](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/pool_config.sh#L242)) and validation returns 0 — so the mistake surfaces an hour later at [`run_in_allocation.sh:231`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L231) as `[FATAL] Slurm hetgroup 1 has N nodes, expected M`, which is exactly what the pre-`sbatch` validator exists to prevent. A table row, and ideally a warning when it's skipped.

### AI-3 — the reasoning-parser contract is one-sided

The argv section ([:94-96](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/README.md#L94-L96)) says parsers belong in the pool definition, but not that they must **match** the consuming Gym server. Gym hard-asserts one direction ([`vllm_model/app.py:440-444`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/vllm_model/app.py#L440-L444)): `uses_reasoning_parser: false` plus a pool passing `--reasoning-parser` raises `NeMo Gym server ... config has explicitly been set to not use a reasoning parser`. One sentence would save that trip.

terrykong (commented):
Reviewed by a team of specialised agents (RL/guidelines, bug-finder, tests, design, Gym integration, devil's advocate), with every finding adversarially challenged before posting. Thanks for this — the hetgroup rework paid off. A few things worth calling out as done well before the list:

- **The PR touches zero pre-existing runtime files.** `ray.sub` and `docs/cluster.md` are byte-identical to `main`; the only non-new files are a docs paragraph and two `pyrefly.toml` lines. That makes it easy to reason about and trivial to revert.
- **`vllm_pool_lb.py` is well-factored.** `Backend` / `BackendPool` / `LoadBalancer` are three clean responsibilities with constructor-injected dependencies, and `pick()` / `_extract_affinity_key()` are pure — which is why the tests can assert real routing behaviour without a Slurm allocation. We checked whether anything wanted extracting and concluded no: the state doesn't partition.
- **The up-front validation block** fails before a single service starts, with specific messages. On an 80-node allocation that's the difference between failing in 2 seconds and burning an hour.
- **The shell is properly tested** — four `run_in_allocation.sh` tests and five `pool_config.sh` tests run the real scripts as subprocesses and assert their `[FATAL]` output, plus the registry and watchdog tests.
- **The generalization is data-driven, not conditional-driven.** Pool differences ride on `POOL_ENV_VARS` / `POOL_VLLM_ARGS` consumed by one uniform loop — no per-pool-type branching anywhere, and no `eval`.
- **The previous round's tautological hetgroup test is properly fixed.** Verified by mutation: swapping every `--het-group=0`↔`=1` in the source now makes `test_launcher_routes_generic_pools_to_explicit_hetgroups` fail. That's the right way to write it.

Five comments below, scoped to the launcher and config surface — the places where a mistake costs a whole allocation rather than a retry. None of them block; the two I'd most want addressed are the `served-model-name` default and the exported `GPUS_PER_NODE`, since both fail *late* and neither is discoverable from the README as written.

**@terrykong — on your hetgroup comment:** the gang-scheduling half landed cleanly; every `srun` is pinned to its component and Slurm does the co-scheduling. The head-node half got answered from a different direction while this PR was in flight: [#3388](https://github.com/NVIDIA-NeMo/RL/pull/3388) landed `DEDICATED_RAY_HEAD` on `main` ([`ray.sub:381`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L381), [`:703-705`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/ray.sub#L703-L705)), and it passes straight through this launcher — `ray.sub` is started via `env … bash "${RAY_SUB}"`, so setting it in the sbatch environment needs no launcher change. It composes better than by accident: the launcher already pins every load balancer to `ray_nodes[0]`, which under `DEDICATED_RAY_HEAD=1` is precisely the GPU-free node. Nothing in the new README mentions this, and the sizing rule (hetgroup 0 = N+1, `cluster.num_nodes` = N, which `ray.sub` warns it "cannot validate for you") is exactly what users get wrong — worth a row. What still isn't expressible is a *third* component for a genuinely GPU-free head: [`:36`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L36) hard-rejects `SLURM_HET_SIZE != 2`, and `docs/cluster.md` notes the head's GPUs today are "idled, not freed". Relaxing that to `>= 2` with the group indices as env vars would make it a config change later rather than a script edit. cc @macandro96.

**On the description:** "backend failover" is doing slightly more work than the code currently supports. The load balancer really does retry across healthy backends, but [`check_service_steps`](https://github.com/NVIDIA-NeMo/RL/blob/ac9e00db52c9a3a19e86ce3d54a5a1fcd25246ea/tools/external_gym_vllm/run_in_allocation.sh#L301-L315) tears the whole job down within ~5s of any replica step exiting, so the "a backend dies and the pool keeps serving" case can't be reached today — per-request failover off a live-but-erroring backend still works. The README's step 7 ("stops training if any required external service exits") states the opposite policy, so the two just need to agree on which one you meant.

**One structural note:** the framework currently ships with no in-tree caller — nothing outside `tools/external_gym_vllm/` invokes `register_external_vllm_pool`, so the contract (including the `served-model-name` coupling in the comment below) isn't exercised by anything that runs in CI. Landing the `ultra_launch.sh` wiring or an `examples/` snippet in a follow-up would give it one real consumer.

**One non-blocking ask:** if you have a smoke run handy, the wall-clock from launch to `[INFO] External vLLM pools are healthy` would tell users whether `POOL_STARTUP_TIMEOUT=3600` has headroom or is uncomfortably tight. It's the one shipped constant nothing else validates.

_Generated by Claude Code_

tools/external_gym_vllm/run_in_allocation.sh
yfw: Addressed in 540db669e37f277de8d63064a0c11de2cd433322

tools/external_gym_vllm/run_in_allocation.sh
yfw: Addressed in 540db669e37f277de8d63064a0c11de2cd433322

tools/external_gym_vllm/run_in_allocation.sh
yfw: Addressed in 540db669e37f277de8d63064a0c11de2cd433322

tools/external_gym_vllm/README.md
yfw: Addressed in 540db669e37f277de8d63064a0c11de2cd433322

tools/external_gym_vllm/pool_config.sh
yfw: Addressed in 540db669e37f277de8d63064a0c11de2cd433322

yfw:
Re `POOL_STARTUP_TIMEOUT=3600`, looking at some recent runs with 8 TP=8 GenRM replicas and 4 TP=4 NL2Bash replicas, it took ~16 mins on average for external vllms to become healthy, so startup time of 3600 leaves substantial headroom.

terrykong (approved):
All five findings verified as addressed in 540db669e37f277de8d63064a0c11de2cd433322 — I checked the code rather than the reply:

- **`LOG_DIR`** — both action items: the `SLURM_RESTART_COUNT` suffix now mirrors `ray.sub`, and `head_ip_*` is added to the startup cleanup.
- **Replica exit** — `wait "${VLLM_PID}"` now always logs and exits non-zero, including the clean-exit case that previously shrank the pool silently.
- **`POOL_SERVED_MODEL_NAME`** — the coupling to Gym's `model` field is documented, and the README example now passes `--served-model-name` explicitly, which is the part that stops people hitting it.
- **`GPUS_PER_NODE`** — documented as set-before-registration and exported to `ray.sub`; the example sets it explicitly, which also closes the registration-time ordering hazard.
- **Docs** — `/lustre` literals replaced with `EXTERNAL_VLLM_SHARED_ROOT`, `NUM_EXTERNAL_SERVICE_NODES` documented plus a runtime warning when it's skipped, and the reasoning-parser ↔ `uses_reasoning_parser` contract spelled out.

Bonus: documenting `DEDICATED_RAY_HEAD` and the hetgroup-0 `N+1` sizing rule answers @terrykong's original ask directly — that was the thing most likely to bite the first user.

Nice touch adding assertions for each fix rather than just making the change.

LGTM.

_Generated by Claude Code_

terrykong (approved):

PR #3388, opened 2026-07-28
feat(ray): support a dedicated gpu-free head node

macandro96:
# What does this PR do ?

Allow the Ray head to run as a control-plane-only node by advertising zero GPUs and excluding it from worker-unit accounting. Can be switched on setting `DEDICATED_RAY_HEAD=1` on in env.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/ray.sub
+++ b/ray.sub
@@ -226,6 +226,14 @@ fi
 
 # Number of GPUs per worker node
 GPUS_PER_NODE=${GPUS_PER_NODE:-8}
+DEDICATED_RAY_HEAD=${DEDICATED_RAY_HEAD:-0}
+case "$DEDICATED_RAY_HEAD" in
+  0|1) ;;
+  *)
+    echo "Error: DEDICATED_RAY_HEAD must be 0 or 1, got '$DEDICATED_RAY_HEAD'." >&2
+    exit 1
+    ;;
+esac
 
 # Detect GRES support and set GRES_ARG
 log_phase "detecting GRES support (sinfo)"
@@ -367,8 +375,15 @@ export RAY_health_check_period_ms="${RAY_health_check_period_ms:-10000}"
 num_retries=3
 WORKER_NUM_RETRIES=20
 
-# Total Ray worker_units expected once every node has registered with the head.
-NUM_ACTORS=$((GPUS_PER_NODE * SLURM_JOB_NUM_NODES))
+# Total Ray worker_units expected once every schedulable GPU node has
+# registered. In dedicated-head mode the first physical node runs only the Ray
+# control plane and driver, and intentionally contributes no GPU worker units.
+NUM_RAY_GPU_NODES=$((SLURM_JOB_NUM_NODES - DEDICATED_RAY_HEAD))
+if (( NUM_RAY_GPU_NODES <= 0 )); then
+  echo "Error: dedicated Ray head requires at least one additional GPU worker node." >&2
+  exit 1
+fi
+NUM_ACTORS=$((GPUS_PER_NODE * NUM_RAY_GPU_NODES))
 
 # Track backgrounded srun client PIDs for head and workers
 declare -A SRUN_PIDS
@@ -665,6 +680,16 @@ sed -i 's/context\.py_executable = " "\.join(self\.nsight_cmd) + " python"/conte
 
 source $LOG_DIR/topology_probe.sh
 
+# Keep model, generation, teacher, and Gym GPU actors off the Ray head. Omitting
+# worker_units and topology resources also prevents topology-constrained
+# placement groups from selecting it as a compute node.
+HEAD_RAY_RESOURCE_ARGS=""
+if [[ "$DEDICATED_RAY_HEAD" == "1" ]]; then
+  RAY_RESOURCES='{\"slurm_managed_ray_cluster\": 1, \"ray_head\": 1}'
+  HEAD_RAY_RESOURCE_ARGS="--num-gpus=0"
+  echo "[INFO] Dedicated Ray head enabled: advertising zero GPUs"
+fi
+
 # Clear MPI/PMIx/SLURM env vars inherited from the srun launcher so TRT-LLM's
 # MPI_Init (in the ray daemon child process below) starts clean.
 # NOTE: keep this after topology_probe.sh, which reads SLURM_TOPOLOGY_ADDR/SLURM_PROCID.
@@ -693,6 +718,7 @@ while [[ \$count -lt $num_retries ]]; do
 ray start --head \
     --disable-usage-stats \
     --resources="\$RAY_RESOURCES" \
+    \$HEAD_RAY_RESOURCE_ARGS \
     --node-ip-address="$head_node_ip" \
     --port=${PORT} \
     --ray-client-server-port=${RAY_CLIENT_SERVER_PORT} \
--- a/ray.sub
+++ b/ray.sub
@@ -384,6 +384,10 @@ if (( NUM_RAY_GPU_NODES <= 0 )); then
   exit 1
 fi
 NUM_ACTORS=$((GPUS_PER_NODE * NUM_RAY_GPU_NODES))
+if (( DEDICATED_RAY_HEAD == 1 )); then
+  echo "[INFO] DEDICATED_RAY_HEAD=1: $NUM_RAY_GPU_NODES of $SLURM_JOB_NUM_NODES allocated nodes are GPU compute nodes."
+  echo "[INFO] Size your driver COMMAND to $NUM_RAY_GPU_NODES nodes (cluster.num_nodes, plus env.nemo_gym.num_gpu_nodes when set); ray.sub cannot validate this for you."
+fi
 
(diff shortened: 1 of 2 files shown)


ray.sub
```diff
@@ -367,8 +375,15 @@ export RAY_health_check_period_ms="${RAY_health_check_period_ms:-10000}"
...
-NUM_ACTORS=$((GPUS_PER_NODE * SLURM_JOB_NUM_NODES))
+# Total Ray worker_units expected once every schedulable GPU node has
+# registered. In dedicated-head mode the first physical node runs only the Ray
+# control plane and driver, and intentionally contributes no GPU worker units.
+NUM_RAY_GPU_NODES=$((SLURM_JOB_NUM_NODES - DEDICATED_RAY_HEAD))
+if (( NUM_RAY_GPU_NODES <= 0 )); then
+  echo "Error: dedicated Ray head requires at least one additional GPU worker node." >&2
+  exit 1
+fi
+NUM_ACTORS=$((GPUS_PER_NODE * NUM_RAY_GPU_NODES))
```
terrykong, line 386: [`ray.sub:378-386`](https://github.com/NVIDIA-NeMo/RL/blob/c12c0df3ccd094f0389d2eaa4f3d0bda78ffc1a9/ray.sub#L378-L386)

Nice feature, and the `0|1` + `NUM_RAY_GPU_NODES <= 0` guards are the right fail-loud shape.

Usability gap (not a correctness bug): `ray.sub` computes the effective GPU-node count here and then discards it, but `cluster.num_nodes` lives in the recipe YAML and `SLURM_JOB_NUM_NODES` has zero references in `nemo_rl/` — so setting `DEDICATED_RAY_HEAD=1` on a working `--nodes=4` job leaves `cluster.num_nodes: 4` requesting 32 GPUs against 24. Bringup passes at 24/24 ([`ray.sub:771`](https://github.com/NVIDIA-NeMo/RL/blob/c12c0df3ccd094f0389d2eaa4f3d0bda78ffc1a9/ray.sub#L771)), then the driver dies minutes later — after 6 retries ([`virtual_cluster.py:809-831`](https://github.com/NVIDIA-NeMo/RL/blob/c12c0df3ccd094f0389d2eaa4f3d0bda78ffc1a9/nemo_rl/distributed/virtual_cluster.py#L809-L831)) — on `Maximum number of retries reached (6). Cluster resources may be insufficient...`, which names no counts and never mentions the flag.

I couldn't find a way to *validate* this in `ray.sub` (`cluster.num_nodes` is inside the user's `COMMAND`), so the pragmatic ask is just to print the answer:

```suggestion
NUM_ACTORS=$((GPUS_PER_NODE * NUM_RAY_GPU_NODES))
if (( DEDICATED_RAY_HEAD == 1 )); then
  echo "[INFO] DEDICATED_RAY_HEAD=1: $NUM_RAY_GPU_NODES of $SLURM_JOB_NUM_NODES nodes are GPU compute nodes. Set cluster.num_nodes=$NUM_RAY_GPU_NODES in your driver COMMAND."
fi
```

**Also:** a `DEDICATED_RAY_HEAD=0` row in the *Advanced Environment Configuration* table ([`docs/cluster.md`](https://github.com/NVIDIA-NeMo/RL/blob/c12c0df3ccd094f0389d2eaa4f3d0bda78ffc1a9/docs/cluster.md#L127)) — `grep -rn DEDICATED docs/` returns nothing today. Worth noting there that `cluster.num_nodes` must be `--nodes - 1`; that the head's GPUs are idled rather than freed (it keeps `--gres` via `COMMON_SRUN_ARGS` under `#SBATCH --exclusive`, so this costs one node's GPUs); and that `--nodes = N+1` can trip `tools/launch`'s segment-divisibility check ([`tools/launch:138-142`](https://github.com/NVIDIA-NeMo/RL/blob/c12c0df3ccd094f0389d2eaa4f3d0bda78ffc1a9/tools/launch#L138-L142)).

ray.sub
```diff
@@ -665,6 +680,16 @@ sed -i 's/context\.py_executable = " "\.join(self\.nsight_cmd) + " python"/conte
 
 source $LOG_DIR/topology_probe.sh
 
+# Keep model, generation, teacher, and Gym GPU actors off the Ray head. Omitting
+# worker_units and topology resources also prevents topology-constrained
+# placement groups from selecting it as a compute node.
+HEAD_RAY_RESOURCE_ARGS=""
+if [[ "$DEDICATED_RAY_HEAD" == "1" ]]; then
+  RAY_RESOURCES='{\"slurm_managed_ray_cluster\": 1, \"ray_head\": 1}'
```
terrykong: [`ray.sub:687-688`](https://github.com/NVIDIA-NeMo/RL/blob/c12c0df3ccd094f0389d2eaa4f3d0bda78ffc1a9/ray.sub#L687-L688)

Low-severity / fine to defer. The head's resource JSON now has two construction sites: `topology_probe.sh` is the canonical incremental builder ([`ray.sub:559-566`](https://github.com/NVIDIA-NeMo/RL/blob/c12c0df3ccd094f0389d2eaa4f3d0bda78ffc1a9/ray.sub#L559-L566)), sourced by head *and* workers, and this line discards its result and re-hardcodes a two-key literal.

That file already carries a "these key names must stay in sync with `NVLINK_DOMAIN_PREFIX`/`TOPO_RANK_KEY` in `virtual_cluster.py`" contract ([`ray.sub:556-558`](https://github.com/NVIDIA-NeMo/RL/blob/c12c0df3ccd094f0389d2eaa4f3d0bda78ffc1a9/ray.sub#L556-L558)), so node-level keys getting added there is a live pattern. When one is, workers get it and the dedicated head silently won't — surfacing as a confusing "one node is missing X" only under `DEDICATED_RAY_HEAD=1`, 130 lines from the code that caused it.

**Action:** one option, of several — make the builder the single source of truth by exporting a marker before `source $LOG_DIR/topology_probe.sh` and branching inside the probe, so only keys deliberately placed in the compute-only branch are dropped:

```bash
# in head_cmd, before sourcing topology_probe.sh
export NRL_DEDICATED_RAY_HEAD=$DEDICATED_RAY_HEAD

# in topology_probe.sh (~L559)
RAY_RESOURCES='{\"slurm_managed_ray_cluster\": 1'
if [[ "\${NRL_DEDICATED_RAY_HEAD:-0}" == "1" ]]; then
  RAY_RESOURCES+=', \"ray_head\": 1'
else
  RAY_RESOURCES+=', \"worker_units\": '"$GPUS_PER_NODE"
  ...
fi
```

terrykong (commented):
Reviewed by an agent team (4 agents: RL/infra expert, bug finder, design reviewer, plus an adversarial pass). Useful feature — the `0|1` validation and the `NUM_RAY_GPU_NODES <= 0` guard are good fail-loud practice, and the core mechanism holds up.

Things we checked and found **correct**, so they don't need defending:
- Heredoc escaping of the new `RAY_RESOURCES` override — reproduced the full `EOF` -> `EOFINNER` -> `bash` chain against a `ray` argv stub; both modes emit valid JSON, at the same escaping level as the existing `topology_probe.sh` path.
- `--num-gpus=0` semantics, verified against Ray 2.47 `resource_spec.py` — autodetect is skipped and the zero-valued `GPU` key is dropped, so the head's `--gres` can't leak back in.
- `NUM_ACTORS` accounting matches the workers actually launched (`_num_workers = N-1`) in both modes.
- `ray status | grep worker_units | awk` still parses correctly when only workers carry the resource.
- Dropping `worker_units`/`topo_rank`/`nvlink_domain_*` from the head breaks nothing on the Python side — `select_segment_nodes` already skips `NVLINK_DOMAIN_UNKNOWN`, and `prepare_segment_topology` gates on `any(...)` not `all(...)`, so topology-aware placement is not disabled cluster-wide.
- The early `exit 1` paths not touching `ENDED` are fine — both fire before any `srun`.

Three comments below: one on user-facing accounting/docs, one open question about NeMo-Gym compatibility (not reproduced — please sanity-check it), and one deferrable maintainability note.

Note we could not run the repo's linters in this environment (the lockfile is Linux-only), so there is no lint signal from us; the diff is shell-only.

_Generated by Claude Code_

terrykong (commented):
lgtm modulo review. created https://github.com/NVIDIA-NeMo/RL/issues/3474 to track the proper fix

macandro96:
/ok to test [1f399f3](https://github.com/NVIDIA-NeMo/RL/pull/3388/commits/1f399f3af34a86133601dcf73c2245ba35a140f6)

macandro96:
/ok to test [c712eca](https://github.com/NVIDIA-NeMo/RL/commit/c712ecaf82510dfc6735a02c79a50021782f1a30)

ray.sub
terrykong: fixed in c712ecaf82510dfc6735a02c79a50021782f1a30

ray.sub
terrykong: fixed in c712ecaf82510dfc6735a02c79a50021782f1a30

terrykong (approved):

PR #3389, opened 2026-07-28
fix(data): honor subset on GSM8KDataset and correct the val-split docs

tianyi-zhang-02:
## What does this PR do ?

Two small follow-ups to the config-key guard merged in #3271.

**1. `GSM8KDataset` now honors `subset`.** It hardcoded the `"main"` Hugging Face config, so a configured `subset` was silently ignored — `openai/gsm8k` also ships a `"socratic"` config (same questions, answers interleaved with Socratic sub-questions). `subset` is now threaded to `load_dataset`, defaulting to `"main"` so existing behavior is unchanged. This is exactly the example the #3271 guard warns about, so that warning now resolves into working behavior for `gsm8k`.

**2. The `split_validation_size` docs were stale.** Both guides enumerated the supporting datasets, and the lists had drifted:

| | datasets listed |
|---|---|
| `docs/guides/sft.md` | 4 |
| `docs/guides/grpo.md` | 5 |
| actually call `split_train_validation` | **12** — arrow_text, audiomcq, avqa, daily_omni, general_conversations, mmpr_tiny, nemotron_cascade2_sft, oasst, openmathinstruct2, openr1_math, response_dataset, tulu3 |

A reader would wrongly conclude the 7 unlisted datasets can't carve a validation split. Rather than re-listing 12 names that will drift again, the docs now state the **rule** that determines support (the dataset class calls `split_train_validation`), keep the common datasets as examples, and note that an unsupported key now warns rather than being silently dropped (#3271).

## Issues

Follow-up to #3270 / #3271. No new issue filed — these are small corrections in the same area.

## Usage

```yaml
data:
  train:
    dataset_name: gsm8k
    subset: socratic   # before: silently loaded "main"; after: loads the socratic config
```

- [x] New tests: `test_gsm8k_subset_selects_hf_config` (asserts `subset` reaches `load_dataset` and answer extraction still works) and `test_gsm8k_subset_defaults_to_main` (backward compatibility), following the mocked pattern used by the AIME test in the same file.
- [x] Validation: both cases verified locally against the real loader; `ruff check` / `format --check` clean; `pyrefly==0.24.2` reports 0 errors on `gsm8k.py` (it is in `project-includes`); all doc links resolve.
- [x] Docs: updated (that's half this PR).

cc @yuki-97 @RayenTian


nemo_rl/data/datasets/response_datasets/gsm8k.py
```diff
@@ -44,7 +48,7 @@ def __init__(
         self.extract_answer = extract_answer
 
         # load from huggingface
-        self.dataset = load_dataset("openai/gsm8k", "main")[split]
+        self.dataset = load_dataset("openai/gsm8k", subset)[split]
```
yuki-97: Two stale references to "subset on GSM8KDataset" survive this PR — both are the canonical example of a silently-swallowed config key, but the hardcode being fixed here is exactly what made them true:

- `nemo_rl/data/datasets/response_datasets/__init__.py:126` — inline comment above the `warn_on_unsupported_dataset_config_keys` call.
- `nemo_rl/data/datasets/utils.py:212` — the helper's own docstring (`...``subset`` on ``GSM8KDataset`` (which hardcodes the ``main`` config)...`).

**Suggest fix:** either drop the `subset` example (leaving the `split_validation_size` one, which stays accurate), or swap in another dataset that still ignores `subset`.

nemo_rl/data/datasets/response_datasets/gsm8k.py
```diff
@@ -29,12 +29,16 @@ class GSM8KDataset(RawDataset):
...
         split: Split name for the dataset, default is "train"
         extract_answer: Whether to extract the answer from the dataset, default is True
     """
 
     def __init__(
         self,
+        subset: str = "main",
         split: str = "train",
         extract_answer: bool = True,
         system_prompt_file: str | None = None,
```
yuki-97: nit: not introduced in this PR but `system_prompt_file` can be removed. it will be set through `set_task_spec‎` in `nemo_rl/data/datasets/response_datasets/__init__.py`.
```suggestion
```

yuki-97 (commented):
LGTM, thanks @tianyi-zhang-02 , only small nits.

nemo_rl/data/datasets/response_datasets/gsm8k.py
tianyi-zhang-02: Thanks! fixed in `2f72364`. Dropped the `subset`/GSM8KDataset example from both places, keeping the `split_validation_size` one since it stays accurate. Went with dropping rather than swapping in another dataset since naming one just moves the staleness to whenever that gets fixed.

nemo_rl/data/datasets/response_datasets/gsm8k.py
tianyi-zhang-02: Done in `2f72364`. It was unused in `gsm8k.py`, and it isn't a guarded key, so configs still setting it are absorbed by `**kwargs` with no warning.

yuki-97 (dismissed):
LGTM, thanks @tianyi-zhang-02

tianyi-zhang-02:
Updated

tianyi-zhang-02:
Current with main. Ready whenever you get a chance, @yuki-97. Thanks!

yuki-97 (approved):

PR #3390, opened 2026-07-28
perf(vllm): return token metadata with chat completions

jthomson04:
## Summary

This PR removes a redundant, awaited `/tokenize` round trip from every raw-vLLM model turn. Under a 64-concurrent multi-turn workload, that round trip was on the session critical path and accumulated into a significant model-serving bottleneck.

- attach exact prompt token IDs, generation token IDs, and selected-token log probabilities to raw-vLLM chat-completion messages
- match vLLM generation outputs to response choices by choice index and validate token/logprob alignment
- preserve the dynamic token metadata in the serialized OpenAI-compatible response
- retain Gym's existing `/tokenize` compatibility fallback unchanged

## Performance impact

In a matched vLLM 0.23.0 benchmark using Gym `eddd5e98`, this change:

- reduced summed model-request-wave span from **2,683.4s to 2,291.3s**: a **392.1s (14.6%) reduction**, or **1.17x speedup**
- eliminated `/tokenize` entirely across **18,078 successful model calls**
- brought raw vLLM to within **3.5% of Dynamo + vLLM** on model-request span
- closed **83% of the previously observed raw-vLLM-to-Dynamo serving gap**

Because Gym awaited `/tokenize` before starting the next turn, this was not background bookkeeping overhead: its latency directly serialized multi-turn sessions. Returning metadata already available in vLLM's final `RequestOutput` removes that critical-path work without changing generation behavior.

## Root cause

The raw-vLLM backend already had the exact prompt and generated token IDs in the final vLLM `RequestOutput`, but discarded them while serializing the chat-completion response. Gym therefore issued a separate awaited `/tokenize` request on every model turn to reconstruct prompt token IDs, adding inter-turn latency under concurrency.

## Gym compatibility

No paired Gym change or submodule bump is required. NeMo-RL's pinned Gym revision, `473f446`, already guards the compatibility fallback on the absence of native metadata:

```python
if self.config.return_token_id_information and "prompt_token_ids" not in choice_dict["message"]:
```

Both `client.create_tokenize(...)` and the subsequent `message_dict.update(...)` are inside that block. Because this PR serializes `prompt_token_ids` on `choice.message`, pinned Gym skips the entire fallback. The fallback remains available for other servers that do not return native token metadata.

## Validation

- focused utility suite: 34 passed
- Ruff check and format checks passed on all changed files
- matched benchmark:
  - 320/320 valid trajectories
  - zero `/tokenize` calls
  - contiguous per-session multi-turn call sequences

The whole-run first-request-to-last-response span was 3,290.9s because this run accumulated 999.6s of between-wave environment delay, dominated by one 1,008s OpenHands trajectory. The benchmark therefore supports model-serving parity, not environment-inclusive wall-time parity.

## Known Gym payload behavior

Gym's native-metadata path skips its compatibility cleanup, so the standard OpenAI `choice.logprobs` block remains alongside `choice.message.generation_log_probs`. NeMo-RL intentionally preserves `choice.logprobs` to maintain the endpoint's OpenAI response contract.

### DIFF
--- a/nemo_rl/models/generation/vllm/utils.py
+++ b/nemo_rl/models/generation/vllm/utils.py
@@ -318,17 +318,105 @@ def attach_routed_experts_to_chat_response_choices(
     return response
 
 
+def attach_token_information_to_chat_response_choices(
+    response: Any,
+    final_request_output: Any,
+) -> Any:
+    """Attach engine-native token information to OpenAI chat response choices."""
+    prompt_token_ids = getattr(final_request_output, "prompt_token_ids", None)
+    if prompt_token_ids is None:
+        raise RuntimeError(
+            "vLLM was asked to return token information for the "
+            "OpenAI-compatible chat endpoint but the final request output did "
+            "not include prompt_token_ids."
+        )
+
+    outputs_by_index = {
+        output.index: output for output in getattr(final_request_output, "outputs", [])
+    }
+    choices = list(getattr(response, "choices", []))
+    choice_indices = [choice.index for choice in choices]
+    if len(outputs_by_index) != len(
+        getattr(final_request_output, "outputs", [])
+    ) or len(set(choice_indices)) != len(choice_indices):
+        raise RuntimeError(
+            "vLLM returned duplicate choice indices while attaching token "
+            "information to the OpenAI-compatible chat response."
+        )
+
+    missing_choice_indices = sorted(set(choice_indices) - outputs_by_index.keys())
+    if missing_choice_indices:
+        raise RuntimeError(
+            "vLLM was asked to return token information for the "
+            "OpenAI-compatible chat endpoint but response choices could not be "
+            "matched to generation outputs: "
+            f"missing_choice_indices={missing_choice_indices}."
+        )
+
+    for choice in choices:
+        generation_details = outputs_by_index[choice.index]
+        output_token_ids = getattr(generation_details, "token_ids", None)
+        if output_token_ids is None:
+            raise RuntimeError(
+                "vLLM was asked to return token information for the "
+                "OpenAI-compatible chat endpoint but generation output "
+                f"choice_idx={choice.index} did not include token_ids."
+            )
+        generation_token_ids = list(output_token_ids)
+
+        logprob_content = getattr(getattr(choice, "logprobs", None), "content", None)
+        if logprob_content is None:
+            if generation_token_ids:
+                raise RuntimeError(
+                    "vLLM was asked to return token information for the "
+                    "OpenAI-compatible chat endpoint but response "
+                    f"choice_idx={choice.index} did not include logprobs."
+                )
+            generation_log_probs = []
+        else:
+            if len(generation_token_ids) != len(logprob_content):
+                raise RuntimeError(
+                    "vLLM returned mismatched generation token IDs and log "
+                    "probabilities for the OpenAI-compatible chat endpoint: "
+                    f"choice_idx={choice.index}, "
+                    f"token_count={len(generation_token_ids)}, "
+                    f"logprob_count={len(logprob_content)}."
+                )
+            generation_log_probs = []
+            for token_id, logprob_entry in zip(generation_token_ids, logprob_content):
+                if logprob_entry.token != f"token_id:{token_id}":
+                    raise RuntimeError(
+                        "vLLM returned a log probability for an unexpected "
+                        "token while attaching token information to the "
+                        "OpenAI-compatible chat response: "
+                        f"choice_idx={choice.index}, token_id={token_id}, "
+                        f"logprob_token={logprob_entry.token!r}."
+                    )
+                generation_log_probs.append(float(logprob_entry.logprob))
+
+        choice.message.prompt_token_ids = list(prompt_token_ids)
+        choice.message.generation_token_ids = generation_token_ids
+        choice.message.generation_log_probs = generation_log_probs
+
+    return response
+
+
 def model_dump_chat_response_with_routed_experts(response: Any) -> dict[str, Any]:
-    """Dump a vLLM OpenAI chat response while preserving dynamic R3 fields."""
+    """Dump a vLLM OpenAI chat response while preserving dynamic message fields."""
     response_dict = response.model_dump()
     for choice, choice_dict in zip(
         getattr(response, "choices", []), response_dict.get("choices", [])
     ):
-        routed_experts = getattr(
-            getattr(choice, "message", None), "routed_experts", None
-        )
-        if routed_experts is not None:
-            choice_dict.setdefault("message", {})["routed_experts"] = routed_experts
+        message = getattr(choice, "message", None)
+        for field_name in (
+            "routed_experts",
+            "prompt_token_ids",
+            "generation_token_ids",
+            "generation_log_probs",
+        ):
+            field_value = getattr(message, field_name, None)
+            if field_value is not None:
+                choice_dict.setdefault("message", {})[field_name] = field_value
     return response_dict
 
 
--- a/nemo_rl/models/generation/vllm/utils.py
+++ b/nemo_rl/models/generation/vllm/utils.py
@@ -26,6 +26,7 @@ from nemo_rl.models.generation.interfaces import (
 from nemo_rl.utils.routed_experts_codec import encode_routed_experts
 
 R3_MISSING_ROUTE_SENTINEL = ROUTED_EXPERTS_MISSING_ROUTE_SENTINEL
+VLLM_LOGPROB_FLOOR = -9999.0
 
 # The expert-id range vs carry dtype is model-constant, so it is verified on the
 # first non-empty routed-experts tensor per process and skipped afterwards.
@@ -331,26 +332,34 @@ def attach_token_information_to_chat_response_choices(
             "not include prompt_token_ids."
         )
 
-    outputs_by_index = {
-        output.index: output for output in getattr(final_request_output, "outputs", [])
-    }
-    choices = list(getattr(response, "choices", []))
-    choice_indices = [choice.index for choice in choices]
-    if len(outputs_by_index) != len(
-        getattr(final_request_output, "outputs", [])
-    ) or len(set(choice_indices)) != len(choice_indices):
+    generation_outputs = list(getattr(final_request_output, "outputs", []))
+    generation_output_indices = [output.index for output in generation_outputs]
+    outputs_by_index = {output.index: output for output in generation_outputs}
+    if len(outputs_by_index) != len(generation_outputs):
         raise RuntimeError(
-            "vLLM returned duplicate choice indices while attaching token "
-            "information to the OpenAI-compatible chat response."
+            "vLLM returned duplicate generation output indices while attaching "
+            "token information to the OpenAI-compatible chat response."
         )
 
-    missing_choice_indices = sorted(set(choice_indices) - outputs_by_index.keys())
-    if missing_choice_indices:
+    choices = list(getattr(response, "choices", []))
+    choice_indices = [choice.index for choice in choices]
+    if len(set(choice_indices)) != len(choice_indices):
+        raise RuntimeError(
+            "vLLM returned duplicate response choice indices while attaching "
+            "token information to the OpenAI-compatible chat response."
+        )
+
+    choice_index_set = set(choice_indices)
+    output_index_set = set(generation_output_indices)
+    missing_choice_indices = sorted(choice_index_set - output_index_set)
+    unexpected_output_indices = sorted(output_index_set - choice_index_set)
+    if missing_choice_indices or unexpected_output_indices:
         raise RuntimeError(
             "vLLM was asked to return token information for the "
             "OpenAI-compatible chat endpoint but response choices could not be "
             "matched to generation outputs: "
-            f"missing_choice_indices={missing_choice_indices}."
+            f"missing_choice_indices={missing_choice_indices}, "
+            f"unexpected_output_indices={unexpected_output_indices}."
         )
 
     for choice in choices:
@@ -364,35 +373,39 @@ def attach_token_information_to_chat_response_choices(
             )
         generation_token_ids = list(output_token_ids)
 
-        logprob_content = getattr(getattr(choice, "logprobs", None), "content", None)
-        if logprob_content is None:
+        generation_logprob_details = getattr(generation_details, "logprobs", None)
+        if generation_logprob_details is None:
             if generation_token_ids:
                 raise RuntimeError(
                     "vLLM was asked to return token information for the "
-                    "OpenAI-compatible chat endpoint but response "
+                    "OpenAI-compatible chat endpoint but generation output "
                     f"choice_idx={choice.index} did not include logprobs."
                 )
             generation_log_probs = []
         else:
-            if len(generation_token_ids) != len(logprob_content):
+            if len(generation_token_ids) != len(generation_logprob_details):
                 raise RuntimeError(
                     "vLLM returned mismatched generation token IDs and log "
                     "probabilities for the OpenAI-compatible chat endpoint: "
                     f"choice_idx={choice.index}, "
                     f"token_count={len(generation_token_ids)}, "
-                    f"logprob_count={len(logprob_content)}."
+                    f"logprob_count={len(generation_logprob_details)}."
                 )
             generation_log_probs = []
-            for token_id, logprob_entry in zip(generation_token_ids, logprob_content):
-                if logprob_entry.token != f"token_id:{token_id}":
+            for token_id, position_logprobs in zip(
+                generation_token_ids, generation_logprob_details
+            ):
+                selected_token_logprob = position_logprobs.get(token_id)
+                if selected_token_logprob is None:
                     raise RuntimeError(
-                        "vLLM returned a log probability for an unexpected "
-                        "token while attaching token information to the "
-                        "OpenAI-compatible chat response: "
-                        f"choice_idx={choice.index}, token_id={token_id}, "
-                        f"logprob_token={logprob_entry.token!r}."
+                        "vLLM generation log probabilities did not include the "
+                        "selected token while attaching token information to "
+                        "the OpenAI-compatible chat response: "
+                        f"choice_idx={choice.index}, token_id={token_id}."
                     )
-                generation_log_probs.append(float(logprob_entry.logprob))
+                generation_log_probs.append(
+                    max(float(selected_token_logprob.logprob), VLLM_LOGPROB_FLOOR)
+                )
 
         choice.message.prompt_token_ids = list(prompt_token_ids)
         choice.message.generation_token_ids = generation_token_ids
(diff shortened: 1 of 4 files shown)


nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -613,19 +614,27 @@ async def capture_result_generator():
...
                     return response
 
-                return attach_routed_experts_to_chat_response_choices(
-                    response,
-                    final_res,
-                    device=torch.device("cpu"),
-                    logger=LOGGER,
-                    routed_experts_dtype=worker_self.routed_experts_dtype,
-                )
+                if request.logprobs and request.return_tokens_as_token_ids:
```
jthomson04: **[BUG]** This gate ignores the server-level `return_tokens_as_token_ids` default, so the metadata is silently omitted for a request shape that Gym actually sends.

The server is constructed with `return_tokens_as_token_ids=True` (line 671 below), and vLLM resolves the per-request field against that default:

```python
# vllm v0.20.0, vllm/entrypoints/openai/chat_completion/serving.py:1716
should_return_as_token_id = (
    return_as_token_id if return_as_token_id is not None
    else self.return_tokens_as_token_ids
)
```

So a client sending `logprobs=True` **without** `return_tokens_as_token_ids` still gets `token_id:N` tokens rendered — but `request.return_tokens_as_token_ids` is `None`, this gate is falsy, and nothing gets attached. Gym then falls back to `/tokenize` and the perf win quietly disappears with no error anywhere.

This isn't hypothetical: Gym's non-capture path explicitly supports caller-supplied `logprobs=True` without the flag (`responses_api_models/vllm_model/app.py:306-310` at pin `473f446`).

Suggested fix — resolve the same way vLLM does (`self.return_tokens_as_token_ids` is available on the mixin via `OpenAIServing.__init__`):

```python
return_as_token_id = (
    request.return_tokens_as_token_ids
    if request.return_tokens_as_token_ids is not None
    else self.return_tokens_as_token_ids
)
if request.logprobs and return_as_token_id:
```

The explicit-`False` case is already handled correctly by the current code; only `None` is wrong.

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -312,17 +312,105 @@ def attach_routed_experts_to_chat_response_choices(
...
+        if output_token_ids is None:
+            raise RuntimeError(
+                "vLLM was asked to return token information for the "
+                "OpenAI-compatible chat endpoint but generation output "
+                f"choice_idx={choice.index} did not include token_ids."
+            )
+        generation_token_ids = list(output_token_ids)
+
+        logprob_content = getattr(getattr(choice, "logprobs", None), "content", None)
+        if logprob_content is None:
```
jthomson04: **[BUG]** This turns `logprobs=True` with an explicit `top_logprobs=null` into an HTTP 500 where it previously returned 200.

vLLM builds `choice.logprobs` only when `request.logprobs and request.top_logprobs is not None` (v0.20.0 `chat_completion/serving.py:743`, `:1315`), and derives sampling params as `logprobs=self.top_logprobs if self.logprobs else None` (`chat_completion/protocol.py:549`). So for `logprobs=true, top_logprobs=null, return_tokens_as_token_ids=true`, `choice.logprobs` is `None` while `output.token_ids` is non-empty → this raises → unhandled by the route's `except VLLMValidationError` → opaque 500.

Gym's capture path pins `top_logprobs=0`, so the shipped flow is safe, but this is a new hard failure for other callers.

Raising is arguably right (silently attaching `generation_log_probs=[]` alongside non-empty `generation_token_ids` would zero the loss mask downstream, which is worse) — but it should surface as an explicit 400 like the `VLLMValidationError` path in `create_chat_completion`, not a generic 500.

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -312,17 +312,105 @@ def attach_routed_experts_to_chat_response_choices(
...
+                raise RuntimeError(
+                    "vLLM returned mismatched generation token IDs and log "
+                    "probabilities for the OpenAI-compatible chat endpoint: "
+                    f"choice_idx={choice.index}, "
+                    f"token_count={len(generation_token_ids)}, "
+                    f"logprob_count={len(logprob_content)}."
+                )
+            generation_log_probs = []
+            for token_id, logprob_entry in zip(generation_token_ids, logprob_content):
+                if logprob_entry.token != f"token_id:{token_id}":
```
jthomson04: This alignment check couples correctness to a vLLM presentation-layer string format. If that prefix ever changes across a vLLM bump, every generation 500s rather than degrading.

`final_res.outputs[i].logprobs` carries the native `{token_id: Logprob}` mapping — sourcing `generation_log_probs` from there instead of parsing the OpenAI `token_id:` strings would drop the coupling entirely. Nothing else depends on the string form, since the token IDs already come from `output.token_ids` above.

(Note this wouldn't rescue the `top_logprobs=null` case — the engine computes no logprobs at all there — so that still needs to error; see the comment above.)

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -312,17 +312,105 @@ def attach_routed_experts_to_chat_response_choices(
...
+            "OpenAI-compatible chat endpoint but the final request output did "
+            "not include prompt_token_ids."
+        )
+
+    outputs_by_index = {
+        output.index: output for output in getattr(final_request_output, "outputs", [])
+    }
+    choices = list(getattr(response, "choices", []))
+    choice_indices = [choice.index for choice in choices]
+    if len(outputs_by_index) != len(
```
jthomson04: Nit: this recomputes `getattr(final_request_output, "outputs", [])` after `outputs_by_index` was already built from it, and the single error message says "duplicate choice indices" even when the duplication is on the engine-output side.

Also worth converging with the sibling: `attach_routed_experts_to_chat_response_choices` checks the same unmatched-choice invariant with `continue` + a post-loop set difference, while this one uses an upfront set difference. Same contract, two shapes.

tests/unit/models/generation/test_vllm_utils.py
```diff
@@ -551,6 +552,62 @@ def test_attach_routed_experts_to_chat_response_choices_raises_for_unmatched_cho
         )
 
 
+def test_attach_token_information_to_chat_response_choices():
```
jthomson04, line 562: **[TEST]** This covers only the happy path, but the new function has five raise branches and none are exercised. The routed-experts analog in this same file has `test_attach_routed_experts_to_chat_response_choices_raises_for_unmatched_choice`.

Worth adding at least two negative cases:
- token/logprob count mismatch (`utils.py:370`)
- unexpected `logprob_entry.token` format (`utils.py:381`) — this one is the version-fragility tripwire, so a test pinning it would fail loudly on a vLLM bump instead of failing in production.

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -1624,9 +1641,13 @@ def _standardize(d: dict) -> dict:
         message = d["choices"][0]["message"]
         for key in ("reasoning", "reasoning_content"):
             message.pop(key, None)
+        message.pop("generation_log_probs", None)
 
         return d
 
+    assert actual_result["choices"][0]["message"]["generation_log_probs"] == [
```
jthomson04, line 1794: **[TEST]** Good assertion here. The higher-value place to pin the new field, though, is the `required_prefix_token_ids` test further down in this file (around line 1987) — it posts a chat completion with `required_prefix_token_ids` but currently only reads `logprobs.content[0]`.

That's precisely where engine-native `prompt_token_ids` diverges from naive template tokenization: the spliced prompt. One line pins the behavior that motivates the whole PR:

```python
assert vllm_http_server_result["choices"][0]["message"]["prompt_token_ids"] == initial_tokenized_query_ids
```

jthomson04 (commented):
Nice change — the restructuring that decouples routed-experts attachment from token attachment is a clear improvement, and `attach_token_information_to_chat_response_choices` mirrors its routed-experts analog closely.

One blocking-ish issue (the gate on `return_tokens_as_token_ids` makes the feature a silent no-op for a request shape Gym itself supports), plus a new 500 path, some missing negative-test coverage, and a nit. Details inline.

One cross-repo consequence worth calling out here since it isn't visible in this diff: at the pinned Gym commit `473f446`, `choice_dict.pop("logprobs")` lives *inside* `if self.config.return_token_id_information and "prompt_token_ids" not in choice_dict["message"]:` (`responses_api_models/vllm_model/app.py:446`, `:499`). Now that the server always supplies `prompt_token_ids`, that block is skipped entirely, so the duplicated per-token `logprobs` payload (token string + logprob + `bytes` + `top_logprobs`) stays on every returned message and flows into logged trajectories — the data the comment there calls "duplicated information". It doesn't break validation (`NeMoGymChoice` inherits `Choice`, which declares `logprobs`), but it partially offsets the payload win. Worth a companion Gym change, or at least a note in the PR description.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Addressed in c3129078b1e0e5fb04d5c2eda20da281fb5e3d3a. The gate now resolves the per-request setting against the serving default, matching vLLM. The HTTP test omits the request field and verifies native metadata is still returned; explicit false remains unchanged.

nemo_rl/models/generation/vllm/utils.py
jthomson04: Addressed in c3129078b1e0e5fb04d5c2eda20da281fb5e3d3a. The full-generator path now raises VLLMValidationError with parameter=top_logprobs before consuming results. The HTTP test posts an explicit null and asserts a 400 with an actionable error.

nemo_rl/models/generation/vllm/utils.py
jthomson04: Addressed in c3129078b1e0e5fb04d5c2eda20da281fb5e3d3a. generation_log_probs now comes from the native CompletionOutput.logprobs selected-token mapping and applies the same -9999 floor as vLLM. The happy-path unit test deliberately supplies unrelated presentation strings to prove they are ignored.

nemo_rl/models/generation/vllm/utils.py
jthomson04: Addressed in c3129078b1e0e5fb04d5c2eda20da281fb5e3d3a. Final outputs are cached once; duplicate engine-output and duplicate response-choice indices now have separate diagnostics, and exact missing and unexpected index sets are checked. I kept the routed-experts sibling unchanged to avoid an unrelated behavior refactor.

yuki-97 (commented):
Overall LGTM on the code — self-review pass already resolved the substantive issues (server-default fallback for `return_tokens_as_token_ids`, HTTP 400 for `top_logprobs=null` instead of 500, native `CompletionOutput.logprobs` instead of the `token_id:` string parse, and comprehensive error-branch coverage on the new helper).

The 3 attached fields (`prompt_token_ids`, `generation_token_ids`, `generation_log_probs`) map to the existing Gym consumer contract at [`responses_converter.py:166`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/nemo_gym/responses_converter.py#L166) and [`:431`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/nemo_gym/responses_converter.py#L431). Until now those fields were synthesized upstream by Gym's `vllm_model` server ([`app.py:462-496`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/vllm_model/app.py#L462-L496)) via a per-turn `/tokenize` round-trip + `token_id:` prefix parsing on `choice.logprobs.content`, guarded by a `START TODO / END TODO` block whose exit condition is exactly "NeMo-RL vLLM natively supports prompt token ids". This PR is that upgrade — the producer now attaches natively, enabling a follow-up Gym change to delete the TODO block.

**[BLOCKER]** perf gain requires the paired Gym change. The currently pinned Gym (`473f446`) [unconditionally calls `/tokenize`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/vllm_model/app.py#L481) and `.update()`-overwrites the natively-attached fields on `message_dict`, so this PR by itself is perf-neutral (negligible extra serialization — see breakdown below). The 14.6% / 392.1s / "zero `/tokenize` calls" numbers land only after `app.py:462-496` is deleted on the Gym side and the submodule pin is bumped — currently the PR description reads as if this PR is standalone-shippable for perf.

**Payload size breakdown** (per response, ~1000-token prompt + ~1000-token generation, for context on the "extra serialization" above):

| Component | Size | Share of total |
|---|---|---|
| Plain text response (no logprobs, no metadata) | ~2 KB | 0.1% |
| + native `choice.logprobs.content` (1000 positions × ~50 B) | +50 KB | 2.4% |
| + `prompt_token_ids` (1000 int, this PR) | +5 KB | 0.2% |
| + `generation_token_ids` (1000 int, this PR) | +5 KB | 0.2% |
| + `generation_log_probs` (1000 float, this PR) | +10-15 KB | 0.6% |
| + `routed_experts` for Qwen3-30BA3B (48 MoE layers × topk=8 × int16, base64 envelope) | +2000 KB | 96.5% |
| Total | ~2.1 MB | 100% |

The 3 new fields sum to ~20 KB — <1% of a Qwen3-30BA3B response and ~40% growth on top of a non-MoE baseline — both dwarfed by the per-turn `/tokenize` round-trip latency they replace.

Two nits attached separately.

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -318,17 +319,117 @@ def attach_routed_experts_to_chat_response_choices(
...
+                )
+
+        choice.message.prompt_token_ids = list(prompt_token_ids)
+        choice.message.generation_token_ids = generation_token_ids
+        choice.message.generation_log_probs = generation_log_probs
+
+    return response
+
+
 def model_dump_chat_response_with_routed_experts(response: Any) -> dict[str, Any]:
```
yuki-97: `model_dump_chat_response_with_routed_experts` now preserves four dynamic message fields (`routed_experts`, `prompt_token_ids`, `generation_token_ids`, `generation_log_probs`); the name still advertises only routed experts.

The updated docstring already says "dynamic message fields", which is where the truth lives — but future readers grepping `routed_experts` or adding another dynamic field will hit the name-vs-behavior gap. Since scope broadened here, renaming is cheapest now (2 call sites: `vllm_worker_async.py:737` and `test_vllm_utils.py`).

**Suggest fix:** rename to something like `model_dump_chat_response_with_dynamic_message_fields`.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -609,19 +626,27 @@ async def capture_result_generator():
...
-                    logger=LOGGER,
-                    routed_experts_dtype=worker_self.routed_experts_dtype,
-                )
+                if request.logprobs and return_as_token_id:
+                    response = attach_token_information_to_chat_response_choices(
+                        response,
+                        final_res,
+                    )
+
+                if worker_self._return_routed_experts_enabled():
```
yuki-97, line 640: **[TEST]** After this PR the two attach helpers can both run on the same response (routed-experts recipes: `router_replay.enabled=True` → `enable_return_routed_experts=True` → C=T, plus the new B=T from `logprobs + return_tokens_as_token_ids`), but no test pins the combined shape.

`test_attach_token_information_to_chat_response_choices` sets only token fields; `test_attach_routed_experts_to_chat_response_choices_*` set only `routed_experts`; `test_model_dump_chat_response_with_routed_experts_preserves_dynamic_field` also sets only `routed_experts` even though the dump function was extended to 4 fields. Nightly MoE recipes exercise this compose path but assert convergence, not attach-field shape.

**Suggest fix:** extend the model_dump test (or add a sibling) to seed both `routed_experts` AND the three token fields on `choice.message` via both helpers, then assert all 4 survive `model_dump_chat_response_with_routed_experts`.

yuki-97:
hi @jthomson04 , nice feature!

> [BLOCKER] perf gain requires the paired Gym change. The currently pinned Gym (473f446) [unconditionally calls /tokenize](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/vllm_model/app.py#L481) and .update()-overwrites the natively-attached fields on message_dict, so this PR by itself is perf-neutral (mildly negative from the extra serialization). The 14.6% / 392.1s / "zero /tokenize calls" numbers land only after app.py:462-496 is deleted on the Gym side and the submodule pin is bumped — currently the PR description reads as if this PR is standalone-shippable for perf.

for this one, want to confirm are you using a different version nemo-gym, or am I missing something?

jthomson04:
Thanks for checking this. The benchmark did use Gym `eddd5e98`, but a paired Gym change is not required for the optimization.

At the pinned Gym revision `473f446`, the `/tokenize` compatibility path is conditional:

```python
if self.config.return_token_id_information and "prompt_token_ids" not in choice_dict["message"]:
```

Both [`client.create_tokenize(...)` and the subsequent `message_dict.update(...)`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/vllm_model/app.py#L446-L499) are inside that block. This PR preserves `prompt_token_ids` directly on `choice.message` during serialization, so the condition is false and pinned Gym skips the entire fallback; it does not overwrite the native fields.

The standard `choice.logprobs` payload remains because that cleanup is also inside the skipped compatibility block. That duplication is intentional here to preserve the OpenAI response contract and is now called out separately in the PR description.

I updated the description to disclose the benchmark Gym revision and make the pinned-Gym behavior explicit. No Gym PR or submodule bump is needed.

nemo_rl/models/generation/vllm/utils.py
jthomson04: Addressed in 9d9676f81d12f6fb1977429b7f0ede4a108e8d5c. Renamed the helper to model_dump_chat_response_with_dynamic_message_fields and updated its import, HTTP call site, and focused tests. Repository search confirms the old name has no remaining consumers.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Addressed in 9d9676f81d12f6fb1977429b7f0ede4a108e8d5c. The serializer test now applies both token-information and routed-expert attachment to the same response, then asserts that routed_experts, prompt_token_ids, generation_token_ids, and generation_log_probs all survive serialization.

jthomson04:
/ok to test [60c006d](https://github.com/NVIDIA-NeMo/RL/commit/60c006d757e2ab99f5b277f8d9428c1911088747)

yuki-97 (approved):
LGTM, thanks for the improvement and the clear clarify!

PR #3400, opened 2026-07-29
feat(grpo): add val_start_at to delay periodic validation

michal2409:
Adds `grpo.val_start_at`: the first training step eligible for periodic validation. It mirrors `val_period`'s required-int convention: `-1` (the exemplar default) disables the delay and keeps today's behavior exactly. When set, periodic validation (`val_period`) only fires once `step >= val_start_at`; `val_at_start` and `val_at_end` are unaffected.

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -253,8 +253,8 @@ class GRPOConfig(TypedDict):
     advantage_clip_high: NotRequired[float | None]
     use_leave_one_out_baseline: bool
     val_period: int
-    # First training step eligible for periodic validation; absent means no delay.
-    val_start_at: NotRequired[int]
+    # First training step eligible for periodic validation; -1 disables the delay.
+    val_start_at: int
     val_batch_size: int | None  # None for NeMo-Gym compatibility
     val_at_start: bool
     # Whether to run validation on the last training step. Setting this to True ensures the
@@ -2588,7 +2588,7 @@ def grpo_train(
     val_at_start = master_config.grpo["val_at_start"]
     val_at_end = master_config.grpo["val_at_end"]
     val_period = master_config.grpo["val_period"]
-    val_start_at = master_config.grpo.get("val_start_at", None)
+    val_start_at = master_config.grpo["val_start_at"]
     colocated_inference = master_config.policy["generation"]["colocated"]["enabled"]
     refit_buffer_size_gb = master_config.policy.get("refit_buffer_size_gb")
 
@@ -3149,7 +3149,7 @@ def grpo_train(
                 # Run validation if it's a validation step or last step with val_at_end
                 if (
                     val_period > 0
-                    and (val_start_at is None or total_steps + 1 >= val_start_at)
+                    and (total_steps + 1) >= val_start_at
                     and (total_steps + 1) % val_period == 0
                 ) or (val_at_end and is_last_step):
                     memory_tracker.snapshot_start_of_stage("Validation", dir())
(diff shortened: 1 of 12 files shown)


tests/unit/algorithms/test_grpo.py
```diff
@@ -2445,6 +2445,80 @@ def test_grpo_train_skips_prev_logprobs_when_force_on_policy_ratio(
     )
 
 
+@pytest.mark.parametrize("train_func", [grpo_train, async_grpo_train])
```
yuki-97: `grpo_train_sync` in `grpo_sync.py` received the same three-line guard (L446, L996) but isn't parametrized here.

nemo_rl/algorithms/grpo.py
```diff
@@ -253,6 +253,8 @@ class GRPOConfig(TypedDict):
     advantage_clip_high: NotRequired[float | None]
     use_leave_one_out_baseline: bool
     val_period: int
+    # First training step eligible for periodic validation; absent means no delay.
+    val_start_at: NotRequired[int]
```
yuki-97: let's mirror `val_period`'s convention: required `int` with `-1` as the disable sentinel.

`val_period: int` on L255 is gated by `if val_period > 0` and set to `-1` in `grpo_smoke.yaml`. The new knob has the same "disable = no delay" semantic, so the `NotRequired[int]` + `.get(..., None)` + `is None or ...` cascade at L2591/L3152/L3858/L4558 (also grpo_sync.py L446/L996) is unnecessary.

**Suggest fix:** declare `val_start_at: int`, add `val_start_at: -1` to `grpo_math_1B.yaml`, direct-read at the call sites, and use `(total_steps + 1) >= val_start_at` in the guards.
also need to change for the other two places in this PR.

```suggestion
    val_start_at: int
```

nemo_rl/algorithms/grpo.py
```diff
@@ -2586,6 +2588,7 @@ def grpo_train(
     val_at_start = master_config.grpo["val_at_start"]
     val_at_end = master_config.grpo["val_at_end"]
     val_period = master_config.grpo["val_period"]
+    val_start_at = master_config.grpo.get("val_start_at", None)
```
yuki-97: we encourage to not use default value in codes and read from yaml.
FYI we are migrating to use BaseModel and put the default value, but now let's still follow the previous style.
```suggestion
    val_start_at = master_config.grpo["val_start_at"]
```

yuki-97 (commented):
thanks @michal2409 , overall LGTM and left some minor comments.

tests/unit/algorithms/test_grpo.py
michal2409: Added in 0a4e31ae — the test is now parametrized over all three trainers; `grpo_train_sync` runs through a new `mock_sync_grpo_infrastructure` helper (mirroring the async one)

nemo_rl/algorithms/grpo.py
michal2409: Done in 0a4e31ae — `val_start_at: int` required with `-1` sentinel, guards are now `(step + 1) >= val_start_at`, and `val_start_at: -1` is recorded in `grpo_math_1B.yaml` plus the other standalone GRPO-family roots (`examples/nemo_gym/grpo_nanov3.yaml`, `grpo_workplace_assistant_nemotron_nano_v2_9b.yaml`, and the four `nemotron-3-super/stage*.yaml`)

nemo_rl/algorithms/grpo.py
michal2409: Done in 0a4e31ae — direct read `master_config.grpo["val_start_at"]` in all three trainers, no call-site default.

yuki-97 (dismissed):
LGTM, thanks @michal2409

michal2409:
The L1 failure was `research/template_project/single_update.py` — it validates the full `MasterConfig` and its config copy was missing the new required key. Added `val_start_at` there in 127db4d0.

yuki-97 (approved):

Issue #3336, opened 2026-07-24
[Dataplane Sync RL] cleanup tracking issue


terrykong:
Tracking issue for loose ends in the Single Controller **sync dataplane** implementation (`grpo_sync.py`), companion to the async/SC tracking issue #2625.
Items will be checked off or split into separate issues as they land.

### Known missing items

[...]

PR #3401, opened 2026-07-29
feat(grpo): validation-only sampling params and grouped pass@k validation

michal2409:
Right now, validation measures the model exactly the way training explores it: one rollout per prompt, using the same sampling settings as training. For agentic RL that makes the validation accuracy a noisy number — a single sample at training temperature can pass or fail a prompt by luck.

This PR lets validation measure differently from how training explores:

- `policy.generation.val_temperature` / `val_top_p` / `val_top_k` — validation-only sampling. The exemplar YAMLs default them to the train values via interpolation (`val_temperature: ${.temperature}`, ...), so validation samples exactly like training unless overridden (typically near-greedy).
- `grpo.val_num_generations_per_prompt` — validation runs each prompt k times. `accuracy` stays the mean reward over all rollouts; when k > 1 a separate `pass_k` metric reports the share of prompts with at least one passing attempt.

How it works:

- `validate()` builds an explicit `SamplingParams` dataclass from the val keys and passes it through the NeMo-Gym rollout entry points; training rollouts are untouched.
- The vLLM async OpenAI server accepts exactly two sampling profiles: train and val (the on-policy guard for training is unchanged).
- `setup()` rejects val sampling that differs from train sampling unless the backend is vLLM and the NeMo-Gym rollout path is used.

Together these turn validation accuracy into a stable, repeatable measurement — which matters whenever it drives a decision, such as accuracy-threshold stopping.


nemo_rl/algorithms/grpo.py
```diff
@@ -259,6 +259,14 @@ class GRPOConfig(TypedDict):
     # final checkpoint has validation metrics, which is required for get_best_checkpoint_path().
     val_at_end: bool
     max_val_samples: int | None  # None for NeMo-Gym compatibility
+    # Validation-only sampling overrides for NeMo-Gym rollouts. Absent means
+    # validation uses policy.generation sampling; when only one is set, the
+    # other falls back to its policy.generation value.
+    val_temperature: NotRequired[float]
```
yuki-97: val_temperature and val_top_p are generation params, not eval-loop knobs — wdyt moving them into `GenerationConfig` in `nemo_rl/models/generation/interfaces.py`? num_val_generations_per_prompt fits GRPOConfig since it drives the eval loop.

Accessors then become `master_config.policy["generation"]["val_temperature"]` / `["val_top_p"]` and `master_config.grpo["num_val_generations_per_prompt"]`.

nemo_rl/algorithms/grpo.py
```diff
@@ -389,6 +397,20 @@ def setup(
     if generation_config["backend"] == "vllm":
         normalize_vllm_refit_config(cast(VllmConfig, generation_config))
 
+    val_temperature = grpo_config.get("val_temperature", None)
```
yuki-97: Default val_temperature / val_top_p in the exemplar YAML via OmegaConf interpolation (`val_temperature: ${.temperature}`) so they inherit from temperature / top_p and are always non-None at runtime, and give num_val_generations_per_prompt a YAML default of `1`.

Then switch to direct subscription (`master_config.grpo["num_val_generations_per_prompt"]`, etc.) and drop every `is None` gate: this whole setup() block (400–411) collapses, and the validate() gate + dict-copy at 3619–3632 does too.

nemo_rl/models/generation/interfaces.py
```diff
@@ -207,6 +207,9 @@ class GenerationConfig(TypedDict):
     _pad_token_id: NotRequired[int]
     # MTP draft weights arrive via refit if the trainer trains the MTP layer.
     _mtp_weights_from_refit: NotRequired[bool]
+    # Internal: resolved validation sampling profile (grpo.val_temperature /
+    # val_top_p), published by grpo.setup() for OpenAI-compatible servers.
+    _validation_generation: NotRequired[dict[str, float] | None]
```
yuki-97: Add a `SamplingParams` dataclass (`temperature`, `top_p`) in `nemo_rl/models/generation/interfaces.py`. 

`run_nemo_gym_rollout_sync` takes an explicit `sampling_param: SamplingParam` arg — validate() constructs one from val_temperature/val_top_p and passes it straight through; the vLLM worker unpacks it into vllm's `SamplingParams` at the boundary.

The `_validation_generation` side-channel then goes away entirely — no field on GenerationConfig (this line), no setup() stash at grpo.py:402-411, no `allowed_sampling` whitelist in vllm_worker_async.py.

nemo_rl/algorithms/grpo.py
yuki-97: currently val_xxx is only supported for vLLM + gym, let's add an assert when the train sampling param is different from the val sampling param when using other rollout backend or not using gym at `setup()` to guard.

nemo_rl/algorithms/grpo.py
```diff
@@ -3652,11 +3697,26 @@ def validate(
...
+                    "Validation rewards must be divisible by "
+                    "grpo.num_val_generations_per_prompt"
+                )
+                accuracy = (
+                    (rewards_t.view(-1, num_val_generations_per_prompt) > 0)
+                    .any(dim=1)
+                    .float()
+                    .mean()
+                    .item()
+                )
```
yuki-97: it's good to add `pass@k`, but here still using the name `accuracy` which is not correct.

could you still keep `accuracy` as before, and add a new `pass_k` metric?

nemo_rl/algorithms/grpo.py
```diff
@@ -259,6 +259,14 @@ class GRPOConfig(TypedDict):
     # final checkpoint has validation metrics, which is required for get_best_checkpoint_path().
     val_at_end: bool
     max_val_samples: int | None  # None for NeMo-Gym compatibility
+    # Validation-only sampling overrides for NeMo-Gym rollouts. Absent means
+    # validation uses policy.generation sampling; when only one is set, the
+    # other falls back to its policy.generation value.
+    val_temperature: NotRequired[float]
+    val_top_p: NotRequired[float]
```
yuki-97: let's also add `val_top_k` to keep the same params list with train side.

nemo_rl/algorithms/grpo.py
```diff
@@ -3588,16 +3616,33 @@ def validate(
             master_config.grpo["max_val_samples"]
             // master_config.grpo["val_batch_size"]
         )
+        val_temperature = master_config.grpo.get("val_temperature", None)
+        val_top_p = master_config.grpo.get("val_top_p", None)
+        assert (val_temperature is None and val_top_p is None) or _should_use_nemo_gym(
```
yuki-97: we can remove it here after https://github.com/NVIDIA-NeMo/RL/pull/3401/changes#r3676037582.

tests/unit/algorithms/test_grpo.py
```diff
@@ -3173,6 +3173,67 @@ def test_validate_works_without_logger(self, mock_grpo_components):
         assert "accuracy" in val_metrics
         assert "avg_length" in val_metrics
 
+    def test_grouped_validation_uses_pass_at_k_as_accuracy(self, mock_grpo_components):
```
yuki-97: can we also add an unit test for different train/val sampling param and num_generations_per_prompt?

nemo_rl/algorithms/grpo.py
```diff
@@ -259,6 +259,14 @@ class GRPOConfig(TypedDict):
...
     val_at_end: bool
     max_val_samples: int | None  # None for NeMo-Gym compatibility
+    # Validation-only sampling overrides for NeMo-Gym rollouts. Absent means
+    # validation uses policy.generation sampling; when only one is set, the
+    # other falls back to its policy.generation value.
+    val_temperature: NotRequired[float]
+    val_top_p: NotRequired[float]
+    # Number of independent validation rollouts generated for each prompt;
+    # accuracy is then reported as pass@k over each prompt's k rollouts.
+    num_val_generations_per_prompt: NotRequired[int]
```
yuki-97: nit
```suggestion
    val_num_generations_per_prompt: NotRequired[int]
```

yuki-97 (commented):
@michal2409 thanks for adding this valuable feature! left some comments.

nemo_rl/algorithms/grpo.py
michal2409: Done in e290c18f. They live in `GenerationConfig` now; `val_num_generations_per_prompt` stays in `GRPOConfig`.

nemo_rl/algorithms/grpo.py
michal2409: Done in e290c18f. The exemplar YAMLs default them via `${.temperature}` etc., the code reads them directly, and both blocks are gone.

nemo_rl/models/generation/interfaces.py
michal2409: Done in e290c18f. `validate()` builds a `SamplingParams` and passes it through the gym rollout entry points; `_validation_generation` is gone.

nemo_rl/algorithms/grpo.py
michal2409: Done in e290c18f. `setup()` now asserts val sampling equals train sampling unless the backend is vLLM and the NeMo-Gym path is used.

nemo_rl/algorithms/grpo.py
michal2409: Done in e290c18f. `accuracy` is the mean reward again; `pass_k` is a separate metric reported when k > 1.

nemo_rl/algorithms/grpo.py
michal2409: Done in e290c18f.

nemo_rl/algorithms/grpo.py
michal2409: Done in e290c18f.

tests/unit/algorithms/test_grpo.py
michal2409: Done in e290c18f. Added a test that validation passes the val sampling params to the gym rollout with k=2 (checking both `accuracy` and `pass_k`), and one for the new `setup()` guard.

nemo_rl/algorithms/grpo.py
michal2409: Done in e290c18f. Renamed; it is a required key now since the exemplar YAMLs default it to 1.

nemo_rl/algorithms/grpo.py
```diff
@@ -3579,6 +3599,12 @@ def validate(
     timer = Timer(context={"worker": "validator"})
     with timer.time("total_validation_time"):
         print(f"▶ Starting validation at step {step}...", flush=True)
+        val_num_generations_per_prompt = master_config.grpo[
+            "val_num_generations_per_prompt"
+        ]
+        assert val_num_generations_per_prompt >= 1, (
```
yuki-97: can we move this to the setup part? so that we can get the error at beginning.

yuki-97 (commented):
thanks @michal2409 for the update, just one nit.

besides, maybe like the CI fail in https://github.com/NVIDIA-NeMo/RL/pull/3404, will need to also update the configs under `examples/nemo_gym/nemotron-3-ultra`, main added some new recipes recently.

yuki-97:
@terrykong could you take a review as well?

nemo_rl/algorithms/grpo.py
michal2409: Done in 38d141b3. The assert now runs in setup() next to the other val-sampling checks.

nemo_rl/algorithms/grpo.py
```diff
@@ -398,6 +403,24 @@ def setup(
     if generation_config["backend"] == "vllm":
         normalize_vllm_refit_config(cast(VllmConfig, generation_config))
 
+    # Validation-only sampling is honored only on the NeMo-Gym vLLM rollout
+    # path; everywhere else validation must sample exactly like training.
+    val_sampling_overridden = (
+        generation_config["val_temperature"] != generation_config["temperature"]
```
terrykong, line 411: [`nemo_rl/algorithms/grpo.py:409`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L409)

**1 action item.**

### AI-1 — 7 nemotron-3-ultra recipes now `KeyError` at startup (PR-introduced)

This is a bare subscript, but the 7 configs under `examples/nemo_gym/nemotron-3-ultra/` are standalone (no `defaults:` key) and their `policy.generation` block was not given the new `val_*` keys — the PR renamed their `grpo.num_val_generations_per_prompt` but left the generation block untouched. This is the gap [@yuki-97 flagged on 7/31](https://github.com/NVIDIA-NeMo/RL/pull/3401#pullrequestreview-PRR_kwDOOJjv8s8AAAABHw8Qsg).

Walk it with `mopd.yaml`:
1. [`mopd.yaml` has no `defaults:` key](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/examples/nemo_gym/nemotron-3-ultra/mopd.yaml#L1), so [`load_config_with_inheritance`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/utils/config.py#L73) never merges in the exemplar.
2. Its generation block ends at [`top_k: null` (L328)](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/examples/nemo_gym/nemotron-3-ultra/mopd.yaml#L328) — no `val_temperature`.
3. `setup()` reaches this line and raises `KeyError: 'val_temperature'`.
4. It fires even though [`mopd.yaml` sets `val_period: -1`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/examples/nemo_gym/nemotron-3-ultra/mopd.yaml#L73) — the guard is unconditional, so recipes that never validate still crash at startup.

Affected: `ifbench_teacher`, `mopd`, `reasoning_teacher`, `rlhf_teacher`, `student_rlvr1`, `student_rlvr2`, `swe_teacher`. [`research/template_project/configs/grpo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/research/template_project/configs/grpo_math_1B.yaml#L285) has the same omission (it got the `grpo` key but not the generation keys); it doesn't call `grpo.setup()` so it won't crash, but it's inconsistent.

**Action:** add the three lines to each file's `policy.generation` block (after its `top_k`), and to the research template config:

```yaml
    val_temperature: ${.temperature}
    val_top_p: ${.top_p}
    val_top_k: ${.top_k}
```

nemo_rl/models/generation/interfaces.py
```diff
@@ -201,6 +202,13 @@ class GenerationConfig(TypedDict):
     temperature: float
     top_p: float
     top_k: int | None
+    # Validation-only sampling. The exemplar YAMLs default these to the train
+    # values above via interpolation (${.temperature}, ...), so validation
+    # samples exactly like training unless overridden. Only honored on the
+    # NeMo-Gym vLLM rollout path (guarded in grpo.setup()).
+    val_temperature: NotRequired[float]
```
terrykong: [`nemo_rl/models/generation/interfaces.py:209`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/generation/interfaces.py#L209)

**1 action item.**

### AI-1 — pick one access contract for the `val_*` keys

They're declared `NotRequired` here, but [`grpo.setup()` bare-subscripts them](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L409) while [`vllm_worker_async` reads them with `.get()`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/generation/vllm/vllm_worker_async.py#L691). The `config-conventions` skill says `NotRequired` fields are accessed with `in` / `.get(key)`, and required fields are read bare with the default living in the exemplar YAML. Right now it's neither consistently — and that mismatch is what turns the missing-key case into the `KeyError` in my other comment.

Since every exemplar YAML defines these, I'd make them required so the fail-loud is intentional and both call sites read the same way:

```suggestion
    val_temperature: float
```

(and correspondingly `val_top_p: float`, `val_top_k: int | None` below, with `vllm_worker_async` switching to plain subscripts). The alternative — keep `NotRequired`, use `.get()` in `setup()`, and let `None` mean "inherit train" — is equally fine; just not the mix.

nemo_rl/models/generation/interfaces.py
```diff
@@ -214,6 +222,31 @@ class GenerationConfig(TypedDict):
     _mtp_weights_from_refit: NotRequired[bool]
 
 
+@dataclass
+class SamplingParams:
```
terrykong: [`nemo_rl/models/generation/interfaces.py:226`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/generation/interfaces.py#L226)

**1 action item.**

### AI-1 — name collides with two existing `*SamplingParams`

`grpo.py` now imports two identically-shaped dataclasses: [`TrainingSamplingParams(top_k, top_p, temperature)`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/logits_sampling_utils.py#L30) at [grpo.py:41](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L41), and this new `SamplingParams(temperature, top_p, top_k)` at [grpo.py:101](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L101). Same three fields, same "sampling params" name, different meaning (train-time logit filtering vs generation-side rollout profile). The bare name also shadows vLLM's own `SamplingParams`.

**Action:** rename to `GenerationSamplingParams` for symmetry with its sibling. Free to reshape now, breaking later.

```suggestion
class GenerationSamplingParams:
```

nemo_rl/algorithms/grpo.py
```diff
@@ -398,6 +403,24 @@ def setup(
...
+        or generation_config["val_top_p"] != generation_config["top_p"]
+        or generation_config["val_top_k"] != generation_config["top_k"]
```
terrykong, line 413: [`nemo_rl/algorithms/grpo.py:411`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L411)

**1 action item.**

### AI-1 — `val_top_k` can never be honored, and fails late

Setting `val_top_k` to anything non-null is guaranteed to fail on the only path this guard permits:
1. This line sees `val_top_k != top_k` and requires vLLM + NeMo-Gym — which passes.
2. At the first validation, [`run_async_nemo_gym_rollout` asserts `not sampling_params.top_k`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/experience/rollouts.py#L2153) → `"Top k is not supported in the sampling params in NeMo-Gym path!"`.
3. Even if it didn't, [`_prepare_nemo_gym_rows` only stamps `temperature` and `top_p`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/experience/rollouts.py#L2023) and [`vllm_worker_async` forces `request.top_k = -1`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/generation/vllm/vllm_worker_async.py#L675).

So the failure is real but arrives at the first val step instead of at startup — the same "get the error at the beginning" concern you already addressed for the other check.

**Action:** add `assert not generation_config["val_top_k"]` here alongside the existing guard (or drop `val_top_k` if it isn't going to be plumbed through).

examples/configs/grpo_math_1B.yaml
```diff
@@ -15,6 +15,8 @@ grpo:
   advantage_clip_low: null
   advantage_clip_high: null
   max_val_samples: 256
+  # Validation rollouts per prompt; k > 1 also reports the pass_k metric.
+  val_num_generations_per_prompt: 1
```
terrykong, line 20: [`examples/configs/grpo_math_1B.yaml:19`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/examples/configs/grpo_math_1B.yaml#L19)

**1 action item.**

### AI-1 — `k` silently changes what `max_val_samples` means

[`max_batches = max_val_samples // val_batch_size`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L3715) is computed on the *unrepeated* batches, and [each batch is then expanded k-fold](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L3722). So with `max_val_samples: 256` and `k: 4` you actually run 1024 rollouts, and `accuracy`'s denominator is 1024. That's a defensible definition (`max_val_samples` = prompts), but nothing says so, and the end-of-validation `Samples processed:` line will report 4x the configured budget.

**Action:** make it explicit in the comment here and on the [`max_val_samples` field](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L250):

```suggestion
  # Validation rollouts per prompt; k > 1 also reports the pass_k metric.
  # max_val_samples counts PROMPTS: total validation rollouts = max_val_samples * k.
  val_num_generations_per_prompt: 1
```

_(Uses the current spelling; if you take the rename revert from the `grpo.py:254` thread, spell it `num_val_generations_per_prompt` here too.)_

nemo_rl/algorithms/grpo.py
```diff
@@ -398,6 +403,24 @@ def setup(
...
+    )
+    if val_sampling_overridden:
+        assert generation_config["backend"] == "vllm" and _should_use_nemo_gym(
+            master_config
+        ), (
+            "generation.val_temperature/val_top_p/val_top_k differing from the "
+            "train sampling params is only supported for vLLM NeMo-Gym rollouts."
+        )
+    assert grpo_config.val_num_generations_per_prompt >= 1, (
```
terrykong, line 429: [`nemo_rl/algorithms/grpo.py:420`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L420)

**1 action item (minor).**

### AI-1 — `stop_at_validation_metric: pass_k` with `k = 1` fails hours in

`pass_k` is only added to `val_metrics` [when k > 1](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L3829). With `k = 1` the run does start, and only at the *first* validation does [`_validation_stop_value`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L2584) raise `grpo.stop_at_validation_metric='pass_k' is not a reported validation metric`. With a large `val_period` that's a long way in — and it's the same "get the error at the beginning" concern you already acted on for the sampling guard.

Since you're already validating the k range right here:

```suggestion
    assert grpo_config.val_num_generations_per_prompt >= 1, (
        "grpo.val_num_generations_per_prompt must be >= 1"
    )
    assert not (
        grpo_config.stop_at_validation_metric == "pass_k"
        and grpo_config.val_num_generations_per_prompt <= 1
    ), (
        "grpo.stop_at_validation_metric='pass_k' requires "
        "grpo.val_num_generations_per_prompt > 1"
    )
```

_(Uses the current spelling; if you take the rename revert from the `grpo.py:254` thread, spell it `num_val_generations_per_prompt` here too.)_

(Arguably the `>= 1` assert wants to be `Field(ge=1)` on the `GRPOConfig` field instead, which would make it enforced at load time for free.)

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -678,9 +678,27 @@ async def create_chat_completion(
...
+            # stability. Validation rollouts are the one exception: they are stamped with
+            # the validation sampling profile (generation.val_temperature / val_top_p),
+            # which is metric-only and safe to serve. Multi-turn agents issue their own
+            # requests, so this server-side check is the one chokepoint they all pass.
+            request_top_p = 1.0 if request.top_p is None else request.top_p
+            is_train_sampling = (
+                request.temperature == generation_config["temperature"]
+                and request_top_p == generation_config["top_p"]
+            )
+            is_val_sampling = request.temperature == generation_config.get(
```
terrykong: [`nemo_rl/models/generation/vllm/vllm_worker_async.py:691`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/generation/vllm/vllm_worker_async.py#L691)

**No action needed on this line** — recording the trade-off, since this weakens an invariant deliberately. (Separate, actionable issue in the same block: see my comment on L686 about the `top_p` fallback.)

The old guard's value was that server-side exactly *one* sampling profile was legal, so no request could go off-policy regardless of who issued it. Now two are legal and nothing tags which phase a request belongs to, so a training request stamped with the val profile would pass. Not reachable from NeMo-RL's own stamping — training always builds from [`SamplingParams.from_generation_config`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/generation/interfaces.py#L239) — and with the default `${.temperature}` interpolation the two profiles are identical, so the check is bit-for-bit as strict as before unless a user opts in. Reasonable trade given multi-turn agents issue their own requests; flagging only so it's a recorded decision rather than a silent relaxation.

One thing worth writing into the comment, since it's the fact the whole relaxation rests on: *only `grpo.validate()` ever constructs a non-train `SamplingParams`.* That's currently implicit, and it's what a future reader would need to know before adding a second caller.

nemo_rl/algorithms/grpo.py
```diff
@@ -3691,19 +3718,31 @@ def validate(
             if batch_idx >= max_batches:
                 break
 
+            if val_num_generations_per_prompt > 1:
+                val_batch = val_batch.repeat_interleave(val_num_generations_per_prompt)
```
terrykong, line 3742: [`nemo_rl/algorithms/grpo.py:3722`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo.py#L3722)

**No action needed in this PR** — FYI, filed as a sync-dataplane port item.

The TransferQueue trainer validates through its own [`validate_sync`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/grpo_sync.py#L236), which has no `repeat_interleave` and no `pass_k` — so with `data_plane.enabled: true` this line never runs and the grouped-validation key is silently a no-op. The *sampling* half of your PR is already fine there, since `examples/run_grpo.py` calls `grpo.setup()` before [dispatching to `grpo_train_sync`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/examples/run_grpo.py#L42), so the guard does fire.

That's a pre-existing feature-parity gap in `grpo_sync.py`, not something to fix here. Filed as the third checkbox in [#3336](https://github.com/NVIDIA-NeMo/RL/issues/3336) ("Port grouped validation ... to `validate_sync`"), which is the sync-dataplane cleanup tracking issue. The async path is unaffected — it calls the updated `validate()` directly.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -678,9 +678,27 @@ async def create_chat_completion(
...
             # The request sampling params need to exactly match those as are set in NeMo RL.
-            # If they do not match, the inference will be off policy and destroy training stability.
-            assert request.temperature == generation_config["temperature"]
-            assert request.top_p == generation_config["top_p"]
+            # If they do not match, the inference will be off policy and destroy training
+            # stability. Validation rollouts are the one exception: they are stamped with
+            # the validation sampling profile (generation.val_temperature / val_top_p),
+            # which is metric-only and safe to serve. Multi-turn agents issue their own
+            # requests, so this server-side check is the one chokepoint they all pass.
+            request_top_p = 1.0 if request.top_p is None else request.top_p
```
terrykong: [`nemo_rl/models/generation/vllm/vllm_worker_async.py:686`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/generation/vllm/vllm_worker_async.py#L686)

**1 action item.**

### AI-1 — this normalizes the *comparison*, not the request, so an unset `top_p` still reaches vLLM

The contrast with the line eight above is the whole thing:

```python
request.top_k = -1                                               # mutates the request
request_top_p = 1.0 if request.top_p is None else request.top_p  # new local; request untouched
```

`top_k` is genuinely pinned — the request vLLM receives carries `-1`. `request_top_p` is a fresh local that is only used in the two comparisons below it; `request.top_p` is never written, and [`request` is handed to vLLM unchanged](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/generation/vllm/vllm_worker_async.py#L704) still carrying `None`.

So the `1.0` never reaches the engine. vLLM resolves the unset `top_p` from the **model's** `generation_config.json`: [unset stays `None`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/openai/chat_completion/protocol.py#L219) → [`default_sampling_params.get("top_p", ...)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/openai/chat_completion/protocol.py#L603-L605) → [`model_config.get_diff_sampling_param()`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/openai/chat_completion/serving.py#L168) → [the model's file, since `generation_config` defaults to `"auto"`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/config/model.py#L1499-L1501). `Qwen/Qwen3-30B-A3B-Instruct-2507` ([used here](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml#L30)) ships `top_p: 0.8`; the gym recipes set `top_p: 1.0`.

**PR-introduced.** vLLM's behavior is old (default since v0.8.0), but [`main`'s bare `assert request.top_p == generation_config["top_p"]`](https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/models/generation/vllm/vllm_worker_async.py#L683) compared `None == 1.0`, failed, and rejected the request before it could reach vLLM. Coalescing first makes exactly that request pass — the assert is silenced, the sampling is not changed.

**Why it bites harder than a temperature mismatch:** top-p truncates and renormalizes, which changes the logprobs, so NeMo-RL mirrors it into training — [`TrainingSamplingParams` is built from `generation_cfg["top_p"]`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L244) and [the logits are rescaled](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L571) only when [`top_p != 1.0`](https://github.com/NVIDIA-NeMo/RL/blob/d2089ab05490f1d0ff9bdc228e511cef8743641d/nemo_rl/algorithms/logits_sampling_utils.py#L52). With config `top_p: 1.0`, training applies **no** filtering while generation sampled at `0.8` — logprobs computed on the full distribution for tokens drawn from a truncated one, with nothing surfacing it.

**Action:** require it explicitly. Pinning it the way `top_k` is pinned isn't available: this guard accepts *two* profiles and the server can't tell which one a request belongs to, so `request.top_p = <train>` would silently force train sampling onto validation requests (and vice versa). `top_k` escapes that because `-1` is legal under either profile.

```suggestion
            # NOTE: this normalizes the comparison only. vLLM resolves an unset top_p
            # from the model's generation_config.json (ModelConfig.generation_config
            # defaults to "auto"), NOT to 1.0, so a request that omits it would sample
            # off-policy while passing the check below.
            assert request.top_p is not None, (
                "top_p must be set explicitly on NeMo-RL requests; an unset top_p is "
                "resolved by vLLM from the model's generation_config.json and would "
                "bypass the on-policy sampling check."
            )
            request_top_p = request.top_p
```

### Follow-up — filed as #3497

The same mechanism applies to `temperature`, `top_k`, `min_p` and `repetition_penalty` engine-wide, and predates this PR. #3497 proposes passing `generation_config="vllm"` at engine construction so unset parameters resolve to vLLM's neutral defaults. Nothing to do here — that issue covers the engine-wide fix, this thread covers only the `top_p` guard regression.

terrykong (commented):
Reviewed the structure for the val sampling params, per @yuki-97's request. **Overall the shape looks right to me** — threading an explicit `SamplingParams` through the rollout call instead of the `_validation_generation` side-channel, guarding at `setup()` rather than whitelisting per-request, and leaving `accuracy` alone while adding `pass_k` beside it are all the right calls. The flat `val_` prefix also matches existing convention (`val_period`, `val_batch_size`, `max_val_samples`, `val_at_start`), so I wouldn't push for a nested block.

Two items I'd like resolved before merge, plus some smaller structural ones — all inline:

1. The 7 standalone `examples/nemo_gym/nemotron-3-ultra/*.yaml` configs never got the new `val_*` generation keys, so they `KeyError` at startup. The gap @yuki-97 flagged on 7/31 is only half-closed — the `grpo` block was updated, the `generation` block wasn't.
2. In `vllm_worker_async.py`, `request_top_p = 1.0 if request.top_p is None else request.top_p` normalizes the *comparison* but not the request, so a request omitting `top_p` now passes the on-policy guard and vLLM samples it from the model's `generation_config.json`. Previously that request was rejected.

Things I checked and found **fine**, so you don't need to re-verify them:
- `${.temperature}` interpolation resolves lazily *after* the `defaults:` deep-merge, so a child recipe overriding `temperature` gets the matching `val_temperature`.
- Rollout ordering is preserved on both paths (`_NemoGymStreamAccumulator` restores by `row_index`; `run_multi_turn_rollout` scatters back via `active_indices`), so `pass_k`'s `view(-1, k)` groups the right rollouts.
- `BatchedDataDict.repeat_interleave` deep-copies list fields, so the k copies don't alias one `message_log`.
- `mock_grpo_components` is function-scoped, so the new `setup()` test doesn't leak `backend: megatron` into its neighbours.

Filed #3497 separately for the engine-wide half of item 2 (vLLM resolving *any* unset sampling parameter from the model's `generation_config.json`) — that one predates this PR and isn't an ask here.

Reviewed at d2089ab05490f1d0ff9bdc228e511cef8743641d; re-checked against 160b5be1b46e58b313837091f9ca27f9cffb3a4b, which only touches `model_utils.py`, so every finding still applies unchanged. I couldn't run the test suite or linters locally (Linux-only lockfile), so nothing here is execution-verified.

_Generated by Claude Code_

nemo_rl/algorithms/grpo.py
michal2409: Done in 51cc9a6f — added the three val_* interpolation lines to all 7 nemotron-3-ultra configs and the research template.

nemo_rl/models/generation/interfaces.py
michal2409: Done in 51cc9a6f — took the required option: the three keys are now required in GenerationConfig and vllm_worker_async reads them with plain subscripts. Since pydantic enforces required TypedDict keys in every algorithm MasterConfig, the eval/mmau, distillation_math, and ppo_math_1B configs (plus their reference configs) also got the interpolation defaults.

nemo_rl/models/generation/interfaces.py
michal2409: Done in 51cc9a6f — renamed to GenerationSamplingParams.

nemo_rl/algorithms/grpo.py
michal2409: Done in 51cc9a6f — setup() now asserts val_top_k is unset when val sampling is overridden, so it fails at startup with a clear message instead of at the first validation step.

examples/configs/grpo_math_1B.yaml
michal2409: Done in 51cc9a6f — comment added here and on the GRPOConfig max_val_samples field.

nemo_rl/algorithms/grpo.py
michal2409: Done in 51cc9a6f — setup() asserts stop_at_validation_metric=pass_k requires val_num_generations_per_prompt > 1.

nemo_rl/models/generation/vllm/vllm_worker_async.py
michal2409: Done in 51cc9a6f — replaced the coalesce with an explicit assert request.top_p is not None (your suggested text), and the comment now records that grpo.validate() is the only caller constructing a non-train profile.

yuki-97 (approved):
thanks @michal2409 , LGTM.

one thing want to check that @terrykong do you think we should do https://github.com/NVIDIA-NeMo/RL/issues/3497 in this PR, or leave it as is now and do in a next PR?

nemo_rl/algorithms/grpo.py
terrykong: fixed in 51cc9a6fb970a2ba3c232d1999d52d993608c66e

nemo_rl/models/generation/interfaces.py
terrykong: fixed in 51cc9a6fb970a2ba3c232d1999d52d993608c66e — keys are required now and `vllm_worker_async` switched to plain subscripts, so both sites read the same way. Hand-built generation configs in tests were brought up to the wider contract in 739a675a49df9466c1f20c296f3997e306d032b1.

nemo_rl/models/generation/interfaces.py
terrykong: fixed in 51cc9a6fb970a2ba3c232d1999d52d993608c66e

nemo_rl/algorithms/grpo.py
terrykong: fixed in 51cc9a6fb970a2ba3c232d1999d52d993608c66e

examples/configs/grpo_math_1B.yaml
terrykong: fixed in 51cc9a6fb970a2ba3c232d1999d52d993608c66e

nemo_rl/algorithms/grpo.py
terrykong: fixed in 51cc9a6fb970a2ba3c232d1999d52d993608c66e

nemo_rl/models/generation/vllm/vllm_worker_async.py
terrykong: fixed in 51cc9a6fb970a2ba3c232d1999d52d993608c66e

nemo_rl/models/generation/vllm/vllm_worker_async.py
terrykong: fixed in 51cc9a6fb970a2ba3c232d1999d52d993608c66e

terrykong:
@yuki-97 @michal2409 i merged. #3497 can be tackled in a follow up. it's sort of a preexisting issue

PR #3163, opened 2026-07-11
feat(grpo): mask env-flagged samples from policy loss

macandro96:
# What does this PR do ?

Adds a per-sample gradient mask driven by the rollout environment. When `instance_config.mask_sample` is `True` on a rollout result, the sample's `loss_multiplier` is zeroed out before the policy update — but the sample is kept for advantage computation (baseline / std).

Cherry-pick commit [here](https://gitlab-master.nvidia.com/terryk/nemo-rl-internal/-/merge_requests/222)

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #3402, opened 2026-07-29
feat(grpo): make env-flagged sample masking optional

michal2409:
NeMo-Gym environments and agents can flag individual rollout samples as "don't train on this" (the mask_sample flag, added in #3163). Flagged samples are dropped from the policy-gradient loss while still counting toward advantage computation. Currently this behavior is always on: every gym rollout batch carries the flags, and any sample the environment flags is silently masked out.

This PR makes that behavior configurable via one optional boolean, `env.should_mask_flagged_samples` (documented in the gym exemplar YAMLs, next to `should_log_nemo_gym_responses`):

- Absent or true (default): behavior is unchanged — flags are carried in the batch and flagged samples are masked from the loss.
- false: the mask_sample key is never added to the rollout batch, so no env-driven masking happens. Downstream loss code already treats a missing key as a no-op, so nothing else needs to change.

The key is read in a single helper (`should_mask_flagged_samples` in `rollouts.py`) used by the sync trainer, the async trajectory collector, and `AsyncNemoGymRolloutImpl` (so the future single-controller flow honors the gate too — there, turning it off drops the flag from the result before it lands in `Completion.env_extras`).

Why you'd want to turn it off: the flags can be too coarse to honor — Gym flags rollouts that hit max iterations even when they solve the task, and those are samples worth training on (https://github.com/NVIDIA-NeMo/RL/pull/3201#issuecomment-4989862073). Also, how many samples get flagged is decided by the environment/agent and varies run to run, so masking changes the effective batch composition non-deterministically. For controlled experiments and benchmark runs we need the loss to see every sample; A/B runs showed the masked-sample ramp changing training dynamics mid-run. The gate makes that choice explicit while leaving the default exactly as it is.


nemo_rl/experience/rollouts.py
```diff
@@ -2437,11 +2444,14 @@ def _postprocess_single_nemo_gym_group(
...
-            # Agent/env-driven mask flag — True means this sample should be masked
-            # from the GRPO gradient (kept for advantage computation).
-            "mask_sample": _extract_mask_sample_flags(results),
         }
     )
+    # Agent/env-driven mask flag — True means this sample should be masked
+    # from the GRPO gradient (kept for advantage computation). Gated by
+    # grpo.mask_env_flagged_samples so runs can keep the key out of the
+    # batch entirely.
+    if mask_env_flagged_samples:
```
yuki-97, line 2468: `mask_env_flagged_samples` is only used for gym, let's move it from `grpo` to `env` and rename to `should_mask_flagged_samples` like `should_log_nemo_gym_responses` in `examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml`.

nemo_rl/algorithms/grpo.py
```diff
@@ -259,6 +259,10 @@ class GRPOConfig(TypedDict):
     # final checkpoint has validation metrics, which is required for get_best_checkpoint_path().
     val_at_end: bool
     max_val_samples: int | None  # None for NeMo-Gym compatibility
+    # Set false to keep env-driven mask_sample flags out of NeMo-Gym rollout
+    # batches, disabling env-driven loss masking; absent/true carries the
+    # flags and masks flagged samples from the loss.
+    mask_env_flagged_samples: NotRequired[bool]
```
yuki-97: could you add a brief description about when and why we need to set it to false here?

yuki-97 (commented):
overall LGTM, left some minor comments.

cc @ananthsub for this WAR for gym issue, details in https://github.com/NVIDIA-NeMo/RL/pull/3201#issuecomment-4989862073.

nemo_rl/algorithms/grpo.py
yuki-97: not a blocker for this PR, but wonder could you help to also add this to `AsyncNemoGymRolloutImpl` in `nemo_rl/experience/rollout_manager.py`? if not is also fine.

in case we forget to add it when you changing to use single controller flow later.

nemo_rl/experience/rollouts.py
michal2409: Done in 203c4a0a. The key is `env.should_mask_flagged_samples` now, read by one helper in `rollouts.py`.

nemo_rl/algorithms/grpo.py
michal2409: Done in 203c4a0a. The when/why is in the helper docstring and as a short comment on the key in the gym example YAMLs.

michal2409: Done in 203c4a0a. When the gate is off, `_result_to_completion` drops the `mask_sample` flag from the result, so later batch building never sees it.

michal2409: Added in 3a2b4b9b. The docstring now also gives the concrete case from #3201: Gym flags rollouts that hit max iterations even when they solve the task, and those are worth training on.

nemo_rl/experience/rollout_manager.py
```diff
@@ -413,6 +414,7 @@ def __init__(
         self._max_seq_len = max_seq_len
         self._max_rollout_turns = max_rollout_turns
         self._generation_config = generation_config
+        self._mask_env_flagged_samples = mask_env_flagged_samples
```
yuki-97, line 417: **[BUG]** The gate is unwired for the SC path — `RolloutManager.__init__` at L680 instantiates `rollout_cls(...)` without forwarding `mask_env_flagged_samples`, and doesn't accept it as a param either. `self._mask_env_flagged_samples` is always the default `True`, so `if not self._mask_env_flagged_samples` at L560 never fires and setting `env.should_mask_flagged_samples: false` in an SC recipe has no effect on this path.

**Suggest fix:** add `mask_env_flagged_samples: bool = True` to `RolloutManager.__init__`, forward it in the `rollout_cls(...)` call at L680, and compute + pass `should_mask_flagged_samples(master_config.env)` from `single_controller_utils/setup.py:435`. `AsyncRolloutImpl` already accepts `**kwargs`, so uniform forwarding is fine.

nemo_rl/experience/rollouts.py
```diff
@@ -1558,6 +1558,21 @@ def get_nemo_gym_thinking_tags(env_config: dict[str, Any]) -> list[str]:
...
+    True (the default): env-driven ``mask_sample`` flags are carried in the
+    rollout batch and flagged samples are dropped from the loss.
+
+    Set false when the flags are too coarse to honor: for example, Gym flags
+    rollouts that hit max iterations even when they solve the task, and those
+    are samples worth training on. It also keeps batch composition
+    deterministic for controlled benchmark runs — how many samples get
+    flagged varies run to run.
+    """
+    return env_config.get("should_mask_flagged_samples") is not False
```
yuki-97, line 1573: No unit coverage for `should_mask_flagged_samples` or either gate site (`_postprocess_single_nemo_gym_group` batched at `rollouts.py:2469`, `AsyncNemoGymRolloutImpl._result_to_completion` streaming at `rollout_manager.py:558`). The sibling `_extract_mask_sample_flags` is covered at `tests/unit/experience/test_reward_penalties.py:90`.

A similar-shape test for the helper (absent / `null` / `true` / `false` → expected bool) plus one assert per gate site (`final_batch` omits `mask_sample` when gate off; `env_extras["instance_config"]` has `mask_sample` popped when gate off) would close the gap — and would have caught the previous [BUG] since it's an end-to-end wiring test.

yuki-97 (commented):
@michal2409 thanks for the update!

the new added one in `nemo_rl/experience/rollout_manager.py` seems not pass the config correctly, and could you help to add some unit tests to guard? flagged inline.

nemo_rl/experience/rollout_manager.py
michal2409: Done in 20f1f6e5. `RolloutManager` now takes `mask_env_flagged_samples` and forwards it to the impl, and `setup_single_controller` passes `should_mask_flagged_samples(master_config.env)`.

nemo_rl/experience/rollouts.py
michal2409: Done in 20f1f6e5. Helper cases sit next to the `_extract_mask_sample_flags` test, each gate site has one test, and a `RolloutManager` forwarding test covers the wiring.

nemo_rl/experience/rollout_manager.py
```diff
@@ -676,6 +686,7 @@ def __init__(
             max_rollout_turns=max_rollout_turns,
             policy_generation=policy_generation,  # type: ignore
             generation_config=generation_config,
+            mask_env_flagged_samples=mask_env_flagged_samples,
```
yuki-97, line 690: let's add a brief comment here to tell `mask_env_flagged_samples` is `AsyncNemoGymRolloutImpl` specific and no use for  `AsyncRolloutImpl`.

michal2409: Done in 79514c2d.

yuki-97 (approved):

PR #3404, opened 2026-07-29
feat(grpo): optional early stop at a validation metric threshold

michal2409:
Train-to-accuracy runs (benchmarking, recipe qualification, ablations) need training to stop as soon as the quality bar is met, rather than at a step count nobody can know in advance. This PR adds an optional early stop on a validation metric, via two config keys defaulted in the exemplar YAMLs:

- `grpo.stop_at_validation_metric` (str | null) — which reported validation metric to compare; `null` (the default) disables early stopping. Any reported metric works (for example a grouped pass@k); a metric that validation does not report fails loudly listing the available keys.
- `grpo.stop_at_validation_threshold` (float | null) — end training once the chosen metric reaches this value; `setup()` asserts it is set whenever the metric is.

The stop takes effect at the end of the stopping step: with checkpointing enabled the step is saved first (forced like a last step, carrying its validation metrics for `get_best_checkpoint_path()`), and every exit flushes pending checkpoint finalization the same way the existing timeout/max-steps early returns do.

Covered by unit tests parametrized over all three trainers (grpo_train, async_grpo_train, grpo_train_sync): periodic threshold stop, initial-validation stop, configurable metric, loud failure on an unreported metric, and checkpoint-on-stop.


tests/unit/algorithms/test_grpo.py
```diff
@@ -2445,6 +2445,79 @@ def test_grpo_train_skips_prev_logprobs_when_force_on_policy_ratio(
     )
 
 
+@pytest.mark.parametrize("train_func", [grpo_train, async_grpo_train])
```
yuki-97: `grpo_train_sync` received the same three-line early-stop guards (`grpo_sync.py:527-541` initial and `1036-1050` periodic) but isn't parametrized here — the initial + periodic threshold paths are only exercised for `grpo_train` and `async_grpo_train`.

Same shape as [the previous comment on PR #3400](https://github.com/NVIDIA-NeMo/RL/pull/3400#discussion_r3675275308); could reuse the `mock_sync_grpo_infrastructure` helper you added there.

nemo_rl/algorithms/grpo.py
```diff
@@ -259,6 +259,12 @@ class GRPOConfig(TypedDict):
     # final checkpoint has validation metrics, which is required for get_best_checkpoint_path().
     val_at_end: bool
     max_val_samples: int | None  # None for NeMo-Gym compatibility
+    # Early stop: end training once the chosen validation metric reaches
+    # this threshold; null disables early stopping.
+    stop_at_validation_threshold: float | None
+    # Which validation metric the early stop compares, e.g. accuracy (always
+    # reported) or pass_k (grouped validation).
+    stop_at_validation_metric: str
```
yuki-97: let's use `stop_at_validation_metric` to decide whether we do the early stop instead of `stop_at_validation_threshold`. still `None` means not do.

and could add an assert in setup to guard when `stop_at_validation_metric` is not None, then `stop_at_validation_threshold` should also be not None.

```suggestion
    stop_at_validation_metric: str | None
```

yuki-97 (commented):
thanks @michal2409 , LGTM, just two minor nits.

yuki-97 (commented):
thanks @michal2409 , LGTM, just two minor nits.

nemo_rl/algorithms/grpo.py
michal2409: Done in aecf0af0. `stop_at_validation_metric` (str | null) is the switch now — null disables — and `setup()` asserts the threshold is set whenever the metric is.

tests/unit/algorithms/test_grpo.py
michal2409: Done in aecf0af0. The stop tests are parametrized over `grpo_train_sync` via `mock_sync_grpo_infrastructure`, plus an initial-validation stop test covering all three trainers.

yuki-97 (dismissed):
thanks @michal2409 , LGTM

michal2409:
Same pre-existing L1 PPO flake as on #3401 (critic loss vs the 6.0 bound since the vLLM 0.25.1 bump): raised the bound to 8.0 with a note pointing at #3412.

yuki-97:
thanks for the bump for L1 PPO threshold, tracked at https://github.com/NVIDIA-NeMo/RL/issues/3412#issuecomment-5162595220 and bump `critic/loss (max)` bound from 6.0 to 8.0 in this PR to unblock main.

yuki-97 (approved):

PR #2341, opened 2026-04-25
perf: Benchmark tuning

guyueh1:
# What does this PR do ?

Tuning qwen30B and deepseek configs in perf benchmarks.

Rebase of https://github.com/NVIDIA-NeMo/RL/pull/2328 and https://github.com/NVIDIA-NeMo/RL/pull/2333

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

Issue #2180, opened 2026-03-31
Race condition: NaN in generation_logprobs when in_flight_weight_updates=true with vLLM PP>1


sharonyu-115:
**Describe the bug**

When using async GRPO with `in_flight_weight_updates=true` and vLLM `pipeline_parallel_size > 1`, weight updates can execute concurrently with vLLM forward passes on the same GPU worker, producing NaN values in `generation_logprobs`. This corrupts 6 training metrics (`gen_kl_error`, `policy_kl_error`, `js_divergence_error`, `token_mult_prob_error`, `max_seq_mult_prob_error`, `approx_entropy`) at ~2% of training steps in my test run.

**Root cause:** vLLM's `RayDistributedExecutor` dispatches forward passes via Ray compiled DAGs (background thread) and weight updates via regular `.remote()` calls (main thread). These can execute concurrently on the same Ray actor. With PP>1, the EngineCore uses `step_with_batch_queue` which returns before the forward pass completes, allowing a weight update to overlap with an in-flight forward pass. The forward pass reads partially-updated model weights, producing NaN logprobs.

PP=1 is safe because the EngineCore's blocking `step()` serializes everything.

**Steps/Code to reproduce bug**

1. Run async GRPO with vLLM PP=2 and `in_flight_weight_updates=true`:

```yaml
grpo:
  async_grpo:
    enabled: true
    max_trajectory_age_steps: 1
    in_flight_weight_updates: true

policy:
  generation:
    colocated:
      enabled: false
    vllm_cfg:
      async_engine: true
      pipeline_parallel_size: 2
```

2. Monitor `train/gen_kl_error` in wandb or grep for `Generation KL Error: nan` in the driver log.

e.g. wandb metrics with nan values:

<img width="2068" height="659" alt="Image" src="https://github.com/user-attachments/assets/0cf65134-92b4-410d-9969-c42479fac982" />

3. To confirm, run the same config with `in_flight_weight_updates: false` (safe) or `pipeline_parallel_size: 1` (control) — both produce zero NaN.

**Experimental results** (Qwen3-8B-Base, 300 steps):

| Run | vLLM PP | `in_flight` | NaN steps | NaN rate |
|-----|---------|------------|-----------|----------|
| Safe | 2 | false | 0 / 278 | 0% |
| **Race** | **2** | **true** | **5 / 300** | **1.7%** |
| Control | 1 | true | 0 / 300 | 0% |
[...]

Issue #2124, opened 2026-03-18
Many layers missing when converting GPTOSS Megatron checkpoint to HF format


dhineshkumar-r:
**Describe the bug**

Many layers are not found in generator model when converting gpt-oss checkpoint Megatron to HF format.

```
load_return: _IncompatibleKeys(missing_keys=['decoder.layers.0.self_attention.linear_proj._extra_state', 'decoder.layers.0.self_attention.linear_qkv._extra_state', 'decoder.layers.0.pre_mlp_layernorm._extra_state', 'decoder.layers.0.mlp.experts.linear_fc1._extra_state', 'decoder.layers.0.mlp.experts.linear_fc2._extra_state', 'decoder.layers.1.self_attention.linear_proj._extra_state', 

....

ayers.22.mlp.experts.linear_fc1._extra_state', 'decoder.layers.22.mlp.experts.linear_fc2._extra_state', 'decoder.layers.23.self_attention.linear_proj._extra_state', 'decoder.layers.23.self_attention.linear_qkv._extra_state', 'decoder.layers.23.pre_mlp_layernorm._extra_state', 'decoder.layers.23.mlp.experts.linear_fc1._extra_state', 'decoder.layers.23.mlp.experts.linear_fc2._extra_state', 'decoder.final_layernorm._extra_state'], unexpected_keys=[])
tokenizer_config.json: 4.20kB [00:00, 20.5MB/s]
tokenizer.json: 100%|█████████████████████████████████████████████████| 27.9M/27.9M [00:00<00:00, 28.0MB/s]
special_tokens_map.json: 100%|██████████████████████████████████████████| 98.0/98.0 [00:00<00:00, 1.09MB/s]
chat_template.jinja: 16.7kB [00:00, 61.2MB/s]
generation_config.json: 100%|█████████████████████████████████████████████| 177/177 [00:00<00:00, 1.87MB/s]
model.safetensors.index.json: 36.4kB [00:00, 92.2MB/s]                               | 0/5 [00:00<?, ?it/s]
model-00002-of-00002.safetensors: 100%|████████████████████████████████| 4.17G/4.17G [00:10<00:00, 398MB/s]
model-00000-of-00002.safetensors: 100%|████████████████████████████████| 4.79G/4.79G [00:14<00:00, 337MB/s]
model-00001-of-00002.safetensors: 100%|████████████████████████████████| 4.80G/4.80G [00:19<00:00, 245MB/s]
[...]

PR #3407, opened 2026-07-29
feat(nemo-gym): train on rollouts from external agent harnesses

ananthsub:
Trains on rollouts from an agent harness that drives its own model calls and returns no token IDs.

Gym captures the exact token IDs and log probabilities inside its model server and rebuilds each rollout into a trainable Responses payload. NeMo-RL assigns rollout identity, constructs the `TokenSource` inside its rollout actor, consumes the rebuilt result, and retires the frozen capture only after the rollout consumer accepts it.

Depends on the Gym stack: NVIDIA-NeMo/Gym#2190, #2124, #2125, #2126, #2341, #2180, #2181, and #2349.

## Control and data flow

```mermaid
sequenceDiagram
    autonumber
    participant RL as NemoGym actor
    participant G as Gym run_examples
    participant H as Agent harness
    participant M as Gym model-server worker
    participant K as TokenSink
    participant S as TokenSource
    participant T as Trainer consumer

    RL->>RL: Run token_source_factory inside actor
    RL->>RL: Assign _ng_rollout_id
    RL->>G: Dispatch selected rows
    G->>H: Run agent
    H->>M: Model calls under /training-token-capture/{rollout_id}
    M->>K: Append TokenEntry records
    G-->>RL: Return rollout without token metadata
    RL->>S: Freeze and read snapshot
    RL->>RL: Rebuild response and preserve original media items
    RL-->>T: Yield trainable rollout
    T-->>RL: Request next rollout
    RL->>S: Conditionally retire accepted snapshot
```

## Integration behavior
- Capture selection comes from Gym's resolved static agent configuration. `token_id_capture.all_agents: true` opts every agent into capture for training-only runs.
- `_build_token_capture_source` uses a local `TokenCaptureStore` when `token_id_capture.dir` is configured.
- A training framework passes `token_source_factory` to `spinup_nemo_gym_actor`. Ray serializes the factory with the actor constructor, and the factory runs inside the actor process and virtual environment.
- The factory must return Gym's `TokenSource` protocol. It may wrap a TransferQueue handle or any other framework-owned transport client.
- Gym model-server workers construct the corresponding configured `TokenSink` and `LineageStore` proxies in their own processes. Their adapter package must be installed in the model-server environment.
- The source, sink, and lineage proxies do not share Python objects or virtual environments. They communicate through the framework-owned transport.
- Frozen records remain available while the rollout is yielded. Conditional retirement runs only after the consumer requests the next item, so failed or abandoned consumption does not delete evidence prematurely.
- `shutdown()` closes the actor-owned source before stopping Gym servers.
- Full UUID rollout IDs avoid collisions across actors and restarts.
- Multimodal postprocessing uses the original response output, while training tokens come from the rebuilt output.
- GRPO grouping keeps dataset prompt identity instead of grouping on the external harness's captured first-call prompt.

## Source construction

The default file adapter is selected with `token_id_capture.dir`. A framework data plane injects an actor-local factory instead of naming a source class in Gym configuration.

```python
source_factory = TransferQueueTokenSourceFactory(queue_handle)
actor = spinup_nemo_gym_actor(
    ...,
    token_source_factory=source_factory,
)
```

The source must implement Gym's `TokenSource` protocol: `freeze`, conditional `drop`, and idempotent `close`. The model-server side uses the corresponding `TokenSink` and `LineageStore` protocols supplied by the framework.

## Metrics

The integration reports `token_capture/rebuilt_fraction`, `rollouts_rebuilt`, `rollouts_unbuilt`, `calls_per_rollout_mean`, `chains_per_rollout_mean`, `delivered_fraction_mean`, `quarantined_fraction_mean`, `empty_generation_calls`, `parent_link_failures`, `masked_rollouts`, and `incomplete_rollouts`.


nemo_rl/environments/nemo_gym.py
```diff
@@ -311,6 +436,26 @@ async def run_rollouts(
...
+            # A rollout whose returned output has no token ids gets them attached from what was
+            # recorded for it; one that already carries them is left alone. Gym decides that from
+            # the rollout itself, so a batch can mix native agents and external harnesses, and this
+            # call does not need to know which this is.
+            if token_source is not None:
+                # run_examples returns the harness result without the correlation key, which the
+                # lookup needs. Gym's rollout collection copies it onto the result for the same
+                # reason; this is that copy for the low-level path.
+                nemo_gym_result[ROLLOUT_ID_KEY_NAME] = nemo_gym_row[ROLLOUT_ID_KEY_NAME]
+                built = await finalize_rollout_token_capture(
```
cmunley1, line 728: Should we preserve the pre-rollout input_ids for GRPO grouping?

The postprocessed capture returns each harness’s first-call prompt as NemoGymRolloutResult.input_ids, and grpo.py replaces the original repeated prompt IDs with it. Harness-specific workspace paths can then split one task into singleton groups, making baseline equal reward and all advantages zero.

I ran into this with OpenCode: 7/16 successful rollouts, but zero advantages. The minimal fix seems to be removing the input_ids = nemo_gym_rollout_result.input_ids assignment in grpo.py, while keeping the captured IDs in repeated_batch["message_log"] for training.

cmunley1: actually advantage estimator also seems to need this fix

PR #3408, opened 2026-07-29
feat(research): add flow_grpo, Flow-GRPO for Qwen-Image

triple-mu:
# What does this PR do ?

Adds Flow-GRPO to NeMo RL as a research project under `research/flow_grpo/`: online RL for image diffusion (flow-matching, text-to-image) models, following [Flow-GRPO](https://arxiv.org/abs/2505.05470), with a Qwen-Image LoRA recipe on an OCR task. The prompt asks for an image with a specific piece of text rendered in it, and the reward measures how well the generated image reads back.

A deterministic ODE sampler has no action density to do policy gradients on, so the rollout follows Flow-GRPO and turns the flow-matching ODE into an SDE whose per-step transition log-probs are computable in closed form. On top of that it is the usual GRPO loop: group-relative advantage, clipped policy-gradient loss.

Rollouts run through NeMo Automodel/Diffusers inside the policy worker and share the training model (no refit step). That differs from core's split between policy-training and generation backends, which is why this lands under `research/` first (see the review discussion below). The research-folder move, worker fixes and a checkpoint-resume fix were contributed by @sajadn via triple-mu/RL#1 on top of the reviewed core implementation.

Layout (following `research/template_project`):

| | |
|---|---|
| `flow_grpo/algorithms/` | training loop and `MasterConfig` (`flow_grpo.py`), clipped policy-gradient loss (`loss/flow_grpo.py`), SDE step and log-prob kernel (`sde.py`) |
| `flow_grpo/models/` | config schemas and protocols (`interfaces.py`), Qwen-Image pipeline adapter (`pipeline.py`), controller-side policy (`policy/flow_grpo_policy.py`), Ray worker with Automodel loading, LoRA and DP all-reduce (`policy/workers/flow_grpo_worker.py`) |
| `flow_grpo/data/`, `flow_grpo/environments/` | prompt dataset and OCR dataset export; pluggable image-reward environment (`dummy`, `jpeg_compressibility`, `pickscore`, `ocr`, `genrm_ocr`) |
| `flow_grpo/actor_environments.py` | registers the policy and reward actors with this project's uv environment |
| `run_flow_grpo.py`, `configs/recipes/` | entry point; exemplar config (`genrm_ocr` judge, 300 steps) and nightly recipe (CPU PaddleOCR reward, 60 steps) |
| `tests/unit`, `tests/functional`, `tests/test_suites` | discovered by the research runners; the nightly recipe is listed in `nightly.txt`, the 300-step exemplar in `disabled.txt` because it needs an externally served judge |
| `README.md` | the guide plus reproduction steps |

Dependencies (`nemo-rl[automodel]`, `diffusers`, `paddleocr`, `paddlepaddle`) are declared in `research/flow_grpo/pyproject.toml`; the project is a uv workspace member and shares the root `uv.lock`. Core changes are limited to the workspace member, `uv.lock`, the suite lists and the research-project section of the testing skill; the shared CI changes for research test discovery are in #4216.

# Usage

```bash
cd research/flow_grpo
uv sync --locked --inexact --group test
uv run --locked --no-sync python -m flow_grpo.data.export_ocr_prompts --out-dir data/ocr

# The exemplar reward is a Qwen3-VL judge served with the repository's vLLM extra:
uv run --locked --directory ../.. --extra vllm vllm serve Qwen/Qwen3-VL-8B-Instruct --tensor-parallel-size 4 --port 30000
export GENRM_BASE_URL=http://<host>:30000/v1

uv run --locked --no-sync python run_flow_grpo.py --config configs/recipes/flow_grpo_qwen_image_ocr.yaml
```

`research/flow_grpo/README.md` covers hyperparameters, reward plugins, metrics to watch and how to reproduce the results below.

- [x] Did you add or update any necessary documentation?

# Additional Information

## Results

Qwen-Image with LoRA, 300 steps on a single 8×B200 node. The validation reward (deterministic ODE, fixed seed, so comparable across steps) rises from 0.889 to 0.979. Re-scoring the step-210 checkpoint with a scorer the model never trained against, PaddleOCR for text accuracy and PickScore for image quality:

| | Base | Trained |
| --- | --- | --- |
| Character similarity (1 − CER) | 0.693 | 0.845 |
| Exact-match rate | 0.465 | 0.648 |
| Word recall | 0.817 | 0.903 |
| PickScore | 23.02 | 23.04 |

Text accuracy jumps while PickScore stays flat. The nightly recipe (CPU PaddleOCR reward, 60 steps, 1×8 B200) passed its gate, val reward 0.670 → 0.897, and @sajadn re-ran the nightly config on the research layout.

## Verification of the research layout

On 4×B200 (`nvcr.io/nvidia/pytorch:26.07-py3`, uv 0.11.29, `NEMO_RL_PY_EXECUTABLES_SYSTEM=1`):

| Check | Result |
|---|---|
| `uv lock --check` (uv 0.11.28, the Dockerfile pin) | passes |
| ruff 0.9.9 `check`, `check --select I`, `format --check` on `research/flow_grpo` | clean |
| `pyrefly check` with `research/flow_grpo/pyrefly.toml` | 0 errors |
| `uv run --no-sync pytest tests/unit` in `research/flow_grpo` | 107 passed |
| `tests/functional/flow_grpo.sh` (1 GPU, `tiny-random/Qwen-Image`, 5 steps) | passes: `step_5/policy/{model,optim}` written, median `train/mean_ratio` 0.99998, max `train/grad_norm` 2.3e-4 |

## Notes

- `test_all_test_scripts_accounted_for_in_test_suites` and the recipe accounting test fail until #4216 lands (research paths in `nightly.txt` / `disabled.txt`); this branch will be rebased once it merges.
- Training is LoRA on single-GPU data-parallel workers; there is no FSDP/Megatron sharding. Other flow-matching pipelines can plug in via the `DiffusionPipelineAdapter` protocol.


nemo_rl/models/diffusion/workers/diffusion_worker.py
```diff
@@ -0,0 +1,789 @@
...
+    optimizer: Any
+    _img_shapes: list[tuple[int, int, int]]
+    adapter: Any
+    _loss_fn: Any
+
+    @staticmethod
+    def configure_worker(
+        num_gpus: int | float | None = None,
+        bundle_indices: tuple[int, list[int]] | None = None,
+    ) -> tuple[dict[str, Any], dict[str, str], dict[str, Any], dict[str, Any]]:
```
sharonyu-115: **[BUG][CRITICAL]** [`diffusion_worker.py:276-285`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/models/diffusion/workers/diffusion_worker.py#L276-L285) [confidence: 98]

`DiffusionPolicyWorker.configure_worker(num_gpus, bundle_indices)` has no `num_gpus_per_node` param (or `**kwargs`). `RayWorkerGroup._create_workers_from_bundle_indices` -> `IsolatedWorkerInitializer.create_worker` unconditionally calls `configure_worker(num_gpus=..., bundle_indices=..., num_gpus_per_node=self.cluster.num_gpus_per_node)` ([worker_groups.py call site](https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/distributed/worker_groups.py)). Every sibling worker (`megatron_policy_worker.py`, `vllm_worker.py`, `megatron_value_worker.py`) accepts this 3rd param — the diffusion worker alone breaks the contract. **Every diffusion-GRPO run crashes with `TypeError` at worker construction, before any rollout runs.** Verified by tracing the full call path (`policy.py` -> `worker_groups.py:335-417,586-593,140-192` -> `diffusion_worker.py:276`) with no wrapper/shim intercepting the call.

Suggested fix:
```suggestion
    @staticmethod
    def configure_worker(
        num_gpus: int | float | None = None,
        bundle_indices: tuple[int, list[int]] | None = None,
        num_gpus_per_node: int | None = None,
    ) -> tuple[dict[str, Any], dict[str, str], dict[str, Any], dict[str, Any]]:
        del num_gpus_per_node  # one GPU per worker; no per-bundle seeding needed
        resources = {"num_gpus": 1, "num_cpus": 4}
```

_rl-expert, confirmed by devil-advocate_

nemo_rl/models/diffusion/sde.py
```diff
@@ -0,0 +1,145 @@
...
+        if prev_sample is None:
+            variance_noise = torch.randn(
+                model_output.shape,
+                generator=generator,
+                device=model_output.device,
+                dtype=model_output.dtype,
+            )
+            prev_sample = prev_sample_mean + std_dev_t * variance_noise
+
+        log_prob = -(prev_sample.detach() - prev_sample_mean).square()
```
sharonyu-115: **[BUG][CRITICAL]** [`sde.py:122-139`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/models/diffusion/sde.py#L122-L139) [confidence: 90]

The `"cps"` branch of `sde_step_with_logprob` computes `log_prob = -(prev_sample - prev_sample_mean).square()`, missing `/(2*std_dev_t^2)`, `-log(std_dev_t)`, and `-log(sqrt(2*pi))` that the sibling `"sde"` branch (the default) correctly includes. Not a valid Gaussian log-density — corrupts the GRPO importance ratio whenever `sde_type="cps"` is set. Both shipped configs default to `sde_type="sde"`, so the exemplar/nightly recipe is unaffected, but `sde_type` is a plain unvalidated `str` field, so this is live, reachable config surface that will silently corrupt training the first time anyone tries `"cps"`.

```suggestion
        std = std_dev_t
        log_prob = (
            -(prev_sample.detach() - prev_sample_mean).square() / (2 * std.square())
            - torch.log(std)
            - math.log(math.sqrt(2 * math.pi))
        )
```

The existing test [`test_sde.py:130-164`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/tests/unit/models/diffusion/test_sde.py#L130-L164) reproduces the same buggy formula as its "expected" value (a change-detector, not an oracle) — see the companion comment on that file for a verified independent-oracle test to add alongside this fix.

_rl-expert and test-agent independently converged on this via different methods (formula inspection vs. `torch.distributions.Normal` oracle comparison); confirmed by devil-advocate_

tests/unit/models/diffusion/test_sde.py
```diff
@@ -0,0 +1,232 @@
...
+            noise_level=noise_level,
+            prev_sample=prev_sample,
+            sde_type="cps",
+        )
+    )
+
+    torch.testing.assert_close(actual_prev_sample, prev_sample)
+    torch.testing.assert_close(actual_logprob, expected_logprob)
+    torch.testing.assert_close(actual_mean, expected_mean)
+    torch.testing.assert_close(actual_std_dev_t, expected_std_dev_t)
```
sharonyu-115: **[TEST]** companion to the [`sde.py:139` cps log-prob bug](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/models/diffusion/sde.py#L139) [confidence: 85]

`test_cps_step_recomputes_logprob_from_prev_sample` re-derives its expected value using the exact same (buggy) formula as the source, so it can't catch a shared-formula bug. test-agent wrote and verified an independent-oracle test against `torch.distributions.Normal` that would have caught this — recommend adding both this test and an equivalent oracle test for the `"sde"` branch:

```suggestion
def test_cps_logprob_is_not_a_normalized_gaussian_density():
    """Documents an asymmetry vs the 'sde' branch: 'cps' log_prob is missing
    the 1/(2*std^2) scale term and the Gaussian normalization constants that
    the 'sde' branch has. If this ever starts passing, the two branches'
    log-prob semantics have changed and ratio/approx_kl scale assumptions
    across sde_type should be re-audited."""
    scheduler = FakeFlowMatchScheduler()
    model_output = torch.tensor([[[0.2, -0.1], [0.4, -0.3]]])
    sample = torch.tensor([[[1.0, -1.0], [0.5, -0.5]]])
    timestep = torch.tensor([750.0])
    prev_sample, log_prob, mean, std = sde_step_with_logprob(
        scheduler, model_output, timestep, sample,
        noise_level=0.4, generator=torch.Generator().manual_seed(0),
        sde_type="cps", reduce_per_sample=False,
    )
    oracle = torch.distributions.Normal(mean, std.expand_as(mean)).log_prob(prev_sample)
    assert not torch.allclose(log_prob, oracle, atol=1e-6)
```
(verified passing standalone against the real `sde.py` source)

_test-agent_

nemo_rl/data/datasets/text_to_image_prompt.py
```diff
@@ -0,0 +1,113 @@
...
+        return len(self._records)
+
+    def __getitem__(self, idx: int) -> DiffusionDatumSpec:
+        rec = self._records[idx]
+        datum: DiffusionDatumSpec = {
+            "prompt": rec["prompt"],
+            "negative_prompt": rec.get("negative_prompt", self.negative_prompt_default),
+            "metadata": rec.get("metadata", {}),
+            "idx": idx,
+            "loss_multiplier": 1.0,
```
sharonyu-115: **[BUG]** [`text_to_image_prompt.py:94`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/data/datasets/text_to_image_prompt.py#L94) [confidence: 85]

`DiffusionDatumSpec.loss_multiplier` ([`interfaces.py:193`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/models/diffusion/interfaces.py#L193)) is a documented per-sample loss-masking field (mirrors the standard `DatumSpec` schema used elsewhere), but `__getitem__` hardcodes it to `1.0` — it never reads it from the jsonl record at all, and nothing downstream reads the collated `batch["loss_multipliers"]` either. The actual masking path ([`diffusion_grpo.py:315-327`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/algorithms/diffusion_grpo.py#L315-L327)) reads a *different*, undocumented nested key: `metadata["loss_multiplier"]`. A user setting `"loss_multiplier": 0.0` in a jsonl record (matching this field's semantics everywhere else in the codebase) is silently ignored.

Suggest either wiring `batch["loss_multipliers"]` through to `sample_mask` directly, or dropping the dead top-level field and documenting that masking goes through `metadata["loss_multiplier"]`.

_bug-finder, confirmed (and found to be understated) by devil-advocate_

nemo_rl/models/diffusion/workers/diffusion_worker.py
```diff
@@ -0,0 +1,789 @@
...
+            self._loss_fn = DiffusionGRPOLossFn(loss_cfg)
+        use_reference = float(loss_cfg["beta"]) > 0
+
+        # Gradient accumulation over sample-dimension chunks so the
+        # with-grad T-step recompute only holds `micro` samples' activations
+        # at a time. Chunk losses are weighted by sample count, which matches
+        # the full-batch masked_mean when masks are uniform across samples.
+        total = int(data["generation_logprobs"].shape[0])
+        # None → the whole rollout batch in one backward pass.
+        micro = self.config["train_micro_batch_size"] or total
```
sharonyu-115: **[GUIDELINE]** [`diffusion_worker.py:678`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/models/diffusion/workers/diffusion_worker.py#L678) [confidence: 85]

`micro = self.config["train_micro_batch_size"] or total` — since `0` and `None` are both falsy in Python, `train_micro_batch_size: 0` silently collapses to the documented `None` sentinel (whole-batch backward) instead of raising a config error. No validator on this field forbids `0` (contrast `ppo_mini_batch_size`, which does have a guard on its analogous field).

```suggestion
            micro = (
                total
                if self.config["train_micro_batch_size"] is None
                else self.config["train_micro_batch_size"]
            )
```

_rl-expert, confirmed by devil-advocate_

nemo_rl/algorithms/diffusion_grpo.py
```diff
@@ -0,0 +1,519 @@
...
+    loss_multiplier: torch.Tensor,
+) -> BatchedDataDict[DiffusionTrainDataSpec]:
+    T = traj["timesteps"].shape[-1]
+    # Keep only the columns inside the SDE window so the training-side
+    # recompute never forwards out-of-window steps. Window starts may differ
+    # per sample (sampled per worker), but the widths must match.
+    mask = traj["timestep_mask"]
+    widths = mask.sum(dim=1).long()
+    w = int(widths.max().item())
+    if 0 < w < T:
```
sharonyu-115: **[GUIDELINE]** [`diffusion_grpo.py:171`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/algorithms/diffusion_grpo.py#L171) [confidence: 82]

`sde_window_size=0` makes `compute_window_mask` return an all-zero mask; `_build_train_data`'s `0 < w < T` check then skips the re-masking block, and `masked_mean`'s `+1e-8` epsilon guard means the loss becomes silently exactly `0.0` every step (no crash, no NaN) — a silent no-op training run. This config family already uses `0` as an explicit "disable" sentinel for `val_period`/`max_val_samples`, so a user could reasonably (and wrongly) expect `sde_window_size: 0` to mean "no windowing" rather than "zero-width window."

Suggest validating `sde_window_size > 0` (or `None`) at config-load time, distinct from the validation path's intentional `sde_window_size=0` override.

_bug-finder, confirmed (and found more plausible than originally framed) by devil-advocate_

nemo_rl/environments/image_reward_environment.py
```diff
@@ -0,0 +1,552 @@
...
+                        {"type": "text", "text": GENRM_OCR_PROMPT},
+                    ],
+                },
+            ],
+            "model": self._model,
+            **self._sampling_params,
+        }
+        # Mirrors verl-omni genrm_ocr: unbounded timeout, no retry; a failed
+        # request raises and fails the training step loudly.
+        response = requests.post(self._url, json=request, timeout=None)
```
sharonyu-115: **[BUG]** [`image_reward_environment.py:400`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/environments/image_reward_environment.py#L400) [confidence: 80]

`requests.post(self._url, json=request, timeout=None)` — `timeout=None` disables both connect and read timeouts. A wedged/unresponsive served judge doesn't raise (which the "fail loud" comment anticipates) — it hangs the synchronous training loop forever with no exception and no log line. Real, deliberately-authored footgun, not a hidden contradiction.

Suggest a finite, generous timeout (e.g. 300s) so the failure still surfaces loudly instead of hanging indefinitely.

_bug-finder, confirmed by devil-advocate_

tests/unit/models/diffusion/test_qwen_image_pipeline_adapter.py
```diff
@@ -0,0 +1,382 @@
...
+def test_apply_true_cfg_preserves_norm_of_conditional():
+    torch.manual_seed(0)
+    a = torch.randn(2, 4)
+    b = torch.randn(2, 4)
+    combined = apply_true_cfg(a, b, true_cfg_scale=3.0)
+    # The last-dim norm of the result should equal the last-dim norm of the
+    # conditional branch.
+    assert torch.allclose(
+        torch.norm(combined, dim=-1), torch.norm(a, dim=-1), atol=1e-5
+    )
```
sharonyu-115: **[TEST]** [`test_qwen_image_pipeline_adapter.py:92-107`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/tests/unit/models/diffusion/test_qwen_image_pipeline_adapter.py#L92-L107) [confidence: 85]

`test_apply_true_cfg_preserves_norm_of_conditional` only checks output norm equals input norm — since [`apply_true_cfg`](https://github.com/NVIDIA-NeMo/RL/blob/f9fdce49573ebaad29827931bd11d4a1e6789c2d/nemo_rl/models/diffusion/pipeline.py#L46-L62) always renormalizes to the norm of whatever it's given as its first argument, a `noise_pred`/`negative_noise_pred` argument swap at the call site would still pass this test undetected.

```suggestion
def test_apply_true_cfg_extrapolates_away_from_negative():
    """Catches a noise_pred/negative_noise_pred swap: preserves-norm alone
    can't distinguish 'extrapolate away from uncond' from its mirror image,
    since both get renormalized to the same cond_norm."""
    noise_pred = torch.tensor([[1.0, 0.0]])
    negative_noise_pred = torch.tensor([[0.0, 1.0]])
    result = apply_true_cfg(noise_pred, negative_noise_pred, true_cfg_scale=2.0)
    expected = torch.tensor([[2.0, -1.0]]) / math.sqrt(5.0)
    assert torch.allclose(result, expected, atol=1e-6)
    swapped = torch.tensor([[-1.0, 2.0]]) / math.sqrt(5.0)
    assert not torch.allclose(result, swapped, atol=1e-6)
```
(verified passing against the real `pipeline.py` — current implementation is correct, this only adds a regression guard)

_test-agent, confirmed by devil-advocate_

nemo_rl/models/diffusion/workers/diffusion_worker.py
```diff
@@ -0,0 +1,789 @@
...
+        if k == "ratio_min":
+            acc[k] = min(acc.get(k, float("inf")), v)
+        elif k == "ratio_max":
+            acc[k] = max(acc.get(k, float("-inf")), v)
+        else:
+            acc[k] = acc.get(k, 0.0) + v * weight
+
+
+@ray.remote
+class DiffusionPolicyWorker:  # pragma: no cover
```
sajadn: One high level comment is you named the classes very generically such as DiffusionPolicyWorker, it's better to name them more specific like `FlowGRPOPolicyWorker` given that we'll probably have some other algorithms added.

examples/configs/flow_grpo_qwen_image_ocr.yaml
```diff
@@ -0,0 +1,121 @@
...
+    max_sequence_length: 256
+    guidance_scale: null
+
+  algo:
+    noise_level: 1.2
+    sde_type: "sde"
+    sde_window_size: 2
+    sde_window_range: [0, 5]
+
+  lora_cfg:
```
sajadn, line 65: do you have any experiment/recipe without Lora? It's more out of curiosity not necessary for this PR.

nemo_rl/algorithms/diffusion_grpo.py
```diff
@@ -0,0 +1,519 @@
...
+class DiffusionDataConfig(BaseModel, extra="allow"):
+    train: DiffusionDataSplitConfig
+    val: DiffusionDataSplitConfig | None = None
+
+
+class DiffusionEnvConfig(BaseModel, extra="allow"):
+    image_reward: ImageRewardEnvConfig
+
+
+class DiffusionCheckpointingConfig(BaseModel, extra="allow"):
```
sajadn: can't we reuse the existing checkpointingConfig? https://github.com/NVIDIA-NeMo/RL/blob/aa739509ce3772a18a4d02da651b86c9e898e8cb/nemo_rl/utils/checkpoint.py#L98

nemo_rl/algorithms/diffusion_grpo.py
```diff
@@ -0,0 +1,519 @@
...
+    enabled: bool = True
+    checkpoint_dir: str = "results/diffusion_grpo"
+    save_period: int = 100
+
+
+class DiffusionMasterConfig(BaseModel, extra="allow"):
```
sajadn: following the other algorithms you can change this class to ```MasterConfig```, look at grpo for example: https://github.com/NVIDIA-NeMo/RL/blob/aa739509ce3772a18a4d02da651b86c9e898e8cb/nemo_rl/algorithms/grpo.py#L357

nemo_rl/algorithms/diffusion_grpo.py
```diff
@@ -0,0 +1,519 @@
...
+    """Schema for `examples/configs/diffusion_grpo_qwen_image*.yaml`."""
+
+    policy: DiffusionPolicyConfig
+    loss_fn: DiffusionLossConfig
+    grpo: DiffusionGRPOAlgoConfig
```
sajadn: let's make this flow_grpo

nemo_rl/algorithms/diffusion_grpo.py
```diff
@@ -0,0 +1,519 @@
...
+
+        if traj["metadata"]:
+            # sample_trajectory replicates metadata K times alongside latents;
+            # guard the contract so a mismatch fails here, not inside the loss.
+            assert len(traj["metadata"]) == rewards.shape[0], (
+                f"metadata length {len(traj['metadata'])} != rollout batch "
+                f"{rewards.shape[0]}"
+            )
+            loss_mult = torch.tensor(
+                [m.get("loss_multiplier", 1.0) for m in traj["metadata"]],
```
sajadn: `loss_multiplier` is a key in `metadata` or is it at the same level at `metadata`? in the datum i see that it's a sibling of `metadata`?

nemo_rl/data/datasets/text_to_image_prompt.py
sajadn: do we really need loss_multiplier given that it's hardcoded

nemo_rl/algorithms/loss/diffusion_grpo.py
```diff
@@ -0,0 +1,177 @@
...
+
+from typing import Any
+
+import torch
+
+from nemo_rl.algorithms.utils import masked_mean
+from nemo_rl.models.diffusion.interfaces import DiffusionLossConfig
+
+
+class DiffusionGRPOLossFn:
```
sajadn: again naming issue

nemo_rl/algorithms/loss/flow_grpo.py
```diff
@@ -0,0 +1,177 @@
...
+            mask = timestep_mask.float() * sample_mask.float().unsqueeze(-1)
+        policy_loss = masked_mean(pg, mask)
+
+        with torch.no_grad():
+            clipfrac_higher = masked_mean(
+                ((ratio - 1.0) > ratio_clip_max).float(), mask
+            )
+            clipfrac_lower = masked_mean(((1.0 - ratio) > ratio_clip_min).float(), mask)
+            clipfrac = clipfrac_higher + clipfrac_lower
+            approx_kl = masked_mean(-log_ratio, mask)
```
sajadn, line 150: can we use k2 or k3 estimator here?

nemo_rl/models/diffusion/workers/diffusion_worker.py
triple-mu: Done in `e35ffe2`. Renamed the algorithm-specific classes after the algorithm and kept the reusable ones on `diffusion`:

- `DiffusionPolicyWorker` → [`FlowGRPOPolicyWorker`](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/models/diffusion/workers/flow_grpo_worker.py#L252), `DiffusionPolicy` → `FlowGRPOPolicy`
- `DiffusionGRPOLossFn` → `FlowGRPOLossFn`, `DiffusionGRPOAlgoConfig` → `FlowGRPOAlgoConfig`, `DiffusionLossConfig` → `FlowGRPOLossConfig`
- Modules, entrypoint, exemplar, recipe and guide follow: `nemo_rl/algorithms/flow_grpo.py`, `examples/run_flow_grpo.py`, `flow_grpo-qwen-image-ocr-1n8g-dp8-lora`.

The two algorithm-specific schemas also moved out of `interfaces.py`: `FlowGRPOAlgoConfig` next to the training loop, `FlowGRPOLossConfig` next to the loss (mirroring `ClippedPGLossConfig` in `loss_functions.py`).

`sde.py`, `pipeline.py`, `interfaces.py`, `image_reward_environment.py` and the prompt dataset keep the `diffusion` name — they aren't flow-GRPO specific, and a second diffusion RL algorithm should reuse them as-is.

examples/configs/flow_grpo_qwen_image_ocr.yaml
triple-mu: No, every run we have is LoRA. The code path does support `policy.lora_cfg.enabled: false` (full-parameter training), but the KL reference is defined as "base model with the adapter disabled", so `loss_fn.beta > 0` has no reference copy without LoRA — [`MasterConfig._kl_requires_lora`](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/algorithms/flow_grpo.py#L153-L161) rejects that combination at config load, and [the worker raises](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/models/diffusion/workers/flow_grpo_worker.py#L638-L643) if it slips through.

That guard isn't what forces LoRA here though — this recipe runs `beta: 0.0`. It's LoRA because the verl-omni baseline we match (`run_qwen_image_ocr_lora.sh`, rank 64 / alpha 128) is, and we wanted the comparison to be apples-to-apples. Happy to run a full-parameter ablation as a follow-up.

nemo_rl/algorithms/diffusion_grpo.py
triple-mu: Done. Dropped `DiffusionCheckpointingConfig` and the hand-rolled `_latest_checkpoint()`. [`MasterConfig.checkpointing`](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/algorithms/flow_grpo.py#L151) is now `nemo_rl.utils.checkpoint.CheckpointingConfig`, and the loop drives a `CheckpointManager` the same way `grpo.py` does: [resume](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/algorithms/flow_grpo.py#L300-L312) via `get_latest_checkpoint_path()` + `load_training_info()`, [save](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/algorithms/flow_grpo.py#L437-L448) via `init_tmp_checkpoint()` → `<tmp>/policy` → `finalize_checkpoint()`. The `tmp_step_N` → `step_N` rename replaces the old `optim/.metadata` completeness probe, and `keep_top_k` pruning comes for free.

Two details worth flagging: `save_optimizer` is plumbed through to the worker so the key isn't silently ignored, and the exemplar sets [`metric_name: null`](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/examples/configs/flow_grpo_qwen_image_ocr.yaml#L32) because validation runs after the save — a `val:reward_mean` ranking would score each checkpoint with the previous validation's number. The functional test moved to the new [`step_N/policy/`](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/tests/functional/flow_grpo.sh#L95) layout in `db929f6`.

nemo_rl/algorithms/diffusion_grpo.py
triple-mu: Done — it's [`MasterConfig`](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/algorithms/flow_grpo.py#L141) now, imported as `FlowGRPOMasterConfig` in `tests/unit/test_config_validation.py` like the other algorithms.

nemo_rl/algorithms/diffusion_grpo.py
triple-mu: Done — the section is [`flow_grpo:`](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/algorithms/flow_grpo.py#L146) across the exemplar, the recipe, the CLI overrides in the test scripts, and the guide. It also removed a wart in `test_config_validation.py`: the diffusion branch no longer has to run before the generic `grpo` branch, it just [checks `"flow_grpo" in config_dict`](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/tests/unit/test_config_validation.py#L147-L149).

nemo_rl/algorithms/diffusion_grpo.py
triple-mu: Good catch — it's a sibling of `metadata`, so `m.get("loss_multiplier", 1.0)` never found it and silently fell back to `1.0` on every sample. Together with your next comment, it's removed end to end.

nemo_rl/data/datasets/text_to_image_prompt.py
triple-mu: No. Removed from `DiffusionDatumSpec`, the dataset, the collate fn and the trainer. The `sample_mask` it fed is gone too — it was always all-ones and no DP padding path uses it, so the loss now [masks with `timestep_mask` alone](https://github.com/NVIDIA-NeMo/RL/blob/db929f654221165ebf31bc1a3855348b00243f3c/nemo_rl/algorithms/loss/flow_grpo.py#L140-L141).

nemo_rl/environments/image_reward_environment.py
```diff
@@ -0,0 +1,552 @@
...
+
+
+class GenRmOcrOptions(BaseModel, extra="allow"):
+    """Extra keys of a `genrm_ocr` plugins entry.
+
+    Defaults mirror verl-omni genrm_ocr.py (`DEFAULT_GRM_MODEL_PATH` and
+    `DEFAULT_SAMPLING_PARAMS`).
+    """
+
+    model: str = "~/models/tiny-random/qwen3-vl"
```
sajadn: can we remove this and make it a required field in the config file

examples/configs/flow_grpo_qwen_image_ocr.yaml
triple-mu: Correcting my earlier answer: verl-omni **does** ship full-parameter recipes. Eight scripts under `examples/flowgrpo_trainer/qwen_image/` set no `lora_rank` (the config default is 0), and `run_qwen_image_ocr.sh` opens with `# Qwen-Image full-weight RL`. Their docs list it at 4×H200 with a measured 0.510 images/GPU/s, 250 s/step. It fits in 4 GPUs because it shards with FSDP and enables `param_offload` + `optimizer_offload`, with CFG off (`true_cfg_scale=1.0`) and lr dropped from 3e-4 to 3e-5.

We can't run it, for two reasons that compound. Implementation: the worker passes [`parallel_scheme=None`](https://github.com/NVIDIA-NeMo/RL/blob/9646b0ee8d3052c3aee4eaf464f642dc9c7ebc28/nemo_rl/models/diffusion/workers/flow_grpo_worker.py#L92), so every Ray actor holds a complete single-GPU replica and gradients sync through a [hand-written per-parameter all-reduce](https://github.com/NVIDIA-NeMo/RL/blob/9646b0ee8d3052c3aee4eaf464f642dc9c7ebc28/nemo_rl/models/diffusion/workers/flow_grpo_worker.py#L706-L716) — no parameter sharding, no offload. The Qwen-Image transformer is 20.43B parameters; with torch's AdamW the moments follow the parameter dtype, so bf16 weights + gradients + two moments is ~163 GB on a single card, before the text encoder (16.6 GB) and VAE that stay resident on the same device. Resources: we only have enough GPUs to run LoRA.

So the full-parameter path is not blocked in code — only `beta > 0` is [rejected](https://github.com/NVIDIA-NeMo/RL/blob/9646b0ee8d3052c3aee4eaf464f642dc9c7ebc28/nemo_rl/models/diffusion/workers/flow_grpo_worker.py#L638-L643), since the KL reference is the base model with the adapter disabled — but it has never been exercised and would OOM today. Making it real means wiring `parallel_scheme` to FSDP2, which is its own change.

nemo_rl/algorithms/loss/diffusion_grpo.py
triple-mu: Already done — this landed 19 minutes after you commented, so you were looking at `f9fdce4`. `DiffusionGRPOLossFn` is now [`FlowGRPOLossFn`](https://github.com/NVIDIA-NeMo/RL/blob/9646b0ee8d3052c3aee4eaf464f642dc9c7ebc28/nemo_rl/algorithms/loss/flow_grpo.py#L57) in `nemo_rl/algorithms/loss/flow_grpo.py`; details in the thread on the worker comment.

One name is still arguably too generic: `DiffusionTrainDataSpec` carries flow-GRPO semantics (advantages, generation log-probs). Happy to rename it here if you want, though it widens an already large diff.

nemo_rl/algorithms/loss/flow_grpo.py
triple-mu: We can — [`calculate_kl(..., kl_type=...)`](https://github.com/NVIDIA-NeMo/RL/blob/9646b0ee8d3052c3aee4eaf464f642dc9c7ebc28/nemo_rl/algorithms/utils.py#L42) supports k1/k2/k3 and is element-wise, so the shapes drop straight in. Worth noting first that this line is [inside `torch.no_grad()`](https://github.com/NVIDIA-NeMo/RL/blob/9646b0ee8d3052c3aee4eaf464f642dc9c7ebc28/nemo_rl/algorithms/loss/flow_grpo.py#L144-L150) and only feeds the metrics dict, so the choice has no effect on the gradient — the KL *penalty* under `loss_fn.beta` is a separate analytic Gaussian KL on the means.

I'd rather keep k1 here, for two reasons. It's the parity target: verl-omni reports `actor/ppo_kl` as k1, and our archived comparison logs show it going negative, which only a signed estimator can do. And at this recipe's scale the log-ratio sits around 1e-7, where the helper's k3 (`exp(logr) - 1 - logr`) degenerates into fp32 rounding noise — I get sign flips. If you want a non-negative diagnostic, k2 is well-behaved at that scale; I'd add it alongside rather than replace k1. Let me know if you'd prefer that.

nemo_rl/environments/image_reward_environment.py
triple-mu: Agreed, and it was worse than a stale default — `~/models/tiny-random/qwen3-vl` is verl-omni's local debug path, so it could never have worked for anyone. [`model` is now required](https://github.com/NVIDIA-NeMo/RL/blob/9646b0ee8d3052c3aee4eaf464f642dc9c7ebc28/nemo_rl/environments/image_reward_environment.py#L346); the [exemplar](https://github.com/NVIDIA-NeMo/RL/blob/9646b0ee8d3052c3aee4eaf464f642dc9c7ebc28/examples/configs/flow_grpo_qwen_image_ocr.yaml#L95) already names `Qwen/Qwen3-VL-8B-Instruct`, and the nightly recipe uses the CPU `ocr` plugin so it needed no change.

The unit test that asserted the old default is replaced by one asserting the `ValidationError`.

triple-mu:
### On moving this to `research/`

Following up on the Slack thread about the discrete-diffusion design doc: happy to move continuous (image) diffusion out of core. I've scoped it and it's mostly mechanical — but there's one thing that breaks, and one question that changes the size of the job by 5x. Rather than guess, four questions.

**1. Is the driver namespace ownership, or scope?**

If it's that image diffusion shouldn't sit on names a dLLM integration will want, that's five bare uses of `diffusion`: the `nemo_rl/models/diffusion/` package, the `diffusion` extra in `pyproject.toml`, `PY_EXECUTABLES.AUTOMODEL_DIFFUSION`, `examples/configs/recipes/diffusion/`, and `tests/test_suites/diffusion/`. Renaming those to something image-specific is roughly a fifth of the work of a move, and it keeps the convergence gate scheduled (see 3). The algorithm-level names are already unambiguous — `flow_grpo` is unique repo-wide after the rename in this PR.

If it's scope — core doesn't want this algorithm line at all — then the move is the right call and I'll do it.

**2. Where will the dLLM work land?**

The design doc isn't in-tree (nothing under `docs/design-docs/`, nothing in the `features.md` roadmap), so I'm guessing. My read is that a dLLM in NeMo-RL is an LM policy — it'd want `nemo_rl/models/policy/` and `nemo_rl/models/generation/` (block-diffusion generate in place of vLLM) and the existing `automodel`/`mcore` extras. This PR's `nemo_rl/models/diffusion/` holds SDE math, a diffusers pipeline adapter and LoRA config — no code overlap. Is that right?

**3. The convergence gate loses its schedule.**

`research/` has auto-discovery for unit tests (`L0_Unit_Tests_Other.sh` scans `research/*/tests/unit`) and functional tests (`L1_Functional_Tests_Other_1.sh` scans `research/*/tests/functional/*.sh`), but not for nightlies. `test_all_test_scripts_accounted_for_in_test_suites` asserts set equality between the suite lists and `glob(tests/test_suites/**/*.sh)`, so putting a `research/` path in `nightly.txt` fails that test, and leaving it out means the OCR gate (1x8, 60 steps, `val/reward_mean` +0.03) never runs on a schedule. `research/template_project/tests/test_suites/llm/single_update_1n8g.sh` is already in exactly that state — present, in no list.

That gate is the only objective evidence this PR has. Options: accept manual triggering, or let me add a general `nightly`-scans-`research/*/tests/test_suites` patch to core (which would also adopt the orphaned template script). If the latter, should it be a separate PR, and who owns it?

**4. Where does the `diffusion` extra go?**

Leaving it in the root `pyproject.toml` means core's surface doesn't actually shrink. Pushing it into a research member's dependencies has no precedent in-tree, and `nemo_rl/utils/venvs.py` runs an unconditional `uv sync --directory <git_root>` (venvs.py:97) before the actor's `py_executable`. Has `uv run --locked --directory research/<proj>` been exercised here?

---

For reference, the target layout I'd use is `research/flow_grpo/` with a flat package (dropping the `models/` and `workers/` shells): 31 files moved, ~8 new, 14 core files reverted, 71 import references across 20 files rewritten. Actor registration is not a problem — `research/template_project/single_update.py` already shows the pattern of inserting into `ACTOR_ENVIRONMENT_REGISTRY` from the driver.

I'd rather not fold this into the current PR: it invalidates every permalink and review thread here, and buries the fixes from this round. Once you've answered 1 and 3 I'll open a dedicated PR.

sajadn:
1. The motivation for moving this to `research/` is the different generation architecture. Rollouts run through Automodel/Diffusers inside the policy worker, sharing the training model and eliminating the refit step. That simplifies this implementation, but differs from NeMo-RL’s existing separation between policy training and generation backends.
We also want contributors to explore alternative architectures without requiring every contribution to follow the current core design. We’d therefore like to introduce this approach in `research/` first, gather customer feedback, and establish its integration and support requirements before considering it for core. This is unrelated to reserving the diffusion namespace and doesn’t rule out future core support.
2. Yes, we don’t expect direct overlap between your image-diffusion components and the planned dLLM implementation.
3. That’s a valid concern. I’ll work with the CI team to support research projects’ test_suites, including extending both the script-accounting and recipe-accounting checks to recognize research paths. We're even okay merging the code without the convergence tests and have them as a fast follow.
4. NeMo-RL already supports research-local pyproject.toml files through uv workspaces. Please follow research/template_project (https://github.com/NVIDIA-NeMo/RL/tree/3056d5db32fbe8cbc72447bee79dd7f7859838c4/research/template_project): declare the project’s dependencies in research/flow_grpo/pyproject.toml, add "research/flow_grpo" to the root workspace members, and regenerate the shared root uv.lock. Please review the lockfile diff for unintended changes to existing dependencies. For the policy worker, register uv run --locked --directory <absolute-research-project-path> directly in ACTOR_ENVIRONMENT_REGISTRY from the research entrypoint. No command-line extras are needed if the project declares those dependencies. Your concern about the root-first sync is valid. It would remain, followed by the research-specific command targeting the same worker environment. I'm validating that path and we'll keep you posted.

sajadn:
i think it'll be simpler if i PR into your branch so you can review. I'll open it soon.

triple-mu:
> think it'll be simpler if i PR into your branch so you can review. I'll open it soon.

Thanks a lot for the advice! I’m also refactoring the code right now — would love to team up if you have any relevant PRs.

sajadn:
@triple-mu I’ve opened https://github.com/triple-mu/RL/pull/1 against your branch with the research-folder refactor, worker fixes, and checkpoint-loading fix. The shared CI changes are handled separately in #4216. Please review and if it looks good to you please merge it into your branch. I ran the nightly config and it seems to be working well:
Screenshot 2026-09-20 at 6 15 38 PM

triple-mu:
Merged triple-mu/RL#1 into the branch as a fast-forward, thanks for the fixes; the resume path had never been exercised on my side. One follow-up commit on top: the research entries added to the root `pyrefly.toml` made the root `pyrefly check` abort ("matched by project-excludes"), because pyrefly treats the nested `research/flow_grpo/pyrefly.toml` as its own project and excludes the directory from the parent. I removed them and let the project config cover those files (0 errors); the same commit sorts one import block ruff's isort flagged.

On 4×B200 with the merged branch: `uv lock --check`, ruff, the project pyrefly check, the 107 research unit tests and the functional smoke all pass. The two accounting tests will stay red until #4216 lands; I'll rebase as soon as it does. Whenever you are ready, could you trigger `/ok to test` on the current head?

tests/test_suites/nightly.txt
```diff
@@ -197,6 +197,13 @@ tests/test_suites/llm/grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron_generation-n
 tests/test_suites/llm/grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron_generation-single-controller-async-lag4.sh
 tests/test_suites/llm/ppo-qwen2.5-1.5b-gsm8k-2n8g-megatron-valuetp2sp-dynbatch-noncolocated-async-single-controller.sh
 
+#############
+# Diffusion #
+#############
+
+# Qwen-Image flow-GRPO LoRA on the Flow-GRPO OCR task (PaddleOCR reward, single-node 8-GPU DP)
+research/flow_grpo/tests/test_suites/flow_grpo-qwen-image-ocr-1n8g-dp8-lora.sh
```
terrykong, line 205: [`nightly.txt`](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/tests/test_suites/nightly.txt#L205) — line 205

**1 action item.**

**TL;DR — two root unit tests fail on this branch as-is, and #4216 (the PR that teaches them about research paths) has not merged yet, so this one must not land first.**

PR-introduced. The two `research/...` entries added here and in [`disabled.txt`](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/tests/test_suites/disabled.txt#L93) have no match in the accounting logic, and this PR does not change that logic:

- [`test_all_test_scripts_accounted_for_in_test_suites`](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/tests/unit/test_recipes_and_test_suites.py#L242) builds its on-disk set from a glob that [globs only `tests/test_suites`](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/tests/unit/test_recipes_and_test_suites.py#L246), so neither research path is in it and the set comparison fails.
- [`test_all_recipe_yamls_accounted_for_in_test_suites`](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/tests/unit/test_recipes_and_test_suites.py#L258) opens with [this length assert](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/tests/unit/test_recipes_and_test_suites.py#L262), which is now off by two.

[`test_nightly_compute_stays_below_4700_hours`](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/tests/unit/test_recipes_and_test_suites.py#L287) needs no action: [`tools/launch`](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/tools/launch#L105) only checks that the path is a file before reading the CONFIG block, so a `research/...` path works there. At [`NUM_NODES=1` / `NUM_MINUTES=360`](https://github.com/NVIDIA-NeMo/RL/blob/30601a13886560c922e6f64866fcbab1d388523a/research/flow_grpo/tests/test_suites/flow_grpo-qwen-image-ocr-1n8g-dp8-lora.sh#L6) the recipe costs 48 GPU-hours, well inside the 4640 to 4700 bump.

### AI-1

Wait for [#4216](https://github.com/NVIDIA-NeMo/RL/pull/4216) to merge, then rebase this branch onto it before merging this one. #4216 is still open as of this review. Not a suggestion block: this is a merge-ordering ask, not a code edit.

terrykong (commented):
Light review, aimed at the changes that reach NeMo-RL core rather than at the research project itself.

**Core surface is clean.** `uv.lock` is a pure addition, 346 insertions and zero deletions, so nothing the root `nemo-rl` package or its extras already resolve to has moved. Every added package comes from `pypi.org`, and the only non-registry entry is `flow-grpo` itself as an editable workspace member. `research/flow_grpo/pyproject.toml` adds no custom index and only `nemo-rl = { workspace = true }`. The new `research/*/tests/**` ruff ignore cannot match the root `tests/` tree.

**Nothing in the new code reaches outside its own project.** The research tests and one nightly recipe run in CI, so the blast radius got a pass: no `eval`, `exec`, `os.system`, `subprocess`, `pickle.loads`, `torch.load` or `yaml.load`; no `conftest.py` or `sitecustomize`; no CI-config change. The only two outbound hosts are `raw.githubusercontent.com` for the OCR prompt list, which is plain text that gets parsed rather than executed, and the user-configured `GENRM_BASE_URL` judge. `flow_grpo/actor_environments.py` writes only its own two `flow_grpo.`-namespaced keys into `ACTOR_ENVIRONMENT_REGISTRY`, the same pattern `research/template_project/single_update.py` already uses. The shell scripts stay inside the project and its experiment directory.

Thanks for the detailed PR description, and for re-scoring the trained checkpoint with PaddleOCR and PickScore. Using two scorers the model never trained against is the right way to show the reward is not just being gamed.

One inline comment, on merge ordering.

_Generated by Claude Code_

terrykong:
lgtm @triple-mu ! thanks for your contribution. the review left is simply a placeholder that we need to wait for @sajadn's #4216 to land first and for this to rebase on top

terrykong (approved):

sajadn (approved):

PR #3417, opened 2026-07-29
feat: add NuminaMath-1.5 response dataset

### LINKED ISSUE #3416: Proposal: add NuminaMath-1.5 as a built-in verifiable math response dataset
## Summary

Propose adding `AI-MO/NuminaMath-1.5` as a built-in response dataset for verifiable-answer math training. It's one of the standard large math corpora and isn't in the registry today; the closest existing entries are GSM8K, DeepScaler, DAPO-Math-17K, OpenMathInstruct-2 and OpenR1-Math-220k (#3131).

## Why it needs a wrapper rather than `ResponseDataset`

Unlike OpenR1-Math-220k, this one can't be mapped straight through: roughly a quarter of the rows have no checkable answer, so a naive `{problem, answer}` mapping would feed `"proof"` / `"notfound"` strings to the math verifier as ground truth. The dataset ships explicit type/quality columns to filter on:

| filter stage | rows remaining |
|---|---|
| all rows | 896,215 |
| `answer` not in `{proof, notfound}` | 752,915 |
| + `question_type != "proof"` | 735,559 |
| + `problem_is_valid == "Yes"` and `solution_is_valid == "Yes"` | **682,710** |

The two proof filters are not redundant — `question_type == "proof"` (110,998 rows) and `answer == "proof"` (81,042 rows) only partially overlap, so filtering on `answer` alone still leaves 17,356 proof-type rows.

For reference, the other column distributions:
- `question_type`: math-word-problem 631,522 / MCQ 146,449 / proof 110,998 / NaN 5,216 / other 2,030
- `problem_is_valid`: Yes 859,356 / Incomplete 30,377 / NaN 5,219 / "More than one problem" 1,105 / "Not a problem" 158
- `solution_is_valid`: Yes 844,988 / Incomplete 32,522 / "Problem not solved" 13,446 / NaN 5,227 / "Not matched with problem" 32

## Proposed implementation

A `RawDataset` wrapper following the GSM8K / OpenR1-Math-220k pattern — emits the same two-message `{problem, answer}` shape and reuses the existing `math_hf_data_processor` / `MathEnvironment` path, no new reward logic — plus:

- `verifiable_only: bool = True` — drop the sentinel answers and `question_type == "proof"`
- `require_valid: bool = True` — keep only rows where both validity flags are `"Yes"`
- `split_validation_size` / `seed` support (calls `split_train_validation`)
- one `DATASET_REGISTRY` entry and unit tests covering the filter behavior

## Questions

1. Is a built-in registry addition welcome here, or would you rather this stay a dotted-import (#2642) dataset given its size?
2. Do the default filters look right? Defaulting both to `True` is the conservative choice (a user asking for math RL gets verifiable rows), but it does mean the default silently drops ~24% of the dataset, so I'm happy to flip either default or expose `question_type` selection (e.g. to include/exclude MCQ) instead.

Happy to send the PR.

cc @yuki-97 @RayenTian

tianyi-zhang-02:
## What does this PR do ?

Adds `AI-MO/NuminaMath-1.5` as a built-in response dataset for verifiable-answer math training. It's one of the standard large competition-math corpora (896,215 rows) and isn't in the registry today.

Unlike OpenR1-Math-220k (NVIDIA-NeMo/RL#3131), this one can't be a plain pass-through wrapper: about a quarter of the rows have no closed-form answer, and the dataset encodes that with literal sentinel strings in `answer`, which would otherwise reach the math verifier as ground truth.

| filter stage | rows remaining |
| -- | -- |
| all rows | 896,215 |
| `answer` not in `{proof, notfound}` | 752,915 |
| * `question_type != "proof"` | 735,559 |
| * `problem_is_valid == "Yes"` and `solution_is_valid == "Yes"` | **682,710** (defaults) |

Both proof checks are needed — `question_type == "proof"` (110,998) and `answer == "proof"` (81,042) only partially overlap, so a row can be tagged `"math-word-problem"` and still carry `answer="proof"` (verified against the live dataset).

## Changes

* `NuminaMath15Dataset` with `verifiable_only` and `require_valid` (both default `True`, both configurable for callers who want the unfiltered corpus), plus `split_validation_size` / `seed` support.
* Registered as `NuminaMath-1.5` in the response `DATASET_REGISTRY`.
* Emits the same two-message `{problem, answer}` shape as the sibling math datasets and reuses the existing `math_hf_data_processor` / `MathEnvironment` path — no new reward logic.
* Added to the built-in dataset examples in the SFT/GRPO guides, and to `pyrefly.toml` `project-includes` (the file type-checks clean). Deliberately not added to the `split_validation_size` enumeration in those guides, because NVIDIA-NeMo/RL#3389 replaces that enumeration with the rule that determines support — avoiding a conflict between the two PRs.

## Issues

Closes NVIDIA-NeMo/RL#3416.

## Usage

```yaml
data:
  train:
    dataset_name: NuminaMath-1.5
    split_validation_size: 0.01
    # verifiable_only: true   # default; drops proof/notfound answers and proof-typed rows
    # require_valid: true     # default; keeps only rows flagged valid by the dataset
```

- [X] New tests, in two layers:
  - **Real data**, matching the OpenR1-Math-220k precedent from NVIDIA-NeMo/RL#3131: `NuminaMath-1.5` added to the parametrized `test_build_in_dataset_with_split_validation` with its own assertion branch, covering registry dispatch, `split_validation_size` and the emitted schema. The assertion checks the property this loader exists for — no sentinel answer survives the default filters. (Note this adds one more real dataset load to that test; happy to drop the entry and rely on the mocked tests if the CI time is unwelcome.)
  - **Mocked**, in `tests/unit/data/datasets/test_numinamath_dataset.py`: one source row per exclusion reason (sentinel `proof`, sentinel `notfound`, `question_type="proof"`, invalid problem, invalid solution, empty answer) plus two that must survive; covers the defaults, each flag independently, the emitted schema and the validation split. Kept as a separate file because the filter behavior needs controlled source rows, which the real-data test cannot express.
- [X] Validation: all cases verified locally against the real loader; `ruff check` / `format` clean; `pyrefly==0.24.2` reports 0 errors.
- [X] Docs: updated both guides' val-split lists.

## Open question

The defaults drop \~24% of the corpus. I've defaulted both filters to `True` on the grounds that a user selecting a math dataset for RL wants verifiable rows, but I'm happy to flip either default, or to expose `question_type` selection (e.g. to include/exclude the 146,449 MCQ rows) instead — whichever you prefer.

cc @yuki-97 @RayenTian


yuki-97 (commented):
@RayenTian to review

RayenTian:
LGTM. Would you please update this branch then we can trigger CI.

tianyi-zhang-02:
Updated with main (`b01c00f`) — clean merge, submodule pointers match. Ready for CI, thanks @RayenTian!

RayenTian:
/ok to test [b01c00f](https://github.com/NVIDIA-NeMo/RL/pull/3417/commits/b01c00f8fa0b68c51fc762fcddc32e449822055e)

yuki-97 (approved):
LGTM, thanks @tianyi-zhang-02

PR #3135, opened 2026-07-09
perf: perf recipe changes to enable force on policy ratio

seonjinn:
# What does this PR do ?

This PR updates GB200/H100 GRPO performance recipes for LLaMa-3 8B, Qwen3-32B, Qwen3-30BA3B:                                                                                                                                                                                           
                                                                                                                                                                                                                                                
  - Enable loss_fn.force_on_policy_ratio: true.                                                                                                                                                                                                 
  - Set policy.train_global_batch_size: 2048 to align policy training with the generated batch size.                                                                                                                                            
  - Cover Llama 3.1 8B, Qwen3-30B-A3B, and Qwen3-32B in both Sync and Async-1off modes.

Across GB200 and H100, enabling force-on-policy ratio:

- reduces logprob time by **41.6%–50.1%** and increases logprob throughput by **70.4%–100.5%**;
- improves E2E throughput by **7.3%–20.9%** in all six GB200 cases and by **5.5%–18.8%** in five of six H100 cases;
- leaves policy-training performance largely unchanged, generally within **±3%**;


## GB200 results 

### Benchmark setup
- 20 training steps per run
- Reported values: arithmetic mean of steps 3–8 inclusive

The GB200 comparison includes both the GBS increase and force-on-policy ratio; it is not a same-GBS isolation.

### E2E impact

| Model | Mode | Step time (s) | Delta | E2E tok/s/GPU | Delta |
|---|---|---:|---:|---:|---:|
| Llama 3.1 8B | Sync | 68.202 → 63.307 | -7.18% | 4,335 → 4,649 | +7.25% |
| Llama 3.1 8B | Async-1off | 61.422 → 53.034 | -13.66% | 5,089 → 5,892 | +15.77% |
| Qwen3-30B-A3B | Sync | 196.702 → 178.055 | -9.48% | 2,115 → 2,340 | +10.62% |
| Qwen3-30B-A3B | Async-1off | 271.973 → 231.400 | -14.92% | 1,531 → 1,800 | +17.56% |
| Qwen3-32B | Sync | 310.435 → 277.021 | -10.76% | 1,360 → 1,525 | +12.13% |
| Qwen3-32B | Async-1off | 161.853 → 133.918 | -17.26% | 1,306 → 1,580 | +20.94% |

### Logprob impact

| Model | Mode | Logprob time (s) | Delta | Logprob tok/s/GPU | Delta |
|---|---|---:|---:|---:|---:|
[...]

PR #3419, opened 2026-07-29
fix: prevent PP hang in MoE aux-loss metric reduction

dafu-wu:
# What does this PR do ?

Fixes a **silent, no-traceback hang** in `get_moe_metrics()` when training a
MoE model with `moe_router_load_balancing_type=aux_loss` and
`pipeline_model_parallel_size > 1`.

## Symptom

With `aux_loss` load balancing enabled and PP > 1, training progresses for a
number of steps and then hangs: no exception, no OOM, no exit — GPUs go idle
and the process never returns. The same recipe with
`moe_router_load_balancing_type=none` runs indefinitely without issue.

## Root cause

`get_moe_metrics()` calls `reduce_aux_losses_tracker_across_ranks()`, which runs
`torch.distributed.all_reduce(values, group=pp_group)` **for each name present
in the local MoE logging tracker**:

```python
# megatron/core/transformer/moe/moe_utils.py
def reduce_aux_losses_tracker_across_ranks(...):
    tracker = get_moe_layer_wise_logging_tracker()
    if track_names is None:
        track_names = tracker.keys()          # <-- only the LOCAL rank's keys
    ...
    for name in track_names:
        values = tracker[name]["values"]
        torch.distributed.all_reduce(values, group=pp_group)   # <-- collective
```

The tracker entry is created **lazily** — `save_to_aux_losses_tracker` only
allocates `torch.zeros(num_layers)` the first time a rank actually saves a loss.
So if some PP rank did **not** save an aux loss this step — e.g. a pipeline
stage that holds no MoE layer, or an MTP MoE layer that lives only on the last
stage — that rank has no key for `load_balancing_loss`, skips the `all_reduce`
for that name, while other PP ranks perform it. The collective then mismatches
its participants across the PP group and hangs.

This is exactly the situation Megatron's own `track_moe_metrics` guards against
via `force_initialize=True`; NeMo-RL calls `reduce_aux_losses_tracker_across_ranks()`
directly and bypassed that guard.

## Fix

Mirror the `force_initialize=True` behaviour inside `get_moe_metrics()`:
before the reduction, pre-initialize the tracker on **every** rank so each
tracked name exists with an equally-sized zero tensor of length
`num_layers + mtp_num_layers` — the same size the router uses in
`save_to_aux_losses_tracker`. `num_layers` / `mtp_num_layers` are threaded from
`model_config` in both the policy and value Megatron workers.

When `num_layers is None` (all existing/other call sites), the behaviour is
completely unchanged — this is a no-op guarded by an opt-in argument.

## Usage

No config change required. The guard activates automatically for MoE models
(`num_moe_experts > 1`) in the Megatron policy/value workers, which now pass
`num_layers` / `mtp_num_layers` from `model_config`. After this change,
`moe_router_load_balancing_type=aux_loss` with PP > 1 no longer hangs.

## Files

- `nemo_rl/models/megatron/common.py` — pre-initialize tracker in `get_moe_metrics()`
- `nemo_rl/models/policy/workers/megatron_policy_worker.py` — pass `num_layers`/`mtp_num_layers`
- `nemo_rl/models/value/workers/megatron_value_worker.py` — pass `num_layers`/`mtp_num_layers`

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Does the PR affect components that other people are working on?

## Additional Information

Repro parallelism where the hang was observed: `PP=4, TP=8, EP=8` on a
Nemotron-3-style 120B MoE SFT run; also reproducible with MTP enabled (the MTP
MoE layer lives only on the last PP stage, guaranteeing the tracker-key
asymmetry across the PP group).


terrykong (commented):
@yfw could you review?

dafu-wu:
@yfw  can you help review it？

nemo_rl/models/megatron/common.py
```diff
@@ -141,7 +144,40 @@ def get_moe_metrics(
...
+        tracker_num_layers = num_layers + (mtp_num_layers or 0)
+        tracker = get_moe_layer_wise_logging_tracker()
+        for name in track_names:
+            if name not in tracker:
+                tracker[name] = {
+                    "values": torch.zeros(tracker_num_layers, device="cuda"),
+                    "reduce_group": None,
+                    "avg_group": None,
+                    "reduce_group_has_dp": False,
+                }
```
yfw: [`common.py:171-179`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/common.py#L171-L179)

**This block doesn't reach Megatron's tracker on the pinned Megatron-Core, so the hang isn't prevented.**

At mcore 0.19.0, [`get_moe_layer_wise_logging_tracker()`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_utils.py#L1028-L1039) is a deprecated shim that **builds and returns a brand-new dict on every call**:

```python
@deprecated(version="0.16", removal_version="0.18", alternative="get_moe_metrics_tracker().metrics")
def get_moe_layer_wise_logging_tracker():
    return {name: {"values": entry.values, ...} for name, entry in get_moe_metrics_tracker().metrics.items()}
```

So `tracker[name] = ...` lands in a temporary. [`reduce_aux_losses_tracker_across_ranks()`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_utils.py#L1012-L1025) then resolves names from the *real* `get_moe_metrics_tracker().metrics`, which still lacks the key — [`_sync_metrics` skips it](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L265-L272) before reaching the `all_reduce`.

To be clear, **the code is correct for the Megatron-Core you were almost certainly building against** — I checked the history rather than guessing. Before [`873678ada` "Refactor and Improve MoE Logging"](https://github.com/NVIDIA/Megatron-LM/commit/873678adafa71702ef7e7a405a7f4a3377dd2727) (2026-05-27, first released in `core_v0.18.0`), the accessor returned the live module global:

```python
def get_moe_layer_wise_logging_tracker():
    global _MOE_LAYER_WISE_LOGGING_TRACKER
    return _MOE_LAYER_WISE_LOGGING_TRACKER      # live handle
```
([pre-refactor source](https://github.com/NVIDIA/Megatron-LM/blob/7521ecbeaf6a2dc9312303135f2572a7250723da/megatron/core/transformer/moe/moe_utils.py#L1221-L1224))

and [`save_to_aux_losses_tracker`](https://github.com/NVIDIA/Megatron-LM/blob/7521ecbeaf6a2dc9312303135f2572a7250723da/megatron/core/transformer/moe/moe_utils.py#L996-L1003) built entries with exactly the shape you write, `"reduce_group_has_dp"` included. The pre-refactor reducer also did `track_names = tracker.keys()`, matching the snippet in your PR description. So this is purely a version-boundary issue, not a misreading on your part:

| megatron-core | this patch |
|---|---|
| ≤ 0.17.x | works as written |
| ≥ 0.18.0 (incl. this repo's 0.19.0 pin) | inert |

Repro on the pinned version (CPU-only, no GPUs needed):
```python
import torch
from megatron.core.transformer.moe.moe_utils import get_moe_layer_wise_logging_tracker
from megatron.core.transformer.moe.moe_logging import get_moe_metrics_tracker

t = get_moe_layer_wise_logging_tracker()
t["load_balancing_loss"] = {"values": torch.zeros(4), "reduce_group": None,
                            "avg_group": None, "reduce_group_has_dp": False}
print(list(get_moe_metrics_tracker().metrics.keys()))   # -> []   (injection lost)
```

mcore added a public primitive for precisely this case in the same refactor: [`MoEMetricsTracker.ensure_initialized`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L219-L235), which is what `report(force_initialize=True)` [itself calls](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L181-L186). It uses the same `num_layers + (mtp_num_layers or 0)` sizing you derived, and picks the device itself (with a CPU fallback), which also removes the hardcoded `device="cuda"`.

```suggestion
        mcore_tracker = get_moe_metrics_tracker()
        for name in track_names:
            mcore_tracker.ensure_initialized(name, tracker_num_layers)
```

This needs one new import at the top of the file:
```python
from megatron.core.transformer.moe.moe_logging import get_moe_metrics_tracker
```

Could you confirm which Megatron-Core version the PP=4/TP=8/EP=8 run used? If it was ≤0.17.x that fully explains the gap, and this becomes a straightforward port rather than a rethink.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -914,6 +914,11 @@ def train(
             moe_metrics = get_moe_metrics(
                 loss_scale=moe_loss_scale,
                 per_layer_logging=self.cfg["megatron_cfg"]["moe_per_layer_logging"],
+                # Pre-initialize the aux-loss tracker on every PP rank so the
+                # cross-PP all_reduce inside get_moe_metrics does not hang when a
+                # rank saved no aux loss this step (e.g. MTP MoE on the last stage).
+                num_layers=getattr(model_config, "num_layers", None),
+                mtp_num_layers=getattr(model_config, "mtp_num_layers", None),
```
yfw, line 1088: [`megatron_policy_worker.py:920-921`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/policy/workers/megatron_policy_worker.py#L920-L921)

There's a **third** `get_moe_metrics()` call site that didn't get these two kwargs — [`megatron_policy_worker.py:1462`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1462-L1465), inside `_finish_train_step_body`:

```python
moe_metrics = get_moe_metrics(
    loss_scale=moe_loss_scale,
    per_layer_logging=self.cfg["megatron_cfg"]["moe_per_layer_logging"],
)
```

It sits inside an identical `num_moe_experts > 1` guard and already has `model_config` in scope [four lines above](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1458), so it reads as an oversight rather than a deliberate exclusion. It's reachable through the public `finish_train_step()` on the single-controller / split-API path, with no MoE or PP gating along the way.

Worth flagging that this path also has **no MoE coverage in CI** — both single-controller entries in [`nightly.txt`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/tests/test_suites/nightly.txt#L163-L165) run dense models (Llama-3.1-8B and Qwen2.5-Math-1.5B), so the `num_moe_experts > 1` branch never fires there. A hang on this path wouldn't be caught by the nightly suite.

Same two lines should do it:
```python
                num_layers=getattr(model_config, "num_layers", None),
                mtp_num_layers=getattr(model_config, "mtp_num_layers", None),
```

(The policy and value worker sites you did update are nicely symmetric with each other — no issue there.)

nemo_rl/models/megatron/common.py
```diff
@@ -141,7 +144,40 @@ def get_moe_metrics(
...
+    # allocates torch.zeros(num_layers) the first time a rank saves a loss). If any PP
+    # rank did not save an aux loss this step, it skips the all_reduce for that name
+    # while other PP ranks perform it -> the collective mismatches participants and hangs.
+    #
+    # Mirror Megatron's own track_moe_metrics(force_initialize=True) guard: allocate a
+    # zero tensor of size (num_layers + mtp_num_layers) for each tracked name on every
+    # rank, matching the size the router uses in save_to_aux_losses_tracker.
+    if num_layers is not None:
+        if track_names is None:
+            track_names = ["load_balancing_loss"]
```
yfw: [`common.py:168-169`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/common.py#L168-L169)

Two things worth considering here:

1. **`track_names` is never passed by any caller.** Neither worker sets it, so the new parameter is currently dead and this default is always what runs.
2. **The default covers one of three reachable aux-loss modes.** The router records separate keys, each driving its own `all_reduce`: [`load_balancing_loss`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L323), [`seq_load_balancing_loss`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L378), and [`global_load_balancing_loss`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L423). [`is_aux_loss_enabled()`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L282-L287) treats all three as first-class, and `routing_type` may even be a **list**, so more than one can be live at once.

`moe_router_load_balancing_type` is a raw passthrough in NeMo-RL ([`setup.py:748`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/setup.py#L748)) with no allowlist, and this repo's own configs name `seq_aux_loss` in their comments (e.g. [`grpo_math_1B_megatron.yaml:111`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/examples/configs/grpo_math_1B_megatron.yaml#L111)) — so it's a mode users reach for. As written, someone on `seq_aux_loss` with PP>1 would still hang, with a fix in the tree that looks like it should have covered them.

Both Megatron ([`training.py:2688-2696`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/training/training.py#L2688-L2696)) and Megatron-Bridge ([`train_utils.py:1029-1042`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/a056408d29ad1070fb926512991075998c9e023f/src/megatron/bridge/training/utils/train_utils.py#L1029-L1042)) derive the list from config instead:

```python
track_names = []
if "aux_loss" in moe_router_load_balancing_type:        track_names.append("load_balancing_loss")
if "seq_aux_loss" in moe_router_load_balancing_type:    track_names.append("seq_load_balancing_loss")
if "global_aux_loss" in moe_router_load_balancing_type: track_names.append("global_load_balancing_loss")
```

Deriving it the same way covers all three modes — and it also neatly handles the `"none"` case in the comment below.

(`z_loss` is the fourth name mcore records, but `moe_z_loss_coeff` isn't currently plumbed through NeMo-RL, so it's latent rather than reachable today.)

nemo_rl/models/megatron/common.py
```diff
@@ -141,7 +144,40 @@ def get_moe_metrics(
...
+    # the pipeline-parallel group for each name present in the *local* tracker. The
+    # tracker entry is created lazily (Megatron save_to_aux_losses_tracker only
+    # allocates torch.zeros(num_layers) the first time a rank saves a loss). If any PP
+    # rank did not save an aux loss this step, it skips the all_reduce for that name
+    # while other PP ranks perform it -> the collective mismatches participants and hangs.
+    #
+    # Mirror Megatron's own track_moe_metrics(force_initialize=True) guard: allocate a
+    # zero tensor of size (num_layers + mtp_num_layers) for each tracked name on every
+    # rank, matching the size the router uses in save_to_aux_losses_tracker.
+    if num_layers is not None:
```
yfw: [`common.py:167`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/common.py#L167)

A knock-on effect to consider once the mechanism is working: this gates on `num_layers is not None`, but the workers only check `num_moe_experts > 1` — not whether load balancing is actually enabled.

With `moe_router_load_balancing_type: "none"` the router [returns before recording](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L296-L298) (`if aux_loss_coeff == 0: return probs`), so today the tracker stays empty and `get_moe_metrics()` returns `{}` — which is falsy, so [the `if moe_metrics:` guard](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/policy/workers/megatron_policy_worker.py#L922) drops it and no `moe_metrics` key is emitted at all.

With an effective pre-init, that becomes `{"load_balancing_loss": 0.0}` — truthy — so a permanently-zero series would start appearing in the logged metrics for runs that never had one. `"none"` is the default ([`setup.py:1873`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/setup.py#L1873)) and the value in every MoE config in the repo, so this would affect all of them rather than an edge case.

Gating the pre-init on the configured balancing type — i.e. the same config-derived `track_names` from the comment above, which comes out empty under `"none"` — resolves both points in one change.

nemo_rl/models/megatron/common.py
```diff
@@ -141,7 +144,40 @@ def get_moe_metrics(
...
+        if track_names is None:
+            track_names = ["load_balancing_loss"]
+        tracker_num_layers = num_layers + (mtp_num_layers or 0)
```
yfw, line 254: [`common.py:170`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/common.py#L170)

Can you add a unit test here? [`tests/unit/models/megatron/test_moe_metrics.py`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/tests/unit/models/megatron/test_moe_metrics.py) already exists and runs CPU-only, so it's cheap — and it would have caught the tracker issue above.

One trap worth flagging: both existing tests stub the accessor with a **persistent** dict ([line 89](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/tests/unit/models/megatron/test_moe_metrics.py#L89), `lambda: tracker`). A plain dict *is* live, so a test written in that house style would pass while production stays a no-op. The test has to assert against the real `get_moe_metrics_tracker().metrics` and must **not** stub `get_moe_layer_wise_logging_tracker`.

This version was verified to **fail on this branch and pass with the suggested fix**:

```python
@pytest.mark.mcore
def test_preinit_registers_entry_on_live_tracker(monkeypatch):
    """A PP rank that saved no aux loss must still enter the collective."""
    from megatron.core.transformer.moe.moe_logging import (
        destroy_moe_metrics_tracker,
        get_moe_metrics_tracker,
    )
    from nemo_rl.models.megatron import common as common_mod
    from nemo_rl.models.megatron.common import get_moe_metrics

    destroy_moe_metrics_tracker()
    try:
        monkeypatch.setattr(
            common_mod, "reduce_aux_losses_tracker_across_ranks", lambda *a, **k: None
        )
        get_moe_metrics(loss_scale=1.0, num_layers=4, mtp_num_layers=1)

        live = get_moe_metrics_tracker().metrics
        assert set(live) == {"load_balancing_loss"}
        assert live["load_balancing_loss"].values.numel() == 5  # num_layers + mtp
    finally:
        destroy_moe_metrics_tracker()
```

Run with: `uv run --group test pytest tests/unit/models/megatron/test_moe_metrics.py --mcore-only -q`

nemo_rl/models/megatron/common.py
```diff
@@ -125,6 +125,9 @@ def get_moe_metrics(
     loss_scale: float,
     total_loss_dict: Optional[dict] = None,
     per_layer_logging: bool = False,
+    num_layers: Optional[int] = None,
+    mtp_num_layers: Optional[int] = None,
+    track_names: Optional[list[str]] = None,
```
yfw, line 203: [`common.py:128-130`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/common.py#L128-L130)

Two small things about the new parameters:

**1. They're missing from the `Args:` block.** [Lines 137-141](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/common.py#L137-L141) still document only the original three; the new ones are described in the trailing `Note:` instead. The sibling `get_mtp_metrics()` at [line 206](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/common.py#L206) is the in-file convention — it documents semantics *and* the default inside each entry. Something like:

```
        num_layers: Total number of transformer layers. When provided, the aux-loss
            tracker is pre-initialized on every rank before the reduction (see Note).
            Defaults to None, which disables pre-initialization.
        mtp_num_layers: Extra layers contributed by Multi-Token Prediction, added to
            ``num_layers`` to size the pre-initialized tensor. Defaults to None.
        track_names: Aux-loss names to pre-initialize; should mirror what the router
            records for the configured ``moe_router_load_balancing_type``.
            Defaults to None, which pre-initializes only "load_balancing_loss".
```

**2. Minor: `num_layers` is now shadowed later in the function.** [Line 189](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/nemo_rl/models/megatron/common.py#L189) does `num_layers = int(loss_list.numel()) ...`, which predates this PR but becomes a rebinding now that a parameter shares the name — and the two mean different things (config layer count vs. per-tensor element count). It's genuinely harmless as written (the parameter is only read at line 167, before the loop), just a tripwire for the next edit. Renaming the loop-local to `num_tracked_layers` would clear it.

yfw (commented):
Thanks for digging into this — the failure mode you describe is real. `force_initialize=True` is exactly the right precedent to reach for.

The main thing I'd like to work through with you: against the Megatron-Core this repo currently pins (**0.19.0**, via `3rdparty/Megatron-Bridge-workspace/Megatron-Bridge/3rdparty/Megatron-LM` @ `cf2f07d`, per [`uv.lock`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/uv.lock#L3331)), the pre-initialization doesn't reach Megatron's tracker — so the hang would still occur. Details in the first inline comment, including a short CPU-only script you can run to check it on your side.

This is a version boundary, not a misreading on your part — I checked the upstream history rather than assuming. Your patch is correct against megatron-core ≤0.17.x, where that accessor returned the live global dict and the entry shape (including `reduce_group_has_dp`) matched exactly what you wrote. The [MoE-logging refactor](https://github.com/NVIDIA/Megatron-LM/commit/873678adafa71702ef7e7a405a7f4a3377dd2727) landed in `core_v0.18.0` and turned that accessor into a copy. So this needs a port to the newer API, not a rethink — the diagnosis itself holds up.

For calibration, the preconditions for the hang look like: an MoE model, **and** a non-zero aux-loss balancing type, **and** PP>1, **and** a layout where some PP stage owns no aux-loss-recording MoE layer. No config shipped in this repo combines those (the `aux_loss` exemplars are PP=1; the MoE+PP ones use `"none"`), so this is a user-config-reachable bug — consistent with how you found it.

Reviewed with a team of specialized agents; the mechanism finding was independently reproduced by execution three times.

_Generated by Claude Code_

dafu-wu:
> Thanks for digging into this — the failure mode you describe is real. `force_initialize=True` is exactly the right precedent to reach for.
> 
> The main thing I'd like to work through with you: against the Megatron-Core this repo currently pins (**0.19.0**, via `3rdparty/Megatron-Bridge-workspace/Megatron-Bridge/3rdparty/Megatron-LM` @ `cf2f07d`, per [`uv.lock`](https://github.com/NVIDIA-NeMo/RL/blob/ddda2fde080daaef6aa641d6504e10ebdad2bd35/uv.lock#L3331)), the pre-initialization doesn't reach Megatron's tracker — so the hang would still occur. Details in the first inline comment, including a short CPU-only script you can run to check it on your side.
> 
> This is a version boundary, not a misreading on your part — I checked the upstream history rather than assuming. Your patch is correct against megatron-core ≤0.17.x, where that accessor returned the live global dict and the entry shape (including `reduce_group_has_dp`) matched exactly what you wrote. The [MoE-logging refactor](https://github.com/NVIDIA/Megatron-LM/commit/873678adafa71702ef7e7a405a7f4a3377dd2727) landed in `core_v0.18.0` and turned that accessor into a copy. So this needs a port to the newer API, not a rethink — the diagnosis itself holds up.
> 
> For calibration, the preconditions for the hang look like: an MoE model, **and** a non-zero aux-loss balancing type, **and** PP>1, **and** a layout where some PP stage owns no aux-loss-recording MoE layer. No config shipped in this repo combines those (the `aux_loss` exemplars are PP=1; the MoE+PP ones use `"none"`), so this is a user-config-reachable bug — consistent with how you found it.
> 
> Reviewed with a team of specialized agents; the mechanism finding was independently reproduced by execution three times.
> 
> _Generated by Claude Code_

@yfw Thanks for the thorough review — the mechanism finding is correct, and I've pushed `2df90ac9` addressing all six comments.

### On the megatron-core version question

The repro ran on NeMo-RL [`37cb7de1`](https://github.com/NVIDIA-NeMo/RL/commit/37cb7de1a9f19faf3cc450f3cf6da6912b8f5182) ("perf: perf recipe changes to enable force on policy ratio", #3135), which pins:

```
NeMo-RL 37cb7de1
  └─ Megatron-Bridge 554c7b93
       └─ Megatron-LM 00225507  =  megatron-core 0.19.0
```

Same minor as this PR's base (`cf2f07d`). So this **wasn't** a ≤0.17.x environment — at [`00225507`](https://github.com/NVIDIA/Megatron-LM/tree/002255075c3728fded9a2e435677840b08560d55):

- [`get_moe_layer_wise_logging_tracker()`](https://github.com/NVIDIA/Megatron-LM/blob/002255075c3728fded9a2e435677840b08560d55/megatron/core/transformer/moe/moe_utils.py#L1028-L1039) was already the copy-returning shim
- the entry field was already `needs_dp_avg`, not `reduce_group_has_dp`
- [`_sync_metrics`](https://github.com/NVIDIA/Megatron-LM/blob/002255075c3728fded9a2e435677840b08560d55/megatron/core/transformer/moe/moe_logging.py#L265-L267) already had the `if name not in self._metrics: continue` skip
- [`ensure_initialized`](https://github.com/NVIDIA/Megatron-LM/blob/002255075c3728fded9a2e435677840b08560d55/megatron/core/transformer/moe/moe_logging.py#L219-L235) was already available

Which means the original patch was **inert on the very environment where I reproduced the hang**. I had confirmed the root cause (py-spy on the stuck ranks showed the PP `all_reduce` participant mismatch) but never re-ran the 4-node repro after writing the patch, so I never validated the fix itself. Your finding is exactly right — and thanks for checking the upstream history rather than assuming.

### What changed in `2df90ac9`

| # | Comment | Fix |
|---|---------|-----|
| 1 | tracker shim | `MoEMetricsTracker.ensure_initialized()` — what `report(force_initialize=True)` calls. Also drops the hardcoded `device="cuda"` and the stale `reduce_group_has_dp` field. |
| 2 | third call site | `_finish_train_step_body` now passes the same kwargs; all three sites symmetric. |
| 3 | `track_names` | New `get_aux_loss_track_names(model_config)` covering `load_balancing_loss` / `seq_load_balancing_loss` / `global_load_balancing_loss` / `z_loss`, incl. the list form of `moe_router_load_balancing_type`. |
| 4 | zero-metric regression | Gate is now `num_layers is not None and track_names` — empty under `"none"`, so nothing is pre-initialized. |
| 5 | tests | Assert against the real `get_moe_metrics_tracker().metrics`, no stubbing of the deprecated accessor. Verified they **fail on `ddda2fde` and pass on `2df90ac9`**. |
| 6 | docstring / shadowing | `Args:` completed per the `get_mtp_metrics()` convention; loop-local renamed `num_tracked_layers`. |

### One deliberate deviation from the suggested snippet

I used **exact matching plus a non-zero coefficient check** rather than the substring test in [`training.py:2689`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/training/training.py#L2688-L2696):

- `"aux_loss" in "seq_aux_loss"` is `True`, so a substring test would pre-initialize `load_balancing_loss` for `seq_aux_loss` users — re-creating the zero-metric problem from comment 4.
- `sft.yaml:142` and `dpo.yaml:157` set `moe_router_load_balancing_type: "aux_loss"` but never set `moe_aux_loss_coeff` (mcore default `0.0`), so the router returns early and records nothing.

This mirrors `MoETopKRouter.get_aux_loss_coeff` / `is_aux_loss_enabled`. Happy to switch to the upstream form if you'd rather stay byte-identical with `training.py`.

nemo_rl/models/megatron/common.py
dafu-wu: Confirmed and fixed. Switched to `get_moe_metrics_tracker().ensure_initialized(name, num_layers + (mtp_num_layers or 0))`, with the import added at module top.

This also removes the hardcoded `device="cuda"` (the API picks the device itself, with a CPU fallback) and the `reduce_group_has_dp` key — renamed to `needs_dp_avg` in the same refactor, which is another sign the original block was written against the pre-refactor API.

Version detail in the main thread: my repro was on 0.19.0, not ≤0.17.x, so the patch was inert there too.

nemo_rl/models/policy/workers/megatron_policy_worker.py
dafu-wu: Fixed — `_finish_train_step_body` now passes `num_layers` / `mtp_num_layers` / `track_names`. All three sites are symmetric now.

Good catch on the nightly gap too: both single-controller entries are dense models, so that branch would never have been exercised.

nemo_rl/models/megatron/common.py
dafu-wu: Good catch — this would have hit every MoE config in the repo. The gate is now `if num_layers is not None and track_names:`, and `track_names` comes out empty under `"none"`, so nothing is pre-initialized and the falsy-guard behaviour is unchanged.

For the record, `examples/nemo_gym/grpo_nanov3.yaml` is PP=2 with `"none"`, so this was reachable in-tree.

nemo_rl/models/megatron/common.py
dafu-wu: Added. The tests assert against the real `get_moe_metrics_tracker().metrics` and do not stub `get_moe_layer_wise_logging_tracker` — thanks for calling out that trap explicitly, it would have been easy to follow the existing house style and end up with a green-but-meaningless test.

Verified the new tests **fail on `ddda2fde` and pass on `2df90ac9`**. Also added coverage for `get_aux_loss_track_names` across all balancing types and the list form.

nemo_rl/models/megatron/common.py
dafu-wu: Both done — the three parameters are documented in the `Args:` block following the `get_mtp_metrics()` convention, and the loop-local is renamed to `num_tracked_layers`.

nemo_rl/models/megatron/common.py
dafu-wu: Fixed via `get_aux_loss_track_names(model_config)`, now passed at all three call sites and covering all four names the router records.

One deviation worth flagging: I used exact matching + a non-zero coefficient check instead of the substring test in `training.py:2689`. `"aux_loss" in "seq_aux_loss"` is `True`, so the substring form would pre-initialize `load_balancing_loss` for `seq_aux_loss` users and re-create the zero-metric issue from your next comment. The coefficient check matters because `sft.yaml` / `dpo.yaml` name `aux_loss` without setting `moe_aux_loss_coeff` (default `0.0`), where the router records nothing. This mirrors `get_aux_loss_coeff` / `is_aux_loss_enabled`.

Agreed `z_loss` is latent — included it anyway so the helper stays correct if `moe_z_loss_coeff` is ever plumbed through.

dafu-wu:
@yfw @terrykong  can you help review it again?

nemo_rl/models/megatron/common.py
```diff
@@ -121,10 +122,85 @@ def broadcast_tensor(
...
+        # Only a real number can enable the loss; anything else (None, or a stand-in
+        # object from a partially-populated config) means "not configured".
+        if isinstance(coeff, (int, float)) and coeff > 0 and name not in track_names:
+            track_names.append(name)
+
+    # z_loss is recorded independently of the load balancing type, gated only on its
+    # own coefficient being set (see MoETopKRouter.apply_z_loss).
+    z_loss_coeff = getattr(model_config, "moe_z_loss_coeff", None)
+    if isinstance(z_loss_coeff, (int, float)):
+        track_names.append("z_loss")
```
yfw: [`common.py:188-192`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L188-L192)

**The pre-init order doesn't match the order Megatron records in, which can silently cross-match the collectives.**

[`common.py:261`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L261) calls `reduce_aux_losses_tracker_across_ranks()` with no arguments, and the shim then falls back to *local* insertion order:

```python
# moe_utils.py:1024
names_list = track_names if track_names is not None else list(tracker.metrics.keys())
```
([`_sync_metrics` issues one `all_reduce` per name in that order](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L265-L272))

But the two ranks disagree on that order:

- A rank **with** MoE layers gets the router's order, which is fixed: [`apply_z_loss` runs first](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L639), then always aux → seq → global ([L679/685/693](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L679)).
- A rank **without** MoE layers gets `track_names` order — aux names in *config* order, with `z_loss` appended last here.

[`clear()` keeps entries](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L214-L217), so whatever order is established on step 1 persists. Verified by running the helper against the router's order:

```
aux_loss + z_loss      helper=[load_balancing_loss, z_loss]   router=[z_loss, load_balancing_loss]   MATCH: False
["seq_aux_loss","aux_loss"]  helper=[seq_load_balancing_loss, load_balancing_loss]                   MATCH: False
```

Same shapes and dtypes, so NCCL won't error — it pairs the buffers positionally and the two metrics get silently cross-reduced. Wrong numbers rather than a hang.

**Not reachable from anything we ship today** — `moe_z_loss_coeff` isn't plumbed in NeMo-RL and no Bridge provider sets it, and the list form would need a matching list `moe_aux_loss_coeff`. So this is a latent trap rather than a live bug, but it's cheap to close now while the helper is new. Note the docstring at [L154-155](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L154-L155) already promises "in the order Megatron records them".

The most robust fix is to stop depending on local order at all — pass the names through, which is exactly what [`report()` does internally](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L176-L188):

```suggestion
    # z_loss is recorded independently of the load balancing type, gated only on its
    # own coefficient being set (see MoETopKRouter.apply_z_loss). It is emitted first
    # because the router applies it before the balancing losses, and
    # reduce_aux_losses_tracker_across_ranks() iterates the tracker in insertion order.
    z_loss_coeff = getattr(model_config, "moe_z_loss_coeff", None)
    if isinstance(z_loss_coeff, (int, float)):
        track_names.append("z_loss")
```

and reorder the loop above it to iterate `_AUX_LOSS_TRACK_NAMES` (whose own order is already aux → seq → global) rather than the config list, looking the coefficient up by `routing_types.index(balancing_type)` to mirror [`get_aux_loss_coeff`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L267-L280). Optionally also pass `track_names` into the reduce call when pre-init ran.

**Heads-up on the test:** [`test_moe_metrics.py:148`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/tests/unit/models/megatron/test_moe_metrics.py#L148) currently asserts `["load_balancing_loss", "z_loss"]`, which pins the divergent order — it'll need flipping to `["z_loss", "load_balancing_loss"]` alongside this. A case pinning the list form to router order would be a good addition too.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -914,6 +917,13 @@ def train(
             moe_metrics = get_moe_metrics(
                 loss_scale=moe_loss_scale,
                 per_layer_logging=self.cfg["megatron_cfg"]["moe_per_layer_logging"],
+                # Pre-initialize the aux-loss tracker on every PP rank so the
+                # cross-PP all_reduce inside get_moe_metrics does not hang when a
+                # rank recorded no aux loss this step (e.g. a stage with no MoE
+                # layer, or MTP MoE on the last stage).
+                num_layers=getattr(model_config, "num_layers", None),
+                mtp_num_layers=getattr(model_config, "mtp_num_layers", None),
+                track_names=get_aux_loss_track_names(model_config),
```
yfw: [`megatron_policy_worker.py:920-926`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/policy/workers/megatron_policy_worker.py#L920-L926)

This block sits outside the `if not eval_mode:` guard, so pre-init also fires on validation steps — where nothing is ever recorded.

`train(eval_mode=True)` calls [`self.model.eval()`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/policy/workers/megatron_policy_worker.py#L649) under `torch.no_grad()`, and both router record paths gate on `self.training and torch.is_grad_enabled()` ([z_loss](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L532), [aux](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L670)). So no rank records, but `num_layers`/`track_names` are config-derived rather than step-derived, so every rank still creates a zeros entry, `if moe_metrics:` becomes truthy, and validation metrics gain a flat-zero `load_balancing_loss` series that wasn't there before.

That's the same permanently-zero-metric concern your own helper docstring cites as the reason for gating on the coefficient — it just isn't applied to the forward-only path. Skipping pre-init there is safe for the hang, since no rank records in eval mode so the tracker is uniformly empty.

```suggestion
                num_layers=getattr(model_config, "num_layers", None),
                mtp_num_layers=getattr(model_config, "mtp_num_layers", None),
                track_names=(
                    [] if eval_mode else get_aux_loss_track_names(model_config)
                ),
```

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1457,6 +1467,13 @@ def _scale_metric(name: str, value: Any) -> Any:
...
```
yfw: [`megatron_policy_worker.py:1476`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1476)

This async `finish_train_step` site is the one round 1 missed, and the guard here is entirely new — but nothing asserts it. Would you consider extending the existing test so a future revert of these three kwargs gets caught?

[`test_megatron_split_state.py:536-537`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/tests/unit/models/policy/test_megatron_split_state.py#L536-L537) already reaches this exact call and asserts only `loss_scale`. One caveat worth knowing: `_make_mock_model()` builds `model.config` as a `MagicMock`, so `get_aux_loss_track_names` returns `[]` unless the test sets real values — which is why the assertion needs them spelled out:

```python
        w.model.config.num_moe_experts = 4
        w.model.config.num_layers = 32
        w.model.config.mtp_num_layers = None
        w.model.config.moe_router_load_balancing_type = "aux_loss"
        w.model.config.moe_aux_loss_coeff = 1e-2
        ...
        kwargs = mock_module_symbols["moe"].call_args.kwargs
        assert kwargs["num_layers"] == 32
        assert kwargs["track_names"] == ["load_balancing_loss"]
```

Fair warning that I could not run this myself — that module skips in my environment (`No module named 'transformer_engine'`), so please treat the snippet as a sketch rather than verified.

nemo_rl/models/megatron/common.py
```diff
@@ -121,10 +122,85 @@ def broadcast_tensor(
...
+    ``moe_aux_loss_coeff`` is a list of the same length (validated by Megatron's
+    ``TransformerConfig``), so more than one aux loss can be live at once.
+
+    Args:
+        model_config: Megatron ``TransformerConfig`` (or any object exposing
+            ``moe_router_load_balancing_type`` / ``moe_aux_loss_coeff``).
+
+    Returns:
+        list[str]: Aux-loss tracker names to pre-initialize, in the order Megatron
+        records them. Empty when no aux loss is enabled.
```
yfw: [`common.py:149-155`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L149-L155)

Two small docstring points:

1. The `Args:` entry lists `moe_router_load_balancing_type` / `moe_aux_loss_coeff`, but the function also reads `moe_z_loss_coeff` at [L190](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L190).
2. `Returns:` promises "in the order Megatron records them" — which is the right invariant to state, but isn't currently true (see the ordering comment above). Worth keeping the sentence and making the code match it.

```suggestion
    Args:
        model_config: Megatron ``TransformerConfig`` (or any object exposing
            ``moe_router_load_balancing_type`` / ``moe_aux_loss_coeff`` /
            ``moe_z_loss_coeff``).

    Returns:
        list[str]: Aux-loss tracker names to pre-initialize, in the order Megatron
        records them (z_loss first, then aux -> seq -> global). The reduction below
        iterates the tracker in insertion order, so this order is load-bearing.
        Empty when no aux loss is enabled.
```

nemo_rl/models/megatron/common.py
```diff
@@ -135,13 +211,53 @@ def get_moe_metrics(
...
+    # lazily (MoEMetricsTracker.record only allocates torch.zeros(num_layers) the first
+    # time a rank records a loss), and _sync_metrics skips names it does not have. If
+    # any PP rank did not record an aux loss this step, it skips the all_reduce for that
+    # name while other PP ranks perform it -> the collective mismatches participants and
+    # hangs with no traceback.
+    #
+    # Mirror Megatron's own report(force_initialize=True) guard, which calls
+    # ensure_initialized() for the same reason. Sizing must be
+    # (num_layers + mtp_num_layers) to match what the router passes to record().
+    if num_layers is not None and track_names:
```
yfw: [`common.py:253`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L253)

Minor, and a judgement call: when `track_names` is non-empty but `num_layers` is `None`, this silently skips the pre-init — and the resulting failure mode is the original no-traceback PP hang, which is the exact symptom the PR exists to remove.

The callers use `getattr(model_config, "num_layers", None)`, so any config object missing the attribute lands here quietly. Megatron itself is loud about the equivalent case:

```python
# moe_logging.py:182-183
raise ValueError("num_layers must be provided when force_initialize=True.")
```
([source](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L181-L186))

A one-line warning would make a misconfiguration diagnosable instead of presenting as a hang:

```python
    if track_names and num_layers is None:
        warnings.warn(
            "MoE aux-loss tracker pre-initialization skipped: num_layers unavailable "
            "on model_config. PP ranks without MoE layers may hang in the aux-loss "
            "all_reduce.",
            stacklevel=2,
        )
```

Happy for you to skip this if you'd rather keep the function quiet — in practice `num_layers` is always present on a real `TransformerConfig`.

nemo_rl/models/megatron/common.py
yfw: [`common.py:188-192`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L188-L192)

**The pre-init order doesn't match the order Megatron records in, so the reduction ends up depending on rank-local dict order.** This is a robustness nit rather than a correctness bug — impact note below.

[`common.py:261`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L261) calls `reduce_aux_losses_tracker_across_ranks()` with no arguments, and the shim then falls back to *local* insertion order:

```python
# moe_utils.py:1024
names_list = track_names if track_names is not None else list(tracker.metrics.keys())
```
([`_sync_metrics` issues one `all_reduce` per name in that order](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L265-L272))

But the two ranks disagree on that order:

- A rank **with** MoE layers gets the router's order, which is fixed: [`apply_z_loss` runs first](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L639), then always aux → seq → global ([L679/685/693](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L679)).
- A rank **without** MoE layers gets `track_names` order — aux names in *config* order, with `z_loss` appended last here.

[`clear()` keeps entries](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L214-L217), so whatever order is established on step 1 persists. Verified by running the helper against the router's order:

```
aux_loss + z_loss      helper=[load_balancing_loss, z_loss]   router=[z_loss, load_balancing_loss]   MATCH: False
["seq_aux_loss","aux_loss"]  helper=[seq_load_balancing_loss, load_balancing_loss]                   MATCH: False
```

Collective order has to agree across ranks on a communicator, so the two ranks pair their operations positionally. Importantly this does **not** corrupt values: a rank that recorded nothing contributes an all-zeros tensor, so each pairing still sums to the correct total. The observable effect is limited to swapped metric *labels* in the local tracker of a PP rank that owns no MoE layer — the ranks that actually hold the losses stay correct.

**Not reachable from any config shipped today.** `moe_z_loss_coeff` isn't plumbed through NeMo-RL and no Bridge provider sets it, so the z_loss trigger is inert; and the list-form trigger would need a list `moe_aux_loss_coeff`, which `setup.py` never assigns (and mcore asserts the two lists match in length). So this is purely defensive — worth doing while the helper is new, not something to hold the PR for.

The most robust fix is to stop depending on local order at all — pass the names through, which is exactly what [`report()` does internally](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/moe_logging.py#L176-L188) and reorder the loop above it to iterate `_AUX_LOSS_TRACK_NAMES` (whose own order is already aux → seq → global) rather than the config list, looking the coefficient up by `routing_types.index(balancing_type)` to mirror [`get_aux_loss_coeff`](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/transformer/moe/router.py#L267-L280). Optionally also pass `track_names` into the reduce call when pre-init ran.

**Heads-up on the test:** [`test_moe_metrics.py:148`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/tests/unit/models/megatron/test_moe_metrics.py#L148) currently asserts `["load_balancing_loss", "z_loss"]`, which pins the divergent order — it'll need flipping to `["z_loss", "load_balancing_loss"]` alongside this. A case pinning the list form to router order would be a good addition too.

While you're in here, two small docstring points at [L149-155](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L149-L155): the `Args:` entry omits `moe_z_loss_coeff`, which the function reads at [L190](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/megatron/common.py#L190); and `Returns:` promises "in the order Megatron records them", which is the right invariant to state but isn't currently true. If you take the fix above, the ordering stops mattering and only the `Args:` line needs updating.

nemo_rl/models/policy/workers/megatron_policy_worker.py
yfw: [`megatron_policy_worker.py:1476`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1476)

This async `finish_train_step` site is the one round 1 missed, and the guard here is entirely new — but nothing asserts it. Would you consider extending the existing test so a future revert of these three kwargs gets caught?

[`test_megatron_split_state.py:536-537`](https://github.com/NVIDIA-NeMo/RL/blob/52538e2536b7e8918564406c35a2bce1f682b50f/tests/unit/models/policy/test_megatron_split_state.py#L536-L537) already reaches this exact call and asserts only `loss_scale`. One caveat worth knowing: `_make_mock_model()` builds `model.config` as a `MagicMock`, so `get_aux_loss_track_names` returns `[]` unless the test sets real values — which is why the assertion needs them spelled out:

```python
        w.model.config.num_moe_experts = 4
        w.model.config.num_layers = 32
        w.model.config.mtp_num_layers = None
        w.model.config.moe_router_load_balancing_type = "aux_loss"
        w.model.config.moe_aux_loss_coeff = 1e-2
        ...
        kwargs = mock_module_symbols["moe"].call_args.kwargs
        assert kwargs["num_layers"] == 32
        assert kwargs["track_names"] == ["load_balancing_loss"]
```

Fair warning that I could not run this myself — that module skips in my environment (`No module named 'transformer_engine'`), so please treat the snippet as a sketch rather than verified.

yfw (approved):

dafu-wu:
@yfw  Thanks for approving the PR! Could you also approve and run the four pending workflows? I don’t have permission to approve them.

vikalluru:
@dafu-wu @yfw - Can we merge the PR now?

yfw (approved):

Issue #3342, opened 2026-07-24
Add vLLM↔TRT-LLM /v1/chat/completions parity test (CI)


terrykong:
## Summary

The TRT-LLM OpenAI HTTP server (`nemo_rl/models/generation/trtllm/trtllm_http_server.py`) and the vLLM server independently re-implement the `/v1/chat/completions` path — prompt construction, tool parsing, reasoning parsing, and EOS-stripping — sharing only [`replace_prefix_tokens`](https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/models/generation/openai_server_utils.py). Nothing currently guarantees the two backends emit **identical tokens** for the same input, yet NeMo-Gym consumes their outputs interchangeably.

This was raised during review of #3130 ([thread](https://github.com/NVIDIA-NeMo/RL/pull/3130#discussion_r3620330266)). That PR added isolated TRT-LLM-side unit coverage (`tests/unit/models/generation/test_trtllm_http_server.py`) and is being merged first to unblock mlperf (validated on a qwen3.5-35B workload against vLLM); the **cross-backend parity test** is deferred to this follow-up.

## Proposed test

Add a greedy (`temperature=0.0`) parity test: the same small model (e.g. Qwen3-0.6B, TP=1) and the same multi-turn scenario driven through each backend's `/v1/chat/completions`, asserting **bitwise-identical** `prompt_token_ids`, `generation_token_ids`, and `content` (the exact fields NeMo-Gym reads back).

The scenario should include a **tool-call turn** (request with `tools=`, assistant emits a `<tool_call>`, then a `tool` result message and a follow-up turn) — this exercises the highest-drift surfaces at once: tool parsing/serialization, `finish_reason="tool_calls"`, and the prefix splice across the tool boundary.

Since vllm/trtllm live in separate extras and CI lanes, the practical shape is a committed golden of `(prompt_token_ids, generation_token_ids, content, tool_calls)` that each backend's lane asserts against (TRT-LLM side in `tests/unit/models/generation/trtllm/test_trtllm_http_server.py` with `pytestmark = pytest.mark.trtllm`), or a functional/L1 nightly launching both backends — which would also cover the "Add CI test(s)" TODO.

SGLang is out of scope — it's driven token-in/token-out and never hits this server layer.

## Links
- PR: #3130
- Review thread: https://github.com/NVIDIA-NeMo/RL/pull/3130#discussion_r3620330266
[...]

PR #3130, opened 2026-07-08
feat(trtllm): support agentic training for SWE rollout

hchings:
# What does this PR do ?

Add support of agentic training for SWE with TRTLLM rollout. 
This MR builds on top of https://github.com/NVIDIA-NeMo/RL/pull/2420/ (merged).

[‼️ Notes for reviewers ‼️] 
- This MR only ensure convergence of trtllm aligns w/ vllm's. Perf optimizations will be in subsequent MRs.

#### 1. `Qwen3.5-35B-A3B` 4 nodes X GB200/GB300
- [TRTLLM wandb, 07/22](https://wandb.ai/nvidia/grpo-dev-erinh/runs/83q0fgo1?nw=nwusererinh)
- [vllm wandb](https://wandb.ai/nvidia/grpo-dev-erinh/runs/bdbrd3jk?nw=nwusererinh)
- Convergence related curves aligned between both backends in the first 20+ steps.

Retested w/ latest commits as of `7/25`.

#### 2. `Qwen3-30B-Thinking`, 16 nodes X H100s
- [TRTLLM wandb](https://wandb.ai/nvidia/swe-benchmark-30b-erinh?nw=nwusererinh), using the recipe provided by Rayen. Please see pinned run w/ suffix `eos-fix-0630-r3`.  
- Convergence aligned with BIn Hu's run with vllm. [vllm wandb](https://wandb.ai/nvidia/binhu-nemo-rl/runs/s9zc8dn3?nw=nwuserbihu), verified by Jia in early July.

Last tested til commit 0591b0ed18c92f3226bdc63053976c8033d860a1. 
Tentatively remove Thinking scripts in https://github.com/NVIDIA-NeMo/RL/pull/3130/changes/07dbfd44ceb08a32873da5e002c62f1e9e34b272 and will add them back in a follow-up MR. 

# Issues
[TRTLLM-13972](https://jirasw.nvidia.com/browse/TRTLLM-13972)

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
[...]

PR #3422, opened 2026-07-30
fix: Add chat completion parity tests for TRTLLM v.s. vLLM and fix disparities

hchings:
# What does this PR do ?

Follow-up task for #3130. Context in https://github.com/NVIDIA-NeMo/RL/issues/3342#issuecomment-5089148270 

### 1. Unit tests for pre-merge -- focus on parity of prompt construction, tool parsing, reasoning parsing, and EOS-stripping. 
- Runs the same Qwen3-0.6B model and multi-turn tool-use scenarios through real TRT-LLM and vLLM backends with greedy decoding.
- Covers Qwen3 and DeepSeek-R1 reasoning parsers, including thinking-enabled and thinking-disabled scenarios.
- Verifies the exact engine prompt token IDs match on the first turn.
- Verifies prompt construction across later turns by comparing the appended prompt suffix after:
- An assistant tool call and deterministic tool result.
- A normal assistant response and follow-up user message.
- Compares first-turn tool-call generation tokens across backends with a minimum similarity threshold of 0.9, allowing expected kernel-level generation differences.
- Validates equivalent structured tool parsing for representative valid, malformed, and mixed-content tool-call outputs.
- Validates reasoning extraction, marker removal, and content separation for both reasoning parsers.
- Verifies generated EOS/stop tokens are retained in response token IDs and logprobs while excluded from parsed response text,
    preserving multi-turn token continuity.
- Uses a committed TRT-LLM golden so the TRT-LLM and vLLM CI lanes can validate the same parity contract independently.

### 2. Nightly -- e2e run comparing reward/logprob error/grad_norm, etc for SWE task.
WIP 

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


tests/unit/models/generation/chat_template_parity_common.py
```diff
@@ -0,0 +1,205 @@
...
+                {"name": "run_bash", "arguments": {"command": "whoami"}},
+            ],
+        },
+    },
+    {
+        "name": "text_before_and_after_tool_call",
+        "raw_output": "before " + _tool_call_text("pwd") + " after",
+        "expected": {
+            "reasoning_content": "",
+            "content": "before ",
```
terrykong: [`chat_template_parity_common.py:138`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/chat_template_parity_common.py#L138)

**1 action item.**

### AI-1
This case cannot pass on the TRT-LLM side, so `test_tool_parser_contracts` fails deterministically in `L0_Unit_Tests_Trtllm`. Two independent strips forbid a trailing space:

1. upstream `Qwen3ToolParser.detect_and_parse` does [`normal_text = text[:idx].strip()`](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/serve/tool_parser/qwen3_tool_parser.py#L53)
2. NeMo-RL strips again: [`return result.normal_text.strip(), calls`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L434)

So actual is `"before"`, expected is `"before "` → `AssertionError: TRT-LLM qwen3 tool contract 'text_before_and_after_tool_call' failed`. The expectation matches vLLM's Hermes parser, which slices without stripping.

This is a genuine adapter divergence worth deciding on rather than papering over: vLLM keeps the trailing space, TRT-LLM strips it, and **both silently drop the `" after"` suffix** (upstream keeps only `text[:idx]`).

**Action:** make the expectation whitespace-insensitive (compare `content.strip()`), or split it per backend and pick a canonical whitespace policy.

```suggestion
            "content": "before",
```

tests/unit/models/generation/test_vllm_http_server_parity.py
```diff
@@ -0,0 +1,419 @@
...
+
+    body1 = {
+        "model": MODEL,
+        "messages": [{"role": "user", "content": USER_MSG}],
+        "tools": [TOOL_DEF],
+        "temperature": 0.0,
+        "top_p": 1.0,
+        "top_k": -1,
+        "max_tokens": 512,
+        "logprobs": True,
```
terrykong, line 145: [`test_vllm_http_server_parity.py:149`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L149)

**1 action item** — applies once you rebase (see the conflict note on `vllm_worker_async.py`).

### AI-1
[#3390](https://github.com/NVIDIA-NeMo/RL/blob/037fb36bdad40f6336edcdfd39f7c5150b152324/nemo_rl/models/generation/vllm/vllm_worker_async.py#L604) merged on 2026-08-10 and added a hard precondition:

```python
if request.logprobs and return_as_token_id and request.top_logprobs is None:
    raise VLLMValidationError("`top_logprobs` must be set when requesting token "
                              "information from the NeMo-RL chat endpoint.", ...)
```

All three request bodies here (L149, L187, L224) send `"logprobs": True` with **no** `top_logprobs`, and NeMo-RL hardcodes [`return_tokens_as_token_ids=True`](https://github.com/NVIDIA-NeMo/RL/blob/037fb36bdad40f6336edcdfd39f7c5150b152324/nemo_rl/models/generation/vllm/vllm_worker_async.py#L686) → `return_as_token_id` is True. The raise is [caught and converted to HTTP 400](https://github.com/NVIDIA-NeMo/RL/blob/037fb36bdad40f6336edcdfd39f7c5150b152324/nemo_rl/models/generation/vllm/vllm_worker_async.py#L743), so `r1.raise_for_status()` throws and every scenario fails on rebased `main`.

**Action:** add `top_logprobs` to all three bodies.

Related simplification: #3390 now puts `generation_token_ids` and `generation_log_probs` directly on the message, so [`_extract_generation_token_ids`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L102) and its `"token_id:"` string-parsing can be deleted in favour of reading the fields — which also makes this test structurally symmetric with the TRT-LLM one that already does `msg["generation_token_ids"]`.

tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py
```diff
@@ -0,0 +1,499 @@
...
+import pytest
+import requests
+
+from nemo_rl.models.generation.trtllm.trtllm_http_server import (
+    _build_reasoning_parser,
+    _build_tool_parser,
+    _ends_with_token_suffix,
+    _make_parse_tool_calls,
+)
+from tests.unit.models.generation.chat_template_parity_common import (
```
terrykong, line 36: [`test_trtllm_http_server_parity.py:37`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py#L37)

**1 action item.**

### AI-1
`ruff check --select I` reports 2 errors — this import block and the one at [`test_vllm_http_server_parity.py:30`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L30). Verified with the pinned ruff (`v0.9.9`, per `.pre-commit-config.yaml`):

```
Found 2 errors.
[*] 2 fixable with the `--fix` option.
```

`pre-commit` runs this hook, so CI is red until it's fixed.

**Action:** `uv run ruff check --select I --fix tests/unit/models/generation/`

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -613,12 +680,19 @@ async def capture_result_generator():
...
-                    or not isinstance(response, ChatCompletionResponse)
+                    not isinstance(response, ChatCompletionResponse)
                     or final_res is None
                 ):
                     return response
 
+                # Return the engine-recorded prompt after prefix replacement.
+                prompt_token_ids = list(final_res.prompt_token_ids or [])
+                for choice in response.choices:
+                    choice.message.prompt_token_ids = prompt_token_ids
```
terrykong: [`vllm_worker_async.py:691`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/nemo_rl/models/generation/vllm/vllm_worker_async.py#L691)

**1 action item** (rebase), **1 follow-up**.

### AI-1 — delete this hunk on rebase
The PR is `CONFLICTING` with `main`. [#3390 "perf(vllm): return token metadata with chat completions"](https://github.com/NVIDIA-NeMo/RL/blob/037fb36bdad40f6336edcdfd39f7c5150b152324/nemo_rl/models/generation/vllm/vllm_worker_async.py#L626) merged 2026-08-10 and supersedes **this hunk specifically**, as a superset — it attaches `prompt_token_ids`, `generation_token_ids` *and* `generation_log_probs`, and generalises the dump helper into a field loop. Its motivation was performance: it removes an awaited `/tokenize` round-trip Gym was making every turn (~1.17x end-to-end).

To be precise about what survives, since this PR touches four regions of the file:

| hunk | what | after rebase |
|---|---|---|
| `@@ -323` | `parse_chat_output` | survives |
| `@@ -393` | compute `model_stop_token_ids` | survives |
| `@@ -561` | pass it to `replace_prefix_tokens` | survives |
| **`@@ -613`** (this one) | `prompt_token_ids` + control flow | **superseded — delete** |

Only this hunk overlaps #3390 (which touched old lines 45, 594-628, 738), so it is the conflict. The `utils.py` and `test_vllm_utils.py` changes go with it. `main` also moved vLLM 0.20.0 -> 0.25.1 in the same window, so re-check anything validated against 0.20.0.

Also note #3390 *gates* the attach on `request.logprobs and return_as_token_id` and adds a `top_logprobs` precondition — which breaks the new parity test (see the comment on [`test_vllm_http_server_parity.py:149`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L149)).

### Follow-up — worth a sentence in the PR description
After rebase, the only change this PR still makes to **vLLM runtime behaviour** is hunks `@@ -393` / `@@ -561`: vLLM now counts `cfg["stop_token_ids"]` and `generation_config.eos_token_id` as turn terminators, where before only `tokenizer.eos_token_id` counted. That is needed to keep both backends symmetric through the shared util — but it is a behaviour change to the backend that wasn't the subject of this PR, in a case that (as far as I can tell) nothing in-repo reaches. Calling that out explicitly would save the next reviewer the same trace.

tests/unit/models/generation/chat_template_parity_common.py
```diff
@@ -0,0 +1,205 @@
...
+            "reasoning_content": "unexpected thought",
+            "content": "Visible answer",
+            "tool_calls": [],
+        },
+    },
+    {
+        "name": "prompt_injected_reasoning_start",
+        "raw_output": "inspect files</think>Visible answer",
+        "enable_thinking": True,
+        "reasoning_at_start": True,
```
terrykong, line 102: [`chat_template_parity_common.py:102`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/chat_template_parity_common.py#L102)

**1 action item.**

### AI-1
`reasoning_at_start` is honored on the TRT-LLM side but **silently ignored on the vLLM side**, so this shared table asserts the same `expected` dict from *different* inputs per backend.

- TRT-LLM: [`_parse_with_trtllm`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py#L387) forwards it — `reasoning_at_start=case["reasoning_at_start"]` ([L394](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py#L394)).
- vLLM: [`_parse_with_vllm`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L295) passes only `raw_output` / `reasoning_parser` / `tool_parser` / `tools` / `enable_thinking`, and [`parse_chat_output`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/nemo_rl/models/generation/vllm/vllm_worker_async.py#L326) has no such parameter.

`prompt_injected_reasoning_start` therefore passes on vLLM only because its parsers happen to treat a missing `<think>` as reasoning-at-start — not because the flag was honored. A shared contract table with a field live on one side and dead on the other is exactly the silent divergence this PR exists to catch.

**Action:** either plumb `reasoning_at_start` through [`parse_chat_output`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/nemo_rl/models/generation/vllm/vllm_worker_async.py#L326), or drop the field from the table and add a comment that vLLM infers it.

tests/unit/models/generation/chat_template_parity_common.py
```diff
@@ -0,0 +1,205 @@
...
+        "raw_output": "before " + _tool_call_text("pwd") + " after",
+        "expected": {
+            "reasoning_content": "",
+            "content": "before ",
+            "tool_calls": [{"name": "run_bash", "arguments": {"command": "pwd"}}],
+        },
+    },
+    {
+        "name": "malformed_tool_call_falls_back_to_content",
+        "raw_output": "<tool_call>{bad json}</tool_call>",
```
terrykong: [`chat_template_parity_common.py:144`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/chat_template_parity_common.py#L144)

**1 action item.**

### AI-1
This case passes on TRT-LLM **by accident**, via a completely different code path than vLLM. TRT-LLM's `has_tool_call` matches the literal `"<tool_call>\n"` ([qwen3_tool_parser.py](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/serve/tool_parser/qwen3_tool_parser.py#L39)); this raw string has no newline, so it returns False and [`_parse` short-circuits](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L410) before any JSON parsing happens. The malformed JSON is never exercised on that backend.

**Action:** add the newlines so both backends actually reach the parser.

```suggestion
        "raw_output": "<tool_call>\n{bad json}\n</tool_call>",
```

tests/unit/models/generation/chat_template_parity_common.py
```diff
@@ -0,0 +1,205 @@
...
+        "expected": {
+            "reasoning_content": "",
+            "content": "<tool_call>{bad json}</tool_call>",
+            "tool_calls": [],
+        },
+    },
+)
+
+
+def token_edit_similarity(left: list[int], right: list[int]) -> float:
```
terrykong, line 154: [`chat_template_parity_common.py:154`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/chat_template_parity_common.py#L154)

**1 action item.**

### AI-1
`token_edit_similarity`, [`inclusive_token_span`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/chat_template_parity_common.py#L175) and [`prompt_suffix_after_turn`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/chat_template_parity_common.py#L196) carry **every** parity assertion in both suites and have no unit tests. They're pure Python, so they can be covered in a cheap lane instead of behind a GPU engine.

I wrote 18 cases and ran them — all pass. Worth adding as `tests/unit/models/generation/test_chat_template_parity_common.py`; one of them pins what the 0.9 threshold actually permits (**2 token edits on a 20-token span**), which is otherwise a mystery number.

Note the new file must also be added to the `L0_Unit_Tests_Vllm_*.sh` line-21 path list next to `test_openai_server_utils.py` — `unit/models/generation/` is ignored by `Models_1..4` and `unit/models/` by `Other`, so a file there matching neither `test_vllm*.py` nor that list runs in **no** lane.

```python
def test_token_edit_similarity_090_threshold_allows_two_edits_in_twenty_tokens():
    left = list(range(20))
    two_edits = [999, 998] + list(range(2, 20))
    assert token_edit_similarity(left, two_edits) == pytest.approx(0.9)
    three_edits = [999, 998, 997] + list(range(3, 20))
    assert token_edit_similarity(left, three_edits) < 0.9


def test_inclusive_token_span_returns_markers_inclusive():
    tokens = [0, 7, 8, 42, 43, 9, 10, 99]
    assert inclusive_token_span(tokens, [7, 8], [9, 10]) == [7, 8, 42, 43, 9, 10]


def test_prompt_suffix_after_turn_returns_appended_tokens():
    turns = [
        {"prompt_token_ids": [1, 2], "generation_token_ids": [3, 4]},
        {"prompt_token_ids": [1, 2, 3, 4, 5, 6], "generation_token_ids": [7]},
    ]
    assert prompt_suffix_after_turn(turns, 0) == [5, 6]
```

tests/unit/models/generation/test_vllm_http_server_parity.py
```diff
@@ -0,0 +1,419 @@
...
+        "Golden model revision does not match; regenerate it"
+    )
+    for scenario_name, _ in PARSER_SCENARIOS[reasoning_parser]:
+        assert scenario_name in golden.get("scenarios", {}), (
+            f"Scenario {scenario_name!r} missing from golden; regenerate it"
+        )
+    return golden
+
+
+@pytest.fixture(scope="module", params=tuple(PARSER_SCENARIOS))
```
terrykong: [`test_vllm_http_server_parity.py:255`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L255)

**2 action items** — both CI cost, neither blocking.

### AI-1
`pytest-shard` assigns **per item**, not per module: `sha256hash(item.nodeid) % num_shards`. A module-scoped fixture can't hold its items together, so this file's 4 items (2 params × 2 tests) scatter across all three `L0_Unit_Tests_Vllm_*` shards — I hashed the real nodeids and got builds in **all 3 shards (3-4 engine builds instead of 2)**. Folding `test_parser_contracts` into `test_parity` (one item per param) would cap it at 2.

### AI-2
Neither parity file is listed in [`tests/unit/excluded_unit_tests.sh`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/excluded_unit_tests.sh), whose stated principle is to exclude tests that "spin up WorkerGroups, Policy, **VllmGeneration**, ... load HF models" — which this file does. Scoped correctly this only affects the opt-in `CI:Lfast` lane (`image_tag` is non-empty only for `Lfast`/`docs`, and the PR default is L1 ⇒ `FAST=0`), but it's two lines to fix:

```
    --ignore=unit/models/generation/test_vllm_http_server_parity.py
    --ignore=unit/models/generation/trtllm/test_trtllm_http_server_parity.py
```

nemo_rl/models/generation/trtllm/trtllm_http_server.py
```diff
@@ -69,6 +99,22 @@ def create_app(
         **(default_chat_template_kwargs or {}),
     }
 
+    # Cache tokenized reasoning markers outside the request path.
+    _reasoning_start_suffixes: tuple[tuple[int, ...], ...] = ()
+    if reasoning_parser == "deepseek-r1":
+        _reasoning_start_suffixes = tuple(
```
terrykong: [`trtllm_http_server.py:105`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L105)

**1 action item, 1 follow-up.** To be clear up front: **this change is a fix, not a regression** — I'm only suggesting hardening it.

The old code set `reasoning_at_start = bool(chat_template_kwargs["enable_thinking"])`, and since [`_server_template_kwargs` hardcodes `"enable_thinking": True`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L97) that was *always* True for Qwen3 — whose template does **not** pre-inject `<think>`. Upstream [`DeepSeekR1Parser.parse`](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/llmapi/reasoning_parser.py#L156) with `reasoning_at_start=True` skips the `<think>` partition entirely, so the literal tag was being swallowed into `reasoning_content`. Inferring from the actual prompt is correct.

### AI-1
The 7 hand-enumerated whitespace variants are the weak point: a template whose marker isn't byte-identical to one of them (e.g. `<think>\n\n\n`) falls through to `False` **silently**, and the entire chain-of-thought is returned as visible `content` with empty `reasoning_content` — no error, no log. Decoding the tail is strictly more robust for one cheap decode:

```python
    def _prompt_opens_reasoning(token_ids: list[int]) -> bool:
        tail = tokenizer.decode(token_ids[-16:], skip_special_tokens=False)
        return tail.rstrip().endswith("<think>")
```

Also note the gate is `reasoning_parser == "deepseek-r1"`, but upstream registers the *same* `DeepSeekR1Parser` under six keys ([`qwen3_5`, `minimax_m2`, ...](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/llmapi/reasoning_parser.py#L111)), three of which default `reasoning_at_start=True`. Dropping the name gate would make the inference apply to all of them.

### Follow-up
`NEMO_RL_GENERATE_PARITY_GOLDEN=1` is the only way to regenerate the golden and is documented nowhere (I grepped `docs/`, `tests/`, `.github/`). A module docstring on the TRT-LLM parity file naming the command, the extra, and the GPU requirement would close it.

tests/unit/models/generation/test_vllm_http_server_parity.py
```diff
@@ -0,0 +1,419 @@
...
+        "resources": {"gpus_per_node": None, "num_nodes": None},
+    },
+    "vllm_kwargs": {
+        "revision": MODEL_REVISION,
+        "tokenizer_revision": MODEL_REVISION,
+    },
+}
+
+
+def _wait_for_server(base_url: str, timeout: int = 180) -> None:
```
terrykong, line 97: [`test_vllm_http_server_parity.py:90`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L90)

**1 action item** — minor.

### AI-1
A bare `requests.get(base_url)` inside `try` returns as soon as the socket accepts, so a 404 from a partially-started ASGI app satisfies it. The first `POST` then dies on a raw `raise_for_status()` HTTPError instead of the clear `TimeoutError`, and in a module-scoped fixture that surfaces as an unrelated-looking failure in whichever test ran first. The TRT-LLM sibling already does this correctly by polling `/health` for a 200.

```suggestion
def _wait_for_server(base_url: str, timeout: int = 180) -> None:
    health_url = base_url.rstrip("/").removesuffix("/v1") + "/health"
    deadline = time.time() + timeout
    while time.time() < deadline:
        try:
            if requests.get(health_url, timeout=3).status_code == 200:
                return
        except Exception:
            pass
        time.sleep(2)
    raise TimeoutError(f"vLLM server at {base_url} did not start within {timeout}s")
```

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -323,6 +323,65 @@ def clear_vllm_logger_metrics(self) -> None:
             self.kv_cache_usage_perc = []
             self.generation_tokens = []
 
+    def parse_chat_output(
```
terrykong: [`vllm_worker_async.py:326`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/nemo_rl/models/generation/vllm/vllm_worker_async.py#L326)

**1 action item** — suggestion, not blocking. Survives the rebase, so it is worth deciding on.

### AI-1 — move `parse_chat_output` into the test file
This is a test-only method on a production Ray worker class: the docstring says *"Run parser parity cases in the vLLM worker environment"*, and it has **zero production callers**. Its only caller invokes it **as a string** — [`run_single_worker_single_data("parse_chat_output", ...)`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L301) — so there is no static reference, and grep / "find usages" will not connect the two.

**It does not need the worker.** The method builds *fresh* parsers rather than reusing the server's:

```python
parser_type = ReasoningParserManager.get_reasoning_parser(reasoning_parser)
parser = parser_type(tokenizer)
```

So running it inside the worker performs the same construction as running it in the pytest process — only in a different process. The one thing it takes from the worker is `self.llm.renderer.tokenizer`, and the fixture [already builds an equivalent tokenizer](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/test_vllm_http_server_parity.py#L270) (`AutoTokenizer.from_pretrained(MODEL, revision=MODEL_REVISION)`). vLLM is importable in the test process too — the lane runs `uv run --extra vllm pytest`, and this file already imports vLLM symbols at module scope.

**This PR's own TRT-LLM side is the proof:** [`test_reasoning_parser_contracts`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py#L419) and [`test_tool_parser_contracts`](https://github.com/NVIDIA-NeMo/RL/blob/5aa8e0ab310645566f2e10ed188b21460871b47b/tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py#L434) run the *identical* `REASONING_PARSER_CONTRACT_CASES` / `TOOL_PARSER_CONTRACT_CASES` tables with a plain `@pytest.mark.parametrize` — no fixture, no Ray, no engine.

```python
# test_vllm_http_server_parity.py — body moved verbatim from the worker method
def _parse_with_vllm(tokenizer, *, raw_output, reasoning_parser, tool_parser,
                     tools, enable_thinking) -> dict[str, Any]:
    from vllm.entrypoints.openai.chat_completion.protocol import ChatCompletionRequest
    from vllm.reasoning.abs_reasoning_parsers import ReasoningParserManager
    from vllm.tool_parsers.abstract_tool_parser import ToolParserManager
    ...
```

Then delete the worker method and drop the `run_single_worker_single_data` hop at the call site.

Payoff: 59 lines out of a production file, 8 parser assertions no longer gated behind a GPU engine build, and the two backends' contract tests become structurally symmetric — which is this PR's stated goal.

I am **not** suggesting `nemo_rl/models/generation/vllm/utils.py` as the destination: every helper there has a production caller, so a zero-caller function would just relocate the problem into a different production file.

**One thing I can't rule out:** if you deliberately want the parsers exercised against the *engine's own* tokenizer instance (to catch a tokenizer-identity mismatch the test tokenizer wouldn't reproduce), that is a real reason to keep it in the worker. Nothing in the current test asserts that, so I am asking rather than assuming — if that is the intent, a one-line comment saying so would prevent someone else from "simplifying" it later.

nemo_rl/models/generation/openai_server_utils.py
```diff
@@ -30,6 +31,7 @@ def replace_prefix_tokens(
     model_prefix_token_ids: list[int],
     template_prefix_token_ids: list[int],
     template_token_ids: list[int],
+    model_stop_token_ids: Collection[int] | None = None,
```
terrykong, line 34: @bxyu-nvidia can you help review? i think this makes sense to have a set in case multiple stop tokens

terrykong (commented):
Reviewed at `5aa8e0a` with an agent team (7 agents, two adversarial passes). Nice work — the parity harness is the right idea, and underneath the tests this fixes **three real production bugs**: the swallowed `<think>` tag, the missing `prompt_token_ids` on vLLM chat responses, and sampled non-EOS stop tokens being dropped during prefix splicing.

**⚠️ Please rebase first.** The PR is currently `CONFLICTING` with `main` in all three vLLM files. [#3390](https://github.com/NVIDIA-NeMo/RL/blob/037fb36bdad40f6336edcdfd39f7c5150b152324/nemo_rl/models/generation/vllm/vllm_worker_async.py#L626) merged on 2026-08-10 and implements the `prompt_token_ids` work as a superset, so the vLLM half of this PR should mostly be *dropped* on rebase rather than merged. #3390 also added a `top_logprobs` precondition that will make the new vLLM parity test return HTTP 400 — details inline.

**Three things block green CI:**
1. `test_tool_parser_contracts` cannot pass — the `text_before_and_after_tool_call` expectation is unreachable on TRT-LLM (two independent `.strip()` calls).
2. 2 × ruff `I001` in the new parity test files.
3. After rebase: the vLLM parity requests need `top_logprobs`.

Everything else is a suggestion. A few notes on how this was checked, so you know what to trust: I had no GPU, so all vLLM/TRT-LLM **runtime** claims are static analysis against pinned upstream source — but the two blockers above are verifiable by inspection and need no GPU. Numbers quoted from the golden fixture were extracted from the committed file, and Qwen3 token IDs were verified against `Qwen/Qwen3-0.6B@c1899de`.

If it helps, I wrote up the four disparities with a worked example built from real tokens in your golden (turn 1 → turn 2 splice, showing exactly which token gets skipped): https://terryk.gitlab-master-pages.nvidia.com/nemo-html/terryk/claude-summaries/pr-3422-trtllm-vllm-parity.html

_Generated by Claude Code_

tests/unit/models/generation/chat_template_parity_common.py
hchings: Addressed in ac18494ae9dd3980cd893e7310b2a4a34f450961.

tests/unit/models/generation/test_vllm_http_server_parity.py
hchings: Addressed in ac18494ae9dd3980cd893e7310b2a4a34f450961.

tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py
hchings: addressed in 96bf6342b266d5e046d631a580395db59a9cc88c.

nemo_rl/models/generation/vllm/vllm_worker_async.py
hchings: Addressed in 96bf6342b266d5e046d631a580395db59a9cc88c

tests/unit/models/generation/chat_template_parity_common.py
hchings: Addressed in 96bf6342b266d5e046d631a580395db59a9cc88c

tests/unit/models/generation/chat_template_parity_common.py
hchings: Addressed in 96bf6342b266d5e046d631a580395db59a9cc88c.

tests/unit/models/generation/test_vllm_http_server_parity.py
hchings: Addressed in 96bf6342b266d5e046d631a580395db59a9cc88c

tests/unit/models/generation/chat_template_parity_common.py
hchings: Addressed in 96bf6342b266d5e046d631a580395db59a9cc88c.

nemo_rl/models/generation/trtllm/trtllm_http_server.py
hchings: Addressed in 96bf6342b266d5e046d631a580395db59a9cc88c.

nemo_rl/models/generation/vllm/vllm_worker_async.py
hchings: Addressed in d6e9c8c7087181d13f5b4819a160cd814b3e7db8

tests/unit/models/generation/test_vllm_http_server_parity.py
hchings: Addressed in d6e9c8c7087181d13f5b4819a160cd814b3e7db8

nemo_rl/models/generation/openai_server_utils.py
bxyu-nvidia: ye looks reasonable

nemo_rl/models/generation/trtllm/trtllm_http_server.py
```diff
@@ -33,18 +33,47 @@
...
+def _tokens_for_response_text(
+    token_ids: list[int], stop_token_ids: set[int]
+) -> list[int]:
+    text_token_end = len(token_ids)
+    while text_token_end and token_ids[text_token_end - 1] in stop_token_ids:
+        text_token_end -= 1
+    return token_ids if text_token_end == len(token_ids) else token_ids[:text_token_end]
+
+
+def _ends_with_token_suffix(
```
terrykong: [`trtllm_http_server.py:45`](https://github.com/NVIDIA-NeMo/RL/blob/c2ed2fc3b79b6e7384b7b93c0e0641f654847b4e/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L45)

**1 action item.**

### AI-1 — `_ends_with_token_suffix` is now dead; please delete it and its test
Switching to `_prompt_opens_reasoning` was the right call, but it left this behind: `_ends_with_token_suffix` has **zero production callers**. The `_reasoning_start_suffixes` tuple it was written to consume no longer exists anywhere in the repo.

Its only remaining references are its own test and that test's import:

```
nemo_rl/models/generation/trtllm/trtllm_http_server.py:45   def _ends_with_token_suffix(...)
tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py:34   (import)
tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py:416  def test_ends_with_token_suffix()
```

The function and the test now exist only to justify each other.

**Action:** delete `_ends_with_token_suffix` and [`test_ends_with_token_suffix`](https://github.com/NVIDIA-NeMo/RL/blob/c2ed2fc3b79b6e7384b7b93c0e0641f654847b4e/tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py#L416). Nothing to relocate — `_prompt_opens_reasoning` fully replaced it.

tests/unit/models/generation/test_vllm_http_server_parity.py
```diff
@@ -0,0 +1,457 @@
...
+        model=MODEL,
+        messages=[{"role": "user", "content": "parser contract"}],
+        tools=tools,
+        chat_template_kwargs={"enable_thinking": enable_thinking},
+    )
+    reasoning_content = ""
+    content: str | None = raw_output
+
+    if reasoning_parser is not None:
+        parser_input = "<think>" + raw_output if reasoning_at_start else raw_output
```
terrykong: [`test_vllm_http_server_parity.py:301`](https://github.com/NVIDIA-NeMo/RL/blob/c2ed2fc3b79b6e7384b7b93c0e0641f654847b4e/tests/unit/models/generation/test_vllm_http_server_parity.py#L301)

**1 action item.** Re-opening my earlier `reasoning_at_start` comment — the field is no longer ignored, but the two cases now collapse into one. Good news: the fix is one line.

### AI-1 — drop the `"<think>"` prefix; vLLM already handles this shape
Prefixing `<think>` makes `prompt_injected_reasoning_start` produce **byte-identical parser input** to `explicit_reasoning_answer`:

```
explicit (reasoning_at_start=False): "<think>inspect files</think>Visible answer"
injected (reasoning_at_start=True) : "<think>" + "inspect files</think>Visible answer"
                                   = "<think>inspect files</think>Visible answer"   <- identical
```

So both cases run the same string through the same parser and assert the same dict. The "prompt already opened the block, so there is no opening tag in the output" shape — the only thing case 2 exists to cover — is never exercised on vLLM.

**The prefix isn't compensating for anything.** vLLM has no `reasoning_at_start` *flag*, but it fully supports the *behaviour*: [`BaseThinkingReasoningParser.extract_reasoning`](https://github.com/vllm-project/vllm/blob/v0.20.0/vllm/reasoning/basic_parsers.py) partitions on the opening tag and, when there isn't one, keeps the text as-is and splits on `</think>` — with the comment *"For models that may not generate start token, assume the reasoning content is always at the start."* Running that logic on the raw case-2 string returns exactly what the table expects:

```
shape 1 (has <think>)     -> ('inspect files', 'Visible answer')
shape 2 (no <think>) RAW  -> ('inspect files', 'Visible answer')   <- no prefix needed
```

**Action:** pass `raw_output` through untouched. Case 2 then genuinely tests the no-opening-tag shape on vLLM, and it passes.

```suggestion
        # vLLM has no reasoning_at_start flag: its parser infers the shape from
        # the text, so shape-2 output (no opening <think>) is passed through
        # untouched. TRT-LLM is told explicitly via the flag instead.
        parser_input = raw_output
```

Walk-through of the mechanism, if useful: https://terrykong.github.io/gh-pages-poc/terryk/pr-3422-reasoning-at-start.html

terrykong (commented):
Re-reviewed at `c2ed2fc` after the rebase. **All 11 findings from my previous review are genuinely fixed** — I verified each against the code rather than the commit messages, since the SHAs in the replies predate the rebase. Several went further than asked: dropping `parse_chat_output` entirely rather than relocating it, and making the contract tests tokenizer-only so they no longer need a GPU engine. I also ran the new `test_chat_template_parity_common.py` locally — **19/19 pass**. Flipping the golden so vLLM generates and TRT-LLM conforms is the right direction.

Two follow-ups below, both minor: one dead-code cleanup, and one re-open where the fix works mechanically but the test collapses into a duplicate.

**Convergence is safe** — I checked whether the `content` whitespace difference between backends (vLLM `"before "` vs TRT-LLM `"before"`) could reach the training tokens. It cannot: `replace_prefix_tokens` keeps prior turns as verbatim token IDs and discards the template's re-render of that region, and `final_prompt_token_ids` is used both as the engine prompt and as the returned `prompt_token_ids`, so no train/inference skew is possible. The shared-splice refactor is also a strict no-op when `model_stop_token_ids` is omitted (0 differences over 108,569 randomised cases).

**Not this PR, for the record:** while tracing the above I found that `top_k` is handled inconsistently across backends — vLLM's HTTP path forces `-1` and ignores `cfg["top_k"]`, while TRT-LLM's HTTP path honours it since #3537, and the two families disagree on whether "disabled" is `-1` or `0`. That is all pre-existing on `main` and nothing here needs to change; I filed https://github.com/NVIDIA-NeMo/RL/issues/3776 to document it.

One process note: the unit lanes still have not run (`CI quality check`: *No tests run: Label CI:L\**). Lint is green so the ruff fix is CI-verified, but the tool-parser contract fix deserves a real TRT-LLM lane run — that test was previously failing deterministically, and the fix has only been verified by inspection.

_Generated by Claude Code_

tests/unit/models/generation/test_vllm_http_server_parity.py
hchings: updated

nemo_rl/models/generation/trtllm/trtllm_http_server.py
hchings: updated

tests/test_suites/nightly_gb200.txt
```diff
@@ -20,6 +20,7 @@ tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_generation-noncolocated-m
 # TRT-LLM generation backend
 tests/test_suites/llm/grpo-qwen3-1.7b-2n4g-fsdp2-trtllm.sh
 tests/test_suites/llm/grpo-qwen2.5-0.5b-1n4g-megatron-trtllm-noncolocated-async.sh
+tests/test_suites/llm/grpo-qwen3-0.6b-1n4g-megatron-trtllm-tp2-noncolocated-gym.sh
```
yuki-97: This nightly can't start as configured — `run_grpo_nemo_gym.py` raises before training begins.

The driver launches [`run_grpo_nemo_gym.py`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/examples/nemo_gym/run_grpo_nemo_gym.py#L193), which calls `setup_nemo_gym_config` unconditionally; that function branches only `vllm`/`megatron` before [`raise ValueError(f"NeMo Gym does not support generation backend {backend!r}.")`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/environments/nemo_gym.py#L950), and this recipe sets `backend: trtllm`. Not introduced by this PR — the branch arrived in 13a10647e (#3920), which also broke the documented [`grpo_qwen3_30b_async_swe_trtllm.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/examples/swe_bench/grpo_qwen3_30b_async_swe_trtllm.yaml) path. `TEST_DRYRUN` [exits](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/test_suites/llm/common.env#L60) before the `uv run` line, so the dry-run test can't catch this.

**Suggest fix:**

1. Add a `trtllm` branch to `setup_nemo_gym_config` that sets `trtllm_cfg.expose_http_server`.
2. Or hold this entry out of `nightly_gb200.txt` until that lands on main.

examples/configs/recipes/llm/grpo-qwen3-0.6b-1n4g-megatron-trtllm-tp2-noncolocated-gym.yaml
```diff
@@ -0,0 +1,81 @@
...
+      async_engine: true
+      gpu_memory_utilization: 0.6
+      max_batch_size: 64
+      max_num_tokens: 4096
+      in_flight_weight_updates: ${grpo.async_grpo.in_flight_weight_updates}
+      recompute_kv_cache_after_weight_updates: ${grpo.async_grpo.recompute_kv_cache_after_weight_updates}
+      expose_http_server: true
+      default_chat_template_kwargs:
+        enable_thinking: false
+      tool_parser: qwen3_coder
```
yuki-97, line 59: `qwen3_coder` can't parse what Qwen3-0.6B emits, so tool calls in this nightly come back empty.

Qwen3's chat template emits a JSON body inside `<tool_call>` tags. `Qwen3CoderToolParser` expects `<function=...>` XML instead, but its `has_tool_call` only tests for the bare `<tool_call>` substring — so it enters parsing, `_parse_block` matches no `<function=`, and it returns zero calls. [`_make_parse_tool_calls`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L437-L438) then hits `if not calls: return text, []` — the raw text survives as content, but the gym env sees no tool call, and there is no fallback parser.

Verified against the TRT-LLM commit this repo actually builds, not the PyPI release: `3rdparty/TensorRT-LLM-workspace/pyproject.toml` `[tool.trtllm]` pins `ref = bf2ef86f9a2652132b11773d4041e292c553c142`.

- [`tool_parser_factory.py:58-59`](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/serve/tool_parser/tool_parser_factory.py#L58-L59) — the YAML string routes through `{"qwen3": Qwen3ToolParser, "qwen3_coder": Qwen3CoderToolParser}`.
- [`tool_parser_factory.py:20-22`](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/serve/tool_parser/tool_parser_factory.py#L20-L22) — upstream's own `MODEL_TYPE_TO_TOOL_PARSER` resolves `model_type: "qwen3"` to `"qwen3"`. `Qwen/Qwen3-0.6B@c1899de2`'s `config.json` is `model_type: "qwen3"`, so `qwen3_coder` is not what upstream would auto-pick for this model — it is not the target of any model_type.
- [`qwen3_coder_parser.py:57-58`](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/serve/tool_parser/qwen3_coder_parser.py#L57-L58) — `has_tool_call` is `self.tool_call_start_token in text` with the token being the bare `<tool_call>`; [`:40-42`](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/serve/tool_parser/qwen3_coder_parser.py#L40-L42) — `tool_call_function_regex = r"<function=(.*?)</function>|<function=(.*)$"` is the only thing `_parse_block` looks for.
- [`qwen3_tool_parser.py:34`](https://github.com/NVIDIA/TensorRT-LLM/blob/bf2ef86f9a2652132b11773d4041e292c553c142/tensorrt_llm/serve/tool_parser/qwen3_tool_parser.py#L34) — the JSON-format parser this model's output actually fits (`bot_token = "<tool_call>\n"`).

In-repo, `tool_parser` always tracks the model's emitted format: the base config this recipe inherits uses `nemotron_json` for Nemotron Nano ([`grpo_workplace_assistant_nemotron_nano_v2_9b.yaml:257`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L257)), and the other Qwen gym recipes use vLLM's JSON-format `hermes`. `qwen3_coder`'s only other use in the repo is a Nemotron-Omni VLM recipe. This PR's own TRT-LLM parity test uses [`_build_tool_parser("qwen3")`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py#L394) for the same model.

```suggestion
      tool_parser: qwen3
```

nemo_rl/models/generation/trtllm/trtllm_http_server.py
```diff
@@ -107,6 +120,10 @@ def create_app(
         **(default_chat_template_kwargs or {}),
     }
 
+    def _prompt_opens_reasoning(token_ids: list[int]) -> bool:
```
yuki-97, line 123: The `True` branch of `_prompt_opens_reasoning` has no coverage, and it's the branch a shipped recipe takes.

It's a closure inside `create_app`, so no test can import it — every other module-level helper in this file is imported and tested by [`test_trtllm_http_server.py`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/trtllm/test_trtllm_http_server.py#L350). The CPU contract test passes `reasoning_at_start` straight into `_build_reasoning_parser`, and all three golden scenarios end their prompt in `assistant\n` or `</think>`, so the e2e path only ever reaches `False`. [`grpo_qwen3_30b_async_swe_trtllm.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/examples/swe_bench/grpo_qwen3_30b_async_swe_trtllm.yaml#L20) pairs `deepseek-r1` with Qwen3-30B-A3B-Thinking, whose template ends `<|im_start|>assistant\n<think>\n`.

**Suggest fix:**

1. Hoist it to module scope taking `tokenizer` as a parameter — that is the only name it captures.
2. Add a table test: `assistant\n` -> False, `<think>\n\n</think>\n\n` -> False, `...<think>\n` -> True.

nemo_rl/models/generation/trtllm/trtllm_http_server.py
```diff
@@ -272,14 +297,9 @@ async def chat_completions(request: Request):
...
 
-        # Strip trailing stop tokens TRT-LLM appends — apply_chat_template doesn't reproduce
-        # <|endoftext|>, so they'd break seen_token_ids contiguity. Trim logprobs in lockstep.
-        while gen_token_ids and gen_token_ids[-1] in _eos_token_ids:
-            gen_token_ids.pop()
-            if gen_logprobs:
-                gen_logprobs.pop()
-
-        gen_text = tokenizer.decode(gen_token_ids, skip_special_tokens=False)
+        # Parse without trailing stops; retain full response metadata.
```
yuki-97, line 300: nit: the comment on `include_stop_str_in_output` still describes the trimming this hunk removed.

[`trtllm_http_server.py:95`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L95) reads "so the adapter can trim tokens and logprobs together", but the lockstep `pop()` loop is gone and the flag is now load-bearing for the opposite reason — stop tokens must be retained so `generation_token_ids` matches vLLM and the next turn's `replace_prefix_tokens` can see `model_ended_with_stop`. `_tokens_for_response_text` is text-only and never touches logprobs.

**Suggest fix:** reword line 95 to "Include generated stop tokens so generation_token_ids stays contiguous with the next turn's prompt."

nemo_rl/models/generation/openai_server_utils.py
```diff
@@ -112,7 +112,8 @@ def replace_prefix_tokens(
         if tid == eos_token_id:
             count_seen += 1
             if count_seen == count_needed:
-                template_cut_start = pos
+                # Keep sampled stops; otherwise let the template add the missing EOS.
```
yuki-97, line 115: nit: the docstring still describes the pre-PR cut.

[Lines 71-73](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/generation/openai_server_utils.py#L71-L73) say the splice cuts "at the N-th EOS in template_token_ids", but this line cuts one past it whenever the model's own last token was a stop. The worked example below still evaluates correctly, so this is one stale sentence rather than a stale example — and `model_stop_token_ids` is not named in the prose at all.

**Suggest fix:** reword to "...cut at the N-th EOS in template_token_ids, or just past it when the model's own last token was a stop token", naming `model_stop_token_ids` as the set that decides.

tests/unit/models/generation/test_vllm_http_server_parity.py
```diff
@@ -0,0 +1,465 @@
...
+        ), f"scenario={scenario_name!r}: turn-one engine prompt mismatch"
+
+        for turn_index in range(len(actual_turns) - 1):
+            actual_suffix = prompt_suffix_after_turn(actual_turns, turn_index)
+            expected_suffix = prompt_suffix_after_turn(expected_turns, turn_index)
+            if actual_suffix != expected_suffix:
+                suffix_mismatches.append(
+                    f"scenario={scenario_name!r}: transition {turn_index + 1}->"
+                    f"{turn_index + 2} appended prompt suffix mismatch "
+                    f"(vLLM={actual_suffix!r}, TRT-LLM={expected_suffix!r})"
```
yuki-97, line 450: nit: the expected side here is the vLLM golden, not TRT-LLM.

[`_load_golden`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/test_vllm_http_server_parity.py#L252) asserts `source_backend == "vllm"` and this test is the only writer of the fixture, so a failure here sends the reader hunting in the wrong backend.

```suggestion
                    f"(actual={actual_suffix!r}, golden={expected_suffix!r})"
```

examples/configs/recipes/llm/grpo-qwen3-0.6b-1n4g-megatron-trtllm-tp2-noncolocated-gym.yaml
```diff
@@ -0,0 +1,81 @@
...
+    moe_per_layer_logging: false
+  optimizer:
+    name: torch.optim.AdamW
+    kwargs:
+      lr: 5.0e-06
+      weight_decay: 0.01
+      betas:
+      - 0.9
+      - 0.999
+      eps: 1.0e-08
```
yuki-97, line 37: nit: nothing reads this block — the megatron path builds its optimizer from `megatron_cfg.optimizer`.

`policy.optimizer` is consumed only by the DTensor/AutoModel workers ([dtensor_policy_worker.py:491](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L491), [automodel/setup.py:759](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/automodel/setup.py#L759)); this recipe inherits `megatron_cfg.enabled: true` and `dtensor_cfg.enabled: false`, and these values duplicate `megatron_cfg.optimizer` in the base exactly. The base sets `optimizer: null`, and the nearest sibling recipe re-asserts it.

**Suggest fix:** drop lines 29-37.

tests/unit/models/generation/test_vllm_http_server_parity.py
```diff
@@ -0,0 +1,465 @@
...
+                golden = {
+                    "model": MODEL,
+                    "model_revision": MODEL_REVISION,
+                    "source_backend": "vllm",
+                    "scenarios": {},
+                }
+            golden.setdefault("scenarios", {})
+            golden["scenarios"][scenario_name] = {"turns": actual_turns}
+            GOLDEN_PATH.parent.mkdir(parents=True, exist_ok=True)
+            GOLDEN_PATH.write_text(json.dumps(golden, indent=2) + "\n")
```
yuki-97, line 432: The committed golden [`trtllm/fixtures/chat_template_parity_golden.json`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/trtllm/fixtures/chat_template_parity_golden.json) is 4289 lines / 80 KB, but under 5% of it is load-bearing — and this writer is most of the reason. The other token-ID fixture in this test's own directory, [`test_vllmasyncgenerationworker_replace_prefix_worker.json`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/test_vllmasyncgenerationworker_replace_prefix_worker.json), stores 3×457 token IDs in **5 lines / 7.7 KB** by keeping the arrays inline and storing only what the test asserts on, deriving the rest at runtime ([`test_vllm_generation.py:2374`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/test_vllm_generation.py#L2374) decodes and slices prefixes itself rather than storing them). Three things here diverge from that:

1. **`indent=2` puts one token ID per line.** Arrays inline instead: 80164 B / 4289 lines → 30040 B / 112 lines.
2. **`content` / `reasoning_content` / `tool_calls` / `finish_reason` are never read back out of the golden.** Both `test_parity` bodies assert only `prompt_token_ids`, the tool-call span carved out of `generation_token_ids`, and `prompt_suffix_after_turn`; the text fields reach nothing but the failure artifact at [`test_trtllm_http_server_parity.py:489-490`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/trtllm/test_trtllm_http_server_parity.py#L489-L490). They do have to stay in `_extract_fields` (turns 1-3 assert on them live), so the filter belongs here at write time.
3. **`deepseek_r1_with_thinking` is byte-identical to `qwen3_with_thinking`** in the committed file — I diffed the parsed objects. That is expected: the golden holds engine prompt + greedy generation token IDs, and the reasoning parser only affects how finished text gets split, which is exactly what is never compared. Suggest keying golden scenarios on thinking alone (`with_thinking` / `without_thinking`) and making [`PARSER_SCENARIOS`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/chat_template_parity_common.py#L43-L46) map `parser -> (golden_key, enable_thinking)`. Both parsers then check against one stored blob — a slightly stronger contract than storing it twice.

All three together: **80164 B / 4289 lines → 14495 B / 40 lines**, with the raw per-turn structure kept, so the golden stays regenerable and `prompt_suffix_after_turn`'s prefix invariant still gets asserted against the golden itself.

The block below covers 1 and 2; it needs `import re` next to the `import json` on line 23.

````suggestion
            golden["scenarios"][scenario_name] = {
                "turns": [
                    {
                        "prompt_token_ids": turn["prompt_token_ids"],
                        "generation_token_ids": turn["generation_token_ids"],
                    }
                    for turn in actual_turns
                ]
            }
            GOLDEN_PATH.parent.mkdir(parents=True, exist_ok=True)
            # Keep token arrays inline, like the other token-ID fixtures here.
            GOLDEN_PATH.write_text(
                re.sub(
                    r"\[[\d,\s]+\]",
                    lambda m: "[" + " ".join(m.group()[1:-1].split()) + "]",
                    json.dumps(golden, indent=2),
                )
                + "\n"
            )
````

File: tests/unit/models/generation/test_vllm_http_server_parity.py (hunk 8 above)
Comment (yuki-97) [reply tests/unit/models/generation/test_vllm_http_server_parity.py]: Re-opening for AI-2 only — AI-1 did land.

AI-1: the two contract tests no longer touch the engine (`_parse_with_vllm` at test_vllm_http_server_parity.py:278 runs in-process now that `parse_chat_output` moved out of the worker), so only the two `test_parity` items build one. That is the cap this asked for, even though the suggested fold-into-`test_parity` was not done — the items still scatter across all three shards, they are just cheap now.

AI-2 did not: `grep -n parity tests/unit/excluded_unit_tests.sh` at 7671c1be returns nothing. The list is sourced only under `FAST=1` (run_unit_shard_common.sh:28-29), so the `CI:Lfast` lane still collects both parity files and builds engines there.

File: nemo_rl/models/generation/vllm/vllm_worker_async.py (hunk 4 above)
Comment (yuki-97) [reply nemo_rl/models/generation/vllm/vllm_worker_async.py]: Re-opening for the follow-up only — AI-1 did land.

AI-1: the `@@ -613` hunk is gone; `git diff origin/main...HEAD` on this file now returns three hunks and main's #3390 version of the attach path stands unmodified.

The follow-up did not: the PR description at 7671c1be still has no sentence about the widened stop-token set. `vllm_worker_async.py:400` reads `model_stop_token_ids = set(self.cfg.get("stop_token_ids") or ())` and folds in `generation_config.eos_token_id` at :401-407, where before only `tokenizer.eos_token_id` counted. The body's nearest line — "Verifies generated EOS/stop tokens are retained in response token IDs and logprobs while excluded from parsed response text" — describes a test assertion, not the vLLM runtime change.

File: tests/test_suites/llm/grpo-qwen3-0.6b-1n4g-megatron-trtllm-tp2-noncolocated-gym.sh [hunk 23]
```diff
@@ -0,0 +1,73 @@
...
+    echo "[ERROR] Expected step $MAX_STEPS, reached $LAST_STEP"
+    exit 1
+fi
+
+uv run tests/check_metrics.py "$JSON_METRICS" \
+    'mean(data["train/reward"]) > 0.05' \
+    'median(data["train/token_mult_prob_error"]) < 1.1' \
+    'mean(data["train/gen_kl_error"]) < 0.02' \
+    'mean(data["train/grad_norm"], 2, 0) > 0.1' \
+    'mean(data["train/grad_norm"], 2, 0) < 30.0'
```
yuki-97, line 71: could you paste the wandb curves or metric results into the PR description to show that the new nightly test can run well?

examples/configs/recipes/llm/grpo-qwen3-0.6b-1n4g-megatron-trtllm-tp2-noncolocated-gym.yaml
yuki-97: traced by claude. I'm not sure if change to `qwen3` is correct. could you help double check?

examples/configs/recipes/llm/grpo-qwen3-0.6b-1n4g-megatron-trtllm-tp2-noncolocated-gym.yaml
```diff
@@ -0,0 +1,81 @@
+defaults: ../../../nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml
+grpo:
+  num_prompts_per_step: 4
+  num_generations_per_prompt: 2
+  adv_estimator:
+    name: reinforce_plus_plus
+    minus_baseline: false
+  async_grpo:
+    in_flight_weight_updates: true
```
yuki-97, line 9: `in_flight_weight_updates: true` has no async loop to apply to — this recipe runs synchronous GRPO.

`enabled` isn't set here, and the base config has [`async_grpo.enabled: false`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L53), so the dispatch at [`run_grpo_nemo_gym.py:289`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/examples/nemo_gym/run_grpo_nemo_gym.py#L289) (`elif config.grpo.async_grpo.enabled:`) falls through to the synchronous branch at :334 and `async_grpo_train` never runs.

It isn't fully inert, though: line 54 interpolates this value into `trtllm_cfg.in_flight_weight_updates`, which [`trtllm_generation.py:474-479`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/generation/trtllm/trtllm_generation.py#L474-L479) turns into `drain=not in_flight` — so refit skips the drain under a sync loop.

Dropping the block is the smaller fix, and it matches the recipe's own name (no `async`); the base already supplies `in_flight_weight_updates: false`. If async was the intent instead, add `enabled: true` alongside it, the way [`grpo-nanov3-30BA3B-2n8g-megatron_generation-noncolocated-async-gym.yaml:3-5`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/examples/configs/recipes/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation-noncolocated-async-gym.yaml#L3-L5) does.

```suggestion
```

nemo_rl/models/generation/trtllm/trtllm_http_server.py
```diff
@@ -33,18 +33,31 @@
...
 
-def _build_reasoning_parser(name: str, chat_template_kwargs: dict[str, Any]) -> Any:
-    from tensorrt_llm.llmapi.reasoning_parser import ReasoningParserFactory
+def _tokens_for_response_text(
+    token_ids: list[int], stop_token_ids: set[int]
+) -> list[int]:
+    text_token_end = len(token_ids)
+    while text_token_end and token_ids[text_token_end - 1] in stop_token_ids:
+        text_token_end -= 1
+    return token_ids if text_token_end == len(token_ids) else token_ids[:text_token_end]
```
yuki-97, line 42: The identity return aliases the caller's list, and `gen_token_ids` is what ships back as training data.

Both branches are behaviourally identical — the only consumer is `tokenizer.decode` at [:302](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L302) — so the conditional buys one skipped copy of at most `max_new_tokens` ints and costs an alias: if anything downstream ever trims `text_token_ids`, it silently mutates the `generation_token_ids` sent at [:331](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/generation/trtllm/trtllm_http_server.py#L331), which is the class of bug this hunk exists to prevent. `test_response_text_excludes_stops_without_mutating_training_tokens` asserts exactly that invariant, while [`test_response_text_reuses_unterminated_generation_tokens`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/tests/unit/models/generation/trtllm/test_trtllm_http_server.py#L110-L115) pins the alias with `is` and so blocks the simpler form.

Separately, `text_token_end` reaching 0 is reachable — a generation that is entirely stop tokens, e.g. `[151645]` — and neither new test covers it. It doesn't crash (`decode([])` is `""`), but it is the boundary the `while text_token_end and ...` guard exists for, so it is worth pinning.

Suggested companion edit, replacing `test_trtllm_http_server.py:110-115`:

```python
def test_response_text_returns_all_tokens_when_unterminated():
    generation_token_ids = [10, 20]

    assert _tokens_for_response_text(generation_token_ids, {2, 3}) == [10, 20]


def test_response_text_is_empty_when_generation_is_all_stops():
    assert _tokens_for_response_text([2, 3], {2, 3}) == []
```

```suggestion
    return token_ids[:text_token_end]
```

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -397,6 +397,14 @@ def _setup_vllm_openai_api_server(self, app: FastAPI) -> FastAPI:
...
         engine_client = self.llm
         model_config = self.llm_async_engine_args.create_model_config()
+        model_stop_token_ids = set(self.cfg.get("stop_token_ids") or ())
+        generation_eos_token_ids = model_config.try_get_generation_config().get(
+            "eos_token_id"
+        )
+        if isinstance(generation_eos_token_ids, int):
+            model_stop_token_ids.add(generation_eos_token_ids)
+        elif generation_eos_token_ids is not None:
+            model_stop_token_ids.update(generation_eos_token_ids)
```
yuki-97, line 407: The generation-config eos should be a fallback, not a union — as written it silently widens an explicitly configured `stop_token_ids`.

`configure_generation_config` only fills this in when it is `None` ([`__init__.py:75-76`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/models/generation/__init__.py#L75-L76)), so a value that reaches the worker is either the user's explicit list or the auto-filled `[tokenizer.eos_token_id]` — in both cases it is the configured answer to "what terminates a turn", and unioning `generation_config.json`'s eos list into it overrides that. A recipe pinning `[151645]` precisely to avoid terminating on `<|endoftext|>` would still get `{151645, 151643}`, `model_ended_with_stop` would fire on the excluded token, and the next turn's prompt would lose the template `<|im_end|>`.

Today the gym path nulls `stop_token_ids` outright ([`nemo_gym.py:952-954`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/environments/nemo_gym.py#L952-L954)), so the fallback is what runs and behaviour is unchanged either way — this is about which precedence the code encodes for when that changes.

```suggestion
        model_stop_token_ids = set(self.cfg.get("stop_token_ids") or ())
        if not model_stop_token_ids:
            generation_eos_token_ids = model_config.try_get_generation_config().get(
                "eos_token_id"
            )
            if isinstance(generation_eos_token_ids, int):
                generation_eos_token_ids = [generation_eos_token_ids]
            model_stop_token_ids = set(generation_eos_token_ids or ())
```

Note the `isinstance` normalisation still has to stay: `eos_token_id` is an `int` for many models and a list for Qwen3, and `replace_prefix_tokens` calls `set()` on whatever it receives (`openai_server_utils.py:100`), which would raise on a bare int.

examples/configs/recipes/llm/grpo-qwen3-0.6b-1n4g-megatron-trtllm-tp2-noncolocated-gym.yaml
```diff
@@ -0,0 +1,81 @@
...
+      betas:
+      - 0.9
+      - 0.999
+      eps: 1.0e-08
+  generation:
+    backend: trtllm
+    max_new_tokens: 512
+    stop_token_ids:
+    - 151643
+    - 151645
```
yuki-97, line 43: Heads-up, pre-existing and no action needed in this PR: these lines cannot take effect on a gym run.

[`setup_nemo_gym_config`](https://github.com/NVIDIA-NeMo/RL/blob/7671c1bee60a59d51f4b6edc015ea2e78fe289fb/nemo_rl/environments/nemo_gym.py#L952-L954) clears both stop settings unconditionally, for every backend, after its backend branch:

```python
    # Stop strings or token ids are not supported
    generation_config["stop_strings"] = None
    generation_config["stop_token_ids"] = None
```

It is the only writer of that value, all three gym entrypoints go through it (`run_grpo_nemo_gym.py:193`, `run_distillation_nemo_gym.py:107`, `run_grpo_single_controller.py:145`), and nothing repopulates it afterwards — `configure_generation_config` runs earlier, at `run_grpo_nemo_gym.py:180`. So under gym the worker always sees `None`, whatever the recipe asks for.

The values are still covered by the server-side fallbacks: `trtllm_http_server.py:154` adds `tokenizer.eos_token_id` (151645) and `:161` adds the `generation_config.json` eos list, which for Qwen3-0.6B is `[151645, 151643]` — both of the ids here. So nothing breaks; the three lines just don't do anything. Worth deciding whether to keep them as documentation of intent or drop them, and worth knowing when reading the recipe.

yuki-97 (commented):
**Scope:** the three production fixes (stop-token threading in `replace_prefix_tokens`, TRT-LLM reasoning-parser state, stop-token retention), the new vLLM/TRT-LLM parity harness and its committed golden, and the new TRT-LLM gym nightly.

**Checked:**

- Splice semantics — hand-derived old vs new `replace_prefix_tokens` across the EOS / non-EOS-stop / max-tokens cases. The token the model side stops cutting and the one the template side now skips are the same `eos_token_id`, so the EOS path is byte-identical to before; only the non-EOS-stop path changes, which is the fix.
- Upstream API usage — `ReasoningParserFactory`, both Qwen3 tool parsers and `try_get_generation_config` read at the pinned `tensorrt-llm==1.3.0rc21` and vLLM versions rather than from docstrings.
- Golden provenance — the fixture is vLLM-generated (`source_backend: "vllm"`) and written only by the vLLM lane, so the cross-backend comparison lives in the TRT-LLM test while the vLLM one is a self-regression against the same file.
- Coverage — each new test resolved against the shard that actually collects it, and the golden decoded turn by turn.
- Recipe layer — inheritance, node counts, suite bookkeeping, and the runnability of the new nightly end to end.

**Remaining comments** are on the new nightly (tool-parser choice, `in_flight_weight_updates`, an unread `policy.optimizer` block, and a request for curves), two production asks (the widened stop-token set and an aliased return), one coverage gap on the reasoning-state inference, the golden's size, two re-opened threads, and three stale-text nits.

PR #3427, opened 2026-07-30
feat: add single controller doc and gb200 nightly tests

yuki-97:
## Summary
- Add single-controller user guide (`docs/guides/single-controller.md`) and wire it into `docs/index.md`.
- Add two GB200 nightly tests + recipes: (validated both passed)
  - `grpo-llama3.1-8b-instruct-2n4g-async-1off-single-controller-streaming2` (async streaming)
  - `grpo-qwen2.5-math-1.5b-instruct-1n4g-megatron-single-controller-sync` (megatron sync)
- Minor tweaks to the paired 2n8g / 1n8g recipes.


docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### 3. `RolloutManager.generate_and_push` (`nemo_rl/experience/rollout_manager.py`)
+
+- One entry point per prompt group: reserve a buffer slot, drive the rollout via `AsyncRolloutImpl` or `AsyncNemoGymRolloutImpl`, and commit with the observed weight versions.
+- `env_handles` provide per-task environments to the rollout implementations.
+
+#### 4. `StalenessSampler` (`nemo_rl/algorithms/async_utils/staleness_sampler.py`)
+
+- Filter-only prompt-group selector over `TQReplayBuffer`. The base `PromptGroupSampler` protocol defines `admit`, `select`, and `evict`.
+- `WindowedSampler`, `WeightFifoSampler`, `InOrderSampler` are the built-in policies (one per row in the [Sampler modes](#sampler-modes) table). `CustomSampler` imports a user-supplied class by FQN.
```
terrykong: `CustomSampler` isn't a symbol — and it sits in a list with `WindowedSampler` ([L203](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/staleness_sampler.py#L203)), `WeightFifoSampler` ([L304](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/staleness_sampler.py#L304)) and `InOrderSampler` ([L342](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/staleness_sampler.py#L342)), which *are* real classes, so a reader will grep for it and find nothing. The `custom` mode is [`CustomSamplerConfig`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/staleness_sampler.py#L415) carrying `target`, with [`create_sampler`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/staleness_sampler.py#L488) doing the FQN import and the `isinstance(..., PromptGroupSampler)` check.

```suggestion
- `WindowedSampler`, `WeightFifoSampler`, `InOrderSampler` are the built-in policies (one per row in the [Sampler modes](#sampler-modes) table). The `custom` mode (`CustomSamplerConfig.target`) makes `create_sampler` import a user-supplied class by FQN and type-check it against `PromptGroupSampler`.
```

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
+# Train with Single-Controller (Async GRPO)
+
+The Single-Controller (SC) path is an alternative async GRPO runtime that runs rollout generation and policy training as two independent *pumps* coordinated by a single Ray actor (`SingleControllerActor`) sitting over a shared TransferQueue (TQ) data plane. Compared to the legacy async GRPO in [async-grpo.md](./async-grpo.md), SC decouples per-prompt rollouts from the per-step batch boundary: producers push finished rollouts into `TQReplayBuffer` at group granularity, and a pluggable `StalenessSampler` decides which groups the trainer consumes on each step.
```
terrykong, line 7: maybe add a small github admonition warning that this is a beta feature and is still being actively developed and does not yet have a stable API. github issues filed are welcome. (Something of that nature)

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+## Async-RL Knobs and Sampler Modes
+
+All SC async-RL runtime knobs live under `async_rl:` in the master config. The most important choice is the `sampler`, which sets the staleness policy shared by the rollout pump (how far it may run ahead) and the train pump (which groups it may consume).
+
+### Sampler modes
+
+The sampler is a discriminated union over `sampler.name`. Four modes are shipped:
+
+
+| `sampler.name` | Rollout gating                                                                                                          | Train selection                                                                                                   | Typical use                                                                                                  |
```
terrykong, line 69: A picture would help a lot here — this table is the hardest thing in the guide to hold in your head, because the three modes key on *different fields* of the buffer rather than being three settings of one rule.

I drew two you're welcome to **screenshot straight into the docs or just take inspiration from** — no attribution needed, and no obligation to use either:

- **[Same buffer, four different training batches](https://terrykong.github.io/gh-pages-poc/terryk/pr-3427-sc-samplers.html)** — six prompt groups finishing at different times, one decision point, and the four modes each selecting a different set. This is the one I'd reach for.
- [Buffer snapshot view](https://terrykong.github.io/gh-pages-poc/terryk/pr-3427-sc-samplers-alt.html) — an alternate framing of the same rules, better for reading the `select` predicate exactly.

Each is one self-contained SVG with every code reference permalinked to `db0aa28`. They render light/dark — **use light theme for a docs screenshot.**

The point the first figure tries to make, in case it's useful even if the figure isn't: `windowed` and `weight_fifo` select on `start_weight` (recorded at dispatch), while `in_order` selects on the `target_step` stamp assigned at `admit`. Because `start_weight` is stamped at [`reserve`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/replay_buffer.py#L683) and therefore only increases down the buffer, the three usually pick the *same* groups — they come apart only when rollouts finish out of order, which is exactly the case the figure draws.

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+| ---------------- | ---------------------------------------- | ------------------------------------------------------------------------------------ |
+| Entrypoint       | `run_grpo.py`                            | `run_grpo_single_controller.py`                                                      |
+| Data-plane       | Direct actor RPC                         | TransferQueue (`data_plane.enabled: true` required)                                  |
+| Rollout batching | Full-batch `AsyncTrajectoryCollector`    | Per-prompt `RolloutManager.generate_and_push` into a group-granular `TQReplayBuffer` |
+| Staleness policy | Single knob (`max_trajectory_age_steps`) | Pluggable `StalenessSampler` (`in_order` / `weight_fifo` / `windowed` / `custom`)    |
+| Batch boundary   | Sampled by target weight                 | Sampler-defined; can decouple rollout dispatch from train batch (streaming)          |
+
+
+If you want a straight port of legacy semantics, use `sampler.name: in_order` with `max_lookahead_versions` matching your old `max_trajectory_age_steps` and `min_groups_for_streaming_train: ${grpo.num_prompts_per_step}`.
```
terrykong: [`docs/guides/single-controller.md:142`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/docs/guides/single-controller.md#L142)

**1 action item.** This is the only migration guidance in the guide, and it covers just the staleness knob. A legacy-config translation table would earn its keep here, because there's a trap this section doesn't warn about:

**SC ignores `grpo.async_grpo:` entirely.** `grep -rn async_grpo` across [`single_controller.py`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller.py), [`single_controller_utils/`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller_utils/config.py) and [`run_grpo_single_controller.py`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/examples/run_grpo_single_controller.py) returns zero hits — SC reads only `async_rl:`. And because the loader is plain `OmegaConf.merge` with no `set_struct` and the configs are `extra="allow"`, a leftover `async_grpo` block is **silently inert rather than an error**.

That isn't hypothetical — the SC llama recipe in this PR inherits one. [`performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml:3-6`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/examples/configs/recipes/llm/performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml#L3) contributes `async_grpo.enabled: true`, `max_trajectory_age_steps: 1` and `in_flight_weight_updates: true` into the resolved config, all dead. The recipe even carries the comment `# Matches grpo.async_grpo.max_trajectory_age_steps=1` next to `max_staleness_versions: 1` — so you clearly knew, but a reader of that YAML sees two async blocks with no way to tell which is live.

**Action:** add a "Migrating a legacy async config" subsection here. Seed below — please sanity-check the last row, I couldn't determine SC's actual behaviour for `in_flight_weight_updates` from the code:

```suggestion
### Migrating a legacy async config

SC reads its async knobs from `async_rl:` and **ignores `grpo.async_grpo:` entirely**. Because the config loader accepts unknown keys, a leftover `async_grpo` block is silently inert rather than an error — delete it when porting, or you will be reading knobs that do nothing.

| Legacy `grpo.async_grpo.*` | SC equivalent |
| -------------------------- | ------------- |
| `enabled: true` | *(none)* — SC is selected by the entrypoint (`run_grpo_single_controller.py`), plus `data_plane.enabled: true` |
| `max_trajectory_age_steps: N` | `async_rl.sampler.name: in_order` with `max_lookahead_versions: N` |
| `recompute_kv_cache_after_weight_updates` | `async_rl.recompute_kv_cache_after_weight_updates` (same meaning) |
| `in_flight_weight_updates` | *(none)* — SC always syncs weights at the end of the train step |

SC-only knobs with no legacy counterpart: `min_groups_for_streaming_train` (set to `${grpo.num_prompts_per_step}` for legacy batch semantics), `max_inflight_prompts`, `max_buffered_rollouts`.
```

Optional follow-up, not for this PR: since a stray `async_grpo` block under `run_grpo_single_controller.py` is always a user error, the entrypoint could warn or raise on it next to the existing `data_plane.enabled` check at [run_grpo_single_controller.py:88](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/examples/run_grpo_single_controller.py#L88) — cheap fail-loud for a silent-misconfiguration path.

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+| Streaming, gated dispatch        | `weight_fifo`  | `max_staleness_versions: >= 1` | `x <= num_prompts_per_step`      | `num_prompts_per_step × (max_staleness_versions + 1)` | / |
+| Streaming, over-sampled          | `windowed`     | `max_staleness_versions: >= 1` | `x <= num_prompts_per_step`      | Larger than the gated capacity (dispatch is ungated)  | [`grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml`](../../examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml) |
+
+
+Field definitions:
+
+- `max_buffered_rollouts` — hard cap on unconsumed rollout groups buffered in the data plane. Validated at setup against the gated sampler's required capacity; a value too small deadlocks the rollout pump, so setup raises instead of silently blocking.
+- `min_groups_for_streaming_train` — minimum ready groups the trainer waits for before dispatching a batch. Set to `num_prompts_per_step` for sync/legacy semantics; lower for streaming.
+
+## Implementation Structure
```
terrykong, line 95: [`docs/guides/single-controller.md:87`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/docs/guides/single-controller.md#L87)

**1 action item.** Consider adding diagrams to this guide. I drew three you're welcome to **screenshot directly or just take inspiration from** — no attribution needed, and no obligation to use any of them:

| | figure | for which section |
|---|---|---|
| 1 | **[Ownership & call model](https://terrykong.github.io/gh-pages-poc/terryk/pr-3427-sc-ownership.html)** | `## Implementation Structure` (this section) |
| 2 | **[Same buffer, four different training batches](https://terrykong.github.io/gh-pages-poc/terryk/pr-3427-sc-samplers.html)** | `### Sampler modes` |
| 3 | [Buffer snapshot view](https://terrykong.github.io/gh-pages-poc/terryk/pr-3427-sc-samplers-alt.html) *(alternate framing of #2)* | `### Sampler modes` |

Each is one self-contained SVG sized to screenshot into the docs, with every code reference as a permalink pinned to `db0aa28`. #2 is the one I'd reach for first. They render light/dark — **use light theme for a docs screenshot.**

**Why a figure earns its keep in this section.** The "Core components" list reads as five peer classes, but they sit at three levels, and that difference is what tells a reader which arrows cross a process boundary:

- [`SingleControllerActor`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller.py#L73) is `@ray.remote(num_cpus=1, num_gpus=0)` — a **CPU-only** process. Its docstring says it better than the prose does: *"All other actors are passive — they expose methods and wait to be called."*
- `TQReplayBuffer`, `RolloutManager` and the sampler are **plain objects inside that actor**, cloudpickled in via [`SingleControllerActorArgs`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller_utils/setup.py#L68). Reaching them is a direct call, not RPC.
- Only [`TQPolicy`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller_utils/setup.py#L210) and the [generation backend](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller_utils/setup.py#L171) are GPU worker groups — the only hops that go over the wire.

Today a reader infers all of that from the `@ray.remote` decorator plus the shape of the args dataclass.

Two invariants worth surfacing wherever they fit, because they're load-bearing and currently invisible:

1. **Why the driver builds everything.** The guide explains the actor "does no construction work ... because running setup inside a nested Ray actor breaks `runtime_env` resolution" — but not the exception, which is more instructive: [`Logger`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller.py#L124) *is* built inside the actor, because wandb/TB backends hold a `_thread.lock` that cloudpickle can't serialize.
2. **The buffer rebind.** [`single_controller.py:120-122`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller.py#L120) does `self._rollout_manager._tq_buffer = self._buffer` because `rollout_manager` and `tq_buffer` are separate fields of the args dataclass, so Ray deserializes them as *two different* buffers — without the rebind the writer and the sampler hold separate copies and the sampler never sees committed groups.

**Separately — one small wording fix in this section.** Under `TQReplayBuffer` the guide says *"On `commit`, the producer tensorizes the prompt group ... and records the `start_weight` and `end_weight` versions."* `start_weight` is actually recorded at [`reserve`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/replay_buffer.py#L683), at dispatch; [`commit`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/replay_buffer.py#L746) only sets `end_weight_list[idx]` and flips `ready_list[idx] = True` (it takes `start_weight_version`, but uses it for [tensorization](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/replay_buffer.py#L721), not the slot).

That detail matters more than it looks: because slots are appended at *reserve*, buffer index order is **dispatch order, not completion order**, and `start_weight` only ever increases down the buffer. That's precisely what makes `windowed`'s "buffer order" selection and [`weight_fifo`'s](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/staleness_sampler.py#L315) "oldest weight" selection usually agree — they diverge only when rollouts finish out of order. Worth one clarifying clause here.

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+        vllm_cfg:
+          async_engine: true
+        colocated:
+          enabled: false
+          resources:
+            num_nodes: 1
+            gpus_per_node: 4  # inference GPUs; remainder go to training
+    ```
+
+3. **One RL step = one optimizer step.** The SC train pump does not support multi-mini-step inside a single RL step (see the validator in `single_controller_utils/config.py`):
```
terrykong: Path doesn't resolve from anywhere — the file is `nemo_rl/algorithms/single_controller_utils/config.py` ([`validate_single_controller_config`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/single_controller_utils/config.py#L87)). Every other component reference in this guide uses the full repo path.

```suggestion
3. **One RL step = one optimizer step.** The SC train pump does not support multi-mini-step inside a single RL step (see `validate_single_controller_config` in [nemo_rl/algorithms/single_controller_utils/config.py](../../nemo_rl/algorithms/single_controller_utils/config.py)):
```

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+      use_importance_sampling_correction: true
+    ```
+
+## Async-RL Knobs and Sampler Modes
+
+All SC async-RL runtime knobs live under `async_rl:` in the master config. The most important choice is the `sampler`, which sets the staleness policy shared by the rollout pump (how far it may run ahead) and the train pump (which groups it may consume).
+
+### Sampler modes
+
+The sampler is a discriminated union over `sampler.name`. Four modes are shipped:
```
terrykong: "Discriminated union" is pydantic vocabulary; it doesn't tell a reader what to put in their YAML. The user-visible fact is that each mode has its **own** knobs — `max_lookahead_versions` isn't a field that exists under `name: windowed` — which is what the table's "Sampler knob" column already encodes.

```suggestion
Pick one of four modes with `sampler.name`. Each mode takes its own knobs, listed below — a knob from one mode has no effect under another:
```

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+| `windowed`     | Ungated — rollout keeps producing until the buffer fills.                                                               | Take any ready group with `start_weight` in `[train - max_staleness_versions, train]`, optionally freshest-first. | Over-sampled streaming; aged groups outside the window are evicted (wasted compute).                         |
+| `custom`       | Determined by the imported class.                                                                                       | Determined by the imported class.                                                                                 | `target: "module:ClassName"` — bring your own `PromptGroupSampler`.                                          |
+
+
+### Config → behavior map
+
+The exemplars shipped in this PR stack cover the four canonical setups:
```
terrykong: "in this PR stack" will read oddly once merged. Also worth softening "the four canonical setups": only three ship an exemplar — `weight_fifo` has none anywhere in the repo (its only references are unit tests, e.g. `tests/unit/single_controller/test_sampler_interface.py:113`).

```suggestion
The shipped exemplars cover three of the four modes:
```

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+| -------------------------------- | -------------- | ------------------------------ | -------------------------------- | ----------------------------------------------------- | -------- |
+| Sync / on-policy                 | `in_order`     | `max_lookahead_versions: 0`    | `${grpo.num_prompts_per_step}`   | `num_prompts_per_step × 1`                            | [`grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync.yaml`](../../examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync.yaml) |
+| Async, exact batch→step matching | `in_order`     | `max_lookahead_versions: >= 1` | `x <= num_prompts_per_step`   | `num_prompts_per_step × (max_lookahead_versions + 1)` | [`grpo_math_1B_megatron_single_controller.yaml`](../../examples/configs/grpo_math_1B_megatron_single_controller.yaml) |
+| Streaming, gated dispatch        | `weight_fifo`  | `max_staleness_versions: >= 1` | `x <= num_prompts_per_step`      | `num_prompts_per_step × (max_staleness_versions + 1)` | / |
```
terrykong: The Exemplar cell is a bare `/`, which renders as a stray slash rather than reading as "none".

```suggestion
| Streaming, gated dispatch        | `weight_fifo`  | `max_staleness_versions: >= 1` | `x <= num_prompts_per_step`      | `num_prompts_per_step × (max_staleness_versions + 1)` | — (none shipped) |
```

docs/guides/single-controller.md
```diff
@@ -0,0 +1,152 @@
...
+- Group-granular replay buffer with reserve/commit slot accounting.
+- On `commit`, the producer tensorizes the prompt group into N training-shaped rows and records the `start_weight` and `end_weight` versions the rollout observed.
+- Tracks `target_step` per group when the sampler assigns one at admit time (used by `in_order`).
+
+#### 3. `RolloutManager.generate_and_push` (`nemo_rl/experience/rollout_manager.py`)
+
+- One entry point per prompt group: reserve a buffer slot, drive the rollout via `AsyncRolloutImpl` or `AsyncNemoGymRolloutImpl`, and commit with the observed weight versions.
+- `env_handles` provide per-task environments to the rollout implementations.
+
+#### 4. `StalenessSampler` (`nemo_rl/algorithms/async_utils/staleness_sampler.py`)
```
terrykong: This header has the same `<Class> (<module path>)` shape as the three above it, which all name real symbols — but there's no `StalenessSampler` class. The protocol is [`PromptGroupSampler`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/staleness_sampler.py#L64) and the shared base is [`BaseSampler`](https://github.com/NVIDIA-NeMo/RL/blob/db0aa2800b74261f3961c5b01ae961832401f9d8/nemo_rl/algorithms/async_utils/staleness_sampler.py#L108). (The generic prose uses at L3 and L138 read fine as a concept — no change needed there.)

```suggestion
#### 4. Samplers (`nemo_rl/algorithms/async_utils/staleness_sampler.py`)
```

terrykong (commented):
Reviewed with a 4-agent pass (docs/config, tests, config-correctness deep dive, adversarial verification). **No blockers.** Everything inline is a documentation improvement — the code, configs and test scripts are sound.

Most of the doc comments are one-click `suggestion` blocks. Two are asks rather than edits: a legacy→SC config translation table, and an offer of diagrams.

Verified clean, for the record:

- **Config correctness.** Both new recipes were fully resolved through their inheritance chains and evaluated against every SingleController validator: `num_prompts_per_step × num_generations_per_prompt == train_global_batch_size` (2048 and 512), `required_buffer_capacity_for_config` (the qwen recipe sits exactly at the boundary, 32 vs 32 — passes), `min_groups_for_streaming_train`, the checkpointing/validation guards, and `data_plane.enabled`. GPU splits are non-degenerate (llama 4 train + 4 inference; qwen 2 + 2) and all parallelism resolves to TP1/PP1, so divisibility holds throughout.
- **Nightly scripts.** The only delta from their merged 8-GPU siblings is the added `GPUS_PER_NODE=4` — which `test_nightly_suites_match_gpus_per_node` *requires* for `nightly_gb200.txt` entries. Naming, suite placement, file modes and copyright exemption all check out, and the yaml↔sh↔suite 1:1 invariants hold.
- **Renames.** No dangling references to the old `-sc` names anywhere in the repo; `log_dir`, `wandb.name` and `checkpoint_dir` were updated together in all four files.
- We looked hard at the `timing/train/total_step_time < 25` bound being carried onto half the training GPUs and concluded it is **fine** — the metric includes generation wait, the right comparison is 4×GB200 vs 8×H100, and the already-shipped `grpo-qwen2.5-math-1.5b-instruct-1n4g-fsdp2tp1.v3.sh` keeps the same bound with *less* walltime and passes nightly. Noting it only so you know it was considered rather than missed.

Two things deliberately **not** raised as asks here, in case they're useful as follow-ups: `nightly_gb200.txt` has no GPU-hour budget guard (`test_nightly_compute_stays_below_3480_hours` covers only `nightly.txt`, so the ~29 GPU-h added here is unguarded), and SC's `_build_clusters` lacks the `inference_gpus_per_node == cluster.gpus_per_node` guard that `grpo.py` has for the multi-node non-colocated case, so a mismatched config would silently strand GPUs. Both are pre-existing and untouched by this PR.

Thanks for writing the guide up — the sampler-mode and config→behavior tables are genuinely useful, and the SC path badly needed a user-facing entry point.

_Generated by Claude Code_

docs/guides/single-controller.md
yuki-97: fixed in 5020b84dfa3db56adf685e0d8fe9d43017b4c012

docs/guides/single-controller.md
yuki-97: fixed in 5020b84dfa3db56adf685e0d8fe9d43017b4c012

docs/guides/single-controller.md
yuki-97: fixed in 5020b84dfa3db56adf685e0d8fe9d43017b4c012

docs/guides/single-controller.md
yuki-97: fixed in 5020b84dfa3db56adf685e0d8fe9d43017b4c012

docs/guides/single-controller.md
yuki-97: fixed in 5020b84dfa3db56adf685e0d8fe9d43017b4c012

docs/guides/single-controller.md
yuki-97: fixed in 5020b84dfa3db56adf685e0d8fe9d43017b4c012

docs/guides/single-controller.md
yuki-97: fixed in 5020b84dfa3db56adf685e0d8fe9d43017b4c012

docs/guides/single-controller.md
yuki-97: fixed in 8363fabe889fc6380cf1c75321fd2afcfc5d6341 + 3eccfc2

docs/guides/single-controller.md
yuki-97: fixed in b696ee5d84de0fe520ec52601ebcbeb0dfb00f3c

docs/guides/single-controller.md
yuki-97: fixed in bb4c762b170247cc6279d0073a2247356eaad5d7

terrykong (dismissed):

docs/guides/single-controller.md
```diff
@@ -0,0 +1,177 @@
...
+| Entrypoint       | `run_grpo.py`                            | `run_grpo_single_controller.py`                                                      |
+| Data-plane       | Direct actor RPC                         | TransferQueue (`data_plane.enabled: true` required)                                  |
+| Rollout batching | Full-batch `AsyncTrajectoryCollector`    | Per-prompt `RolloutManager.generate_and_push` into a group-granular `TQReplayBuffer` |
+| Staleness policy | Single knob (`max_trajectory_age_steps`) | Pluggable `StalenessSampler` (`in_order` / `weight_fifo` / `windowed` / `custom`)    |
+| Batch boundary   | Sampled by target weight                 | Sampler-defined; can decouple rollout dispatch from train batch (streaming)          |
+
+
+### Migrating a legacy async config
+
+SC reads its async knobs from `async_rl:` and **ignores `grpo.async_grpo:` entirely**.
```
terrykong: **1 action item.** This sentence now contradicts the code added in this PR. [`run_grpo_single_controller.py:88`](https://github.com/NVIDIA-NeMo/RL/blob/d4b148578140206b987cbc8c809f4e06b4092bab/examples/run_grpo_single_controller.py#L88) raises `SC requires \`grpo.async_grpo: null\`` — so the block is no longer *ignored*, it's **rejected**. A reader of this migration section concludes they can leave the legacy block in place, then hits a hard error on the first run. (Both halves are new here — my earlier comment described the old silently-ignored behaviour, and the guard changed it.)

```suggestion
SC reads its async knobs from `async_rl:` and **requires `grpo.async_grpo: null`** — `run_grpo_single_controller.py` raises if a legacy block is still present, so null it out when porting rather than leaving it in place.
```

examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync.yaml
```diff
@@ -2,17 +2,21 @@
...
+  log_dir: logs/grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync
   wandb:
-    name: grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-sc
+    name: grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync
 
 # SC does not support validation and checkpointing yet.
 grpo:
+  # Null out the legacy async_grpo block inherited from
+  # performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml: SC reads
+  # async_rl instead and would silently ignore any values left here.
```
terrykong: **1 action item.** The provenance in this comment is a copy-paste from the llama recipe — this config inherits `async_grpo` from `grpo-qwen2.5-math-1.5b-instruct-1n8g-fsdp2tp1.v3.yaml` → `grpo_math_1B.yaml`, not from `performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml`. Also, *"would silently ignore any values left here"* is no longer true in any of the four configs — [the entrypoint now raises](https://github.com/NVIDIA-NeMo/RL/blob/d4b148578140206b987cbc8c809f4e06b4092bab/examples/run_grpo_single_controller.py#L88) instead of ignoring, which is the whole point of the guard.

```suggestion
  # Null out the legacy async_grpo block inherited from grpo_math_1B.yaml:
  # SC reads async_rl instead, and the entrypoint raises if this block is set.
```

examples/configs/grpo_math_1B_megatron_single_controller.yaml
```diff
@@ -3,6 +3,10 @@ defaults: grpo_math_1B.yaml
 
 # SC does not support validation yet.
 grpo:
+  # Null out the legacy async_grpo block inherited from
+  # performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml: SC reads
+  # async_rl instead and would silently ignore any values left here.
```
terrykong: **1 action item.** The provenance in this comment is a copy-paste from the llama recipe — this config inherits `async_grpo` from `grpo_math_1B.yaml` (its direct parent), not from `performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml`. Also, *"would silently ignore any values left here"* is no longer true in any of the four configs — [the entrypoint now raises](https://github.com/NVIDIA-NeMo/RL/blob/d4b148578140206b987cbc8c809f4e06b4092bab/examples/run_grpo_single_controller.py#L88) instead of ignoring, which is the whole point of the guard.

```suggestion
  # Null out the legacy async_grpo block inherited from grpo_math_1B.yaml:
  # SC reads async_rl instead, and the entrypoint raises if this block is set.
```

examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
```diff
@@ -6,6 +6,10 @@ defaults: ./grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async.yaml
 
 # SC does not support validation and checkpointing yet.
 grpo:
+  # Null out the legacy async_grpo block inherited from
+  # performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml: SC reads
+  # async_rl instead and would silently ignore any values left here.
```
terrykong: **1 action item.** The provenance in this comment is a copy-paste from the llama recipe — this config inherits `async_grpo` from `grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async.yaml` (its direct parent, which sets `async_grpo:` at line 3), not from `performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml`. Also, *"would silently ignore any values left here"* is no longer true in any of the four configs — [the entrypoint now raises](https://github.com/NVIDIA-NeMo/RL/blob/d4b148578140206b987cbc8c809f4e06b4092bab/examples/run_grpo_single_controller.py#L88) instead of ignoring, which is the whole point of the guard.

```suggestion
  # Null out the legacy async_grpo block inherited from grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async.yaml:
  # SC reads async_rl instead, and the entrypoint raises if this block is set.
```

examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml
```diff
@@ -2,17 +2,21 @@
...
+  log_dir: logs/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2
   wandb:
-    name: grpo-llama3.1-8b-instruct-2n8g-async-1off-sc
+    name: grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2
 
 # SC does not support validation and checkpointing yet.
 grpo:
+  # Null out the legacy async_grpo block inherited from
+  # performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml: SC reads
+  # async_rl instead and would silently ignore any values left here.
```
terrykong: **1 action item.** Provenance is right here, but the second clause isn't. Also, *"would silently ignore any values left here"* is no longer true in any of the four configs — [the entrypoint now raises](https://github.com/NVIDIA-NeMo/RL/blob/d4b148578140206b987cbc8c809f4e06b4092bab/examples/run_grpo_single_controller.py#L88) instead of ignoring, which is the whole point of the guard.

```suggestion
  # Null out the legacy async_grpo block inherited from performance/grpo-llama3.1-8b-instruct-2n8g-async-1off.yaml:
  # SC reads async_rl instead, and the entrypoint raises if this block is set.
```

terrykong (commented):
Re-reviewed after the fixes — **all 10 previous threads verified fixed against the commits, no blockers.** I also confirmed the new `async_grpo` guard doesn't break anything: the `null` propagates correctly through inheritance into both new GB200 recipes (I resolved both chains — `async_grpo=None`), and both SC functional tests still pass it (`grpo_dp_single_controller.sh` picks up the default SC config, and the gym one was patched in d4b148578140206b987cbc8c809f4e06b4092bab).

Everything here is small and follows *from* those fixes — every comment is a one-click suggestion.

- **`single-controller.md:157`** is the one I'd actually push for: it still says SC *ignores* `grpo.async_grpo:`, but the new guard **rejects** it, so the migration section now contradicts the runtime.
- The four config comments each had a copy-pasted provenance (only the llama one named its real parent) and all four claimed SC "would silently ignore any values left here" — which is exactly what the guard removed.

_Generated by Claude Code_

terrykong (dismissed):

RayenTian (approved):

PR #3429, opened 2026-07-30
feat(sc): add checkpoint save/restore to SingleController

haitian-nvidia:
# What does this PR do ?

Adds checkpoint save/resume to the SingleController path, closing the
`NotImplementedError` placeholder left in `setup_single_controller`.

**Save** (in the train pump, after each weight sync):
- Policy weights via `policy.save_checkpoint`, and optimizer state when
  `checkpointing.save_optimizer` is set. Megatron `async_save` is
  supported: the tmp→step rename is deferred with
  `begin_finalization(wait_fn=policy.finalize_async_save)` and flushed at
  the next save or on exit, so training/rollouts continue while the weight
  write completes in the background.
- Training state (`training_info.json`), dataloader position
  (`train_dataloader.pt`), and — when the sampler supports it
  (windowed/over-sampled) — the TQ replay buffer's committed prompt groups
  (`replay_buffer.pt`, meta + DataPlane payloads).
- Triggers: `save_period` boundary, last step, and
  `checkpoint_must_save_by` (timeout save + early stop).
  `latest_checkpoint_status.json` is refreshed after each save.

**Resume**:
- `setup_single_controller` resolves the latest checkpoint, loads
  `training_info.json`, feeds `weights_path`/`optimizer_path` to the
  trainer, and restores the dataloader position (with the existing
  dataset-swap guard).
- `SingleControllerActor` restores its counters, seeds the sampler
  dispatch cursor via the new `resume_from_step` argument, and reloads
  replay-buffer groups (re-acquiring one buffer-capacity permit per
  group) before the pumps start.
- New sampler surface: `create_sampler(..., resume_from_step=...)` and a
  `supports_buffer_checkpoint` property on `PromptGroupSampler` (True for
  the windowed sampler; gated samplers skip buffer save/restore).

# Usage

```bash
uv run examples/run_grpo_single_controller.py \
    checkpointing.enabled=true \
    checkpointing.checkpoint_dir=results/my-run \
    checkpointing.save_period=10
# Re-running the same command resumes from the latest step_N checkpoint.
```

**Pre checks**:

# Additional Information

- Unit tests: `tests/unit/single_controller/` — 108 passed, including 32
  new tests in `test_sc_checkpointing.py` (save triggers, async-save
  finalization ordering and failure propagation, metric_name handling,
  dataloader round-trip, replay-buffer persistence, setup resume wiring).
- E2E on GB200 (Qwen3-0.6B, megatron backend, `async_save: true`):
  a 4-step run saves complete `step_2`/`step_4` checkpoints with no
  `tmp_step_*` leftovers; a `checkpoint_must_save_by` run saves and stops
  early; resuming restores the dataloader + replay buffer and continues
  to step 4.
- No documentation changes needed (config surface reuses the existing
  `checkpointing` block).


tests/unit/single_controller/test_tq_replay_buffer.py
```diff
@@ -363,3 +382,200 @@ def test_size_and_len(self):
...
+        assert [c["sample_ids"] for c in dp.get_calls] == [
+            list(metas[0].sample_ids),
+            list(metas[1].sample_ids),
+        ]
+        assert dp.get_calls[0]["select_fields"] == list(metas[0].fields)
+        assert state["groups"][0]["fields_data"] == {
+            "payload_for": list(metas[0].sample_ids)
+        }
+
+    def test_round_trip_restores_lists_and_rows(self):
```
yuki-97, line 547: Round-trip only exercises `start_weight == end_weight` and `target_step is None`. `_add_group(weight=w)` defaults `end_weight=weight`, and `commit` never sets `target_step` — so a load bug that swapped `start_weight` / `end_weight` at `replay_buffer.py:1002-1003` or silently dropped `target_step` (the `InOrderSampler`'s primary key at `staleness_sampler.py:431`) would still pass.

**Suggest fix:** parameterize with `start=1, end=2` on one group and `target_step=7` on another.

tests/unit/single_controller/test_tq_replay_buffer.py
```diff
@@ -98,6 +99,24 @@ def clear_samples(self, sample_ids: list[str] | None, partition_id: str) -> None
         for sid in ids:
             self._rows.pop(sid, None)
 
+    def get_samples(
```
yuki-97, line 104: `FakeDataPlaneClient.get_samples` returns a hand-crafted `{"payload_for": sample_ids}` dict and `load_state_dict` re-puts it verbatim — but the production replay-buffer file goes through `torch.save(buffer_state, replay_buffer.pt)` on TensorDict-valued `fields_data` (see `single_controller.py:290,766`). Any bug in that real serialization path (dtype/device promotion, non-contiguous strides, custom-Tensor pickle) is uncaught here.

**Suggest fix:** one round-trip that stores an actual TensorDict payload through an in-memory `BytesIO` `torch.save`/`torch.load` cycle.

tests/unit/single_controller/test_tq_replay_buffer.py
yuki-97, line 547: Three round-trip shapes aren't covered: (a) empty buffer (common resume shape when no group was committed before the checkpoint); (b) middle-unready `[ready, unready, ready]` (`state_dict` skips unready by index; the trailing-unready test wouldn't catch an off-by-one in that skip); (c) `target_step != None` — the `InOrderSampler` resume path, which needs `_make_group_entry(target_step=7)` to construct.

tests/unit/single_controller/test_sc_checkpointing.py
```diff
@@ -0,0 +1,1083 @@
...
+        with patch("ray.cluster_resources", return_value={"GPU": 0}):
+            await actor._train_pump()
+        if flush:
+            actor._checkpointer.shutdown()
+        return actor
+
+    return asyncio.run(_main())
+
+
+def _run_actor_run(mc: MasterConfig, actor_args: SingleControllerActorArgs):
```
yuki-97, line 423: All four `_run_actor_run` call sites use `max_num_steps=0`, so `_train_pump` never runs post-restore. `TestCounterRestore` covers constructor seeding and `TestSaveTrigger` covers the fresh pump, but nothing composes the two to prove that resuming with `save_state["current_step"]=N` and running the pump yields `_train_steps = N + k` (not `k`, not `2N + k`).

**Suggest fix:** one test that primes `save_state["current_step"]=2`, runs with `max_num_steps=4`, and asserts final `_train_steps == 4`.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -352,6 +372,20 @@ def setup_single_controller(
         drop_last=True,
         num_workers=data_config["num_workers"],
     )
+    if last_checkpoint_path is not None:
+        dataloader_state_path = os.path.join(
+            last_checkpoint_path, "train_dataloader.pt"
+        )
+        if os.path.exists(dataloader_state_path):
```
yuki-97: SC guards `load_dataloader_state` with `if os.path.exists(dataloader_state_path)` and prints a warning on miss, while GRPO's `init_train_dataloader` (`grpo.py:451-452`) calls it unconditionally and lets a missing file raise. SC always writes `train_dataloader.pt` on save (line 759), so absence means externally-tampered / partial ckpt — silent-fresh-start then replays already-seen samples with no signal to the user.

**Suggest fix:** drop the `exists` guard and let `load_dataloader_state` raise, matching GRPO's contract.

nemo_rl/algorithms/async_utils/staleness_sampler.py
```diff
@@ -113,11 +123,26 @@ class BaseSampler(abc.ABC):
...
+            raise ValueError(
+                f"resume_from_step must be non-negative, got {resume_from_step}"
+            )
         self._buffer = buffer
-        # Pre-incremented before each admitted batch, so -1 lets the first
-        # batch through a zero-staleness gate.
-        self._dispatch_index: int = -1
+        # Pre-incremented before each admitted batch, so the cursor trails
+        # the run's starting step by one.
+        self._dispatch_index: int = resume_from_step - 1
```
yuki-97: `resume_from_step` is threaded through 6 signatures + a special-case `!= 0` gate for `CustomSamplerConfig` at line 561-562 that silently overrides YAML `model_extra` on resume but silently loses to it on fresh runs.

Since only `_dispatch_index` needs to be restored, wdyt replacing the kwarg with a `BaseSampler.set_dispatch_index(resume_from_step)` post-construction hook, called once from `SingleControllerActor.__init__` after `create_sampler`? Removes the 5 sub-constructor signature changes, the factory kwarg, and the `CustomSampler` asymmetry.

nemo_rl/algorithms/single_controller.py
```diff
@@ -206,7 +234,21 @@ async def run(self) -> dict[str, Any]:
...
+            except Exception:
+                if not propagating:
+                    raise
+                warnings.warn(
+                    "Checkpoint finalization failed while handling an "
+                    "exception; the original exception will be re-raised.",
+                    stacklevel=2,
+                )
+            finally:
+                self._logger.finish()
```
yuki-97: `sys.exc_info()[0] is not None` + try/except buys "shutdown errors don't mask training errors" — but Python's built-in exception chaining already preserves the original via `__context__`, and the shape reads against `skills/error-handling`'s fail-loud policy. The whole block flattens to:

```suggestion
            self._logger.finish()
            await asyncio.to_thread(self._checkpointer.shutdown)
```

nemo_rl/algorithms/single_controller.py
```diff
@@ -573,6 +675,108 @@ async def _train_pump(self) -> None:
...
+        save_state = self._save_state
+        save_state["current_step"] = self._train_steps
+        save_state["total_steps"] = self._train_steps
+        save_state["current_epoch"] = self._current_epoch
+        save_state["consumed_samples"] = self._consumed_samples
+        save_state["total_valid_tokens"] = self._total_valid_tokens
+        # Snapshot before any await so it can't interleave with
+        # _rollout_pump iterating this same dataloader.
+        dataloader_state = self._dataloader.state_dict()
+        if val_metrics is not None:
```
yuki-97: for the val part, is it make sense to you to remove it and add back when supporting val?
I'm also ok with keeping it here since val should be supported soon, so just a notice to add some related tests when supporting val.

nemo_rl/algorithms/single_controller.py
```diff
@@ -226,6 +268,44 @@ async def ping(self) -> dict[str, Any]:
...
+    async def _maybe_restore_replay_buffer(self) -> None:
+        """Restore replay-buffer groups from the previous run's checkpoint.
+
+        No-op unless the sampler supports_buffer_checkpoint (ungated only).
+        """
+        if not (
+            self._sampler.supports_buffer_checkpoint
+            and self._last_checkpoint_path is not None
+        ):
+            return
```
yuki-97: instead of `supports_buffer_checkpoint`, I think maybe better to
1. save `sampler_name` in ckpt.
2. if `sampler_name` equal with current, load buffer
3. if `sampler_name` not equal, warning and return.

something like below:
```suggestion
        if not self._last_checkpoint_path is not None:
            return

        if sampler_name != current_sampler_name:
            print(...)
            return

```

examples/configs/grpo_math_1B_megatron_single_controller.yaml
```diff
@@ -21,7 +21,6 @@ async_rl:
   # Enable per-rollout diagnostic prints (prompt content / completion previews).
   diagnostics: false
 
-# SC does not support checkpointing yet.
```
yuki-97, line 27: can we add functional tests like `tests/functional/grpo_async_replay_buffer_checkpoint.sh` and `tests/functional/sft_resume_diamond.sh`?

yuki-97 (commented):
@haitian-nvidia thanks for supporting this! left some comments.

nemo_rl/algorithms/single_controller_utils/setup.py
haitian-nvidia: Done in 26c0f926e — dropped the `os.path.exists` guard; a resume checkpoint missing `train_dataloader.pt` now raises (matching GRPO's contract). The unit test now expects `FileNotFoundError`.

nemo_rl/algorithms/async_utils/staleness_sampler.py
haitian-nvidia: Done in 26c0f926e — replaced the `resume_from_step` kwarg threading with a `BaseSampler.set_dispatch_index(resume_from_step)` post-construction hook (also on the `PromptGroupSampler` protocol), called once from `SingleControllerActor.__init__`. The factory kwarg and the custom-sampler fresh-vs-resume asymmetry are gone.

nemo_rl/algorithms/single_controller.py
haitian-nvidia: Done in 26c0f926e — flattened to `self._logger.finish()` + a bare `await asyncio.to_thread(self._checkpointer.shutdown)`, relying on exception chaining.

tests/unit/single_controller/test_tq_replay_buffer.py
haitian-nvidia: Done in cbbbe2b21 — `test_round_trip_preserves_end_weight_and_target_step` parameterizes exactly as suggested: one group with `start=1, end=2` and one with `target_step=7`, asserting all three lists survive the round-trip. Built via the real `reserve(target_step=...)` → `commit` path (rather than a hand-built envelope) so the save side of target_step serialization is exercised too.

tests/unit/single_controller/test_tq_replay_buffer.py
haitian-nvidia: Done in cbbbe2b21 — `test_round_trip_tensordict_payload_through_torch_save` stores a real TensorDict as `fields_data` (mixed int64/float32 dtypes + a non-contiguous `[:, ::2]` view) through an in-memory BytesIO `torch.save`/`torch.load(weights_only=False)` cycle (matching the production load), then asserts per-key `torch.equal` after `load_state_dict` re-puts.

tests/unit/single_controller/test_tq_replay_buffer.py
haitian-nvidia: Done in cbbbe2b21 — (a) `test_round_trip_empty_buffer`; (b) `test_state_dict_skips_middle_unready` (`[ready, unready, ready]`, asserting the neighbouring groups' fields don't shift); (c) covered by `test_round_trip_preserves_end_weight_and_target_step` (see the thread above) via the real `reserve(target_step=7)` path.

tests/unit/single_controller/test_sc_checkpointing.py
haitian-nvidia: Done in cbbbe2b21 — `test_resumed_pump_continues_to_max_steps` primes `save_state["current_step"]=2`, runs the live pump with `max_num_steps=4`, and asserts `_train_steps == 4` (not 2, not 6), plus that only the post-resume `step_4` boundary is checkpointed and `consumed_samples` accumulates on top of the restored value.

nemo_rl/algorithms/single_controller.py
haitian-nvidia: Done in 60de5aa3f, with one refinement on where the name lives: `supports_buffer_checkpoint` is gone — every save writes `replay_buffer.pt` and records `async_rl.sampler.name` into **training_info.json** (run metadata, alongside `current_step`), rather than into the buffer envelope. On resume the saved name is compared against the current sampler before the `torch.load`, so a mismatched (or pre-this-change) checkpoint warns and skips without reading the buffer file. `TQReplayBuffer` stays sampler-agnostic.

nemo_rl/algorithms/single_controller.py
haitian-nvidia: Done in 60de5aa3f — removed the val placeholder from `_save_checkpoint` (the `val_metrics` parameter and its branch); we'll add it back with the validation loop, along with tests. The stale `val_reward` sentinel from `_default_grpo_save_state` is still dropped so it never lands in training_info.json.

examples/configs/grpo_math_1B_megatron_single_controller.yaml
haitian-nvidia: Done in 7c1d6da5c — added `tests/functional/grpo_checkpoint_single_controller.sh` (registered in `L1_Functional_Tests_SingleController.sh`, full tier). Same two-run shape as `grpo_async_replay_buffer_checkpoint.sh`: run 1 uses a 1-second `checkpoint_must_save_by` so the timeout save fires deterministically at the first step boundary and stops early (also covering the timeout-save path); run 2 resumes from `step_1`, asserting the dataloader + replay-buffer restore lines and reaching `step_4`. Verified end to end on 2 GPUs (~8 min runtime).

haitian-nvidia:
@mehraakash

nemo_rl/algorithms/single_controller.py
```diff
@@ -574,6 +665,105 @@ async def _train_pump(self) -> None:
...
+        """
+        save_state = self._save_state
+        save_state.current_step = self._train_steps
+        save_state.total_steps = self._train_steps
+        save_state.current_epoch = self._current_epoch
+        save_state.consumed_samples = self._consumed_samples
+        save_state.total_valid_tokens = self._total_valid_tokens
+        # The restore skips the replay buffer when the resuming run uses a
+        # different sampler (its stamps may never be selectable there).
+        save_state.sampler_name = self._async_cfg.sampler.name  # type: ignore[attr-defined]
```
yuki-97: `save_state.sampler_name = ...  # type: ignore[attr-defined]` here attaches a runtime-only field to `GRPOSaveState`. Because the dataclass doesn't declare it, `_get_grpo_save_state` (`grpo.py:315-318`) filters it out on load and `single_controller_utils/setup.py:340-341` has to hand-re-attach with a second `# type: ignore`.

Two silent-drop paths across three files: tightening `_get_grpo_save_state` to strict extraction, or forgetting the re-attach on a future SC-only save-state helper, silently disables the sampler-name gate and restores the buffer under the wrong sampler.

**Suggest fix:** add `sampler_name: Optional[str] = None` to `GRPOSaveState`, default in `_initial_grpo_save_state`, drop both `# type: ignore` sites + the setup.py re-attach block.

nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -836,6 +836,194 @@ async def remove(self, idxs: list[int], remove_in_dp: bool) -> int:
...
+            print(
+                "TQReplayBuffer capacity changed: "
+                f"checkpoint={state['saved_capacity']}, current={max_groups}. "
+                "Using current config value."
+            )
+        num_truncated = 0
+        if len(groups) > max_groups:
+            num_truncated = len(groups) - max_groups
+            # Keep the freshest max_groups groups, preserving original order.
+            prioritized = sorted(
```
yuki-97, line 1055: **[BUG]** `sorted(range(...), key=(start_weight, i))[num_truncated:]` (freshest-first) is only safe for staleness samplers whose `_stamp` returns None. For InOrderSampler, `start_weight` and `target_step` are both monotonic, so freshest-first keeps the largest target (far-future) and drops the near-term ones. `InOrderSampler.evict` only drops `target < current_train_weight`, so far-future permits stay held; rollout blocks on `_buffer_capacity.acquire()` while pump stalls waiting for the near-term target → silent deadlock. Triggered when `max_buffered_rollouts` shrinks across save/resume.

Keep-smallest doesn't really help either — rollout would just re-produce the dropped in-order targets from the restored dataloader position (see the separate note on `set_dispatch_index`).

**Suggest fix:** when any loaded group carries `target_step` (InOrder-family), assert loud on `len(groups) > max_groups` rather than truncate — there's no meaningful "keep half" for shrunk-capacity in-order resume. Non-InOrder samplers: current freshest-first is fine.

tests/unit/single_controller/test_sc_checkpointing.py
```diff
@@ -0,0 +1,1095 @@
...
+        assert not (ckpt_dir / "step_2").exists()
+
+        trainer.release.set()
+        actor._checkpointer.shutdown()
+
+        assert (ckpt_dir / "step_2").is_dir()
+        assert not (ckpt_dir / "tmp_step_2").exists()
+        assert trainer.finalize_calls == 1
+
+    def test_failed_background_finalization_raises_at_next_save(self, tmp_path):
```
yuki-97, line 725: `test_failed_background_finalization_raises_at_next_save` covers only the next-save path. The prod comment at `single_controller.py:757-758` says the async rename is "flushed at the next save **or on exit**" — the exit path runs through `run()`'s finally block → `self._checkpointer.shutdown()` → `finalize_pending` (`checkpoint.py:340`). If a last-save background failure is never picked up by a subsequent save, nothing today asserts `shutdown()` re-raises.

**Suggest fix:** companion test with `max_num_steps=2, save_period=2` + `_FailingFinalizeTrainer`, drive `_train_pump` then `_checkpointer.shutdown()`, assert `RuntimeError` at shutdown.

tests/unit/single_controller/test_sc_checkpointing.py
```diff
@@ -0,0 +1,1095 @@
...
+        # identity lands in training_info.json for the restore-side check.
+        assert buffer.state_dict_calls == [4]
+        assert _training_info(ckpt_dir, 2)["sampler_name"] == "windowed"
+
+    def test_run_restores_replay_buffer_and_permits(self, tmp_path):
+        ckpt_dir = tmp_path / "resume_ckpt"
+        ckpt_dir.mkdir()
+        envelope = {"groups": ["g0", "g1", "g2"]}
+        torch.save(envelope, ckpt_dir / "replay_buffer.pt")
+        mc = _actor_master_config(tmp_path, max_num_steps=0)
```
yuki-97, line 1112: All four `_run_actor_run` sites (:1002, :1032, :1058, :1084) pass `max_num_steps=0`, so `_train_pump` skips its loop body — same trivially-zero shape as [the previous comment](https://github.com/NVIDIA-NeMo/RL/pull/3429#discussion_r3691256355). The counter-restore path got its composed resume + pump test (`test_resumed_pump_continues_to_max_steps`, :480); the replay-buffer restore has no equivalent, so a regression that leaves restored permits un-released under a live pump would still pass this suite.

**Suggest fix:** add a companion that restores a buffer with `load_return=k>0`, runs `_train_pump` for a few steps, asserts the pump does not stall and permits are released.

nemo_rl/algorithms/single_controller.py
```diff
@@ -574,6 +665,105 @@ async def _train_pump(self) -> None:
...
+        # SC has no validation loop yet; drop the default sentinel instead of
+        # persisting a bogus val_reward.
+        if hasattr(save_state, "val_reward"):
+            delattr(save_state, "val_reward")
+
+        full_metric_name = self._master_config.checkpointing["metric_name"]
+        if full_metric_name is not None:
+            assert full_metric_name.startswith(
+                "train:"
+            ) or full_metric_name.startswith("val:"), (
```
yuki-97: Line 699 accepts `val:` prefixes and line 706 sets `metrics_source = None` for the val branch, so `if not metrics_source:` (:707) fires the "no val metrics were collected" warning at every save. SC has no validation loop yet (`# SC has no validation loop yet` at :690) — top-k retention silently becomes a no-op instead of failing at setup.

Per `skills/error-handling` "fail loud on misconfig", the validate step is the right home for this.

**Suggest fix:** in `validate_single_controller_config` add `assert metric_name is None or metric_name.startswith("train:"), "SC has no validation loop yet; use 'train:<name>' or leave metric_name=null."`

examples/configs/grpo_math_1B_megatron_single_controller.yaml
yuki-97: To actually verify correctness (not just save/restore mechanics), can we add a diamond check like `tests/functional/sft_resume_diamond.sh`? SC with `windowed` is non-deterministic, but `async_rl.sampler.max_staleness_versions=0` makes it sync — `train/lr`, `train/global_valid_seqs`, `train/global_valid_toks`, `train/num_unmasked_tokens` should `np.isclose` between the 1+3 resumed run and a 4-step baseline. Would catch optimizer/dataloader/scheduler restore bugs the existence checks miss.

nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -836,6 +836,194 @@ async def remove(self, idxs: list[int], remove_in_dp: bool) -> int:
...
+                "put_samples",
+                sample_ids=list(meta.sample_ids),
+                partition_id=self._partition_id,
+                fields=group["fields_data"],
+                tags=[dict(t) for t in meta.tags],
+            )
+            self.meta_list.append(meta)
+            self.start_weight_list.append(group["start_weight"])
+            self.end_weight_list.append(group["end_weight"])
+            self.target_step_list.append(group["target_step"])
```
yuki-97, line 1074: for the in-order sampler, we always assuming each target_step will have exactly N prompts groups.

assuming N=16, we resuming 3 for this step, then we start a new batch from dataloader to generate 16. this will have finished prompts groups > N.

wdyt to check at rollout only submit when ready + inflight < N, if >= N just drop the remaining prompt of current `prompt_batch`, this should be a relatively easy way to fix.

also these dropped data may could be fixed when supporting "Generation-level backpressure", so I think it's fine to just drop for now. (like what the legacy code do)

tests/unit/single_controller/test_sc_checkpointing.py
```diff
@@ -0,0 +1,1095 @@
...
+                "max_groups": 4,
+                "expected_partition_id": _PARTITION_ID,
+                "expected_group_size": 2,
+            }
+        ]
+        # Each restored group holds one _buffer_capacity permit.
+        assert actor._buffer_capacity._value == 4 - 3
+        assert result["train_steps"] == 0
+
+    def test_run_restore_at_full_capacity_does_not_hang(self, tmp_path):
```
yuki-97: `test_run_restore_at_full_capacity_does_not_hang` claims to guard "K==capacity acquisitions don't block", but that's guaranteed by `asyncio.Semaphore(N)` semantics — K acquires with N permits available always succeed instantly. What it actually asserts (`_buffer_capacity._value == 0`, one `load_calls`) is the same permit-accounting shape as `test_run_restores_replay_buffer_and_permits` (:994) with K=3 instead of K=capacity.

Suggest: parametrize L994 with `[K=3, K=max_buffered_rollouts]` and drop this test — or delete outright.

nemo_rl/algorithms/single_controller.py
haitian-nvidia: Done in `325769930` — field declared on `GRPOSaveState`, both `# type: ignore` sites and the `setup.py` re-attach removed. Side effect: every algorithm's `training_info.json` now carries `sampler_name: null`, which tripped `test_get_grpo_save_state_handles_legacy_checkpoint_and_filters_metrics` (updated in the same commit).

nemo_rl/algorithms/single_controller.py
haitian-nvidia: Done in `325769930` — moved into `validate_single_controller_config`. Two deviations: `raise ValueError` instead of `assert` (matches the other checks there), and gated on `checkpointing.enabled` since `metric_name` only drives top-k.

It caught a live misconfig: all SC configs inherit `metric_name: "val:accuracy"` + `keep_top_k: 3` from `grpo_math_1B.yaml`, so top-k has silently been a no-op. Same commit sets `metric_name: null` in the SC exemplar and in the one SC recipe that leaves checkpointing enabled.

nemo_rl/algorithms/async_utils/replay_buffer.py
haitian-nvidia: Done in `8b8bfa847` — over-capacity raises when any group carries `target_step`, naming both ways out; unstamped samplers keep freshest-first. Guard sits before any DataPlane write, so preflight ordering holds.

nemo_rl/algorithms/async_utils/replay_buffer.py
haitian-nvidia: Done in `8b8bfa847`, but counting once after `admit()` rather than per prompt: `reserve()` runs inside `generate_and_push` (`rollout_manager.py:742`), so a per-prompt check reads a stale count and would dispatch all N anyway. Counting at admit is exact — each batch owns a unique `target_step`, so `buffered` is precisely the restored count. New `TQReplayBuffer.count_for_target_step()` covers your `ready + inflight`. Dropping the tail, as agreed.

tests/unit/single_controller/test_sc_checkpointing.py
haitian-nvidia: Done in `c039949ed` — `test_failed_background_finalization_raises_at_shutdown`, built as suggested. Also asserts the rename never happened (`tmp_step_2` present, no `step_2`) before the raise.

tests/unit/single_controller/test_sc_checkpointing.py
haitian-nvidia: Done in `c039949ed` — new `_run_restore_then_train_pump` helper (restore, then a live `_train_pump` bounded by `wait_for`). Restores `K=4 == max_buffered_rollouts`, runs two steps, asserts the semaphore returns to full.

tests/unit/single_controller/test_sc_checkpointing.py
haitian-nvidia: Agreed — deleted in `c039949ed`. The full-capacity shape survives in the new live-pump test above, which uses `K == max_buffered_rollouts` and also proves the permits come back.

examples/configs/grpo_math_1B_megatron_single_controller.yaml
haitian-nvidia: Async GRPO checkpointing doesn't guarantee exact-once prompt consumption across a resume. The checkpoint snapshots the dataloader position and the *ready* contents of the replay buffer while rollout generation keeps running; in-flight prompts are deliberately not persisted, and after a resume the shortfall is filled from later dataloader batches. SingleController follows the same semantics.

So even with a zero-staleness sampler, a resumed run isn't guaranteed to consume the same prompts as an uninterrupted baseline. Since the two runs train on different data, data-dependent metrics like `train/global_valid_toks` and `train/num_unmasked_tokens` can't be compared directly, so a diamond check isn't possible here.

nemo_rl/algorithms/single_controller.py
```diff
@@ -525,6 +605,28 @@ async def _train_pump(self) -> None:
                     )
                     await self._sync_weights(calibration_data=calibration_data)
 
+                # Checkpointing (mirrors async_grpo_train's save block).
+                self._consumed_samples += grpo_cfg.num_prompts_per_step
+                self._total_valid_tokens += step_metrics.get("global_valid_toks", 0)
+                self._timeout.mark_iteration()
+
+                is_last_step = self._train_steps >= grpo_cfg.max_num_steps
```
terrykong: [`single_controller.py:613`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L613)

**1 action item.**

**TL;DR — on resume, `is_last_step` can never become true, so the run stops short of `max_num_steps` and writes no final checkpoint; with `in_order` it exits silently with status 0.**

PR-introduced.

**Background.** `_clamp_max_num_steps` sets `max_num_steps = min(max_num_steps, max_num_epochs * len(dataloader))` ([`setup.py:250`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller_utils/setup.py#L250)). It runs *after* `load_dataloader_state` restores the position ([`setup.py:370-374`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller_utils/setup.py#L370-L374)), but `len(dataloader)` is the full per-epoch batch count regardless of where the loader was restored to. So a resumed run gets a fresh-run step budget against a loader holding only the remaining batches.

`grpo_train` doesn't have this problem because it anchors last-step on **dataloader position** as well as step count ([`grpo.py:3222-3226`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/grpo.py#L3222-L3226)):

```python
is_last_step = total_steps + 1 >= max_num_steps
if not master_config.data["use_multiple_dataloader"]:
    is_last_step = is_last_step or (
        (current_epoch + 1 == max_num_epochs)
        and (current_step + 1 == len(wrapped_dataloader))
    )
```

SC keeps only the first half. That missing disjunct is the whole defect.

**How it breaks** (SC exemplar, `checkpointing.enabled=true`, `save_period=2`; resume with `replay_buffer.pt` deleted):
1. Run 1 saves at train step 2 while the rollout pump has consumed ~6 batches — `_GatedSampler.admit` runs ahead by `gate_window` ([`staleness_sampler.py:316`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/staleness_sampler.py#L316)).
2. Resume: `restored = 0`, so the buffer starts empty and no partial groups exist.
3. `max_num_steps` is re-clamped back to the full `L`, but only `L - 6` batches remain.
4. `min_groups_for_streaming_train: ${grpo.num_prompts_per_step}` means `select` returns `(None, 0)` until a whole batch is ready ([`staleness_sampler.py:212-214`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/staleness_sampler.py#L212-L214)) — so `groups_dispatched == 0` at every step boundary.
5. Loader empties → `_rollout_exhausted` set → `train_meta is None`, `buffered_groups == 0`, `groups_dispatched == 0` → **silent `return`** at [`:488-494`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L488-L494).

**Symptom:** exit code 0, `run()` returns `train_steps` short of `max_num_steps`, `is_last_step` never fired so **no final checkpoint was written** — the run looks like a normal completion. With `windowed` you get the other branch instead: `RuntimeError: rollout exhausted before a complete training step was assembled` ([`:495`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L495)), at the end of a long run.

Worth noting the trigger is the PR's own advice: `load_state_dict`'s over-capacity error tells users to "delete `replay_buffer.pt` from the checkpoint to resume with an empty buffer" ([`replay_buffer.py:1004`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/replay_buffer.py#L1004)), and that empty-buffer start is exactly what makes the shortfall a whole number of batches and the exit silent.

**Action:** in `_train_pump`, mirror `grpo_train`'s disjunct so a resumed run still saves when the epoch runs out before `max_num_steps`:

```suggestion
                is_last_step = self._train_steps >= grpo_cfg.max_num_steps or (
                    self._rollout_exhausted.is_set() and len(self._buffer) == 0
                )
```

(one way to do it — anchoring on rollout exhaustion is the SC-native equivalent of grpo's dataloader-position check; clamping against *remaining* batches in `_clamp_max_num_steps` would also work.)

nemo_rl/algorithms/single_controller.py
```diff
@@ -525,6 +605,28 @@ async def _train_pump(self) -> None:
...
+                # _train_steps was already incremented above, so it equals
+                # the legacy loop's 1-indexed `step + 1`.
+                should_save_by_step = (
```
terrykong, line 639: [`single_controller.py:616`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L616)

**1 action item.**

`should_save_by_step` omits the `ft_save_period` term. PR-introduced.

Every other algorithm reads it and ORs it into the save trigger — [`grpo.py:3367`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/grpo.py#L3367), [`grpo.py:4828`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/grpo.py#L4828), [`grpo_sync.py:1132`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/grpo_sync.py#L1132), [`dpo.py:659`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/dpo.py#L659), [`ppo.py:1509`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/ppo.py#L1509), [`rm.py:595`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/rm.py#L595), [`distillation.py:994`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/distillation.py#L994) — and `CheckpointManager.remove_old_checkpoints` still honors `ft_keep_latest_k`.

**Failure:** a user sets `checkpointing.ft_save_period: 5` / `ft_keep_latest_k: 2` on an SC recipe. Zero fault-tolerance checkpoints are ever written, no warning is emitted, and crash recovery silently falls back to the `save_period` cadence. The config is accepted and ignored.

**Action:** read it once in `__init__` and add the term, matching `grpo_sync.py:1129-1134`:

```suggestion
                ft_save_period = self._master_config.checkpointing.get("ft_save_period")
                should_save_by_step = (
                    is_last_step
                    or self._train_steps
                    % self._master_config.checkpointing["save_period"]
                    == 0
                    or (
                        ft_save_period is not None
                        and self._train_steps % ft_save_period == 0
                    )
                )
```

If FT checkpointing is intentionally out of scope for SC, reject it in `validate_single_controller_config` instead — either way it shouldn't be silently dropped.

examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
```diff
@@ -10,6 +10,7 @@ grpo:
 
 checkpointing:
   checkpoint_dir: results/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller
+  metric_name: null
```
terrykong, line 16: [`grpo-qwen3-30ba3b-…-single-controller.yaml:13`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml#L13)

**1 action item.**

Line 7 of this file still reads `# SC does not support validation and checkpointing yet.` — now false for the checkpointing half, and this PR edited line 13 six lines below it.

Two more carry the same stale line and aren't in this diff:
- `examples/configs/recipes/llm/grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync.yaml:9`
- `examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml:9`

**Action:** narrow rather than delete in all three — the validation half is still true and is the premise for the new `metric_name` `train:`-only guard, plus `grpo.val_period: 0` sits right below it. For this file (line 7):

```
# SC does not support validation yet.
```

tests/functional/grpo_checkpoint_single_controller.sh
```diff
@@ -0,0 +1,141 @@
...
+    logger.wandb_enabled=false
+    logger.monitor_gpus=false
+    checkpointing.enabled=true
+    checkpointing.checkpoint_dir=$CKPT_DIR
+    checkpointing.save_period=2
+    checkpointing.metric_name=null
+    data_plane.enabled=true
+    data_plane.impl=transfer_queue
+    data_plane.backend=simple
+    async_rl.sampler.name=windowed
```
terrykong, line 57: **1 action item.** Follow-up to the [diamond thread](https://github.com/NVIDIA-NeMo/RL/pull/3429#discussion_r3736453392) — your objection there was right; here's why, verified:

- **The suggested knob is a no-op for dispatch.** `WindowedSampler.admit` has no version gate at any `max_staleness_versions` ([`staleness_sampler.py:252-254`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/staleness_sampler.py#L252-L254)); staleness only narrows eviction. A diamond built on it would flake.
- **An exact diamond isn't definable anyway.** Prompts in flight at save are consumed-but-dropped ([`:698`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L698), [`replay_buffer.py:843`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/replay_buffer.py#L843)), so a resumed run trains on different prompts — even at `max_lookahead_versions=0`.

**Action — say so in the header:**

```suggestion
# Verifies checkpoint save/restore MECHANICS only. Metric equivalence across
# resume is not well-defined here: prompts in flight at save time are dropped
# while the dataloader cursor has already advanced past them (not exact-once).
    async_rl.sampler.name=windowed
```

Optional: `train/lr` alone IS prompt-independent (pure function of scheduler step count) and a 2-run compare at indices 2-3 would catch an LR-schedule reset on resume — but it needs TB enabled + a baseline run, so skip if too heavy for L1. (Assumes `lr` is emitted on the SC path: `"lr"` ∈ `_MB_METRIC_MEAN` ([`utils.py:34-36`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller_utils/utils.py#L34-L36)), not runtime-verified.)

nemo_rl/algorithms/single_controller.py
```diff
@@ -574,6 +674,87 @@ async def _train_pump(self) -> None:
...
+        save_state.total_steps = self._train_steps
+        save_state.current_epoch = self._current_epoch
+        save_state.consumed_samples = self._consumed_samples
+        save_state.total_valid_tokens = self._total_valid_tokens
+        # The restore skips the replay buffer when the resuming run uses a
+        # different sampler (its stamps may never be selectable there).
+        save_state.sampler_name = self._async_cfg.sampler.name
+        # Snapshot before any await so it can't interleave with
+        # _rollout_pump iterating this same dataloader.
+        dataloader_state = self._dataloader.state_dict()
```
terrykong, line 760: [`single_controller.py:698`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L698)

> 📊 [Visual explainer of the cursor-advance mechanism](https://terrykong.github.io/gh-pages-poc/terryk/pr3429-cursor-advance.html)

**1 action item — please fix in this PR.** (Full partial-rollout persistence stays tracked in #3594; this asks only for the prompts.)

**TL;DR — prompts in flight (or yielded-but-undispatched) at save time are silently lost forever on resume; persisting their `DatumSpec`s and re-dispatching them through the normal path closes the gap without touching generation state.**

**The mechanism, in four steps:**
1. The cursor advances the moment a batch is *yielded* ([`:375`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L375)); dispatch inside the batch is prompt-by-prompt behind permits ([`:398-400`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L398-L400)).
2. This snapshot records that already-advanced cursor.
3. `state_dict` drops unready reservations ([`replay_buffer.py:843`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/replay_buffer.py#L843)) — and it couldn't save their prompts even if it wanted to: `reserve()` stores `meta=None` until commit ([`:685`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/replay_buffer.py#L685)), so the buffer never knows the prompt.
4. On resume the loader continues past them → up to `max_inflight_prompts` + one full batch of prompts per save are never trained on. The bias is non-uniform: long-generation prompts are the ones most likely to be in flight at a save boundary, so they are disproportionately dropped. `in_order`'s top-up ([`:384`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L384)) preserves the per-step *count* but fills from fresh prompts; `windowed` preserves nothing.

**Action — persist the pending prompts and re-enqueue on resume.** One way to do it (the prompts are all known driver-side; no partial-generation state needed):

1. **Track:** keep `self._pending_prompts: dict[str, DatumSpec]` in the actor — add the whole batch's prompts when the rollout pump yields it (this also captures the undispatched tail), remove each on successful commit. Mutations happen between awaits on the loop, so `_save_checkpoint`'s synchronous window snapshots it atomically for free.
2. **Save:** include `list(self._pending_prompts.values())` in the checkpoint — e.g. a `pending_prompts` key in the `replay_buffer.pt` envelope (one atomic artifact, and `load_state_dict`'s preflight can validate it).
3. **Resume:** before drawing from the dataloader, the rollout pump dispatches the restored prompts as its first synthetic batch(es), through the normal `admit`/permit path. For `in_order` this dovetails with the existing top-up — the first admit's shortfall (`batch.size − count_for_target_step`) is exactly the number of lost prompts for that target step, so the restored prompts fill precisely the holes. No dataloader rewind needed anywhere.

One gating nuance: the buffer restore is rightly skipped on a `sampler_name` mismatch ([`:270`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L270)), but the pending *prompts* should be re-enqueued regardless — raw prompts are sampler-agnostic; it's the ready groups' weight/target stamps that aren't. Handle the two envelope keys separately in `_maybe_restore_replay_buffer`.

Backward compat is free: the new key is optional, and a checkpoint without it restores exactly as today.

**Testing this deterministically — gate the fake rollouts on per-prompt events.** The suite already has both halves of the idiom: a never-finishing rollout via `await asyncio.Event().wait()` ([`test_rollout_pump.py:218`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/tests/unit/single_controller/test_rollout_pump.py#L218)) and a `threading.Event`-gated finalize fake ([`test_sc_checkpointing.py:146`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/tests/unit/single_controller/test_sc_checkpointing.py#L146)). Generalize to a gated rollout manager and the partial state becomes exactly controllable — no sleeps, no races:

```python
class _GatedRolloutManager(_FakeRolloutManager):
    def __init__(self, *a, **kw):
        super().__init__(*a, **kw)
        self.gates: dict[str, asyncio.Event] = {}   # prompt id -> gate

    async def generate_and_push(self, prompt, *a, **kw):
        gate = self.gates.setdefault(_pid(prompt), asyncio.Event())
        await gate.wait()                # frozen "in flight" until the test releases it
        return await super().generate_and_push(prompt, *a, **kw)
```

Then one test expresses the whole failure geometry: release batch-3's gates (→ ready slots), leave batch-4's unset (→ provably in flight), let batch 5 park on a full buffer; fire the save; assert the envelope's `pending_prompts` is exactly batch 4 + batch 5 and the ready groups are exactly batch 3. A second test restores into a fresh actor with a recording rollout manager and asserts the first dispatches are exactly the restored prompts (and, for `in_order`, that the top-up shortfall arithmetic holds). That pair pins both save-side capture and resume-side re-dispatch, and would have caught this gap in the first place.

nemo_rl/algorithms/single_controller.py
```diff
@@ -574,6 +674,87 @@ async def _train_pump(self) -> None:
...
+            buffer_state,
+            os.path.join(checkpoint_path, "replay_buffer.pt"),
+        )
+        # Rename happens in the background once the async weight writes
+        # finish; flushed at the next save or on exit.
+        self._checkpointer.begin_finalization(
+            checkpoint_path,
+            wait_fn=self._trainer.finalize_async_save,
+        )
+        _write_latest_checkpoint_status(
```
terrykong: [`single_controller.py:754`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L754)

**1 action item** (nit).

`_write_latest_checkpoint_status` reads, JSON-parses and rewrites a file inline on the event loop — the same loop servicing the rollout pump. The other five blocking calls in this method are all wrapped in `asyncio.to_thread`: [`finalize_pending` :715](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L715), [`init_tmp_checkpoint` :719](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L719), [`save_checkpoint` :727](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L727), and both `torch.save`s ([:736](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L736), [:744](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L744)).

Pre-empting the obvious rebuttal — this isn't "`grpo.py:3478` does the same, house idiom". `grpo_train` is **synchronous**; there's no event loop to block. `_save_checkpoint` is `async def` and its own docstring says *"rollouts keep running throughout"*, so this is an internal inconsistency in this PR's new async code, not a deviation from grpo.

**Action:**

```suggestion
        await asyncio.to_thread(
            _write_latest_checkpoint_status,
            self._checkpointer,
            last_checkpoint_step=self._train_steps,
        )
```

nemo_rl/algorithms/async_utils/staleness_sampler.py
```diff
@@ -104,6 +104,10 @@ def required_buffer_capacity(self, groups_per_step: int) -> Optional[int]:
         """Buffer-capacity the policy needs, or ``None`` if unconstrained."""
         ...
 
+    def set_dispatch_index(self, resume_from_step: int) -> None:
```
terrykong, line 116: [`staleness_sampler.py:107`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/staleness_sampler.py#L107)

**1 suggestion** ([DESIGN], low severity — fine to defer to a tracking issue).

**TL;DR — nothing in the sampler API says "my state survives a resume"; today that's true for the built-ins only by inventory, and a stateful custom sampler would silently reset.**

For the three built-ins this PR's restore is genuinely complete: the only mutable sampler state anywhere in this file is `_dispatch_index` ([`:318`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/async_utils/staleness_sampler.py#L318) is the sole mutation), and `set_dispatch_index` restores it. But that completeness is a fact about the current implementations, not a contract. A `CustomSamplerConfig` sampler with real internal state (adaptive curriculum weights, per-prompt bookkeeping, an RNG) passes the resume gate — both runs report `sampler_name == "custom"` — gets the replay buffer restored under it, and has its private state silently reset to `__init__` values. No warning fires; the restore log reads as success.

**Suggestion — one way to do it:** generalize `set_dispatch_index` into a state pair on the protocol, with `BaseSampler` providing the built-ins' complete implementation:

```python
class PromptGroupSampler(Protocol):
    def state_dict(self) -> dict[str, Any]: ...
    def load_state_dict(self, state: dict[str, Any], *, resume_from_step: int) -> None: ...

class BaseSampler(abc.ABC):
    def state_dict(self) -> dict[str, Any]:
        return {"dispatch_index": self._dispatch_index}

    def load_state_dict(self, state, *, resume_from_step: int) -> None:
        # Built-ins re-derive the cursor from the trainer step; customs
        # may override to restore whatever they track.
        self.set_dispatch_index(resume_from_step)
```

Persist the dict next to `sampler_name` (it already rides in `training_info.json`), and call `load_state_dict` at the current `set_dispatch_index` site ([`single_controller.py:157`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/single_controller.py#L157)). A custom sampler that doesn't override the pair gets today's behavior; one that does gets a real hook instead of a silent reset.

To be explicit: this is **not** re-proposing the `supports_buffer_checkpoint` boolean that was (rightly) removed earlier in this review — that asked "can I be checkpointed?", which a sampler can't meaningfully answer. This gives samplers the *mechanism* to round-trip their state, of which the dispatch cursor becomes the base case.

Since no in-repo custom sampler exists yet, deferring this behind a tracking issue is perfectly reasonable — but the protocol is being widened in this PR anyway, so the seam is cheapest to add now.

terrykong: @haitian-nvidia this came up in the review and in general i would imagine staleness sampler could be stateful, so this could be a good thing to track, but just leaving this for discussion

nemo_rl/algorithms/single_controller.py
```diff
@@ -127,15 +130,31 @@ def __init__(
         self._logger.log_hyperparams(master_config.model_dump())
         self._timer = Timer()
 
+        # Also built here, not on the driver: TimeoutChecker must capture
+        # wall-clock start times inside the actor, not at driver setup time.
+        # actor_args only carries the driver-side restore products
```
terrykong, line 141: **No action needed** — FYI, recorded so nobody re-flags this placement later.

The actor-side construction here is correct and matches legacy: `grpo_train` and `async_grpo_train` also build their `TimeoutChecker` post-setup, at the top of the train function ([`grpo.py:2615`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/grpo.py#L2615), [`:3943`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/algorithms/grpo.py#L3943)) — so the comment above defending the actor-side anchor stands.

The shared, pre-existing hazard (all paths, not this PR): the budget anchors at construction ([`timer.py:433`](https://github.com/NVIDIA-NeMo/RL/blob/c039949edb91e972cb0b40393a11f285458d5373/nemo_rl/utils/timer.py#L433)), i.e. *after* Ray/cluster/model setup — so `checkpoint_must_save_by` silently excludes setup time, and a user budgeting from allocation start (4h allocation, `3h30` budget) can have the timeout save fire after the allocation dies. Tracked in #3596 (anchor at `SLURM_JOB_START_TIME` when present, with a `time.time()` fallback). No change requested in this PR.

terrykong (commented):
## Agent-team review — 6 agents

Thanks for this — closing the `NotImplementedError` on the SC path is a real unblock, and the test work is genuinely strong. `test_tq_replay_buffer.py`'s additions in particular (a real `torch.save`/`torch.load` cycle over a non-contiguous TensorDict view, preflight tests that also assert nothing reached the DataPlane) are the best part of the PR. Reusing `CheckpointManager.begin_finalization` rather than re-implementing a second flush protocol inside the actor was the right call.

**Please rebase — and by hand.** The PR is `CONFLICTING` with `main`. The conflicts come from a single commit, `0e687e6d0` (#3499, NeMo-Gym spinup), which touches none of the checkpointing logic — but it lands on the exact `_build_trainer(weights_path=…, optimizer_path=…)` call sites this PR adds, so please resolve manually rather than with `-X ours`/`-X theirs`.

**Two findings worth acting on before merge**, both about save *triggers* rather than the save path itself: the missing dataloader-position disjunct in `is_last_step` (a resumed run can end short and write no final checkpoint), and `ft_save_period` being silently ignored. The other four are a doc fix, a test-scope note, a docstring, and a one-line `to_thread` nit.

**On the open threads:** we verified the other seven unresolved threads against HEAD and they're genuinely addressed — the `8b8bfa847` truncation fix and the `count_for_target_step` top-up were each re-derived independently by two reviewers and are correct (the top-up is exactly permit-balanced, which is the non-obvious part). Those can be resolved. The one comment below on the functional test follows up on the diamond thread, where your position turns out to be right and the suggested determinism knob turns out to be a no-op.

**Verification caveat, stated plainly:** this review is static. The repo lockfile is Linux-only and the review host is macOS, so neither `pre-commit` nor any test was executed — there is no lint or test signal behind these findings, only source reading. Anything needing a runtime is flagged as unverified inline.

_Generated by Claude Code_

nemo_rl/algorithms/single_controller.py
macandro96: @terrykong @haitian-nvidia I was looking at this comment and i believe this overlaps with #3507 which was built on top of TQ x Gym integration to support partial rollout checkpointing.
#3507  currently uses a slightly different invariant instead of a separate _pending_prompts map: it acquires prompt-batch capacity before advancing the dataloader and couples dataloader advancement with rollout-ledger reservation under the checkpoint barrier. The snapshot then includes the dataloader cursor, rollout ledger, replay metadata, and native TQ state. On restore it reconstructs those groups, reuses sealed siblings, and redispatches only missing generations.
So I agree the gap must be fixed, but we should avoid having both pending_prompts and the rollout ledger represent the same in-flight prompts. Would you prefer that we extract a minimal prompt-ledger/cursor fix from #3508 into #3429 and have #3508 extend it with TQ partial-state reconciliation, or keep the implementation in #3508/#3594?
Let me know what you think?

nemo_rl/algorithms/single_controller.py
haitian-nvidia: Done in `6a3408d50` — took the rollout-exhaustion disjunct as suggested. Once exhaustion + a drained buffer are known, every completed step is potentially the last, so each one saves. New unit test (`test_rollout_exhaustion_saves_final_checkpoint`) drives a pump whose sampler exhausts after step 2 of `max_num_steps=4` and asserts steps 1–2 both checkpoint and the pump still exits cleanly at 2.

nemo_rl/algorithms/single_controller.py
haitian-nvidia: Done in `6a3408d50` — ORed into `should_save_by_step`, matching `grpo_sync.py:1129-1134`; read via `.get` since the key is `NotRequired`. New test: `save_period=100`, `ft_save_period=2`, `max_num_steps=3` → exactly `{step_2, step_3}`.

examples/configs/recipes/llm/grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml
haitian-nvidia: Done in `6a3408d50` — narrowed to `# SC does not support validation yet.` in all three recipes.

tests/functional/grpo_checkpoint_single_controller.sh
haitian-nvidia: Done in `6a3408d50` — header now carries the mechanics-only statement with your wording. Skipped the optional `train/lr` two-run compare per the skip-if-too-heavy call: it needs TB enabled plus a baseline run, which is heavy for L1.

nemo_rl/algorithms/single_controller.py
haitian-nvidia: Done in `6a3408d50` — wrapped in `asyncio.to_thread`, matching the other five blocking calls in `_save_checkpoint`.

nemo_rl/algorithms/single_controller.py
haitian-nvidia: @terrykong @macandro96 I think checkpointing in-flight rollout and dataloader cursor advance mechanism is out of the scope of this PR, we can make them as follow-up PRs. For the current one we just keep it consistent with the legacy.

nemo_rl/algorithms/async_utils/staleness_sampler.py
haitian-nvidia: Let's keep the current design for now and move to a `state_dict`/`load_state_dict` pair when a sampler with more state than the dispatch cursor actually shows up

nemo_rl/algorithms/single_controller.py
terrykong: resolving for now to merge the PR, let's keep discussin on the issue

terrykong (dismissed):

haitian-nvidia:
/ok to test https://github.com/NVIDIA-NeMo/RL/commit/ab447cb6f9e8f5ea9b6a5cddc7b435b75e183334

terrykong (dismissed):

yuki-97 (approved):

PR #3066, opened 2026-07-05
fix(async-grpo): add startup barrier to prevent checkpoint-resume deadlock

snowmanwwg:
## Problem

When resuming async GRPO from a checkpoint whose replay-buffer snapshot already contains a complete batch for the current training step, training silently deadlocks — no Python traceback, no error, just a stall that eventually hits the job timeout.

**Root cause:**

1. The startup wait loop checks `has_complete_batch(step)` and breaks immediately when `True`.
2. After consuming those restored trajectories and triggering a refit, the collector's `_calculate_target_weights` takes the non-initial branch and returns `[weight_version+1, ...]`, skipping the just-refitted step entirely.

Timeline on a step-30 resume with 8 trajectories for target 30 in the checkpoint:

| Event | State |
|---|---|
| `load_state_dict(step=30)` | 8 trajectories for target 30 in buffer |
| Startup wait: `has_complete_batch(30)` → True, break | Target 31 never checked |
| `sample(weight_version=30)` consumes 8 trajectories | |
| Refit → `weight_version = 31` | |
| `_calculate_target_weights(31)`: non-initial → reserves target 32 | Target 31 nobody generates |
| `sample(weight_version=31)` → 0 found → stall | |

This race is invisible during normal uninterrupted training because target N+1 is generated while step N trains, so the lookahead is always present before a refit. On a checkpoint resume, target N is already complete, so startup returns before the lookahead exists. First surfaced by PR #2651 (replay-buffer checkpointing).

## Fix

Before breaking out of the startup wait, also require `has_complete_batch(step+1)` to be True (or `step+1 >= max_num_steps`). The collector's initial branch of `_calculate_target_weights` already covers `[start_step, start_step + max_age]`, so it fills the lookahead while the barrier holds. Once both targets are ready, consuming the current step and refitting is safe.

The barrier is a no-op on fresh starts (target 0 is not in the buffer, so the loop never reaches the `if current_step_ready` branch until both targets have been filled naturally).

## Changes

- `nemo_rl/algorithms/grpo.py` — lookahead check inside the `if current_step_ready` branch, guarded by `max_trajectory_age_steps > 0` and `step + 1 < max_num_steps`
- `tests/unit/algorithms/test_async_utils.py` — `test_resume_deadlock_precondition_detectable` constructs the exact buffer state (current complete, lookahead absent) and asserts the readiness values the barrier branches on

## Related

[...]

PR #3435, opened 2026-07-31
fix(async): preserve rollout lookahead on resume

yfw:
# What does this PR do ?

Follow-up to #3066, which fixed the async GRPO checkpoint-resume deadlock by waiting for the next rollout batch before resuming training. This change preserves that protection while allowing training and rollout generation to overlap.

## Background

After restoring a checkpoint with a complete batch for step N, #3066’s startup barrier waits until the full batch for N+1 is available. This prevents the collector from advancing past N+1, but it serializes an entire rollout round before training can resume.

The barrier only needs to know that N+1 has a producer. Once the collector has reserved N+1, training step N can safely begin while those rollouts are generated.

## Changes

- Include actively reserved target weights in trajectory collector status.
- Consider the startup pipeline ready when:
    - the current training batch is complete; and
    - the lookahead target is either complete or actively claimed by the collector.

- Continue blocking when the lookahead is neither complete nor claimed.
- Preserve the existing last-step and no-lookahead behavior.

## Result

On checkpoint resume:

- #3066 behavior: wait for the complete N+1 rollout batch, then train N.
- New behavior: wait only until N+1 is claimed, then train N while N+1 is generated.

This keeps the deadlock protection introduced by #3066 without adding a full, unoverlapped rollout round to every resume.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


nemo_rl/algorithms/grpo.py
```diff
@@ -4595,11 +4622,24 @@ def async_grpo_train(
...
+                "dataloader exhausted"
+                if collector_status["data_exhausted"]
+                else "collector errored"
+            )
+            recovery_advice = (
+                "Increase data.train.max_num_epochs or use a larger dataset."
+                if collector_status["data_exhausted"]
+                else "Inspect the preceding trajectory collector error."
+            )
             raise RuntimeError(
```

**TL;DR — the `continue` this PR deletes used to skip the exhausted/errored guard entirely, so a resume with a complete batch for step N and a dead dataloader spun at 1 Hz forever; that hang-to-raise fix is the most valuable change in the PR and nothing covers it.**

PR-introduced. To be clear this is a note on a *good* change, not a defect.

Pre-PR the lookahead branch ended in `wait_iterations += 1; time.sleep(1.0); continue`, which sits **above** this guard — so the driver could never reach it. `check_health()` at the top of the loop doesn't cover the gap either: it only re-raises `_fatal_error_message`, which [`_run_rollout_batch_worker`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1367) sets from *worker* failures, not [`_mark_collection_failed`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/async_utils/trajectory_collector.py#L435). Post-PR the same state falls through to [`grpo.py:4637`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/grpo.py#L4637) and raises.

The happy path through this loop is well covered — every `async_grpo_train` test traverses it. The **blocking** path and this new `RuntimeError` are not, and the fixture is why: [`StubReplayBuffer.has_complete_batch`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/tests/unit/algorithms/test_grpo.py#L1031) is `lambda _target_step, num_prompts_per_step, *_args: self._size >= num_prompts_per_step` — it discards the target step, so no existing test can express "step N complete, N+1 absent", which is exactly the state this branch needs.

### AI-1 — add a test for the abort path

I verified this appended to a copy of the real `test_grpo.py`: **`1 passed in 24.75s`**, `ruff format --check` and `ruff check` both clean. Two mutation checks, so you know it has teeth: `target=1` → `target=2` fails with `Regex pattern did not match`, and `call_count == 2` → `== 1` fails with `assert 2 == 1` — which is what proves the barrier *blocked for a full iteration* before raising rather than falling straight through.

```python
def test_async_grpo_startup_aborts_when_lookahead_never_claimed(
    mock_grpo_components, tmp_path
):
    """Startup raises instead of spinning when the lookahead can never be claimed.

    Regression for the pre-fix ``continue``, which skipped the
    exhausted/errored check whenever the lookahead was missing and left the
    driver waiting on a target no producer would ever claim.
    """
    master_config = mock_grpo_components["master_config"]
    master_config.grpo.max_num_steps = 2
    master_config.grpo.val_period = 0
    master_config.grpo.val_at_start = False
    master_config.grpo.val_at_end = False
    master_config.grpo.use_dynamic_sampling = False
    master_config.policy["generation"]["colocated"]["enabled"] = False
    checkpointer = mock_grpo_components["checkpointer"]
    checkpointer.get_latest_checkpoint_path.return_value = None
    checkpointer.checkpoint_dir = tmp_path

    mock_batch = next(iter(mock_grpo_components["train_dataloader"]))
    mock_rollout_metrics = {"mean_gen_tokens_per_sample": 2.0}

    # Restored buffer: step 0 is complete, step 1 is absent.
    complete_batch = MagicMock()
    complete_batch.remote = MagicMock(
        side_effect=lambda target_step, *_args, **_kwargs: target_step == 0
    )

    # First poll: collector alive but has not claimed step 1 -> keep waiting.
    # Second poll: collector stopped with the dataset exhausted -> must raise.
    statuses = [
        {
            "running": True,
            "data_exhausted": False,
            "errored": False,
            "error": None,
            "inflight_workers": 1,
            "generating_targets": [],
        },
        {
            "running": False,
            "data_exhausted": True,
            "errored": False,
            "error": None,
            "inflight_workers": 0,
            "generating_targets": [],
        },
    ]
    status = MagicMock()
    status.remote = MagicMock(side_effect=statuses)

    with (
        patch.object(
            StubReplayBuffer,
            "has_complete_batch",
            property(lambda self: complete_batch),
        ),
        patch.object(
            StubAsyncTrajectoryCollector, "get_status", property(lambda self: status)
        ),
        patch("nemo_rl.algorithms.grpo.time.sleep", return_value=None),
        mock_async_grpo_infrastructure(mock_batch, mock_rollout_metrics),
        pytest.raises(
            RuntimeError,
            match=r"dataloader exhausted.*lookahead claim at target=1",
        ),
    ):
        async_grpo_train(
            mock_grpo_components["policy"],
            _mock_policy_generation(),
            mock_grpo_components["train_dataloader"],
            mock_grpo_components["val_dataloader"],
            mock_grpo_components["tokenizer"],
            mock_grpo_components["loss_fn"],
            mock_grpo_components["task_to_env"],
            mock_grpo_components["val_task_to_env"],
            mock_grpo_components["logger"],
            checkpointer,
            _initial_grpo_save_state(),
            master_config,
        )

    # Two polls means the barrier blocked for a full iteration before raising.
    assert status.remote.call_count == 2
```

The `match` also pins the new `stop_reason` split, so it covers the exhausted-vs-errored wording you added at [`grpo.py:4627-4636`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/grpo.py#L4627).

One honest limitation, worth a comment in the test if you keep it: on `origin/main` these same inputs `continue` forever, so this **hangs** rather than going red. It locks in the new abort behavior; it would not have caught the original bug.

nemo_rl/algorithms/grpo.py
```diff
@@ -4703,6 +4743,11 @@ def async_grpo_train(
                         collector_status = ray.get(
                             trajectory_collector.get_status.remote()
                         )
+                        awaited_target = step
+                        print(
+                            f"   Awaiting target {awaited_target}; claimed by collector: "
+                            f"{awaited_target in collector_status.get('generating_targets', ())}"
+                        )
```
yfw, line 5045: **1 action item.** Adjacent footgun rather than something this PR broke — but you are editing this exact block, and you already fixed its twin.

The starvation guard six lines below ([`grpo.py:4759-4764`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/grpo.py#L4759)) fires on `data_exhausted` **or** `errored`, but the message only tells the exhausted story:

> `Trajectory collector stopped: dataloader exhausted at training_step={step}. The dataset ran out of data before training could complete. ... Increase data.train.max_num_epochs or use a larger dataset.`

`errored=True, data_exhausted=False` is directly reachable: [`_collection_loop`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/async_utils/trajectory_collector.py#L533)'s `except Exception` calls `_mark_collection_failed(e)` and then its `finally` sets `self.running = False`, leaving `data_exhausted` at `False` because `dataloader_exhausted` was never set. So an operator whose collection thread crashed with, say, a tokenizer error is told to buy a bigger dataset, and the real exception is only visible buried inside the `Collector status: {...}` dump.

**Action:** in `async_grpo_train`'s main-loop starvation branch, reuse the `stop_reason` / `recovery_advice` pattern you just introduced at [`grpo.py:4627-4636`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/grpo.py#L4627) instead of hardcoding the exhausted text — the two messages are otherwise the same message. No `suggestion` block because line 4759 falls six lines past the end of the last diff hunk (which ends at 4753), so GitHub would reject a comment anchored there.

nemo_rl/algorithms/grpo.py
```diff
@@ -4117,6 +4117,41 @@ def aggregate_rollout_metrics(
...
+    max_trajectory_age_steps: int,
+    max_num_steps: int,
+) -> bool:
+    """Return whether async training can overlap safely with lookahead generation.
+
+    A complete current step is sufficient only after the collector has either
+    completed or claimed the next target. The claim preserves the same
+    training/generation overlap used by steady-state async training without
+    allowing an unrelated reservation to open the startup barrier.
+    """
```
yfw, line 4362: **1 action item.**

Extracting the helper dropped this comment from the loop with no replacement anywhere (it is a pure deletion in the diff):

> `A restored buffer may already contain `step`, allowing startup to consume it before the collector generates `step + 1`. After refit, the collector advances to targets starting at `step + 2`, leaving `step + 1` permanently missing.`

The new docstring states the **rule** ("complete or claimed") but not the **mechanism** that makes the barrier necessary — that the post-refit window advances to `step + 2`, so an unclaimed `step + 1` is orphaned with no producer. Since the helper is now the only home for that rationale, a reader can no longer derive *why a claim is sufficient* from anything in the file. That mechanism is also exactly what your new `_release_target` assertion in [`test_startup_barrier_clears_when_collector_reserves_lookahead`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/tests/unit/algorithms/test_async_utils.py#L2531) encodes.

**Action:** restore it into the docstring. I verified this exact text applied to a copy of `grpo.py`: `ruff format --check` → `1 file already formatted`, `ruff check` → `All checks passed!`

```suggestion
    """Return whether async training can overlap safely with lookahead generation.

    A restored buffer may already contain ``step``, letting startup consume it
    before the collector generates ``step + 1``. After the refit that follows
    ``step``, the collector's target window advances to ``step + 2``, so an
    unclaimed ``step + 1`` would then be permanently missing. Requiring the
    next target to be complete *or* actively claimed keeps that gap closed
    while preserving the training/generation overlap used by steady-state async
    training, and without letting an unrelated reservation open the barrier.
    """
```

yfw (commented):
Reviewed the resume barrier change. The mechanism holds: a claim really does carry its target across the refit boundary, because the batch worker writes to the `target_weight_version` it reserved at dispatch regardless of the weight-version advance, and `prepare_for_refit` either drains in-flight workers or lets them complete in flight. No blocking issues.

Worth calling out explicitly — moving the exhausted/errored check onto the fall-through path converts a silent 1 Hz hang into an actionable error. `data_exhausted` is assigned once at [`trajectory_collector.py:538`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/async_utils/trajectory_collector.py#L538) and never reset, so the state is terminal rather than transient, which makes raising the right call; and a resume that only needs to finish its last step still exits cleanly, because `step + 1 == max_num_steps` makes `need_lookahead` False. That is the most valuable change here, which is why the inline comment asks for a test to pin it.

Async PPO carries the same #3066 barrier at [`ppo.py:2343-2367`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/ppo.py#L2346) and still requires a *complete* lookahead, and it restores its replay buffer too ([`ppo.py:2242`](https://github.com/NVIDIA-NeMo/RL/blob/3fc2964d9727726760c2ab3c9707904fdde728bc/nemo_rl/algorithms/ppo.py#L2242)), so async PPO resumes still pay the serialized rollout round this PR removes for GRPO. It is not a drop-in — PPO passes `initial_buffer_max_age` derived from its warmup schedule and omits the `max_trajectory_age_steps > 0` term. Tracked separately in #3828; nothing to do here.

`pre-commit run --all-files` passes on this head.

_Generated by Claude Code_

nemo_rl/algorithms/grpo.py
yfw: Addressed by accdaf47e09730246ab8563f6a831b06203cd821

nemo_rl/algorithms/grpo.py
yfw: Addressed by accdaf47e09730246ab8563f6a831b06203cd821

nemo_rl/algorithms/grpo.py
yfw: Addressed by accdaf47e09730246ab8563f6a831b06203cd821

nemo_rl/algorithms/grpo.py
```diff
@@ -4192,6 +4192,44 @@ def aggregate_rollout_metrics(
     return aggregated
 
 
+def _startup_pipeline_ready(
```
yuki-97: The barrier now opens once N+1 is merely [claimed](https://github.com/NVIDIA-NeMo/RL/blob/c6291e23fdd99eddab11667cc1b7f3d536ae56f7/nemo_rl/algorithms/grpo.py#L4195), not complete — but a claim isn't a guarantee. If that worker fails, [`_release_target`](https://github.com/NVIDIA-NeMo/RL/blob/c6291e23fdd99eddab11667cc1b7f3d536ae56f7/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1114) drops it with no retry, and [`_calculate_target_weights`](https://github.com/NVIDIA-NeMo/RL/blob/c6291e23fdd99eddab11667cc1b7f3d536ae56f7/nemo_rl/algorithms/async_utils/trajectory_collector.py#L258) permanently excludes that version once [`set_weight_version(N+1)` runs, before `resume_after_refit` re-enables dispatch](https://github.com/NVIDIA-NeMo/RL/blob/c6291e23fdd99eddab11667cc1b7f3d536ae56f7/nemo_rl/algorithms/grpo.py#L5223-L5229) — true for both engine configs. The main loop then spins forever, since a lone sub-threshold failure never trips `errored`, reopening on resume the exact deadlock class PR #3066 closed by requiring N+1 fully complete.

**Suggest fix:** on release of a claimed-but-incomplete target still below the window, re-queue it for retry, or fail loud via check_health regardless of the global max_generation_failures count.

nemo_rl/algorithms/grpo.py
```diff
@@ -4786,11 +4834,22 @@ def async_grpo_train(
                             and not collector_status["running"]
                             and collector_status["inflight_workers"] == 0
                         ):
+                            stop_reason = (
```
yuki-97: The collector-stopped guard is duplicated almost verbatim in two places: the startup wait loop (~L4695-4721) and here in the main training loop (~L4829-4853). Both compute the same `stop_reason`/`recovery_advice` from `collector_status` and raise the same-shaped `RuntimeError`, differing only in wording ("start" vs "continue", `target=` vs `training_step=`).

**Suggest fix:** extract the shared guard into a helper so the two sites can't drift:

```python
def _raise_if_collector_stopped(
    collector_status: dict[str, Any],
    *,
    awaited_target: str,
    awaited_work: str,
    action: str,  # "start" or "continue"
) -> None:
    if not (
        (collector_status["data_exhausted"] or collector_status.get("errored", False))
        and not collector_status["running"]
        and collector_status["inflight_workers"] == 0
    ):
        return
    stop_reason = (
        "dataloader exhausted"
        if collector_status["data_exhausted"]
        else "collector errored"
    )
    recovery_advice = (
        "Increase data.train.max_num_epochs or use a larger dataset."
        if collector_status["data_exhausted"]
        else "Inspect the preceding trajectory collector error."
    )
    raise RuntimeError(
        f"Trajectory collector stopped ({stop_reason}) while waiting for "
        f"{awaited_work} at {awaited_target}. "
        f"Training cannot {action} without the required trajectories. "
        f"Collector status: {collector_status}. "
        f"{recovery_advice}"
    )
```

Call sites become one-liners: `_raise_if_collector_stopped(collector_status, awaited_target=..., awaited_work=..., action="start"|"continue")`.

yfw: Addressed in 92f3ed924c2fcd9986006afe68302fc70ba86858

yuki-97 (approved):

PR #2964, opened 2026-06-26
fix: resolve HybridEP sequence packing padding

seonjinn:
# What does this PR do?

- Builds the physical padding mask for packed THD inputs and keeps its CP/SP-local layout aligned with the model input.
- For supported HybridEP sequence-packed workloads, aligns unequal TP x EP input lengths once in NeMo-RL before model collectives start.
- Preserves NeMo-owned pre-padding after Megatron-Bridge validation and disables MCore's repeated per-layer padding collective for this path.
- Fails closed unless the workload uses the HybridEP flex dispatcher, NeMo-owned sequence packing, PP=1, and MTP disabled.

The opt-in is `megatron_cfg.moe_hybridep_prepad_packed_inputs=true`. Other layouts continue to use MCore's padding behavior.

# Why?

Nemotron3 Super with EP=32 exposed a collective-ordering hang. Some ranks entered MCore's per-layer scalar token-count MAX all-reduce while other ranks entered the distributed optimizer's expert-parameter all-gather on the same communicator sequence number. The earlier one-time pre-padding strategy present at `cc9610abad51cbd44e9c7931af5961468b9a806b` avoided that collision. This PR restores that ownership with explicit supported-layout guards while retaining the current THD padding-mask handling.

# Dependencies

- Companion x86 HybridEP package enablement: #3436.
- MCore must exclude physical padding rows from routing and broadcast the padding mask across the expert dimension. The successful Super runs use the routing-mask corrections corresponding to NVIDIA/Megatron-LM#5542 and NVIDIA/Megatron-LM#6114; the latter is still open. This PR does not vendor a custom Megatron-LM fork.

# Validation

- GCP-NRT B200 job `507684`: 12/12 focused tests passed at PR head `eecfeeb0` against Megatron-Bridge `fcbabe78` and stock MCore `14499ef0`.
- Canonical strict-candidate jobs Qwen3-30B-A3B `506437`, Qwen3-235B `506438`, and Nemotron3 Super `506434`: 20/20 steps, exit `0:0`, no training fatal.
- NeMo-RL-main-pinned Bridge jobs Qwen3-30B-A3B `507111`, Qwen3-235B `507112`, and Nemotron3 Super `506895`: 20/20 steps, exit `0:0`, no classified runtime fatal. These runs used the MCore routing-mask candidate described above.
[...]

PR #2224, opened 2026-04-07
feat: add Gemma4 support

sharonyu-115:
# What does this PR do ?

Adds Gemma 4 support to NeMo-RL with DAPO and GRPO recipes across dense and MoE variants, plus a VLM recipe. 

# Issue 
List issues that this PR closes: 
https://github.com/NVIDIA-NeMo/RL/issues/2212

**Summary of code changes:** 

### `nemo_rl/models/policy/utils.py`
Registers `gemma4` → `AutoModelForImageTextToText` in both the HF and
NeMo-AutoModel factory maps (2-line addition). Makes Gemma4 load as a
VLM-capable architecture.

### `nemo_rl/models/generation/vllm/vllm_worker.py`
Adds `Gemma4ForConditionalGeneration` to the two arch lists that gate special
vLLM handling (alongside Gemma3 and Qwen3.5).

### `nemo_rl/models/automodel/train.py`
Two Gemma4 forward-path fixes:
- New `_needs_kv_cache_for_shared_layers()` helper + threading a `use_cache`
  arg through `model_forward` / `forward_with_post_processing_fn`. Gemma4 E2B
  uses KV-sharing (`num_kv_shared_layers > 0`) and needs `use_cache=True` so
  `DynamicCache` feeds K/V from anchor layers to shared layers — otherwise
  shared layers fall back to untrained projections and produce garbage.
  Temporary; removable after transformers ≥5.5.2 (HF #45312).
- Injects `mm_token_type_ids` (zeros) for `model_type == "gemma4"` even on
  text-only inputs, mirroring existing Gemma3 `token_type_ids` handling.

### `nemo_rl/models/automodel/setup.py`
- New `_disable_automodel_checkpoint_dtype_restore()`: monkeypatches
  Automodel's `_restore_loaded_model_dtype` to a no-op so NeMo-RL's
  `torch_dtype=float32` master-weight load isn't silently downcast back to
  bf16 (which broke AdamW — the nano-v2-12b reward-stuck-at-0.18 bug).
  Added unit test to detect when the patch is on longer needed. 
- **Removes the unconditional visual-encoder freeze** for text-only training —
  now handled declaratively via recipe `freeze_config`. (This is the
  behavioral change flagged in PR review, now documented in
  `docs/model-quirks.md`.)

### `nemo_rl/models/automodel/checkpoint.py`
New `_patch_qwen_vl_vision_key_mapping()` import-time monkeypatch re-adding the
[...]

PR #3436, opened 2026-07-31
perf: enable HybridEP on H100/B200 x86 systems

seonjinn:
## Summary

- Make the same DeepEP HybridEP dependency available on Linux x86_64 and aarch64.
- Pin every dependency group and the lockfile to `17cfb817bccec3a9c247013360cc550c2bac441e`.

## Multi-node requirement

`HYBRID_EP_MULTINODE` defaults to `0` and is evaluated when DeepEP is built. Multi-node execution therefore requires a fresh wheel or actor environment built with `HYBRID_EP_MULTINODE=1`; recipe variables cannot change an existing single-node binary.

For H100/B200 x86, `8` means ranks per NVLink domain, not maximum EP size. The EP group must be divisible by 8. The H100 runs used the DOCA/NCCL path without a DeepEP source or DMA-BUF patch.

## Validation

| Model | H100 topology | EP | Result | W&B |
|---|---:|---:|---:|---|
| Qwen3-30B-A3B | 4 × 8 GPUs | 8 | 20/20 | [run](https://wandb.ai/nvidia/sna-hybridep-h100-validation/runs/46beq3cz) |
| Qwen3-235B-A22B | 16 × 8 GPUs | 16 | 20/20 | [run](https://wandb.ai/nvidia/sna-hybridep-h100-validation/runs/esept1ll) |
| Nemotron 3 Super 120B-A12B | 32 × 8 GPUs | 32 | 20/20 | [run](https://wandb.ai/nvidia/sna-hybridep-h100-validation/runs/195tut82) |

An exact, unpatched `17cfb817` wheel also completed 20/20 steps for Qwen3-30B-A3B on 4 × 8 B200 GPUs; the DMA-BUF workaround was not required.

These runs included #2964 and #3438. Nemotron 3 Super also used the Megatron-LM fixes from NVIDIA/Megatron-LM#5542 and NVIDIA/Megatron-LM#6114, so the results establish compatibility of the pin rather than isolation of this dependency-only diff.

## Boundary and testing

- `uv lock --check` passed with 560 packages resolved.
- The 59/59 B200-hosted test from #3438 covers resolved recipes, not DeepEP runtime.
- The larger B200 Nemotron 3 Super/EP32 path remains unresolved; this PR does not vendor or rely on the experimental DMA-BUF workaround.


pyproject.toml
```diff
@@ -132,17 +132,15 @@ automodel = [
   "causal-conv1d",
   "nv-grouped-gemm",
   "transformer-engine[pytorch,core_cu13] @ git+https://github.com/NVIDIA/TransformerEngine.git@v2.14.1",
-  "deep_ep @ git+https://github.com/deepseek-ai/DeepEP.git@29d31c095796f3c8ece47ee9cdcc167051bbeed9 ; platform_machine == 'x86_64'",
-  "deep_ep @ git+https://github.com/deepseek-ai/DeepEP.git@a48493600c4886c1b297aaa78db0e1ebc2d8dd6c ; platform_machine == 'aarch64'",
+  "deep_ep @ git+https://github.com/deepseek-ai/DeepEP.git@17cfb817bccec3a9c247013360cc550c2bac441e",
```
yuki-97, line 148: Would help to link an H100 run of one existing DeepEP recipe on pin `17cfb817` in the PR body — e.g. `tests/test_suites/llm/grpo-qwen3.5-35ba3b-2n8g-automodel-ep16.sh`.

None of the L0 / fast-L1 tiers turn on DeepEP, so their `success` here does not cover the actual path this bump touches.

pyproject.toml
```diff
@@ -366,8 +363,7 @@ override-dependencies = [
   "llguidance>=1.3.0,<1.4.0",
   # Override setuptools range in other dependencies to address CVE GHSA-58pv-8j8x-9vj2
   "setuptools>=80.10.2",
-  "deep_ep @ git+https://github.com/deepseek-ai/DeepEP.git@29d31c095796f3c8ece47ee9cdcc167051bbeed9 ; platform_machine == 'x86_64'",
-  "deep_ep @ git+https://github.com/deepseek-ai/DeepEP.git@a48493600c4886c1b297aaa78db0e1ebc2d8dd6c ; platform_machine == 'aarch64'",
+  "deep_ep @ git+https://github.com/deepseek-ai/DeepEP.git@17cfb817bccec3a9c247013360cc550c2bac441e",
```
yuki-97, line 414: On GB200, two things would help close the "Validation boundary" gap:

1. Could you link the B200 validation from #3438 (job `507030`, 59/59 on head `d68c6fa8`) in the PR body? Since that PR depends on this bump, that job is effectively the runtime evidence for `17cfb817` on B200.

2. Is the local DMA-BUF fix still required on top of `17cfb817` (i.e. was job `507030` also patched)? If so, could you upload it so the shipped pin actually matches what B200 needs at runtime — otherwise people may meet issues.

yuki-97 (commented):
besides the two inline comments about test and patch, will need to rebase main since there's a conflict.

pyproject.toml
yuki-97: I remember previously `grpo-qwen3.5-35ba3b-2n8g-automodel-ep16` had some issue before bump deep_ep to `29d31c095796f3c8ece47ee9cdcc167051bbeed9` in https://github.com/NVIDIA-NeMo/RL/pull/2224, so it's good to have a test on this.

seonjinn:
Contributor self-review refresh for `919a6d102dc223cf3e19341e8a04eab30870aa8d` after the latest main and lockfile refresh.

The DeepEP pin is consistent across x86 and aarch64 resolution paths, and lock/config checks pass. Two runtime evidence items remain before merge readiness: link or rerun the H100 recipe validation, and confirm whether the B200 DMA-BUF workaround is still required on the exact pinned stack.

pyproject.toml
seonjinn: Thanks — I re-checked the provenance.

For H100, multi-node runtime validation used the exact, unpatched DeepEP pin `17cfb817` and a wheel built with `HYBRID_EP_MULTINODE=1` before the actor environment was created. The NeMo-RL validation stack also included #2964 and the matching recipe enablement from #3438, so this is runtime compatibility evidence for the pin rather than an isolation test of this dependency-only diff.

| Model | H100 80 GB topology | HybridEP result | E2E TPS/GPU vs matched AllToAll | W&B |
|---|---:|---:|---:|---|
| Qwen3-30B-A3B | 4 nodes × 8 GPUs | 20/20 steps | +0.60% | [run](https://wandb.ai/nvidia/sna-hybridep-h100-validation/runs/46beq3cz) |
| Qwen3-235B-A22B | 16 nodes × 8 GPUs | 20/20 steps | +4.34% | [run](https://wandb.ai/nvidia/sna-hybridep-h100-validation/runs/esept1ll) |
| Nemotron 3 Super 120B-A12B | 32 nodes × 8 GPUs | 20/20 steps | +8.74% | [run](https://wandb.ai/nvidia/sna-hybridep-h100-validation/runs/195tut82) |

W&B history was re-checked and contains train steps 1–20 for all three H100 runs. The multi-node flag is a build-time requirement at this pin; #3436 selects the source revision but does not itself rebuild an existing single-node DeepEP binary.

Job `507030` was a B200-hosted configuration test, not a DeepEP runtime test. It ran `tests/unit/tools/test_hybridep_default_8g_recipes.py` and passed 59/59; it did not build, import, or execute DeepEP. It was therefore not patched, and I do not think it should be presented as runtime evidence for `17cfb817`. It is valid resolved-recipe coverage for #3438.

Separately, a 4-node × 8-GPU B200 Qwen3-30B-A3B run used an exact, unpatched wheel built from `17cfb817` and completed 20/20 steps with exit code 0. That establishes that the DMA-BUF fallback is not required for this validated B200 path.

The local DMA-BUF fallback was an experimental workaround for `ibv_reg_mr errno=14` observed in a larger Super/EP32 configuration. The patched Super attempts did not establish a completed result because they later failed in other runtime paths, so there is not enough evidence to vendor/upload that workaround or make this dependency bump rely on it. The validation boundary should distinguish the 59/59 recipe test, the exact-pin Qwen runtime pass, and the still-unresolved large-EP corner case.

pyproject.toml
yuki-97: validated `grpo-qwen3.5-35ba3b-2n8g-automodel-ep16` pass.

yuki-97 (approved):

PR #3437, opened 2026-07-31
refactor(vllm): defer HTTP server imports in async worker

tianyi-zhang-02:
## What does this PR do?

Defers the direct `fastapi` and `uvicorn` imports in `vllm_worker_async.py` until the HTTP server is actually created.

This is a small import-layering cleanup, not a claim that `nemo_rl.algorithms.grpo` can run without those packages: other import paths still load them today. #3552 carries this change together with the broader fixes that produce the measurable GRPO import improvement.

## Testing

- Added a runtime regression that reloads the async worker and records its direct imports, failing if `fastapi` or `uvicorn` moves back to module scope
- `ruff`, formatting, and `git diff --check` pass

Superseded by #3552, so I am closing this focused version.

cc @RayenTian @yuki-97

### DIFF
--- a/nemo_rl/models/generation/vllm/vllm_worker_async.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker_async.py
@@ -20,12 +20,14 @@ import threading
 import time
 import uuid
 import warnings
-from typing import Any, AsyncGenerator, Optional, cast
+from typing import TYPE_CHECKING, Any, AsyncGenerator, Optional, cast
 
 import ray
 import torch
-import uvicorn
-from fastapi import FastAPI
+
+if TYPE_CHECKING:
+    import uvicorn
+    from fastapi import FastAPI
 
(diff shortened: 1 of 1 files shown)


tianyi-zhang-02:
Closing in favour of #3552, which carries this commit and does the rest of the job.

This change on its own does not move anything — module count stays at exactly 7150 with it applied. `vllm_sparse_refit.py:29-31` imports `fastapi`/`uvicorn` at module scope too, and more to the point the vLLM workers are not where `fastapi` comes from: `algorithms/grpo.py:127` → `utils/memory_tracker.py:19` → `from ray.scripts.scripts import memory_summary`, which pulls in Ray's dashboard stack.

#3552 defers the real sources: 7150 → 5256 modules, 9.04s → 3.77s to `import nemo_rl.algorithms.grpo`.

PR #3438, opened 2026-07-31
perf: enable HybridEP in H100/B200 performance recipes

seonjinn:
## Dependencies

Requires #2964. The HybridEP dependency enablement from #3436 is already merged. Nemotron 3 Super also requires the merged Megatron-LM fixes NVIDIA/Megatron-LM#5542 and NVIDIA/Megatron-LM#6114. This PR does not vendor a custom Megatron-LM fork.

These recipes configure runtime topology only. Multi-node DeepEP must be built with `HYBRID_EP_MULTINODE=1` before the actor environment is created. On H100/B200 x86, `8` denotes ranks per NVLink domain, not maximum EP size; H100 EP16 and EP32 completed 20/20 steps. See #3436 for build and runtime evidence.

## Summary

- Enable HybridEP in the canonical H100/B200 MoE performance recipes.
- Set explicit x86 and GB200 topology values in the affected recipes.
- Enable one-time prepadding only for supported PP=1, MTP-disabled packed workloads.
- Bound Qwen3-30B-A3B H100 logprob memory with `logprob_chunk_size: 1024` and deferred FP32 logits.

## Matched H100 performance

Each pair used the same model, node/GPU topology, batch sizes, sequence length, parallelism, and logprob settings from the performance recipe. The only functional A/B difference was AllToAll versus HybridEP (including the required HybridEP topology and padding settings). Values are means over steps 3-19 of completed 20-step runs on H100 80 GB.

| Model / dispatcher | Result | E2E step (s) | E2E tokens/s/GPU | Policy step (s) | Policy tokens/s/GPU | Logprob step (s) | Logprob tokens/s/GPU |
|---|---:|---:|---:|---:|---:|---:|---:|
| [Qwen3-30B-A3B / AllToAll](https://wandb.ai/nvidia/sna-hybridep-h100/runs/qt0a1hqn) | 20/20 | 158.82 | 1,313.62 | 38.20 | 5,443.38 | 11.89 | 17,494.02 |
| [Qwen3-30B-A3B / HybridEP](https://wandb.ai/nvidia/sna-hybridep-h100/runs/yajqw1jm) | 20/20 | **153.78** | **1,356.24** | **34.67** | **5,995.00** | **10.44** | **19,911.95** |
| Qwen3-30B-A3B improvement | — | **3.18% faster** | **+3.24%** | **9.24% faster** | **+10.13%** | **12.20% faster** | **+13.82%** |
| [Nemotron3 Super / AllToAll](https://wandb.ai/nvidia/sna-hybridep-h100/runs/neebcx3k) | 20/20 | 80.42 | 39.22 | 6.71 | 487.02 | 4.27 | 729.60 |
| [Nemotron3 Super / HybridEP](https://wandb.ai/nvidia/sna-hybridep-h100/runs/xd53tixy) | 20/20 | **77.92** | **40.83** | **4.68** | **682.40** | **3.83** | **815.65** |
| Nemotron3 Super improvement | — | **3.11% faster** | **+4.12%** | **30.17% faster** | **+40.12%** | **10.45% faster** | **+11.79%** |

Qwen3-30B-A3B used `logprob_chunk_size: 1024` and deferred FP32 logits in both arms. Nemotron3 Super used its unchanged performance-recipe memory settings in both arms. This avoids attributing memory-setting differences to HybridEP.

Qwen3-235B HybridEP also completed all 20 training steps with the canonical performance recipe. Its matched baseline rerun is still in progress, so no unmatched speedup is reported here.

## Testing

- H100 Qwen3-30B-A3B and Nemotron3 Super: matched 20-step runs linked above.
- H100 Qwen3-235B HybridEP: 20/20 training steps completed.
- 59 resolved-recipe unit cases passed.
- B200 resolved-recipe validation passed.
- `git diff --check`


seonjinn:
Contributor self-review status: this recipe PR remains blocked on a clean HybridEP runtime dependency. The current `moe_hybridep_prepad_packed_inputs` setting has no typed runtime consumer on latest main, while the intended dependency path still relies on the over-broad #2964 branch. Recommendation: first land/replace the runtime support and #3436 pin, then restack and validate this recipe; do not retrigger CI on the current head.

nemo_rl/models/megatron/setup.py
```diff
@@ -950,6 +950,109 @@ def _apply_parallelism_config(model_cfg: Any, config: PolicyConfig) -> None:
...
+            raise ValueError(
+                "NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN="
+                f"{ranks_text!r} must be a positive integer"
+            ) from exc
+        os.environ["NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN"] = str(ranks_per_domain)
+        ranks_source = "megatron_cfg"
+    elif "NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN" in os.environ:
+        ranks_source = "environment"
+    else:
+        ranks_per_domain = min(ep_size, 64)
```
yuki-97: **[BUG]** The auto-fallback can compute a value that the check below immediately rejects.

`min(ep_size, 64)` yields 64 whenever `expert_model_parallel_size > 64`, and [`:994`](https://github.com/NVIDIA-NeMo/RL/blob/67b127aeb61eac5b6b4dc1113ba35f1277c4e815/nemo_rl/models/megatron/setup.py#L994) then raises `must divide expert_model_parallel_size` for any EP that is not a multiple of 64 — 80, 96 and 144 all abort, naming an env var the user never set. On main this path warned and ran. DeepEP asserts the same invariant itself, so the check is right and the fallback is what is wrong.

**Suggest fix:**

1. use `math.gcd(ep_size, 64)` here — same value for every EP that works today, and divisible by construction.
2. add `import math`.

examples/configs/recipes/llm/performance/grpo-nemotron3-super-120BA12B-32n8g.yaml
```diff
@@ -20,6 +20,16 @@ policy:
   dtensor_cfg:
     enabled: false
   megatron_cfg:
+    # HybridEP settings
+    moe_token_dispatcher_type: flex
+    moe_flex_dispatcher_backend: hybridep
+    moe_hybridep_num_sms: 32
+    moe_hybridep_prepad_packed_inputs: true
```
yuki-97, line 27: nit: enabling HybridEP here leaves the 4-GPU sibling documenting a limitation that no longer exists.

[`grpo-nemotron3-super-120BA12B-32n4g.yaml:36-37`](https://github.com/NVIDIA-NeMo/RL/blob/9dfe0afe10629b4e4d5533541857b4b81d9c987d/examples/configs/recipes/llm/performance/grpo-nemotron3-super-120BA12B-32n4g.yaml#L36-L37) still says HybridEP is disabled `due to incompatibility with sequence packing`, but #2964 fixed exactly that: [`hybridep.py:56-62`](https://github.com/NVIDIA-NeMo/RL/blob/9dfe0afe10629b4e4d5533541857b4b81d9c987d/nemo_rl/models/megatron/hybridep.py#L56-L62) gates prepadding on PP=1 and MTP disabled, and both 4-GPU recipes satisfy it. Same block in [`-32n4g-async-1off.yaml:47-48`](https://github.com/NVIDIA-NeMo/RL/blob/9dfe0afe10629b4e4d5533541857b4b81d9c987d/examples/configs/recipes/llm/performance/grpo-nemotron3-super-120BA12B-32n4g-async-1off.yaml#L47-L48).

**Suggest fix:**

1. drop the `due to incompatibility with sequence packing` clause in both 4-GPU recipes, or replace it with the real reason.
2. keep them on alltoall — [`test_hybridep_default_8g_recipes.py:72-77`](https://github.com/NVIDIA-NeMo/RL/blob/9dfe0afe10629b4e4d5533541857b4b81d9c987d/tests/unit/tools/test_hybridep_default_8g_recipes.py#L72-L77) asserts they must not carry HybridEP keys.

tests/unit/tools/test_hybridep_default_8g_recipes.py
```diff
@@ -0,0 +1,220 @@
...
+        _megatron_config(_resolve_recipe("grpo-deepseek-v3-32n8g.yaml"))
+    )
+    child_environment = _environment(
+        _megatron_config(_resolve_recipe("grpo-deepseek-v3-32n4g.yaml"))
+    )
+
+    unrelated_parent_environment_keys = (
+        set(parent_environment) - GB200_HYBRIDEP_ENVIRONMENT_KEYS
+    )
+    assert unrelated_parent_environment_keys <= set(child_environment)
```
yuki-97: nit: this assertion cannot fail.

`grpo-deepseek-v3-32n8g.yaml`'s effective `env_vars` is exactly the four `GB200_HYBRIDEP_ENVIRONMENT_KEYS`, so the subtraction is empty and this reduces to the empty set being a subset of anything — it passes even if the child had no `env_vars` at all. No recipe chain here has a parent-only unrelated key, so there is nothing to repoint it at.

**Suggest fix:** drop it — [`:174-190`](https://github.com/NVIDIA-NeMo/RL/blob/9dfe0afe10629b4e4d5533541857b4b81d9c987d/tests/unit/tools/test_hybridep_default_8g_recipes.py#L174-L190) already asserts `32n4g`'s four env values.

nemo_rl/models/megatron/setup.py
```diff
@@ -950,6 +950,109 @@ def _apply_parallelism_config(model_cfg: Any, config: PolicyConfig) -> None:
...
+        os.environ["NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN"] = str(ranks_per_domain)
+        ranks_source = "fallback"
+        warnings.warn(
+            "HybridEP: NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN not "
```
yuki-97: nit: both auto-default warnings lost the only actionable thing they carried.

Pre-PR they named the override key and the derivation — `Auto-setting to min(expert_model_parallel_size=8, 64) = 8. Set 'hybridep_num_ranks_per_nvlink_domain' in megatron_cfg to override.` Both now print a bare number, and the rank-0 diagnostic reports `source=fallback` without the key name either.

**Suggest fix:** put the override key name back in both messages, here and at [`:1009`](https://github.com/NVIDIA-NeMo/RL/blob/67b127aeb61eac5b6b4dc1113ba35f1277c4e815/nemo_rl/models/megatron/setup.py#L1009).

tests/unit/tools/test_hybridep_default_8g_recipes.py
```diff
@@ -0,0 +1,220 @@
...
+    recipe_name: str,
+) -> None:
+    megatron_cfg = _megatron_config(_resolve_recipe(recipe_name))
+
+    assert megatron_cfg["moe_token_dispatcher_type"] == "flex"
+    assert megatron_cfg["moe_flex_dispatcher_backend"] == "hybridep"
+    assert megatron_cfg["moe_hybridep_num_sms"] == 32
+    assert X86_HYBRIDEP_ENVIRONMENT.items() <= _environment(megatron_cfg).items()
+
+
```
yuki-97: nit: nothing catches a new 8-GPU recipe that is added later and belongs in one of these lists.

The four tuples name 29 files while `examples/configs/recipes/llm/performance/` holds 43, so a future MoE 8-GPU recipe silently escapes every HybridEP assertion here. The nearest analog already guards this — [`test_mxfp8_rollout_recipes.py:279-281`](https://github.com/NVIDIA-NeMo/RL/blob/67b127aeb61eac5b6b4dc1113ba35f1277c4e815/tests/test_mxfp8_rollout_recipes.py#L279-L281).

The glob is exact today: `*n8g*.yaml` is 19 files and `MOE_8G_RECIPES | DENSE_8G_RECIPES` is the same 19.

```suggestion


def test_every_8gpu_performance_recipe_is_classified() -> None:
    eight_gpu_recipes = {path.name for path in _recipe_dir().glob("*n8g*.yaml")}
    assert eight_gpu_recipes == set(MOE_8G_RECIPES) | set(DENSE_8G_RECIPES)


```

yuki-97 (commented):
**Scope:** the 19 HybridEP performance recipes and the new `tests/unit/tools/test_hybridep_default_8g_recipes.py`.

**Checked:**

- Effective config — resolved every `defaults:` chain and confirmed the ranks/EP divisibility, the `NVLINK_DOMAIN_SIZE`/`USE_MNNVL` pairing, and the PP=1 and MTP-disabled prepad rule.
- Upstream contract — the x86 env block against Megatron-Bridge's H100 perf recipes.
- Test reachability — CI shard collection for the new file, and whether the nightly `.sh` drivers override any of these keys.

Nothing blocking. **Both remaining comments are nits:** a stale sequence-packing note on the 4-GPU nemotron recipes, and an assertion in the new test that cannot fail.

examples/configs/recipes/llm/performance/grpo-nemotron3-super-120BA12B-32n8g.yaml
seonjinn: Fixed in fb0bf1338.

Confirmed the stale reason: both 4-GPU recipes leave `pipeline_model_parallel_size` at its default of 1 and set `mtp_num_layers: 0`, so the `hybridep.py` prepad gate (PP=1, MTP disabled) passes for them. The `due to incompatibility with sequence packing` clause is wrong post-#2964.

Replaced it in both files with the real reason - the 4-GPU topology has not been perf-validated with HybridEP - and kept them on alltoall, so `test_4g_non_hybridep_recipes_do_not_set_hybridep_topology` still holds.

tests/unit/tools/test_hybridep_default_8g_recipes.py
seonjinn: Correct, dropped in fb0bf1338.

Verified the assertion was vacuous: `grpo-deepseek-v3-32n8g.yaml` chains to `grpo_math_1B.yaml`, which sets no `megatron_cfg.env_vars`, so the parent resolves to exactly the four `GB200_HYBRIDEP_ENVIRONMENT_KEYS`. The subtraction is empty and the subset check passes unconditionally. `test_gb200_qwen3_235b_16n4g_preserves_parent_environment` still covers real parent-environment inheritance.

yuki-97 (approved):

Issue #2140, opened 2026-03-23
NCCL timeout when reproducing Qwen3-235B performance test


jp7c5:
**Describe the bug**

Reproducing performance experiment on on-policy Qwen3-235B, I always got NCCL timeout during computing logprob.
I'm using H200 nodes, and on the same cluster I succeeded on reproducing results on Qwen3-30B-A3B and Qwen3-32B.

```bash
▶ Computing logprobs...
[36m(MegatronPolicyWorker[rank=2] pid=2739419, ip=10.1.40.56)[0m /nemoRL_r0.5.0/nemo_rl/distributed/model_utils.py:700: UserWarning: Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x[tuple(seq)] instead of x[seq]. In pytorch 2.9 this will be interpreted as tensor index, x[torch.tensor(seq)], which will result either in an error or a different result (Triggered internally at /pytorch/torch/csrc/autograd/python_variable_indexing.cpp:345.)
[36m(MegatronPolicyWorker[rank=2] pid=2739419, ip=10.1.40.56)[0m   ids_chunks.append(input_ids[slices])
[36m(MegatronPolicyWorker pid=357821, ip=10.1.82.182)[0m WARNING:megatron.core.datasets.megatron_tokenizer:megatron_tokenizer.py:27: You’re using the legacy tokenizer system, which is deprecated and will be removed in a future release. Please migrate to the new tokenizer system (`megatron.core.tokenizers.MegatronTokenizer`).[32m [repeated 127x across cluster][0m
[36m(MegatronPolicyWorker[rank=30] pid=3264028, ip=10.1.101.236)[0m /nemoRL_r0.5.0/nemo_rl/distributed/model_utils.py:700: UserWarning: Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x[tuple(seq)] instead of x[seq]. In pytorch 2.9 this will be interpreted as tensor index, x[torch.tensor(seq)], which will result either in an error or a different result (Triggered internally at /pytorch/torch/csrc/autograd/python_variable_indexing.cpp:345.)[32m [repeated 16x across cluster][0m
[36m(MegatronPolicyWorker[rank=30] pid=3264028, ip=10.1.101.236)[0m   ids_chunks.append(input_ids[slices])[32m [repeated 16x across cluster][0m
[...]

PR #3441, opened 2026-07-31
perf(modelopt): skip redundant weight fake-quant during frozen-weight logprobs

### LINKED ISSUE #3440: Avoid repeated weight fake-quantization during QAT logprob rescoring
## Summary

With ModelOpt QAT enabled, Megatron GRPO repeatedly fake-quantizes unchanged weights during `get_logprobs`.

`get_logprobs` runs without optimizer updates, but every microbatch still executes the ModelOpt quantized forward wrappers and recomputes the quantized weights. On Qwen2.5-32B-Instruct with W4A16, this makes logprob rescoring approximately 1.76x slower than bf16.

## Measurements

Configuration:

- Qwen2.5-32B-Instruct
- GRPO with the Megatron backend
- ModelOpt `nvfp4_a16`
- TP=8, PP=1, DP=4
- 32 H100-80GB GPUs
- Sequence length 16384
- Global batch size 256
- Microbatch size 1

`policy_and_reference_logprobs`, measured over two consecutive steps:

| configuration | step 1 | step 2 |
|---|---:|---:|
| bf16 | 148.43 s | 160.26 s |
| W4A16 QAT | 261.69 s (1.763x) | 281.30 s (1.755x) |

## Where the repeated work occurs

The repeated work comes from the interaction of these two paths:

1. NeMo-RL's [`MegatronPolicyWorkerImpl.get_logprobs`](https://github.com/babyplutokurt/RL/blob/60b1e3085150b8868760d32234881085a76164ed/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1495-L1572) runs the forward-only microbatch loop under `torch.no_grad()`.
2. In the pinned ModelOpt revision, [`_QuantFunctionalMixin.forward`](https://github.com/NVIDIA/Model-Optimizer/blob/c3b913b9cc1d82d5a0af9fa77b4db87829e6f158/modelopt/torch/quantization/plugins/custom.py#L49-L73) installs the quantized linear functional on every forward. That functional calls [`self.weight_quantizer(weight)`](https://github.com/NVIDIA/Model-Optimizer/blob/c3b913b9cc1d82d5a0af9fa77b4db87829e6f158/modelopt/torch/quantization/nn/modules/quant_linear.py#L45-L53) each time.

The weights do not change between these microbatches, so the weight quantization result is recomputed for the same parameter throughout the stage.

## Proposed direction

Avoid recomputing frozen-weight snaps during `get_logprobs`:

- Cache each weight snap for recipes that still require the ModelOpt wrappers for activation quantization.
- For weight-only recipes, materialize each snapped weight once for the stage and bypass eligible per-forward wrappers.

The optimization must remain limited to frozen-weight evaluation. It must not change the training forward or bypass enabled activation or attention quantizers.

## Scope

Megatron ModelOpt QAT logprob rescoring only. The training forward and its straight-through gradient path are out of scope.

babyplutokurt:
## What does this PR do?

  Adds an opt-in `policy.quant_fold_frozen_weight_snap` flag that eliminates redundant weight fake-quantization during the QAT logprob stages.

  In ModelOpt QAT the weight quantizer sits inside the linear, so every forward recomputes `weight_quantizer(weight)`. During the `get_logprobs` re-scoring stage the weights are frozen (`no_grad`, no
  optimizer step between microbatches), so that result is identical across every microbatch and is pure wasted work.

  With the flag set, the stage is wrapped in a context manager that folds each enabled fake-quant weight quantizer once, using ModelOpt's fold formula (`QuantModule.fold_weight`): the fake-quantized
  value is written into the parameter and the weight quantizer is disabled, which is exactly the frozen-weight steady state. Forwards then read an already-snapped weight instead of re-snapping per
  microbatch. Weights and quantizers are restored on exit, including on exception. Default off; when unset the path is byte-for-byte the base implementation.

  ## Why fold per pair instead of calling `mtq.fold_weight`

  The upstream utility selects quantizers on `fake_quant` alone and dereferences `weight.data` unconditionally (`quant_module.py`, line 147). On Megatron models with tied word embeddings the
  `output_layer` is built with `skip_weight_param_allocation` and carries `weight = None` (the embedding weight is borrowed at forward time), while ModelOpt still attaches a `weight_quantizer` to it.
  `mtq.fold_weight` crashes there with `AttributeError: 'NoneType' object has no attribute 'data'`. Reproduced end to end on Qwen3-0.6B QA-GRPO, which ties embeddings; HF never hits this because it
  expresses tying as an aliased tensor.

  The per-pair fold applies the identical upstream formula but:
  - skips pairs whose weight is not a tensor (the tied `output_layer` case), and
  - skips disabled quantizers, whose forward is the identity: folding them is a no-op and cloning their weights for restore wastes memory (the disabled `lm_head`/embedding quantizers hold roughly 40% of
  the quantized-weight bytes on Qwen3-0.6B).

  Discovery uses the same `*_weight_quantizer` attribute-name suffix scan as `fold_weight`, so fused and MoE modules (`w13_weight_quantizer`, `gate_up_proj_weight_quantizer`) are folded and, critically,
  restored.

  ## Correctness and performance verification

  - Unit tests (10, running the real ModelOpt library): forward output is bit-identical folded vs unfolded, weights restore through their original storage, amax survives, restoration holds on exception,
  fused quantizers are restored, tied-embedding modules with `weight = None` are skipped, disabled quantizers are untouched.
  - Mechanism check on GPU (NVFP4 W4A16 and W4A4, bf16): logits and `state_dict` bit-identical with and without the fold. Only the weight quantizer is disabled, so W4A4 keeps its activation quantizers
  running.
  - E2E A/B on Qwen3-0.6B QA-GRPO (2 GPUs, fixed seed): both arms complete 3/3 steps; step-1 metrics identical; 112 quantizers folded and restored per policy logprob pass, 0 on the reference pass (which
  runs unquantized by design).
  - Speedup scales as 1 + Wq/G (weight-quant cost over GEMM cost per forward), so the win is largest for short-sequence/small-microbatch scoring: up to ~1.7x on the logprob stage at ~1k tokens per
  forward on the smoke setup, shrinking toward 1.0x for long-sequence recipes such as the 30k-token DAPO config.

  ## Usage

  ```yaml
  policy:
    quant_cfg: "examples/modelopt/quant_configs/nvfp4_a16.yaml"
    quant_fold_frozen_weight_snap: true
  ```
  Enabled in the QA recipe configs in this PR; costs one temporary copy of each folded weight shard for the duration of the stage. Megatron QAT only.

### DIFF
--- a/nemo_rl/modelopt/models/policy/workers/weight_folding.py
+++ b/nemo_rl/modelopt/models/policy/workers/weight_folding.py
@@ -1,0 +1,118 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Temporarily fold fake-quantized weights for frozen-weight QAT stages.
+
+In ModelOpt QAT the weight quantizer sits inside the linear, so every forward recomputes
+``weight_quantizer(weight)`` — an elementwise pass over the weight shard, uncached.
+During the ``get_logprobs`` re-scoring stage the weights are frozen (``torch.no_grad()``,
+no optimizer step between microbatches), so that result is identical across every
+microbatch and is pure wasted work.
+
+ModelOpt already ships the primitive: :func:`modelopt.torch.quantization.fold_weight`
+writes the fake-quantized value into the weight and disables the weight quantizer, which
+is exactly the frozen-weight steady state. This module only makes that fold *reversible*,
+so it can wrap a single stage of an otherwise-continuing QAT run.
+"""
+
+import contextlib
+from collections.abc import Iterator
+
+import torch
+
+import modelopt.torch.quantization as mtq
+from modelopt.torch.quantization.nn.modules.tensor_quantizer import TensorQuantizer
+
+_QUANTIZER_SUFFIX = "weight_quantizer"
+
+
+def _foldable_weight_quantizers(
+    model: torch.nn.Module,
+) -> Iterator[tuple[torch.Tensor, TensorQuantizer]]:
+    """Yield the ``(weight, quantizer)`` pairs that ``fold_weight`` will actually fold.
+
+    Mirrors ``QuantModule.fold_weight``'s own discovery — any attribute whose name ends in
+    ``weight_quantizer`` holding a ``fake_quant`` ``TensorQuantizer``, paired with the
+    weight named by dropping the ``_quantizer`` suffix. Matching it exactly matters for
+    fused and MoE modules, which expose names like ``w13_weight_quantizer`` and
+    ``gate_up_proj_weight_quantizer``: a plain ``module.weight_quantizer`` lookup would
+    miss those, leaving them folded and disabled after the stage.
+
+    ``SequentialQuantizer`` weight quantizers (W4A4 double-quant) are skipped, because
+    ``fold_weight`` skips them too — it subclasses ``nn.Sequential``, not
+    ``TensorQuantizer`` — so there is nothing to fold or restore for those modules.
+    """
+    for module in model.modules():
+        for name in dir(module):
+            if not name.endswith(_QUANTIZER_SUFFIX):
+                continue
+            quantizer = getattr(module, name, None)
+            if not isinstance(quantizer, TensorQuantizer) or not quantizer.fake_quant:
+                continue
+            weight = getattr(module, name[: -len("_quantizer")], None)
+            if isinstance(weight, torch.Tensor):
+                yield weight, quantizer
+
+
+@contextlib.contextmanager
+def temporarily_fold_weights(
+    model: torch.nn.Module,
+    *,
+    verbose: bool = False,
+    rank: int = 0,
+) -> Iterator[None]:
+    """Fold fake-quantized weights into the parameters for the duration of the block.
+
+    Wraps :func:`modelopt.torch.quantization.fold_weight`, which snaps each weight into
+    its existing parameter storage and disables the weight quantizer, then restores the
+    original weights and re-enables the quantizers on exit. Forwards inside the block read
+    an already-snapped weight and short-circuit out of the disabled quantizer, so the snap
+    happens once per stage instead of once per microbatch.
+
+    ``keep_attrs=True`` is required: without it ``fold_weight`` also deletes the
+    quantizers' ``_amax`` / ``_pre_quant_scale`` (and SVDQuant LoRA) buffers, and
+    re-enabling a quantizer whose calibration state has been dropped would silently change
+    quantization once training resumes.
+
+    Only valid while the weights are frozen — ``no_grad`` with no optimizer step — which
+    is exactly the ``get_logprobs`` re-scoring stage. Folding across a live training step
+    would corrupt the weights, since the fold is written into the parameter itself.
+
+    Activation quantization is unaffected: ``fold_weight`` disables only the *weight*
+    quantizer, so a recipe's ``input_quantizer`` / ``output_quantizer`` keep running and
+    W4A4 logprobs are unchanged.
+
+    Costs one temporary copy of each folded weight shard, held only for the block.
+    """
+    folded = list(_foldable_weight_quantizers(model))
+    original_weights = [(weight, weight.detach().clone()) for weight, _ in folded]
+    enabled_quantizers = [quantizer for _, quantizer in folded if quantizer.is_enabled]
+
+    try:
+        # fold_weight writes through `.data`, but running it under grad would still build
+        # a throwaway graph for the snap itself.
+        with torch.no_grad():
+            mtq.fold_weight(model, keep_attrs=True)
+        yield
+    finally:
+        with torch.no_grad():
+            for weight, original in original_weights:
+                weight.data.copy_(original)
+        for quantizer in enabled_quantizers:
+            quantizer.enable()
+    if verbose and rank == 0:
+        print(
+            f"[weight_folding] frozen-weight stage: folded and restored "
+            f"{len(folded)} weight quantizer(s)."
+        )
--- a/docs/guides/quantization-aware-rl.md
+++ b/docs/guides/quantization-aware-rl.md
@@ -78,6 +78,43 @@ for most models, but it is not guaranteed for every architecture or recipe. If
 you encounter errors with the standard Megatron layer specs, leave it unset or
 set it to `false` to exercise ModelOpt's Megatron layer-spec path.
 
+## Frozen-Weight Logprob Optimization
+
+QARL can avoid repeatedly fake-quantizing the same frozen weights during the
+no-gradient policy and reference logprob passes. Set
+`policy.quant_fold_frozen_weight_snap: true` to enable it; it defaults to `false`.
+
+> **Requires nvidia-modelopt 0.47 or newer**, which is where
+> `mtq.temporarily_fold_weights` was added. NeMo RL currently pins an older
+> commit, so enabling the option raises a `RuntimeError` naming this requirement
+> rather than silently doing nothing. The shipped QARL configs leave it `false`
+> until the pin moves.
+
+When enabled, NeMo RL wraps the pass in ModelOpt's
+`mtq.temporarily_fold_weights`, which folds each fake-quant weight quantizer once
+at the start of the pass: the fake-quantized value is written into the existing
+parameter storage and the weight quantizer is disabled. Forwards during the pass
+then read an already-quantized weight instead of re-quantizing it on every
+microbatch. ModelOpt snapshots the affected weights beforehand and restores them,
+along with the quantizer state, before training resumes -- including if the pass
+raises.
+
+`policy.quant_fold_snapshot_device` selects where those snapshots live. Leave it
+unset to keep each snapshot on its parameter's device, which is fastest; set it to
+`cpu` to trade a host round trip for accelerator memory on memory-tight runs.
+
+This applies to any quantization format. Only the *weight* quantizer is disabled,
+so recipes that also quantize activations (such as W4A4) keep their input and
+output quantizers running and produce unchanged logprobs.
+
+ModelOpt does not support weight quantizers built as a `SequentialQuantizer`
+(as used by double-quantized recipes) and raises `NotImplementedError` for them,
+so do not enable this option for such a recipe. None of the quantization configs
+shipped under `examples/modelopt/quant_configs/` build one.
+
+The option costs one temporary copy of each folded weight shard, held only for the
+duration of the pass.
+
 ## Quantization-Aware GRPO (QA-GRPO)
 
 ### Configuration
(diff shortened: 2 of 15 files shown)


babyplutokurt:
Hi @terrykong 

This PR improve the speed of the QAT Training, by stashing the fake quantization cache and remove unncessary per-forward wrapper. 62% speed up compare to current in-house w4a16 QAT in logprobs stage.

terrykong:
@sharonyu-115 @mxinO could you review?

mxinO:
Thanks for the cotribution, this is a good idea. But this is over complicated solution, modelopt has a `fold_weight` utility, this can be a simple context manager, and we don't need split to two different methods and introduce so many helpers, an example can be,
```python
@contextmanager
def temporarily_fold_weights(model):
    modules = [
        m
        for m in model.modules()
        if hasattr(m, "weight") and hasattr(m, "weight_quantizer")
    ]
    original_weights = [(m.weight, m.weight.detach().clone()) for m in modules]
    active_quantizers = [m.weight_quantizer for m in modules if m.weight_quantizer.is_enabled]

    try:
        mtq.fold_weight(model, keep_attrs=True)
        yield
    finally:
        with torch.no_grad():
            for weight, original in original_weights:
                weight.copy_(original)
        for quantizer in active_quantizers:
            quantizer.enable()
```
and use it in the `get_logprobs`,
```python
def get_logprobs(self, *args, **kwargs):
    if not self.cfg.get("quant_fold_frozen_weight_snap", False):
        return super().get_logprobs(*args, **kwargs)

    with temporarily_fold_weights(self.model):
        return super().get_logprobs(*args, **kwargs)
```

babyplutokurt:
@mxinO   Thanks, this is much better, and I've reimplemented it your way. Pushed as a single commit.

  **Result:** Three context managers and eight helpers collapse to one context manager and one helper; two mutually-exclusive config keys (with precedence rules) collapse to a single
  `policy.quant_fold_frozen_weight_snap`. All the monkeypatching is gone, no more patching `TensorQuantizer.forward`, `DynamicModule.__getattr__`, or MRO `forward` shadowing. `get_logprobs` is now the
  four lines you sketched.

  You were also right that the split was unnecessary. I'd assumed W4A4 needed its own mode, but `fold_weight` disables only the *weight* quantizer, so `input_quantizer`/`output_quantizer` keep running and
  activation-quantized recipes are unaffected. Verified empirically: forward output is bit-identical folded vs. unfolded with `input_quantizer` enabled.

babyplutokurt:
I've moved off `mtq.fold_weight`. It crashes on tied-embedding Megatron models.

  While validating the fold end-to-end (QA-GRPO, Megatron + vLLM), the first `get_logprobs`
  call with the flag enabled died inside the utility:

  ```
  modelopt/torch/quantization/nn/modules/quant_module.py:147, in fold_weight
      weight.data.copy_(attr(weight.float()).to(weight.dtype))
  AttributeError: 'NoneType' object has no attribute 'data'
  ```
  
  **Root cause:** `fold_weight` selects pairs on `fake_quant` alone; it checks neither
  `is_enabled` nor whether the weight is a tensor. On Megatron models with tied word
  embeddings the tied `output_layer` is built with `skip_weight_param_allocation`, so it
  carries `weight = None` and borrows the embedding weight at forward time. ModelOpt still
  attaches an `output_layer.weight_quantizer` (disabled by every standard recipe via
  `'*output_layer*': enable: false`, but `fake_quant` stays `True`), so the filter passes
  and line 147 dereferences the None weight. Reproduced on Qwen3-0.6B
  (`tie_word_embeddings: true`). HF models don't hit it because they express tying as a
  shared tensor rather than `None`, which is why unit tests alone missed it.

  There's no way to filter from outside: `mtq.fold_weight(model)` walks all modules, and
  even per-module `module.fold_weight()` can't skip an individual pair inside a fused
  module. So the context manager now applies the utility's own fold formula (its lines
  147 and 148, verbatim) to each pair it discovers:

  ```python
  weight.data.copy_(quantizer(weight.float()).to(weight.dtype))
  quantizer.disable()
  ```

  with two guards the upstream walk lacks: skip `weight is None`, and skip disabled
  quantizers. A disabled quantizer's forward is the identity, so folding it is a no-op,
  and cloning its weight for restore was pure waste (the disabled `lm_head`/embedding
  pairs hold ~40% of the quantized-weight bytes on Qwen3-0.6B). A side benefit is that
  `keep_attrs` becomes moot, since the `_amax`/`_pre_quant_scale` deletion lives in the
  code path we no longer invoke.

  Same math, verified: logits are bit-identical folded vs unfolded (W4A16 and W4A4,
  bf16, real ModelOpt), the full `state_dict` round-trips bit-identical, and the
  previously crashing E2E run now completes with correct fold/restore counts. Added a
  regression test that builds the tied-`output_layer` shape (`weight=None` plus an
  attached quantizer).

  Worth an upstream ModelOpt issue for the missing guards at `quant_module.py:147`; if
  that lands, this module can go back to delegating.

babyplutokurt:
Filed an upstream issue related to mtq.fold_weight: https://github.com/NVIDIA/Model-Optimizer/issues/2131

babyplutokurt:
fixed ModelOpt upstream at: https://github.com/NVIDIA/Model-Optimizer/pull/2132

babyplutokurt:
@mxinO @sharonyu-115 @terrykong ^^

mxinO:
Thanks, the `fold_weight` issue should be handled in modelopt side, rather than duplicate the logic here, I will make a pr for this https://github.com/NVIDIA/Model-Optimizer/pull/2140

babyplutokurt:
Thanks @mxinO, I looked at the diff in NVIDIA/Model-Optimizer#2140 and mtq.temporarily_fold_weights looks like exactly the right upstream home for this.

Question on sequencing: since this PR is opt-in and already verified against the ModelOpt commit NeMo-RL currently pins, can we merge it first and then migrate the internals to mtq.temporarily_fold_weights in a follow-up once the pin is bumped past #2140? Or should we wait for nemo_RL track updated modelOpt that contains your PR?

babyplutokurt:
@mxinO What's your thoughput? ^^

nemo_rl/modelopt/models/policy/workers/weight_folding.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Beyond the upstream scan, pairs are skipped when:
+
+    * the quantizer is disabled — its forward is the identity, so folding it is a no-op
+      that would only cost a wasted restore clone (recipes routinely disable
+      ``lm_head``/embedding quantizers, which hold ~40% of the quantized-weight bytes);
+    * the weight is not a tensor — Megatron tied-embedding ``output_layer`` modules
+      carry ``weight = None`` and borrow the embedding weight at forward time
+      (upstream ``fold_weight`` crashes on these);
+    * the quantizer is a ``SequentialQuantizer`` (W4A4 double-quant) — it subclasses
```
mxinO: The seq quantizer use case is very rare, we can skip support of it for now.

nemo_rl/modelopt/models/policy/workers/weight_folding.py
```diff
@@ -0,0 +1,140 @@
...
+            if not name.endswith(_QUANTIZER_SUFFIX):
+                continue
+            quantizer = getattr(module, name, None)
+            if (
+                not isinstance(quantizer, TensorQuantizer)
+                or not quantizer.fake_quant
+                or not quantizer.is_enabled
+            ):
+                continue
+            weight = getattr(module, name[: -len("_quantizer")], None)
```
mxinO: This is not reliable, use `iter_weights_for_calibration`

nemo_rl/modelopt/models/policy/workers/weight_folding.py
```diff
@@ -0,0 +1,140 @@
...
+    parameter itself.
+
+    Activation quantization is unaffected: only *weight* quantizers are disabled, so a
+    recipe's ``input_quantizer`` / ``output_quantizer`` keep running and W4A4 logprobs
+    are unchanged.
+
+    Costs one temporary copy of each folded weight shard, held only for the block.
+    """
+    folded = list(_foldable_weight_quantizers(model))
+    original_weights = [(weight, weight.detach().clone()) for weight, _ in folded]
```
mxinO: We might want to move this to cpu.

nemo_rl/modelopt/models/policy/workers/weight_folding.py
```diff
@@ -0,0 +1,140 @@
...
+
+    try:
+        with torch.no_grad():
+            for weight, quantizer in folded:
+                # Exact upstream fold formula (quant_module.py::fold_weight).
+                weight.data.copy_(quantizer(weight.float()).to(weight.dtype))
```
mxinO: this won't work for cases with PQS and rotation, though we don't have use case in RL now.

mxinO (commented):
I still suggest wait of https://github.com/NVIDIA/Model-Optimizer/pull/2132, I will push the merge of it.

mxinO:
https://github.com/NVIDIA/Model-Optimizer/pull/2140 is merged, can you use the new added temeporary fold weight? we can update the modelopt commit pin.

babyplutokurt:
@mxinO Done and pushed

The fold now goes through `mtq.temporarily_fold_weights`. I left `snapshot_device` defaulting to the parameter's device since the snapshot is freed at the end of the stage, with `policy.quant_fold_snapshot_device` available for memory-tight runs, happy to default it to `cpu` if you'd prefer.

On the pin, I couldn't bump it directly: [`c3b913b9`](https://github.com/NVIDIA/Model-Optimizer/commit/c3b913b9cc1d82d5a0af9fa77b4db87829e6f158) → [`7c494088`](https://github.com/NVIDIA/Model-Optimizer/commit/7c494088fa51ccea22af6df466c392756a8243ee) crosses 0.46 → 0.47 and breaks two things in [`vllm_quant_patch.py`](https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/modelopt/models/generation/vllm_quant_patch.py) that are unrelated to weight folding:

1. [`vllm_fused_moe_package`](https://github.com/NVIDIA/Model-Optimizer/blob/c3b913b9cc1d82d5a0af9fa77b4db87829e6f158/modelopt/torch/quantization/plugins/vllm.py#L90) is gone in 0.47, but we still read it at [line 169](https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/modelopt/models/generation/vllm_quant_patch.py#L169).
2. 0.47 [registers `RoutedExperts` itself](https://github.com/NVIDIA/Model-Optimizer/blob/7c494088fa51ccea22af6df466c392756a8243ee/modelopt/torch/quantization/plugins/vllm.py#L710), so our `_register_routed_experts_quant_module` guard early-returns and our subclass never installs, meaning upstream's strict `type(quant_method) is UnquantizedFusedMoEMethod` assert applies, where we had relaxed it to a substring check for vLLM 0.25.

The import is guarded meanwhile, so this PR is safe against the current pin, I'll drop the guard once it lands.

babyplutokurt:
I also set the config files to use `quant_fold_frozen_weight_snap: false` right now to avoid the try catch error.

we can flip them on once the pin is bumped

babyplutokurt:
@mxinO friendly ping on this one, it has been a few weeks since the switch to `mtq.temporarily_fold_weights`.

Status update:

- Rebased onto today's `main` (`d4b44688`), no conflicts. Head is now `4755ee6f6da57c9fe78d60aead2c987e2c342f41`.
- Unit tests pass locally on the mcore extra: `tests/unit/models/policy/test_weight_folding.py` (the four fold contract tests run against the real 0.47 API, plus the worker-routing test) and the existing `tests/unit/models/policy/test_megatron_quant_worker.py`.
- Still safe against the current pin (`c3b913b9`): the option is opt-in, the import is guarded, and all shipped QARL configs carry `quant_fold_frozen_weight_snap: false`, so behaviour on `main` is unchanged.

Could you take another look for approval, and could a maintainer trigger CI with `/ok to test 4755ee6f6da57c9fe78d60aead2c987e2c342f41`?

On the pin: I checked and `main` still pins `c3b913b9`, and the only open PR that moves it (#3566) targets a rev that predates NVIDIA/Model-Optimizer#2140, so it would not bring the fold API either. I'm happy to open the bump as a follow-up PR (pin to `7c494088` or newer, fix the two `vllm_quant_patch.py` call sites, flip the configs to `true`, drop the import guard) once this one lands, unless you would rather own it.

cc @sharonyu-115 @terrykong

PR #3453, opened 2026-07-31
feat: nccl v2.30.7 bump

youngeunkwon0405:
# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


pyproject.toml
```diff
@@ -354,9 +354,9 @@ override-dependencies = [
   "nvidia-cublas==13.5.1.27; sys_platform != 'darwin'",
   "nvidia-cudnn-cu13==9.20.0.48; sys_platform != 'darwin'",
   "nvidia-cudnn-frontend==1.23.0",
-  # NCCL 2.30.4 fixes a context-parallel hang on GB200 when GPUs are non-contiguous.
+  # NCCL 2.30.7 contains fixes required by NCCL M2N resharding.
```
terrykong: [`pyproject.toml:357`](https://github.com/NVIDIA-NeMo/RL/blob/1c9088738c3bcf2ed2ba6a39e8f67821e932fdd9/pyproject.toml#L357)

**1 action item.**

### AI-1

2.30.7 is on the same minor line as 2.30.4, so it still carries the GB200 context-parallel fix — meaning this pin now holds back two defects but records only one. Replacing the comment rather than extending it means whoever next tries to relax or re-float the pin only sees the m2n reason and may conclude the GB200 hang is unrelated. Pre-existing text, but this PR is what drops it.

The m2n claim also has no citation — NCCL publishes no release body for these tags, so it can't be confirmed from public sources. A link (release note or internal bug) makes it checkable.

**Action:** keep both reasons and link the m2n fix. One way to do it — of several:

```suggestion
  # NCCL 2.30.4 fixes a context-parallel hang on GB200 when GPUs are non-contiguous;
  # 2.30.7 additionally fixes hangs in the NCCL m2n reshard (<link to NCCL release note / internal bug>).
```

terrykong: i guess agent makes a good point, we should probably say >=2.30.4 fixes that hang

terrykong (commented):
LGTM on the substance — this is now a clean, self-contained wheel bump. The base-image half is gone (reverted to `26.05-cuda13.2` in 1c9088738c3bcf2ed2ba6a39e8f67821e932fdd9), and with it the CUDA 13.3 toolchain break that was reddening both container builds; [GB200](https://github.com/NVIDIA-NeMo/RL/actions/runs/30871874417/job/92117664322) and [H100](https://github.com/NVIDIA-NeMo/RL/actions/runs/30871874417/job/92117664070) builds are green, as is the rest of CI. The TRT-LLM ceiling raise and the lock regen are consistent with the override.

One small doc ask inline on [`pyproject.toml`](https://github.com/NVIDIA-NeMo/RL/blob/1c9088738c3bcf2ed2ba6a39e8f67821e932fdd9/pyproject.toml#L357); everything below is context.

**Context — no action.**

*The wheel does reach the m2n path.* `nccl4py` 0.3.1 — which provides the `nccl.m2n.reshard` entry point [`xferdtensor.py`](https://github.com/NVIDIA-NeMo/RL/blob/1c9088738c3bcf2ed2ba6a39e8f67821e932fdd9/nemo_rl/weight_sync/xferdtensor.py#L25) imports — declares no dependency on `nvidia-nccl-cu13` (its `uv.lock` deps are `cuda-core`, `cuda-pathfinder`, `numpy`, `packaging`), so it resolves `libnccl` at runtime rather than linking the wheel at install time. That resolution prefers the wheel: at the pinned `cuda-pathfinder` 1.5.6, [`EARLY_FIND_STEPS = (find_in_site_packages, find_in_conda)`](https://github.com/NVIDIA/cuda-python/blob/cuda-pathfinder-v1.5.6/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py#L211-L215) run *before* the already-loaded check and the system search.

*Staying on `26.05` re-introduces a system/wheel NCCL skew* — the image ships 2.30.4 while the override installs 2.30.7. That's the alignment [#3352](https://github.com/NVIDIA-NeMo/RL/commit/f76dbf3d6c8e358c29c8cb6b9b290372cf8a04cb) had established, and it's benign here for the reason above. Worth flagging only because re-aligning is blocked for now: per the NGC tag list for `cuda-dl-base`, `26.05` is the last cuda13.2 build (`26.06` and `26.07` are both cuda13.3), and moving to 13.3 needs a torch that supports it — [`torch==2.11.0`](https://github.com/NVIDIA-NeMo/RL/blob/1c9088738c3bcf2ed2ba6a39e8f67821e932fdd9/pyproject.toml#L364) is a `+cu130` build whose `cpp_extension.py` sets no g++ bounds for 13.3, which is what broke DeepEP and nv-grouped-gemm inside torch's own `List_inl.h:202` header ([reproduced outside this repo](https://github.com/LMCache/LMCache/issues/4090)). So the skew is the right call for now; a follow-up issue to re-align once torch supports 13.3 would be useful.

*Minor:* the PR body is still the unfilled template — a one-liner on what the m2n reshard hang was would help whoever reads this in `git log` later.

_Generated by Claude Code_

terrykong (approved):

Issue #2190, opened 2026-04-02
How save Lora Weight into normal format


hanguangmic:
I used https://github.com/NVIDIA-NeMo/RL/blob/main/examples/configs/recipes/llm/grpo-nanov3-30BA3B-2n8g-megatron-lora.yaml here to train the model and got the dist torch lora weight. but how to make it works  for the inference?
The following is my peft config

{'_target_': 'megatron.bridge.peft.lora.LoRA',
 'a2a_experimental': False,
 'alpha': 64,
 'canonical_mapping': {},
 'dim': 32,
 'dropout': 0.0,
 'dropout_position': 'post',
 'exclude_modules': ['*out_proj*'],
 'lora_A_init_method': 'xavier',
 'lora_B_init_method': 'zero',
 'lora_dtype': 'None',
 'params_to_save': {'decoder.layers.0.mixer.in_proj.adapter.linear_in.weight',
  'decoder.layers.0.mixer.in_proj.adapter.linear_out.weight',
  'decoder.layers.1.mlp.shared_experts.linear_fc1.adapter.linear_in.weight',
  'decoder.layers.1.mlp.shared_experts.linear_fc1.adapter.linear_out.weight',
  'decoder.layers.1.mlp.shared_experts.linear_fc2.adapter.linear_in.weight',
  'decoder.layers.1.mlp.shared_experts.linear_fc2.adapter.linear_out.weight',
  'decoder.layers.10.mlp.shared_experts.linear_fc1.adapter.linear_in.weight',
  'decoder.layers.10.mlp.shared_experts.linear_fc1.adapter.linear_out.weight',
  'decoder.layers.10.mlp.shared_experts.linear_fc2.adapter.linear_in.weight',
  'decoder.layers.10.mlp.shared_experts.linear_fc2.adapter.linear_out.weight',
  'decoder.layers.11.mixer.in_proj.adapter.linear_in.weight',
  'decoder.layers.11.mixer.in_proj.adapter.linear_out.weight',
  'decoder.layers.12.self_attention.linear_proj.adapter.linear_in.weight',
  'decoder.layers.12.self_attention.linear_proj.adapter.linear_out.weight',
  'decoder.layers.12.self_attention.linear_qkv.adapter.linear_in.weight',
  'decoder.layers.12.self_attention.linear_qkv.adapter.linear_out.weight',
  'decoder.layers.13.mlp.shared_experts.linear_fc1.adapter.linear_in.weight',
[...]

PR #3455, opened 2026-07-31
feat(sc): gate-authoritative token capture on the Gym tokidcap stack

pthombre:
## What

Token-in/token-out capture for NeMo-Gym rollouts in the async SingleController GRPO path, migrated onto the upstream Gym token-id-capture PR stack (NVIDIA-NeMo/Gym #2190→#2124→#2125→#2126→#2180→#2181→#2182) per `docs/design-docs/tq-gym-tokidcap-migration.md`. The Gym submodule is pinned to `pthombre/tq-tokidcap-capture` (= stack top `81ac2736` + the 7-commit gate/staging series; companion Gym PR opened against the stack).

With `token_capture.enabled=true`: the Gym gate holds each rollout's token lineage (base `LineageIndex`, capacity sized from the training config), vLLM workers stage per-call token deltas + logprobs directly into TransferQueue before the response releases, agent-facing messages and the Ray return are token-free, and the `BlackboxFinalizer` rebuilds verified always-N training rows from staged deltas. With `enabled=false` (default) every legacy codepath is byte-identical.

Highlights of this migration vs. the previous fork-based MVP:

- **Identity**: no metadata side-channel. `RolloutManager` puts `{group_id}_g{i}` on the run body (`_ng_rollout_id`); agents stamp `/ng-rollout/<id>` on every model call via stack helpers (all agent impls unmodified); the middleware-minted `model_call_id` is the call id. TQ sample ids ARE the capture keys.
- **Lineage**: the stack's dialect-portable assistant-turn fingerprinting replaces the explicit marker (no `openai_utils`/`responses_converter` plumbing); fallback causes are instrumented (`ambiguity` separately — the data-driven trigger for a future marker).
- **Security/robustness pulled into Stage 1**: control plane requires a per-run bearer token (a sandboxed harness could otherwise forge a sibling's seal); every control-plane call has a hard deadline (gate death surfaces as failed dispatches + placeholders instead of a silent retry stall); terminal-aware chain selection prevents silent training on a sub-agent fork.
- Rebuild delegates to the stack's `run_builder(prefix_merging)`; unresolved final-call retries publish placeholders instead of training on a guess.

Working log with the workaround ledger (upstream asks → contained workarounds → what deletes when each ask lands) and a stack-bump checklist: `docs/design-docs/tq-gym-tokidcap-migration-log.md`.

## Tests

- Gym side (submodule): full unit suite at the pin — 1925 passed / 0 failed.
- RL side: capture unit suites (`--nemo-gym-only`) + replay buffer / rollout manager / SC setup suites; 2-GPU flag-off functional (`grpo_async_gym_single_controller.sh`) as the pin-bump regression and the same with `++token_capture.enabled=true` (results recorded in the migration log).


pthombre:
Companion Gym PR (the gate/staging layer on the tokidcap stack): NVIDIA-NeMo/Gym#2278

/ok to test 624bb277d92a28bcaa846ca4ab75fe2ca365214f

PR #3457, opened 2026-08-02
feat(vllm): route NeMo-Gym rollouts through vLLM Router

aoshen02:
## Summary
- allow NeMo-Gym to use an external vLLM Router endpoint
- keep NeMo RL-owned Ray workers for lifecycle and refit/sync
- preserve the existing per-worker URL behavior when no router is configured

## Validation
- `git diff --check`
- `/home/inf-aoshen/vllm/.venv/bin/python -m compileall -q nemo_rl/models/generation/vllm nemo_rl/algorithms/single_controller_utils/setup.py nemo_rl/algorithms/distillation.py nemo_rl/algorithms/grpo.py`

This change is AI-assisted and has been reviewed locally. It is not a duplicate of the open NeMo-Gym or Responses API PRs: those do not route NeMo-Gym's vLLM backend through an external vLLM Router while retaining NeMo RL worker ownership.

### DIFF
--- a/nemo_rl/algorithms/distillation.py
+++ b/nemo_rl/algorithms/distillation.py
@@ -544,7 +544,7 @@ def setup(
                     nemo_gym_dict.setdefault("uv_venv_dir", uv_venv_dir)
                 nemo_gym_cfg = NemoGymConfig(
                     model_name=generation_config["model_name"],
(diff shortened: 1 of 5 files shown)


aoshen02:
Retargeting this change to the author fork main branch as requested.

PR #3458, opened 2026-08-02
feat(vllm): route native async rollouts through vLLM Router

aoshen02:
## Summary
- add an HTTP `/v1/completions` path for NeMo RL native async rollouts
- send pre-tokenized prompts through an external vLLM Router
- preserve prompt/token IDs, logprobs, truncation, and optional routed-expert metadata
- keep Ray workers responsible for lifecycle and weight refit/sync

## Validation
- `git diff --check`
- `/home/inf-aoshen/vllm/.venv/bin/python -m compileall -q nemo_rl/models/generation/vllm`

This change is AI-assisted and has been reviewed locally. It is separate from the open NeMo-Gym/Responses API work: it implements the native async generation transport and token-preserving completion response required by NeMo RL.


aoshen02:
Retargeting this change to the author fork main branch as requested.

PR #3463, opened 2026-08-03
feat(draft): support sequence packing with online eagle3 draft training

yuekaizhang:
## What does this PR do?

Adds sequence packing support to online EAGLE-3 draft training (previously the
two features were mutually exclusive). With packing enabled, policy training
time is cut roughly in half (~2100 vs ~1040 tok/s/GPU on a Qwen3-30B-A3B 8K
GRPO run), so draft co-training no longer pays the unpacked-training penalty.

Key changes:
- Per-segment teacher-logit/embedding shifts in the packed [1, T] layout
  (boundary-safe rolls; no leakage across packing boundaries)
- Packed draft-loss path in `DraftLossWrapper` with a packed token mask
- THD attention support in the eagle module forward (`padding_causal`
  mask-type override for TE kernels)
- Guard relaxed from draft ⊥ sequence-packing to draft ⊥ CP>1

Unit tests prove packed/unpacked numerical equivalence (full-vocab and
d2t-mapped drafts); verified end-to-end on 1n8g (8B) and 4n8g (30B-A3B) GRPO
runs with no acceptance regression.

Blue curve: without the PR
Orange curve: with the PR

image


yuekaizhang:
@yuki-97  @isomap  Hi, I was wondering if you could help review the PR when you have some time. Many thanks!

yuekaizhang:
/ok to test [ac8845e](https://github.com/NVIDIA-NeMo/RL/pull/3463/commits/ac8845ee2d60ace846ee6dcd73b0d30d51976ffa)

sharonyu-115:
Hi @yuekaizhang, you can flag for @seonjinn to review this PR once it is ready. Thanks.

yuekaizhang:
Hi @seonjinn, the pr is ready. Would you mind helping review it? Thanks!

nemo_rl/models/megatron/train.py
```diff
@@ -245,16 +246,35 @@ def forward_with_post_processing_fn(
...
+            # correct sequence-parallel layout for free). no_grad matches the
+            # capture hooks, which hand the draft detached embeddings.
+            with torch.no_grad():
+                shifted_input_ids = _pack_input_ids(
+                    data_dict["input_ids"],
+                    packed_seq_params.cu_seqlens_q,
+                    packed_seq_params.cu_seqlens_q_padded,
+                    roll_shift=-1,
+                )
+                shifted_input_embeds = capture.model.embedding(
```
seonjinn, line 331: [`train.py:263`](https://github.com/NVIDIA-NeMo/RL/blob/6efb671054934a622f2726d203ced415222a40da/nemo_rl/models/megatron/train.py#L249-L265)

With `policy.draft.enabled=true`, sequence packing enabled, and `pipeline_model_parallel_size > 1`, this path runs on the [post-process chunk that owns the draft model](https://github.com/NVIDIA-NeMo/RL/blob/6efb671054934a622f2726d203ced415222a40da/nemo_rl/models/megatron/draft/utils.py#L969-L979). However, [pinned MCore constructs `GPTModel.embedding` only on pre-process/MTP stages](https://github.com/NVIDIA/Megatron-LM/blob/6513e3e23d6b5eda6a1c934990b15e804237732b/megatron/core/models/gpt/gpt_model.py#L167-L175), so this call raises `AttributeError: 'GPTModel' object has no attribute 'embedding'`.

**Action:** compute and transport the shifted embeddings from the embedding-owning stage, or reject PP>1 during setup and add a packed PP regression test.

nemo_rl/algorithms/loss/wrapper.py
```diff
@@ -248,10 +266,61 @@ def __init__(
...
+        self.student_logits = student_logits
+        if cu_seqlens_q is not None and student_logits is None:
+            raise ValueError("student_logits must be passed explicitly in packed mode.")
+        if cu_seqlens_q is None and prepare_fn is None:
+            raise ValueError("prepare_fn is required in unpacked mode.")
         self.draft_loss_fn = DraftCrossEntropyLossFn(
             vocab_parallel_group=vocab_parallel_group,
         )
 
+    def _packed_draft_loss(
```
seonjinn, line 283: [`wrapper.py:283`](https://github.com/NVIDIA-NeMo/RL/blob/6efb671054934a622f2726d203ced415222a40da/nemo_rl/algorithms/loss/wrapper.py#L283-L322)

**1 action item.**

This PR changes convergence-sensitive online EAGLE training loss and packed execution semantics, but it currently uses `CI:Lfast`. The repository's [CI policy assigns convergence-affecting changes to `CI:L2`](https://github.com/NVIDIA-NeMo/RL/blob/6efb671054934a622f2726d203ced415222a40da/.agents/contributor-skills/cicd/SKILL.md#L43-L64).

**Action:** switch the PR to `CI:L2` and rerun CI for `6efb671054934a622f2726d203ced415222a40da`.

nemo_rl/models/policy/lm_policy.py
```diff
@@ -122,12 +122,17 @@ def __init__(
                 "policy.draft.enabled=true is only supported with the Megatron backend. "
                 "Set policy.megatron_cfg.enabled=true or disable policy.draft."
             )
-        if draft_enabled and bool(
-            config.get("sequence_packing", {}).get("enabled", False)
+        if (
+            draft_enabled
+            and int(config.get("megatron_cfg", {}).get("context_parallel_size", 1)) > 1
```
seonjinn: [`lm_policy.py:127`](https://github.com/NVIDIA-NeMo/RL/blob/6efb671054934a622f2726d203ced415222a40da/nemo_rl/models/policy/lm_policy.py#L125-L128)

**1 action item.**

`context_parallel_size` is required and is indexed directly later, so the new fallback `1` conflicts with the repository's [centralized config-default convention](https://github.com/NVIDIA-NeMo/RL/blob/6efb671054934a622f2726d203ced415222a40da/.agents/contributor-skills/config-conventions/SKILL.md).

**Action:** read the required value directly here: `config["megatron_cfg"]["context_parallel_size"]`.

seonjinn (commented):
Self-review completed with a coordinated multi-agent pass and independent adversarial verification. Three actionable items survived review. Existing PR-head Linux CI passed; local macOS pre-commit could not run because the lockfile supports Linux only

seonjinn:
@yuekaizhang Could you also include bigger model test results as well?  like Qwen3-235B.

nemo_rl/models/megatron/train.py
yuekaizhang: Good catch — fixed in 5cac01d6e: packed draft training with `pipeline_model_parallel_size > 1` is now rejected at setup with a clear error (the packed path re-embeds the shifted ids via the model embedding, which only exists on the first PP stage), and guard regression tests were added in `tests/unit/models/policy/test_lm_policy_draft_guards.py`. Transporting embeddings across stages is left as a follow-up if PP support becomes needed.

nemo_rl/algorithms/loss/wrapper.py
yuekaizhang: Done — switched the PR label to `CI:L2` and re-triggered CI on the new head.

nemo_rl/models/policy/lm_policy.py
yuekaizhang: Fixed in 5cac01d6e: the required `megatron_cfg` parallel sizes are now read by direct indexing (the guard runs after the draft-requires-Megatron check, so `megatron_cfg` is guaranteed present). `sequence_packing` stays on `.get` since it is `NotRequired` in `PolicyConfig`.

yuekaizhang:
@seonjinn Thanks for the review — all three action items are addressed (packed-draft PP>1 guard + regression tests, direct config indexing, and the label is now CI:L2).

Re: Qwen3-235B — we haven't run it yet; results so far are from Qwen3-8B (1n8g) and Qwen3-30B-A3B (4n8g). We could run a 235B validation as a follow-up if there is suitable pretrained ealge-3 draft model.

yuekaizhang:
/ok to test [b5be6b9](https://github.com/NVIDIA-NeMo/RL/pull/3463/commits/b5be6b97270c0065aa3945d8e1af921ea1ca687c)

yuekaizhang:
/ok to test [41196b8](https://github.com/NVIDIA-NeMo/RL/pull/3463/commits/41196b84da04831fbe21a23551bd2f1ce2949e67)

nemo_rl/models/megatron/train.py
```diff
@@ -544,6 +564,31 @@ def __call__(
                 vocab_parallel_group=get_tensor_model_parallel_group(),
                 context_parallel_group=get_context_parallel_group(),
             )
+            if "student_logits" in data_dict:
```
yuki-97, line 572: This guard can't fire on the branch it's on. For GRPO, [`grpo.py:672`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/nemo_rl/algorithms/grpo.py#L672) already rejects fused + sequence packing at setup, so the packed arm never sees `use_fused_linear_logprobs=true`. (It is live for SFT, which has no fused+packing guard.)

The reachable combination is fused + draft with packing **off** — the [`else:` arm below](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/nemo_rl/models/megatron/train.py#L601) and the layout of [the shipped eagle3 recipe](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/examples/configs/recipes/llm/grpo-qwen3-1.7b-1n8g-megatron-eagle3.yaml#L28-L29). Nothing guards it, and draft is absent from the fused limitations list at [`grpo.md:397-401`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/docs/guides/grpo.md#L397-L401), so it dies on a 2-D teacher tensor inside `map_teacher_logits_to_draft_vocab` instead of printing this message.

**Suggest fix:** move it beside the existing draft guards in [`lm_policy.py:126-155`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/nemo_rl/models/policy/lm_policy.py#L126-L155) as a setup-time `raise ValueError` — one place, both layouts, all algorithms — and add the constraint to the `grpo.md` list.

nemo_rl/algorithms/loss/utils.py
```diff
@@ -34,6 +34,98 @@
...
+        # The packer absorbs bin-level alignment padding into the last
+        # sequence's effective length (see _pack_input_ids), so cu_seqlens can
+        # exceed the unpacked row width; only real tokens carry mask.
+        copy_len = min(actual_len, token_mask.shape[1])
+        if copy_len <= 1:
+            continue
+        packed[packed_start : packed_start + copy_len - 1] = (
+            token_mask[i, 1:copy_len] * sample_mask[i]
+        )
+    return packed.unsqueeze(0)
```
yuki-97: This loop re-implements [`_pack_input_ids`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/nemo_rl/algorithms/loss/utils.py#L292-L339)'s padded-offset walk plus [`roll_packed_seq_dim`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/nemo_rl/algorithms/loss/utils.py#L73-L94), including the `min(actual_len, row_width)` clamp verbatim — and this copy of the clamp is unexercised. The fixture's `cu_seqlens = [0, 5, 12, 16]` are unpadded and all <= `max_len = 7`, but production passes `packed_seq_params.cu_seqlens_q`, which *is* `cu_seqlens_padded` ([`data.py:1129`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/nemo_rl/models/megatron/data.py#L1129)) with the bin-fill deficit folded into its last entry ([`data.py:1031-1036`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/nemo_rl/models/megatron/data.py#L1031-L1036)). Dropping the `min()` keeps the suite green but raises a shape-mismatch `RuntimeError` under draft + packing + FP8 ([`data.py:1306-1320`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/nemo_rl/models/megatron/data.py#L1306-L1320)).

**Suggest fix:** compose the two existing helpers — element-wise identical on the fixture, on `copy_len <= 1`, and on an inflated last segment, and it inherits the clamp's regression test at [`test_sequence_packing_fusion.py:443`](https://github.com/NVIDIA-NeMo/RL/blob/41196b84da04831fbe21a23551bd2f1ce2949e67/tests/unit/algorithms/test_sequence_packing_fusion.py#L443):

```python
    packed = _pack_input_ids(
        token_mask * sample_mask.unsqueeze(-1), cu_seqlens, cu_seqlens_padded
    )
    return roll_packed_seq_dim(packed, cu_seqlens_padded, seq_dim=1)
```

(`_pack_input_ids(..., roll_shift=-1)` alone won't do — `seq.roll` is circular within `padded_len` and wraps `mask[0]` into each segment's last slot.)

yuki-97: Can reuse the existing helpers instead of re-implementing `_pack_input_ids`'s padded-offset walk (including its untested `min(actual_len, ...)` clamp) inline:

```python
    packed = _pack_input_ids(
        token_mask * sample_mask.unsqueeze(-1), cu_seqlens, cu_seqlens_padded
    )
    return roll_packed_seq_dim(packed, cu_seqlens_padded, seq_dim=1)
```

Also add a unit test that exercises `actual_len > token_mask.shape[1]` (padded segment inflated past the unpacked row width) — none of the current fixtures (`test_draft_loss_wrapper.py`, `test_sequence_packing_fusion.py:443`) trigger that clamp branch.

nemo_rl/models/megatron/train.py
yuekaizhang: You're right — for GRPO the packed-branch assert was dead code, and the live fused+draft+unpacked combination fell through to a shape error. Fixed in 68f1b8cd3: the check now lives with the other draft guards in `lm_policy.py` as a setup-time `ValueError` (both layouts, all algorithms), the dead assert is removed, the constraint is documented in the grpo.md fused-limitations list, and `test_lm_policy_draft_guards.py` covers both layouts.

nemo_rl/algorithms/loss/utils.py
yuekaizhang: Done in 68f1b8cd3 — `pack_rolled_draft_token_mask` now composes `_pack_input_ids` + `roll_packed_seq_dim` exactly as suggested, and a new unit test exercises the bin-alignment clamp (`cu_seqlens` inflated past the unpacked row width). The existing packed-vs-unpacked equivalence tests confirm the refactor is numerically identical.

yuekaizhang:
/ok to test [e63c4f9](https://github.com/NVIDIA-NeMo/RL/pull/3463/commits/e63c4f969c20b1b4205762d27732c7e9f0fe5d0f)

yuekaizhang:
/ok to test [349dbcd](https://github.com/NVIDIA-NeMo/RL/pull/3463/commits/349dbcdacaf157079236fe38489aa5309b234507)

yuekaizhang:
/ok to test [af6ebb4](https://github.com/NVIDIA-NeMo/RL/pull/3463/commits/af6ebb49277f034efce424c9e99cde5f031c70cd)

yuki-97 (approved):

yuki-97:
af6ebb4 passed on `CI:L1` (run w/ `CI:L2`) except the PPO tests already failed in main branch, merged main and changed to `CI:Lfast` for merge.

PR #3468, opened 2026-08-03
feat: add inflight batchsize and kvcache size metrics and for trtllm rollout

Superjomn:
# What does this PR do ?

Add several metrics for inspect the inflight batch size and kvcache usage.

wandb [link](https://wandb.ai/nvidia/nemotron-swe-debugging-chunweiy/runs/wip6d45a?nw=nwuserchunweiy)

image
image

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


yuki-97 (commented):
@RayenTian to review

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
```diff
@@ -256,7 +290,248 @@ async def post_init_async(self) -> None:
...
+                "📋[TRT-LLM Metric Logger] stats drain task started",
+                flush=True,
+            )
+
+    # ------------------------------------------------------------------ #
+    #  In-flight batching telemetry
+    # ------------------------------------------------------------------ #
+
+    @staticmethod
+    def _get_stat_field(obj: Any, *keys: str) -> Any:
```
hchings, line 295: I wonder whether we can put all metrics utilities (same for the ones in `trtllm_generation.py`) into some sort of Mixin  or utility class, so we can put and manage all/most of the metric-related implementation into a separate python file.  I much prefer we don't bake metrics utils into the current three core trtllm_*.py files, it makes it hard to maintain.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
```diff
@@ -0,0 +1,89 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Inject a sync, Ray-picklable iteration-stats fetch into TRT-LLM's worker.
+
+This lets the TRT-LLM in-flight-batching metrics logger read per-iteration
+stats via ``collective_rpc``.
+
+Why this is needed (TRT-LLM 1.3.0rc21, ``_torch`` ``AsyncLLM`` + ``RayExecutor``):
```
hchings: Has this MR been tested w/ newer TRTLLM version? rc21 is a bit old. 

Also nit: can we streamline these agent comments to make it much shorter?

nemo_rl/models/generation/trtllm/trtllm_generation.py
```diff
@@ -132,7 +132,9 @@ def __init__(
         # broadcast; colocated shares the policy's NCCL group so don't touch it.
         env_vars: dict[str, str] = {}
         if not self.colocated_enabled:
-            env_vars["NCCL_CUMEM_ENABLE"] = "1"
+            env_vars["NCCL_CUMEM_ENABLE"] = os.environ.get(
+                "NRL_GEN_NCCL_CUMEM_ENABLE", "1"
```
hchings: What is the motivation of this change? 
I saw other places like `nemo_rl/models/megatron/setup.py:333` still have this hard coded as 1 due to some know issue.

nemo_rl/models/generation/trtllm/config.py
```diff
@@ -49,6 +49,15 @@ class TrtllmSpecificArgs(TypedDict):
     #   "qwen3_coder" -> Qwen3CoderToolParser  (XML format: <function=...>)
     tool_parser: NotRequired[str]
     reasoning_parser: NotRequired[str]
+    # Enable per-iteration in-flight batching telemetry. When set, the async
+    # worker constructs the engine with enable_iter_perf_stats=True and drains
+    # get_stats_async() into per-worker inflight-batch-size timelines (the
+    # TRT-LLM analog of vLLM's per_worker_inflight_batch_sizes). Requires
+    # async_engine=True.
+    enable_trtllm_metrics_logger: NotRequired[bool]
```
hchings, line 54: Do we know whether there's a perf impact when the metric logger is on? Or is it neglectable?

hchings, line 54: We should also discuss w/ Nemo-RL team on a proper place to add TRTLLM docs, including documented these metrics utils. Not a blocker for ML-perf, so I'm fine if it's a follow-up MR.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
hchings: Mutating a vendored library’s source file on disk seems to be quite a WAR. I'm not sure whether we want this in main and what's the longer-term solution, if any.

nemo_rl/models/generation/trtllm/trtllm_generation.py
```diff
@@ -435,6 +437,65 @@ def finish_generation(self, *args: Any, **kwargs: Any) -> bool:
             print(f"Error in finish_generation: {e}")
             return False
 
+    def get_logger_metrics(self) -> dict[str, Any]:
+        """Collect in-flight batching telemetry from the DP-leader workers.
+
+        Returns the same shape the consumer (algorithms/utils.py) asserts on:
+        ``{metric_name: {dp_idx: list[...]}}`` with ``inflight_batch_sizes`` and
+        ``num_pending_samples`` keys. Mirrors vLLM's get_vllm_logger_metrics.
```
hchings: nit: docstring not fully aligned w/ codes. (e.g., kv_cache_usage_perc)

nemo_rl/models/generation/trtllm/config.py
hchings, line 54: Should also add a few unittests for the newly added functionalities.

nemo_rl/models/generation/trtllm/trtllm_generation.py
Superjomn: This change is unnecessary, removed.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
RayenTian: Patching the vendored file on disk is fine, but could we follow the approach in [nemo_rl/models/generation/vllm/patches.py](https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/models/generation/vllm/patches.py) and also rename this file into nemo_rl/models/generation/trtllm/patches.py? It reads the file back after writing to confirm the patch landed and warns if it didn't. The current bare open(..., "a+") append swallows errors, so if it silently fails the metrics just show up as a flat line of zeros with no warning.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
```diff
@@ -0,0 +1,64 @@
...
```
RayenTian: The copyright year should be the current year (2026) for a new file, matching the other files added here.

```suggestion
# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```

examples/configs/grpo_math_1B_trtllm.yaml
```diff
@@ -17,6 +17,8 @@ policy:
       # diverge from grpo.async_grpo (single source of truth).
       in_flight_weight_updates: ${grpo.async_grpo.in_flight_weight_updates}
       recompute_kv_cache_after_weight_updates: ${grpo.async_grpo.recompute_kv_cache_after_weight_updates}
+      enable_trtllm_metrics_logger: false # Set to true to enable TRT-LLM per-iteration IFB/kv-cache telemetry (async_engine only); turn off for better performance
+      trtllm_metrics_logger_interval: 5.0 # Interval in seconds to collect TRT-LLM logger metrics
```
RayenTian: Just a question, not a blocker — the default here is `5.0s`, while vLLM uses `0.5s` ([`grpo_math_1B.yaml:400`](https://github.com/NVIDIA-NeMo/RL/blob/84132d96eb2bacf6d3deeb59c2e2f3af0344f8ae/examples/configs/grpo_math_1B.yaml#L400)). With a short rollout, would 5s sample often enough to catch the IFB/kv-cache spikes, or might a smaller interval be worth it? Happy either way if 5s is intentional.

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
```diff
@@ -256,7 +280,200 @@ async def post_init_async(self) -> None:
...
+            )
+
+    async def _drain_stats_loop(self) -> None:
+        """Continuously drain per-iteration stats from the engine.
+
+        Records the in-flight batching size (context + generation requests
+        scheduled this iteration).
+        """
+        assert self.llm is not None
+        interval = float(self.cfg["trtllm_cfg"]["trtllm_metrics_logger_interval"])
```
RayenTian, line 345: The interval is read without checking it's set and positive, unlike vLLM ([`vllm_worker_async.py:248-254`](https://github.com/NVIDIA-NeMo/RL/blob/84132d96eb2bacf6d3deeb59c2e2f3af0344f8ae/nemo_rl/models/generation/vllm/vllm_worker_async.py#L248)). Enabling the logger without this key gives a bare `KeyError`, and `0` would busy-loop the drain task. Consider mirroring vLLM's assert (present and `> 0`).

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
```diff
@@ -0,0 +1,64 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Inject a sync, Ray-picklable ``fetch_stats_serialized`` onto TRT-LLM's BaseWorker.
+
+Needed because ``collective_rpc("fetch_stats")`` returns unpicklable C++ bindings,
+the RPC-client ``get_stats()`` path hits a stats-less engine and always returns [],
+and the built-in serialized fetches are async (unusable via the sync
+``RayGPUWorker.call_worker_method``). This sync method returns JSON strings that
+survive the Ray RPC round-trip. Loaded in RayGPUWorker processes via a ``.pth``
```
RayenTian: This says the patch reaches the worker processes "via a `.pth` file" (also on line 62), but the code doesn't create any `.pth` — [`_install_ifb_stats_patch`](https://github.com/NVIDIA-NeMo/RL/blob/84132d96eb2bacf6d3deeb59c2e2f3af0344f8ae/nemo_rl/models/generation/trtllm/trtllm_worker_async.py#L342) appends the patch to `base_worker.py` instead. Could we update the wording to match, so the next reader doesn't go looking for a `.pth` file that isn't there?

nemo_rl/models/generation/trtllm/trtllm_generation.py
```diff
@@ -435,6 +435,59 @@ def finish_generation(self, *args: Any, **kwargs: Any) -> bool:
...
+                logger_metrics["inflight_batch_sizes"][dp_idx] = inflight
+            pending = stats.get("num_pending_samples")
+            if pending:
+                logger_metrics["num_pending_samples"][dp_idx] = pending
+            kv = stats.get("kv_cache_usage_perc")
+            if kv:
+                logger_metrics["kv_cache_usage_perc"][dp_idx] = kv
+        return logger_metrics
+
+    def clear_logger_metrics(self) -> None:
```
RayenTian, line 482: Small nit: this override has no docstring, while its sibling `get_logger_metrics` (line 438) and the vLLM analog ([`vllm_generation.py:1117`](https://github.com/NVIDIA-NeMo/RL/blob/84132d96eb2bacf6d3deeb59c2e2f3af0344f8ae/nemo_rl/models/generation/vllm/vllm_generation.py#L1117)) both have one.

```suggestion
    def clear_logger_metrics(self) -> None:
        """Clear logger metrics for performance reporting."""
```

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
```diff
@@ -256,7 +280,200 @@ async def post_init_async(self) -> None:
...
+                    if isinstance(stats, (str, bytes, bytearray)):
+                        try:
+                            stats = json.loads(stats)
+                        except (ValueError, TypeError):
+                            stats = None
+                    if stats is not None:
+                        ifb = self._get_stat_field(
+                            stats, "inflightBatchingStats", "numScheduledRequests"
+                        )
+                        ctx = self._get_stat_field(
```
RayenTian: Just to confirm my understanding — this fills `num_pending_samples` from `numContextRequests`, which is the number of context/prefill requests **already scheduled** in the current batch, right? On the vLLM side the same key comes from `vllm:num_requests_waiting`, which is ["Number of requests waiting to be processed"](https://github.com/vllm-project/vllm/blob/70b84f0bcbb6d0a35b74b1035673a1c934089dbb/rust/src/metrics/src/scheduler.rs#L211) — i.e. still queued, not yet running. Since both plot under the same "num_pending_samples", could we use a field closer to the waiting-queue count, or rename/note this so the two backends aren't compared as if they mean the same thing?

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
```diff
@@ -256,7 +280,200 @@ async def post_init_async(self) -> None:
...
+                if _err_logged < 5:
+                    _err_logged += 1
+                    print(
+                        f"⚠️[TRT-LLM Metric Logger] stats drain error: {e}",
+                        flush=True,
+                    )
+            # Back off between cycles to avoid busy-spinning when the engine returns immediately.
+            await asyncio.sleep(interval)
+
+    def get_trtllm_logger_metrics(self) -> dict[str, Any]:
```
RayenTian, line 434: Non-blocker: vLLM also tracks a `generation_tokens` series ([`vllm_worker_async.py:283`](https://github.com/NVIDIA-NeMo/RL/blob/84132d96eb2bacf6d3deeb59c2e2f3af0344f8ae/nemo_rl/models/generation/vllm/vllm_worker_async.py#L283)), but the TRT-LLM side doesn't.

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
RayenTian: Agree with pulling this out of the core file. Two parts:

- **Patch machinery** (`_install_ifb_stats_patch` + `_ifb_stats_patch.py`): move it into the `nemo_rl/models/generation/trtllm/patches.py` suggested in [this comment](https://github.com/NVIDIA-NeMo/RL/pull/3468#discussion_r3747263774), so the vendored-file patching lives in one place (mirroring `vllm/patches.py`).
- **The get/clear/drain metrics methods**: I'd avoid a Mixin — the drain reads `self.llm`/`self.cfg` and mutates the buffers on `self`, so a Mixin would turn that host dependency into an implicit, unchecked contract. Since vLLM also keeps its metrics logger inline in [`vllm_worker_async.py`](https://github.com/NVIDIA-NeMo/RL/blob/84132d96eb2bacf6d3deeb59c2e2f3af0344f8ae/nemo_rl/models/generation/vllm/vllm_worker_async.py#L233), I think leaving these as methods here is fine for now; extracting these into a small collector object (rather than a Mixin) could be a follow-up if the author is interested, but I’m also fine leaving them here for now. Happy to get @yuki-97 's take on this part as well.

RayenTian:
Thanks for adding this @Superjomn ! Left some comments.

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
Superjomn: Change this to use [numQueuedRequests](https://github.com/nvidia/TensorRT-LLM/blob/07a559193f73a64de3d04d40e43cc7b7695a427d/cpp/include/tensorrt_llm/executor/types.h#L388) instead.

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
Superjomn: Applied with a new `patches.py`.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
Superjomn: Let me verify with the latest RC; I will update here.
And cleaned up the comments.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
Superjomn: Applied with a new patches.py

nemo_rl/models/generation/trtllm/trtllm_generation.py
Superjomn: Updated.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
Superjomn: Updated.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
Superjomn: Ideally, it should be compatible with newer trtllm, so I removed "1.3.0rc21" keywords from the code.

examples/configs/grpo_math_1B_trtllm.yaml
Superjomn: Updated the defaults to 0.5 to align with vllm.

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
Superjomn: Added two asserts on L341 and L342 as above.

nemo_rl/models/generation/trtllm/config.py
Superjomn: 4 tests added — tests/unit/models/generation/trtllm/test_trtllm_patches.py (patch idempotency across co-located replica actors; read-back warning path) and 2 _drain_stats_loop tests in test_trtllm_worker_async.py (numQueuedRequests preferred, numContextRequests fallback)

Superjomn: the drain loop is one collective_rpc per interval per DP leader, off the request path; it is gated behind enable_trtllm_metrics_logger and the yaml comment documents turning it off for best perf. State honestly whether a measurement was taken — do not invent numbers.

nemo_rl/models/generation/trtllm/_ifb_stats_patch.py
Superjomn: Fixed.

PR #3470, opened 2026-08-03
feat(sc): contain rollout failures in the SingleController path

asolergi-nv:
Part 1/4 of #3454

# What this fixes

Two failure modes in the SingleController async-GRPO path, both **silent**. Neither produces an exception, a log line, or an unhealthy Ray actor — which is why they have not been noticed.

**1. The GRPO path silently corrupts training.** `AsyncRolloutImpl` caught every exception from generation and committed the rollout anyway, with empty text and zero reward. Those rows enter the GRPO leave-one-out baseline as legitimate zero-reward samples, so a transient generation failure does not fail the run — it quietly biases the advantage estimate. This is a live data-correctness bug independent of any resiliency feature.

**2. The NeMo-Gym path silently wedges.** Gym's HTTP client is constructed with `ClientTimeout()` — every timeout disabled — and retries `ClientOSError` / `ServerDisconnectedError` in an uncapped loop. A dead vLLM endpoint therefore parks the rollout forever. Each parked rollout holds a `max_inflight_prompts` permit, so the rollout pump eventually blocks entirely while the train pump spins on an empty buffer. Ray reports every actor healthy throughout.

# What it does

- **Stops swallowing generation errors** (`rollout_manager.py`). The single highest-value change here: failures are classified and raised instead of being committed as zero-reward rows.
- **A failure taxonomy** separating infrastructure failures (retry elsewhere) from data failures (retrying will not help). HTTP is classified **by status, not exception type** — a `ClientResponseError` carrying 503 is infrastructure, one carrying 400 is data.
- **Deadlines** on rollouts, generation turns and environment steps, all defaulting to`None` (disabled).
- **Re-dispatch with two separate budgets.** A prompt that fails on infrastructure is re-dispatched to a different engine rather than dropped; a prompt that fails on its own data is retried a smaller number of times and then either fails the run or is skipped under an explicit budget. Two budgets because "the engine died" and "this prompt is malformed" deserve opposite treatment.
- **A watchdog** that publishes rollout counters and reports stalls — the failures that otherwise produce no signal at all.
- **Partial NeMo-Gym re-dispatch**: only the rows that never arrived are re-sent, rather than the whole prompt group.

Inert by default — all timeouts default to `None`. The two always-on changes are that errors stop being swallowed and retries are bounded; both convert silent failure into loud failure, which can surface pre-existing problems that were previously invisible. That is the intent.

# Tests 

- Unit tests for the taxonomy, deadlines, re-dispatch budgets, watchdog, and partial gym re-dispatch, including a regression test for the stall condition above.
- `ruff` and `pyrefly` clean.
- End-to-end on 2xA6000: `grpo_dp_single_controller.sh` (all 5 metric checks) and `grpo_async_gym_single_controller.sh` (10 steps, median `gen_kl_error` 0.0010579, reward 0.5).

## The chaos harness

`tests/functional/grpo_dp_single_controller_chaos.sh` SIGKILLs a generation worker and asserts the job fails **fast and attributably** rather than wedging. It found the stall-detector bug above, and it is **registered in the SingleController L1 lane** 

Registered **twice**, because Ray retitles a worker for the duration of each call and the two states fail by different routes:

| variant | victim | outcome |
|---|---|---|
| default (`idle`) | `ray::VllmAsyncGenerationWorker` | 222s, `RolloutStall` — the loss must be *detected* |
| `VICTIM_STATE=serving` | `ray::VllmAsyncGenerationWorker.generate_async` | 12s, `GenerationUnavailable` + `RolloutRedispatchExhausted` — an in-flight RPC dies |

Victim selection is pinned rather than left to `pgrep ... | head -1`, which matched three distinct processes across five states and so silently chose a different scenario run to run. It never failed — every scenario ends in a bounded attributable failure, which is all the test asserted — so the divergence was only visible as wall-clock. Three consecutive `idle` runs now give an identical victim title and 222s.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


asolergi-nv:
/ok to test [97b309a](https://github.com/NVIDIA-NeMo/RL/pull/3470/commits/97b309aaa6114bfae177796c0e90f3691af149d4)

asolergi-nv:
/ok to test [97b309a](https://github.com/NVIDIA-NeMo/RL/pull/3470/commits/97b309aaa6114bfae177796c0e90f3691af149d4)

asolergi-nv:
/ok to test [85afdb2](https://github.com/NVIDIA-NeMo/RL/pull/3470/commits/85afdb2bb89326b04ca402f57c0c1d6826ad0070)

asolergi-nv:
/ok to test [85afdb2](https://github.com/NVIDIA-NeMo/RL/pull/3470/commits/85afdb2bb89326b04ca402f57c0c1d6826ad0070)

asolergi-nv:
/ok to test [8bb7148](https://github.com/NVIDIA-NeMo/RL/pull/3470/commits/8bb7148f52a94326c960d062f34eac245012d3f7)

asolergi-nv:
/ok to test [9a3e424](https://github.com/NVIDIA-NeMo/RL/pull/3470/commits/9a3e424900bd8c7c63863aaad1acd86f54a7e169)

asolergi-nv:
/ok to test [933cf55](https://github.com/NVIDIA-NeMo/RL/pull/3470/commits/933cf550f2c85347f4689a17383e18b06e7cc70a)

nemo_rl/experience/failures.py
```diff
@@ -0,0 +1,232 @@
...
+    if isinstance(exc, RolloutDataFailure):
+        return False
+
+    # An HTTP error is classified by status, not by type. A 4xx from the generation
+    # engine describes the request -- "This model's maximum context length is ..." comes
+    # back as a 400 -- so retrying it on another shard would fail identically. 5xx and
+    # the retriable 4xx mean the endpoint is unwell, which another shard can serve.
+    status = _http_status(exc)
+    if status is not None:
+        return status >= 500 or status in _RETRIABLE_HTTP_STATUSES
```
terrykong: [`nemo_rl/experience/failures.py:197`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/failures.py#L197)

**2 action items — please fix in this PR.**

**TL;DR — every NeMo-Gym HTTP failure crosses the `NemoGym` Ray actor boundary, which destroys the exception's type and `.status`, so this status-based INFRA/DATA split never runs in production: a 503 from a sick vLLM endpoint classifies DATA and kills the run after 2 attempts under the default `on_data_exhausted: fail_fast` — the 5-attempt `max_attempts_per_prompt` re-dispatch never engages.**

PR-introduced (defect in PR-introduced code). Mechanism:
1. Gym raises `aiohttp.ClientResponseError` inside the `NemoGym` actor ([nemo_gym.py#L533](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/environments/nemo_gym.py#L533)).
2. Ray pickles the cause into `RayTaskError`. A real `ClientResponseError` carries `CIMultiDictProxy` headers (aiohttp's `raise_for_status` passes `headers=self.headers`), which cannot pickle — so Ray replaces the cause with a bare `RayError(str)`.
3. On the driver: `isinstance(e, ClientResponseError)` is `False`, `getattr(e, "status", None)` is `None`, `__cause__` is `None`. Verified empirically (live Ray 2.57 + aiohttp 3.14.3): 400 and 503 both arrive as bare `RayTaskError` → DATA.
4. The unit tests stay green because they construct `ClientResponseError(None, (), ...)` — headerless, hence picklable — and never cross a Ray boundary.

Two facts make this the *dominant* gym failure path, not a corner case: Gym's exception middleware converts inner-server errors to **500** ([server_utils.py#L562-L571](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/nemo_gym/server_utils.py#L562-L571)) — exactly the status this branch would call INFRA — and Gym's transport layer retries `ServerDisconnectedError`/`ClientOSError` in an uncapped `while True` ([server_utils.py#L219-L245](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/nemo_gym/server_utils.py#L219-L245)), so those never reach NeMo-RL at all. Net: out of the box, `max_attempts_per_prompt` effectively cannot apply on the gym path — the path whose dead-endpoint scenario motivates this PR. To be clear, this is an unmet promise rather than a regression: main was 1-attempt-fatal here, this PR makes it 2.

The same boundary affects the native path: `RayTaskError` itself is not in `_INFRA_TYPES` (and `issubclass(RayTaskError, RayActorError)` is `False`), so an exception raised inside a still-*alive* worker — vLLM `EngineDeadError`, CUDA OOM, both `RuntimeError`s — arrives as `RayTaskError(RuntimeError)` → DATA → 2 attempts → `fail_fast`, bypassing the re-dispatch machinery built for exactly that. The chaos test kills the worker *process* (→ `RayActorError` → INFRA), so it only exercises the covered branch.

### AI-1
Classify on the raising side, where the status is still in hand — in `NemoGym.run_rollouts`'s except block ([nemo_gym.py#L533](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/environments/nemo_gym.py#L533)), raise a picklable typed failure before the Ray boundary. One way to do it:
```python
status = getattr(error, "status", None)
if status is not None:
    detail = f"NeMo-Gym /run failed with HTTP {status}: {error}"
    if status >= 500 or status in (408, 429):
        raise GymTransportError(detail) from None
    raise RolloutDataFailure(detail) from None
raise
```
(`GymTransportError`/`RolloutDataFailure` are plain `Exception` subclasses with a str arg, so they pickle cleanly and `classify_rollout_failure`'s explicit-class fast path wins.)

### AI-2
Harden `_INFRA_TYPES` ([failures.py#L136](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/failures.py#L136)) for the native vector: add `ray.exceptions.ObjectLostError` (the parent of the already-listed `OwnerDiedError`/`ObjectFetchTimedOutError`), `ray.exceptions.OutOfMemoryError`, `ObjectStoreFullError`, `RaySystemError` — and consider treating a `RayTaskError` whose cause was degraded as INFRA rather than DATA (the infra budget is bounded and backed off, so a genuine bug still surfaces after 5 attempts, whereas a genuine infra event under DATA is unrecoverable).

A test that would have caught this: round-trip the exception the way Ray does (`cloudpickle.dumps` on a `ClientResponseError` carrying `CIMultiDictProxy` headers) before classifying, or assert whatever `NemoGym.run_rollouts` raises is pickle-round-trippable and still classifies INFRA.

nemo_rl/algorithms/single_controller.py
```diff
@@ -580,6 +606,79 @@ async def _train_pump(self) -> None:
...
+
+        Environments without the method are skipped rather than treated as unhealthy;
+        only NeMo-Gym has subprocess servers to lose.
+        """
+        for env_name, handle in self._env_handles.items():
+            health_check = getattr(handle, "health_check", None)
+            if health_check is None:
+                continue
+            try:
+                await self._ray_get(health_check.remote())
```
terrykong: [`nemo_rl/algorithms/single_controller.py:676`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/algorithms/single_controller.py#L676)

**2 action items, 1 follow-up.**

PR-introduced. Two ways `_check_env_health` escapes the watchdog's own safety contract:

1. A health failure raises `RuntimeError` out of `_watchdog_pump`, which `run()` awaits — so the run dies even under the default `stall_action: "warn"`, whose doc says it decides whether the watchdog "only reports, or ends the run". With `gym_subprocess_check` defaulting to `true`, this is a new run-killing path enabled by default, in a PR whose stated posture is inert-by-default.
2. `await self._ray_get(health_check.remote())` has no deadline. `NemoGym` is an asyncio actor, so if it is wedged — the motivating case for this watchdog — the await never returns, the pump never reaches the stall check again, and stall detection is dead exactly when it is needed.

### AI-1
Route health-check failures through `stall_action` (warn → print, abort → raise), matching the stall path.

### AI-2
Wrap the health RPC in `asyncio.wait_for(..., timeout=<interval_s>)` and treat a timeout as unhealthy (reported per AI-1) rather than blocking the pump.

### Follow-up
`gym_subprocess_check` gates a generic hook — `_check_env_health` iterates *every* env handle exposing `health_check`; only NeMo-Gym has one today. `env_health_check` would be the honest name once another env grows one (fine to fold into this PR or defer).

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -35,11 +42,97 @@
...
+
+
 class AsyncRLConfig(BaseModel, extra="allow"):
     # Staleness policy shared by the rollout and train pumps.
     sampler: SamplerConfig = Field(
         default_factory=InOrderSamplerConfig,
     )
+    # Deadline for one NeMo-Gym prompt-group rollout, covering the whole streaming
+    # response. None disables.
+    rollout_timeout_s: Optional[PositiveFloat] = None
```
terrykong: [`nemo_rl/algorithms/single_controller_utils/config.py:125`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/algorithms/single_controller_utils/config.py#L125)

**1 action item.**

PR-introduced. The fault-tolerance keys are split across two levels with mixed scoping, and the path-specific ones are silent no-ops when set for the wrong path: `rollout_timeout_s` is read only by `AsyncNemoGymRolloutImpl` ([rollout_manager.py#L815](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L815)); `generation_timeout_s`/`env_timeout_s` only by `AsyncRolloutImpl` ([#L504](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L504), [#L404](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L404)); `rollout_failure.max_gym_row_attempts` only by the gym impl — while the rest of `rollout_failure` governs both paths via the shared retry loop in `generate_and_push`. A native-path operator who sets `rollout_timeout_s: 900` (the most generic-sounding key) gets no deadline and no warning; the YAML comments say all this, but nothing enforces or structures it.

A second cost of the current shape: the three timeouts dangle directly on `async_rl` among unrelated pump/buffer knobs (`max_inflight_prompts`, `min_groups_for_streaming_train`, ...), even though they are fault-tolerance controls — while `rollout_failure` and `watchdog` are already blocks. Gathering the whole FT surface into one block fixes both problems at once.

**Action:** collect the fault-tolerance config into one block — shared budgets at the top level, path-specific knobs in named sub-blocks — so the YAML structure itself carries the scoping and nothing FT-related dangles at the root (one way to shape it, following the repo's nested-BaseModel convention; naming the block `fault_tolerance` instead of `rollout_failure` also works):

```python
class NativeRolloutFTConfig(BaseModel, extra="allow"):
    # Deadline for a single generate_async turn. None disables.
    generation_timeout_s: Optional[PositiveFloat] = None
    # Deadline for one environment step. None disables.
    env_timeout_s: Optional[PositiveFloat] = None

class NemoGymRolloutFTConfig(BaseModel, extra="allow"):
    # Deadline for one whole prompt-group rollout, re-dispatches included. None disables.
    rollout_timeout_s: Optional[PositiveFloat] = None
    # Re-dispatch just the rows that never arrived before retrying the whole group.
    max_row_attempts: PositiveInt = 3

class RolloutFailureConfig(BaseModel, extra="allow"):
    # ── shared: consumed by generate_and_push, which sits above the impl split ──
    max_infra_attempts_per_prompt: PositiveInt = 5
    max_data_attempts_per_prompt: PositiveInt = 2
    backoff_base_s: PositiveFloat = 1.0
    max_backoff_s: PositiveFloat = 30.0
    # 0 = fail fast on first data-budget exhaustion; N > 0 = skip up to N prompts.
    # (Collapses on_data_exhausted + max_skipped_prompts -- see the thread on L66.)
    max_skipped_prompts: NonNegativeInt = 0
    # ── path-specific ──
    native: NativeRolloutFTConfig = Field(default_factory=NativeRolloutFTConfig)
    nemo_gym: NemoGymRolloutFTConfig = Field(default_factory=NemoGymRolloutFTConfig)
```

```yaml
async_rl:
  rollout_failure:                     # shared budgets — both rollout paths
    max_infra_attempts_per_prompt: 5
    max_data_attempts_per_prompt: 2
    backoff_base_s: 1.0
    max_backoff_s: 30.0
    max_skipped_prompts: 0             # 0 = fail fast; N > 0 = skip up to N prompts
    native:                            # AsyncRolloutImpl only
      generation_timeout_s: null
      env_timeout_s: null
    nemo_gym:                          # AsyncNemoGymRolloutImpl only
      rollout_timeout_s: null
      max_row_attempts: 3
```

Two things to carry along with the move: (1) `_check_watchdog_outlasts_rollouts` currently compares `stall_timeout_s` only against the gym-only `rollout_timeout_s`, so the invariant it advertises is unchecked against `generation_timeout_s`/`env_timeout_s` on the native path — re-point it at the new locations and cover the native pair; (2) nesting fixes readability but a populated wrong-path sub-block is still a silent no-op, so a cheap setup-time check in `validate_single_controller_config` (reject/warn when the inapplicable sub-block carries non-default values — same idiom as the `val_temperature` guard in [grpo.py#L411-L432](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/algorithms/grpo.py#L411-L432)) closes the loop.

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -35,11 +42,97 @@
...
+
+    Infrastructure failures re-dispatch the prompt onto a different generation shard;
+    data failures are deterministic, so their budget is small and exhausting it is
+    reported rather than absorbed. Nothing here ever discards a prompt silently.
+    """
+
+    # Attempts for infrastructure failures (timeout, dead shard, transport). Each retry
+    # re-enters shard selection, so it lands elsewhere. Exhausting this means the fleet
+    # is broken rather than the prompt, and the run fails.
+    max_attempts_per_prompt: PositiveInt = 5
```
terrykong: [`nemo_rl/algorithms/single_controller_utils/config.py:56`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/algorithms/single_controller_utils/config.py#L56)

**1 action item.**

PR-introduced. `max_attempts_per_prompt` beside `max_data_attempts_per_prompt` reads as "total budget" + "sub-budget", but the two counters are disjoint: `infra_attempts` gates the loop ([rollout_manager.py#L1095](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L1095)) while `data_attempts` is tracked independently ([#L1128](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L1128)) — under the shipped defaults (5, 2) one prompt can consume **6** attempts. The internal dataclass already has the honest name, `max_infra_attempts`; only `_build_retry_policy` records the equivalence, so grepping this key never reaches the retry loop.

**Action:** rename to `max_infra_attempts_per_prompt` (update the exemplar YAML), and add one docstring line: the two budgets are independent — worst case is their sum minus one.

nemo_rl/experience/rollout_manager.py
```diff
@@ -511,36 +730,121 @@ def _build_inputs(self, input_sample: DatumSpec) -> list[dict]:
...
+                            or attempt == self._max_gym_row_attempts
+                        ):
+                            raise
+                    else:
+                        if timing_metrics is not None:
+                            env_timing_metrics = timing_metrics
+
+            missing = [i for i, result in enumerate(results) if result is None]
+            if missing:
+                raise GymTransportError(
```
terrykong: [`nemo_rl/experience/rollout_manager.py:844`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L844)

**2 action items.**

PR-introduced. When an intermediate re-dispatch attempt absorbs an INFRA error and a later attempt ends the stream cleanly-but-short, this raise carries no `from` — the actual transport error (the reason the rows are missing) is lost at exactly the point the operator reads it. Separately, row-level re-dispatches never call `record_redispatch`, so `rollout/redispatch_total` — "the only externally visible sign that the fleet is degrading" per `RolloutStats`' own docstring — stays flat while gym silently retries rows all run.

### AI-1
Keep `last_error` across attempts and `raise GymTransportError(...) from last_error`.

### AI-2
Count row re-dispatches (thread the stats object or a callback into `AsyncNemoGymRolloutImpl`) under their own reason key.

nemo_rl/algorithms/single_controller.py
```diff
@@ -120,6 +125,14 @@ def __init__(
...
         self._buffer = actor_args.tq_buffer
         self._rollout_manager = actor_args.rollout_manager
+        # getattr, not attribute access: `env_handles` is a real field of
+        # SingleControllerActorArgs, so every production caller has it -- but upstream's
+        # unit tests build actor_args as a SimpleNamespace listing only what upstream's
+        # own __init__ reads, and upstream never reads this one. Requiring it here turns
+        # our feature into a construction requirement for tests that do not exercise it,
+        # which is how test_logs_setup_timing_metrics broke on the main sync. Same shape
+        # as the `timeouts` regression in the previous sync.
+        self._env_handles = getattr(actor_args, "env_handles", {})
```
terrykong: [`nemo_rl/algorithms/single_controller.py:135`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/algorithms/single_controller.py#L135)

**1 action item.**

PR-introduced. `env_handles` is a required, non-defaulted field of `SingleControllerActorArgs` ([setup.py#L88](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/algorithms/single_controller_utils/setup.py#L88)) and every production caller passes it; this `getattr` default exists for test fixtures. The cost is a silent failure mode: if the field is ever renamed or dropped, `watchdog.gym_subprocess_check: true` degrades to a no-op health check with no error — the class of silent failure this PR removes elsewhere.

**Action:** use `actor_args.env_handles` directly and add `env_handles={}` to the one `SimpleNamespace` fixture that omits it (the other fixtures in `tests/unit/single_controller/` already pass it).

tests/functional/grpo_dp_single_controller_chaos.sh
```diff
@@ -0,0 +1,369 @@
...
+#
+# Usage: bash tests/functional/grpo_dp_single_controller_chaos.sh
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
+PROJECT_ROOT=$(realpath "$SCRIPT_DIR"/../..)
+git config --global --add safe.directory "$PROJECT_ROOT"
+
+set -eou pipefail
+
+EXP_NAME=$(basename "$0" .sh)
```
terrykong: [`tests/functional/grpo_dp_single_controller_chaos.sh:26`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/functional/grpo_dp_single_controller_chaos.sh#L26)

**1 action item.**

PR-introduced. Both SingleController-lane entries run this script ([L1_Functional_Tests_SingleController.sh#L44](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/functional/L1_Functional_Tests_SingleController.sh#L44) idle, [#L55](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/functional/L1_Functional_Tests_SingleController.sh#L55) serving) with the same `EXP_NAME`, and the script begins with `rm -rf "$EXP_DIR"` ([#L141](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/functional/grpo_dp_single_controller_chaos.sh#L141)) — so the serving run deletes the idle run's `run.log` and tensorboard artifacts. When the serving variant fails in CI, the idle evidence you'd want to compare against is gone. `VICTIM_STATE` is only assigned below at [#L71](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/functional/grpo_dp_single_controller_chaos.sh#L71), hence the `:-idle` default:

```suggestion
EXP_NAME=$(basename "$0" .sh)-${VICTIM_STATE:-idle}
```

nemo_rl/experience/rollout_manager.py
```diff
@@ -511,36 +730,121 @@ def _build_inputs(self, input_sample: DatumSpec) -> list[dict]:
...
+        ).remote(pending, self._tokenizer, timer_prefix):
+            rowidx, result, timing_metrics = await result_ref
+            # Validated against the original group, not the pending subset: on a
+            # re-dispatch the row keeps its original index so results stay ordered.
+            if not isinstance(rowidx, int) or not 0 <= rowidx < total_rows:
+                raise ValueError(
+                    f"NeMo-Gym returned invalid row index {rowidx!r} for "
+                    f"{total_rows} inputs"
+                )
+            if rowidx not in dispatched:
```
terrykong, line 846: [`nemo_rl/experience/rollout_manager.py:768`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L768)

**1 action item.**

PR-introduced test gap. Of the three row-index guards in `_stream_rows`, only the duplicate guard is pinned by a test (`match="duplicate row index 0"` in [test_rollouts.py#L1482](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/unit/experience/test_rollouts.py#L1482)); this "not dispatched in this attempt" guard — new in this PR, and the thing that stops a stale echo of an already-landed row from silently overwriting a good `results[rowidx]` during re-dispatch — has none. Delete these two lines and the suite stays green.

**Action:** add a case next to the partial-re-dispatch tests: attempt 1 lands row 0 then dies; attempt 2's stream echoes row 0 again → expect `ValueError` matching `"which was not dispatched"`.

nemo_rl/experience/rollout_manager.py
```diff
@@ -728,31 +1053,117 @@ async def run_rollout(self, input_sample: DatumSpec) -> PromptGroupRecord:
...
+                            "dataset or sequence-length configuration is likely wrong"
+                        ) from error
+                    print(
+                        f"skipping prompt idx={input_sample['idx']} after "
+                        f"{data_attempts} deterministic failure(s) ({reason}: {error})",
+                        flush=True,
+                    )
+                    self._stats.skipped += 1
+                    return RolloutOutcome.SKIPPED
+                self._stats.record_redispatch(reason)
```
terrykong: [`nemo_rl/experience/rollout_manager.py:1147`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L1147)

**1 action item.**

PR-introduced. Data-failure retries are recorded into the same `redispatches_by_reason` namespace as infra re-dispatches, so `rollout/redispatch_total` conflates "fleet degrading" with "prompts are bad" — the very distinction the two-budget design exists to make.

**Action:** record data retries under their own key (e.g. `rollout/data_retries_total/<reason>`).

tests/functional/grpo_dp_single_controller_chaos.sh
```diff
@@ -0,0 +1,369 @@
...
+# This is the scenario the whole SC resiliency effort exists for. Before the P0
+# work, a dead generation endpoint left rollouts parked forever -- each holding a
+# max_inflight_prompts permit -- until the rollout pump blocked and the train pump
+# spun, with no exception raised anywhere. The pass condition here is therefore not
+# "training succeeds"; it is "the job stops, quickly, with an attributable error".
+#
+# Registered in the SingleController L1 lane (full mode). It was originally kept out as
+# "timing-sensitive", but that no longer justifies exclusion: the death deadline is now
+# 600s against an observed 222s, the victim selection is asserted rather than assumed, and
+# the recovery tests in the same lane kill processes the same way. Leaving it out meant
```
terrykong: [`tests/functional/grpo_dp_single_controller_chaos.sh:14`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/functional/grpo_dp_single_controller_chaos.sh#L14)

**1 action item.**

PR-introduced. This line, [#L79](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/functional/grpo_dp_single_controller_chaos.sh#L79) and [#L240](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/tests/functional/grpo_dp_single_controller_chaos.sh#L240) justify decisions by pointing at "the recovery test(s)", which don't exist in this branch — the lane contains only `grpo_dp_single_controller.sh` and `grpo_async_gym_single_controller.sh`, neither of which kills processes. Presumably a forward reference to the next PR in the #3454 stack.

**Action:** name the target explicitly ("the recovery harness added in <next PR>") or inline the one-sentence reason at each site.

nemo_rl/experience/rollout_manager.py
```diff
@@ -44,6 +56,194 @@
...
+class RolloutRetryPolicy:
+    """Retry budgets for one prompt, resolved from ``async_rl.rollout_failure``.
+
+    The defaults here reproduce the historical single-attempt behaviour so that a
+    directly-constructed ``RolloutManager`` does not silently gain retries;
+    ``setup_single_controller`` passes the configured values in.
+    """
+
+    # Attempts for infrastructure failures. 1 means no retry.
+    max_infra_attempts: int = 1
```
terrykong: [`nemo_rl/experience/rollout_manager.py:79`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L79)

**1 action item** (low severity).

PR-introduced. This dataclass carries a full second set of defaults (1/1/1) that contradict `RolloutFailureConfig`'s (5/2/3). The intent — a directly-constructed `RolloutManager` must not silently gain retries — is documented and sound, but defaults now live in two places, and a reader here walks away believing the default budget is 1.

**Action:** make these fields required and have the `None`-fallbacks construct the explicit single-attempt policy (`RolloutRetryPolicy(max_infra_attempts=1, max_data_attempts=1, max_gym_row_attempts=1)`), so the only *defaults* in the codebase are the BaseModel's.

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -35,11 +42,97 @@
...
+    # Attempts for deterministic, prompt-specific failures. One retry separates a
+    # transient empty response from a genuinely bad prompt; a second identical failure
+    # confirms the prompt is at fault.
+    max_data_attempts_per_prompt: PositiveInt = 2
+    # First infra-retry delay, doubled per attempt.
+    backoff_base_s: PositiveFloat = 1.0
+    # Ceiling on the exponential backoff, so a long outage retries at a steady rate.
+    max_backoff_s: PositiveFloat = 30.0
+    # After max_data_attempts_per_prompt: fail the run, or continue without the prompt.
+    on_data_exhausted: Literal["fail_fast", "skip"] = "fail_fast"
```
terrykong: [`nemo_rl/algorithms/single_controller_utils/config.py:66`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/algorithms/single_controller_utils/config.py#L66)

**1 action item.**

This enum/int pair encodes exactly two meaningful states, and the `_check_consistent` branch below exists only to reject their one degenerate combination (`skip` + `0`). A single knob makes the illegal state unrepresentable:

- `max_skipped_prompts: 0` (default) → fail fast on the first data-budget exhaustion (re-raise the original error — current `fail_fast`), and
- `max_skipped_prompts: N > 0` → skip up to N prompts, then fail (current `skip`).

Nothing is lost: `fail_fast` with a positive skip count is meaningless today, and `skip` already requires N ≥ 1. The name self-documents ("allow at most 0 skipped prompts" reads as fail-on-first).

**Action:** drop `on_data_exhausted`, keep only `max_skipped_prompts`. This also deletes the `skip`+`0` validator branch, its test, and `RolloutRetryPolicy.skip_on_data_exhausted` — the retry loop in [`generate_and_push`](https://github.com/NVIDIA-NeMo/RL/blob/933cf550f2c85347f4689a17383e18b06e7cc70a/nemo_rl/experience/rollout_manager.py#L1129-L1146) just compares against the budget (at 0 the first exhaustion re-raises, which is exactly today's `fail_fast`).

terrykong (commented):
Team review — 5 agents (RL expert, bug finder, test reviewer, design reviewer, devil's advocate); every finding below survived an adversarial verification pass (4 candidate findings were dropped by it).

This is a well-built PR solving a real, previously unowned problem — the failure taxonomy's module-function shape, the pydantic→dataclass layering (with its dependency-direction rationale), the watchdog pump's mirroring of the sibling pumps, and the chaos harness's pinned victim selection are all genuinely good work, and the unit suite is unusually thorough (it even cross-checks Gym's own retry-status list against the classifier).

**Two items we'd ask to address in this PR** (both inline): the Ray-boundary classification defect — as written, the status-based INFRA/DATA split and the `max_attempts_per_prompt` re-dispatch budget cannot engage on the NeMo-Gym path, which is where the motivating dead-endpoint failure lives — and the env-health check's bypass of `stall_action` plus its unbounded await. Everything else is a suggestion or follow-up.

On the two pre-flagged reviewer questions: (1) the attempt budgets in `rollout_failure` apply to **both** rollout paths (the retry loop sits above the impl split); the genuinely path-specific, silently-ignored keys are the `async_rl` timeout trio — see the inline comment proposing a setup-time guard. (2) Neither "prompt longer than max_model_len" nor "generated to the token limit" can kill a run: the native path returns a zero-generation / truncated result without raising, and Gym absorbs the context-length 400 into an empty `finish_reason="length"` completion before it ever reaches NeMo-RL — verified on both paths.

Linters: all pre-commit hooks pass locally (pyrefly not runnable on this host; CI covers it). `failures.py` is correctly registered in `pyrefly.toml`.

_Generated by Claude Code_

terrykong:
@asolergi-nv I've mirrored this branch (and parts 2–4) from your fork into `NVIDIA-NeMo/RL` so the series can be linked as a real PR stack — `gh stack` can't operate on fork branches, and with all four based on `main` each PR showed the cumulative diff rather than its own layer.

**This PR now continues as #3589** (stack #3593):

| Part | Mirror PR | Original | Diff vs. its base |
|---|---|---|---|
| 1/4 contain rollout failures | #3589 | #3470 | +3446 −111 |
| 2/4 fleet health + Gym router | #3590 | #3471 | +2186 −54 |
| 3/4 keep training on shard death | #3591 | #3472 | +3437 −59 |
| 4/4 restart + re-admission | #3592 | #3473 | +1153 −65 |

Branches were pushed unchanged — head SHA here is identical (`933cf550f2c85347f4689a17383e18b06e7cc70a`). Leaving this open for history and the existing review threads; **please move new work and discussion to #3589**. Push to the upstream branch `feat/sc-resiliency-01-containment` (you should have access — let me know if not).

nemo_rl/experience/failures.py
asolergi-nv: Confirmed and fixed. Reproduced across a live Ray actor:

```
503, before   RayTaskError                      -> DATA
503, after    RayTaskError(GymTransportError)   -> INFRA
400, after    RayTaskError(RolloutDataFailure)  -> DATA
```

**AI-1** (29299d009): `NemoGym.run_rollouts` now classifies at the source and raises a picklable `GymTransportError`/`RolloutDataFailure` before the boundary. `http_status_is_infra` is public so both sides share one rule, with a test asserting they agree. `TestTheRayActorBoundary` builds the realistic headers-carrying error, asserts it *cannot* be pickled, and checks that what we raise survives and still classifies.

**AI-2, partly**: added `ObjectLostError`, `OutOfMemoryError`, `ObjectStoreFullError`, `RaySystemError` — all confirmed uncovered.

**Not taken**: treating any degraded `RayTaskError` as INFRA inside `classify_rollout_failure`. That function is also consulted from the main retry loop and the gym row-retry loop, so it would reach env-step and gym-postprocessing failures — a reward-fn bug or `Generation logprobs contain NaN` would get up to 15 dispatches before surfacing wrapped, and would inflate `redispatch_total`, the conflation the :1147 thread asks us to remove.

Scoped it to the generation call site instead (c000b31fe): a Ray-boundary error from a *generation* RPC is INFRA, since we know the RPC's purpose there. That also closed a real inconsistency — 2/4's `_generate_on_shard` already maps `RayError` to `GenerationUnavailable`, so the same exception was INFRA there and DATA here. Happy to widen it if you'd rather, though it's probably worth deciding together with 2/4, which makes the same trade implicitly.

nemo_rl/algorithms/single_controller.py
asolergi-nv: Both fixed in ac6565ef9.

**AI-1**: `_check_env_health` returns the problems it found; `_watchdog_pump` routes them through `stall_action`.
**AI-2**: each probe is wrapped in `asyncio.wait_for(..., interval_s)`, and a timeout counts as unhealthy.

The second was the sharper catch — the watchdog could be silently disabled by exactly the wedge it exists to detect. Tests cover abort-raises, warn-only-warns, and a health check that never returns not stopping the pump.

On the rename: agreed `env_health_check` is the honest name. Not done yet — happy to fold in here or take it as a follow-up.

nemo_rl/algorithms/single_controller_utils/config.py
asolergi-nv: Done in 1590e2403, close to your sketch: shared budgets at the top of `rollout_failure`, path-specific knobs under `native`/`nemo_gym` (`max_gym_row_attempts` → `nemo_gym.max_row_attempts`).

Both carry-alongs:
- `_check_watchdog_outlasts_rollouts` now checks all three deadlines rather than only the gym one — that was a genuine gap on the native path; tests parametrize over all three.
- `validate_single_controller_config` rejects a populated wrong-path sub-block. It immediately caught the chaos harness setting the gym-only `rollout_timeout_s` on a native run, where it had never done anything.

Added one thing beyond the sketch: since every key here is renamed or moved under `extra="allow"`, the old spellings would parse and quietly stop taking effect — for `on_data_exhausted: skip` that is a behaviour change with no diagnostic. Both levels now reject the old names and name the replacement.

nemo_rl/algorithms/single_controller_utils/config.py
asolergi-nv: Agreed — done in 1590e2403. `on_data_exhausted` is gone; `max_skipped_prompts: 0` (the default) fails on the first exhaustion and re-raises the original error, exactly as `fail_fast` did. The `skip`+`0` validator branch, its test, and `RolloutRetryPolicy.skip_on_data_exhausted` are all removed.

nemo_rl/algorithms/single_controller_utils/config.py
asolergi-nv: Renamed to `max_infra_attempts_per_prompt` in 1590e2403, with the independence of the two budgets stated where the defaults live (worst case is their sum minus one). Exemplar YAML updated, and the old key now raises rather than being silently accepted.

86582577c cleans up what the rename left behind — including the `RolloutRedispatchExhausted` message, which was telling operators to look for a key that no longer exists.

nemo_rl/experience/rollout_manager.py
asolergi-nv: Done in e3d043039 — data retries record under `rollout/data_retry_total/<reason>`.

Also gave every family an aggregate while in there: data failures previously had only per-exception series, so alerting on "any data failure" required knowing the exception names in advance.

nemo_rl/experience/rollout_manager.py
asolergi-nv: Both in e3d043039.

**AI-1**: `last_error` is kept across attempts and chained (`raise GymTransportError(...) from last_error`), with a test where attempt 1 dies on transport and attempt 2 ends cleanly-but-short.
**AI-2**: row re-dispatches now increment `rollout/gym_row_redispatch_total`, which meant threading the owning manager's stats object into `AsyncNemoGymRolloutImpl`.

nemo_rl/experience/rollout_manager.py
asolergi-nv: Added in e3d043039: `test_a_stale_echo_of_a_landed_row_is_rejected` — attempt 1 lands row 0 then dies, attempt 2 echoes row 0.

Checked it the way you framed it: with the two lines deleted the new test fails, and with them restored it passes.

nemo_rl/experience/rollout_manager.py
asolergi-nv: Done in e3d043039. The three attempt budgets are now required, and `RolloutRetryPolicy.single_attempt(**overrides)` is the explicit no-retry construction used by the direct-construction fallbacks. `RolloutFailureConfig` is the only place a retry default lives now.

Used a named constructor rather than making callers spell out all three, so tests don't have to state budgets irrelevant to what they're testing.

tests/functional/grpo_dp_single_controller_chaos.sh
asolergi-nv: Applied in bc16af89c. `submit-ci.sh` globs `tests/functional/*/run.log`, so both directories are collected without any change there.

tests/functional/grpo_dp_single_controller_chaos.sh
asolergi-nv: Fixed in 19a9506cb — two are inlined with the actual reason, and the third names the part of the series that adds the harness.

nemo_rl/algorithms/single_controller.py
asolergi-nv: Done in 61fad3d06 — direct `actor_args.env_handles`, with `env_handles={}` added to the one fixture that omitted it.

Agreed on the reasoning: the getattr default degrades `gym_subprocess_check: true` into a health check that iterates nothing and reports nothing, which is the failure class this PR exists to remove. `_fleet_monitor`/`_policy_router` in 2/4 keep the idiom for a different reason — `None` there is a meaningful value meaning the feature is off.

asolergi-nv:
Thanks — this was a genuinely useful review. All twelve are addressed; the work is pushed to **#3589**, which is where the series continues.

| # | thread | commit |
|---|---|---|
| 6 | Ray-boundary classification | 29299d009, c000b31fe |
| 2 | env health check | ac6565ef9 |
| 12 | chaos `EXP_NAME` collision | bc16af89c |
| 3, 4, 5 | fault-tolerance config | 1590e2403, 86582577c |
| 7, 9, 10 | retry counters and policy | e3d043039 |
| 8 | stale-echo re-dispatch test | e3d043039 |
| 1 | `env_handles` | 61fad3d06 |
| 11 | dangling forward references | 19a9506cb |

Two things worth surfacing rather than leaving in the threads.

**One divergence.** On the classification finding I took the source-side fix and the explicit `_INFRA_TYPES` additions, but scoped the `RayTaskError` widening to the generation call site rather than putting it in `classify_rollout_failure` — reasoning in that thread. It's the one place I didn't follow the suggested action, and it's worth deciding alongside 2/4, whose `_generate_on_shard` already makes the same trade.

**A correction to something I had wrong.** Your answer to my second question stands, and it fixes a mistake of mine: I had measured the native path (over-long prompt → `generation_lengths=[0]`, `truncated=False`, no exception) but claimed the gym path failed loudly with a 400 → DATA → fail-fast. It does not. The Ray-boundary finding is exactly why my check looked convincing — I had built a headerless `ClientResponseError`, which production never produces, so I was testing the same fiction the unit tests were.

The related gap that remains, and is orthogonal to this PR: an over-long prompt on the native path produces an empty rollout marked `truncated=False`, so it is indistinguishable from a model that chose to emit nothing. Measured on GB200. Worth a follow-up to raise `RolloutDataFailure` when `allowed_new_tokens == 0`.

PR #3471, opened 2026-08-03
feat(sc): track generation fleet health and route NeMo-Gym through NeMo-RL

asolergi-nv:
Part 2/4 of #3454

# What this fixes

Containment (Part 1/4, #3470) stops a dead engine from wedging or corrupting a run, but the fleet still has no idea which engines are alive, so both rollout paths keep sending work to a dead one:

- **GRPO** picks a generation shard by round-robin over the full worker list.
- **NeMo-Gym** picks a policy endpoint by static round-robin over a list fixed at process start, with no health input and **no failover** — `_resolve_client` never re-resolves. A dead endpoint keeps receiving roughly 1/N of new rollouts for the rest of the run.

# What it does

**A fleet health model** (`fleet_health.py`) — a pure state machine with I/O injected, so it is testable without Ray or GPUs:

```
HEALTHY ⇄ SUSPECT → DEAD → RESTARTING → STALE → HEALTHY
                              ↓
                           RETIRED
```

`SUSPECT` exists so a single failed probe does not cost a shard's throughput. `DEAD → HEALTHY` is deliberately unreachable: only a completed refit returns a shard to service, because an engine that answers `/health` says nothing about whether its weights are current.

**GRPO shard selection** now picks among healthy shards (least-outstanding), and generation failures are reported back to the monitor — the routing adapters see failures a liveness probe cannot, such as a shard that answers `/health` and errors on every generation.

**A NeMo-RL-owned router for NeMo-Gym.** Rather than change Gym, hand it a single URL that we own. Gym's `base_url` takes one string, so its round-robin becomes a no-op and the routing decision moves next to the health data. **Zero NeMo-Gym changes.**

Three decisions:

  - **The router's URL never changes.** The port is reserved once and passed in, so Ray recreating a restarted actor rebinds the same address. This matters because Gym never re-resolves: if the router allocated a fresh free port on restart — the way everything else in this codebase allocates ports — Gym would hold a dead URL forever. There is a test pinning it.

  - **All router state is built in `__init__`, including the server thread.** The deliberate inverse of the `NemoGym._spinup` shape, where the servers are started from a separate method Ray never re-runs on restart — leaving a live actor that cannot serve.

  - **The "no healthy backend" status is load-bearing and validated at config load.** Gym retries 429/500/502/503/504/520 and *raises its own retry ceiling* on the rate-limit subset, so answering with a 503 would spin forever and recreate the exact hang this removes. The default is 409, and the config rejects anything in Gym's retry set rather than leaving it to be discovered in production.

# Tests

- Fleet-health state machine tests, including that `DEAD → HEALTHY` is unreachable and that DEAD/RESTARTING/STALE ignore successful probes.
- 19 router tests run against **real aiohttp servers**, not fakes — a proxy is precisely the component fakes flatter, since header handling, streaming and status propagation only misbehave over a real socket. Covers all four endpoints Gym calls, including `/tokenize`, which is *not* under `/v1` because Gym's `create_tokenize` strips the suffix; plus a 512 KB streamed response and status propagation.
- 8 config-validation tests, including that every status in Gym's retry set is rejected.
- **End-to-end**: `grpo_async_gym_single_controller.sh` with the router enabled. Gym's only endpoint was the router (`:6081`) while vLLM served on `:3038`, and the run completed. The strongest signal is `gen_kl_error`, which compares vLLM's logprobs against the trainer's recomputation — a proxy corrupting or truncating a response blows it up. It did not move. Now **registered in `L1_Functional_Tests_SingleController.sh`** (full mode): the router previously had no functional coverage at all, so a regression in the proxy fronting every Gym rollout would have shipped silently.
- **The chaos test shows the health layer earning its keep.** Same scenario and box as PR 1: kill the only generation shard mid-run. PR 1 fails in **222s**; this PR fails in **62s**, and the log says why:

  ```
  fleet: shard 0 healthy -> suspect
  fleet: shard 0 suspect -> dead
  GenerationFleetExhausted
  ```
   #3470 has no monitor, so its only backstop is the stall detector: it reports `RolloutStall` once the rollout has been quiet long enough — true, but that names the symptom, and the timeout has to be generous enough not to fire on a slow-but-healthy rollout. The monitor here probes the shard directly and trips the `min_healthy_shards` floor, reporting `GenerationFleetExhausted`, which names the cause. Both are bounded and attributable; this one is ~4x faster and more actionable.

  Both numbers are from runs with the victim state **pinned** to `idle`, so this is reproducible rather than an artifact of which state the kill happened to catch. Ray renames the process title during a call, and a worker killed mid-`generate_async` fails much faster by a different route; #3470 makes that a separate registered variant instead of a coin flip.
- Full suite on 2xA6000: 644 unit passed / 3 skipped, plus `grpo_dp_single_controller.sh`, both Gym variants, and the chaos test. Gym with defaults confirms inertness — `PolicyRouterConfig(enabled=False)` and zero router startups.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


terrykong:

**This PR now continues as #3590**, based on #3589 — so it shows only this layer's diff (+2186 −54, 17 files) instead of the cumulative +5616. Head SHA unchanged (`8fec9443c7ce32d657aced1f1ed64a4229a6d2c0`), upstream branch `feat/sc-resiliency-02-fleet-health-router`.

Leaving this open for history; please move new work and discussion to #3590.

PR #3472, opened 2026-08-03
feat(sc): keep training when a generation shard dies

asolergi-nv:
Part 3/4 of #3454 

# What this fixes

With #3470 and #3471 in place a dead engine no longer wedges rollouts and no longer receives traffic — but the run still dies, and it dies in the worst possible way.

The refit communicator (`model_update_group`) is built once at setup over every training and inference rank. A NCCL broadcast requires **every** rank in the communicator to take part, so when a generation rank is gone the next weight sync blocks forever *inside NCCL*: no exception, no progress, no CPU burn, and Ray still reporting every actor healthy. This is the failure the whole effort exists to remove.

It is also ordered so that the hang comes first: `sync_weights` does `ray.get(futures_train)` before the `ray.get(futures_inference)` that would surface the dead actor, so the trainer blocks before anything can report the death.

# What it does

**Reconcile before every refit, not on a death event.** `reconcile_communicator()` is called at the top of `_sync_weights` — the one point where the refit group is provably idle and every rank is synchronized, which matters because the operations that change membership are themselves collectives. Doing it every time is idempotent and converges after a missed or reordered health update, instead of needing replay.

**Rebuild over the survivors**, so training continues on what is left. Trainers are never excluded, so rank 0 stays a trainer and the broadcast root is stable.

**Both NCCL transports recover**, by different routes. `nccl_reshard` needs more than the plain broadcast: both communicator families rebuilt *and* the refit plan regenerated.

Four decisions:

  - **"Absent" is deliberately not the complement of "serving".** The obvious implementation reads `serving_shards()`, and it is a bug. `SUSPECT` (failing probes, not yet condemned) and `STALE` (reloaded, holding old weights) are both withheld from traffic while their processes are alive and join a refit normally. Using the serving set would abort a run on one failed probe — and would abort it precisely when a STALE shard is waiting to be refit, which is the recovery, not the failure. Hence `GenerationFleetMonitor.absent_shards()` over `{DEAD, RESTARTING, RETIRED}`.

  - **Rebuilding the communicator is only half a recovery.** Every refit *dispatch* goes through `run_all_workers_*`, which walks the whole worker group — so after a loss it kept calling the dead shard's Ray actor and the next refit failed with `RayActorError`. Both dispatches now address only surviving DP leaders. (This shipped broken in an intermediate commit here and is fixed within this PR; kept as separate commits because the fix is instructive.)

  - **`nccl_reshard` cannot simply be resized.** Its bulk path is a mesh-to-mesh redistribute, not a broadcast: `prepare_nccl_reshard_refit_info` derives each parameter's destination placements from `gen_world_size`. Reusing a plan built for the old fleet does not error — a stale mesh is still a valid mesh — it just has survivors writing the slices the dead shard owned and leaving their own unwritten. So the plan is regenerated, and `init_communicator` and the rebuild share one `_build(membership)` so that arithmetic exists once and every normal run exercises the rebuild path.

  - **Survivors are compacted to contiguous prefixes**, not left with a hole. The reshard destination mesh is `torch.arange(offset, offset + num_gpus)`, so a gap silently misaligns every parameter rather than erroring. The rank arithmetic lives in a pure `weight_sync/membership.py`, separate from Ray dispatch, because it is the part that must be exactly right and the part that cannot be exercised without ≥3 GPUs.

# Tests
- `StatelessProcessGroup.abort()` verified on 2×A6000: with a peer SIGKILLed mid-broadcast, a survivor blocked in the collective was released **0.15 s** after another thread called `abort()`. `abort()` and not `destroy()` — NCCL documents `destroy` as an intra-node collective every rank must call or it hangs, which is exactly what a dead rank cannot do.
- `tests/functional/grpo_sc_generation_shard_recovery.sh` is the gate, runs on both refit transports in the SingleController lane, and self-skips below 3 GPUs rather than passing vacuously.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


terrykong:

**This PR now continues as #3591**, based on #3590 — so it shows only this layer's diff (+3437 −59, 32 files) instead of the cumulative +9051. Head SHA unchanged (`b04fe19f845626398cf17fd03321f5932c881691`), upstream branch `feat/sc-resiliency-03-elastic-recovery`.

Leaving this open for history; please move new work and discussion to #3591.

PR #3473, opened 2026-08-03
feat(sc): restart dead generation shards and re-admit them at the next refit

asolergi-nv:
Part 4/4 of #3454 

# What this fixes

After #3472 a lost shard is dropped and the run continues on the rest — but the fleet only ever **shrinks**. A job that sheds a few transient failures over a long run ends permanently smaller, and nothing ever brings capacity back.

Two things were missing, and the second was a genuine defect:

**1. Re-admission was structurally impossible.** `reconcile_communicator` keyed off "is anything absent", so the moment a restarted shard stopped being absent the answer was `False`, no rebuild happened, and the shard stayed out of both the communicator and the refit dispatch for the rest of the run. Restarting an engine would have achieved nothing.

**2. Nothing promoted a restarted shard back into service.** `GenerationFleetMonitor.report_refit` — documented as "a completed refit is the only way back into the serving set" — **was never called anywhere in `nemo_rl`**. A restarted shard therefore reached `STALE`, rejoined the communicator, received weights, and then sat idle for the rest of the run. The fleet recovered on paper and not in throughput. Every supervisor unit test passed because each one called `report_refit` by hand; the tests now drive it through the controller, which is what makes the wiring observable.

# What it does

**Membership is compared against what was built**, not against "is anything absent", so a shard leaving and a shard returning are one operation in opposite directions. `desired_membership()` expresses it as "what it should be" rather than "what changed", which is what makes both directions fall out of one comparison. An unrecorded membership means the full fleet, not "unknown".

**`EngineSupervisor`** restarts dead shards, one background task each. Handover is entirely through fleet-health states:
```
DEAD ─(restart starts)→ RESTARTING ─(engine up)→ STALE ─(refit)→ HEALTHY
```
`RESTARTING` is absent from collectives, so a rebuild landing mid-restart correctly excludes it. `STALE` is present but not serving — that is what lets the refit write current weights before the shard takes traffic again.

**The restart is not awaited.** Reloading a model takes minutes, and the loop it runs from
also drives the rollout pump, the watchdog and the refit; blocking it would stall training
that the surviving shards can still do.

**`RayWorkerGroup.recreate_worker`** replays the recorded creation call for one worker rather than re-deriving it — re-deriving means reproducing the placement group, bundle, venv and rank bookkeeping, a second implementation that would drift. The old actor is killed first: usually already dead, but an unresponsive one still holding the GPU would make the replacement fail on memory rather than on anything informative.

# Testing

- 18 supervisor tests: state handover, the attempt cap and retirement, failed restarts not propagating, and that `tick()` returns while the restart is still blocked.
- 8 `recreate_worker` tests pinning the replay — same bundle and placement group, old actor killed first, name uniquified per incarnation, and the recorded spec left unmutated so names do not compound as `-r1-r2-r3`.
- `RESTART_DEAD_SHARDS=true` mode added to the recovery functional test and registered in the SingleController lane. It asserts a restart was attempted, an engine came back, **two** rebuilds happened (one dropping the shard, one taking it back), and a `stale -> healthy` transition occurred. Completion plus one rebuild would only prove the fleet shrank and carried on.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


terrykong:

**This PR now continues as #3592**, based on #3591 — so it shows only this layer's diff (+1153 −65, 21 files) instead of the cumulative +10139 across 56 files. Head SHA unchanged (`bce9eb03ed9a502282724d2069f7a55758310212`), upstream branch `feat/sc-resiliency-04-restart-readmission`.

Leaving this open for history; please move new work and discussion to #3592.

PR #3477, opened 2026-08-03
feat(refit): support BF16 to MXFP8 NCCL reshard

seonjinn:
## Summary

Enables NCCL-Reshard refit from a BF16 Megatron trainer to an MXFP8 vLLM rollout engine.

During refit, NCCL first reshards each BF16 weight into a worker-local buffer. vLLM then quantizes that local shard into E4M3 values and E8M0 scales, loads the existing MXFP8 runtime layout, and reuses the initialized MoE kernel across later refits.

Existing matching BF16 and blockwise-FP8 paths remain unchanged. Blockwise-FP8 trainer storage to MXFP8 rollout remains rejected because their scale formats are incompatible. Batched MoE layout optimization is handled separately in #3478.

The shared `quantize_mxfp8_weight` helper is also used by the existing IPC/collective MXFP8 load path. Restoring grouped-expert scale shapes and replacing FlashInfer zero-block E8M0 scale byte 0 with byte 1 there is intentional, so both refit transports satisfy the same MXFP8 consumer contract.

## Why

NCCL-Reshard previously required matching trainer and rollout storage, so BF16 training with MXFP8 rollout could not use the sharded refit path. This PR adds the required receiver-side conversion without trainer-side prequantization or changes to the collective protocol.

## Performance

Matched 20-step Qwen3-235B-A22B GRPO A/B on 16 B200 nodes (128 GPUs): 8 trainer nodes, 8 rollout nodes, MoE-only MXFP8 rollout. Steps 3-20 are averaged; only `refit_transport` differs.

| Metric | Legacy | NCCL-Reshard | Change |
| --- | ---: | ---: | ---: |
| Refit total | 18.71 s | 13.35 s | **-28.7%** |
| Transfer/update | 19.81 s | 14.13 s | **-28.7%** |
| E2E step time | 315.90 s | 310.33 s | **-1.8%** |
| E2E throughput | 101.42 | 103.16 tok/s/GPU | **+1.7%** |

The 5.36 s refit reduction accounts for 5.57 s of E2E improvement. Generation, training, and logprob timings remain within 0.6%.

[Legacy W&B](https://wandb.ai/nvidia/sna-pr3477-qwen235b-refit-ab/runs/mmuggufa) · [NCCL-Reshard W&B](https://wandb.ai/nvidia/sna-pr3477-qwen235b-refit-ab/runs/yx7bj48u)

Both runs recorded all 20 steps. The NCCL run finalized W&B before a post-run Ray teardown failure, so the reported metric history is complete.

### Correctness reruns after the grouped-expert placement fix

Both corrected runs completed 20 logged steps. The table averages steps 3-20 (18 steady-state steps with no missing rows).

| Model | `train/gen_kl_error`, mean / max | Refit total | Run |
| --- | ---: | ---: | --- |
| Qwen3-235B-A22B | 0.01159 / 0.01276 | 1.66 s | [W&B](https://wandb.ai/nvidia/sna-pr3477-qwen235b-refit-ab/runs/o7oa9jh0) |
| Qwen3-30B-A3B | 0.00399 / 0.00425 | 0.385 s | [W&B](https://wandb.ai/nvidia/sna-qwen30-matched-bf16-vs-mxfp8/runs/r0sfm4qc) |

The earlier `train/gen_kl_error > 2.0` failure is not reproduced in either corrected configuration. These runs validate the refit path and weight placement; they are not a claim of full convergence parity.

## Validation

- Covers supported and rejected precision pairs, including non-blockwise FP8 storage.
- Covers dense and grouped-MoE BF16-to-E4M3/E8M0 value and scale mappings.
- Rejects unknown or mismatched wire dtypes before raw refit copies.
- Validates grouped-expert shape restoration, MXFP8 scale shape/dtype, and one-time MoE-kernel initialization.
- Full GitHub CI is running for the latest review-fix commit.


seonjinn (commented):
Self-review completed with the upstream `/review-pr-team` skill on `98d00ab49a7a9cc9f86eaaba9e74e3527d26205c`. The review used the RL, bug-finding, design, test-coverage, existing-thread, and devil's-advocate agents.

Actionable findings resolved:
- Unknown or missing wire dtypes now fail loudly instead of silently bypassing BF16-to-MXFP8 receiver quantization.
- Added focused coverage for unsupported wire dtypes, missing scale tensors, invalid scale dtype, and non-divisible MXFP8 K dimensions.
- Expanded the quantization test and documentation to verify that only zero E8M0 scale bytes are sanitized while nonzero scale bytes are preserved.

Potential findings rejected after code-level verification:
- `merged_slice` is applied exactly once in the receiver pre-hook; the outer slice is validation-only.
- E8M0 byte 1 for all-zero blocks is intentional for TRTLLM kernel stability and is numerically equivalent because the associated FP8 values are zero.

Local verification passed: `ruff check`, `ruff format --check`, `compileall`, and `git diff --check`. Linux/vLLM tests are also being run on Lyris with the exact PR SHA; GitHub CI is requested below.

seonjinn:
Self-review validation update for `98d00ab49a7a9cc9f86eaaba9e74e3527d26205c`:

- Lyris GB200 SLURM job: `2677546`
- Environment: PyTorch `2.11.0+cu130`, vLLM `0.25.1`
- Result: **25 passed** in the targeted vLLM suites
  - `tests/unit/models/generation/test_nccl_reshard_backend.py`
  - `tests/unit/models/generation/test_vllm_fp8_quantization.py`
- Includes all four new metadata guard cases and CUDA batched-MoE coverage.

The earlier local limitation was macOS/Linux-lockfile incompatibility; the exact pushed SHA is now verified in the cluster container.

seonjinn (commented):
Self-review completed with `/review-pr-team` against commit `98d00ab49` and resolved in `778c218f6`.

The multi-agent review produced 8 candidate findings; adversarial validation filtered 6. I resolved the two remaining items as follows:

- Removed the redundant local `_STR_TO_DTYPE` import in `VllmInternalWorkerExtension._receive_and_load_misc_params`.
- Rechecked the trainer precision guard against the actual NeMo-RL policy setup. `policy.precision` is required and its canonical accepted values are `float32`, `bfloat16`, and `float16`; unlike vLLM precision, `bf16`, `auto`, and `None` are not valid policy values. I retained the exact `bfloat16` guard, documented this distinction, and added regression cases proving that `bf16` and `None` are rejected rather than silently accepted.

Validation on exact HEAD `778c218f6c1663d9ae13a0c3a6bb8b6559888e6a`:

- Local `ruff check`, `ruff format --check`, and `git diff --check`: passed.
- Lyris GB200 job `2677994`, PyTorch `2.11.0+cu130`, vLLM `0.25.1`: 25 vLLM/MXFP8 tests passed and 56 NCCL-reshard utility tests passed (81 total).
- Log: `/lustre/fsw/coreai_dlalgo_llm/users/sna/experiments/pr3477-exact-head/selfreview-778c218f6-final/slurm-2677994.out`

No unresolved correctness or design findings remain from this self-review.

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -632,17 +633,35 @@ def check_nccl_reshard_refit_support(master_config: dict) -> None:
...
-                    "policy.megatron_cfg.fp8_cfg.fp8_param=True "
-                    "(BF16→FP8 train-side quantization is not implemented yet)."
-                )
-            elif fp8_recipe != "blockwise":
-                violations.append(
-                    "policy.megatron_cfg.fp8_cfg.fp8_recipe must be 'blockwise' "
-                    f"when fp8_param=True (got {fp8_recipe!r}); other recipes "
-                    "don't produce export-ready scale_inv tensors."
+                    "policy.megatron_cfg.fp8_cfg.fp8_param=True, or "
+                    "is_mx=True for BF16-to-MXFP8 refit."
```
youngeunkwon0405, line 723: [`nemo_rl/weight_sync/nccl_reshard_utils.py:650-664`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/weight_sync/nccl_reshard_utils.py#L650-L664)

**[DOC]** Nice, complete precision matrix. One follow-through: [`docs/design-docs/nccl-reshard-refit.md:36-38`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/docs/design-docs/nccl-reshard-refit.md#L36-L38) still reads "BF16 train ↔ FP8 gen is not supported yet", which this PR makes false — and since no recipe YAML ships with the feature, that bullet is the main discovery path for users. Consider adding the new supported combination (BF16 train ↔ MXFP8 gen via `vllm_cfg.precision: fp8` + `vllm_cfg.is_mx: true`, receiver-side quantization) and noting the blockwise-FP8 → MXFP8 rejection there.

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -632,17 +633,35 @@ def check_nccl_reshard_refit_support(master_config: dict) -> None:
...
+                    )
+            elif vllm_cfg.get("is_mx"):
+                # Policy precision uses the canonical NeMo-RL spelling; unlike
+                # vLLM precision, it does not accept "bf16", "auto", or None.
+                if trainer_precision != "bfloat16":
+                    violations.append(
+                        "policy.generation.vllm_cfg.is_mx=True with "
+                        "policy.megatron_cfg.fp8_cfg.fp8_param=False requires "
+                        "policy.precision='bfloat16' for receiver-side MXFP8 "
+                        f"quantization (got {trainer_precision!r})."
```
youngeunkwon0405, line 717: [`nemo_rl/weight_sync/nccl_reshard_utils.py:650-658`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/weight_sync/nccl_reshard_utils.py#L650-L658)

**[GUIDELINE, pre-existing — non-blocking]** `is_mx` is only consulted inside the `gen_precision == "fp8"` branch here, and [`init_fp8` only engages when fp8 weights are in use](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/models/generation/vllm/quantization/fp8.py#L208-L212) — so a config with `vllm_cfg.is_mx: true` but `precision` unset/bf16 is silently ignored end-to-end: the user asks for MXFP8 and gets a BF16 rollout with no signal. This hole predates the PR and affects the IPC path identically, so it doesn't belong to this gate — flagging only because this PR makes `is_mx` load-bearing. A warn-or-raise where `is_mx` is consumed (init_fp8 / vllm_worker) would cover both refit paths, if you want to take it here or in a follow-up.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -441,6 +441,24 @@ def _is_fp8_weight(name, model):
...
+        mxfp8_e4m3_quantize,
+    )
+
+    value, scale = mxfp8_e4m3_quantize(weight)
+    value = value.reshape(weight.shape)
+    scale = scale.reshape(*weight.shape[:-1], weight.shape[-1] // 32)
+    scale = torch.where(scale == 0, torch.ones_like(scale), scale)
+    return value, scale
+
+
```
youngeunkwon0405, line 464: [`nemo_rl/models/generation/vllm/quantization/fp8.py:444-461`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/models/generation/vllm/quantization/fp8.py#L444-L461)

**[SCOPE + DOCSTRING]** Two related asks on `quantize_mxfp8_weight`:

1. This helper is also called from the existing `load_weights` path ([fp8.py:473](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/models/generation/vllm/quantization/fp8.py#L473)), so the new 0→1 scale remap and the 3-D scale reshape change the behavior of the **existing IPC/collective MXFP8 refit** as well, not just NCCL reshard — very likely fixing a latent grouped-expert issue there (vLLM 0.25.1's FlashInfer path restores the flat scale view [only for 2-D inputs](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/quantization/utils/mxfp8_utils.py), per its own `x_scales.ndim == 1 and x.ndim == 2` guard). Could you confirm that's intended and mention it in the PR description? It changes the review scope for anyone reading the title.

2. The docstring justifies the remap with "the FP8 values in those blocks remain zero", but `torch.where(scale == 0, ...)` rewrites *every* byte-0 block unconditionally — the equivalence relies on FlashInfer emitting byte 0 only for all-zero blocks, which the code doesn't check (numerically it would take a block amax below 2^-127 to matter, so this is a wording ask, not a bug claim). Consider stating the FlashInfer guarantee being relied on, or conditioning the substitution on the block values actually being zero.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -819,30 +820,119 @@ def build_hf_to_local_param_map(self, refit_info: dict) -> HFToLocalParamMap:
...
+                    )
+                specs[hf_name] = _mxfp8_receiver_quant_spec(
+                    vllm_param, scale_param, merged_slice
+                )
+            else:
+                specs[hf_name] = (
                     LocalParamSpec(base=vllm_param.data)
                     if merged_slice is None
                     else _merged_param_spec(vllm_param, merged_slice)
                 )
```
youngeunkwon0405: [`nemo_rl/models/generation/vllm/vllm_backend.py:926-934`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/models/generation/vllm/vllm_backend.py#L926-L934)

**[BUG, low]** `wire_dtype` is now computed for every param, but only the (bf16 wire, fp8 param) combination acts on it — every other mismatch falls through to the plain byte-copy `LocalParamSpec`, where e.g. a float32/float16 wire dtype landing on a bf16 param is silent corruption (exactly the failure class the MXFP8 branch guards against). There's no trainer-vs-gen dtype cross-check elsewhere either — `check_nccl_reshard_refit_support` passes `policy.precision: float32` with gen precision unset, and xferdtensor infers dtype from the destination. Since the receiver now has the information in hand, a one-line fail-loud in the else branch closes it:

```suggestion
            if wire_dtype != vllm_param.dtype:
                raise ValueError(
                    f"nccl_reshard: wire dtype {wire_dtype} != vLLM param dtype "
                    f"{vllm_param.dtype} for {hf_name!r} (unsupported refit combination)"
                )
            mapping[hf_name] = LocalParamSpec(base=vllm_param.data)
```
(adjust to the actual local variable names — the shape of the check is the point.)

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -819,30 +820,119 @@ def build_hf_to_local_param_map(self, refit_info: dict) -> HFToLocalParamMap:
...
+            if wire_dtype is None:
+                raise ValueError(
+                    f"build_hf_to_local_param_map: unsupported wire dtype "
+                    f"{wire_dtype_value!r} for {hf_name!r}"
+                )
+            if wire_dtype == torch.bfloat16 and vllm_param.dtype == torch.float8_e4m3fn:
+                vllm_name = vllm_names_by_id.get(id(vllm_param))
+                if vllm_name is None:
+                    raise ValueError(
+                        f"build_hf_to_local_param_map: resolved vLLM target for "
```
youngeunkwon0405, line 890: [`nemo_rl/models/generation/vllm/vllm_backend.py:871-890`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/models/generation/vllm/vllm_backend.py#L871-L890)

**[GUIDELINE, low]** Two maintainability notes in one: (a) recovering the vLLM param *name* via the `{id(param): name}` reverse lookup re-walks `named_parameters()` a second time and silently keeps the last name if two entries alias one tensor — `_build_hf_to_gen_backend_mapping` already has `vllm_name` in hand at every `mapping[hf_name] = ...` site, so widening the mapping value to `(vllm_name, vllm_param, merged_slice)` deletes both the lookup and the "is not a registered model parameter" branch (unreachable by construction). (b) nit: `torch.uint8` and the `// 32` literals here are hardcoded while the sibling MXFP8 code imports [`MXFP8_SCALE_DTYPE` / `MXFP8_BLOCK_SIZE`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/models/generation/vllm/quantization/fp8.py#L776-L778) from vLLM — importing the constants keeps a future vLLM bump from flipping this assertion into rejecting every valid config.

tests/unit/weight_sync/test_nccl_reshard_utils.py
```diff
@@ -62,6 +63,56 @@ def test_check_nccl_reshard_refit_support_accepts_valid_config() -> None:
...
+
+    check_nccl_reshard_refit_support(config)
+
+
+def test_check_nccl_reshard_refit_support_rejects_bf16_to_blockwise_fp8() -> None:
+    config = _valid_nccl_reshard_config()
+    config.policy["generation"]["vllm_cfg"]["precision"] = "fp8"
+
+    with pytest.raises(ValueError, match="is_mx=True for BF16-to-MXFP8 refit"):
+        check_nccl_reshard_refit_support(config)
```
youngeunkwon0405, line 119: [`tests/unit/weight_sync/test_nccl_reshard_utils.py:85-101`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/tests/unit/weight_sync/test_nccl_reshard_utils.py#L85-L101)

**[TEST]** The `fp8_recipe != "blockwise"` violation ([nccl_reshard_utils.py:644-649](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/nemo_rl/weight_sync/nccl_reshard_utils.py#L644-L649)) lost direct coverage in the restructure — it's now nested under `if fp8_param:` behind the `is_mx` check, and `fp8_recipe` only ever appears as `"blockwise"` in this file. Verified-passing addition (3/3 in a CPU venv):

```python
@pytest.mark.parametrize("fp8_recipe", ["tensorwise", "mxfp8", None])
def test_check_nccl_reshard_refit_support_rejects_non_blockwise_fp8_param(
    fp8_recipe: str | None,
) -> None:
    config = _valid_nccl_reshard_config()
    config.policy["generation"]["vllm_cfg"]["precision"] = "fp8"
    config.policy["megatron_cfg"]["fp8_cfg"] = {
        "fp8_param": True,
        "fp8_recipe": fp8_recipe,
    }

    with pytest.raises(ValueError, match="must be 'blockwise'"):
        check_nccl_reshard_refit_support(config)
```

tests/unit/models/generation/test_nccl_reshard_backend.py
```diff
@@ -313,3 +322,170 @@ def test_build_hf_to_local_param_map_specs_and_roundtrip():
...
+        "layer_names": ["model.layers.0"],
+        "per_layer_params": {
+            "model.layers.0": [
+                {
+                    "name": "model.layers.0.mlp.experts.gate_proj.weight",
+                    "global_shape": [E, Pl, H],
+                    "dtype": "torch.bfloat16",
+                    "grouped_expert_proj": "gate_proj",
+                },
+                {
```
youngeunkwon0405, line 340: [`tests/unit/models/generation/test_nccl_reshard_backend.py:327-340`](https://github.com/NVIDIA-NeMo/RL/blob/778c218f6c1663d9ae13a0c3a6bb8b6559888e6a/tests/unit/models/generation/test_nccl_reshard_backend.py#L327-L340)

**[TEST, optional]** All four new backend tests use 3-D grouped MoE experts; the dense gated-MLP MXFP8 path is untested — there the 1-D `merged_slice` is applied to both a 2-D `gate_up_proj` value `[2I, K]` and its scale `[2I, K/32]`, tensors with different trailing extents. It's correct (the slice only touches dim 0, and the build-time `expected_scale_shape` check fails loud on a wrong slice), but a dense analog of this test asserting gate and up land in disjoint `[0:I)` / `[I:2I)` halves of *both* the value and scale params would pin the slice-reuse contract cheaply.

youngeunkwon0405 (commented):
Team review: 4 agents (RL expert, test reviewer, bug finder, devil's advocate) + lead. Every finding was independently re-verified before posting; several initial findings were dropped or corrected during adversarial review.

**This is a solid, well-evidenced feature.** Verified clean against the pinned vLLM 0.25.1: scale shapes/naming (incl. the w13 merged-slice math for both value and scale tensors), E8M0 floor-rounding safety, 32-group vs TP/EP shard alignment (allocation-time %32 checks make straddling impossible), cross-refit re-entrancy (one-time MoE kernel init + copy_-based reload keeping captured tensor refs live), and misc-path quantization coverage. The A/B perf table is internally consistent (all rows recomputed). One notable verification: the 3-D scale reshape is load-bearing on the production path — vLLM 0.25.1's FlashInfer branch (the only one reachable on Blackwell) restores the flat scale view only for 2-D inputs, so grouped experts genuinely need it; a claim that it was a no-op (based on the torch fallback path) was raised and rejected during review. A one-line confirmation from a GB200 run that `mxfp8_e4m3_quantize` returns a flat scale for 3-D input would close the loop on that.

7 inline comments, none blocking: a design-doc support-matrix update (main ask), a scope/intent confirmation for the `quantize_mxfp8_weight` change that also affects the existing IPC/collective MXFP8 path, two test suggestions (one pre-verified), a cheap fail-loud dtype check, one combined maintainability note, and a pre-existing `is_mx` silent-drop observation explicitly not owned by this PR.

Coordination note: open PRs #3618/#3619 touch the same files (`nccl_reshard_refit` finalization tail, `check_nccl_reshard_refit_support`, the shared `_valid_nccl_reshard_config` fixture). The overlaps are textually adjacent but semantically independent — whoever lands second has a trivial rebase.

Linter (ruff check + format): clean. 56/56 CPU-runnable tests pass; the vllm-marked test files are correctly picked up by the L0 Vllm shards.

_Generated by Claude Code_

nemo_rl/weight_sync/nccl_reshard_utils.py
seonjinn: Agreed that silently ignoring `is_mx=true` when generation precision is not FP8 should be fixed. This behavior predates this PR and affects both IPC and NCCL-Reshard, so I am keeping it out of this transport-focused change and will handle it in a follow-up validation change at the shared `is_mx` consumption point.

nemo_rl/weight_sync/nccl_reshard_utils.py
seonjinn: Updated the precision requirements in the design doc. It now documents BF16 training to MXFP8 generation with `precision=fp8` and `is_mx=true`, receiver-side quantization, and the blockwise-FP8 to MXFP8 rejection. Included in 9d2a1d0c2.

nemo_rl/models/generation/vllm/quantization/fp8.py
seonjinn: Confirmed that the shared helper change is intentional for both NCCL-Reshard and the existing IPC/collective MXFP8 load path. I clarified the FlashInfer zero-block assumption in the docstring and updated the PR description to state the broader scope. Included in 9d2a1d0c2.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Added a fail-loud check for every wire/target dtype mismatch except the explicitly supported BF16-to-MXFP8 conversion, plus focused unit coverage. Included in 9d2a1d0c2.

tests/unit/weight_sync/test_nccl_reshard_utils.py
seonjinn: Added the parametrized negative test for `tensorwise`, `mxfp8`, and unset recipes with `fp8_param=True`. Included in 9d2a1d0c2.

tests/unit/models/generation/test_nccl_reshard_backend.py
seonjinn: Added the dense gated-MLP MXFP8 test. It verifies that gate/up values and their scale tensors land in disjoint `[0:I)` and `[I:2I)` regions. Included in 9d2a1d0c2.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Thanks. I am deferring the mapping tuple/name refactor because it changes the mapping contract and all of its tests without changing this feature behavior. The second `named_parameters()` walk happens only during refit setup, not on each refit. I am also keeping the MXFP8 format checks local in this PR rather than adding another vLLM-private import. I will address both as a focused cleanup if we standardize this mapping API.

nemo_rl/weight_sync/nccl_reshard_utils.py
seonjinn: Addressed in 68bf4c927. `init_fp8()` now rejects `is_mx=True` unless `precision="fp8"`, before loading the model config. Added coverage for unset, `auto`, `bf16`, and `bfloat16`, so the validation applies equally to IPC, conventional collective, and NCCL-Reshard initialization.

seonjinn:
/ok to test [68bf4c9](https://github.com/NVIDIA-NeMo/RL/pull/3477/commits/68bf4c92752909a2bf859cce7ce1ac188b9fac0e)

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -441,6 +443,25 @@ def _is_fp8_weight(name, model):
...
+    affected block is zero.
+    """
+    from vllm.model_executor.layers.quantization.utils.mxfp8_utils import (
+        mxfp8_e4m3_quantize,
+    )
+
+    value, scale = mxfp8_e4m3_quantize(weight)
+    value = value.reshape(weight.shape)
+    scale = scale.reshape(*weight.shape[:-1], weight.shape[-1] // 32)
+    scale = torch.where(scale == 0, torch.ones_like(scale), scale)
```
youngeunkwon0405, line 461: If TRTLLM breaks when it gets scale byte 0, then isn't that a TRTLLM MXFP8 kernel bug? Is there an ongoing upstream fix for this?

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -819,30 +820,124 @@ def build_hf_to_local_param_map(self, refit_info: dict) -> HFToLocalParamMap:
...
                 region = vllm_param.data[merged_slice]
                 return RefitCtx(buf=torch.empty_like(region), extra={"region": region})
 
-            def post(ctx):
+            def post(ctx: RefitCtx) -> None:
                 ctx.extra["region"].copy_(ctx.buf)
 
             return LocalParamSpec(base=vllm_param, pre=pre, post=post)
 
+        def _mxfp8_receiver_quant_spec(
```
youngeunkwon0405: I think this function name should be `_bf16_to_mxfp8_receiver_quant_spec`, since it assumes a bf16 input and also considering for the future `_mxfp8_to_mxfp8` case support.

youngeunkwon0405: Logic-wise, it's excellent!

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -819,30 +820,124 @@ def build_hf_to_local_param_map(self, refit_info: dict) -> HFToLocalParamMap:
...
+                if isinstance(wire_dtype_value, torch.dtype)
+                else _STR_TO_DTYPE.get(wire_dtype_value)
+            )
+            if wire_dtype is None:
+                raise ValueError(
+                    f"build_hf_to_local_param_map: unsupported wire dtype "
+                    f"{wire_dtype_value!r} for {hf_name!r}"
+                )
+            if wire_dtype == torch.bfloat16 and vllm_param.dtype == torch.float8_e4m3fn:
+                vllm_name = vllm_names_by_id.get(id(vllm_param))
```
youngeunkwon0405, line 887: I think all of these mxfp8-related gatings are unnecessary.

youngeunkwon0405: I guess just doing something like 
```
vllm_name = vllm_names_by_id[id(vllm_param)]
scale_param = vllm_params[f"{vllm_name}_scale_from_checkpoint"]

specs[hf_name] = _mxfp8_receiver_quant_spec(
    vllm_param,
    scale_param,
    merged_slice,
)
```
might be sufficient.

youngeunkwon0405 (commented):
Hi @seonjinn, I reviewed the code line by line, the implementation looks good to me in general. However, in your wandb reports in the PR note, I was able to see that both leagcy run and the nccl-reshard refit run are borken. The gen KL error is very high (>2.0) and Loss and Avg Reward are 0 for the both run. I think the code path is broken from more deeper place. I think we need to debug that issue first and confirm that leagacy refit works correctly and then validate this new feature.

seonjinn:
> Hi @seonjinn, I reviewed the code line by line, the implementation looks good to me in general. However, in your wandb reports in the PR note, I was able to see that both leagcy run and the nccl-reshard refit run are borken. The gen KL error is very high (>2.0) and Loss and Avg Reward are 0 for the both run. I think the code path is broken from more deeper place. I think we need to debug that issue first and confirm that leagacy refit works correctly and then validate this new feature.

Youngeun, thank you so much! This was caused by incorrect quantization scope inclusion after we bumped up vLLM version (0.20.2 -> 0.25.1) I made a PR to resolve this issue. #3652

nemo_rl/models/generation/vllm/quantization/fp8.py
seonjinn: Yes, I agree this should be treated as an upstream TRTLLM/FlashInfer MXFP8 compatibility issue. I could not find an existing public FlashInfer issue or PR that fixes scale byte 0 handling. The 0-to-1 remap here is therefore a temporary NeMo-RL workaround; it relies on FlashInfer emitting byte 0 only for all-zero blocks, so changing the scale byte preserves the dequantized zero values. I will track the upstream issue separately and link it here once one is available.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Thanks. I kept the dtype gate because this mapping also contains ordinary BF16 destinations; applying the MXFP8 receiver spec unconditionally would quantize parameters that should remain BF16. I also kept the fail-loud name, shape, and dtype checks so an incompatible vLLM parameter layout does not silently corrupt a refit. The direct `_scale_from_checkpoint` lookup is not sufficient for all supported models: the corrected Qwen3-235B run exposed routed-expert runtime parameters named with `_scale` instead. Commit 6f2b4fb9d therefore resolves either `_scale_from_checkpoint` or `_scale`, then builds `_bf16_to_mxfp8_receiver_quant_spec`; its unit test covers the runtime `_scale` case.

seonjinn:
Contributor self-review refresh for `563f10941165acf3f2b681c7f4236d362167d5ba` after the latest main merge.

No new local source-level blocker was found, and Ruff/format/config checks pass. This PR is not yet merge-ready based on source review alone: the earlier GPU runs reported invalid convergence signals, so a corrected BF16-to-MXFP8 convergence run is still required before merge.

nemo_rl/models/generation/vllm/quantization/fp8.py
youngeunkwon0405: Can you file an NVBug for this?

youngeunkwon0405 (approved):
LGTM, thanks for the contribution!
@terrykong, could you please provide your final review?

seonjinn:
> LGTM, thanks for the contribution! @terrykong, could you please provide your final review?

@youngeunkwon0405  Thank you so much for your help! Thanks to you, this PR was significantly improved in terms of correctness and functionality!

terrykong (approved):

PR #3478, opened 2026-08-03
perf(vllm): batch MXFP8 MoE refit shuffles

seonjinn:
## What does this PR do?

This PR extracts the MXFP8 MoE layout-transform optimization from #3294 into an independently reviewable change.

- Replace the per-expert W13/W2 value and scale shuffle loop with four batched row gathers.
- Cache the value-independent W13/W2 row permutations on each layer.
- Keep batched gather outputs transient so CUDA tensor objects are not retained across vLLM sleep/wake boundaries.
- Derive layout conversion from the selected MoE backend: FlashInfer TRTLLM always uses the batched path, the per-expert implementation remains a test oracle, and unsupported backends fail explicitly until they have a backend-specific conversion.

The optimization is transport-independent and does not modify weight-transfer metadata or precision negotiation. BF16-training/MXFP8-rollout support and the required vLLM 0.25.1 MXFP8 MoE kernel lifecycle are owned by #3477. The current branch temporarily retains the lifecycle block so it remains testable against `main`; it will be removed by rebasing after #3477 merges. The final #3478 diff will contain only the batched-shuffle optimization and its tests.

## Why?

Each MoE layer has W13 and W2 value tensors plus their E8M0 scale tensors. The reference path launches the same four value-dependent row shuffles once per expert. Stacking experts preserves the exact row permutation while reducing Python dispatch and kernel-launch overhead.

The resulting layout is specific to `Fp8MoeBackend.FLASHINFER_TRTLLM`. Selecting another backend now fails before mutating weights rather than silently applying the FlashInfer layout.

## Performance

Matched Qwen3-30B-A3B MXFP8, 8-node/32-GPU GB200, 20-step NCCL-Reshard A/B. Both arms used source `9706cdc`, the same container, runtime, topology, rollout schedule, and instrumentation; only the shuffle implementation changed. Values are arithmetic means over Steps `3-9,11-19`.

| Metric | Per-expert reference | Batched | Change |
|---|---:|---:|---:|
| Transfer + update (`n=15`) | 5.899 s | 0.894 s | -84.85% |
| E2E step time | 208.136 s | 204.397 s | -1.80% |
| E2E tokens/s/GPU | 1228.204 | 1250.614 | +1.82% |

Step 11 lacks the transfer/update metric in both logs. The non-refit workload phases remained within 0.85%, and mean generation length was identical. Reference: job `5884885`, [W&B](https://wandb.ai/nvidia/sna-vllm0251-flashinfer-trtllm/runs/3wcyte6u). Batched: job `5883555`, [W&B](https://wandb.ai/nvidia/sna-vllm0251-flashinfer-trtllm/runs/hnpwwpwm).

## Correctness

CUDA tests compare the batched output bitwise with the per-expert reference for gated aligned shapes, gated shapes requiring scale padding, and non-gated MoE shapes. Tests also verify backend-derived dispatch, explicit rejection of unsupported backends, stable runtime parameter objects and storage across refits, and no caller-retained gather output buffers.


seonjinn:
Final validation for `52f30c7e8d99fbc7cb9138b2972546f4d7fed8ed`:

### Matched 20-step A/B

Qwen3-30B-A3B MXFP8 on 8 OCI-HSG GB200 nodes / 32 GPUs with NCCL Reshard. Both arms used source `9706cdc`, the same immutable container, runtime, topology, rollout schedule, and instrumentation. Only `refit_batched_moe_shuffle` changed. Arithmetic means use Steps `3-9,11-19`.

| Metric | Per-expert reference | Batched | Change |
|---|---:|---:|---:|
| Transfer + update (`n=15`) | 5.899 s | 0.894 s | -84.85% |
| E2E step time | 208.136 s | 204.397 s | -1.80% |
| E2E tokens/s/GPU | 1228.204 | 1250.614 | +1.82% |

Step 11 lacks the transfer/update metric in both logs.

- Reference: job `5884885`, [W&B](https://wandb.ai/nvidia/sna-vllm0251-flashinfer-trtllm/runs/3wcyte6u)
- Batched: job `5883555`, [W&B](https://wandb.ai/nvidia/sna-vllm0251-flashinfer-trtllm/runs/hnpwwpwm)

Both jobs recorded and synced all 20 steps. After training, both hit the same unrelated Ray Python-finalization assertion and exited `1:0`; all requested comparison metrics were already complete.

### Regression and CI evidence

- Current CUDA test: job `5883915`, `3 passed`.
- Parent implementation negative control: job `5884340`, failed at the intended retained-output-buffer assertion.
- [Full CI run 30990382419](https://github.com/NVIDIA-NeMo/RL/actions/runs/30990382419): all 54 executed jobs passed; patch coverage is 100%.

youngeunkwon0405:
Should this behavior be derived from the selected backend rather than exposed as a user-facing knob? Both branches produce the FlashInfer TRTLLM layout, so setting `refit_batched_moe_shuffle=false` only selects the slower FI implementation; it does not make non-FlashInfer backends safe. Since the batched path is already verified bitwise against the reference, could we invoke it unconditionally when `self.mxfp8_backend == Fp8MoeBackend.FLASHINFER_TRTLLM`, keep the per-expert implementation only as a test oracle, and either dispatch the proper backend-specific refit conversion or fail loudly for other backends?

youngeunkwon0405:
For the other parts, looks reasonable to me.

youngeunkwon0405 (approved):
LGTM

seonjinn:
@youngeunkwon0405 Thank you so much for your time and for reviewing this PR, Youngeun!

PR #3494, opened 2026-08-04
feat(vlm): Nemotron Super Omni GRPO with MTP under context parallelism

DanialTaheri:
## Summary

Enables Nemotron Super Omni GRPO on the multimodal NeMo-Gym path, including Multi-Token Prediction under context parallelism.

## Fixes

**Multimodal rollouts**
- `nemo_gym.py` — the Super chat template zeroes its structured-image count when prompt text already contains a literal `<image>`. 1,149 of 60,299 rows in the blended Super dataset do (100% of two MMPR/NLVR2 sources), so the template emitted 601 placeholders against 600 projected features and `_merge_projected_media` aborted the step. Adds `sanitize_image_placeholders` and `pad_dynamic_image_shapes` for ragged tiling.
- `vllm_worker_async.py` — the embedded OpenAI server is built directly rather than through vLLM's CLI, where chat-template file *paths* are normally loaded. `OpenAIServingRender` expects literal Jinja, so a path was rendered as a template and every `<image>` marker was dropped.
- `nemo_rl/__init__.py`, `worker_groups.py` — prefer an explicit `HF_MODULES_CACHE` and make it importable in the current interpreter, and forward it to isolated worker initializers. Without this a `trust_remote_code` actor cannot import `transformers_modules` while unpickling its constructor arguments.

**MTP under context parallelism** (three bugs, each masked by the previous)
- `data.py` — Nemotron Omni consumes a full THD row and applies CP itself after rebuilding multimodal embeddings, so `input_ids` is passed unsharded. The MTP loss mask always took the CP-local shard, so mask and tokens had different layouts. Both call sites previously raised `NotImplementedError` rather than reconcile them; now the mask follows the same layout rule as `input_ids`.
- `train.py` — a model that slices CP itself returns `(output, sliced_loss_mask)` when handed a full-sequence `loss_mask`. That tuple reached the loss wrapper, which called `.narrow()` on it.
- `run_multimodal_grpo_nemo_gym.py` — did not pass `trains_mtp` to `configure_generation_config` (the text entrypoint does), so `_mtp_weights_from_refit` stayed `False` and the worker was routed into `load_mtp_weights_from_disk`. That path expects the MTP layer indexed positionally after the backbone; Nemotron checkpoints use an `mtp.*` namespace, so MTP speculative decoding failed with `No MTP layer weights for layers [88] found`. It also means the trained MTP head is now refit each step rather than the drafter staying frozen.

**Tower controls** — `radio_force_cpe_eval_mode` and the four `freeze_*` keys, applied only when present so provider defaults survive omission.

**Launcher** — Hydra's delete of `compilation_config.cudagraph_capture_sizes` errors when a recipe does not pin it; gated behind `MTP_DROP_CUDAGRAPH_CAPTURE_SIZES` (default unchanged).

## Known dependency

MTP at CP>1 also needs a Megatron-LM fix, submitted separately: the MTP block calls the model's embedding directly and never applies the sequence-parallel scatter that multimodal models rely on their outer forward to perform, so `_concat_embeddings` fails (8192 vs 1024 at 16384 tokens, CP=2/TP=8). Carried here as `patches/mtp-sp-scatter.patch` until that lands.

## Testing

Nemotron Super Omni 120B-A12B, TP8/EP16/CP2, 16 nodes, async GRPO:

- 50 steps at 32 prompts x 16 generations
- 50 steps at 128 prompts x 16 generations (2,048 rollouts/step), mean reward 0.45 over steps 1-25 rising to 0.53 over 26-50
- MTP trained at `mtp_loss_scaling_factor=0.1` with speculative decoding at ~75% acceptance, flat across the run
- `grad_norm` stable in 0.035-0.145

Full metrics, curves and per-run configuration: [W&B report — Upstream Super MTP to main NeMoRL](https://wandb.ai/nvidia/grpo-superv3omni/reports/Upstream-Super-MTP-to-main-NemoRL--VmlldzoxNzY1OTA3Ng)


tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.sh
```diff
@@ -0,0 +1,58 @@
...
+# ship with the repo, so this driver is parameterized rather than
+# self-contained, and is listed in disabled.txt rather than a nightly suite.
+# examples/nemo_gym/nemotron-3-super-omni/super_omni_launch.sh is the Slurm
+# wrapper that exports these and submits the job.
+: "${MODEL_PATH:?MODEL_PATH must point at the Super Omni HF checkpoint}"
+: "${TRAIN_PATH:?TRAIN_PATH must point at the Gym-format training jsonl}"
+VAL_PATH="${VAL_PATH:-$TRAIN_PATH}"
+
+cd $PROJECT_ROOT
+uv run examples/nemo_gym/run_multimodal_grpo_nemo_gym.py \
```
yfw: [`vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.sh:29`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.sh#L29)

**Blocker: all three new drivers invoke an entrypoint that does not exist.** `examples/nemo_gym/run_multimodal_grpo_nemo_gym.py` is not in the tree — and it has **never existed on `main`**. It lived only on PR #3414's pre-squash development branches; commit `faa185a36` ("consolidate into one entrypoint") folded its functionality into `run_grpo_nemo_gym.py` (+23 lines: the `is_vlm` processor branch, the `skip_tokenizer_init` assert, the `data_tokenizer` swap, and `processor=` into `setup()`), and #3414 reached `main` as the squash `154aebad2`.

Same line in all three: [`...-async-gym.v1.sh:29`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.sh#L29), [`...-mtp.v1.sh:33`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp.v1.sh#L33), [`...-mtp-specdec.v1.sh:34`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.sh#L34).

**The launcher already gets this right** — [`super_omni_launch.sh:36`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/nemo_gym/nemotron-3-super-omni/super_omni_launch.sh#L36) has `ENTRYPOINT="${ENTRYPOINT:-examples/nemo_gym/run_grpo_nemo_gym.py}"`. So the path was fixed in one place and missed in three, which is the signature of a rebase across the #3414 squash.

We confirmed `run_grpo_nemo_gym.py` fully covers the VLM path and is untouched by this PR: `is_vlm` processor branch at L155-161, `trains_mtp` at L169-178 feeding `configure_generation_config`, the `skip_tokenizer_init` assert at L180, and `processor=processor` into `setup()` at L251. So the fix is a one-line path change in each driver.

docs/guides/nemotron-3-super-omni.md
```diff
@@ -0,0 +1,123 @@
...
+The launcher turns W&B logging on, and the driver creates the run before any
+worker starts, so a missing credential ends the job about two minutes into a
+full allocation. Clusters that do not mount `/home` into the training
+container cannot read a `wandb login` credential from `~/.netrc`; only
+`WANDB_API_KEY` in the submitting environment reaches the job. Pick one:
+
+```bash
+export WANDB_API_KEY=<key>       # log live
+export WANDB_MODE=offline        # log locally, `wandb sync <run-dir>` later
+EXTRA_OVERRIDES="logger.wandb_enabled=false"   # skip W&B entirely
```
yfw: [`nemotron-3-super-omni.md:64`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/docs/guides/nemotron-3-super-omni.md#L64)

**The documented override variable isn't the one the launcher reads, and the W&B example actively backfires.** The guide uses `EXTRA_OVERRIDES` at L64, L71, L76 and L89; `super_omni_launch.sh` only ever reads `EXTRA_HYDRA_ARGS` (declared L40, checked L82, appended L154, and named in its own error text at L86).

The sharp case is this line. The guide offers `EXTRA_OVERRIDES="logger.wandb_enabled=false"` as the way to skip W&B, but the preflight guard at [`super_omni_launch.sh:82`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/nemo_gym/nemotron-3-super-omni/super_omni_launch.sh#L82) regexes `EXTRA_HYDRA_ARGS`:

```bash
if [[ ! "${EXTRA_HYDRA_ARGS}" =~ logger\.wandb_enabled=([Ff]alse|0) ]] \
    && [[ "${WANDB_MODE}" == "online" && -z "${WANDB_API_KEY:-}" ]]; then ... exit 1
```

With `EXTRA_OVERRIDES` set, `EXTRA_HYDRA_ARGS` stays empty, the regex doesn't match, and the launcher **exits 1 with the very error the guide is teaching you to avoid**. The documented 8-node smoke at L89-102 has the same problem, so a reader following the guide can't run it without a W&B key.

The launcher is self-consistent; the guide is the side to fix. Renaming all four occurrences to `EXTRA_HYDRA_ARGS` is the smallest change.

examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp.v1.yaml
```diff
@@ -0,0 +1,37 @@
...
+    # the main next-token objective is unchanged by the auxiliary head.
+    mtp_num_layers: 1
+    mtp_use_repeated_layer: false
+    mtp_detach_heads: true
+    # Train the MTP head: its loss is added to the training objective at this
+    # weight. Set 0.0 instead to keep the head present and refit into vLLM (so
+    # speculative decoding can use it) without training it, which is what the
+    # reference Super Omni CP=2 recipe does.
+    #
+    # Note the pinned Megatron-LM has no separate MTP gradient clipping
```
yfw: [`vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp.v1.yaml:22`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp.v1.yaml#L22)

**This WARNING is inverted, and it will misdirect whoever debugs `grad_norm` next.** The comment says the pinned Megatron-LM has no separate MTP gradient clipping — "neither clip_grads.py nor optimizer_config.py references MTP" — and concludes that MTP gradients are clipped jointly with the backbone, "the known cause of exploding grad_norm."

The premise is literally true and the conclusion is false: the mechanism lives in `optimizer.py`, not the two files that were grepped. At the pinned Megatron-LM (`6513e3e2`):

```python
MTP_GRAD_NORM_GROUP = 'mtp'                    # optimizer.py:150
GRAD_NORM_GROUP_ATTR = 'grad_norm_group'       # :151
SEPARATE_GRAD_NORM_GROUPS = (MTP_GRAD_NORM_GROUP,)  # :152
```
with `clip_grad_norm`'s docstring at `:373-378` stating *"Parameters in a registered grad_norm_group are excluded from the main gradient norm and clipped independently using their group norm"*, and the tagging at `multi_token_prediction.py:1686-1689`:
```python
if self.config.mtp_detach_heads:
    for param in self.parameters():
        param.grad_norm_group = 'mtp'
```
Both MTP recipes set `mtp_detach_heads: true`, so that path is exactly what fires. This came in via Megatron-LM `5b16c9964869`, which is present in **both** the current pin and #3568's `d12f6c8c9` — so the rebase won't fix it.

Three places carry this claim: here, [`...-mtp-specdec.v1.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.yaml#L346), and [`...-mtp-specdec.v1.sh:19-20`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.sh#L19). Since MTP grads *are* clipped separately, the observed `grad_norm` behaviour needs a different explanation — worth correcting rather than deleting, since the operational advice to watch `train/grad_norm` is still good.

examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.yaml
```diff
@@ -0,0 +1,141 @@
+defaults: ../../../nemo_gym/grpo_nanov3.yaml
+grpo:
+  num_prompts_per_step: 256
+  max_num_steps: 1000
+  val_period: 0
```
yfw, line 5: [`vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.yaml:5`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.yaml#L5)

**Validation is off, but checkpoint selection still keys on a validation metric — so best-checkpoint selection degrades to "latest", after warning on every save.** This affects all three recipes: expanding this file with `tools/config_cli.py expand` yields `val_period: 0` alongside an inherited `metric_name: val:total_reward/mean`, `higher_is_better: true`. The flattened specdec recipe makes it visible in one place ([L27-30, L84-86](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.yaml#L27): `val_period: 0`, `val_at_start: false`, `val_at_end: false`, `metric_name: val:total_reward/mean`).

The chain, traced through source:
1. `grpo.py:3435-3438` only validates the `val:`/`train:` prefix, so nothing fails at startup. `grpo.py:3445-3452` then warns on **every** checkpoint save — *"You asked to save checkpoints based on total_reward/mean but no val metrics were collected. This checkpoint will not be saved as top-k."* — so this is noisy rather than silent, though easy to lose in a 16-node Ray driver log.
2. `checkpoint.py:482` sorts on `x[2].get(self.metric_name, -float("inf"))` — every checkpoint ties at `-inf`.
3. `checkpoint.py:518` filters `[c for c in checkpoint_history if self.metric_name in c[2]]` — empty.
4. `checkpoint.py:532-537` then warns *"No checkpoints contain metric ... Returning latest checkpoint. Consider enabling val_at_end or adjusting val_period"* and returns the latest.

Two things suggest this is an oversight rather than a choice: the warning text names this exact misconfiguration, and the value is inherited from [`grpo_nanov3.yaml:62`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/nemo_gym/grpo_nanov3.yaml#L62) where it was coherent because that parent sets `val_period: 5`.

It's currently masked by `keep_top_k: 1000000` (nothing is pruned, so no checkpoint is lost) — but that masking is accidental. Simplest fix is `metric_name: null`, which the schema allows (`checkpoint.py:123`, `str | None`); alternatively enable `val_at_end`.

patches/README.md
```diff
@@ -0,0 +1,35 @@
+# Out-of-tree changes for Super Omni MTP at CP>1
```
yfw: [`patches/README.md`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/patches/README.md)

**Since this PR is landing on #3568, this whole directory can go — and it becomes actively misleading if it stays.** #3568 bumps Megatron-Bridge to `0c565c9a0`, which pins Megatron-LM `d12f6c8c9`, and that revision already contains the fix: it landed upstream as [`67700f79fbc7` — *fix(mtp): scatter the MTP embedding when the model's embedding does not* (#6256)](https://github.com/NVIDIA/Megatron-LM/commit/67700f79fbc70453917fabc19dbbf824f704be13) on 2026-08-08, byte-identical to this patch including the comment text.

We verified the patch **rejects** against the post-bump file (`git apply` -> `patch does not apply` at line 1103), while still applying cleanly at today's pin. So after the rebase the apply command in this README fails, and the "check whether applied" recipe at L21-23 would report "applied" for any unrelated local edit.

Deleting `patches/` also cleans up five references that would otherwise dangle:
- `patches/mtp-sp-scatter.patch`
- [`tests/test_suites/disabled.txt:21`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/test_suites/disabled.txt#L21) — "The two MTP variants additionally require patches/mtp-sp-scatter.patch applied..."
- [`...-mtp.v1.sh:20,22`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp.v1.sh#L20) and [`...-mtp-specdec.v1.sh:22,23`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.sh#L22)
- [`docs/guides/nemotron-3-super-omni.md:111-113`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/docs/guides/nemotron-3-super-omni.md#L111) — "This integration uses the repository-pinned Gym, Megatron Bridge, Megatron-LM, and vLLM paths instead", which is false today and becomes true only once the patch is gone.

Two notes if it survives for any reason: the README cites Megatron-LM `8590b5210` (not a valid object in NVIDIA/Megatron-LM), Bridge `2f0f8c80` and MLM `cd4afffa6` — the latter two are real ancestors but not what HEAD pins (`573e088c9c` / `6513e3e23`). And nothing in `docker/`, `pyproject.toml`, `ray.sub`, or CI applies or verifies the patch, so an unpatched tree dies mid-run inside `_concat_embeddings` after the 120B conversion has already burned the allocation.

For what it's worth the diagnosis behind the patch was correct and useful — the bug affects every Bridge provider with `scatter_embedding_sequence_parallel = False` once MTP is on, not just Omni.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -664,6 +665,25 @@ class NeMoRLOnlineRenderer(NeMoRLOpenAIServingMixin, OnlineRenderer):
         serving_chat_kwargs = serving_chat_default_kwargs | self.cfg["vllm_cfg"].get(
             "http_server_serving_chat_kwargs", dict()
         )
+        # The embedded server is constructed directly instead of through
```
yfw, line 668: [`vllm_worker_async.py:668`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/models/generation/vllm/vllm_worker_async.py#L668)

**This reimplements a first-class vLLM parameter, and the mechanism is dead in all three recipes you ship.**

The comment says *"vLLM 0.20's `OpenAIServingChat.__init__` does not accept `chat_template_kwargs`"*, but this repo pins **vLLM 0.25.1**, where the supported parameter exists as `default_chat_template_kwargs` — including on `OnlineRenderer.__init__` (`vllm/renderers/online_renderer.py`), which stores it and merges it request-last in `preprocess_chat`. That's the same "server default, request wins" semantics hand-rolled at L759-766 and L835-845. (Two smaller things in the same comment: it cites 0.20 against a 0.25.1 pin, and names `OpenAIServingRender`, which doesn't exist at 0.25.1 — the class here is `OnlineRenderer`.)

The repo already has the native pattern: [`vlm_grpo-nemotron-omni-30ba3b-circle-click-2n8g-megatron-tp2ep8.v1.yaml:108`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-circle-click-2n8g-megatron-tp2ep8.v1.yaml#L108) on `main` sets `default_chat_template_kwargs` inside `http_server_serving_chat_kwargs`.

It's also unreachable as shipped: all three new recipes set `chat_template_kwargs` under `env.nemo_gym.policy_model.responses_api_models.vllm_model.extra_body` (a per-request Gym path), not under `http_server_serving_chat_kwargs` — so the `pop` always returns `None` and both merge blocks are skipped by their own `if`.

Since `online_renderer` is shared by both `NeMoRLOpenAIServingChat` (L705) and `NeMoRLServingTokenization` (L827), passing `default_chat_template_kwargs=` to the `NeMoRLOnlineRenderer(...)` construction at L687 replaces **both** blocks in one line — and removes the `/tokenize` vs `/v1/chat/completions` divergence the second block exists to paper over.

The other half of this hunk — `load_chat_template` at L673-677 — is correct and worth keeping: the recipes pass a `.jinja` path, vLLM expects literal content, and the call is idempotent for anyone already passing inline Jinja.

tests/unit/models/megatron/test_megatron_data.py
```diff
@@ -427,21 +427,61 @@ def test_process_microbatch_marks_single_sequence_trailing_padding(
...
+            model_slices_context_parallel_inputs=True,
+            straggler_timer=MagicMock(),
+        )
+
+        assert result.mtp_loss_mask is not None
+        # The full row, not the CP-sharded one.
+        assert torch.equal(result.mtp_loss_mask, packed_idx0)
+        # And it lines up with the tokens the model will slice.
+        assert result.input_ids_cp_sharded is result.input_ids
+        assert result.mtp_loss_mask.shape[1] == result.input_ids_cp_sharded.shape[1]
```
yfw, line 509: [`test_megatron_data.py:484`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/unit/models/megatron/test_megatron_data.py#L484)

**Two easy hardening additions to what is otherwise a good new test.** The `torch.equal(result.mtp_loss_mask, packed_idx0)` assert genuinely discriminates — index 0 is width 8 and index 1 is width 4, so reading the CP-local shard fails — and together with the pre-existing sibling at L547 both arms of the new conditional are pinned. Two gaps though:

1. **This line is redundant rather than load-bearing.** `mock_pack.return_value` is a single tuple reused for *both* the `input_ids` and `mtp_loss_mask` pack calls, and on this path `input_ids_cp_sharded is input_ids` — so when the implementation is correct both sides are the same object. It *would* fail on a regression to index 1 (width 4 vs 8), but the preceding `torch.equal(result.mtp_loss_mask, packed_idx0)` already catches that, so this assert adds nothing on top. The docstring presents it as the load-bearing invariant ("validates that every token-aligned tensor it shards has the same sequence length"), which oversells it. Giving `mock_pack` a `side_effect` that returns a distinct tuple per call would make it a genuine cross-tensor check.

2. **Nothing asserts what was packed.** Unlike the sibling at [L584](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/tests/unit/models/megatron/test_megatron_data.py#L584) (`assert mock_pack.call_count == 2`, with the comment *"called once for input_ids and once for mtp_loss_mask"*), this test never touches `mock_pack`. A regression that re-packed `input_ids` in place of `mtp_loss_mask` would pass. One line closes it:

```suggestion
        assert mock_pack.call_count == 2
        assert torch.equal(mock_pack.call_args_list[1].args[0], data_dict["mtp_loss_mask"])
```

We ran the file: `--mcore-only` gives 36 passed / 3 failed / 3 skipped, this test passing, with all 3 failures being a missing `transformer_engine` in our environment and zero assertion failures.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -736,6 +756,14 @@ async def create_chat_completion(
                 f"val_top_p={generation_config['val_top_p']})"
             )
 
+
```
yfw: [`vllm_worker_async.py:759`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/models/generation/vllm/vllm_worker_async.py#L759)

**`ruff format --check` fails on this file**, which will block pre-commit and CI:
```
Would reformat: nemo_rl/models/generation/vllm/vllm_worker_async.py
1 file would be reformatted
```
It's a doubled blank line introduced just above this hunk. `ruff format` fixes it; `ruff check` is otherwise clean across the whole PR.

examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.yaml
```diff
@@ -0,0 +1,499 @@
+# Self-contained recipe: Nemotron Super Omni 120B-A12B, async GRPO + NeMo-Gym,
```
yfw: [`vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.yaml)

**`configs-minimize-check` will block both MTP recipes, and this one is the only recipe in the repo with no `defaults:` key.** `grep -L "^defaults:" examples/configs/recipes/{vlm,llm}/*.yaml` returns exactly this file.

We understand why: [`super_omni_launch.sh:38`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/nemo_gym/nemotron-3-super-omni/super_omni_launch.sh#L38) reads `cluster.num_nodes` with `awk` and can't resolve `defaults:`, so the block has to be literal in whatever file the launcher is pointed at. But flattening 499 lines is a heavy way to satisfy that — the sibling [`...-mtp.v1.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp.v1.yaml#L36) keeps `defaults:` and simply repeats the `cluster:` block, which is the same workaround at a fraction of the cost. The repo already ships the resolver, so the fix is small:
```
$ ./tools/config_cli.py expand <recipe>.yaml
cluster:
  gpus_per_node: 8
  num_nodes: 16
```
Pointing the launcher's `awk` at `expand` output instead of the raw file removes the justification for **both** the flattening and the `cluster.num_nodes` duplication, clears the pre-commit hook, and would also let the hardcoded `--gres=gpu:8` at [`super_omni_launch.sh:208`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/nemo_gym/nemotron-3-super-omni/super_omni_launch.sh#L208) derive from `cluster.gpus_per_node` like everything else in `SBATCH_ARGS`.

The cost of hand-flattening is already visible. `mcore_generation_config.refit_backend` is `nvshmem` here with the comment *"Options: gloo or nvshmem"* — a value **and** comment that appear nowhere in the declared four-layer chain; they match `examples/configs/grpo_math_1B.yaml:389`, while the real parent `grpo_nanov3.yaml:247` has `nccl` / *"Options: gloo or nccl"*. It's inert (`grpo.py:1424-1427` reads it only when `backend == "megatron"`, and these use vLLM), but it falsifies this file's header claim to be "the fully resolved merge of what were previously four layers".

Three smaller artifacts of the same flattening: a comment says PIECEWISE is set "above" when `cudagraph_mode: PIECEWISE` is ~21 lines below; `val_batch_size` / `val_num_generations_per_prompt` / `max_val_samples` are carried into a recipe with validation disabled; and `grpo.async_grpo.max_generation_failures: 0` present in the chain was dropped here (value-inert, since `AsyncGRPOConfig` defaults it to 0, but a fourth sign the merge was done by hand). `./tools/config_cli.py compare` against the mtp recipe lists the full set.

Also, both MTP recipes' `cluster:` comment names `super_launch.sh` — the launcher that reads these files is `super_omni_launch.sh`; `super_launch.sh` is the text-only Super launcher, which this PR's own launcher header says it deliberately doesn't share.

docs/guides/nemotron-3-super-omni.md
```diff
@@ -0,0 +1,123 @@
...
+| Setting | Value |
+|---|---|
+| Total allocation | 16 nodes, 8 GPUs per node |
+| Policy parallelism | TP=8, EP=16, CP=2 |
+| Generation allocation | 8 non-colocated nodes |
+| Prompts and generations | 256 prompts, 16 generations per prompt |
+| Train global batch size | 4096 |
+| Maximum sequence length | 16,384 |
+| Async policy lag | at most one version |
+| MTP and speculative decoding | disabled |
```
yfw: [`nemotron-3-super-omni.md:24`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/docs/guides/nemotron-3-super-omni.md#L24)

**Two documentation gaps, both small.**

1. This guide isn't registered in [`docs/index.md`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/docs/index.md#L295). The Guides toctree lists `guides/nemotron-3-nano-omni.md` and `guides/nemotron-3-super.md` but not this page, so Sphinx will emit an "isn't included in any toctree" warning and the page won't be reachable from the nav. It belongs right after `guides/nemotron-3-super.md`.

2. The guide covers only the base recipe — this table says "MTP and speculative decoding | disabled", and the two MTP recipes are never named. Given the PR is titled *"with MTP under context parallelism"*, the headline feature is the undocumented one.

docs/guides/nemotron-3-super-omni.md
```diff
@@ -0,0 +1,123 @@
...
+in the range 0.5 to 2.0, in-flight weight updates, and the existing NeMo RL
+policy-to-vLLM refit path.
+
+The visual encoder and projection remain trainable. The sound encoder and
+projection are frozen for this image-and-text workload. The model's own chat
+template is passed to both the Hugging Face tokenizer and the vLLM chat server.
+
+## Launch
+
+The launcher defaults to the checkpoint, blended Gym data, Slurm account, and
```
yfw: [`nemotron-3-super-omni.md:36`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/docs/guides/nemotron-3-super-omni.md#L36)

**This paragraph describes defaults the launcher doesn't have.** It says the launcher "defaults to the checkpoint, blended Gym data, Slurm account, and cache settings used by the migration source script", and that "its default NeMo RL image uses the Python 3.13 runtime".

In practice every one of those is a placeholder — [`super_omni_launch.sh:20-27`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/nemo_gym/nemotron-3-super-omni/super_omni_launch.sh#L20) sets `/path/to/nemotron-super-omni-hf-checkpoint`, `/path/to/super-omni-gym-train.jsonl`, `/path/to/nemo-rl.sqsh`, `your_slurm_account` — and the preflight at L62-75 **hard-exits** on exactly those values. There is no working default image.

The placeholder-plus-fail-fast design is the right one; the text just predates it. Something like "every path must be exported; the launcher fails fast on the placeholders" would match the behaviour.

nemo_rl/environments/nemo_gym.py
```diff
@@ -354,6 +470,19 @@ class NemoGym(EnvironmentInterface):
 
     def __init__(self, cfg: NemoGymConfig):
         self.cfg = cfg
+        initial_global_config_dict = cfg.get("initial_global_config_dict") or {}
+        self._pad_dynamic_image_shapes = bool(
```
yfw: [`nemo_gym.py:474`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/environments/nemo_gym.py#L474)

**These two new config fields are declared but never populated, so the primary lookup is dead — and the default ends up in a pattern `config-conventions` forbids.**

```python
self._pad_dynamic_image_shapes = bool(
    cfg.get("pad_dynamic_image_shapes",
            initial_global_config_dict.get("pad_dynamic_image_shapes", False)))
```

`pad_dynamic_image_shapes` and `sanitize_image_placeholders` are declared as top-level `NemoGymConfig` fields at [L130-135](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/environments/nemo_gym.py#L130), but the only constructor — `spinup_nemo_gym_actor` at [L1034-1044](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/environments/nemo_gym.py#L1034) — builds `NemoGymConfig(...)` from an explicit keyword list that doesn't include them. So `cfg.get(...)` can never hit and the value always comes from the `initial_global_config_dict` fallback.

The siblings right above show the established pattern: `invalid_tool_call_patterns`, `thinking_tags` and `tokenizer_config` are popped off `nemo_gym_dict` in the builder at [L1021-1023](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/environments/nemo_gym.py#L1021) and passed as real fields, under a comment saying NeMo-RL-side knobs belong there rather than in Gym's global config. Following it here would fix all three issues at once — the dead declarations, the chained `.get()` with hidden `False` defaults, and the need for the compensating pops at [L533-534](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/environments/nemo_gym.py#L533):

```suggestion
        self._pad_dynamic_image_shapes = bool(cfg.get("pad_dynamic_image_shapes"))
        self._sanitize_image_placeholders = bool(cfg.get("sanitize_image_placeholders"))
```

nemo_rl/__init__.py
```diff
@@ -281,11 +281,16 @@ def _patch_nsight_file():
 # Need to set PYTHONPATH to include transformers downloaded modules.
 # Assuming the cache directory is the same cross venvs.
 def patch_transformers_module_dir(env_vars: dict[str, str]):
-    hf_home = os.environ.get("HF_HOME", None)
-    if hf_home is None:
-        return env_vars
+    # Prefer the explicit cache. A per-run HF_MODULES_CACHE keeps concurrent
+    # runs from racing on generated module names, and deriving the directory
+    # from HF_HOME instead would silently point at a different tree.
+    module_dir = os.environ.get("HF_MODULES_CACHE")
```
yfw, line 289: [`__init__.py:287`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/__init__.py#L287)

**This precedence change breaks 8 of 10 existing tests whenever `HF_MODULES_CACHE` is set in the environment** — and this PR's own launcher exports it ([`super_omni_launch.sh:109`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/examples/nemo_gym/nemotron-3-super-omni/super_omni_launch.sh#L109)), so it's likely to bite the exact developers working on this feature. Reproduced:

```
$ pytest tests/unit/test_transformers_module_patching.py -q
10 passed

$ HF_MODULES_CACHE=/tmp/probe pytest tests/unit/test_transformers_module_patching.py -q
8 failed, 2 passed
```

The cause is that those tests set up with `patch.dict(os.environ, {"HF_HOME": tmpdir})` without `clear=True`, and `main` never read `HF_MODULES_CACHE` at all, so an ambient value was previously harmless. The precedence itself looks deliberate and is well-commented — it's the tests that need a hermetic fixture:

```suggestion
    @pytest.fixture(autouse=True)
    def _isolate_hf_modules_cache(self, monkeypatch):
        """HF_MODULES_CACHE now outranks HF_HOME, so it must not leak in."""
        monkeypatch.delenv("HF_MODULES_CACHE", raising=False)
```
(verified: 10 passed with the var set). Worth preferring this over `clear=True`, which would also wipe unrelated ambient vars. One caveat for future reuse: `raising=False` on an absent variable records nothing, so this clears an ambient value but won't undo a write made *inside* a test.

Neither of the two genuinely new behaviours has coverage — the `HF_MODULES_CACHE`-wins branch, and the `sys.path` mutation below (including that it doesn't duplicate on repeated calls). We have 6 passing tests for these if useful.

nemo_rl/environments/nemo_gym.py
```diff
@@ -210,6 +217,80 @@ def _looks_like_image_src(src: str) -> bool:
...
+    """Count structured image items in a Responses-API Gym example."""
+    count = 0
+    input_items = nemo_gym_example.get("responses_create_params", {}).get("input", [])
+    if not isinstance(input_items, list):
+        return count
+    for item in input_items:
+        if not isinstance(item, dict):
+            continue
+        content = item.get("content", [])
+        if not isinstance(content, list):
```
yfw: [`nemo_gym.py:237`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/environments/nemo_gym.py#L237)

**This is the largest new surface in the PR and has zero test coverage.** We grepped `tests/` and `origin/main` for `sanitize_nemo_gym_example_image_placeholders`, `_count_image_payloads` and `_normalize_image_placeholders` — no hits anywhere. It's pure, deterministic dict logic needing no GPU, so it's about as cheap to test as code gets, and it encodes several contracts the comments call out explicitly: non-mutation (rollout batches share example objects), no fabricated `content` key on `function_call`/`reasoning` items, and multimodal -> `""` vs text-only -> `"image"`.

The ragged `pixel_values` path at L398-457 is likewise uncovered — the `return_tensors=None` switch, the `len(images) > 1` gate, the inner `PackedTensor(pad_to_max_shape=True)`, and both new `ValueError`s.

One question rather than a finding: `_normalize_image_placeholders` strips `` and `</img>` unconditionally in **both** branches, including the text-only branch that carefully rewrites `<image>` to the word `"image"` to avoid destroying content. Is the asymmetry intentional? With `mcqa` and `string_match` servers in these recipes, a prompt legitimately quoting those tags would be silently altered, and there's no debug line recording how many rows were touched.

nemo_rl/models/megatron/setup.py
```diff
@@ -647,6 +650,14 @@ def setup_model_config(
     if "layernorm_epsilon" in config["megatron_cfg"]:
         model_cfg.layernorm_epsilon = config["megatron_cfg"]["layernorm_epsilon"]
 
+    # Provider objects loaded from checkpoint metadata otherwise retain the
+    # serialized defaults. Apply explicit recipe controls before model
+    # construction so RADIO positional encoding and frozen towers are stable
+    # and consistent between logprob and training passes.
+    for vlm_key, vlm_value in iter_vlm_config_overrides(config["megatron_cfg"]):
+        if hasattr(model_cfg, vlm_key):
```
yfw: [`setup.py:658`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/models/megatron/setup.py#L658)

**This `hasattr` guard turns a misconfiguration into a silent no-op.** Same shape at [`community_import.py:132-135`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/models/megatron/community_import.py#L132).

The obvious defence is that the guard protects text-only models that have no `freeze_vision_model` — but that case never reaches here. `iter_vlm_config_overrides` already yields only keys the recipe explicitly set (`if key in megatron_config`), and its docstring says so: *"omitting one keeps the provider's own default rather than silently forcing False"*. So a text model yields nothing, and the only way to reach `hasattr == False` is that a user asked for `freeze_vision_model: true` on a provider that doesn't support it — and gets an unfrozen, silently-trained tower with no diagnostic. That's the silent-on-misconfiguration case `CLAUDE.md` asks us to flag, and Megatron-Bridge's own contributor guide carries the mirror-image rule.

```suggestion
        if not hasattr(model_cfg, vlm_key):
            raise ValueError(
                f"megatron_cfg set '{vlm_key}' but {type(model_cfg).__name__} has no "
                "such field; this provider does not support that tower control."
            )
        setattr(model_cfg, vlm_key, vlm_value)
```

Two small extras: `iter_vlm_config_overrides` is a generator with no return annotation (`-> Iterator[tuple[str, Any]]`), which the linting skill asks for on new functions.

nemo_rl/__init__.py
```diff
@@ -294,6 +299,14 @@ def patch_transformers_module_dir(env_vars: dict[str, str]):
     else:
         env_vars["PYTHONPATH"] = f"{module_dir}:{env_vars['PYTHONPATH']}"
 
+    # Updating PYTHONPATH only affects interpreters started after this process.
+    # Ray actors may import NeMo RL before unpickling trust_remote_code objects,
+    # so make the dynamic module package importable in the current interpreter.
+    if env_vars is os.environ:
```
yfw: [`__init__.py:305`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/__init__.py#L305)

**Small readability point: this branches on caller identity to decide whether to mutate global `sys.path`.** The parameter is typed `dict[str, str]`, and nothing in the signature reveals that passing `os.environ` specifically has a second, process-global effect — a caller passing `dict(os.environ)` (a natural thing to do when building a worker environment) silently gets different behaviour.

In practice it isn't really a branch: `patch_transformers_module_dir` is called exactly once in production, at L313 with `os.environ`; every other call site is in the unit tests. So the condition is unconditionally true in production and false only under test, which makes it a test-visibility artifact controlling a global side effect. An explicit keyword would make the contract visible and directly testable:

```suggestion
    if apply_to_current_interpreter:
```
with `apply_to_current_interpreter: bool = False` in the signature and `True` at the L313 call site.

One related note: the docstring above says a per-run `HF_MODULES_CACHE` "keeps concurrent runs from racing on generated module names", but nothing in `nemo_rl/` establishes a per-run cache — that comes solely from the launcher's `${PERSISTENT_CACHE}/hf_modules/${EXP_NAME}` default, so two runs sharing an `EXP_NAME` still share the directory.

nemo_rl/environments/nemo_gym.py
```diff
@@ -339,6 +436,25 @@ def _attach_multimodal_data_to_user_message(
         if key not in processed:
             continue
         value = processed[key]
+        if key == "pixel_values" and isinstance(value, list):
```
yfw: [`nemo_gym.py:439`](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/environments/nemo_gym.py#L439)

**The ragged path can raise `AttributeError: 'list' object has no attribute 'ndim'`.** When `allow_ragged_output` is taken the processor is called with `return_tensors=None` (L406), so `processed["pixel_values"]` is a plain Python list — but the `imgs_sizes` reconstruction guard just above at [L412-417](https://github.com/NVIDIA-NeMo/RL/blob/df8a81ad950839d72a4dca46032e63fa473970ce/nemo_rl/environments/nemo_gym.py#L412) evaluates `processed["pixel_values"].ndim` before this normalization runs:

```python
if (uses_placeholder and "pixel_values" in processed
        and "imgs_sizes" not in processed
        and processed["pixel_values"].ndim == 4):   # <- list has no .ndim
```

The chain short-circuits only when `imgs_sizes` **is** present — and this branch exists precisely for processors that don't emit it. Your own comment at L410-411 says so: *"Historical checkpoints may emit dynamic image tiles without imgs_sizes."* So the crash sits inside the compatibility path written to handle that very case.

To be clear about severity: this is **latent for the three recipes you ship**, since the Nemotron Omni processor does emit `imgs_sizes` — the other preconditions are all satisfied (`pad_dynamic_image_shapes: true`, `limit_mm_per_prompt.image: 3`, and `uses_image_placeholder` is asserted at L498). It would bite a historical-checkpoint processor.

The fix is free: hoist this ragged->tensor normalization above the guard, or add an `isinstance(processed["pixel_values"], torch.Tensor)` term to the chain.

yfw (commented):
Thanks for this — enabling Nemotron Super Omni GRPO with MTP under context parallelism is a substantial piece of work, and the validation in the description (50 steps at 128x16, mean reward 0.45 -> 0.53, ~75% MTP acceptance held flat, `grad_norm` 0.035-0.145, plus the W&B report) is exactly the evidence we want on a change like this.

**The core MTP-under-CP work holds up under scrutiny, and we want to say that clearly before the list below.** We verified:

- The MTP loss-mask layout change is correct, and it's the *only* layout the model accepts — `_apply_context_parallel_sharding` raises on inconsistent tensor lengths, so the previous CP-local read would have failed loudly at step 1 rather than silently mis-training. `NemotronOmniModel` is the only class in the pinned Bridge tree setting `model_slices_context_parallel_inputs = True`, so this is live code for a real model, not scaffolding.
- The `train.py` tuple unwrap is necessary and safe. We hunted for a model that would lose element `[1]` and found none: `HybridModel`/`GPTModel` return Tensors, the fused-CE path returns a Tensor, Qwen3VL's dist-train PP path returns a `dict`, and mcore's `output_processor` hook (the one real candidate) is never installed anywhere in `nemo_rl/`.
- All five tower-control fields exist on the Bridge provider, are non-deprecated, and are applied the way Bridge's own recipes apply them.
- Your new unit test **passes** (`pytest tests/unit/models/megatron/test_megatron_data.py --mcore-only` -> 36 passed; the 3 failures are a missing `transformer_engine` in our environment, with zero assertion failures).
- Config validation is green: 584 passed across `test_config_validation`, `test_config_v2`, and `test_recipes_and_test_suites`.
- No cluster paths, account names, or usernames leak into any new file — only the intended `/path/to/...` placeholders, with a hard-fail preflight behind them.

**Note that the review assumes a dependency on #3568 .** We checked the coupling since this is meant to land on that bump: the Omni contract is unchanged at Bridge `0c565c9a0`, and `unwrap_model` at Megatron-LM `d12f6c8c9` already handles the `FullyShardedDataParallelV1/V2` split. That last one is worth knowing — had it not, `getattr(chunk, "model_slices_context_parallel_inputs", False)` would have quietly returned `False` and the Omni path would have regressed to CP-local slicing with no error. There is also no merge conflict: your only hunk in `megatron_policy_worker.py` is ~2300 lines from #3568's. One thing we could **not** verify: `_apply_context_parallel_sharding` gains a `padding_mask` parameter at the bumped pin — almost certainly inert since NeMo-RL never passes it, but worth one grep after you rebase.

Most of what follows is mechanical. The one blocker is that all three shipped test drivers invoke a script that has never existed on `main`.

_Reviewed by a team of six agents plus two independent adversarial passes, which dropped seven candidate findings (including two that would have been wrong to send you) before this review._

_Generated by Claude Code_

tests/test_suites/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.sh
DanialTaheri: Resolved in commit ca57f9185187c9a2c0a68e0d5c3858fd54bb87d0

docs/guides/nemotron-3-super-omni.md
DanialTaheri: Addressed in d29658d186dfb0fa205458f1eb36ff1e755cf74b

examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp.v1.yaml
DanialTaheri: Addressed in d0628e5f3b5f3ec622f8e474059ee8f6b63970a0

examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym.v1.yaml
DanialTaheri: Resolved in 81446a8f5e4a2ef1ce4c0e04e998772a759e9886

nemo_rl/models/generation/vllm/vllm_worker_async.py
DanialTaheri: Resolved in d3088b4bada84ae7ba0fa389fdf61d488dc36e0b

examples/configs/recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-16n8g-megatron-tp8ep16cp2-async-gym-mtp-specdec.v1.yaml
DanialTaheri: Resolved in a258750f49a601bd9a626617ea386b4e4569b495

nemo_rl/models/generation/vllm/vllm_worker_async.py
DanialTaheri: Addressed in 649f95eaec23f14143f736aecf235d184eb15c0f

docs/guides/nemotron-3-super-omni.md
DanialTaheri: Addressed in 9a131788be78978a1deef5b03800226c8bcb0912

docs/guides/nemotron-3-super-omni.md
DanialTaheri: Resolved d06c194ac461a0654b1a33bb5c492941d863d448

nemo_rl/environments/nemo_gym.py
DanialTaheri: Resolved in 951f7388480d2c6a88384ba910706f16a7a205e5

nemo_rl/environments/nemo_gym.py
DanialTaheri: This is a problem. The Megatron LM tries to match image tokens with <image> in the user prompts. We need sanitization to match this number on the multimodal. For instance if we have extra <image> tokens in the user prompt we remove them.
The issue is with the text data and training with VLMs that use Llava in Megatron. In samples with <image> token and no image, Llava returns an error. That is why we modified that to "image". I see this introducing problem in the text side potentially.

DanialTaheri: I do not like _sanitization functions and what we have here. But seems to be the only way till we fix the MLM.

nemo_rl/environments/nemo_gym.py
DanialTaheri: Addressed in commit 0c3d0667d21f268ffd75c56ca0e47506319cded2

patches/README.md
DanialTaheri: Addressed in fc11b496f700d0cef980858df0f7966a10767d16

nemo_rl/__init__.py
yfw: This one is my fault — apologies. That suggestion block contained a pytest fixture meant for `tests/unit/test_transformers_module_patching.py`, but I anchored it to a comment on this source file, so applying it pasted the fixture into `patch_transformers_module_dir` and **replaced** the `module_dir = os.environ.get("HF_MODULES_CACHE")` line it landed on.

At `fc11b496`, `import nemo_rl` raises `NameError: name 'pytest' is not defined`. Because this function runs at module scope (`:316`), the whole package is unimportable — `ruff check` also reports `F821` for `pytest`, `module_dir`, and `apply_to_current_interpreter`.

Worth flagging the collateral: the `HF_MODULES_CACHE`-first read is gone entirely now, so the original finding is un-fixed as a side effect.

I did check locally that restoring the assignment and threading the flag through as a keyword argument gets `import nemo_rl` working again. The fixture belongs in the test file, where I should have put it. Neither of the two new behaviours (the `HF_MODULES_CACHE` precedence, and the `sys.path` insert) has a test yet, if you want to pick those up in the same pass.

nemo_rl/models/megatron/setup.py
yfw: Also mine, sorry. My suggestion replaced the `if hasattr(...)` line but silently assumed it also consumed the `setattr` beneath it. Applied as-is, the original survived at its old indentation:

```python
        setattr(model_cfg, vlm_key, vlm_value)
            setattr(model_cfg, vlm_key, vlm_value)
```

`nemo_rl/models/megatron/setup.py` no longer compiles (`SyntaxError: Unexpected indentation`, :664); removing the stray line clears it, which I confirmed locally.

Two things still open on this finding: the sibling site at `nemo_rl/models/megatron/community_import.py:133` is untouched and still silently drops an explicitly-configured override, and `iter_vlm_config_overrides` (`community_import.py:26`) still has no return annotation. (My original comment said `nemo_rl/utils/` — wrong path, that was my error too.)

nemo_rl/__init__.py
yfw: The rename landed at the call site but not in the signature, so `apply_to_current_interpreter` is an undefined name at `:308` while the signature is still `(env_vars: dict[str, str])`.

Either add it as a keyword-only parameter defaulting to `False` and pass `True` at the module-level call site, or simply restore `env_vars is os.environ` — this was only a readability nit on my part and it isn't worth churn. Your call.

nemo_rl/models/generation/vllm/vllm_worker_async.py
yfw: My recommendation here was wrong, and acting on it introduced a regression — sorry about that. I said passing `default_chat_template_kwargs=` to the shared renderer would cover both endpoints. Reading vLLM 0.25.1 as installed, it doesn't:

- `ServingTokenization` takes and reads its **own** `default_chat_template_kwargs` (`vllm/entrypoints/serve/tokenize/serving.py:40,54`) and passes *that* into `preprocess_chat` (`:89`) — not the renderer's. So `/tokenize` now renders with `{}`, which is the divergence the block you deleted was preventing.
- `OpenAIServingChat` likewise stores its own (`chat_completion/serving.py:142`) and uses it to build the **reasoning parser** (`:190`, `:257`). Since `vllm_worker_async.py:683` now pops the native key out of `serving_chat_kwargs`, that class receives `{}`. On the merge-base the native spelling passed straight through, so this path worked before and doesn't now. It's latent for the recipe that uses it (`vlm_grpo-nemotron-omni-30ba3b-circle-click...v1.yaml:108` sets `enable_thinking: true`, matching the fallback), but `false` would silently mis-configure the parser.

vLLM's own server passes it to all three consumers (`entrypoints/openai/api_server.py:368/383/393`), which is probably the shape to follow — normalize the legacy alias into the native key and hand it to the renderer, the chat serving, and the tokenization serving, rather than popping it.

Also in the same block: `pop(A) or pop(B)` short-circuits, so a recipe that sets both spellings leaves `chat_template_kwargs` in the dict and `OpenAIServingChat(**serving_chat_kwargs)` raises `TypeError`.

The `load_chat_template` half of this hunk is correct and worth keeping.

nemo_rl/environments/nemo_gym.py
yfw: Verified fixed. I ran the old and new guard blocks over four input shapes: two differently-shaped CHW arrays now yield `[[224, 336], [448, 112]]` — genuinely per-image sizes rather than one shape repeated — where the old code raised `AttributeError: 'list' object has no attribute 'ndim'`. Stacked 4-D behaviour is unchanged, and a malformed 4-D-per-entry list still falls through to your `ValueError`. Nice fix.

One leftover from the config change: the compensating pops at `nemo_gym.py:535-536` are now dead, since the builder already removes those keys before they reach `initial_global_config_dict`. Harmless, but deletable.

docs/guides/nemotron-3-super-omni.md
yfw: Rename verified across all four sites, and the W&B guard now matches — thanks.

One leftover in the same block: `:101` is `EXTRA_HYDRA_ARGS="logger.wandb_enabled=false"` with no `export`, while the two options just above it (`:99-100`) do use `export`. A plain assignment doesn't reach the child `bash super_omni_launch.sh`, so a reader who copies that line still hits the "W&B logging is on but WANDB_API_KEY is unset" abort it's meant to avoid. The prefix form used at `:112-114` works; `export` on `:101` would too.

nemo_rl/environments/nemo_gym.py
yfw: Two of my original points I withdraw outright:

- **The ``/`</img>` strip.** I assumed those were ordinary text in the text-only branch. They are not — `megatron/core/tokenizers/vision/libraries/multimodal_tokenizer.py:18-22` defines `IMAGE_TAGS` with `"internvl": ("", "</img>")`, and `_apply_image_tag` wraps `<image>` in them, matching the `...<image>*N...</img>` expansion your docstring describes. Stripping them uniformly is consistent, not collateral damage.
- **Per-row vs per-item counting.** The invariant is per-sequence, so deriving the replacement once from the row's image count is the right granularity. My comment was wrong.

Still worth doing, all small and none blocking: a unit test (the function has no coverage today and is pure dict logic, so it is cheap); a debug-level count of mutated rows, so the blast radius is auditable when this is pointed at a new data blend; and a docstring line naming the enforcement point and the tag pair — the current docstring tells only the `<image>` story, which is what sent me to the wrong model in the first place.

> The issue is with the text data and training with VLMs that use Llava in Megatron. In samples with `<image>` token and no image, Llava returns an error. That is why we modified that to "image". I see this introducing problem in the text side potentially.

We may still need to fix this for Llava model, but just want to double check, after Ali's change to use `NemotronOmniModelProvider`, we're no longer using Llava model for omni right? Then is this still an issue for Omni?

yfw (commented):
Re-reviewed at `fc11b496`. Thanks for turning this around so quickly. The substantive work all landed, and I verified it rather than taking the commit messages at face value.

**Confirmed fixed:** the entrypoint blocker (all three drivers; zero repo-wide references to the dead path), the `EXTRA_OVERRIDES` -> `EXTRA_HYDRA_ARGS` rename, the MTP grad-clipping warning (the *new* text checks out against Megatron-LM `d12f6c8c9`), checkpoint selection on all three recipes, `minimize-check` (exit 0 across 197 configs, re-confirmed with a deliberate control probe), the toctree entry and MTP section, the ragged `imgs_sizes` reconstruction (verified on genuinely differently-shaped tiles), the MTP test hardening (now load-bearing), and the `patches/` deletion — the SP-scatter fix really is in the pinned Megatron-LM, so dropping it is correct.

**Three of the follow-ups below are my fault, not yours.** Two of my earlier comments used ` ```suggestion ` blocks that weren't valid at the line they were anchored to, and applying them left `nemo_rl` unimportable (`NameError` at `__init__.py:287`) and `nemo_rl/models/megatron/setup.py` non-compiling. A third was bad advice from me about vLLM's `default_chat_template_kwargs`. Details in the threads.

**Merge note:** the only conflict is `uv.lock`. #3568 landed on main (`46ab18ce1`) and relocked it there, while this branch bumped Megatron-Bridge without relocking — merging main and taking its `uv.lock` should clear it, and your own bump commit is now redundant since main already pins `0c565c9a0`.

_Generated by Claude Code_

nemo_rl/__init__.py
yfw: Confirmed the import is repaired — `import nemo_rl` works again, `ruff check` is clean, and the new `test_apply_to_current_interpreter_puts_module_dir_first` pins the keyword behaviour. Thanks for untangling a mess I made.

One piece of the original finding is still open, and it got slightly worse with the new tests: the suite fails when `HF_MODULES_CACHE` is set in the environment.

```
$ pytest tests/unit/test_transformers_module_patching.py
30 passed

$ HF_MODULES_CACHE=/tmp/probe pytest tests/unit/test_transformers_module_patching.py
9 failed, 5 passed
```

Cause is unchanged: the tests set up with `patch.dict(os.environ, {"HF_HOME": tmpdir})` without clearing `HF_MODULES_CACHE`, and that variable now outranks `HF_HOME`. Before this PR it was never read, so an ambient value was harmless. The new cases inherit the same pattern, which is why the count went from 8 to 9.

The fix is an autouse fixture on `TestPatchTransformersModuleDir` that does `monkeypatch.delenv("HF_MODULES_CACHE", raising=False)`. Two notes from trying it locally:

- **`tests/unit/test_transformers_module_patching.py` does not import `pytest`** — adding just the fixture gives a collection error and takes the whole file down. It needs the import too. (I walked straight into this myself, which is a fitting end to this particular thread.)
- Prefer the fixture over adding `clear=True` to the nine `patch.dict` calls; `clear=True` would also wipe unrelated ambient variables.

With both lines in place I get 14 passed clean and 14 passed with `HF_MODULES_CACHE` set, `ruff` clean.

Entirely reasonable to do this as a follow-up rather than hold the PR — it does not affect CI or any training run.

yfw (commented):
Follow-up after the repair round. Everything from the last review is verified fixed — import restored, both VLM override sites aligned, the chat-template defaults now reaching all three serving consumers, the W&B opt-out exported, and new coverage for the placeholder sanitiser. One item from the original findings is still open; details inline.

_Generated by Claude Code_

nemo_rl/environments/nemo_gym.py
yfw: Addressed in follow up PR https://github.com/NVIDIA-NeMo/RL/pull/3631

yfw (dismissed):

yfw (dismissed):

yfw (approved):

Issue #3495, opened 2026-08-04
Distillation loss path upcasts the full [B, S, V] logits to fp32 when only K columns are read


tianyi-zhang-02:
*(Edited: this originally led with the `logprob_chunk_size` point, now the second section. While checking my shapes against the shipped configs I found something simpler and more general, so I reordered and corrected the framing.)*

## The full logits are upcast to fp32 before it is known that only K columns are needed

`get_distillation_topk_logprobs_from_logits` opens with

```python
student_logits = student_logits.to(torch.float32)   # [B, S, V]
```

Three of its four branches read only `K` columns from that tensor:

- `zero_outside_topk=False`, non-TP → `student_logits.gather(-1, teacher_topk_indices)`
- `zero_outside_topk=False`, TP → `gather_logits_at_global_indices(...)`, which casts each chunk to fp32 internally
- `zero_outside_topk=True`, TP → `ChunkedDistributedGatherLogprob`, likewise

Only `zero_outside_topk=True` on the non-TP path requires full-vocab fp32, for its `log_softmax`.

Gathering before the cast is equivalent, since `gather` is a selection and `to()` is elementwise, and produces bitwise-identical values and gradients. Measured on an RTX 4090 at B=1, V=151,936, K=64, the values shipped in `distillation_math.yaml` (`train_micro_batch_size: 1`, `topk_logits_k: 64`, `Qwen3-1.7B-Base`):

| seq_len | cast-then-gather (current) | gather-then-cast | saved |
|---:|---:|---:|---:|
| 2048 | 1.74 GiB | 0.58 GiB | 1.16 GiB |
| 8192 (config default) | 6.96 GiB | 2.33 GiB | 4.63 GiB |

At seq 8192 the gather also drops from 9.2 ms to 0.1 ms, since the current form moves 2.3 GiB through memory to retain 64 columns per position. This is an improvement on both axes rather than a tradeoff, and `zero_outside_topk=False` is the default, so it applies to the common path.

Filed as #3496.

## Separate point: `logprob_chunk_size` does not reach these paths

`policy.logprob_chunk_size` is threaded through the logprob paths (`automodel/train.py:733,744`, `dtensor_policy_worker.py:1224,1234`, `megatron/train.py:596`, `loss/utils.py:93,354`) but not through these two:

[...]

PR #3496, opened 2026-08-04
perf(loss): defer the fp32 upcast in the distillation top-k path

tianyi-zhang-02:
## What does this PR do?

Defers the fp32 upcast in `get_distillation_topk_logprobs_from_logits` until after the non-TP top-k gather. This avoids materializing an fp32 `[B, S, V]` tensor when only `K` logits are used.

The TP/CP and `zero_outside_topk=True` paths keep the existing full upcast. `DistillationLossFn` now requires both masks instead of falling back to an unmasked mean; every in-tree caller already supplies them, and the invariant is required for gradient equivalence at padded duplicate indices.

Refs #3495. `logprob_chunk_size` remains out of scope.

## Validation

Refreshed head: `252f378c3c50c2c671e5a2811537ffd452ca23db` (includes `main` at `fd7112c374af23b3bcc392637fb737495f8a3f31`).

On 2026-09-21, I ran both changed unit-test files on Runpod Secure Cloud, 1× H100 80 GB, `nvcr.io/nvidia/nemo-rl:v0.7.0`, driver 580.126.09, PyTorch 2.11.0+cu130, CUDA 13.0. The current-main test fixture required `pytest==9.0.3` and the pinned `nemo-lens` commit (`b85578f`), installed only in the temporary test environment.

| Run | Result | Scope |
| --- | --- | --- |
| H100 visible | 119 passed, 24 skipped | `test_loss_functions.py` and `test_model_utils.py`; all skips require 2 GPUs |
| CPU-only (`CUDA_VISIBLE_DEVICES=""`) | 65 passed, 78 skipped | Same files; GPU-dependent cases skipped |

Earlier, on head `7534040209a051be3c7db0d0de8b582a065834ea`, I measured the real helper in separate processes at `B=1`, `V=151,936`, `K=64`, using `main → candidate → candidate → main` for each sequence length. Every repeat produced identical output and gradient hashes; outputs were fp32, gradients bf16, and all values were finite.

| S | main peak delta | this PR | reduction |
| ---: | ---: | ---: | ---: |
| 64 | 76.1 MiB | 18.6 MiB | 4.09× |
| 2,048 | 2.32 GiB | 0.58 GiB | 3.99× |
| 8,192 | 9.28 GiB | 2.32 GiB | 3.99× |

These are isolated PyTorch allocator peak deltas, not end-to-end training HBM measurements. The microbenchmark was not rerun after the main-only refresh; 2-GPU TP tests also remain unrun in this 1-GPU session.


yuki-97 (commented):
@fujial-code to review

nemo_rl/distributed/model_utils.py
```diff
@@ -1996,9 +2003,19 @@ def get_distillation_topk_logprobs_from_logits(
...
+            # Gathering K columns and then widening is bitwise equal to
+            # widening the whole vocab axis first: bf16 -> fp32 is exact and
+            # gather is pure selection.
+            #
+            # Duplicate indices do occur -- padded sequence positions carry K
+            # copies of index 0, because dtensor_policy_worker.py pads
+            # topk_indices with value=0 -- so the gather backward accumulates
+            # there rather than scattering. The equality holds anyway, because
+            # DistillationLossFn masks those positions to exactly 0.0 before
+            # reduction, so the accumulated gradient is zero either way.
```
fujial-code, line 2185: The loss also has an unmasked fallback (kl_loss = per_token_kl.mean()) when token_mask/sample_mask are absent, where padded duplicate-index positions would carry nonzero grads and the accumulation order would differ (bf16 post-rounding vs fp32 pre-rounding). Consider asserting that the masks are present in DistillationLossFn, so the invariant this comment relies on is enforced rather than assumed.

tests/unit/distributed/test_model_utils.py
```diff
@@ -1583,3 +1584,92 @@ def test_distributed_vocab_topk_ops(
...
+    )[:, :-1, :]
+    expected.sum().backward()
+
+    assert h_all is None
+    assert topk_logprobs.dtype == torch.float32
+    assert torch.equal(topk_logprobs, expected)
+    assert torch.equal(student.grad, reference.grad)
+
+
+def test_distillation_topk_gathers_before_upcasting(monkeypatch):
```
fujial-code, line 1714: Nit: An equivalent refactor to torch.gather(...) / index_select would keep the optimization but fail this test as a false alarm. Maybe note this in the docstring so a future refactorer knows to update the spy rather than suspect a real regression.

tianyi-zhang-02:
Thanks — the unmasked fallback point is right, and it was the weak part of that comment.

`DistillationLossFn` had `kl_loss = per_token_kl.mean()` when the masks were absent, and on that path the padded positions get real gradient. That is exactly where the deferral stops being equivalent: those positions carry K duplicate copies of index 0 (`dtensor_policy_worker.py` pads `topk_indices` with `value=0`), so the gather backward accumulates rather than scatters, and the accumulation rounds in bf16 instead of fp32. My comment asserted the equality held because the loss masks them — true only on the branch that has masks.

Enforced it as you suggested: the loss now raises instead of falling back. Nothing reached the fallback — `distillation.py` always sets `token_mask`, and `setup_distillation_test_data` sets both — so this is a dead path becoming a loud one rather than a behaviour change. Added a CPU test that the raise fires for either mask missing; it goes red if the silent `mean()` comes back.

Also took the docstring nit — the spy is coupled to `Tensor.gather`, so a rewrite to `torch.gather`/`index_select` would keep the deferral and still trip it. Said so in the docstring, pointing at updating the spy rather than suspecting a regression.

nemo_rl/distributed/model_utils.py
tianyi-zhang-02: Right, and that was the weak part of the comment.

`DistillationLossFn` fell back to `kl_loss = per_token_kl.mean()` when the masks were absent, and on that branch the padded positions get real gradient. Those are exactly the positions carrying K duplicate copies of index 0, since `dtensor_policy_worker.py` pads `topk_indices` with `value=0`. So the gather backward accumulates there instead of scattering, and it rounds in bf16 rather than fp32. That is the one case where deferring the upcast stops being bitwise equivalent. My comment claimed the equality held because the loss masks those positions — true only on the branch that has masks.

Enforced it the way you suggested: it raises now instead of falling back. Nothing reached the fallback anyway — `distillation.py` always sets `token_mask`, and `setup_distillation_test_data` sets both — so this turns a dead path into a loud one rather than changing behaviour. Added a CPU test that the raise fires for either mask missing. Put the silent `mean()` back and it goes red :)

tests/unit/distributed/test_model_utils.py
tianyi-zhang-02: Taken. The spy is bound to `Tensor.gather` specifically, so a rewrite to `torch.gather` or `index_select` would keep the deferral and still trip it. The docstring now says so, and points at updating the spy rather than reading it as a regression :)

tianyi-zhang-02:
Head moved to `8a57e3b14` since your `/ok to review 006a4eb7`, so that pin is stale — ready for a fresh one whenever convenient.

Also: I re-measured on the 4090 and the description's memory table was wrong in our favour. It now reads 2.90 → 1.16 GiB at `B=1, S=2048` and 11.61 → 4.65 at `S=8192`, i.e. 2.5x rather than the 1.5x originally claimed. Body updated with the correction inline.

tianyi-zhang-02:
Final H100 follow-up on SHA `7534040209a051be3c7db0d0de8b582a065834ea`:

- kept the required-mask invariant from your review and merged current `main`;
- ran the real top-k helper in isolated processes on an H100 80 GB (`torch 2.11.0+cu130`, driver `580.126.09`) at `B=1`, `V=151,936`, `K=64`, in `main → candidate → candidate → main` order;
- output and gradient hashes are identical in all 12 runs;
- PyTorch allocated peak delta improves by `4.09× / 3.99× / 3.99×` at sequence lengths `64 / 2,048 / 8,192` (about `57 MiB / 1.74 GiB / 6.96 GiB` saved);
- 4 focused tests pass, and Ruff check/format are clean.

This measures the isolated helper's allocator peak, not total end-to-end training HBM.

cc @fujial-code — this is the final branch after your mask feedback.

Issue #2937, opened 2026-06-25
Perf test failing with NCCL error


kajalj22:
**Describe the bug**

Perf test llm_performance_grpo_qwen3_30ba3b_4n4g  and llm_performance_grpo_qwen3_30ba3b_4n4g_async_1off failing with NCCL error

Similar related failure from release test: llm_grpo_gptoss_20b_8n4g_megatron

**Steps/Code to reproduce bug**

Tests failing off the main branch:

```
(VllmGenerationWorker pid=331005) INFO 06-24 09:11:00 [cumem.py:216] CuMemAllocator: sleep freed 105.73 GiB memory in total, of which 56.88 GiB is backed up in CPU and the rest 48.85 GiB is discarded directly. [repeated 15x across cluster]
(VllmGenerationWorker pid=331005) INFO 06-24 09:11:00 [gpu_worker.py:175] Sleep mode freed 106.05 GiB memory, 9.47 GiB memory is still in use. [repeated 15x across cluster]
(VllmGenerationWorker pid=331005) INFO 06-24 09:11:00 [abstract.py:331] It took 1.106435 seconds to fall asleep. [repeated 15x across cluster]
(raylet) A worker died or was killed while executing a task by an unexpected system error. To troubleshoot the problem, check the logs for the dead worker. Lease ID: 4b000000c32da20c959c26eea79a6299c72312ccb66b90fbc405925650cde385 Worker ID: bfe6d8300c58644324432b13a85e5ce95f5aa912dd6ac2bdc78f0451 Node ID: 4b955aeb106f6f5a0942e5c9d4367d91f03b94b0fa24358464c1489b Worker IP address: 10.109.18.239 Worker port: 10015 Worker PID: 1530113 Worker exit type: SYSTEM_ERROR Worker exit detail: Worker unexpectedly exits with a connection error code 2. End of file. Some common causes include: (1) the process was killed by the OOM killer due to high memory usage, (2) ray stop --force was called, or (3) the worker crashed unexpectedly due to SIGSEGV or another unexpected error.
Traceback (most recent call last):
  File "/lustre/fs1/portfolios/coreai/projects/coreai_dlalgo_nemorl/nemo_rl_ci/main/55699249/RL/code_snapshots/grpo-qwen3-30ba3b-4n4g/examples/run_grpo.py", line 225, in <module>
    main()
    ~~~~^^
[...]

PR #3498, opened 2026-08-05
build(automodel): bump Automodel to r0.6.0

### LINKED ISSUE #3319: Automodel bump up + Refactor to use the generic CP interface to be provided by Automodel
**Is your feature request related to a problem? Please describe.**
See the detailed problem descriptions and discussions in https://github.com/NVIDIA-NeMo/Automodel/issues/2861

**Describe the solution you'd like**
Refactor NeMo RL side of logic to use the unified CP interface from Automodel https://github.com/NVIDIA-NeMo/Automodel/pull/2937 once ready.

jQizhang:
# What does this PR do?

Upgrade the NeMo-RL DTensor v2 Automodel backend to Automodel r0.6.0, align compatible dependency forks on Transformers 5.12.1, and migrate context parallelism (CP) to Automodel's model-owned `ContextParallelSharder` / `ShardLayout` protocol.

For detailed API changes, support boundaries, before/after GRPO and x-token workflows, and experiment results, see [`docs/design-docs/automodel-context-parallel.md`](https://github.com/jQizhang/NeMo_RL/blob/bump-automodel-r0.6.0/docs/design-docs/automodel-context-parallel.md).

Full before/after results for all 10 workloads (pre-upgrade CP1 and post-upgrade CP1/CP2/CP4) are available in the [W&B `nemorl-cp` project](https://wandb.ai/bigbigqi/nemorl-cp/table?nw=nwuserbigbigqi).

### Automodel r0.6.0 upgrade

- Bump the Automodel submodule from `24b47e85` to [`7e9493d7`](https://github.com/NVIDIA-NeMo/Automodel/commit/7e9493d7dde987cc753caf10737ef785e71f3d8c) and track the `r0.6.0` branch.
- Migrate model construction to `MeshContext.build(...)`, `ParallelismSizes`, and one `DistributedSetup` passed to `from_pretrained()`; follow relocated imports and updated checkpoint APIs.
- Wait for asynchronous DCP writes before promoting `tmp_step_N`, preventing incomplete model or optimizer checkpoints.
- Mirror r0.6.0's `megatron-fsdp` Git constraint, pin TileLang / `apache-tvm-ffi` 0.1.11, and expose Z3 libraries to symlinked worker environments.
- Remove obsolete token-classification, checkpoint-dtype, and Gemma 4 KV-sharing workarounds.

### Transformers 5.12.1 upgrade

Raise the shared Transformers resolution to 5.12.1 and add exact pins to compatible backend extras. TRT-LLM retains its backend-constrained version.

| Module / extra | `origin/main` declaration | PR declaration | `origin/main` lock | PR lock | Upgrade status |
| --- | --- | --- | --- | --- | --- |
| Base / main environment | `>=5.5.0,<5.9.0` | `>=5.5.0,<=5.12.1` | `5.5.4` | **`5.12.1`** | Upgraded |
| FSDP | Inherits Base | `==5.12.1` | `5.5.4` | **`5.12.1`** | Added an exact pin and upgraded |
| Automodel | `>=5.5.0,<5.6.0` | `==5.12.1` | `5.5.0` | **`5.12.1`** | Upgraded for Automodel r0.6.0 |
| MCore / Megatron | Depends on Base and transitive constraints from Megatron-Bridge | `==5.12.1` | `5.8.1` | **`5.12.1`** | Added an exact pin and upgraded |
| vLLM | No standalone pin; inherits Base / transitive dependencies | `==5.12.1` | `5.8.1` | **`5.12.1`** | Added an exact pin and upgraded |
| SGLang | `==5.12.1` | `==5.12.1` | `5.12.1` | `5.12.1` | Unchanged from current `origin/main` |
| TRT-LLM | No root-level Transformers pin; constrained by TRT-LLM | No root-level Transformers pin; constrained by TRT-LLM | `5.5.4` | `5.5.4` | Unchanged |
| ModelOpt | Inherits Base | Inherits Base | `5.5.4` | **`5.12.1`** | Upgraded with Base |
| NVRx | Inherits Base | Inherits Base | `5.5.4` | **`5.12.1`** | Upgraded with Base |
| NeMo Gym | Inherits Base | Inherits Base | `5.5.4` | **`5.12.1`** | Upgraded with Base |

`uv` conflicts isolate TRT-LLM from extras using Transformers 5.12.1, including SGLang. The SGLang fork also preserves non-CUDA tensor reductions.

### Context-parallel integration

- Keep canonical RL/loss tensors separate from the private model batch, preserve the direct `CP=1` path, and construct a `ContextParallelSharder` only for `CP>1`.
- Use the actual `ShardLayout` to shard next-token targets and restore logprobs, top-k/full logits, and loss inputs instead of reconstructing the legacy round-robin layout.
- Make gradient fanout explicit: replicated GRPO-style losses use `loss × DP`, while partitioned x-token KD/CE retains `loss × DP × CP`; preserve x-token CUDA IPC, heterogeneous TP/CP, projection, alignment, and multi-teacher aggregation.

The upstream CP API is described in [Automodel #2937](https://github.com/NVIDIA-NeMo/Automodel/pull/2937); the distributed setup migration follows [Automodel #2266](https://github.com/NVIDIA-NeMo/Automodel/pull/2266).

### Tests and documentation

- Add focused CP layout/sharder, loss, checkpoint, Automodel setup/train/data, policy-worker, and SGLang regression tests.
- Add an upgrade design document and CP1/CP2/CP4 comparison plots for Gemma 4 E2B, Gemma 4 26B, Qwen3.5 35B-A3B, and Qwen3.5 9B.

# Issues

Closes #3319: [Automodel bump up + Refactor to use the generic CP interface to be provided by Automodel](https://github.com/NVIDIA-NeMo/RL/issues/3319).

Follow-up: [#3842 — Track nightly OOMs after the Automodel r0.6.0 upgrade](https://github.com/NVIDIA-NeMo/RL/issues/3842) tracks fixes for the four affected nightly cases after this PR merges.

# Usage

There is no new user-facing training API. Existing Automodel recipes continue to set `context_parallel_size`; NeMo-RL now delegates model-side token layout to Automodel.

# Validation

Targeted CUDA/H100 and Ray validation completed during development (before the final rebase) covered CP migration, Automodel data/setup/train, loss, x-token, worker autocast, and distributed x-token TP2/CP1 and TP1/CP2 tests. Ruff and format checks also passed.

After the final rebase, `uv lock --check` with uv 0.11.28, `git diff --check origin/main...HEAD`, and `python -m py_compile nemo_rl/models/automodel/train.py` passed. Current GitHub checks pass for submodule fast-forward, lockfile freshness, copyright, secret scanning, DCO, and PR-title validation.

Short CP1/CP2/CP4 GRPO/DAPO runs show similar reward, generated-token, and generation-KL trends; these are functional and short-run numerical checks, not long-horizon convergence results. Full before/after results for all 10 workloads (pre-upgrade CP1 and post-upgrade CP1/CP2/CP4) are available in the [W&B `nemorl-cp` project](https://wandb.ai/bigbigqi/nemorl-cp/table?nw=nwuserbigbigqi).

# Additional information

- The CP refactor is limited to Automodel DTensor v2; Megatron CP behavior is unchanged.
- VLM training with `CP>1` remains rejected. X-token distillation keeps contiguous CP loss/IPC windows and does not support sequence packing.
- Full-sequence gather remains a correctness-first migration boundary; a fully CP-native loss implementation can be pursued separately.


docs/design-docs/automodel-upgrade-intro.md
```diff
@@ -0,0 +1,264 @@
...
+
+### 1.1 Environment Upgrade
+
+Automodel r0.6.0 requires the following dependency updates in `pyproject.toml`; `uv.lock`
+is regenerated accordingly:
+
+| Dependency | Before | After |
+| --- | --- | --- |
+| Base `transformers` range | `>=5.5.0,<5.9.0` | `>=5.5.0,<5.13.0` |
+| Automodel extra | `transformers>=5.5.0,<5.6.0` | `transformers>=5.12.1,<5.13.0` |
```
jQizhang: **1 action item.**

The "After" column doesn't match what the PR actually pins — the numbers here look like they were taken from the unrelated `<5.13.0` assert in [`_patch_transformers_tokenizer_class_set`](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/nemo_rl/models/policy/__init__.py#L45), not from the extras.

Actual: [`pyproject.toml:57`](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/pyproject.toml#L57) is `transformers>=5.5.0,<=5.12.1` and [`pyproject.toml:134`](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/pyproject.toml#L134) is `transformers==5.12.1`. The table also doesn't mention that `fsdp`, `vllm` and `mcore` each gained an exact `transformers==5.12.1` pin ([L113](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/pyproject.toml#L113), [L144](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/pyproject.toml#L144), [L190](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/pyproject.toml#L190)) while `sglang` deliberately stays at `5.6.0` — that split is the most surprising part of the change and is exactly what a design doc is for.

### AI-1

```suggestion
| Base `transformers` range | `>=5.5.0,<5.9.0` | `>=5.5.0,<=5.12.1` |
| Automodel extra | `transformers>=5.5.0,<5.6.0` | `transformers==5.12.1` |
| `fsdp` / `vllm` / `mcore` extras | No standalone pin | `transformers==5.12.1` |
```

**Context — no action.** Separately, the bullet at [L61-62](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/docs/design-docs/automodel-upgrade-intro.md#L61-L62) ("`BackendConfig` now comes from `components.models.common.utils`") describes no change in this PR — that import path is byte-identical at the base commit (`git show 13b9dae09:nemo_rl/models/policy/__init__.py`). Worth dropping while you're in the file.

nemo_rl/models/automodel/train.py
```diff
@@ -703,45 +714,33 @@ def __call__(
...
-                device_mesh=self.cp_mesh,
-                placements=[Shard(sequence_dim)],
-            )
-
-            logits = redistribute_logits_for_cp(
-                logits, self.device_mesh, self.cp_mesh, sequence_dim
-            )
-
-            token_logprobs = get_logprobs_from_vocab_parallel_logits(
+        if cp_sharder is not None:
```
jQizhang, line 726: **1 action item.**

**TL;DR — this PR rewrote the CP branch of all four post-processors but added CP tests for only two of them, so a layout-restore regression in `LogprobsPostProcessor` (the GRPO/SFT logprob path) or `TopkLogitsPostProcessor` would ship silently.**

[`test_automodel_context_parallel.py`](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/tests/unit/models/automodel/test_automodel_context_parallel.py#L1) exercises `LossPostProcessor` and `FullLogitsPostProcessor` under CP, and covers [`get_cp_sharded_next_token_logprobs`](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/nemo_rl/distributed/model_utils.py#L939) directly — but never reaches these two processors. Every `LogprobsPostProcessor`/`TopkLogitsPostProcessor` call in `test_automodel_train.py` passes `cp_sharder=None` (lines 669, 706, 748, 1784, 1822, 1883).

What that leaves untested is specific, not incidental: this branch reads `data_dict["input_ids"]` rather than `processed_inputs.input_ids` precisely because `_build_model_batch` clones the model-facing tensors under CP — swapping it back would be a silent wrong-answer regression with no test pressure. Nor is the `assert token_logprobs.shape[1] == seq_len - 1`, the zero-prepend, or the `input_lengths` padding mask covered. For [`TopkLogitsPostProcessor`](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/nemo_rl/models/automodel/train.py#L919-L940) the risk is `gather_token_tensor` being applied to `vals` but not `idx`, which silently mis-pairs top-k indices with their values on the way into distillation.

### AI-1

Two tests, both verified passing locally, to add to `tests/unit/models/automodel/test_automodel_context_parallel.py` (where the `_PermutationTokenLayout` fake already lives). Add `LogprobsPostProcessor` and `TopkLogitsPostProcessor` to the existing `nemo_rl.models.automodel.train` import block.

```python
@pytest.mark.automodel
def test_logprobs_postprocessor_cp_matches_cp1_and_masks_padding() -> None:
    """The CP branch must restore canonical order and mask right padding."""
    torch.manual_seed(3)
    order = torch.tensor([0, 3, 1, 2])
    layout = _PermutationTokenLayout(order)
    input_ids = torch.tensor([[1, 2, 3, 4], [5, 6, 7, 8]])
    canonical_logits = torch.randn(2, 4, 9)
    local_logits = canonical_logits.index_select(1, order)

    data_dict = BatchedDataDict(
        {"input_ids": input_ids, "input_lengths": torch.tensor([4, 2])}
    )
    processed_inputs = ProcessedInputs(
        input_ids=input_ids,
        seq_len=4,
        attention_mask=torch.ones(2, 4, dtype=torch.bool),
        position_ids=torch.arange(4).repeat(2, 1),
    )
    processor = LogprobsPostProcessor(cfg={})
    kwargs = dict(
        data_dict=data_dict,
        processed_inputs=processed_inputs,
        original_batch_size=2,
        original_seq_len=4,
    )

    cp = processor(logits=local_logits, cp_sharder=layout, **kwargs)
    cp1 = processor(logits=canonical_logits, cp_sharder=None, **kwargs)

    torch.testing.assert_close(cp, cp1)
    assert cp.shape == (2, 4)
    torch.testing.assert_close(cp[1, 2:], torch.zeros(2))


@pytest.mark.automodel
def test_topk_postprocessor_cp_matches_cp1() -> None:
    """vals and idx must both be restored to canonical order."""
    torch.manual_seed(5)
    order = torch.tensor([0, 3, 1, 2])
    layout = _PermutationTokenLayout(order)
    input_ids = torch.tensor([[1, 2, 3, 4]])
    canonical_logits = torch.randn(1, 4, 9)
    local_logits = canonical_logits.index_select(1, order)

    tp_mesh = MagicMock()
    tp_mesh.get_group.return_value = object()
    processor = TopkLogitsPostProcessor(cfg={}, tp_mesh=tp_mesh, k=3)
    kwargs = dict(
        data_dict=BatchedDataDict(
            {"input_ids": input_ids, "input_lengths": torch.tensor([4])}
        ),
        processed_inputs=ProcessedInputs(
            input_ids=input_ids,
            seq_len=4,
            attention_mask=torch.ones(1, 4, dtype=torch.bool),
            position_ids=torch.arange(4).unsqueeze(0),
        ),
        original_batch_size=1,
        original_seq_len=4,
    )

    def fake_topk(logits, *, k, tp_group, vocab_start_index, vocab_end_index):
        vals, idx = torch.topk(logits, k=k, dim=-1)
        return vals, idx + vocab_start_index

    with (
        patch("nemo_rl.models.automodel.train.distributed_vocab_topk", fake_topk),
        patch("torch.distributed.get_rank", return_value=0),
    ):
        cp_vals, cp_idx = processor(logits=local_logits, cp_sharder=layout, **kwargs)
        cp1_vals, cp1_idx = processor(
            logits=canonical_logits, cp_sharder=None, **kwargs
        )

    torch.testing.assert_close(cp_vals, cp1_vals)
    torch.testing.assert_close(cp_idx, cp1_idx)
```

The first was mutation-checked: feeding the CP-permuted logits without the layout restore produces a different result, so it fails if the branch regresses.

nemo_rl/distributed/model_utils.py
```diff
@@ -1862,6 +2009,7 @@ def get_distillation_topk_logprobs_from_logits(
     vocab_parallel_rank: Optional[int] = None,
     vocab_parallel_group: Optional[torch.distributed.ProcessGroup] = None,
     context_parallel_group: Optional[torch.distributed.ProcessGroup] = None,
+    cp_sharder: Optional["ContextParallelSharder"] = None,
 ):
     """Compute top-k log probabilities from logits."""
```
jQizhang, line 2014: **1 action item.**

Nine parameters and a one-line docstring, with an undocumented 3-tuple return. The non-obvious part is that `cp_sharder` doesn't compose with `context_parallel_group` — it *neutralizes* it: [`cp_group = None if cp_sharder is not None else context_parallel_group`](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/nemo_rl/distributed/model_utils.py#L2077), which also skips the DTensor mesh-derived CP branch below it. Passing both silently ignores one of them, and nothing currently says so.

### AI-1

One way to do it — shapes verified against the code (the next-token shift is applied *after* the CP gather at [L2181-2186](https://github.com/NVIDIA-NeMo/RL/blob/7b5e3a0fb5631e5d3ded5584c57c5d2437717dc6/nemo_rl/distributed/model_utils.py#L2181-L2186), so the return matches the non-CP path):

```suggestion
    """Compute top-k log probabilities from logits.

    Args:
        cp_sharder: Automodel ``ContextParallelSharder`` owning this forward's
            sequence layout (V2 automodel worker with cp_size > 1). When set,
            ``student_logits`` is this rank's CP-local shard while
            ``teacher_topk_indices`` stays canonical, and the student top-k
            log-probs and entropy are gathered back to canonical order before
            the shared next-token shift, so the return shape is unchanged from
            the non-CP path. Mutually exclusive with ``context_parallel_group``:
            a sharder neutralizes that group and the DTensor mesh-derived CP
            branch, since the two layout mechanisms are not additive.

    Returns:
        ``(student_topk_logprobs, teacher_topk_logprobs, H_all)`` with shapes
        ``[B, S-1, K]``, ``[B, S-1, K]``, and ``[B, S-1]`` (``H_all`` is None
        unless ``calculate_entropy``).
    """
```

tests/unit/models/automodel/test_automodel_context_parallel.py
```diff
@@ -0,0 +1,431 @@
...
+except ImportError:
+    pytest.skip("nemo_automodel not available", allow_module_level=True)
+
+from nemo_automodel.components.distributed.context_parallel.sharder import (
+    ContextParallelSharder,
+    ShardLayout,
+    round_robin_local_indices,
+    shard_batch_identity,
+)
+from nemo_rl.algorithms.loss.interfaces import LossInputType
```
jQizhang, line 40: **1 action item.** `pre-commit run --all-files` rewrites this file — ruff's import-sort wants a blank line between the `nemo_automodel` and `nemo_rl` import blocks. Everything else in the hook suite is green.

### AI-1

```suggestion

from nemo_rl.algorithms.loss.interfaces import LossInputType
```

jQizhang (commented):
Team review of the Automodel r0.6.0 bump — 6 agents (RL expert, Automodel submodule expert, bug finder, design reviewer, test agent, devil's advocate), with every finding put through an adversarial verification pass before posting.

Two things worth calling out positively, because they are the parts most likely to be undone by a later change:

- **The CP sharder seam is the right boundary.** `ProcessedInputs` no longer carries `cp_buffers`/`seq_index`, and `PreparedModelForward` holds the per-microbatch distributed state instead — data prep no longer knows about parallelism. And because the seam NeMo-RL depends on is two duck-typed verbs (`shard_token_tensor`/`gather_token_tensor`) rather than the concrete `ContextParallelSharder`, the CP math is now testable on CPU with a ~20-line fake. The code it replaced built `DTensor.from_local(..., device_mesh=cp_mesh)` and was only reachable with a live mesh. That testability win is worth protecting.
- **The deleted-test audit came back clean.** All 16 removed tests were traced to APIs that no longer exist at HEAD, per-file counts reconcile, and both deletions that looked like they might be dropping a guard turned out not to be — the VLM+CP guard survives at `setup.py:592-595` and is still covered, and the KV-sharing tripwire fired exactly as its own contract specified.

Independent verification worth recording: the CP shard/restore round trip and both gradient-fanout branches were validated numerically on 2×H100 (max gradient error 0.0, with a sequence length deliberately not divisible by `2*cp` so CP padding is exercised), and every migrated Automodel API was checked signature-by-signature against the pinned submodule SHA.

The comments below are all low-to-moderate severity; nothing here blocks on correctness.

_Generated by Claude Code_

nemo_rl/models/automodel/checkpoint.py
```diff
@@ -217,6 +237,12 @@ def update_checkpointer_config(
             if k == "model_save_format":
                 # Ensure enum type
                 v = SerializationFormat[v.upper()] if isinstance(v, str) else v
+            elif k == "save_consolidated":
+                # Automodel normalizes legacy bools/strings to SaveConsolidatedMode in
+                # CheckpointingConfig.__post_init__, which setattr bypasses. Without this
+                # a bool True would silently compare unequal to SaveConsolidatedMode.EVERY
+                # and disable consolidated HF export.
+                v = _normalize_supported_save_consolidated(v)
```
terrykong, line 249: [`nemo_rl/models/automodel/checkpoint.py:240-245`](https://github.com/NVIDIA-NeMo/RL/blob/8041224e00e03aba88fd1c05481666f024ddbb62/nemo_rl/models/automodel/checkpoint.py#L240-L245)

**1 action item.**

**TL;DR — with `save_consolidated: true`, checkpoint consolidation runs its `all_gather_object` on the default NCCL group from a background thread while the training loop is using that same communicator, so the run hangs or silently corrupts a reduction.** PR-introduced: this `setattr` branch is what makes the flag take effect at all.

**[Illustrated walkthrough of this bug](https://terrykong.github.io/gh-pages-poc/terryk/pr-3498-consolidated-save-hang.html)** — one page: how the two threads end up on the same channel, where the setting gets lost, and the two ways to fix it.


Background: every save, Automodel can also write consolidated HF safetensors. That needs a collective across ranks. r0.6.0 runs it on a **separate Gloo group** so it cannot collide with training traffic — and it decides whether to create that group **once, in `Checkpointer.__init__`**, from `config.save_consolidated`:

```python
and self.config.save_consolidated != SaveConsolidatedMode.FALSE
```
[`checkpointing.py:551-563`](https://github.com/NVIDIA-NeMo/Automodel/blob/1814c6c93a66b9d59d254960ef6a99a64249b671/nemo_automodel/components/checkpoint/checkpointing.py#L551-L563)

How it goes wrong here:

1. The worker builds the `Checkpointer` before it knows `save_consolidated`, so it defaults to `False` -> `FALSE` ([`checkpoint.py:189-191`](https://github.com/NVIDIA-NeMo/RL/blob/8041224e00e03aba88fd1c05481666f024ddbb62/nemo_rl/models/automodel/checkpoint.py#L189-L191)) and `_consolidation_process_group` stays `None`.
2. The real value arrives later, here — **this is the step that succeeds when it shouldn't.** The config now says "consolidate" while the object built for `FALSE` has no group. [`_rebuild_checkpointer_addons`](https://github.com/NVIDIA-NeMo/RL/blob/8041224e00e03aba88fd1c05481666f024ddbb62/nemo_rl/models/automodel/checkpoint.py#L250) rebuilds `_addons` only, never the group.
3. Automodel falls back to `self.process_group`, and NeMo-RL never passes one ([`checkpoint.py:201-207`](https://github.com/NVIDIA-NeMo/RL/blob/8041224e00e03aba88fd1c05481666f024ddbb62/nemo_rl/models/automodel/checkpoint.py#L201-L207)) -> the **default NCCL group**: `consolidation_process_group = self._consolidation_process_group if ... is not None else self.process_group` ([`checkpointing.py:618-620`](https://github.com/NVIDIA-NeMo/Automodel/blob/1814c6c93a66b9d59d254960ef6a99a64249b671/nemo_automodel/components/checkpoint/checkpointing.py#L618-L620)).
4. `is_async=True` is unconditional for this worker ([`dtensor_policy_worker_v2.py:292-299`](https://github.com/NVIDIA-NeMo/RL/blob/8041224e00e03aba88fd1c05481666f024ddbb62/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py#L292-L299)), so consolidation is deferred onto a **daemon thread** ([`checkpointing.py:1331-1333`](https://github.com/NVIDIA-NeMo/Automodel/blob/1814c6c93a66b9d59d254960ef6a99a64249b671/nemo_automodel/components/checkpoint/checkpointing.py#L1331-L1333)) — whose own docstring says collectives "run on the dedicated Gloo group created at init".
5. That thread calls `dist.all_gather_object(failures, local_failure, group=process_group)` ([`consolidate_hf_safetensors.py:966-969`](https://github.com/NVIDIA-NeMo/Automodel/blob/1814c6c93a66b9d59d254960ef6a99a64249b671/nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py#L966-L969)) while the main thread issues the next step's collectives on the same communicator. Nothing joins the thread first — `finalize_async_save` only runs at promotion time, many collectives later.

Symptom: mismatched collective order on one NCCL communicator — a hang ending in the torch watchdog's `Watchdog caught collective operation timeout`, or a silently wrong reduction. Not verified on hardware (no GPU here); the mechanism is read from the pinned sources above.

This is new with the bump: at the old pin `24b47e85`, `save_model` gated all-ranks consolidation on `not is_async`, and NeMo-RL always sets `is_async=True`, so no thread and no group existed. It is also uncovered — all 20 `save_consolidated` occurrences under `examples/` and `tests/unit/reference_configs/` are `false`, so this bites the first user who turns the documented knob on.

### AI-1

Rebuild the `Checkpointer` (or create the Gloo group) when `save_consolidated` transitions off `FALSE`, rather than `setattr`-ing over an object built for `FALSE`. Every rank runs the same update, so creation order stays consistent. Not a `suggestion` block — the fix spans this loop plus the group setup, which is more than one region.

The cheapest safe version, matching the pattern you already use two lines up: have `_normalize_supported_save_consolidated` reject `true` with a clear message the same way it rejects `final`, until the group is wired.

Two follow-ons if you take this:
- `_normalize_supported_save_consolidated` and this branch have no test (`grep -rn "_normalize_supported_save_consolidated\|SaveConsolidatedMode" tests/` returns nothing). The `True` -> `SaveConsolidatedMode.EVERY` mapping is worth pinning, since your own comment says getting it wrong disables HF export **silently**.
- §2.3 of the new `automodel-context-parallel.md` says NeMo-RL normalizes with `_normalize_save_consolidated()` after `setattr`. That is incomplete for the same reason — the consolidation group also has to be rebuilt.

docs/design-docs/automodel-upgrade-intro.md
terrykong: this file is great btw, but do we need to check it in? would it be sufficient to just have this as the pr description?

terrykong: assumign this gets cleaned up, just a note to clean up the pngs too

terrykong (commented):
Re-review at `8041224e`, after the rebase. One finding left, inline on `checkpoint.py`.

**The one thing to look at** is `save_consolidated: true` pushing checkpoint-consolidation collectives onto the training NCCL channel from a background thread. It is not reachable by anything shipped — every `save_consolidated` in the repo is `false` — but it is new with this bump, and it is a hang rather than an error. There is an [illustrated walkthrough](https://terrykong.github.io/gh-pages-poc/terryk/pr-3498-consolidated-save-hang.html) if the mechanism is easier to follow as a picture.

Two small leftovers, neither worth its own thread:

- The PR description still says the Automodel submodule goes to `7e9493d7`. It is now pinned at `1814c6c93a66b9d59d254960ef6a99a64249b671`, so anyone auditing the API migration against the description reads the wrong tree.
- `pre-commit` still rewrites `tests/unit/models/automodel/test_automodel_context_parallel.py` — ruff wants a blank line before the `nemo_rl` import block.

Everything else in the earlier pass either got fixed in the rebase or did not survive a second look. Worth recording what held up, since these are the parts most likely to be undone later:

- **Both workaround tripwires fired exactly as designed.** The token-classification shim and the fp32 dtype patch were each written with a self-deleting assertion, and this bump removed both cleanly. That is the pattern to repeat for the next workaround pinned to an Automodel version.
- **The deleted-test audit came back clean.** Every removed test traces to an API that no longer exists, and each guard worth worrying about kept its coverage — VLM+CP, CP+sequence-packing, CP+sequence-parallel and Gemma3 all still have tests, and the reward-model/critic CP=1 case gained coverage.
- **The gradient-fanout change is correct.** `gather_token_tensor` backs onto an all-gather whose backward is a summing reduce-scatter, so a replicated CP loss applies each rank's gradient `cp_size` times, and dividing by `cp_gradient_fanout` cancels it exactly — while the x-token path correctly keeps fanout 1, since its KD windows are disjoint and its CE uses an identity-backward sum. Comparing CP1 against CP2 and CP4 is precisely the run that would separate if that divisor were wrong, so the results linked from the description are the right evidence for it.

Bundling the dependency bump with the CP refactor is the right call rather than a problem — r0.6.0 replaced the CP API, so splitting them would leave `main` unbuildable in between.

_Generated by Claude Code_

terrykong (dismissed):

terrykong (dismissed):

sharonyu-115 (dismissed):
The test case change looks good. Approve.

terrykong (approved):

PR #3499, opened 2026-08-05
feat(sc): overlap NeMo-Gym spinup with deferred vLLM load

yuki-97:
## Summary
- Parallel NeMo-Gym init on the SC path, mirroring grpo.py: reserve vLLM URLs up-front (via `defer_model_load=True`) so NeMo-Gym spins up in parallel with the vLLM weight load and trainer build.
- Extract `SetupTimingMetrics` dataclass + `print_setup_timing_summary` into `nemo_rl.algorithms.metric_utils`, shared by SC and grpo.
- SC uses backend-agnostic `generation_init_time_s` (with `generation_init_reserve_time_s` / `generation_init_load_time_s` split populated when the NeMo-Gym overlap path runs); `print_setup_timing_summary` renders the split as `Generation init: X.Xs (reserve X.Xs + load X.Xs)`. grpo.py keeps the existing per-backend `<backend>_init_time_s` fields.
- SC now emits `worker_setup_time_s` and drops `parallel_wall_time_s` / `parallel_init_enabled` (grpo-only).
- Fix `vllm_init_time_s` under-count on grpo.py's NeMo-Gym path: the deferred `VllmGeneration.__init__` elapsed (worker-group spawn + port bind, no weight load) was previously discarded, so `vllm_init_time_s` on gym-on runs only counted `load_and_start`. Now folded in.
- Update `grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2`'s threshold since that value was tested by non-streaming instead of streaming 2 by mistake in https://github.com/NVIDIA-NeMo/RL/pull/3266. w/ the fix in https://github.com/NVIDIA-NeMo/RL/pull/3282, `mean(data["train/grad_norm"], 2, 0)` is `0.07307618069979879`; w/o the fix, it's `0.04454974799106518`, so update to `> 0.06`.

## Test plan
- [x] `uv run pytest -q tests/unit/single_controller/`
- [x] SC nightlies:
  - `grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2`
  - `grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync`
- [x] `tests/functional/grpo_async_gym_single_controller.sh`

## Worker Initialization Timing (before / after)

From `tests/functional/grpo_async_gym_single_controller.sh`. Current `main` doesn't print the SC setup-timing summary at all, so the baseline is captured on [`90fa6c06d`](https://github.com/NVIDIA-NeMo/RL/commit/90fa6c06da00e26b10335166a8bad9e0e86f3389) (the earliest commit that had SC timing metrics but not this PR's deferred-vLLM overlap) with a small typo fix on top, not `main` HEAD.

**Baseline (`90fa6c06d` + typo fix):**
```
▶ Worker Initialization Timing:
  vllm init: 32.4s
  Policy init: 38.2s
  NeMo-Gym init: 11.8s
  Other setup: 1.9s
  Total setup: 51.9s
```

**Branch (`yukih/sc-parallel-init`):**
```
▶ Worker Initialization Timing:
  Generation init: 32.9s (reserve 11.0s + load 21.9s)
  Policy init: 36.3s
  NeMo-Gym init: 11.9s
  Other setup: 1.7s
  Total setup: 49.0s
```

On this smoke config total wall-time is 51.9s → 49.0s (~3s), and the new `reserve/load` split makes the regime visible: `reserve_time (11.0s) ≈ nemo_gym_init (11.9s)`. Reserve is serial by design (need vLLM URLs before spinning gym), so once gym is only that long the parallel executor has essentially nothing to absorb; the delta here is at the noise level. Reassuring that `reserve + load (32.9s) ≈ baseline vllm init (32.4s)`, so the deferred flow adds no measurable overhead. Once NeMo-Gym init grows in heavier setups (multi-resource-server / agent+tool prefetch / large dataset warmup) so gym > reserve, the parallel-init win becomes visible; this smoke just doesn't exercise that regime.

### DIFF
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -455,6 +455,12 @@ def setup_single_controller(
     train_cluster, inference_cluster = _build_clusters(master_config)
     colocated = generation_config["colocated"]["enabled"]
 
+    # Create build tasks for generation / trainer / (nemo-gym) workers
+    build_tasks: dict[str, Callable[[], Any]] = {}
+    generation = None
+    defer_generation_model_load = False
+    gen_reserve_time = 0.0
+
     def _build_generation_then_trainer(
         defer_generation_model_load: bool, generation=None
     ) -> tuple[Any, Any, dict[str, float]]:
@@ -488,13 +494,9 @@ def setup_single_controller(
 
         return generation, trainer, time_metrics
 
-    # Create build tasks for generation / trainer / (nemo-gym) workers
-    build_tasks: dict[str, Callable[[], Any]] = {}
-    generation = None
-    defer_generation_model_load = False
     if use_nemo_gym:
         # defer generation, only get base_urls for nemo_gym spinup
-        generation, _ = _build_generation(
+        generation, gen_reserve_time = _build_generation(
             inference_cluster,
             master_config=master_config,
             defer_model_load=True,
@@ -546,12 +548,18 @@ def setup_single_controller(
 
     if colocated:
         generation, trainer, time_metrics = results["generation_trainer"]
-        setattr(setup_timing_metrics, gen_init_time_key, time_metrics["gen_time"])
+        setattr(
+            setup_timing_metrics,
+            gen_init_time_key,
+            time_metrics["gen_time"] + gen_reserve_time,
+        )
         setup_timing_metrics.policy_init_time_s = time_metrics["trainer_time"]
     else:
-        generation, gen_time = results["generation"]
+        generation, gen_load_time = results["generation"]
         trainer, trainer_time = results["trainer"]
-        setattr(setup_timing_metrics, gen_init_time_key, gen_time)
+        setattr(
+            setup_timing_metrics, gen_init_time_key, gen_load_time + gen_reserve_time
+        )
         setup_timing_metrics.policy_init_time_s = trainer_time
 
     worker_setup_time = time.perf_counter() - setup_start_time
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -464,7 +464,8 @@ def setup_single_controller(
         """Build generation then trainer serially.
 
         Args:
-            defer_generation_model_load: If True, `generation` is a pre-reserved handle and this call finishes its model load; if False, builds generation from scratch.
+            defer_generation_model_load: If True, generation is a pre-reserved handle and this call
+                finishes its model load; if False, builds generation from scratch.
             generation: Pre-reserved generation handle when defer_generation_model_load=True; None otherwise.
 
         Returns:
@@ -534,6 +535,8 @@ def setup_single_controller(
             _build_trainer,
             train_cluster=train_cluster,
             master_config=master_config,
+            tokenizer=tokenizer,
+            processor=processor,
         )
 
     # Submit build tasks and get results
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -25,7 +25,7 @@ import time
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from functools import partial
-from typing import Any, Optional, cast
+from typing import Any, Callable, Optional, cast
 
 from torchdata.stateful_dataloader import StatefulDataLoader
 from transformers import AutoProcessor
@@ -492,7 +492,7 @@ def setup_single_controller(
         return generation, trainer, time_metrics
 
     # Create build tasks for generation / trainer / (nemo-gym) workers
-    build_tasks = {}
+    build_tasks: dict[str, Callable[[], Any]] = {}
     generation = None
     defer_generation_model_load = False
     if use_nemo_gym:
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -181,8 +181,8 @@ def _build_generation(
         defer_model_load: If True (for the NeMo-Gym flow), reserve OpenAI server URLs without loading weights; caller runs gen.load_and_start() later.
 
     Returns:
-        gen: The generation object (VllmGeneration or SGLangGeneration).
-        elapsed_s: Wall time spent in this call.
+        A tuple of (generation object, wall time spent in this call). The
+        generation object is a VllmGeneration or SGLangGeneration.
     """
     t0 = time.perf_counter()
     generation_config = master_config.policy["generation"]
@@ -232,8 +232,7 @@ def _finish_deferred_generation(generation: Any) -> tuple[Any, float]:
         generation: The deferred generation object.
 
     Returns:
-        generation: The finished generation object.
-        elapsed_s: Wall time spent in this call.
+        A tuple of (finished generation object, wall time spent in this call).
     """
     t0 = time.perf_counter()
     generation.load_and_start()
@@ -256,8 +255,7 @@ def _build_trainer(
         processor: Optional AutoProcessor for VLM paths.
 
     Returns:
-        trainer: The TQPolicy trainer.
-        elapsed_s: Wall time spent in this call.
+        A tuple of (TQPolicy trainer, wall time spent in this call).
     """
     t0 = time.perf_counter()
     loss_config = master_config.loss_fn
@@ -284,8 +282,7 @@ def _spinup_gym(master_config: MasterConfig, base_urls: list[str]) -> tuple[Any,
         base_urls: Reserved vLLM OpenAI server URLs.
 
     Returns:
-        actor: The NeMo-Gym actor.
-        elapsed_s: Wall time spent in this call.
+        A tuple of (NeMo-Gym actor, wall time spent in this call).
     """
     t0 = time.perf_counter()
     policy_config = master_config.policy
@@ -363,8 +360,8 @@ def setup_single_controller(
         partition_id: TQ partition the rollout writer + sampler share.
 
     Returns:
-        actor_args: Pre-built SC actor args.
-        setup_timing_metrics: Driver-side per-phase timings, logged by the SC actor.
+        A tuple of (pre-built SC actor args, driver-side per-phase timings
+        logged by the SC actor).
     """
     validate_single_controller_config(master_config)
 
@@ -469,9 +466,8 @@ def setup_single_controller(
             generation: Pre-reserved generation handle when defer_generation_model_load=True; None otherwise.
 
         Returns:
-            generation: The finalized generation object.
-            trainer: The TQPolicy trainer.
-            time_metrics: Per-phase wall times keyed as "gen_time" and "trainer_time".
+            A tuple of (finalized generation object, TQPolicy trainer,
+            per-phase wall times keyed as "gen_time" and "trainer_time").
         """
         time_metrics = {}
 
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -446,7 +446,13 @@ def setup_single_controller(
     setup_start_time = time.perf_counter()
     setup_timing_metrics = SetupTimingMetrics()
     gen_backend = generation_config["backend"]
-    gen_init_time_key = f"{gen_backend}_init_time_s"
+    # TODO: SC doesn't support megatron generation yet; the megatron branch is
+    # defensive — mirrors grpo.py so the metric key stays valid if we add it.
+    gen_init_time_key = (
+        "megatron_generation_init_time_s"
+        if gen_backend == "megatron"
+        else f"{gen_backend}_init_time_s"
+    )
 
     # Create clusters
     train_cluster, inference_cluster = _build_clusters(master_config)
@@ -562,7 +568,7 @@ def setup_single_controller(
     weight_synchronizer = create_weight_synchronizer(
         policy=trainer,
         generation=generation,
-        gen_backend=gen_backend,
+        generation_backend=gen_backend,
         colocated=colocated,
         train_cluster=train_cluster,
         inference_cluster=inference_cluster,
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -38,6 +38,10 @@ from nemo_rl.algorithms.grpo import (
 )
 from nemo_rl.algorithms.loss import ClippedPGLossFn
 from nemo_rl.algorithms.loss.interfaces import LossFunction
+from nemo_rl.algorithms.metric_utils import (
+    SetupTimingMetrics,
+    print_setup_timing_summary,
+)
 from nemo_rl.algorithms.single_controller_utils.config import (
     MasterConfig,
     validate_single_controller_config,
@@ -268,7 +272,7 @@ def setup_single_controller(
     *,
     processor: Optional[AutoProcessor] = None,
     partition_id: str = "rollout_data",
-) -> tuple[SingleControllerActorArgs, dict[str, Any]]:
+) -> tuple[SingleControllerActorArgs, SetupTimingMetrics]:
     """Build the full SC actor args driver-side.
 
     Args:
@@ -359,7 +363,7 @@ def setup_single_controller(
     # Setup Clusters & Workers
     # ==========================
     setup_start_time = time.perf_counter()
-    setup_timing_metrics: dict[str, Any] = {}
+    setup_timing_metrics = SetupTimingMetrics()
     gen_backend = generation_config["backend"]
     gen_init_time_key = f"{gen_backend}_init_time_s"
 
@@ -381,9 +385,9 @@ def setup_single_controller(
         # comes up before the policy.
         generation, gen_time = _timed_build_generation()
         policy, policy_time = _timed_build_trainer()
-        setup_timing_metrics[gen_init_time_key] = gen_time
-        setup_timing_metrics["policy_init_time_s"] = policy_time
-        setup_timing_metrics["parallel_init_enabled"] = 0.0
+        setattr(setup_timing_metrics, gen_init_time_key, gen_time)
+        setup_timing_metrics.policy_init_time_s = policy_time
+        setup_timing_metrics.parallel_init_enabled = 0.0
     else:
         # Non-colocated: generation + policy run on disjoint GPUs, so
         # bring them up in parallel.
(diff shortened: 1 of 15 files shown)


tests/unit/single_controller/test_single_controller_setup.py
```diff
@@ -423,20 +418,134 @@ def test_nemo_gym_wires_env_handle(self, patched_factories):
...
+        with (
+            patch.object(sc_setup_mod, "_should_use_nemo_gym", return_value=True),
+            patch.object(
+                sc_setup_mod, "spinup_nemo_gym_actor", return_value=MagicMock()
+            ),
+            patch.object(sc_setup_mod, "router_replay_enabled", return_value=False),
+        ):
+            _, metrics = setup_single_controller(mc, MagicMock(pad_token_id=0))
+
+        assert metrics.nemo_gym_init_time_s is not None
```
terrykong: [`test_single_controller_setup.py:529`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller_setup.py#L529)

**1 action item, 2 follow-ups.**

### AI-1 — the three-task fan-out has no unit coverage

`build_tasks` has four possible shapes. Three are covered; the one that isn't is the PR's headline feature:

| colocated | gym | `build_tasks` | covered by |
|---|---|---|---|
| yes | no | `{generation_trainer}` | [`:441`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller_setup.py#L441) |
| no | no | `{generation, trainer}` | [`:457`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller_setup.py#L457) |
| yes | yes | `{nemo_gym, generation_trainer}` | [`:477`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller_setup.py#L477), [`:503`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller_setup.py#L503) |
| no | yes | `{nemo_gym, generation, trainer}` | **nothing** |

That last row is the only path reaching [`build_tasks["generation"] = partial(_finish_deferred_generation, ...)`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L523) — and it's the config the functional test actually runs ([`grpo_async_gym_single_controller.sh:72`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/functional/grpo_async_gym_single_controller.sh#L72) sets `policy.generation.colocated.enabled=false`). So the headline path's only coverage today is a GPU functional test.

**Action:** this suggestion replaces `test_setup_timing_includes_nemo_gym_when_enabled` (see Follow-up A) with a `colocated=False` version. Untested locally — please run it.

```suggestion
    def test_nemo_gym_noncolocated_finishes_deferred_load(self, patched_factories):
        """Non-colocated + gym fans out gym / deferred-load / trainer together."""
        mc = _make_master_config(colocated=False, backend="vllm")
        mc.policy["generation"]["model_name"] = "test-model"
        mc.policy["generation"]["stop_strings"] = None
        mc.policy["generation"]["stop_token_ids"] = None
        mc.policy["generation"]["top_k"] = None
        patched_factories["setup_response_data"].return_value = (list(range(8)), None)

        with (
            patch.object(sc_setup_mod, "_should_use_nemo_gym", return_value=True),
            patch.object(
                sc_setup_mod, "spinup_nemo_gym_actor", return_value=MagicMock()
            ),
            patch.object(sc_setup_mod, "router_replay_enabled", return_value=False),
        ):
            actor_args, metrics = setup_single_controller(mc, MagicMock(pad_token_id=0))

        # _build_generation runs once (URL reservation only); the load is finished
        # by _finish_deferred_generation inside the executor.
        patched_factories["_build_generation"].assert_called_once()
        _, gen_kwargs = patched_factories["_build_generation"].call_args
        assert gen_kwargs.get("defer_model_load") is True
        patched_factories["fake_gen"].load_and_start.assert_called_once_with()
        assert actor_args.gen_handle is patched_factories["fake_gen"]
        assert metrics.nemo_gym_init_time_s is not None
        assert metrics.vllm_init_time_s is not None
        assert metrics.policy_init_time_s is not None
```

### Follow-up A — this test is redundant

`test_setup_timing_includes_nemo_gym_when_enabled` ([`:529-547`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller_setup.py#L529)) has byte-identical setup to `test_nemo_gym_records_timing_metrics` ([`:503`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller_setup.py#L503)), and its entire body is `assert metrics.nemo_gym_init_time_s is not None` — verbatim [`:521`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller_setup.py#L521). Hence replacing rather than adding.

### Follow-up B — `metric_utils.py` has no direct test

The closest thing, [`test_logs_setup_timing_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tests/unit/single_controller/test_single_controller.py#L171), asserts `log_metrics.assert_called_once_with(setup_metrics.to_metrics_dict(), ...)` — it calls the method under test to build its own expectation, so it pins the wiring but cannot catch a `to_metrics_dict` regression. A pure-CPU file under `tests/unit/algorithms/` covering the None-drop, `0.0` surviving the filter (`parallel_init_enabled=0.0` must not be dropped), and the `extras` merge would be cheap. Not drafting it here since I couldn't run it.

**Context — no action.** Marks/placement are right: `tests/unit/single_controller/` is swept by `L0_Unit_Tests_Other.sh`, and no `vllm` mark is needed since `VllmGeneration` is never constructed.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -185,30 +213,56 @@ def _build_generation(
...
+    Args:
+        train_cluster: Ray virtual cluster the trainer workers run on.
+        master_config: SC MasterConfig.
+        tokenizer: Tokenizer used by the policy.
+        processor: Optional AutoProcessor for VLM paths.
+
+    Returns:
+        trainer: The TQPolicy trainer.
+        elapsed_s: Wall time spent in this call.
     """
```
terrykong, line 259: [`setup.py:252`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L252)

Adding `Args:`/`Returns:` here is a clear improvement, but it replaced a load-bearing paragraph that was on `main`:

> Driver-side on purpose: instantiating TQPolicy inside another Ray actor nests runtime_envs and triggers Ray's `get_accelerator_ids_for_accelerator_resource` IndexError. Keep this here until PolicyTrainerActor (PR #2692) lands.

That's a named crash, a named cause, and a named expiry condition — and PR #2692 is precisely the change that would move this function into an actor. After this PR the file gives the next person no reason not to.

**Action:** keep the new `Args:`/`Returns:` block and restore the rationale above it.

```suggestion
    Driver-side on purpose: instantiating TQPolicy inside another Ray
    actor nests runtime_envs and triggers Ray's
    get_accelerator_ids_for_accelerator_resource IndexError. Keep this
    here until PolicyTrainerActor (PR #2692) lands.

    Args:
        train_cluster: Ray virtual cluster the trainer workers run on.
        master_config: SC MasterConfig.
        tokenizer: Tokenizer used by the policy.
        processor: Optional AutoProcessor for VLM paths.

    Returns:
        trainer: The TQPolicy trainer.
        elapsed_s: Wall time spent in this call.
    """
```

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -443,9 +608,16 @@ def setup_single_controller(
         tq_buffer=tq_buffer,
     )
 
-    return SingleControllerActorArgs(
+    # Print setup timing metrics
+    total_setup_time = time.perf_counter() - setup_start_time
+    setup_timing_metrics.total_setup_time_s = total_setup_time
+    setup_timing_metrics.other_setup_time_s = total_setup_time - worker_setup_time
+    print_setup_timing_summary(setup_timing_metrics, gen_init_time_key)
```
terrykong: [`setup.py:615`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L615)

**1 action item.**

The PR's whole thesis is that NeMo-Gym spinup now overlaps the vLLM weight load instead of running after it — but there's no recorded number for the win anywhere on the PR (comments are copy-pr-bot plus four `/ok to test`, and there are no reviews yet).

Your test plan shows both SC nightlies and the functional test already ran, so this should be near-free: could you paste the `> Worker Initialization Timing` block from a `main` run and a branch run? Specifically `nemo_gym_init_time_s`, `vllm_init_time_s`, and `worker_setup_time_s` — that last one is the metric this PR adds, and it's exactly the one that would show the overlap.

**Action:** attach the before/after timing block (or a W&B link) to the PR description.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -356,64 +443,142 @@ def setup_single_controller(
...
         generation=generation,
-        generation_backend=backend,
+        generation_backend=gen_backend,
         colocated=colocated,
         train_cluster=train_cluster,
         inference_cluster=inference_cluster,
         refit_buffer_size_gb=policy_config.get("refit_buffer_size_gb"),
     )
     weight_synchronizer.init_communicator()
+    setup_timing_metrics.collective_init_time_s = time.perf_counter() - t0
```
terrykong, line 574: [`setup.py:581`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L581)

**1 action item.** `vllm_init_time_s` measures a different span with NeMo-Gym than without it. Pre-existing in grpo, newly inherited by SC.

### Why the two spans differ

[`generation, _ = _build_generation(..., defer_model_load=True)`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L501) throws away its elapsed time. That call isn't a cheap handle-grab — `VllmGeneration.__init__` builds the whole worker group (actor spawn, per-node vLLM venv resolution, placement groups, port bind). Only `load_model` is deferred, per [`vllm_generation.py:482-493`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/models/generation/vllm/vllm_generation.py#L482) — *"Workers have bound ports during `__init__`"*.

So the two paths time different things:

| | what `vllm_init_time_s` covers |
|---|---|
| gym **off** — [`_build_generation` at `:529`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L529) | worker-group construction **+** port bind **+** weight load |
| gym **on** — [`_finish_deferred_generation` at `:524`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L524) | weight load only |

### Worked example

Illustrative numbers — I don't have a GPU environment, so these are the arithmetic implied by the two code paths, not a measured run. Say worker-group construction takes **40s** and the weight load **90s** on a given recipe:

```
gym OFF:  vllm_init_time_s = 40 + 90 = 130s
gym ON:   vllm_init_time_s =      90 =  90s     # the 40s went into `_`
```

Two consequences:

1. **Enabling NeMo-Gym looks like it made vLLM init 31% faster.** It didn't — identical physical work, 40s of it just stopped being counted. That's awkward for this PR specifically, since `vllm_init_time_s` is one of the series someone would read to judge the overlap win.
2. **The printed summary stops adding up.** `worker_setup_time_s` is measured from [`setup_start_time`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L446) and *does* include the 40s, so `Generation init + Policy init` silently under-sums the worker phase by exactly the discarded amount — with no line in the block accounting for the gap.

### Action

Capture the elapsed at `:501` and fold it into the generation-init metric — into `time_metrics["gen_time"]` on the colocated path and the `gen_time` unpacked at [`:556`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L556) on the non-colocated one — so both paths mean "everything vLLM-side."

Adding it as `extras["vllm_reserve_time_s"]` is a nice extra series and makes the reserve/load split visible, but on its own it leaves the headline metric still meaning two things — so that's worth doing *as well*, not *instead*.

terrykong (commented):
Reviewed with an agent team (5 agents: RL expert, bug finder, test reviewer, design reviewer, devil's advocate). **No correctness defect found** — nice, clean port.

The overlap is a faithful mirror of the already-shipped grpo.py pattern, and I verified the concurrency is safe rather than assuming it: `spinup_nemo_gym_actor` only plumbs `base_urls` into a config dict and never issues HTTP before `load_and_start()`, and [`_reserve_port`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/models/generation/vllm/vllm_worker_async.py#L131) binds *and holds* the socket, so there's no steal window between reservation and uvicorn taking over. SC is actually safer than grpo here — it binds `base_urls` eagerly at `partial()` creation ([`setup.py:511`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/single_controller_utils/setup.py#L511)), whereas grpo reads the attribute inside the worker thread while `load_and_start()` reassigns it.

Also affirming the design, since it came up in review: `SetupTimingMetrics` is the right call. A plain dict gave zero typo-safety and is how `worker_init_timing_metrics` grew a free-form key space; the flat schema with unset-fields-dropped is what lets a dashboard read the same key from a grpo run and an SC run. The `extras` bucket is correctly scoped to genuinely interpolated keys.

Five comments below, all low-severity: one test-coverage gap, one deleted comment worth restoring, a citation fix, an ask for numbers, and a metrics-consistency note.

**Follow-up, not for this PR:** the extraction is 2-of-4 — [`ppo.py:459-689`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/nemo_rl/algorithms/ppo.py#L459) still carries the old `worker_init_timing_metrics` dict *and* a verbatim copy of the print block this PR just extracted (plus a `value_init_time_s` key with no field on the dataclass), and [`tools/refit_verifier.py:651`](https://github.com/NVIDIA-NeMo/RL/blob/a50365b956d04b085ca839ea43306a861e6f4c53/tools/refit_verifier.py#L651) still has the old `init_time_key` signature. Both log to the same `timing/setup` prefix. Worth a tracking issue.

**Caveats on this review:** no GPU here, so the overlap win itself is unverified. The repo linters could not run (macOS host; the uv lockfile only supports linux x86_64/aarch64), so `pre-commit` results are absent — though `metric_utils.py` is correctly added to `pyrefly.toml` `project-includes`. The suggested test was not executed.

_Generated by Claude Code_

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: fixed in 5a7821e922505fb9fa6e3d914d4e8f76dd9dd7984, updated to generation_init_time_s (with explicit reserve/load split fields) in 5f1a32e547807a295102dd3f72bae71a4c08599a

tests/unit/single_controller/test_single_controller_setup.py
yuki-97: fixed in c60ebe6b06f7a9efb3fd073e8978f155d088145e, added metric_utils direct tests (Follow-up B) in 5f1a32e547807a295102dd3f72bae71a4c08599a

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: Sharing the `Worker Initialization Timing` blocks from `tests/functional/grpo_async_gym_single_controller.sh`. Note that current `main` doesn't print the SC setup-timing summary at all, so the "before" side is captured on [`90fa6c06d`](https://github.com/NVIDIA-NeMo/RL/commit/90fa6c06da00e26b10335166a8bad9e0e86f3389) (the earliest commit that had SC timing metrics but not this PR's deferred-vLLM overlap) with a small typo fix on top, not `main` HEAD.

**Baseline (`90fa6c06d` + typo fix):**
```
▶ Worker Initialization Timing:
  vllm init: 32.4s
  Policy init: 38.2s
  NeMo-Gym init: 11.8s
  Other setup: 1.9s
  Total setup: 51.9s
```

**Branch (`yukih/sc-parallel-init`):**
```
▶ Worker Initialization Timing:
  Generation init: 32.9s (reserve 11.0s + load 21.9s)
  Policy init: 36.3s
  NeMo-Gym init: 11.9s
  Other setup: 1.7s
  Total setup: 49.0s
```

On this smoke config total wall-time is 51.9s → 49.0s (~3s), and the new `reserve/load` split makes the regime visible: `reserve_time (11.0s) ≈ nemo_gym_init (11.9s)`. Reserve is serial by design (need vLLM URLs before spinning gym), so once gym is only that long the parallel executor has essentially nothing to absorb; the delta here is at the noise level. Reassuring that `reserve + load (32.9s) ≈ baseline vllm init (32.4s)`, so the deferred flow adds no measurable overhead. Once NeMo-Gym init grows in heavier setups (multi-resource-server / agent+tool prefetch / large dataset warmup) so gym > reserve, the parallel-init win becomes visible; this smoke just doesn't exercise that regime.

Also added the same block to the PR description.

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: we are not planning to use `PolicyTrainerActor` since from test it will cause some issues, so it's intended to delete.

nemo_rl/algorithms/single_controller_utils/setup.py
terrykong: fixed in 5a7821e92505fb9fa6e3d914d4e8f76dd9dd7984 and 5f1a32e547807a295102dd3f72bae71a4c08599a — reserve time is folded in on both the grpo.py and SC paths, and the reserve/load split is exposed too. Thanks for going the extra step with generation_init_time_s.

tests/unit/single_controller/test_single_controller_setup.py
terrykong: fixed in c60ebe6b06f7a9efb3fd073e8978f155d088145e and 5f1a32e547807a295102dd3f72bae71a4c08599a — AI-1 and both follow-ups covered.

nemo_rl/algorithms/single_controller_utils/setup.py
terrykong: Thanks — this is exactly what I was after, and I appreciate you flagging that the baseline had to come from 90fa6c06d rather than main, and that the win is at noise level on this smoke config rather than dressing it up. The reserve (11.0s) ~= nemo_gym_init (11.9s) reading explains why cleanly, and reserve + load (32.9s) ~= baseline vllm init (32.4s) is a good negative result: deferral costs nothing. Resolving.

terrykong (approved):

PR #3423, opened 2026-07-30
chore: bump TQ to v0.19

macandro96:
# What does this PR do ?

1. Bumps TQ to v0.19 to enable checkpointing (Ref [here](https://github.com/Ascend/TransferQueue/blob/b75d570d88c50bbfcbe2171baa727fadd7216f76/docs/checkpoint.md))
2. Fixes for TQ v0.1.9 changing Mooncake reads: uniform rows were returned as nested tensors

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #2935, opened 2026-06-25
feat(data_plane): mooncake cpu rdma backend

ZhiyuLi-Nvidia:
# What does this PR do ?

Makes `data_plane.backend=mooncake_cpu` use CPU RDMA, and only CPU RDMA.

**RDMA is now the transport, not an option.** `_mooncake_transport_config()` auto-detects a RoCE-capable `mlx5` device and passes it to MooncakeStore. There is no TCP mode: `MC_MOONCAKE_PROTOCOL` is gone, and a host without a suitable device raises rather than degrading. A silent TCP fallback only ever served to let a run report mooncake/RDMA coverage it had not actually exercised.

**This is an enabling fix, not just a default flip.** The device probe on `main` used two `dirname`s where it needed three, so `basename $(dirname $(dirname .../mlx5_3/ports/1/link_layer))` resolved to the literal string `ports` — not a device. Setting `MC_MOONCAKE_PROTOCOL=rdma` on `main` therefore produced `device_name="ports"` and could never have worked. The probe is now `glob` + `Path.parents[2]`, which removes the shell entirely along with that class of bug.

**Sizing defaults corrected: 512 + 64 GiB → 64 + 4 GiB per client process.** `global_segment_size` and `local_buffer_size` are registered *per client process* — one per GPU — so a node pays `gpus_per_node x (segment + buffer)`. Under RDMA that memory is pinned via `ibv_reg_mr` and resident from setup, not lazily faulted in, so the old defaults meant 4.6 TiB/node on an 8-GPU host. That was survivable only while the transport was TCP and the mapping lazy; making RDMA the default converted a virtual reservation into a real allocation. Upstream TransferQueue defaults to 4 + 1 GiB per client.

**Deletes the TQ actor `runtime_env` injection.** `_patch_tq_actor_runtime_env` monkey-patched TransferQueue's actor classes to pip-install TQ from a git ref at every actor spawn, putting a network call on the actor-creation path. Its own TODO scoped it to removal once the nightly image shipped TQ via `pyproject.toml`, which it now does. Verified at 256 actors: no `ModuleNotFoundError`.

**Tests skip instead of falling back.** The mooncake unit fixture and `grpo_dp_mooncake.sh` share the adapter's `roce_device()` predicate and skip when no RoCE device is present, so they exercise the real transport where hardware allows and are honestly absent where it does not. InfiniBand is not auto-selected — RoCE is the fabric this is validated on — but `MC_MOONCAKE_DEVICE` names any device explicitly.

# Verification

[...]

PR #2957, opened 2026-06-26
feat(vlm): route VLM GRPO through TQ trainer when data_plane.enabled

ZhiyuLi-Nvidia:
## Summary

Route VLM GRPO through the TransferQueue data plane when `data_plane.enabled=true`, and thread VLM multimodal payloads (`pixel_values`, `image_grid_thw`, `mm_token_type_ids`, `imgs_sizes`, `num_frames`) through the TQ wire.

- `examples/run_vlm_grpo.py` mirrors the `run_grpo.py` launcher pattern: dispatch to `grpo_train_sync` when the data plane is enabled, otherwise stay on the legacy `grpo_train`.
- Multimodal fields cross the wire as a `torch.nested` payload plus per-row `tags` carrying the true segment shapes. `PackedTensor.to_wire` / `from_wire` are the only boundary between the domain wrapper and the wire; there are no `<key>__lengths` companion columns and nothing is padded in transit.
- Consumers that are not sequence-aligned on dim 1 (`materialize`'s pad-to-seqlen skip, `truncate_tensors`, both per-backend seq-dim validators, the TQ fetch list) dispatch through two registries in `nemo_rl/data/multimodal_utils.py`.
- No behavior change for `data_plane.enabled=false` (default).

## What crosses the wire

The codec dispatches on value type only (`Tensor` vs object `ndarray`); which lane a tensor takes is a registry lookup on the field name.

```
                                            what crosses the wire
 TEXT      message_log, strings, task_names   ndarray(dtype=object), untouched
   └─ passthrough                             (tensordict 0.12.2 loses identity
                                              on NonTensorStack; object arrays
                                              round-trip intact)

 TOKEN     input_ids, *logprobs, advantages,  torch.jagged nested, one row per
           token_mask, sample_mask, returns,  sample, sliced to its own length
           routed_experts, + the per-token    (drops mcore SP's TP-multiple
           multimodal maps mm_token_type_ids  padding on the way in)
           and token_type_ids
   └─ jagged by lengths                       TOKEN_ALIGNED_FIELDS =
                                                _TEXT_TOKEN_ALIGNED_FIELDS
                                                | PER_TOKEN_MULTIMODAL_FIELDS

 TENSOR    everything else rectangular        .detach().contiguous(), as-is
   └─ passthrough

 PIXEL     pixel_values, pixel_values_videos, PackedTensor -> to_wire():
           image_grid_thw, video_grid_thw,      flatten each segment to 1-D,
[...]

PR #3501, opened 2026-08-05
feat(data_plane): add Mooncake GDR support

zyzhou5:
# What does this PR do?

Adds the minimal NeMo-RL configuration plumbing needed to enable TransferQueue v0.1.9's Mooncake GPUDirect RDMA (GDR) path. TransferQueue v0.1.9 is already pinned on current `main` by #3423.

- adds optional `data_plane.use_gdr`; omitting it defaults to `false`, so existing recipes are unchanged;
- passes `protocol: rdma` and `use_gdr: true` to Mooncake when GDR is enabled;
- leaves RDMA device selection local to each Mooncake client unless explicitly configured;
- requires CUDA to be initialized before a GPU worker attaches its GDR-enabled TQ client, avoiding TQ's CPU fallback;
- retains NeMo-RL's existing transfer/decode paths and TQ's default staging-buffer size; and
- documents how users select TCP, CPU RDMA, or GPU RDMA.

Persistent queued Mooncake objects remain in CPU RAM. GDR changes the CUDA client's transfer path by using a registered GPU staging buffer; it does not turn the persistent queue into GPU storage.

# User configuration

```yaml
data_plane:
  enabled: true
  impl: transfer_queue
  backend: mooncake_cpu
  use_gdr: true
```

# Relationship to #2935

This PR does not stack on or require #2935. The TQ dependency update is already on `main` via #3423. The runtime-environment cleanup proposed by #2935 is separate and is not included here.

# Relationship to #3837 

To use GDR on the put side instead of falling back to CPU RDMA, we need #3837 to get merged since TQ must be initialized inside the generation worker after CUDA is initialized so it selects GDR for PUTs.
### E2E test

| Backend | W&B run |
| --- | --- |
| Simple | [View run](https://wandb.ai/nvidia/gdre2e-0903/runs/gdr-profile-3501-rebased-fix-simple-20260906-r1) |
| CPU RDMA | [View run](https://wandb.ai/nvidia/gdre2e-0903/runs/gdr-profile-3501-rebased-fix-cpu-rdma-20260906-r1) |
| GDR | [View run](https://wandb.ai/nvidia/gdre2e-0903/runs/gdr-profile-3501-rebased-fix2-gdr-20260906-r2) |

### DIFF
--- a/nemo_rl/data_plane/README.md
+++ b/nemo_rl/data_plane/README.md
@@ -437,6 +437,8 @@ data_plane:
     local_buffer_size:    4294967296   # 4 GiB/process
     reuse_registered_buffers: true     # reuse RDMA-registered buffers
     staging_buffer_size:   268435456   # 256 MiB/pool slot; bigger transfers bypass the pool
+    use_gdr: false                      # GPU-memory RDMA staging in CUDA clients
+    gdr_staging_buffer_mb: 1024         # persistent MiB per active GDR client
   # observability:                     # NotRequired
   #   enabled: false
 ```
@@ -445,13 +447,29 @@ These keys used to sit directly under `data_plane:`. That spelling is not
 rejected — it is simply never read. A config still using it silently gets
 this backend's defaults instead of its own values: an inherited config
 supplies the nested block, so a surviving flat key always loses the merge,
-with no warning either way.
+with no warning either way. Earlier revisions of PR #3501 also used flat
+`use_gdr` and `gdr_staging_buffer_mb` keys; those two spellings are rejected
+with a migration error because silently disabling GDR would invalidate a run.
 
 Backend choice:
 - **`simple`** — ZMQ-backed; lowest setup overhead. Default for tests
   and small runs.
-- **`mooncake_cpu`** — Mooncake transfer engine; higher throughput at
-  scale. Required for multi-node clusters with large bulk volume.
+- **`mooncake_cpu`** — Mooncake's RDMA-only transfer engine. By default,
+  tensors transfer through registered CPU staging. Set
+  `mooncake_cpu.use_gdr: true` to let CUDA-initialized clients use
+  TransferQueue's GDR staging path. CPU-only clients, such as a
+  SingleController producer, continue to use CPU RDMA. GDR changes the
+  client-side tensor transfer and staging path; queued objects still reside in
+  Mooncake-managed host-memory segments.
+
+The CPU host staging pool's `staging_buffer_size` is independent of the GDR
+buffer. `gdr_staging_buffer_mb` is the persistent GPU staging capacity per
+active CUDA client and defaults to 1024 MiB. Transfers through that per-client
+buffer are serialized. Aggregate fetches may exceed the buffer and are split
+into groups. In the mixed CPU-producer/GDR-receiver flow used by
+SingleController, however, each individual tensor must currently fit because
+the CPU PUT path does not create the chunk metadata required by an oversized
+GDR GET.
 
 Capacity rule of thumb (any backend):
 
--- a/nemo_rl/data_plane/README.md
+++ b/nemo_rl/data_plane/README.md
@@ -447,9 +447,7 @@ These keys used to sit directly under `data_plane:`. That spelling is not
 rejected — it is simply never read. A config still using it silently gets
 this backend's defaults instead of its own values: an inherited config
 supplies the nested block, so a surviving flat key always loses the merge,
-with no warning either way. Earlier revisions of PR #3501 also used flat
-`use_gdr` and `gdr_staging_buffer_mb` keys; those two spellings are rejected
-with a migration error because silently disabling GDR would invalidate a run.
+with no warning either way.
 
 Backend choice:
 - **`simple`** — ZMQ-backed; lowest setup overhead. Default for tests
--- a/nemo_rl/data_plane/adapters/transfer_queue.py
+++ b/nemo_rl/data_plane/adapters/transfer_queue.py
@@ -1053,8 +1053,11 @@ class TQDataPlaneClient(DataPlaneClient):
                 for key in wire_fields.keys()
             )
         )
-        gdr_staging = None
         if confirm_gdr_put:
+            # Checked before the put, not after: TQ fixes GDR eligibility when
+            # the client attaches, so this is decidable up front — and once
+            # `kv_batch_put` returns, the rows are already durable and the
+            # controller has been notified, so raising then would strand them.
             tq_client = tq.get_client()
             storage_manager = getattr(tq_client, "storage_manager", None)
             storage_client = getattr(storage_manager, "storage_client", None)
@@ -1074,11 +1077,6 @@ class TQDataPlaneClient(DataPlaneClient):
             tags=user_tags,
         )
         if confirm_gdr_put:
-            if not getattr(gdr_staging, "_initialized", False):
-                raise RuntimeError(
-                    "TransferQueue tensor PUT completed without initializing "
-                    "the requested GDR staging buffer"
-                )
             LOGGER.info(
                 "TransferQueue GDR tensor PUT active (partition=%s)", partition_id
             )
--- a/nemo_rl/data_plane/adapters/transfer_queue.py
+++ b/nemo_rl/data_plane/adapters/transfer_queue.py
@@ -684,6 +684,8 @@ def _init_tq(cfg: DataPlaneConfig) -> None:
                     "metadata_server": f"{local_ip}:50050",
                     "master_server_address": f"{local_ip}:50051",
                     **_mooncake_transport_config(),
+                    "use_gdr": bool(mooncake_cfg.use_gdr),
+                    "gdr_staging_buffer_mb": int(mooncake_cfg.gdr_staging_buffer_mb),
                 },
             },
         }
--- a/nemo_rl/data_plane/adapters/transfer_queue.py
+++ b/nemo_rl/data_plane/adapters/transfer_queue.py
@@ -776,6 +776,12 @@ def _from_wire(td: TensorDict) -> TensorDict:
 class TQDataPlaneClient(DataPlaneClient):
     """Adapter façade — maps NeMo-RL calls onto TransferQueue's public API."""
 
+    # Class-level so ``put_samples`` stays readable on an instance built
+    # without ``__init__`` — ``object.__new__`` in tests, or a process that
+    # unpickles a client without running the constructor.
+    _gdr_requested: bool = False
+    _gdr_put_confirmed: bool = False
+
     def __init__(self, cfg: DataPlaneConfig, *, bootstrap: bool = True) -> None:
         """Construct a TQ-backed client.
 
--- a/nemo_rl/data_plane/adapters/transfer_queue.py
+++ b/nemo_rl/data_plane/adapters/transfer_queue.py
@@ -27,6 +27,7 @@ import glob
 import importlib
 import ipaddress
 import json
+import logging
 import os
 import resource
 import socket
@@ -54,6 +55,8 @@ from nemo_rl.data_plane.interfaces import (
     backend_config,
     data_plane_supports_checkpointing,
 )
+
+LOGGER = logging.getLogger(__name__)
 
 # ──────────────────────────────────────────────────────────────────────────
 # Backend init — lifted from rl-arena/arena/backends.py.
@@ -819,6 +822,13 @@ class TQDataPlaneClient(DataPlaneClient):
 
         self._backend = cfg["backend"]
         self._supports_checkpointing = data_plane_supports_checkpointing(cfg)
+        # GDR is a mooncake_cpu-only transport knob, so key it off the backend
+        # directly rather than off any incidental per-backend flag.
+        self._gdr_requested = self._backend == "mooncake_cpu" and bool(
+            backend_config(cfg).use_gdr
+        )
+        self._gdr_put_confirmed = False
+
         if bootstrap:
             _init_tq(cfg)
         else:
@@ -1033,6 +1043,28 @@ class TQDataPlaneClient(DataPlaneClient):
             wire_fields = detached_fields
             field_names = [str(key) for key in detached_fields.keys()]
 
+        confirm_gdr_put = bool(
+            self._gdr_requested
+            and not self._gdr_put_confirmed
+            and torch.cuda.is_initialized()
+            and wire_fields is not None
+            and any(
+                isinstance(wire_fields.get(key), torch.Tensor)
+                for key in wire_fields.keys()
+            )
+        )
+        gdr_staging = None
+        if confirm_gdr_put:
+            tq_client = tq.get_client()
+            storage_manager = getattr(tq_client, "storage_manager", None)
+            storage_client = getattr(storage_manager, "storage_client", None)
+            gdr_staging = getattr(storage_client, "_gdr_staging", None)
+            if not getattr(storage_client, "use_gdr", False) or gdr_staging is None:
+                raise RuntimeError(
+                    "GDR was requested for a CUDA-initialized TransferQueue "
+                    "client, but TransferQueue selected CPU RDMA for tensor PUTs"
+                )
+
         self._mark_data_operation_started()
         # TQ's wire vocabulary is `keys=` — translation point.
         tq.kv_batch_put(
@@ -1041,6 +1073,16 @@ class TQDataPlaneClient(DataPlaneClient):
             fields=wire_fields,
             tags=user_tags,
         )
+        if confirm_gdr_put:
+            if not getattr(gdr_staging, "_initialized", False):
+                raise RuntimeError(
+                    "TransferQueue tensor PUT completed without initializing "
+                    "the requested GDR staging buffer"
+                )
+            LOGGER.info(
+                "TransferQueue GDR tensor PUT active (partition=%s)", partition_id
+            )
+            self._gdr_put_confirmed = True
 
         return KVBatchMeta(
             partition_id=partition_id,
--- a/nemo_rl/data_plane/interfaces.py
+++ b/nemo_rl/data_plane/interfaces.py
@@ -124,9 +124,7 @@ class DataPlaneConfig(TypedDict):
     ``local_buffer_size`` used to sit at this level. A config still using that
     spelling is not rejected — the flat key is simply never read, and
     :func:`backend_config` resolves the nested block (or its defaults) as if it
-    were absent. Earlier revisions of PR #3501 also put ``use_gdr`` and
-    ``gdr_staging_buffer_mb`` here; those two spellings are rejected explicitly
-    so a stale GDR recipe cannot silently run with GDR disabled. See there.
+    were absent. See there.
     """
 
     enabled: bool
@@ -138,14 +136,6 @@ class DataPlaneConfig(TypedDict):
     controller_address: NotRequired[str]
     ack_timeout_ms: NotRequired[int]
     observability: NotRequired["ObservabilityConfig"]
-
-    # Declared only so Pydantic preserves the first GDR integration's flat
-    # spellings until backend_config can raise the migration error below.
-    use_gdr: NotRequired[bool]
-    gdr_staging_buffer_mb: NotRequired[int]
-
-
-_FIRST_GDR_FLAT_KEYS = ("use_gdr", "gdr_staging_buffer_mb")
 
 
 _CHECKPOINTABLE_BACKENDS: frozenset[str] = frozenset({"simple"})
@@ -175,16 +165,9 @@ def backend_config(cfg: DataPlaneConfig) -> Any:
     (block already coerced to a model) or as a plain dict from a test.
 
     Sizing is read only from the nested block. A config still using the
-    pre-nesting flat sizing spelling gets this backend's defaults, not its own
-    values. The first GDR integration's flat spelling raises instead because
-    silently disabling a requested transport would invalidate a benchmark.
+    pre-nesting flat spelling gets this backend's defaults, not its own values.
     """
     backend = cfg["backend"]
-    stale_gdr = [key for key in _FIRST_GDR_FLAT_KEYS if key in cfg]
-    if stale_gdr:
-        keys = ",".join(stale_gdr)
-        raise ValueError(f"data_plane.{{{keys}}} moved under data_plane.mooncake_cpu")
-
     nested = cfg.get(backend) or {}
     if isinstance(nested, BaseModel):
         nested = nested.model_dump(exclude_unset=True)
--- a/nemo_rl/data_plane/interfaces.py
+++ b/nemo_rl/data_plane/interfaces.py
@@ -40,7 +40,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Annotated, Any, Callable, Literal, NotRequired, Sequence, TypedDict
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, PositiveInt
 from tensordict import TensorDict
 
 DATA_PLANE_CHECKPOINT_SCHEMA_VERSION = 2
@@ -80,6 +80,11 @@ class MooncakeCpuConfig(BaseModel, extra="allow"):
     admitted and never shrink — so raise it only when a per-key payload (one
     sample of one field) genuinely exceeds it, not for headroom.
 
+    ``use_gdr`` lets CUDA-initialized clients transfer through TransferQueue's
+    persistent GPU staging buffer. ``gdr_staging_buffer_mb`` is the positive
+    HBM capacity of that buffer per active GDR client. CPU-only clients keep
+    using the registered host-buffer path.
+
     Every RDMA rail on the host is offered to mooncake (see ``rdma_devices``).
     That is only safe with ``MC_ENABLE_DEST_DEVICE_AFFINITY=1``, which pins each
     transfer's peer rail to the local one by name; on a rail-isolated RoCE
@@ -91,6 +96,8 @@ class MooncakeCpuConfig(BaseModel, extra="allow"):
     local_buffer_size: int = 4294967296  # 4 GiB per client process
     reuse_registered_buffers: bool = True
     staging_buffer_size: int = 268435456  # 256 MiB per pool slot
+    use_gdr: bool = False
+    gdr_staging_buffer_mb: PositiveInt = 1024
 
 
 class DataPlaneConfig(TypedDict):
@@ -117,7 +124,9 @@ class DataPlaneConfig(TypedDict):
     ``local_buffer_size`` used to sit at this level. A config still using that
     spelling is not rejected — the flat key is simply never read, and
     :func:`backend_config` resolves the nested block (or its defaults) as if it
-    were absent. See there.
+    were absent. Earlier revisions of PR #3501 also put ``use_gdr`` and
+    ``gdr_staging_buffer_mb`` here; those two spellings are rejected explicitly
+    so a stale GDR recipe cannot silently run with GDR disabled. See there.
     """
 
     enabled: bool
@@ -129,6 +138,14 @@ class DataPlaneConfig(TypedDict):
     controller_address: NotRequired[str]
     ack_timeout_ms: NotRequired[int]
     observability: NotRequired["ObservabilityConfig"]
+
+    # Declared only so Pydantic preserves the first GDR integration's flat
+    # spellings until backend_config can raise the migration error below.
+    use_gdr: NotRequired[bool]
+    gdr_staging_buffer_mb: NotRequired[int]
+
+
+_FIRST_GDR_FLAT_KEYS = ("use_gdr", "gdr_staging_buffer_mb")
 
 
 _CHECKPOINTABLE_BACKENDS: frozenset[str] = frozenset({"simple"})
@@ -158,9 +175,16 @@ def backend_config(cfg: DataPlaneConfig) -> Any:
     (block already coerced to a model) or as a plain dict from a test.
 
     Sizing is read only from the nested block. A config still using the
-    pre-nesting flat spelling gets this backend's defaults, not its own values.
+    pre-nesting flat sizing spelling gets this backend's defaults, not its own
+    values. The first GDR integration's flat spelling raises instead because
+    silently disabling a requested transport would invalidate a benchmark.
     """
     backend = cfg["backend"]
+    stale_gdr = [key for key in _FIRST_GDR_FLAT_KEYS if key in cfg]
+    if stale_gdr:
+        keys = ",".join(stale_gdr)
+        raise ValueError(f"data_plane.{{{keys}}} moved under data_plane.mooncake_cpu")
+
     nested = cfg.get(backend) or {}
     if isinstance(nested, BaseModel):
         nested = nested.model_dump(exclude_unset=True)
--- a/tests/unit/data_plane/README.md
+++ b/tests/unit/data_plane/README.md
@@ -107,10 +107,11 @@ Generated audit of every test function under `tests/unit/data_plane/` with a one
 - `test_local_node_ip_returns_empty_on_exception` — DNS exception → returns empty string (no crash).
 - `test_mc_tcp_bind_address_overwrites_existing` — TQDataPlaneClient `__init__` uses direct assignment (not `setdefault`).
 
-## `test_mooncake_gdr.py` (3 tests)
+## `test_mooncake_gdr.py` (4 tests)
 
 - `test_init_tq_forwards_nested_gdr_config_and_keeps_rdma` — Nested GDR settings reach TransferQueue without changing #2935's RDMA/all-rail transport.
 - `test_cpu_only_client_may_attach_with_gdr_config` — A CPU-only producer may attach because GDR eligibility is client-local.
+- `test_gdr_tensor_put_is_confirmed_once_and_never_falls_back` — A CUDA client's tensor PUT raises rather than downgrading to CPU RDMA, and logs the GDR confirmation exactly once.
 - `test_gdr_receiver_requires_cuda_initialized` — A policy receiver must initialize CUDA before attaching its GDR client.
 
 ## `test_message_log_decompose.py` (11 tests)
--- a/tests/unit/data_plane/README.md
+++ b/tests/unit/data_plane/README.md
@@ -107,6 +107,12 @@ Generated audit of every test function under `tests/unit/data_plane/` with a one
 - `test_local_node_ip_returns_empty_on_exception` — DNS exception → returns empty string (no crash).
 - `test_mc_tcp_bind_address_overwrites_existing` — TQDataPlaneClient `__init__` uses direct assignment (not `setdefault`).
 
+## `test_mooncake_gdr.py` (3 tests)
+
+- `test_init_tq_forwards_nested_gdr_config_and_keeps_rdma` — Nested GDR settings reach TransferQueue without changing #2935's RDMA/all-rail transport.
+- `test_cpu_only_client_may_attach_with_gdr_config` — A CPU-only producer may attach because GDR eligibility is client-local.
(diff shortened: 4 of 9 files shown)


zyzhou5:
/ok to test [d2edefb](https://github.com/NVIDIA-NeMo/RL/commit/d2edefbafba83f4242f894b862d668e1ab34fdba)

zyzhou5:
/ok to test [31f9117](https://github.com/NVIDIA-NeMo/RL/commit/31f91179aa59c23217cc725216dd1123ae4afeeb)

zyzhou5:
/ok to test [5bdc8e7](https://github.com/NVIDIA-NeMo/RL/commit/5bdc8e7ecc4e093c5bc53d14a3d35669fad592e4)

nemo_rl/data_plane/adapters/transfer_queue.py
```diff
@@ -726,6 +731,10 @@ def __init__(self, cfg: DataPlaneConfig, *, bootstrap: bool = True) -> None:
         # squeezes the trailing 1 back on get. Drop when upstream TQ
         # unifies the schema/data shapes for 1D fields.
         self._promote_1d = cfg["backend"] == "mooncake_cpu"
+        self._gdr_requested = bool(
+            self._promote_1d and backend_config(cfg).use_gdr
+        )
+        self._gdr_put_confirmed = False
```
zyzhou5: [`transfer_queue.py:734-737`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/adapters/transfer_queue.py#L734-L737)

**1 action item.**

**TL;DR — the GDR master switch is derived from `_promote_1d`, a 1-D-schema shim that `main` has already deleted, so this line cannot survive the rebase and both obvious resolutions are wrong.**

PR-introduced.

`_promote_1d` is documented three lines above as a temporary workaround ("Drop when upstream TQ unifies the schema/data shapes for 1D fields"). It happens to equal `backend == "mooncake_cpu"` today, so it works — but "is this the mooncake backend" and "does TQ still need the 1-D shim" are two different questions with two different lifetimes, and the second one has now ended:

- `git grep -n "_promote_1d\|PROMOTE_1D" origin/main -- nemo_rl/` returns nothing. Both `self._promote_1d` and `nemo_rl.data_plane.schema.PROMOTE_1D_FIELDS` are gone from `main`, removed by [`da39e1785`](https://github.com/NVIDIA-NeMo/RL/commit/da39e178598a287e6970d92ab143511d79287c2c) (#2957), which fixes the schema at the source via `_patch_scalar_field_schema()` instead.
- This branch is 52 commits behind and `mergeable: CONFLICTING`. `git merge-tree` gives 4 conflict hunks, and one lands exactly on these lines — `main` replaces this whole region with `self._backend`, `self._supports_checkpointing`, and `_patch_scalar_field_schema()`.

Both natural resolutions fail:

1. Take main's side and re-append the GDR lines → `AttributeError: 'TQDataPlaneClient' object has no attribute '_promote_1d'` at client construction, on **every** backend, not just GDR runs. `tests/unit/data_plane/test_tq_lifecycle.py:56` constructs a `backend: "simple"` client and would fail immediately.
2. Default `_gdr_requested = False` → `confirm_gdr_put` is never true, the entire fail-closed guard from `480ccc707` becomes dead code, and a GDR run silently falls back to CPU RDMA — the exact outcome that commit exists to prevent.

**Action:** key the predicate off the backend directly, which is rebase-clean regardless of how the conflict resolves.

```suggestion
        self._gdr_requested = cfg["backend"] == "mooncake_cpu" and bool(
            backend_config(cfg).use_gdr
        )
        self._gdr_put_confirmed = False
```

**Follow-up (optional, same rebase):** this predicate now exists in two spellings — here and at [`worker_mixin.py:148-152`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/worker_mixin.py#L148-L152). A single `gdr_enabled(cfg)` helper next to `backend_config` in `interfaces.py` would stop them drifting. `interfaces.py` is the right home: it is a pure function of the config, and `worker_mixin.py` already imports `backend_config` from there while importing nothing from `adapters/`.

tests/unit/data_plane/test_mooncake_gdr.py
```diff
@@ -0,0 +1,160 @@
...
+        tq_adapter, "_patch_mooncake_staging_buffers", lambda *_, **__: None
+    )
+    monkeypatch.setattr(tq_adapter.os, "environ", dict(tq_adapter.os.environ))
+
+    client = tq_adapter.TQDataPlaneClient(_gdr_cfg(), bootstrap=False)
+
+    assert client._closed is False
+
+
+def test_gdr_tensor_put_is_confirmed_once_and_never_falls_back(
```
zyzhou5: [`test_mooncake_gdr.py:103`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/tests/unit/data_plane/test_mooncake_gdr.py#L103)

**1 action item.**

**TL;DR — CI has never run on the head commit, so this test and the code it covers are completely unvalidated.**

PR-introduced (process, not code).

`gh api repos/NVIDIA-NeMo/RL/commits/<sha>/check-runs`:

- `5bdc8e7ec` (commit 1) — **83** check-runs, including `L0_Unit_Tests_Other` (the shard that owns `tests/unit/data_plane/`), `L0_Unit_Tests_Data`, `Coverage (unit-test)` and Lint. All green.
- `480ccc707` (commit 2, **the head**) — **5** check-runs, all metadata: `semantic-pull-request` x2, `Post submodule check comment`, `triage`, `DCO`. No lint, no pyrefly, no unit shards.

The three `/ok to test` comments name `d2edefb`, `31f9117` and `5bdc8e7` — none names `480ccc7`, and copy-pr-bot gates workflows on the authorized SHA.

Commit 2 changed exactly two files (+87/-1): the +39 lines in `transfer_queue.py` that add the whole `confirm_gdr_put` probe and both new `RuntimeError`s, and the +49 lines here that add this test. So the newest and most intricate part of the PR has never had a test run against it. (Everything as of commit 1 — both YAMLs, both READMEs, `interfaces.py`, `worker_mixin.py`, `test_backend_config.py` — does have a green matrix.)

**Action:** rebase first, then `/ok to test <new head>`. Doing it in that order avoids buying a green tick on code that a 52-commit rebase is about to rewrite.

Minor rider, worth knowing before the run: `ruff-format` will likely collapse the `self._gdr_requested = bool(...)` wrap at `transfer_queue.py:734-736` — joined it is 84 chars against ruff's default 88, with no magic trailing comma to pin the split. That line is in commit 2, so lint has never seen it. It becomes moot if you take the suggestion on that line.

nemo_rl/data_plane/adapters/transfer_queue.py
```diff
@@ -915,13 +924,45 @@ def put_samples(
...
+        if confirm_gdr_put:
+            if not getattr(gdr_staging, "_initialized", False):
+                raise RuntimeError(
+                    "TransferQueue tensor PUT completed without initializing "
+                    "the requested GDR staging buffer"
+                )
+            LOGGER.info(
+                "TransferQueue GDR tensor PUT active (partition=%s)", partition_id
+            )
+            self._gdr_put_confirmed = True
```
zyzhou5: [`transfer_queue.py:956-965`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/adapters/transfer_queue.py#L956-L965)

**1 action item.**

**TL;DR — this post-PUT check raises after `kv_batch_put` has already committed the rows and notified the controller, so it cannot fail closed; it orphans data at the one call site where it is reachable.**

PR-introduced. To be clear up front: the **pre-PUT** check above is correct and worth keeping — this is only about the block after the write.

By the time this code runs, the write has landed:

1. `tq.kv_batch_put` resolves to `KVStorageManager.put_data`, which runs the Mooncake upsert at [`managers/base.py#L661`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/managers/base.py#L661) and then `await notify_data_update(...)` at [`#L689`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/managers/base.py#L689). Both complete before it returns.
2. So the rows are durable in Mooncake **and** registered with the controller — visible to `kv_list`, claimable by consumers.
3. Raising here means the caller never receives its `KVBatchMeta`.
4. Of the five `put_samples` call sites, only `ReplayBuffer.commit` ([`replay_buffer.py:846-887`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/algorithms/async_utils/replay_buffer.py#L846-L887)) rolls back via `clear_samples`. The one site where `confirm_gdr_put` is actually reachable — `column_io.write_columns`, reached from [`TQWorkerMixin._write_back`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/worker_mixin.py#L420) in the CUDA-initialized policy/value workers — has no rollback at all.

It also appears unreachable as a genuine failure. The pre-PUT check already tests TQ's exact routing condition — `use_gdr_path = self.use_gdr and self._gdr_staging is not None` at [`mooncake_client.py#L155`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L155) — and `_put_tensors_gdr` then calls `lazy_init` unconditionally at [`#L215`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L215), which sets `_initialized = True` at [`mooncake_utils.py#L145`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/utils/mooncake_utils.py#L145) or raises. The tell is that the test has to fake the only path that exercises it (`def put_with_gdr(**kwargs): staging._initialized = True`).

**Action:** drop the raise, keep the confirmation log and the latch.

```suggestion
        if confirm_gdr_put:
            LOGGER.info(
                "TransferQueue GDR tensor PUT active (partition=%s)", partition_id
            )
            self._gdr_put_confirmed = True
```

If you would rather keep a post-hoc probe, it needs to be `LOGGER.warning`, never `raise` — or move it before `kv_batch_put`.

Worth noting this block is in the commit CI has never run, so there is no green run pointing either way; the above is from reading TQ's source.

nemo_rl/data_plane/interfaces.py
```diff
@@ -142,9 +159,16 @@ def backend_config(cfg: DataPlaneConfig) -> Any:
...
-    pre-nesting flat spelling gets this backend's defaults, not its own values.
+    pre-nesting flat sizing spelling gets this backend's defaults, not its own
+    values. The first GDR integration's flat spelling raises instead because
+    silently disabling a requested transport would invalidate a benchmark.
     """
     backend = cfg["backend"]
+    stale_gdr = [key for key in _FIRST_GDR_FLAT_KEYS if key in cfg]
+    if stale_gdr:
+        keys = ",".join(stale_gdr)
+        raise ValueError(f"data_plane.{{{keys}}} moved under data_plane.mooncake_cpu")
```
zyzhou5: [`interfaces.py:167-170`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/interfaces.py#L167-L170)

**1 action item.**

**TL;DR — this migration error guards a spelling that never shipped, and the two `NotRequired` declarations added to make it fire are not actually load-bearing on the path production takes.**

PR-introduced.

`git log -S"use_gdr" 3cea9e1ed` returns zero commits, and `use_gdr` appears nowhere on `main`. The flat spelling existed only in earlier revisions of this branch, so the population being migrated is not "few users" — it is zero. In exchange the merged tree carries seven artifacts permanently: two `NotRequired` keys on the public [`DataPlaneConfig`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/interfaces.py#L139-L145), `_FIRST_GDR_FLAT_KEYS`, this raise, two doc paragraphs, and two tests.

The sharpest cost is the type itself. `use_gdr: NotRequired[bool]` at the top level *declares the key legal*, so a reader who trusts the annotation writes exactly the config this then rejects.

Two things worth weighing:

1. **The declarations may not be needed even if you keep the raise.** They are justified by "pydantic drops undeclared TypedDict keys", which is true for a bare `TypeAdapter(DataPlaneConfig)` — the case `test_pydantic_preserves_flat_gdr_keys_for_migration_error` covers. But in production `DataPlaneConfig` is reached as a field of `class MasterConfig(BaseModel, extra="allow")` ([`config.py:539,553`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/algorithms/single_controller_utils/config.py#L539)), and pydantic's `_typed_dict_schema` only defaults `extra_behavior='ignore'` when no enclosing config sets it — a TypedDict without `__pydantic_config__` does not push its own wrapper, so the model's `extra="allow"` stays in scope. If that reading is right, a stale flat key survives validation without the declarations and still reaches this raise. **Could you confirm by deleting the two declarations and re-running the suite?** That is a 30-second check and settles it; we could not run pydantic here.
2. **The rule is asymmetric.** The docstring justifies raising because "silently disabling a requested transport would invalidate a benchmark" — but `backend_config` reads only the nested block ([`#L172`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/interfaces.py#L172)), so flat `reuse_registered_buffers: false` is silently discarded, which leaves the pool **on** when the config asked for it off. Same harm, opposite handling, three lines apart. And flat `global_segment_size` really did ship in #2439 (`git show 4454fa64e:examples/configs/grpo_math_1B.yaml`, 512 GiB) and is silently dropped to the 64 GiB default today — an 8x change with no warning. The loud guard is on the impossible case.

**Action:** consider dropping the mechanism — the two declarations, `_FIRST_GDR_FLAT_KEYS`, this raise, the two doc paragraphs, and the two tests that pin it. If you would rather keep a net, making it general (warn on any flat key matching a nested field name) would cover all six spellings and remove the asymmetry rather than documenting it.

Two riders that disappear with it: the message tells a `backend: simple` user to move the key under `data_plane.mooncake_cpu`, a block that backend never reads; and `README.md:450` / `interfaces.py:124` cite "PR #3501" in shipped prose, which will not mean anything to a reader after merge (every other `PR #NNNN` in `nemo_rl/` points at an upstream or already-merged PR).

Also, unrelated to the mechanism but worth fixing either way: the PR description's **"User configuration"** block uses the flat spelling, so copy-pasting it raises today.

nemo_rl/data_plane/adapters/transfer_queue.py
```diff
@@ -915,13 +924,45 @@ def put_samples(
...
+        if confirm_gdr_put:
+            tq_client = tq.get_client()
+            storage_manager = getattr(tq_client, "storage_manager", None)
+            storage_client = getattr(storage_manager, "storage_client", None)
+            gdr_staging = getattr(storage_client, "_gdr_staging", None)
+            if not getattr(storage_client, "use_gdr", False) or gdr_staging is None:
+                raise RuntimeError(
+                    "GDR was requested for a CUDA-initialized TransferQueue "
+                    "client, but TransferQueue selected CPU RDMA for tensor PUTs"
+                )
```
zyzhou5: [`transfer_queue.py:938-947`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/adapters/transfer_queue.py#L938-L947)

**2 action items, 1 follow-up.**

**TL;DR — the probe reads five TQ attributes (two private) with falsy `getattr` defaults, so an upstream rename becomes a hard failure with a confidently wrong message, and the tests cannot catch it.**

PR-introduced. The attribute names are all correct at the current pin — we verified each hop — so this is about what happens when the pin moves, which is routine (#3423 last moved it).

The chain is `tq_client.storage_manager` -> `.storage_client` -> `.use_gdr` / `._gdr_staging`, then `._initialized` on the staging object. The last two are TQ-private. Because both `getattr` calls use falsy defaults, "this attribute moved" is indistinguishable from "GDR is off":

1. A TQ bump renames `_gdr_staging` (plausible — TQ could generalise staging beyond GDR).
2. `getattr(storage_client, "_gdr_staging", None)` returns `None`.
3. This raises `RuntimeError: GDR was requested for a CUDA-initialized TransferQueue client, but TransferQueue selected CPU RDMA for tensor PUTs` on the first tensor PUT.
4. That message is false — GDR was working; only the attribute moved. The operator debugs their RDMA fabric instead of the pin.
5. `test_gdr_tensor_put_is_confirmed_once_and_never_falls_back` stays green, because its `SimpleNamespace` still spells `_gdr_staging`.

**AI-1 — distinguish the two failures using this file's own idiom.** [`_tq_shape_drift_error`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/adapters/transfer_queue.py#L285-L303) exists 650 lines up for exactly this, with three call sites (331, 380, 390), and its docstring says so: "All three guards fire for the same reason — the pinned `transfer_queue` revision no longer has the internals a patch depends on." Raising it when the attribute is absent yields the correct diagnosis ("The TQ pin in pyproject.toml has moved: re-point the patch at the new attribute") and keeps the CPU-RDMA message for the case it actually describes.

**AI-2 — pin the fake to the real class.** The test's `SimpleNamespace` mirrors the shape but pins none of the names. A contract test closes that, and `pytest.importorskip` is already the house idiom in this directory ([`test_mooncake_staging_pool.py:99`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/tests/unit/data_plane/test_mooncake_staging_pool.py#L99) imports the real `mooncake_client` module). `GdrStaging.__init__` only assigns fields — `cudaMalloc` is deferred to `lazy_init` — so this needs no GPU:

```python
def test_gdr_confirmation_reads_upstream_private_attribute() -> None:
    """The PUT guard reads TransferQueue's private ``_initialized``; a rename
    upstream would fail every GDR run while the fakes above stayed green."""
    from transfer_queue.utils.mooncake_utils import GdrStaging

    assert "_initialized" in vars(GdrStaging(1))
```

(We could not execute it — no Python environment on the review host — so please run it before keeping it.)

**Follow-up:** giving the chain one address — a small `_gdr_staging_handle(tq_client)` beside `_patch_mooncake_register_check` / `_patch_mooncake_staging_buffers`, which is already this file's designated place for reaching into upstream internals — would make a future pin bump a one-line edit instead of five scattered reads.

Small rider: [`test_cpu_only_client_may_attach_with_gdr_config`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/tests/unit/data_plane/test_mooncake_gdr.py#L87-L100) asserts `client._closed is False`, but `_closed = False` is assigned unconditionally at the end of `__init__`, so it only restates "the constructor returned". `assert client._gdr_requested is True` would actually pin the requested-but-not-active state the test is named for.

nemo_rl/data_plane/interfaces.py
```diff
@@ -88,6 +93,8 @@ class MooncakeCpuConfig(BaseModel, extra="allow"):
     local_buffer_size: int = 4294967296  # 4 GiB per client process
     reuse_registered_buffers: bool = True
     staging_buffer_size: int = 268435456  # 256 MiB per pool slot
+    use_gdr: bool = False
+    gdr_staging_buffer_mb: PositiveInt = 1024
```
zyzhou5: [`interfaces.py:97`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/interfaces.py#L97)

**2 action items.**

**TL;DR — this is an opt-in performance feature with no numbers, and on the one PUT path where GDR is live today NeMo-RL hands TQ CPU-resident tensors, which may make it slower rather than faster.**

PR-introduced. The capability is clearly worth having, and the config plumbing looks right — this is a request for the data that would let a reviewer decide whether to switch it on.

**AI-1 — one A/B.** On a stated model and cluster shape, the rollout->policy transfer with `mooncake_cpu.use_gdr: false` vs `true`: effective GB/s and transfer-phase wall-clock, plus end-to-end step time (same seed and batch shape, median of 3 steps). This is measurable today — the GPU workers' bulk GETs and their write-back PUTs already take the GDR path, so it does not need to wait for #3837, which only moves the *generation worker's* bulk PUT.

**AI-2 — the HBM bill.** Measured `torch.cuda.memory_reserved()` delta per client at the 1024 MiB default, and the per-node total at your `gpus_per_node`. At 8 clients/node that is roughly 8 GiB taken from the training budget, and it is the number that decides whether a memory-tight recipe can enable this at all.

**Context — one thing worth measuring specifically.** On the write-back path, [`worker_mixin.py:462`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/worker_mixin.py#L462) does `val.detach().to("cpu")` before the PUT, so `prev_logprobs`, `reference_policy_logprobs` and `values` arrive at TQ as **CPU** tensors. TQ routes PUTs at client level, not per tensor — [`use_gdr_path = self.use_gdr and self._gdr_staging is not None`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L155), no device check, and its own comment at [`#L219`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L219) says "Tensors may be on CPU or CUDA". So with `use_gdr: true` those CPU tensors are staged through the GPU buffer — an extra host-to-device hop the CPU-RDMA path does not make — and [`_put_tensors_gdr` calls `torch.cuda.synchronize()`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L228) unconditionally, a full-device sync on the training worker once per write-back batch.

That may well be fine or even a wash, but it is the sort of thing an A/B would surface, and it is worth knowing whether the win is concentrated on the GET side.

If the honest answer is "config plumbing only, benchmarks once #3837 lands", saying so explicitly is a perfectly good outcome — it just tells reviewers to judge this as plumbing rather than as a performance claim.

nemo_rl/data_plane/adapters/transfer_queue.py
zyzhou5: **1 action item.**

`_gdr_requested` is derived from `_promote_1d`, but `main` deleted both `_promote_1d` and `PROMOTE_1D_FIELDS` in [`da39e1785`](https://github.com/NVIDIA-NeMo/RL/commit/da39e178598a287e6970d92ab143511d79287c2c) (#2957). This branch is 52 commits behind, and one of the four merge conflicts lands on exactly these lines — so you'll have to resolve it by hand, and both obvious resolutions are wrong:

- keep `_promote_1d` → reintroduces a shim `main` deliberately removed, and `tests/unit/data_plane/test_tq_lifecycle.py:56` (a `backend: "simple"` client) fails with `AttributeError`;
- default `_gdr_requested = False` → silently disables the fail-closed guard this commit exists to add.

Keying off the backend directly is rebase-clean either way:

```suggestion
        self._gdr_requested = cfg["backend"] == "mooncake_cpu" and bool(
            backend_config(cfg).use_gdr
        )
        self._gdr_put_confirmed = False
```

nemo_rl/data_plane/adapters/transfer_queue.py
zyzhou5, line 1089: **1 action item.**

This raises *after* `kv_batch_put` has committed the rows and notified the controller ([`put_data`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/managers/base.py#L661), [`notify_data_update`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/managers/base.py#L689)), so it can't fail closed — and the one call site where it's reachable, `column_io.write_columns` via [`TQWorkerMixin._write_back`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/worker_mixin.py#L420), has no rollback, so the rows are left orphaned.

It also looks unreachable as a genuine failure: the pre-PUT check above already tests TQ's actual routing condition ([`use_gdr_path`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L155)), and `_put_tensors_gdr` then calls [`lazy_init`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L215) unconditionally, which sets `_initialized`. The test has to fake it (`staging._initialized = True`) to reach this branch.

The pre-PUT check is the valuable half and should stay — this is only about the post-PUT raise:

```suggestion
        if confirm_gdr_put:
            LOGGER.info(
                "TransferQueue GDR tensor PUT active (partition=%s)", partition_id
            )
            self._gdr_put_confirmed = True
```

tests/unit/data_plane/README.md
```diff
@@ -112,6 +112,12 @@ Generated audit of every test function under `tests/unit/data_plane/` with a one
 - `test_local_node_ip_returns_empty_on_exception` — DNS exception → returns empty string (no crash).
 - `test_mc_tcp_bind_address_overwrites_existing` — TQDataPlaneClient `__init__` uses direct assignment (not `setdefault`).
 
+## `test_mooncake_gdr.py` (3 tests)
+
+- `test_init_tq_forwards_nested_gdr_config_and_keeps_rdma` — Nested GDR settings reach TransferQueue without changing #2935's RDMA/all-rail transport.
+- `test_cpu_only_client_may_attach_with_gdr_config` — A CPU-only producer may attach because GDR eligibility is client-local.
+- `test_gdr_receiver_requires_cuda_initialized` — A policy receiver must initialize CUDA before attaching its GDR client.
```
zyzhou5: **1 action item.**

`480ccc707` added a fourth test but this section (written in `5bdc8e7ec`) still says 3.

```suggestion
## `test_mooncake_gdr.py` (4 tests)

- `test_init_tq_forwards_nested_gdr_config_and_keeps_rdma` — Nested GDR settings reach TransferQueue without changing #2935's RDMA/all-rail transport.
- `test_cpu_only_client_may_attach_with_gdr_config` — A CPU-only producer may attach because GDR eligibility is client-local.
- `test_gdr_tensor_put_is_confirmed_once_and_never_falls_back` — A CUDA client's first tensor PUT must engage GDR; a CPU-RDMA fallback raises instead of downgrading silently.
- `test_gdr_receiver_requires_cuda_initialized` — A policy receiver must initialize CUDA before attaching its GDR client.
```

nemo_rl/data_plane/interfaces.py
zyzhou5, line 100: **1 action item, not blocking.**

Could you add some numbers for this? An A/B on one stated model/cluster (`use_gdr` false vs true — transfer GB/s and end-to-end step time), plus the measured HBM cost per client at the 1024 MiB default. At 8 clients/node that's ~8 GiB off the training budget, which is what decides whether a memory-tight recipe can enable it.

One thing worth measuring specifically: on the write-back path [`worker_mixin.py:462`](https://github.com/NVIDIA-NeMo/RL/blob/480ccc707c2009625471097137a35bd8f72c9444/nemo_rl/data_plane/worker_mixin.py#L462) does `.to("cpu")` before the PUT, and TQ picks the GDR path per *client*, not per tensor ([`use_gdr_path`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L155); its comment at [`#L219`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L219) notes tensors may be on CPU or CUDA). So those tensors would be staged through the GPU buffer, with a [`torch.cuda.synchronize()`](https://github.com/Ascend/TransferQueue/blob/c51614308b68c8d7a87c9b3ef62d59e14c69bde2/transfer_queue/storage/clients/mooncake_client.py#L228) per batch. Possibly a wash, but the A/B would show it.

"Config plumbing only, benchmarks once #3837 lands" is a perfectly good answer here.

nemo_rl/data_plane/interfaces.py
zyzhou5: **Optional — no action needed if you'd rather keep it.**

This rejects flat `data_plane.use_gdr`, but that spelling never shipped — `git log -S"use_gdr" 3cea9e1ed` is empty, so no merged config can contain it. The cost is that `DataPlaneConfig` now declares `use_gdr` / `gdr_staging_buffer_mb` as `NotRequired`, i.e. documents them as legal top-level keys, purely so this check can reject them.

Not a defect, just permanent schema surface for zero affected configs — your call.

zyzhou5:
## Self-review with `/review-pr-team`

Run with the repo's multi-agent review skill — RL-codebase, bug-finder, test and design reviewers, plus an adversarial devil's-advocate pass that cut roughly two thirds of the raw findings. Everything was re-verified against the pinned TransferQueue source (`c516143`). Five inline comments on this PR.

**Note on the first comment.** It was written before the rebase, so it shows as outdated. That one is already fixed: the branch is now rebased onto `cbbb57e7a`, and `_gdr_requested` no longer rides on `_promote_1d`, which `main` deleted in #2957 — it keys off the backend directly.

**What checks out.** All three claims the new README prose makes about TransferQueue are accurate — the GDR staging buffer really is serialized behind a single lock, `split_by_bytes` really does group aggregate fetches, and the CPU PUT path really does omit the chunk metadata an oversized GDR GET needs. `gdr_staging_buffer_mb: PositiveInt` is a good catch: TQ treats `0` as "silently fall back to CPU RDMA even with `use_gdr=True`", and this turns that into a config-time error instead. The CUDA guard in `TQWorkerMixin` is on the right seam — all four non-mixin client-construction sites are genuinely CPU-only, so `test_cpu_only_client_may_attach_with_gdr_config` is a correct spec rather than a gap. The two exemplar YAMLs are in sync, and the new test file's pytest marks and CI shard are right.

**Still outstanding: CI has never run on this code.** The old head `480ccc707` had 5 check-runs, all metadata; `5bdc8e7ec` had 83 including the full L0 matrix and lint, all green. The last `/ok to test` names `5bdc8e7`, so the fail-closed commit has never been validated — and the rebased head needs a fresh `/ok to test`.

**Method caveat.** Nothing was executed: the review host has no `uv`, `pytest`, or importable `torch`/`pydantic`. Every finding is static reading of this branch plus pinned TransferQueue source, and anything that would need a run to confirm is labelled as such inline.

Several candidate findings were investigated and deliberately dropped rather than posted: three docstring asks (no ruff rule requires a `Raises:` section, and `D417` is ignored here), a suspected missing `pydantic` import (it is imported, and pre-dates this PR), a YAML comment-alignment nit (lint was green on the commit containing it), a hardening ask against a hypothetical future TransferQueue rename (all attribute names are correct at the current pin), and several items with no constructible failing case.

_Generated by Claude Code_

zyzhou5:
/ok to test [1376417](https://github.com/NVIDIA-NeMo/RL/pull/3501/commits/13764175031d94e1c01583b264041514945ef46d)

nemo_rl/data_plane/README.md
```diff
@@ -445,13 +447,29 @@ These keys used to sit directly under `data_plane:`. That spelling is not
...
+with a migration error because silently disabling GDR would invalidate a run.
 
 Backend choice:
 - **`simple`** — ZMQ-backed; lowest setup overhead. Default for tests
   and small runs.
-- **`mooncake_cpu`** — Mooncake transfer engine; higher throughput at
-  scale. Required for multi-node clusters with large bulk volume.
+- **`mooncake_cpu`** — Mooncake's RDMA-only transfer engine. By default,
+  tensors transfer through registered CPU staging. Set
+  `mooncake_cpu.use_gdr: true` to let CUDA-initialized clients use
```
ZhiyuLi-Nvidia, line 457: Do you think we should change the name?

Option one

* moocnake
   - cpu
   - gdr

Option 2

* mooncake_cpu
* mooncake_gdr

nemo_rl/data_plane/README.md
```diff
@@ -437,6 +437,8 @@ data_plane:
     local_buffer_size:    4294967296   # 4 GiB/process
     reuse_registered_buffers: true     # reuse RDMA-registered buffers
     staging_buffer_size:   268435456   # 256 MiB/pool slot; bigger transfers bypass the pool
+    use_gdr: false                      # GPU-memory RDMA staging in CUDA clients
+    gdr_staging_buffer_mb: 1024         # persistent MiB per active GDR client
```
ZhiyuLi-Nvidia, line 441: "plus the measured HBM cost per client at the 1024 MiB default. At 8 clients/node that's ~8 GiB off the training budget"
Can we also put this in doc as well if not added?

ZhiyuLi-Nvidia, line 441: To get the right buffer number, any logic we should add here?
* like we need this buffer size at least bigger than per sample per field payload size
* what's the trade off between speed and memory usge.

"~8 GiB off the training budget"" that's ~10% of hbm in H100 so might be no trivial.

ZhiyuLi-Nvidia:
@zyzhou5 awesome PR!

ZhiyuLi-Nvidia:
@zyzhou5 I remember you mentioned we need to some changes in both put and get side to avoid unnecessary H2D or D2H.

I think the best we can do is 
vllm/put(gpu tensor) ===GDR===> TQ cpu storage ====GDR===> policy/get (gpu tensor)

Have we achieved that?

nemo_rl/data_plane/README.md
zyzhou5: This sounds reasonable to me. But maybe we should have a separate renaming PR since it touches a lot of files.

zyzhou5: created an issue to track this #4049

zyzhou5:
> @zyzhou5 I remember you mentioned we need to some changes in both put and get side to avoid unnecessary H2D or D2H.
> 
> I think the best we can do is vllm/put(gpu tensor) ===GDR===> TQ cpu storage ====GDR===> policy/get (gpu tensor)
> 
> Have we achieved that?

not yet... the PUT operation starts from CPU so there is an extra H2D. The PUT inference-side fix would need a bigger change so should be placed in another PR. Created an issue to track it #4051.

zyzhou5:
/ok to test [9896a67](https://github.com/NVIDIA-NeMo/RL/pull/3501/commits/9896a67bbc6a65f52ed71bcad93e783680ba3a3b)

zyzhou5:
/ok to test [2b15799](https://github.com/NVIDIA-NeMo/RL/pull/3501/commits/2b15799829e6aeea3950a977747e847dcfc72f4d)

terrykong (approved):

ZhiyuLi-Nvidia (approved):
LGTM.

PR #3503, opened 2026-08-05
feat(refit): support BF16 to NVFP4 NCCL reshard

seonjinn:
## What does this PR do?

This PR stacks on #3477 and adds BF16-to-NVFP4 NCCL Reshard refit for vLLM rollout.

- Supports W4A16 rollout from a BF16 Megatron policy.
- Supports W4A4 rollout with a validated static calibration artifact.
- Keeps the NCCL wire tensor in BF16 and converts complete routed-expert groups to ModelOpt NVFP4 tensors on the vLLM receiver.
- Finalizes each layerwise vLLM reload after NCCL bulk and misc weights arrive.
- Fails before collective initialization for unsupported precision, topology, projection, calibration, and metadata combinations.

## Why?

#3477 supports BF16-to-MXFP8 refit. NVFP4 needs grouped receiver-side materialization because one routed-expert group produces packed values, block scales, and global scales. W4A4 additionally requires immutable input calibration. Without this path, disaggregated W4A16 and W4A4 rollout cannot use NCCL Reshard with a BF16 trainer.

## Scope

| Trainer | Rollout | Status |
|---|---|---|
| BF16 Megatron | W4A16 vLLM | Supported |
| BF16 Megatron | W4A4 vLLM | Supported with calibration artifact |
| BF16 Megatron | MXFP8 vLLM | Provided by #3477 |

The NVFP4 path currently supports routed MoE expert projections with generation TP=1, EP=1, and PP=1. TP greater than 1 is rejected because NCCL already produces TP-local expert shards and the vLLM checkpoint loader would shard rank-2 expert tensors again. Existing BF16, matching-FP8, MXFP8, and legacy refit behavior remains unchanged.

## Validation

- Merged #3477 head `9f2170a84` and upstream `main` `af451b1a1`.
- Formatting, Ruff, syntax compilation, and diff checks pass locally.
- Focused GCP-NRT NCCL/vLLM tests on the pre-merge source: 147 passed, 7 skipped.
- GCP-NRT two-step smoke validation on TP1: W4A16 job `498843` and W4A4 job `498845` completed with exit code 0.
- Full CI rerun requested for head `49a61520e`.

## Review boundary

This stacked PR contains the NVFP4 receiver, W4A4 calibration, NCCL initialization and validation integration, and focused tests. Benchmark recipes, launch scripts, reports, and experiment artifacts are intentionally excluded.


seonjinn:
## Self-review (agent team + adversarial verification)

Reviewed at head `0aa1f1ac1` (46 commits behind `main`, merge-base `3d873ebc2`). Three independent passes: production correctness, test quality, and main-drift analysis. Findings ranked by severity.

### 1. Blocker: the support validator unconditionally rejects every real NVFP4 config this PR adds

`nemo_rl/weight_sync/nccl_reshard_utils.py:617-622` is a pre-existing block, untouched by this diff:

```python
if generation.get("real_quant"):
    violations.append(
        "policy.generation.real_quant must be False "
        "(nccl_reshard_refit's bulk xferdtensor writes bypass the "
        "layerwise-reload weight loaders that ModelOpt real quant requires)."
    )
```

This PR then defines `real_nvfp4 = bool(generation.get("real_quant"))` and adds the new `if real_nvfp4:` validation block below it. Both append into the same `violations` list, which raises a single `ValueError` at the end of `check_nccl_reshard_refit_support`. So any config with `real_quant=True` fails the old check before the new NVFP4-specific checks matter, and the feature is unreachable through the validator that `grpo.py` calls before the reshard path is used.

This is confirmed three ways: direct code read; executing the validator against this PR's own `_valid_nvfp4_config` fixture (raises `ValueError: ... real_quant must be False`); and it is exactly the current CI failure on this head - `L0_Unit_Tests_Other` fails on `tests/unit/weight_sync/test_nvfp4_nccl_refit.py::test_validator_accepts_plain_bf16_to_supported_nvfp4[w4a16]` (1 failed / 1904 passed).

Fix: delete the old block or narrow it to reject only real-quant configurations that the NVFP4 path does not support, letting the `real_nvfp4` block own the constraints.

### 2. Major: merging latest main needs manual reconciliation in `build_hf_to_local_param_map`

`git merge-tree --write-tree` against current main reports 3 conflicting files:

- `nemo_rl/models/generation/vllm/vllm_backend.py` - real conflict. Main independently grew the same method with a new wire-dtype resolution/validation block ending in a strict `elif wire_dtype != vllm_param.dtype: raise`, while this PR adds the `_allows_receiver_dtype_conversion()` escape hatch on that same `elif`. Neither existed at the merge-base, so the resolution has to combine main's stricter validation with the escape hatch by hand, and a test should assert the hatch still fires under main's path.
- `nemo_rl/models/generation/vllm/quantization/fp8.py:200` - trivial: main added `trust_remote_code=True` to `AutoConfig.from_pretrained`; keep main's version.
- `tests/unit/models/generation/test_vllm_fp8_quantization.py` - insertion adjacency only; keep both tests.

Everything else (`nccl_reshard_utils.py`, the synchronizer, `config.py`, `distillation.py`, the new `nemo_rl/modelopt/` files) auto-merges clean. Main's new `_supports_unquantized_flashinfer_trtllm_refit` gate does not overlap; main already added the `False` override in `vllm_quant_backend.py`.

### 3. Minor: codecov/patch is at 28.98% of the diff vs the 80% target

The check is failing; the untested surface is concentrated in the new production files.

### 4. Nit: stale docstring in `tests/unit/models/generation/test_nccl_reshard_backend.py`

The `_make_ext` helper's docstring says `named_modules()` drives FusedMoE layout detection in `_build_hf_to_gen_backend_mapping`; in current code `named_modules()` is only used in the unrelated `draft_model` path (`vllm_backend.py:403`), and layout resolution happens during `process_weights_after_loading`. Inert for the test itself; docstring-only fix.

### Verified clean

- NVFP4 quantization math in `nvfp4_refit.py` (block scaling with the 6.0*448.0 denominator, W4A16 vs W4A4 branching, shape/finite guards) and routed-expert family completeness validation - fails closed on incomplete or mixed families.
- `calibration_artifact.py` identity validation: model id/revision/quant-cfg exact match, SHA-256 digest for file-based configs, exact-set projection-name match - no silent acceptance of a wrong artifact found.
- Fail-before-collective ordering: both `prepare_refit_info` calls (which run the new validation) are `ray.get`-blocked strictly before `init_collective()` in the synchronizer, so misconfiguration cannot reach the first NCCL collective.
- Receiver dispatch is purely additive via the subclass hook; the MXFP8 path from #3477 is undisturbed. NVFP4 staging refuses to load until each `w13`/`w2` expert group is completely received. `distillation.py` rejects `nccl_reshard` explicitly instead of silently falling back.
- Test suite: no vacuous asserts, no mock-identity pitfalls, no bypassed production paths, no stale regexes found; the distillation reject-path test asserts real state, not mocks.

Follow-up worth checking once the blocker is fixed: whether `prepare_refit_info`'s `results[0]`-only view is complete under trainer PP>1, since the NVFP4 routed-expert classification is built from that name list (pre-existing pattern, not introduced here).

PR #3506, opened 2026-08-05
feat(rollout): add stable partial-rollout lineage

macandro96:
# What does this PR do?

Adds the durable identity and in-memory lineage foundation for recovering partially completed NeMo-Gym prompt groups.

- Gives each prompt group and GRPO sibling a stable logical identity.
- Tracks physical Gym/Gate attempts separately, so redispatch can create a new attempt without changing canonical training-row identity.
- Records sibling receipts, rewards, and TQ staging keys as soon as Gym seals them.
- Models ownership through generation, finalization, training claim, and uncheckpointed application.
- Preserves sealed siblings when another sibling fails and abandons only unfinished attempts.
- Stores compact attempt UUIDs and small `PromptRef` values while excluding full runtime prompts from serialized lineage.
- Validates the versioned serialized state and rejects unsupported or invalid statuses.

This layer intentionally does not persist or restore the ledger. Persistence and selective redispatch are added by #3507.

## Stack

- Layer 1 of 4.
- Base: `amahishi/sc-tq-token-capture-recovery`, refreshed with the current capture/recovery integration.
- Next: #3507.

## Scope

This is internal plumbing for the existing `token_capture.enabled=true` Single Controller NeMo-Gym path. It adds no user-facing configuration.

Tensor payloads remain in TQ. The controller ledger contains identity, ownership state, compact prompt references, receipts, rewards, and storage references.

## Validation

- Unit coverage is included for identity, lineage transitions, retries, finalization ownership, training claims, serialization, and invalid persisted states.
- The PR remains draft while the refreshed focused suite, recovery functionals, and full CI are rerun.

PR #3507, opened 2026-08-05
feat(rollout): persist and recover partial rollout groups

macandro96:
# What does this PR do?

Persists partial-rollout lineage with native TQ state and recovers it through the Single Controller restart lifecycle.

- Saves rollout lineage beside native TQ state, replay metadata, the dataloader cursor, and controller checkpoint metadata.
- Restores TQ through the bootstrap data-plane client before runtime consumers are constructed.
- Reconstructs finalized replay ownership from metadata while tensor payloads remain authoritative in TQ.
- Validates canonical rows, live staging ownership, receipt references, and lineage digest before recovery.
- Clears unreferenced staging rows and drops unfinished groups that are stale under the restored sampler policy.
- Reuses sealed siblings and redispatches only missing or abandoned generation indices.
- Applies bounded infrastructure retries during startup recovery.
- Resolves compact prompt references through the restored input stream rather than serializing large prompts.
- Makes the checkpoint barrier re-entrant.
- Runs restored recovery beside fresh dataloader dispatch after ownership and buffer permits are reconstructed; both share the bounded rollout-worker budget.

Frequent rollout-only snapshots are added by #3508.

## Stack

- Layer 2 of 4.
- Previous: #3506.
- Next: #3508.

## Recovery contract

1. Load the selected trainer checkpoint and native TQ snapshot.
2. Restore replay metadata, lineage, sampler state, and dataloader position.
3. Validate TQ inventory against finalized and unfinished ownership.
4. Reserve restored buffer capacity before fresh dispatch.
5. Recover unfinished groups concurrently with fresh work.
6. Reuse sealed siblings and redispatch only incomplete siblings.

The persisted controller sidecar remains metadata-only.

## Validation

- Unit coverage includes persistence, inventory validation, selective redispatch, finalizer replay, stale cleanup, capacity ownership, sampler restore, and recovery/fresh-dispatch overlap.
- The PR remains draft while the refreshed focused suite, recovery functionals, and full CI are rerun.

### DIFF
--- a/nemo_rl/algorithms/async_utils/replay_buffer.py
+++ b/nemo_rl/algorithms/async_utils/replay_buffer.py
@@ -158,16 +158,33 @@ class DataPlaneCheckpointBarrier:
         self._condition = asyncio.Condition()
         self._checkpoint_active = False
         self._active_mutations = 0
+        self._mutation_depth_by_task: dict[asyncio.Task[Any], int] = {}
 
     @asynccontextmanager
     async def mutation(self) -> AsyncIterator[None]:
         """Enter a commit/clear section, waiting only for an active checkpoint."""
+        task = asyncio.current_task()
+        if task is None:
+            raise RuntimeError("data-plane mutation must run inside an asyncio task")
+        depth = self._mutation_depth_by_task.get(task, 0)
+        if depth:
+            # Replay-buffer helpers may join a controller-owned mutation. Do not
+            # wait behind a checkpoint that is already waiting for this outer
+            # mutation, or the two tasks deadlock.
+            self._mutation_depth_by_task[task] = depth + 1
+            try:
+                yield
+            finally:
+                self._mutation_depth_by_task[task] -= 1
+            return
         async with self._condition:
             await self._condition.wait_for(lambda: not self._checkpoint_active)
             self._active_mutations += 1
+            self._mutation_depth_by_task[task] = 1
         try:
             yield
         finally:
+            del self._mutation_depth_by_task[task]
             async with self._condition:
                 self._active_mutations -= 1
                 if self._active_mutations == 0:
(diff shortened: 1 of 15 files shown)


tianyi-zhang-02:
@macandro96 — heads-up that this draft and my #3783 rewrite the same three lines in `SingleControllerActor.run()`, and I'd rather flag it now than have you hit it at rebase.

Both replace

```python
done, _ = await asyncio.wait(set(tasks), return_when=asyncio.FIRST_COMPLETED)
```

with a loop that keeps waiting on the still-pending set, closing the same window: the rollout pump exhausting normally is the *first* completion, so under the current code a watchdog raise during the train pump's drain lands on a task nobody awaits and gets swallowed by the `finally: gather(..., return_exceptions=True)`.

Two differences worth knowing before you rebase:

- **`probe_task`.** Your version has no `probe_task` branch — `grep probe_task` on this diff returns nothing — but `main` now creates one when `generation_fleet_health` is enabled (`single_controller.py:353-366`). As written the rewrite would drop that branch.
- **Ordering.** `for task in done: await task` iterates a set, so if the watchdog and a pump complete in the same `wait` batch, which exception surfaces is nondeterministic. #3783 keeps the explicit probe → watchdog → rollout → train order, which is what makes a stall get reported ahead of a pump whose only symptom is "waiting".

#3783 is 177/-16 on two files, based on `main`, mergeable, and adds `tests/unit/single_controller/test_run_supervision.py` covering the drain case and the probe path — this draft has no test of the supervision loop (its `test_single_controller.py` addition is a `_train_pump` test).

Suggestion, entirely yours to take or leave: let #3783 land the loop on its own, then drop your copy at rebase — it's incidental to the rollout-recovery feature this PR is actually about, and it'd stop gating a small fix behind a 1877-line stack. No new-file collision either way. :)

macandro96:
Will be superseded by https://github.com/NVIDIA-NeMo/RL/pull/3923

Issue #2909, opened 2026-06-24
Formalize the advantage-estimator interface (shared Protocol + explicit metrics, unified return type)


yfw:
**Is your feature request related to a problem? Please describe.**

The advantage estimators in `nemo_rl/algorithms/advantage_estimator.py` share no base class or `Protocol`, so the contract is entirely informal and fragile:

1. **Inconsistent return types.** `compute_advantage` returns a bare tensor in `GRPOAdvantageEstimator`, `GDPOAdvantageEstimator`, `ReinforcePlusPlusAdvantageEstimator`, and `OPDAdvantageEstimator`, but a `(advantages, returns)` tuple in `RawRewardAdvantageEstimator` and `GeneralizedAdvantageEstimator`. Callers must "know" which is which.
2. **Metrics via side-channel.** `OPDAdvantageEstimator` emits logging metrics as a side effect into `self.last_metrics` (`advantage_estimator.py:537,586`) rather than returning them, and the caller peeks at the attribute out-of-band, guarded by `hasattr` only because the other estimators don't define it (`grpo.py:3852-3856`):
   ```python
   if hasattr(adv_estimator, "last_metrics") and adv_estimator.last_metrics:
       rollout_metrics.update(adv_estimator.last_metrics)
   ```

Nothing enforces the attribute or return shapes, so adding an estimator means matching undocumented conventions.

**Describe the solution you'd like**

- A shared `AdvantageEstimator` base class / `Protocol` with an explicit metrics hook, e.g. `get_metrics() -> dict[str, float]` defaulting to `{}`. The call site becomes `rollout_metrics.update(adv_estimator.get_metrics())`.
- A unified return type, e.g. an `AdvantageResult(advantages, returns, metrics)` dataclass (`returns=None` where unused), removing the bare-tensor-vs-tuple split.

Touches all estimators in `advantage_estimator.py` plus the `compute_advantage` call sites in `grpo.py`.

**Additional context**

Follow-up from the MOPD PR #2780 review.

tianyi-zhang-02:
Opened #3512 with the proposed unified `AdvantageResult` and migrated the six estimators and their call sites. Metrics travel on the result instead of a separate hook; `_compute_gae` stays unchanged because it is private. @yfw, happy to adjust that shape during review.

PR #3512, opened 2026-08-06
refactor(algorithms): standardize advantage-estimator results

tianyi-zhang-02:
## What does this PR do?

Standardizes all six advantage estimators on one explicit result type:

```python
AdvantageResult(advantages, returns=None, metrics={})
```

GRPO, synchronous GRPO, PPO, and SingleController now consume named fields instead of branching on tensors versus tuples. OPD metrics travel in the result instead of through the `last_metrics` side channel.

The earlier `AdvantageEstimator` Protocol and its fake `None` defaults are removed. `GDPOAdvantageEstimator.repeated_batch` and `GeneralizedAdvantageEstimator.values` remain honestly required, while the typed estimator configs merged in #3773 stay unchanged.

Addresses #2909.

## Validation

Current head: `87d7322d23545ff51e4932cb379510d9c1735927`, including `main` at `fd7112c374af23b3bcc392637fb737495f8a3f31`.

On 2026-09-21, all five changed unit-test files passed on Runpod Secure Cloud: **363 passed**, 0 failed/skipped. Environment: 1× H100 80 GB, `nvcr.io/nvidia/nemo-rl:v0.7.0`, driver 580.126.09, CUDA 13.0, PyTorch 2.11.0+cu130. The current-main test fixture required `pytest==9.0.3` and pinned `nemo-lens` commit `b85578f`, installed only in the temporary Pod environment.

That run caught three tests added since the previous validation that still treated estimator returns as tensors/tuples. This PR now migrates those assertions to `.advantages` / `.returns`. Ruff check and format-check pass on both corrected test files.


tianyi-zhang-02:
Rebased onto current main and pushed a follow-up commit. Removing the `isinstance(result, tuple)` branch in `ppo.py` means every `compute_advantage` test double has to return an `AdvantageResult` too, and five were missed — four of them in tests that landed on main after this branch was cut:

- `tests/unit/algorithms/test_ppo.py` — `DummyAdvantageEstimator` returned a bare tuple, so `test_ppo_train_excludes_overlong_samples_from_advantage` and `test_ppo_train_wires_logprob_mask_to_advantage_training_and_metrics` failed with `AttributeError: 'tuple' object has no attribute 'advantages'`.
- `tests/unit/algorithms/test_grpo.py` — three `MagicMock` `return_value`s set to bare tensors.
- `tests/unit/single_controller/test_train_pump.py` — `_FakeAdvEstimator` returned a bare tensor; this branch had not touched that file at all.

Verified by running `test_ppo.py`, `test_grpo.py` and `test_advantage_estimator.py` against `origin/main` and against this branch and diffing the results: identical, with no failures introduced. Before the fix this branch had two failures that main did not.

Worth flagging one thing I noticed while doing this: the `AdvantageEstimator` `Protocol` added here has no reference anywhere else in the tree, so nothing is actually checked against it today. It is documentation rather than enforcement unless a call site is annotated with it — happy to either annotate one or drop the Protocol and keep just the dataclass, whichever you prefer.

tianyi-zhang-02:
Fixed something I should have caught before opening this: the protocol was defined and then referenced nowhere, and two estimators didn't satisfy it — `GDPOAdvantageEstimator` required `repeated_batch` and `GeneralizedAdvantageEstimator` required `values`, both positional and outside the contract. A protocol nothing is checked against is just a docstring that can drift.

Latest push makes it real:

- The contract's parameters are keyword-only, which is how both loops call it. Each passes the union of what all estimators need and leans on `**kwargs` for the rest, so "may accept extra arguments, may not require them" is the actual invariant.
- `repeated_batch` and `values` get keyword defaults with explicit errors. Both are still required in practice — the defaults only stop the signatures from demanding more than the shared contract. One side effect worth having: `ppo.py` adds `values` to the kwargs only `if "values" in train_data`, so GAE without a critic used to fail with a bare `TypeError` about a missing positional argument. It now says to enable the critic.
- Both `_create_advantage_estimator` functions are annotated with the protocol.

On the test: `runtime_checkable` only checks that `compute_advantage` exists, which every estimator passed even with an incompatible signature — so an `isinstance` test would have been close to vacuous. It inspects the signature instead and asserts no estimator requires anything outside the contract. Removing either default turns it red.

yuki-97 (commented):
hi @tianyi-zhang-02 , sorry for late reply and thanks for so many contributions! will assign reviewers.

@bg51717 to review

tianyi-zhang-02:
Small follow-up on my last push. The note explaining why `values` carries a default went into `_compute_gae`'s Args block instead of `compute_advantage`'s — and `_compute_gae` has neither the default nor the check the note describes (`values: torch.Tensor` is still required there, and the guard sits in `compute_advantage` *above* it, so "checked explicitly below" pointed the wrong way too).

Moved to `compute_advantage`, which turned out to have no Args block at all — which is how it ended up in the wrong place: the string I matched on was unique in the file, so it looked right.

Nothing about the behaviour changes; the docstring in a PR about documenting the estimator contract just shouldn't be the part that's wrong.

nemo_rl/algorithms/advantage_estimator.py
```diff
@@ -450,9 +505,30 @@ def compute_advantage(
...
         Returns:
-            Tuple of (advantages, returns), each of shape [batch_size, seq_len].
+            AdvantageResult with ``advantages`` and ``returns``, each of shape
+            [batch_size, seq_len].
         """
+        if values is None:
+            raise ValueError(
+                "GAE needs per-token value estimates; enable the critic so that "
+                "'values' is present in the training batch, or switch "
+                "grpo.adv_estimator.name to an estimator that needs no value model."
```
bg51717: Should this be `ppo.adv_estimator.name`? GAE is constructed from PPO's adv_estimator config.

nemo_rl/algorithms/ppo.py
```diff
@@ -1062,7 +1063,7 @@ def dynamic_sampling(
     return batch_to_return, is_batch_complete, batch_cache, dynamic_sampling_metrics
 
 
-def _create_advantage_estimator(master_config: MasterConfig):
+def _create_advantage_estimator(master_config: MasterConfig) -> AdvantageEstimator:
     """Create and return an advantage estimator based on configuration.
 
     PPO's training loop consumes a `(advantages, returns)` pair from a
```
bg51717: This still describes the old tuple return contract. Could we update it ?

nemo_rl/algorithms/advantage_estimator.py
```diff
@@ -118,7 +162,7 @@ def compute_advantage(
         prompt_ids,
         rewards,
         mask,
-        repeated_batch,
+        repeated_batch=None,
```
bg51717: @yuki-97, could you take a quick look at the API design here? The PR uses an `AdvantageResult` dataclass for the common return value and an `AdvantageEstimator` Protocol instead of a shared base class. To conform to this Protocol, `GDPOAdvantageEstimator`, which requires `repeated_batch`, now gives that argument a default value of `None` and checks at runtime whether a valid value was actually provided. I'm not sure whether this is a reasonable trade-off.

bg51717 (commented):
Thanks for the refactor. Unifying the estimator return type and removing the `last_metrics` side channel makes the call sites clearer. I left a few small comments.

nemo_rl/algorithms/advantage_estimator.py
tianyi-zhang-02: You are right — GAE is built in `ppo.py::_create_advantage_estimator` from `master_config.ppo["adv_estimator"]`, so `grpo.adv_estimator.name` sends the reader to a key that does not control it. Fixed to `ppo.adv_estimator.name`.

nemo_rl/algorithms/ppo.py
tianyi-zhang-02: Fixed. It now says that both supported estimators return an `AdvantageResult`, that `gae` populates `returns` while `raw_reward` leaves it None, and that the loop writes `train_data["returns"]` only when it is present.

I got this wrong once on the way: my first rewrite claimed both populate `returns`. `RawRewardAdvantageEstimator` returns `AdvantageResult(adv)` with `returns` left None, which is exactly why the loop still guards with `if result.returns is not None`.

nemo_rl/algorithms/advantage_estimator.py
tianyi-zhang-02: Fair question — the fake default is the part I like least too.

Here is why it is there. Both loops call `compute_advantage` with the union of what every estimator might need, and lean on `**kwargs` for the rest. So the real invariant is "may accept extra arguments, may not *require* any beyond prompt_ids/rewards/mask". `repeated_batch` and `values` were the only two breaking it. A protocol that two of six estimators fail is not enforcing much.

It did buy one concrete thing. `ppo.py` adds `values` only `if "values" in train_data`. So GAE without a critic was reachable, and it failed with a bare `TypeError` about a missing positional argument. Now it tells you to enable the critic. That is strictly better than the failure it replaced.

Three ways to go:

1. Leave it. Uniform signature, explicit error, docstring says it is required.
2. Drop the Protocol, keep `AdvantageResult`. Signatures stay honest, and the dataclass is the part carrying its weight anyway. This is what the PR was two pushes ago.
3. Pass one typed context object instead of loose kwargs. Then every estimator takes one argument honestly. I would want this long-term, but it touches every estimator and both loops, so it should be its own PR.

If the fake default is the sticking point, I would rather do 2 here and propose 3 separately. Your call :)

tianyi-zhang-02:
Also merged current main in, which surfaced a real break worth flagging: the async PPO loop added since this branch was cut still had the tuple branching this PR removes elsewhere —

```python
if isinstance(result, tuple):
    advantages, returns = result
else:
    advantages, returns = result, None
```

An `AdvantageResult` is not a tuple, so it took the `else` branch and handed the dataclass itself to `torch.masked_select`. `test_ppo_train_critic_warmup_reuses_generation_until_policy_update` fails on the merge and passes on main, so this was mine to fix, not a pre-existing failure. Now matches the sync path directly above it.

That is the second time a new `compute_advantage` consumer has landed while this sat — worth keeping in mind for whenever it merges. `tests/unit/algorithms/`, `tests/unit/single_controller/`: 859 passed, 54 skipped.

tianyi-zhang-02:
@bg51717 @yuki-97 — new information that bears on the API question at `advantage_estimator.py:165`, which I left as three options and which is still the thing holding this up.

#3773 (`feat(sc): support PPO in single controller`) rewrites the same file from the other side. It replaces `estimator_config: dict` with a pydantic `GAEConfig(BaseModel)` and turns `estimator_config["gae_lambda"]` into `estimator_config.gae_lambda`, and it narrows `AdvantageEstimatorConfig.name` to a `Literal`.

That is option 3 from my earlier reply — "pass one typed object instead of loose lookups" — applied to the config side. So the direction question looks answered: typed objects, not dicts. Which makes the return side of this PR (`AdvantageResult` instead of a bare tensor or a `(advantages, returns)` tuple) the matching half rather than a competing style.

It also means the two will conflict. Both rewrite the estimator `__init__`s and the `compute_advantage` signatures. #3773 is the feature and this has been open since 2026-08-06, so I would rather not make you resolve that: happy to rebase this onto #3773 once it settles, or to close this and hand you the `AdvantageResult` change to fold into #3773 directly if that is less churn. Whichever you prefer.

On the `repeated_batch=None` default that prompted the question: if #3773 lands the typed-config direction, option 3 stops being a big separate change and the fake default can just go away — every estimator would take its own typed config plus the shared tensors, and nothing would need a runtime "was this actually passed" check. I am happy to do that as the follow-up rather than defend the default.

tianyi-zhang-02:
Follow-up on the API concern from your review, on final SHA `6eb502a06122c27eb4763645a40a5175467f4427`:

- took option 2 from the thread: removed the unused Protocol and kept `AdvantageResult` as the common return contract;
- removed the fake `None` defaults, so GDPO's `repeated_batch` and GAE's `values` are required again;
- left #3773's typed estimator configs intact and migrated the GRPO/PPO/SingleController callers, including OPD metrics;
- Runpod validation: 14 targeted tests and 296 related CPU tests pass, 7 GPU-only tests skip, and Ruff check/format are clean.

cc @bg51717 — this directly resolves the trade-off you flagged without keeping a runtime-only signature contract.

PR #2518, opened 2026-05-18
refactor(grpo): migrate TypedDict configs to pydantic BaseModel

NolenLiang:
# What does this PR do ?

**Add a one line overview of what this PR aims to accomplish.**

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Results before / after the changes
image
image
image

PR #2325, opened 2026-04-23
refactor: update TypedDict to BaseModel (MasterConfig & ClippedPGLossConfig)

yuki-97:
## Related issues
1. https://github.com/NVIDIA-NeMo/RL/issues/1675
2. https://github.com/NVIDIA-NeMo/RL/issues/2102

## Summary

This PR refactors `MasterConfig` (in all algorithm entry points: `grpo.py`, `sft.py`, `dpo.py`, `rm.py`, `distillation.py`, `eval.py`) and `ClippedPGLossConfig` from `TypedDict` to `pydantic.BaseModel`.

We can now get four abilities when overriding config fields via Hydra/OmegaConf:

1. **Default values** — Fields omitted from the YAML use their Python-defined defaults.
   ```python
   # disable_ppo_ratio is defined with `= False` in ClippedPGLossConfig
   # but absent from examples/configs/grpo_math_1B.yaml — resolves to False
   loss_config.disable_ppo_ratio  # False
   ```
2. **Custom/extra fields** — Ad-hoc fields can be injected from the command line without modifying the config class.
   ```bash
   uv run python examples/run_grpo.py ++loss_fn.custom=1
   # cfg.custom is now accessible inside ClippedPGLossFn.__init__
   ```
3. **Type validation** — Pydantic validates field types at runtime.
   ```bash
   uv run python examples/run_grpo.py ++loss_fn.disable_ppo_ratio=1.2
   # Will raise error: "Input should be a valid boolean [type=bool_type, input_value=1.2, input_type=float]"
   ```
4. **Go-to-definition** — IDE navigation works on attribute access. Clicking on `config.disable_ppo_ratio` jumps directly to the field definition in `ClippedPGLossConfig`, whereas the previous dict-style access (`config["disable_ppo_ratio"]`) was opaque to static analysis tools.

## Changes

- `MasterConfig` in `grpo.py`, `sft.py`, `dpo.py`, `rm.py`, `distillation.py`, `eval.py`: `TypedDict` → `BaseModel(extra="allow")`
- `ClippedPGLossConfig` in `loss_functions.py`: `TypedDict` → `BaseModel(extra="allow")`
- All dict-style access (`config["key"]`) on these two config types updated to attribute access (`config.key`)
- `skills/config-conventions/SKILL.md` updated to reflect the new `BaseModel` convention

## Results before / after the changes

image

PR #3514, opened 2026-08-06
refactor(ppo): migrate config schema to BaseModel

### LINKED ISSUE #3462: refactor(ppo): remove performance metrics config compatibility after BaseModel migration
## Context

PR #2518 migrates the GRPO config from TypedDict to Pydantic BaseModel. `print_performance_metrics` is shared by GRPO and PPO, so it temporarily supports both attribute-style GRPO config and dict-style PPO config.

Once PPO config is migrated to BaseModel, remove this compatibility path and use direct typed attribute access throughout.

## Cleanup scope

- Replace the GRPO/PPO `getattr` algorithm detection in `print_performance_metrics` with typed config access.
- Remove the `isinstance(algo_config, dict)` branch for prompt and generation counts.
- Remove the remaining temporary top-level GRPO guard around async metrics if the shared API is split or typed accordingly.
- Update the corresponding unit tests.

Tracked from the review discussion on #2518: https://github.com/NVIDIA-NeMo/RL/pull/2518#discussion_r3692119732

fujial-code:
# What does this PR do ?

Migrate PPO configuration from the legacy `TypedDict` schema to Pydantic `BaseModel`, following the GRPO migration in #2518 and the config-v2 conventions established by #2325 and #2650.

The PR:

- converts `PPOConfig` to `BaseModel(extra="allow")` and moves PPO defaults into the schema;
- removes PPO's duplicate `AdvEstimatorConfig` and extends the shared config with the GAE/raw-reward fields used by PPO;
- converts `PPOSaveState` to a dataclass while preserving compatibility with partial and legacy checkpoint dictionaries;
- replaces PPO dict-style config access with typed attribute access;
- makes `print_performance_metrics` accept the algorithm config explicitly, removing the temporary GRPO/PPO config detection and dict compatibility path;
- updates GRPO, synchronous GRPO, and PPO callers of the shared metrics helper;
- updates PPO config/reference fixtures and adds coverage for nested model defaults, instance isolation, legacy checkpoint loading, and typed PPO performance metrics; and
- removes the temporary PPO Automodel metric comments and aligns its critic bounds with the PPO Megatron functional test.

This is intended to preserve existing YAML behavior. `extra="allow"` retains compatibility with older or extended configs, while schema-owned defaults and typed nested models remove hidden call-site defaults.

# Issues

Closes #3462

# Usage

No user-facing YAML migration is required. Existing PPO configs continue to load through `MasterConfig`; Python consumers now use typed attribute access, for example:

```python
num_prompts_per_step = master_config.ppo.num_prompts_per_step
```

# Validation

- `python -m py_compile`: passed
- `ruff format --check`: passed
- `ruff check`: passed
- `git diff --check`: passed
- PPO unit tests on H100: `42 passed`
- PPO config-v2/reference/example validation: `14 passed, 541 deselected`
- Shared performance metrics and full reference-config regression: `34 passed`
- Pyrefly baseline/after comparison: no after-only error signatures; the change removes 7 existing errors in the touched files
- PPO Megatron functional metrics: 9/9 assertions passed
- 70-step PPO baseline/after GPU smoke test: both runs finished all 70 steps with no NaN/Inf

## Results before / after the changes

Both runs used the same Qwen2.5 1.5B model, dataset/config, seed, and 70-step smoke-test setup. Each step used 2 prompts x 2 generations. The two runs executed concurrently on separate GPUs on the same node, so timing is used only to check for a regression, not to claim a speedup.

| Metric (70-step mean) | Baseline | After |
| --- | ---: | ---: |
| `train/reward` | -0.964286 | -0.964286 |
| `train/token_mult_prob_error` | 1.014539 | 1.014417 |
| `train/probs_ratio` | 1.000000000 | 1.000000001 |
| `train/critic/loss` | 0.061495 | 0.062201 |
| `timing/train/total_step_time` | 25.404 s | 24.619 s |

The reward distributions are identical: 66 steps at `-1.0`, 3 steps at `-0.5`, and 1 step at `0.0`. Reward is intentionally sparse in this small smoke test because four binary correctness rewards are scaled from `[0, 1]` to `[-1, 1]` and averaged per step.

- [Baseline W&B run](https://wandb.ai/nvidia/fujial-nemo-rl/runs/u760zjm3)
- [After-change W&B run](https://wandb.ai/nvidia/fujial-nemo-rl/runs/cdje974l)

## Known functional baseline limitation

The stricter Automodel `train/critic/grad_norm < 350` check does not currently pass on either side of the migration with the matching NeMo RL image:

- clean baseline maximum: `574.9854`
- after-change maximum: `972.4849`

Both Automodel runs complete training and write their metrics; the after-change functional script exits at the final metric assertion. The clean baseline passes the previous placeholder `<1500` bound but fails when the same `<350` bound is applied. The PPO Megatron functional test passes all assertions with the aligned bounds. This is therefore not introduced by the config migration, but it remains an explicit known red check while the stricter Automodel bound is included in this diff.

**Pre checks**:

- [x] Added or updated the necessary tests
- [x] Ran the relevant unit and functional tests locally/on a GPU node
- [x] No public documentation update is required for this internal schema migration

# Additional Information

The implementation follows the final reviewed shape of the GRPO config migration in #2518, including nested `default_factory` usage, reference-config updates, and runtime reconstruction of checkpoint state instead of relying on a type-only `cast()`.

### DIFF
--- a/examples/configs/ppo_math_1B.yaml
+++ b/examples/configs/ppo_math_1B.yaml
@@ -21,6 +21,7 @@ ppo:
   dynamic_sampling_max_gen_batches: 10
   batch_multiplier: 1
   skip_reference_policy_logprobs_calculation: true  # No KL, so skip ref logprobs
+  calculate_advantages_on_gpu: false
 
   reward_shaping:
     enabled: true
--- a/nemo_rl/algorithms/advantage_estimator.py
+++ b/nemo_rl/algorithms/advantage_estimator.py
@@ -42,7 +42,7 @@ from nemo_rl.algorithms.utils import (
 
 
 class AdvEstimatorConfig(BaseModel, extra="allow"):
-    """Configuration for advantage estimator (GRPO, GDPO, or Reinforce++)."""
+    """Configuration for GRPO- and PPO-compatible advantage estimators."""
 
     name: str = "grpo"  # "grpo", "gdpo", or "reinforce_plus_plus"
     # GRPO specific
@@ -52,6 +52,15 @@ class AdvEstimatorConfig(BaseModel, extra="allow"):
     reward_weights: list[float] | None = None
     # Reinforce++ specific
     minus_baseline: bool = True
+    # PPO GAE/raw-reward specific
+    gae_lambda: float = 0.95
+    gae_gamma: float = 1.0
+    normalize_advantages: bool = True
+    # VAPO decoupled GAE (None = standard GAE, no decoupling)
+    gae_lambda_value: float | None = None
+    gae_lambda_policy: float | None = None
+    # Length-adaptive lambda_policy = 1 - 1/(alpha * length). 0 = disabled.
+    length_adaptive_alpha: float = 0.0
 
 
 class GRPOAdvantageEstimator:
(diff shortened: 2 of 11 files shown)


fujial-code:
/ok to test https://github.com/NVIDIA-NeMo/RL/commit/d6db1154e6a579170cfca90098d005d53283e7c6

Issue #3288, opened 2026-07-20
feat: extend checkpoint-engine refit to SGLang (and Megatron) generation backends


terrykong:
## Summary

Checkpoint-engine refit (added in #2608) currently supports only the **vLLM** generation backend. Requests with a SGLang or Megatron generation backend are rejected at setup:

```python
# nemo_rl/algorithms/grpo.py
if checkpoint_engine_config is not None and backend != "vllm":
    raise NotImplementedError(
        "checkpoint-engine refit is only supported for the vLLM generation "
        f"backend, but policy.generation.backend={backend!r}. ..."
    )
```

This is a feature request to **extend checkpoint-engine refit (and the NIXL backend) to the other generation backends** — at minimum SGLang, and Megatron generation if applicable.

## Scope / clarification

The restriction is on the **generation** side only. Checkpoint-engine already supports **Megatron, DTensor, and DTensor v2 policy (training) backends** on the send side — so this request is specifically about the *receive* side (the generation engine) being limited to vLLM.

## Motivation

Users on SGLang rollouts cannot use the RDMA/NIXL non-colocated refit path and must fall back to the existing transports. A generation-backend-agnostic checkpoint-engine receive path would let SGLang (and any future backend) benefit from the same streamed, bucketed weight transfer.

## Notes

- The design doc already lists this under current limitations: "SGLang checkpoint-engine refit is not implemented." (`docs/design-docs/checkpoint-engines.md`).
- The `CheckpointEngine` transport abstraction is backend-agnostic; the work is on the generation-worker receive/apply side (an SGLang analog of `VllmCheckpointEngineMixin` / the worker-extension wiring).
- Tracking issue to be referenced from the setup `NotImplementedError` so users hitting the guard can follow progress (mirrors how #3275 tracks the PPO/distillation gap for sparse-delta refit).

_Filed as a follow-up to the #2608 review._

tianyi-zhang-02:
I'd like to take this. After tracing both backends I went with Option A: receive checkpoint-engine batches in the SGLang generation worker, then reuse the existing `flattened_bucket` CUDA-IPC handoff to the server process, so there is no second weight-ingest format to maintain. Dense and standard full-weight MoE are covered; the vLLM-specific `shard_expert_weights=true` path can follow separately.

Implementation is in #3519.

tianyi-zhang-02:
[...]

PR #3515, opened 2026-08-06
fix(policy): surface the uncovered-rank error during HTTP weight refit

tianyi-zhang-02:
## What does this PR do?

Makes the existing uncovered-rank error in `stream_weights_via_http_impl` reachable.

The function currently reads cached IPC topology before resolving the rollout-engine URL, so an uncovered rank raises `KeyError: 'gather_src'`. This moves those reads below URL resolution, allowing the intended `RuntimeError` to report the bad rank mapping.

Covered-rank behavior is unchanged.

## Testing

- Added uncovered-rank and covered-rank regression tests: 2 passed locally
- `ruff check`, `ruff format`, and `pyrefly` pass
- Three existing ZMQ IPC tests fail identically on `main` and this branch on Windows because `ipc://` is unsupported

## Issues

Found while investigating #3288. Draft PRs #3330 and #3426 also touch this function.

cc @RayenTian @yuki-97


yuki-97 (commented):
@bg51717 to review

bg51717 (approved):
LGTM. This change ensures uncovered policy ranks raise the intended `RuntimeError` instead of a `KeyError`, with tests covering both covered and uncovered rank paths.

bg51717:
Merged the latest upstream/main into this branch and removed the redundant explanatory comment.

bg51717:
/ok to test [78f147d](https://github.com/NVIDIA-NeMo/RL/pull/3515/commits/78f147d6913a61d8e9c79379fc3560ffcc123da5)

tianyi-zhang-02:
Two things on the red checks.

`L0_Unit_Tests_Sglang` was flaky, not a real failure. That run ended with

```
=========== 79 passed, 10 skipped, 61 warnings in 1026.58s (0:17:06) ===========
./tests/run_unit.sh: line 44: 20394 Segmentation fault (core dumped) pytest $pytest_args
```

All 79 passed. The segfault is at interpreter shutdown, after the summary line. The job is judged by `tail -n 1 err.log | grep "Finished successfully."`, and that never gets written when the process dies on a signal, so the whole shard goes red. `CI quality check` is just the roll-up of it. Nothing here runs at shutdown — the PR moves two reads inside `stream_weights_via_http_impl`.

I also brought the branch up to date. Should be clean merge now.

Needs a fresh `/ok to test` when convenient :)

tianyi-zhang-02:
Closing this — main fixed it, and better than I did.

`stream_weights_via_http_impl` was rewritten since I opened this. It now reads the topology with `.get()` instead of `[...]`, and an uncovered rank takes an explicit placeholder branch that drains the bucket iterator and returns. That's the right answer: the iterator lazily drives `export_hf_weights`, whose PP broadcast and TP gather are collective over the whole trainer world, so raising there — which is what this PR did — would have hung the other ranks.

Thanks for the review @bg51717 and @yuki-97, sorry it went stale. :)

PR #3516, opened 2026-08-06
fix(sglang): expose cfg on SGLangGeneration

tianyi-zhang-02:
## What does this PR do?

Adds a read-only `cfg` property to `SGLangGeneration`, aliasing its existing `sglang_cfg`.

`create_weight_synchronizer` reads `generation.cfg` before backend dispatch. Without this property, colocated SGLang raises `AttributeError` instead of constructing its `HTTPWeightSynchronizer`.

## Testing

- Added a CPU-only factory regression test using the real `SGLangGeneration` class
- Test passes locally
- `ruff check` and `ruff format` pass

## Issues

Found while investigating #3288. Draft PR #3330 contains the same property as part of a larger change.

cc @RayenTian @yuki-97

### DIFF
--- a/nemo_rl/models/generation/sglang/sglang_generation.py
+++ b/nemo_rl/models/generation/sglang/sglang_generation.py
@@ -122,6 +122,17 @@ class SGLangGeneration(GenerationInterface):
         if init_handles:
             ray.get(init_handles)
 
+    @property
+    def cfg(self) -> SGLangConfig:
+        """Full generation config, matching the ``GenerationInterface`` contract.
+
+        ``sglang_cfg`` already is that config, so alias it rather than keep a
+        second reference. Backend-agnostic callers such as
+        ``nemo_rl.weight_sync.factory.create_weight_synchronizer`` read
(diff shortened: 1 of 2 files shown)


tianyi-zhang-02:
Closing this in favour of #3519, which carries the identical commit as its first commit — the `cfg` property and `test_colocated_sglang_accepts_real_generation_object` are both in that branch verbatim, so nothing is lost by closing this and it is one less PR in the queue.

The underlying bug is unchanged and still worth fixing: `create_weight_synchronizer` reads `generation.cfg` before any backend dispatch (`nemo_rl/weight_sync/factory.py:69`), while `SGLangGeneration` only sets `self.sglang_cfg`, so the single-controller path raises `AttributeError` for `backend=sglang`. vLLM, TRT-LLM and Megatron all expose `cfg`; `megatron_generation.py:96` documents it as the `GenerationInterface` contract.

Happy to reopen as a standalone five-line PR if you would rather land it separately from the checkpoint-engine work — just say so.

PR #3276, opened 2026-07-17
feat: ultra recipes for gb200

macandro96:
# What does this PR do ?

Upstreaming ultra recipes for main

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #3529, opened 2026-08-06
feat(ultra): serve GenRM and NL2Bash as external vLLM pools

macandro96:
Add `EXTERNAL_JUDGES=1` to `ultra_launch.sh`, which registers the GenRM and NL2Bash judges as fixed-model vLLM pools in a second Slurm heterogeneous component instead of serving them inside the training Ray cluster. The allocation wrapper starts each replica, brings up one load balancer per pool, waits for health, and substitutes the resolved endpoints into the driver command before ray.sub starts on the training component.

Also add `NL2BASH_BASE_URL` support for routing that judge at an already running service.

Set `moe_backend=flashinfer_cutlass` on the policy vLLM across the Ultra stage configs, required by the vLLM 0.25 bump.

# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/docs/guides/nemotron-3-ultra.md
+++ b/docs/guides/nemotron-3-ultra.md
@@ -251,6 +251,18 @@ pools serve, and Gym addresses them by `GENRM_SERVED_MODEL_NAME` /
 `GENRM_BASE_URL` or `NL2BASH_BASE_URL` by hand is rejected, since the wrapper
 supplies those.
 
+Each pool is registered only when its model variable is set, so set at least
+one of them. Stages declare only the judges they use — `rlhf_teacher` has no
+NL2Bash block, `reasoning_teacher` has no GenRM block, `swe_teacher` has
+neither — and setting a judge the stage does not declare makes the driver fail
+on an unknown config key.
+
+`BASE_LOG_DIR` and `EXTERNAL_VLLM_TOOLS_DIR_HOST` (the repo's
+`tools/external_gym_vllm`) must resolve under `EXTERNAL_VLLM_SHARED_ROOT`,
+because the wrapper bind-mounts that root into the service containers. Since
+`RESULTS_DIR` defaults to a path relative to the repo, pass an explicit
+`BASE_LOG_DIR` on the shared filesystem when the checkout lives elsewhere.
+
 | Variable | Default | Purpose |
 |---|---|---|
 | `EXTERNAL_JUDGES` | `0` | `1` serves GenRM and NL2Bash in a second hetgroup |
@@ -270,20 +282,31 @@ plugin, and `BASE_LOG_DIR` must live under `EXTERNAL_VLLM_SHARED_ROOT`
 (`/lustre` by default), which is mounted into the service containers.
 `INTERACTIVE=1` is not supported in this mode.
 
+Building on the [Phase 1](#phase-1--49k-context-128-steps) invocation, add:
+
 ```bash
 EXTERNAL_JUDGES=1 \
 GENRM_MODEL=/path/to/genrm-checkpoint \
-GENRM_REASONING_PARSER=/path/to/ultra_v3_reasoning_parser.py \
+GENRM_REASONING_PARSER=/lustre/path/to/ultra_v3_reasoning_parser.py \
 GENRM_REPLICAS=16 \
 NL2BASH_JUDGE_MODEL=/path/to/nl2bash-judge-checkpoint \
 NL2BASH_REPLICAS=4 \
-NUM_GYM_NODES=2 \
-# ... plus the required variables above
+NUM_TRAIN_NODES=64 \
+NUM_GEN_NODES=172 \
+NUM_GYM_NODES=4 \
+BASE_LOG_DIR=/lustre/path/to/ray_logs \
+EXP_NAME=... CONFIG_PATH=... MODEL_PATH=... TRAIN_PATH=... VAL_PATH=... \
+CONTAINER=... SANDBOX_CONTAINER=... PERSISTENT_CACHE=... \
+SLURM_PARTITION=$SLURM_PARTITION SLURM_ACCOUNT=$SLURM_ACCOUNT \
+SAFETY_JUDGE_MODEL=nvidia/Nemotron-Content-Safety-Reasoning-4B \
 bash examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh
 ```
 
-That example allocates 16 GenRM + 4 NL2Bash = 20 external nodes alongside the
-training component.
+That allocates 64 + 172 + 4 = 240 NeMo RL nodes (a multiple of `SEGMENT_SIZE`,
+with Gym now hosting only the safety judge) plus 16 GenRM + 4 NL2Bash = 20
+external nodes. Set only `GENRM_MODEL` or only `NL2BASH_JUDGE_MODEL` to move
+just that judge out of Gym — useful for the teacher stages, which declare only
+the judges they use.
 
 ## Stage 1 — Student RLVR
 
--- a/docs/guides/nemotron-3-ultra.md
+++ b/docs/guides/nemotron-3-ultra.md
@@ -201,6 +201,7 @@ Set the following before each `bash examples/nemo_gym/nemotron-3-ultra/ultra_lau
 | `GENRM_MODEL` | GenRM judge: HF repo id or local path. Default is [`nvidia/NVIDIA-Nemotron-3-Ultra-550B-A55B-GenRM`](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Ultra-550B-A55B-GenRM). Served in-cluster from the gym pool at TP=4 (DP varies by stage). Required unless `GENRM_BASE_URL` is set. |
 | `GENRM_BASE_URL` | _Optional._ URL of a separately-deployed GenRM service (e.g. `http://genrm-host:9213/v1`). If set, routes judging to that endpoint and ignores `GENRM_MODEL`. Useful when sharing a single GenRM deployment across many training jobs. |
 | `NL2BASH_JUDGE_MODEL` | NL2Bash / general-purpose judge: HF repo id or local path. Default judge is `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8`. |
+| `NL2BASH_BASE_URL` | _Optional._ URL of a separately-deployed NL2Bash judge service. Same semantics as `GENRM_BASE_URL`. |
 | `SAFETY_JUDGE_MODEL` | Content-safety judge: HF repo id or local path. Default is [`nvidia/Nemotron-Content-Safety-Reasoning-4B`](https://huggingface.co/nvidia/Nemotron-Content-Safety-Reasoning-4B). |
 
 > **Serving GenRM outside Gym.** For a separately deployed OpenAI-compatible
@@ -209,10 +210,8 @@ Set the following before each `bash examples/nemo_gym/nemotron-3-ultra/ultra_lau
 > deployment to back multiple training runs.
 >
 > To co-schedule dedicated model servers with training in one Slurm
-> heterogeneous allocation, use the
-> [external Gym vLLM pool helpers](https://github.com/NVIDIA-NeMo/RL/blob/main/tools/external_gym_vllm/README.md).
-> They place the server fleet in a second hetgroup, start load balancers on the
-> training component, and inject the resolved endpoints into the driver command.
+> heterogeneous allocation, set `EXTERNAL_JUDGES=1` — see
+> [External judge services](#external-judge-services) below.
 
 Optional knobs:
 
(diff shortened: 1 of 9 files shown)


examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh
```diff
@@ -117,6 +128,171 @@ fi
...
+  export EXTERNAL_VLLM_LB_PYTHON EXTERNAL_VLLM_POOLS EXTERNAL_VLLM_TOOLS_DIR_HOST
+fi
+
+# Judge endpoints: a base_url routes the Gym server at a remote service and
+# skips the local deployment; otherwise Gym serves the model itself.
+GENRM_OVERRIDE=""
+if [[ -n "${GENRM_BASE_URL}" ]]; then
+  GENRM_OVERRIDE="++env.nemo_gym.genrm_model.responses_api_models.genrm_model.base_url=${GENRM_BASE_URL}"
+  if [[ -n "${GENRM_API_MODEL_NAME}" ]]; then
+    GENRM_OVERRIDE="${GENRM_OVERRIDE} env.nemo_gym.genrm_model.responses_api_models.genrm_model.model=${GENRM_API_MODEL_NAME}"
```
yfw, line 305: [`ultra_launch.sh:281`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L281)

**`EXTERNAL_JUDGES=1` aborts the driver on 3 of the 7 stage configs — including `rlhf_teacher`, which is the case this feature most exists for.**

[`:149-150`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L149-L150) make *both* judges mandatory, [`:160`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L160)/[`:177`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L177) set both `*_BASE_URL` to placeholders, and [`:163-164`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L163-L164)/[`:180-181`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L180-L181) default `*_API_MODEL_NAME` to the served name `model` — so it is never empty and this line (and [`:290`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L290)) always appends a **bare**, non-`++` override.

Overrides go through real Hydra against a struct-mode config ([`config.py:186-189`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/nemo_rl/utils/config.py#L186-L189)). Feeding the exact four overrides this mode emits into the repo's own `parse_hydra_overrides`, against each real stage config:

```
swe_teacher          OverridesError: Could not override 'env.nemo_gym.genrm_model...'
rlhf_teacher         OverridesError: Could not override 'env.nemo_gym.nl2bash_judge_model...'
reasoning_teacher    OverridesError: Could not override 'env.nemo_gym.genrm_model...'
student_rlvr1        OK
```

which lines up with which judge blocks each stage declares:

| stage config | `genrm_model` | `nl2bash_judge_model` |
|---|---|---|
| student_rlvr1 / student_rlvr2 / ifbench_teacher / mopd | yes | yes |
| reasoning_teacher | **absent** | yes |
| rlhf_teacher | yes | **absent** |
| swe_teacher | **absent** | **absent** |

`swe_teacher` is arguably academic (it uses no judges), but `rlhf_teacher` is the opposite — it uses GenRM and not NL2Bash, and offloading the 550B GenRM is the motivating use case for this feature.

**Nothing catches it at submit time**: `EXTERNAL_JUDGES=1 CONFIG_PATH=.../swe_teacher.yaml DRY_RUN=1` exits **0**, emitting both placeholders. The failure lands in the driver, after both hetgroups are allocated, every replica has loaded its checkpoint, and the wrapper has waited for health and substituted the URLs.

This is also the opt-out the launcher documents at [`:96-98`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L96-L98) — *"The SWE teacher uses code-execution rewards and needs none of them. Set per recipe; unset vars skip the corresponding override."* `SAFETY_JUDGE_MODEL` still keeps exactly that escape hatch at [`:845`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L845), which is why `swe_teacher` works today and only breaks under this flag.

`++` on both lines would stop the crash but is worse: force-add fabricates a dead `genrm_model` subtree on a config with no GenRM server, and you still allocate and health-wait on a pool the stage never queries. Gating each pool on whether the stage declares that judge follows the precedent already in this file ([`:121-123`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L121-L123) greps `CONFIG_PATH` for `${sif_dir}`):

```bash
EXTERNAL_GENRM=0;   grep -q 'genrm_model:'         "${CONFIG_PATH}" && EXTERNAL_GENRM=1
EXTERNAL_NL2BASH=0; grep -q 'nl2bash_judge_model:' "${CONFIG_PATH}" && EXTERNAL_NL2BASH=1
if (( EXTERNAL_GENRM == 0 && EXTERNAL_NL2BASH == 0 )); then
  echo "ERROR: EXTERNAL_JUDGES=1 but ${CONFIG_PATH} declares no external-servable judge." >&2
  exit 1
fi
```

with each judge's `:?` requirement, registration and override emission moved inside its own guard. `rlhf_teacher` then gets an external GenRM without a wasted 4-node NL2Bash pool.

examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh
```diff
@@ -653,6 +855,9 @@ ${MOPD_OVERRIDES} \
 ${*}"
 
 export COMMAND="${TRAIN_CMD}"
+if (( NUM_EXTERNAL_SERVICE_NODES > 0 )); then
+  validate_external_vllm_submission "${COMMAND}" "${NUM_EXTERNAL_SERVICE_NODES}"
```
yfw, line 904: [`ultra_launch.sh:859`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L859)

**`EXTERNAL_JUDGES=1` can't be submitted with the launcher's own defaults.** Reproduced independently by three reviewers.

This call requires `BASE_LOG_DIR` and `EXTERNAL_VLLM_TOOLS_DIR_HOST` to be absolute *and* under `EXTERNAL_VLLM_SHARED_ROOT` ([`pool_config.sh:262-269`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/tools/external_gym_vllm/pool_config.sh#L262-L269) -> [`:91-101`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/tools/external_gym_vllm/pool_config.sh#L91-L101)), but both launcher defaults violate it:

- [`:307`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L307) `RESULTS_DIR="${RESULTS_DIR:-results/${EXP_NAME}}"` -> [`:320`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L320) `BASE_LOG_DIR="${BASE_LOG_DIR:-${RESULTS_DIR}/ray_logs}"` — **relative**.
- [`:198`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L198) `EXTERNAL_VLLM_TOOLS_DIR_HOST` defaults to `${PROJECT_ROOT}/tools/external_gym_vllm` — so the repo checkout itself must live under `/lustre`.

Observed:
```
ERROR: BASE_LOG_DIR must be an absolute path (got 'results/<exp>/ray_logs')
ERROR: EXTERNAL_VLLM_TOOLS_DIR_HOST must be under /lustre (got '<repo>/tools/external_gym_vllm')
```

Credit where due: it fails at [`:859`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L859), **before** the `sbatch` at [`:1024`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L1024), so no allocation is wasted, and the messages name the offending variable exactly. The `BASE_LOG_DIR` constraint is also stated in the prose this PR adds ([`docs:269`](docs/guides/nemotron-3-ultra.md#L269)).

What makes it worth fixing rather than documenting: the *default* invocation can never satisfy it, so nobody can turn this feature on by following the guide. Two things would close it — validate both paths inside the `EXTERNAL_JUDGES=1` block near [`:148`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L148) (before the run dirs, the cache purge and the code snapshot at [`:672`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L672), so a failed attempt doesn't leave a snapshot behind), and add `BASE_LOG_DIR=/lustre/...` to the docs example. The checkout-must-be-under-`/lustre` requirement isn't documented anywhere yet either — [`docs:269`](docs/guides/nemotron-3-ultra.md#L269) lists model paths, the parser plugin and `BASE_LOG_DIR`, but not the repo.

examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh
```diff
@@ -84,15 +98,12 @@ set -euo pipefail
 # code-execution rewards and needs none of them. Set per recipe; unset vars
 # skip the corresponding override.
 NL2BASH_JUDGE_MODEL="${NL2BASH_JUDGE_MODEL:-}"
+NL2BASH_BASE_URL="${NL2BASH_BASE_URL:-}"
+NL2BASH_API_MODEL_NAME="${NL2BASH_API_MODEL_NAME:-}"
```
yfw, line 102: [`ultra_launch.sh:102`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L102)

**The `NL2BASH_BASE_URL` knob this PR adds can't work end-to-end — Gym gets `model: null` and fails config validation.**

The `++` fix on `base_url` is real and correct. But `*_API_MODEL_NAME` get the served-name default only *inside* the `EXTERNAL_JUDGES=1` block ([`:163-164`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L163-L164), [`:180-181`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L180-L181)). On the hand-set `*_BASE_URL` path they stay empty, so no `model=` override is emitted ([`:280-282`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L280-L282), [`:289-291`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L289-L291)) and the YAML's `model: null` survives ([`student_rlvr1.yaml:601`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/student_rlvr1.yaml#L601), [`:510`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/student_rlvr1.yaml#L510)).

Gym requires it. `VLLMModelConfig` declares `model: str` with no default, and the subclasses actually used re-declare only `base_url`/`api_key` — `model` is inherited un-overridden:
- [`vllm_model/app.py#L48-L51`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/vllm_model/app.py#L48-L51) — `model: str`
- [`local_vllm_model/app.py#L45-L49`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/local_vllm_model/app.py#L45-L49) — overrides `base_url` and `api_key` only
- [`genrm_model/app.py#L82-L85`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_models/genrm_model/app.py#L82-L85) — adds only `supports_principle_role`

Setting `GENRM_MODEL` alongside doesn't help — [`:278-285`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L278-L285) is `if/elif`, and the docs say `GENRM_BASE_URL` "ignores `GENRM_MODEL`".

Fair framing on scope: for `GENRM_BASE_URL` this is inherited, not introduced — `ultra_launch.sh` has only two commits, and in the first (`64cb9f985`, #3276) the override was already bare while the YAML already had `model: null`, so that path died earlier in Hydra and nobody ever reached the pydantic error. This PR fixes that first half; the ask is only that `NL2BASH_BASE_URL`, which it newly documents, ships working:

```suggestion
NL2BASH_API_MODEL_NAME="${NL2BASH_API_MODEL_NAME:-${NL2BASH_JUDGE_MODEL}}"
```

(same for `GENRM_API_MODEL_NAME` at [`:106`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh#L106) from `${GENRM_MODEL}`). Established by source inspection, not by running a Gym server.

docs/guides/nemotron-3-ultra.md
```diff
@@ -228,6 +227,64 @@ Optional knobs:
...
+
+```bash
+EXTERNAL_JUDGES=1 \
+GENRM_MODEL=/path/to/genrm-checkpoint \
+GENRM_REASONING_PARSER=/path/to/ultra_v3_reasoning_parser.py \
+GENRM_REPLICAS=16 \
+NL2BASH_JUDGE_MODEL=/path/to/nl2bash-judge-checkpoint \
+NL2BASH_REPLICAS=4 \
+NUM_GYM_NODES=2 \
+# ... plus the required variables above
```
yfw: [`nemotron-3-ultra.md:281`](https://github.com/NVIDIA-NeMo/RL/blob/cb135591da019a4f24eefcb5812a1ebb8e89dc58/docs/guides/nemotron-3-ultra.md#L281)

**This example silently launches with none of its variables set, so following the guide gives you a normal in-Gym run rather than external judges.**

The `#` comment sits inside a backslash-continuation. Bash removes the backslash-newline, `#` then terminates that logical line, and `bash ultra_launch.sh` on the next line runs as a *separate* command:

```
$ bash -c 'FOO=1 \
BAR=2 \
# comment
env | grep -E "^(FOO|BAR)=" || echo "NO ENV VARS SET"'
NO ENV VARS SET
```

No error, no warning — `EXTERNAL_JUDGES` is simply never set.

And once that's fixed the node counts fail the launcher's own check, because setting `NUM_GYM_NODES=2` alone leaves the other two at their defaults (`64 + 172 + 2 = 238`, and `238 % 16 = 14`):

```
ERROR: NeMo RL nodes=238 is not divisible by SEGMENT_SIZE=16.
  Training=64 + Generation=172 + Gym=2 + Teachers=0 = 238
```

The "plus the required variables above" note doesn't cover this — the required table at [`:189-205`](docs/guides/nemotron-3-ultra.md#L189-L205) has no `NUM_*` row; node counts appear only in the optional table at [`:222`](docs/guides/nemotron-3-ultra.md#L222) with defaults `64, 172, 20`, and the divisibility rule is restated in prose 15 lines above this snippet. Every other launch example in the guide sets the full triple and lands on a legal total (80, 80, 128, 128, and 224 once MOPD's teacher nodes count), so matching that convention fixes both problems at once:

```suggestion
NUM_TRAIN_NODES=64 \
NUM_GEN_NODES=172 \
NUM_GYM_NODES=4 \
bash examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh
```

(plus `BASE_LOG_DIR=/lustre/...`, per the comment on the validation call.)

macandro96:
/ok to test [5c90fe8](https://github.com/NVIDIA-NeMo/RL/pull/3529/commits/5c90fe8b041bb8da141938ff639d567dbc4c6673)

docs/guides/nemotron-3-ultra.md
macandro96: addressed in [86137f0](https://github.com/NVIDIA-NeMo/RL/pull/3529/commits/86137f0640cb8a4ee6bf56e65c1a7b0bc0ff7679)

examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh
macandro96: addressed in [86137f0](https://github.com/NVIDIA-NeMo/RL/pull/3529/commits/86137f0640cb8a4ee6bf56e65c1a7b0bc0ff7679)

examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh
macandro96: addressed in [86137f0](https://github.com/NVIDIA-NeMo/RL/pull/3529/commits/86137f0640cb8a4ee6bf56e65c1a7b0bc0ff7679)

examples/nemo_gym/nemotron-3-ultra/ultra_launch.sh
macandro96: addressed in [86137f0](https://github.com/NVIDIA-NeMo/RL/pull/3529/commits/86137f0640cb8a4ee6bf56e65c1a7b0bc0ff7679)

yfw (approved):

PR #3532, opened 2026-08-07
fix: mtp gating for the nccl-reshard-refit and fix the error when doing ray RPC.

youngeunkwon0405:
# What does this PR do ?

This PR is doing two independent things. I split this PR into two.

1. Gate the MTP layers from the nccl-reshard path. It will force the use of the conventional refit path for the MTP layer.

Goes to #3619

2. Make the `nccl_reshard_refit_info` object wire-safe before sending it to the generation worker during initialization. 
`nccl_reshard_refit_info` contains torch tensors, created by Megatron. The problem is that Megatron replaces the
`torch.storage._load_from_bytes` function that is required for unpickling to a Megatron-specific function. 
So, during unpickling from the generation worker, it will call `import megatron` which will cause an import error.
This PR fixes that issue.

Goes to #3618

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -217,6 +217,8 @@ class SingleControllerActor:
             await asyncio.gather(rollout_task, train_task, return_exceptions=True)
             try:
                 self._weight_synchronizer.shutdown()
+            except Exception as e:  # teardown must not mask the original failure
+                print(f"Error during weight-synchronizer shutdown: {e}", flush=True)
             finally:
                 self._logger.finish()
 
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -207,7 +207,10 @@ class SingleControllerActor:
             rollout_task.cancel()
             train_task.cancel()
             await asyncio.gather(rollout_task, train_task, return_exceptions=True)
-            self._logger.finish()
+            try:
+                self._weight_synchronizer.shutdown()
+            finally:
+                self._logger.finish()
 
         return {
             "train_steps": self._train_steps,
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -215,7 +215,10 @@ class SingleControllerActor:
             rollout_task.cancel()
             train_task.cancel()
             await asyncio.gather(rollout_task, train_task, return_exceptions=True)
-            self._logger.finish()
+            try:
+                self._weight_synchronizer.shutdown()
+            finally:
+                self._logger.finish()
 
         return {
             "train_steps": self._train_steps,
--- a/nemo_rl/weight_sync/nccl_reshard_utils.py
+++ b/nemo_rl/weight_sync/nccl_reshard_utils.py
@@ -191,8 +191,12 @@ def is_nccl_reshard_param(param_name: str) -> bool:
     ``load_weights`` path.
 
     Shared-expert FFN weights (``*.shared_expert.*``) are routed to misc path.
+    Co-trained MTP weights are also routed to misc because vLLM keeps the MTP
+    drafter separate from the main model and updates it through ``load_weights``.
     """
     if "shared_expert" in param_name:
+        return False
+    if param_name.startswith("mtp."):
         return False
     return param_name.endswith(FFN_PROJ_WEIGHT_SUFFIXES) or param_name.endswith(
         FFN_GROUPED_EXPERT_SUFFIXES
@@ -245,6 +249,44 @@ def _restore_placement(p):
             return Shard(p["dim"])
         return Replicate()
     return Replicate()
+
+
+def make_nccl_reshard_refit_info_wire_safe(refit_info: dict) -> dict:
+    """Copy refit metadata into types safe for vLLM's subprocess RPC.
+
+    Megatron patches Torch's tensor pickle reducer to use its safe_globals
+    loader. vLLM worker subprocesses intentionally do not install Megatron, so
+    even the small CPU rank tensors in MeshInfo cannot cross that pickle
+    boundary. Convert meshes and placements to the plain representation that
+    restore_refit_info_placements accepts.
+    """
+
+    def _wire_mesh(mesh):
+        if isinstance(mesh, MeshInfo):
+            return {"mesh": mesh.mesh.tolist()}
+        return mesh
+
+    def _wire_placement(placement):
+        if isinstance(placement, Shard):
+            return {"dim": placement.dim}
+        if isinstance(placement, Replicate):
+            return {}
+        return placement
+
+    wire_info = dict(refit_info)
+    wire_layers = {}
+    for layer_name, params in refit_info.get("per_layer_params", {}).items():
+        wire_params = []
+        for param_info in params:
+            wire_param = dict(param_info)
+            for key in ("src_mesh_info", "dst_mesh_info"):
+                wire_param[key] = _wire_mesh(wire_param.get(key))
+            for key in ("src_placements", "dst_placements"):
+                wire_param[key] = [_wire_placement(p) for p in wire_param.get(key, [])]
+            wire_params.append(wire_param)
+        wire_layers[layer_name] = wire_params
+    wire_info["per_layer_params"] = wire_layers
+    return wire_info
 
 
 def restore_refit_info_placements(refit_info: dict) -> dict:
--- a/nemo_rl/weight_sync/nccl_reshard_utils.py
+++ b/nemo_rl/weight_sync/nccl_reshard_utils.py
@@ -254,11 +254,10 @@ def _restore_placement(p):
 def make_nccl_reshard_refit_info_wire_safe(refit_info: dict) -> dict:
     """Copy refit metadata into types safe for vLLM's subprocess RPC.
 
-    Megatron patches Torch's tensor pickle reducer to use its safe_globals
-    loader. vLLM worker subprocesses intentionally do not install Megatron, so
-    even the small CPU rank tensors in MeshInfo cannot cross that pickle
-    boundary. Convert meshes and placements to the plain representation that
-    restore_refit_info_placements accepts.
+    Importing ``megatron.core`` replaces ``torch.storage._load_from_bytes`` with
+    a Megatron function, so Tensor pickles require Megatron at unpickle time.
+    vLLM subprocesses may not have it importable; convert the metadata to the
+    plain representation accepted by ``restore_refit_info_placements``.
     """
 
     def _wire_mesh(mesh):
--- a/nemo_rl/weight_sync/nccl_reshard_utils.py
+++ b/nemo_rl/weight_sync/nccl_reshard_utils.py
@@ -19,6 +19,8 @@ This module provides:
   shared torch.distributed process group (needed for cross-world transfers)
 - Placement rules: mapping param names to TP/EP sharding strategies
 - build_nccl_reshard_refit_info: compute per-layer param metadata for refit
+- make_nccl_reshard_refit_info_wire_safe: convert placements and meshes into
+  plain dicts/lists before the refit metadata crosses a process boundary
 - restore_refit_info_placements: undo msgspec dict-flattening of placements
   and meshes on the receiving side
 
@@ -279,9 +281,11 @@ def make_nccl_reshard_refit_info_wire_safe(refit_info: dict) -> dict:
         for param_info in params:
             wire_param = dict(param_info)
             for key in ("src_mesh_info", "dst_mesh_info"):
-                wire_param[key] = _wire_mesh(wire_param.get(key))
+                if key in wire_param:
+                    wire_param[key] = _wire_mesh(wire_param[key])
             for key in ("src_placements", "dst_placements"):
-                wire_param[key] = [_wire_placement(p) for p in wire_param.get(key, [])]
+                if key in wire_param:
+                    wire_param[key] = [_wire_placement(p) for p in wire_param[key]]
             wire_params.append(wire_param)
         wire_layers[layer_name] = wire_params
     wire_info["per_layer_params"] = wire_layers
--- a/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
+++ b/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
@@ -210,6 +210,8 @@ class NcclReshardWeightSynchronizer(WeightSynchronizer):
             make_nccl_reshard_refit_info_wire_safe,
         )
 
+        # SingleController sends this across vLLM's subprocess boundary, so avoid
+        # Megatron-dependent Tensor pickles.
         wire_refit_info = make_nccl_reshard_refit_info_wire_safe(
             nccl_reshard_refit_info
         )
--- a/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
+++ b/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
@@ -210,8 +210,10 @@ class NcclReshardWeightSynchronizer(WeightSynchronizer):
             make_nccl_reshard_refit_info_wire_safe,
         )
 
-        # SingleController sends this across vLLM's subprocess boundary, so avoid
-        # Megatron-dependent Tensor pickles.
+        # nccl_reshard_refit_info contains Torch Tensors that are created by
+        # Megatron, which will invoke `import megatron` when unpickled.
+        # To avoid this, we remove the Torch Tensors and restore them later,
+        # via `restore_nccl_reshard_refit_info()`.
         wire_refit_info = make_nccl_reshard_refit_info_wire_safe(
             nccl_reshard_refit_info
         )
--- a/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
+++ b/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
@@ -210,10 +210,11 @@ class NcclReshardWeightSynchronizer(WeightSynchronizer):
             make_nccl_reshard_refit_info_wire_safe,
         )
 
-        # nccl_reshard_refit_info contains Torch Tensors that are created by
-        # Megatron, which will invoke `import megatron` when unpickled.
-        # To avoid this, we remove the Torch Tensors and restore them later,
-        # via `restore_nccl_reshard_refit_info()`.
+        # nccl_reshard_refit_info holds MeshInfo rank tensors created under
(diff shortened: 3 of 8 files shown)


nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -191,9 +191,13 @@ def is_nccl_reshard_param(param_name: str) -> bool:
     ``load_weights`` path.
 
     Shared-expert FFN weights (``*.shared_expert.*``) are routed to misc path.
+    Co-trained MTP weights are also routed to misc because vLLM keeps the MTP
+    drafter separate from the main model and updates it through ``load_weights``.
```
youngeunkwon0405, line 197: [`nemo_rl/weight_sync/nccl_reshard_utils.py:194-195`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_utils.py#L194-L195)

**[BUG] The MTP drafter never gets post-load weight processing on the nccl-reshard path.** Routing `mtp.*` to the misc path sends these weights through [`_load_weights` → `_maybe_refit_mtp_drafter()`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/models/generation/vllm/vllm_backend.py#L593) → `draft_model.load_weights(...)`, but [`nccl_reshard_refit`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/models/generation/vllm/vllm_backend.py#L1076-L1092) only finalizes the main model and the FP8 KV cache — it never calls [`_maybe_process_mtp_drafter_after_loading()`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/models/generation/vllm/vllm_backend.py#L482), which the IPC/collective paths do call via [`_weight_update_lifecycle.finalize()`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/models/generation/vllm/vllm_backend.py#L624). Its own docstring says it mirrors the main-model post-processing so freshly refit MTP layers get converted to their runtime (grouped-GEMM/quant) layout — and MTP MoE expert weights are exactly the class that needs it. Without it the drafter holds raw un-converted weights: no crash, just silently degraded speculative-decoding acceptance.

The gap is technically pre-existing (non-FFN `mtp.*` params already rode misc), but this PR is what routes the MTP FFN/expert weights there, making it live for the target config. Suggested fix: call `self._maybe_process_mtp_drafter_after_loading()` right after the `process_weights_after_loading(...)` call in `nccl_reshard_refit`.

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -191,9 +191,13 @@ def is_nccl_reshard_param(param_name: str) -> bool:
...
     """
     if "shared_expert" in param_name:
         return False
+    if param_name.startswith("mtp."):
+        return False
```
youngeunkwon0405, line 202: [`nemo_rl/weight_sync/nccl_reshard_utils.py:199-200`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_utils.py#L199-L200)

**[LIMITATION] `startswith("mtp.")` covers the bare-prefix MTP naming but not DeepSeek-V3-style naming.** NemotronH exports MTP weights under HF names like `mtp.layers.*.mixer.up_proj.weight` ([nemotron_h_bridge.py:479-480](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py#L479-L480)), so the gate works there. DeepSeek-V3, however, exports MTP as extra *main-model* layer indices — [`model.layers.{mtp_layer + num_transformer_layers}.mlp.gate_proj.weight`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/deepseek/common.py#L138-L139) for the dense MLP and [`...mlp.experts.*`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/deepseek/common.py#L148-L154) for routed experts — so those still take the bulk path (the shared-experts variant is already excluded by the `shared_expert` substring check). With PP>1 that hits the `KeyError` at [nccl_reshard_utils.py:855](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_utils.py#L855) whose message already names MTP as the expected cause. Since DeepSeek MTP + nccl-reshard was equally broken before this PR, this is an uncovered case rather than a regression — a one-line note in this docstring documenting the limitation (or a setup-time check) would help the next person. The [design doc's bulk-path description](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/docs/design-docs/nccl-reshard-refit.md#L57) could also gain a clause mentioning the MTP exclusion.

**[TEST]** The new branch has no row in the existing parametrized table at [test_nccl_reshard_utils.py:192-211](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/tests/unit/weight_sync/test_nccl_reshard_utils.py#L192), which already covers the `shared_expert` exclusion. Suggested rows using real bridge-exported NemotronH names (drafted; not executed — no GPU/Python env on the review node):

```python
        # NemotronH co-trained MTP -> misc (vLLM drafter updated via load_weights)
        ("mtp.layers.0.mixer.up_proj.weight", False),
        ("mtp.layers.0.mixer.down_proj.weight", False),
        ("mtp.layers.0.mixer.experts.3.up_proj.weight", False),
        # a normal param must be unaffected by the new prefix check
        ("model.layers.0.mlp.up_proj.weight", True),
```

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -247,6 +251,43 @@ def _restore_placement(p):
     return Replicate()
 
 
+def make_nccl_reshard_refit_info_wire_safe(refit_info: dict) -> dict:
```
youngeunkwon0405, line 256: [`nemo_rl/weight_sync/nccl_reshard_utils.py:254`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_utils.py#L254)

**[TEST]** `make_nccl_reshard_refit_info_wire_safe` and its inverse [`restore_refit_info_placements`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_utils.py#L291) have no test coverage (the only production call sites are the synchronizer and [vllm_backend.py:803](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/models/generation/vllm/vllm_backend.py#L803)). Both are pure-CPU functions, and `tests/unit/weight_sync/test_nccl_reshard_utils.py` is already markless CPU-only, so a round-trip test is cheap. Three properties worth pinning: (1) restore reproduces the original placements/meshes, (2) the input dict is not mutated (the shallow-copy discipline is what's most likely to regress), and (3) the wire output contains no `torch.Tensor` anywhere — the actual invariant this fix exists to guarantee.

**[DOC]** Minor: the module docstring's feature list ([L17-27](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_utils.py#L17-L27)) enumerates `build_nccl_reshard_refit_info` and `restore_refit_info_placements` — worth adding a bullet for this new peer.

(For the record, the docstring's premise checks out on the pinned deps: megatron-core pinned via Megatron-Bridge does [`torch.storage._load_from_bytes = safe_load_from_bytes`](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/__init__.py#L58) at import time, which makes Tensor pickles resolve a `megatron.core.safe_globals` reference at unpickle time.)

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -247,6 +251,43 @@ def _restore_placement(p):
...
+    wire_info = dict(refit_info)
+    wire_layers = {}
+    for layer_name, params in refit_info.get("per_layer_params", {}).items():
+        wire_params = []
+        for param_info in params:
+            wire_param = dict(param_info)
+            for key in ("src_mesh_info", "dst_mesh_info"):
+                wire_param[key] = _wire_mesh(wire_param.get(key))
+            for key in ("src_placements", "dst_placements"):
+                wire_param[key] = [_wire_placement(p) for p in wire_param.get(key, [])]
```
youngeunkwon0405: [`nemo_rl/weight_sync/nccl_reshard_utils.py:281-284`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_utils.py#L281-L284)

**[NIT]** `wire_param[key] = _wire_mesh(wire_param.get(key))` materializes an explicit `None` for a missing mesh key, and `.get(key, [])` materializes `[]` for missing placements, so the wire dict's key set can differ from what the builder produced. Harmless today (the builder always sets all four keys, and restore tolerates `None`), but `if key in wire_param:` would keep the copy faithful.

nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
```diff
@@ -205,9 +205,23 @@ def init_communicator(self) -> None:
             train_world_size,
             inference_world_size,
         )
-        self._generation.prepare_nccl_reshard_refit_info(nccl_reshard_refit_info)
+
+        from nemo_rl.weight_sync.nccl_reshard_utils import (
+            make_nccl_reshard_refit_info_wire_safe,
+        )
```
youngeunkwon0405: [`nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py:209-211`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py#L209-L211)

**[GUIDELINE]** Consider hoisting this import to the module top: `nccl_reshard_utils` imports only torch + stdlib, so there's no circular-import risk, and this module is constructed on the driver where torch is already loaded. The repo's linting guideline asks that in-function imports carry a one-line reason (cycle-breaking or heavy optional dep) — neither applies here.

nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
```diff
@@ -205,9 +205,23 @@ def init_communicator(self) -> None:
...
+
+        # nccl_reshard_refit_info contains Torch Tensors that are created by
+        # Megatron, which will invoke `import megatron` when unpickled.
+        # To avoid this, we remove the Torch Tensors and restore them later,
+        # via `restore_nccl_reshard_refit_info()`.
```
youngeunkwon0405: [`nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py:213-216`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py#L213-L216)

**[DOC]** `restore_nccl_reshard_refit_info()` doesn't exist — the function that undoes this on the vLLM side is [`restore_refit_info_placements()`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_utils.py#L291), called from [vllm_backend.py:803](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/models/generation/vllm/vllm_backend.py#L803). Also “we remove the Torch Tensors” is slightly off — the meshes are converted to nested lists (the originals are untouched).

```suggestion
        # nccl_reshard_refit_info holds MeshInfo rank tensors created under
        # Megatron, whose pickles resolve a Megatron-patched storage loader and
        # therefore need `import megatron` on unpickle. Convert them to plain
        # lists here; the vLLM worker rebuilds them in
        # `restore_refit_info_placements()`.
```

nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
```diff
@@ -205,9 +205,23 @@ def init_communicator(self) -> None:
...
+        )
+        self._generation.prepare_nccl_reshard_refit_info(wire_refit_info)
 
     def shutdown(self) -> None:
         # The NCCL process groups' lifecycle is managed by Ray actor teardown;
         # the workers that own the groups are destroyed with the cluster.
-        pass
+        # Drop the local wrapper reference explicitly without changing how the
+        # object graph is serialized into SingleControllerActor.
+        self._generation = None
```
youngeunkwon0405, line 229: [`nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py:222-227`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py#L222-L227)

**[NIT]** This line has no observable effect that the team could find: the synchronizer is cloudpickled into `SingleControllerActor` at construction time ([setup.py:619](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/algorithms/single_controller_utils/setup.py#L619)), and `shutdown()` runs inside the actor long after that, mutating only the actor's copy — so serialization can't be affected either way. Every sibling synchronizer's `shutdown()` is either a real teardown or a documented `pass` ([collective_weight_synchronizer.py:124-128](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/collective_weight_synchronizer.py#L124-L128)). If the intent is breaking the `VllmGeneration ↔ synchronizer` reference cycle for GC, saying so (and the observed symptom) in the comment would help; otherwise consider dropping the line — it also makes any post-shutdown `sync_weights()` fail with a bare `AttributeError` on `None`.

nemo_rl/algorithms/single_controller.py
```diff
@@ -215,7 +215,10 @@ async def run(self) -> dict[str, Any]:
             rollout_task.cancel()
             train_task.cancel()
             await asyncio.gather(rollout_task, train_task, return_exceptions=True)
-            self._logger.finish()
+            try:
+                self._weight_synchronizer.shutdown()
+            finally:
+                self._logger.finish()
```
youngeunkwon0405, line 223: [`nemo_rl/algorithms/single_controller.py:218-221`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/algorithms/single_controller.py#L218-L221)

**[BUG] A raising `shutdown()` here masks the real training exception.** This block is inside `run()`'s `finally`, so an exception from `shutdown()` *replaces* any in-flight exception from the `try` body — the user then sees the teardown error instead of the actual rollout/train failure. That's a real path for other transports: [`CheckpointEngineWeightSynchronizer.shutdown` does `ray.get(...)`](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/checkpoint_engine_weight_synchronizer.py#L250-L257) and [`VllmRemoteSparseWeightSynchronizer.shutdown` does `ray.cancel` + a worker RPC](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/weight_sync/vllm_remote_sparse_weight_synchronizer.py#L394-L404), both of which raise when actors are already dead — exactly the state on the error path. [`VllmGeneration.shutdown()` wraps the identical call in try/except](https://github.com/NVIDIA-NeMo/RL/blob/1fdd2a057acd55f797179868f2275cd3fe051d93/nemo_rl/models/generation/vllm/vllm_generation.py#L937-L948) for this reason.

```suggestion
            try:
                self._weight_synchronizer.shutdown()
            except Exception as e:  # teardown must not mask the original failure
                print(f"Error during weight-synchronizer shutdown: {e}", flush=True)
            finally:
                self._logger.finish()
```

Also worth noting in passing: this is the first call to `weight_synchronizer.shutdown()` from the single-controller loop, so it newly activates checkpoint-engine finalize / sparse ZMQ teardown for all transports and runs them twice (here + driver-side `VllmGeneration.shutdown()`). The team verified both are idempotent, so no action needed — just flagging since the PR description doesn't mention it. If you take this change, a small `run()`-teardown unit test (shutdown called; `logger.finish()` still runs when shutdown raises) would fit `tests/unit/single_controller/test_single_controller.py`.

youngeunkwon0405 (commented):
Team review: 4 agents (RL expert, test reviewer, bug finder, devil's advocate) + lead. All findings below were independently re-verified by the devil's advocate against the pinned submodule SHAs before posting.

**Overall**: both halves of the PR address real, reproducible failures on the pinned dependencies, and the wire-safe conversion was verified to round-trip exactly (`Shard`/`Replicate`/`MeshInfo`) without mutating the trainer-side original. The pickle mechanism described in the docstring is accurate for the pinned megatron-core: [`torch.storage._load_from_bytes = safe_load_from_bytes`](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/__init__.py#L58).

8 inline comments: 1 bug on the vLLM side (MTP drafter post-load processing), 1 exception-masking issue in teardown, 1 documented limitation (DeepSeek-V3 MTP naming), 2 test asks, and 3 small doc/style items. Linter (ruff check + format): clean.

_Generated by Claude Code_

terrykong:
closed since covered by the other prs referencing this one

PR #3535, opened 2026-08-07
feat(megatron): support model config overrides

CarlosGomes98:
# What does this PR do ?
Allow arbitrary options to be passed to MBridge model_cfg

# Issues

# Usage
Some feature flags MBridge exposes are not explicitly forwarded by NeMo-RL. This means we must patch NeMo-RL, or submit an MR with that flag. In some cases, it is useful to just directly pass options to MBridge.

```yaml
...
policy:
  megatron_cfg:
    model_overrides:
      gdn_pre_gated_delta_rule_fusion: true
...
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/models/megatron/setup.py
+++ b/nemo_rl/models/megatron/setup.py
@@ -769,9 +769,8 @@ def _merge_model_overrides(
             )
         return merged_mapping
 
-    # explicitly collect the allowed fields so we can trace 
-    # a potential the error to the config / override path
-    # that triggered it
+    # Explicitly collect the allowed fields so we can trace a potential
+    # error back to the config / override path that triggered it.
     dataclass_init_fields = None
     if is_dataclass(target):
         dataclass_init_fields = {field.name for field in fields(target) if field.init}
--- a/nemo_rl/models/megatron/setup.py
+++ b/nemo_rl/models/megatron/setup.py
@@ -18,6 +18,8 @@ import os
 import threading
 import time
 import warnings
+from collections.abc import MutableMapping
+from dataclasses import is_dataclass
 from typing import Any, Callable, Optional, TypeVar
 
 import torch
@@ -647,6 +649,12 @@ def setup_model_config(
     if "layernorm_epsilon" in config["megatron_cfg"]:
         model_cfg.layernorm_epsilon = config["megatron_cfg"]["layernorm_epsilon"]
 
+    # Apply the generic escape hatch last so it can override provider attributes
+    # that also have first-class NeMo-RL config fields.
+    model_overrides = config["megatron_cfg"].get("model_overrides")
+    if model_overrides is not None:
+        _apply_model_overrides(model_cfg, model_overrides)
+
     # Validate chunking configuration
     _validate_chunking_config(config)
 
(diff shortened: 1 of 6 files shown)


nemo_rl/models/megatron/setup.py
```diff
@@ -647,6 +649,12 @@ def setup_model_config(
     if "layernorm_epsilon" in config["megatron_cfg"]:
         model_cfg.layernorm_epsilon = config["megatron_cfg"]["layernorm_epsilon"]
 
+    # Apply the generic escape hatch last so it can override provider attributes
+    # that also have first-class NeMo-RL config fields.
+    model_overrides = config["megatron_cfg"].get("model_overrides")
+    if model_overrides is not None:
+        _apply_model_overrides(model_cfg, model_overrides)
```
ZhiyuLi-Nvidia: [correctness, non-blocking] `model_overrides` can silently shadow first-class NeMo-RL fields (parallelism, precision, MoE, etc.). When that happens, `model_cfg` has the override but `config["megatron_cfg"][X]` retains the original — every downstream reader of `config` (validators, checkpoint config, other `_apply_*` helpers) sees the stale value. Silent split-brain.

Suggest: reject overlap. If a key in `model_overrides` matches a known first-class `megatron_cfg` field, raise a `ValueError` naming the first-class alternative. Keeps `model_overrides` scoped to the stated use case (MBridge flags NeMo-RL doesn't forward) and rules out the fork at the writer boundary.

Also revisit the docstring at line 652–653 — it currently invites users to override first-class fields, which is exactly the conflict case.

ZhiyuLi-Nvidia (commented):
Direction: error out on conflict between `model_overrides` and first-class megatron_cfg fields.

nemo_rl/models/megatron/setup.py
ZhiyuLi-Nvidia: [correctness, non-blocking] Follow-up to the split-brain comment above — the same `setattr` design also breaks checkpoint round-trip.

The *saved* checkpoint's `run_config.yaml` is authoritative for downstream use (evaluation, fine-tuning from this checkpoint, resume with a modified YAML). If Bridge serializes `model_cfg`'s current attribute state, the mutations happen to persist; if it reconstructs from an earlier kwargs/factory path (as `hf_config_overrides` already warns at line 587–596), overrides silently drop out. Either way the guarantee is accidental, not designed.

On resume: `load_model_config` reads `run_config.yaml` as authoritative, then `_apply_model_overrides` re-applies from the current YAML. If the two disagree (user edited overrides between train and resume, or Bridge saved a pre-mutation snapshot), reloaded weights were trained under one config and are now being run under another. No error, silent drift.

Fix direction: merge overrides into the provider construction path rather than post-hoc `setattr` — e.g. `dataclasses.replace(model_cfg, **overrides)` or route through Bridge's factory kwargs — so the final config is one merged object and `run_config.yaml` records the actual training state.

Also add a regression test: after `setup_model_config`, the returned `model_cfg` (and the value that would be serialized to `run_config.yaml`) equals the base provider config with `model_overrides` merged in. Today the integration test at `test_applies_model_overrides_to_loaded_provider` only asserts the attribute was set on a `SimpleNamespace` with a mocked `__post_init__` — it does not verify the saved/final config would reflect the override end-to-end.

ZhiyuLi-Nvidia (commented):
Follow-up: merge overrides into provider construction, not post-hoc setattr; add a test that the saved/final model config equals the merged one.

nemo_rl/models/megatron/setup.py
CarlosGomes98: That makes sense to me, will edit to reject duplication of fields in model_overrides

CarlosGomes98: Makes sense, edited so the merge takes lace earlier and returns a new dataclass through dataclasses.replace

nemo_rl/models/megatron/setup.py
```diff
@@ -697,6 +713,109 @@ def setup_model_config(
...
+            override_path = f"{path}.{name}"
+            current_value = target.get(name)
+            merged_mapping[name] = _merge_model_override_value(
+                current_value, value, override_path
+            )
+        return merged_mapping
+
+    # explicitly collect the allowed fields so we can trace 
+    # a potential the error to the config / override path
+    # that triggered it
```
ZhiyuLi-Nvidia: The Lint CI is failing on `trim trailing whitespace` here — line 772 has a trailing space after `trace`. Applying this suggestion also tightens the comment slightly (`potential the error` → `potential error`). One-click accept:

```suggestion
    # Explicitly collect the allowed fields so we can trace a potential
    # error back to the config / override path that triggered it.
```

ZhiyuLi-Nvidia (commented):
Lint: strip trailing whitespace at setup.py:772 (one-click suggestion).

ZhiyuLi-Nvidia:
Thank you @CarlosGomes98 a lot! Awesome feature.
LGTM!

ZhiyuLi-Nvidia (approved):

PR #3536, opened 2026-08-07
feat(nemo-gym): add SMG router backend

NolenLiang:
# What does this PR do ?

Adds SMG 1.9.0 as a selectable NeMo Gym inference-router backend.

- generalizes the vLLM Router process from #3518 into a backend-selectable inference router
- maps each backend's launch module, supported policy names, and session-affinity header
- validates unsupported backend/policy combinations before startup
- routes NeMo Gym policy traffic through either vLLM Router or SMG without changing worker endpoints
- pins SMG 1.9.0 in the `nemo_gym` optional dependency set

This is a stacked PR whose base is #3518, so its diff contains only the SMG integration increment.

# Issues

Depends on:

- #3518
- [NVIDIA-NeMo/Gym#2347](https://github.com/NVIDIA-NeMo/Gym/pull/2347), which forwards session-affinity headers and consumes inline token IDs without requiring a router `/tokenize` endpoint

# Usage

Configure the `router` block under `nemo_gym`:

```yaml
router:
  enabled: true
  backend: smg
  policy: consistent_hash
```

Set `backend: vllm_router` to use the vLLM Router implementation through the same configuration shape. The shared `consistent_hash` value maps to SMG's `consistent_hashing` CLI policy when SMG is selected.

**Pre checks**:

- [x] Read and followed the [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [x] Added unit coverage for both router backends, policy validation, lifecycle handling, NeMo Gym wiring, and port allocation
- [x] Ran the relevant unit and functional checks locally
- [x] Kept this first integration PR minimal; repository documentation is intentionally not included

# Additional Information

Validation completed:

- 16 relevant NeMo RL unit tests passed
- 4 relevant NeMo Gym tests passed
- Ruff check and format check passed
- Pyrefly passed with 0 errors across 219 modules
- `uv lock --check` passed with the repository's container-pinned UV 0.11.28
- an SMG 1.9.0 smoke test with two mock vLLM workers verified readiness, consistent session routing, distribution across both workers, inline token-ID pass-through, and graceful shutdown with exit code 0


NolenLiang:
The earlier submodule failure on `d45a886b` is resolved.

- #3518 now points the Gym submodule to `550912f2`.
- This PR was rebased onto the updated #3518 head (`1ee6e4f11`).
- The current head (`318cdf575`) no longer changes the Gym submodule.
- The current submodule fast-forward check passes.

PR #3537, opened 2026-08-07
fix(trtllm): Align HTTP server sampling params, use generate_async

shuyixiong:
# What does this PR do ?

**Make the TRT-LLM HTTP rollout path sample the same way the direct`generate()` path already does, and serve it without a worker thread.**
`TrtllmAsyncGenerationWorker` reaches the engine two ways: `generate()` / `generate_async()` call `_build_sampling_params()` directly, while NeMo-Gym rollouts go through the HTTP server, which built its own `SamplingParams`. The two had drifted:

| | direct path | HTTP path (before) |
|---|---|---|
| `top_k` | applied | **not passed — silently ignored** |
| logprob format | `logprobs_simple_format=True` | dict per generated token |
| engine call | native async | `asyncio.to_thread(llm.generate, ...)` |

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/models/generation/trtllm/trtllm_http_server.py
+++ b/nemo_rl/models/generation/trtllm/trtllm_http_server.py
@@ -17,7 +17,6 @@ Returns NeMoGym fields (prompt_token_ids, generation_token_ids, generation_log_p
 Supports Qwen3 tool calling, DeepSeekR1Parser reasoning, and prefix token splicing.
 """
 
-import asyncio
 import logging
 import threading
 import time
@@ -214,9 +213,8 @@ def create_app(
         )
 
         try:
-            outputs = await asyncio.to_thread(
-                llm.generate,
-                [{"prompt_token_ids": adj_prompt}],
+            output = await llm.generate_async(
+                {"prompt_token_ids": adj_prompt},
                 sampling_params=sampling,
             )
         except RequestError as e:
@@ -228,7 +226,6 @@ def create_app(
                 )
             raise
 
-        output = outputs[0]
         gen = output.outputs[0]
         gen_token_ids = list(gen.token_ids)
 
--- a/nemo_rl/models/generation/trtllm/trtllm_http_server.py
+++ b/nemo_rl/models/generation/trtllm/trtllm_http_server.py
@@ -47,6 +47,45 @@ def _build_reasoning_parser(name: str, chat_template_kwargs: dict[str, Any]) -> 
     return ReasoningParserFactory.create_reasoning_parser(name, chat_template_kwargs)
 
 
+def _build_sampling_params(
+    sampling_params_cls: Any,
+    *,
+    sampling_config: dict[str, Any],
+    stop_token_ids: list[int] | None,
+    max_tokens: int,
+) -> Any:
+    """Build the TRT-LLM sampling params for one HTTP rollout request.
+
+    Mirrors the direct generate() path
+    (``TrtllmAsyncGenerationWorkerImpl._build_sampling_params``) so both paths
+    sample from the same distribution for a given generation config.
+
+    Args:
+        sampling_params_cls: ``tensorrt_llm.SamplingParams``, injected so this
+            helper stays importable and testable without the TRT-LLM runtime.
+        sampling_config: NeMo-RL generation config (temperature / top_p / top_k).
+        stop_token_ids: Extra stop tokens from the generation config, if any.
+        max_tokens: Output cap for this request, already clamped to the context window.
+
+    Returns:
+        A ``SamplingParams`` instance to hand to ``llm.generate_async``.
+    """
+    # TRT-LLM spells "no top-k restriction" as 0, the generation config as null.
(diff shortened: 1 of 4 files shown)


shuyixiong:
/ok to test [612f958](https://github.com/NVIDIA-NeMo/RL/commit/612f958f660178a50eb8c37ee696b300ebad37bf)

shuyixiong:
/ok to test [bf2602e](https://github.com/NVIDIA-NeMo/RL/commit/bf2602e046ba532a4b21005476c8d80df2b44ca9)

nemo_rl/models/generation/trtllm/trtllm_http_server.py
```diff
@@ -199,19 +198,23 @@ async def chat_completions(request: Request):
         from tensorrt_llm import SamplingParams as TrtSamplingParams
         from tensorrt_llm.executor.utils import RequestError
 
+        # Same convention as the direct path (_build_sampling_params): TRT-LLM
+        # spells "no top-k restriction" as 0, the generation config as null.
+        top_k_cfg = sampling_config.get("top_k")
         sampling = TrtSamplingParams(
             temperature=float(sampling_config["temperature"]),
             top_p=float(sampling_config["top_p"]),
+            top_k=int(top_k_cfg) if top_k_cfg is not None else 0,
```
yuki-97: is it possible to use the same format as `top_p` and `temperature`?
`grpo_math_1B.yaml` should be almost all the recipes' default recipe and has `top_k`. is there any recipes will break if doing so?
```suggestion
            top_k=int(top_k_cfg),
```

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
```diff
@@ -295,6 +295,7 @@ def start_http_server(self, port: int = 0) -> str:
             sampling_config={
                 "temperature": self.cfg["temperature"],
                 "top_p": self.cfg["top_p"],
+                "top_k": self.cfg["top_k"],
```
yuki-97, line 298: could you help to add top_k to an existing trtllm sampling_config unit test, or adding a new one if there's no existed sampling_config test?

nemo_rl/models/generation/trtllm/trtllm_http_server.py
```diff
@@ -199,19 +198,23 @@ async def chat_completions(request: Request):
...
             max_tokens=int(max_tokens),
             # Include generated stop tokens so the adapter can trim tokens and logprobs together.
             include_stop_str_in_output=True,
             logprobs=True,
+            logprobs_simple_format=True,
```
yuki-97: Not a blocker / issue for this PR — noting as a follow-up.

HTTP path omits `end_id`/`stop_token_ids` while the direct path passes both, but on the pinned `tensorrt-llm==1.3.0rc21` this is functionally aligned: `LLM._prepare_sampling_params` (`llm.py:1531-1546`) auto-fills `end_id = self.tokenizer.eos_token_id` when `None` via `sampling_params.py:362-366`, and both paths share the same `AsyncLLM`.

So the docstring at `trtllm_worker_async.py:573` ("TRT-LLM requires it to be passed explicitly as end_id") is misleading and `_resolve_end_id()` + explicit `end_id=` look redundant. `stop_token_ids` also formally diverges but is moot under NeMo-Gym's `assert not stop_token_ids` (`rollouts.py:2146`).

yuki-97 (commented):
@shuyixiong thanks for the fix! left some minor comments.

nemo_rl/models/generation/trtllm/trtllm_http_server.py
shuyixiong: In [grpo_math_1B.yaml](https://github.com/NVIDIA-NeMo/RL/blob/0e687e6d07623d780a4174310e92382ce738a8a2/examples/configs/grpo_math_1B.yaml#L349), top_k is default to be null, so the if condition here is necessary here.

nemo_rl/models/generation/trtllm/trtllm_worker_async.py
shuyixiong: Added a test in [3ecc4a1](https://github.com/NVIDIA-NeMo/RL/pull/3537/commits/3ecc4a15da256a188a78fe940c3b2f0124f9f7d2)

nemo_rl/models/generation/trtllm/trtllm_http_server.py
shuyixiong: Thanks for the clarification. Removed _resolve_end_id() in [889aaed](https://github.com/NVIDIA-NeMo/RL/pull/3537/commits/889aaed6715cde651f2a49226005bda9079a6c1f)

shuyixiong:
/ok to test [889aaed](https://github.com/NVIDIA-NeMo/RL/pull/3537/commits/889aaed6715cde651f2a49226005bda9079a6c1f)

yuki-97 (approved):

PR #3538, opened 2026-08-07
ci: AUT-1397 support non-root local container tests

svcnemo-autobot:
## Background

The local container test wrappers forced root, while the CI image installed uv, managed Python, and its symlinked cache under `/root`. Mapping the host UID therefore could not run the baked environment.

## What changed

- install uv globally and keep managed Python/cache under `/opt/uv`
- preserve CI cache export/import at the new path
- run local unit and functional wrappers as the caller UID/GID
- use a writable mounted home/cache and reuse baked dependencies with `--no-sync`
- document the custom-checkout workflow and add contract tests

## Details

```mermaid
flowchart LR
  A[Custom checkout] --> B[Existing CI image]
  B --> C[Caller UID and GID]
  C --> D[Test suite]
```

Dependency fingerprints remain the boundary: source-only commits can reuse the image; dependency changes require rebuilding it.

## Tested

- `uv run --no-sync --group dev pre-commit run --all-files --show-diff-on-failure`
- `uv run --no-sync pytest -q tests/unit/test_meta.py` — 4 passed
- `bash -n tests/run_unit_in_docker.sh tests/run_functional_in_docker.sh`
- synthetic UID 1000 `uv run --no-sync` with `/opt`-managed Python and venv


chtruong814 (dismissed):

docker/Dockerfile
```diff
@@ -50,6 +51,10 @@ FROM scratch AS trtllm-ccache-output-base
 FROM scratch AS custom-setup
 COPY docker/install_apptainer.sh /
 
+# Pin the official uv image so build-time tooling is fetched through the
+# container registry's content-addressed distribution path.
+FROM ghcr.io/astral-sh/uv:${UV_VERSION}@sha256:0f36cb9361a3346885ca3677e3767016687b5a170c1a6b88465ec14aefec90aa AS uv
```
chtruong814: Why is it using this approach here unlike the other repos?

ko3n1g: @svcnemo-autobot please answer charlies  question

svcnemo-autobot: Got it — I'm mid-turn right now, so I've queued your message and will fold it into the next turn as soon as the current one finishes. No need to resend.

svcnemo-autobot: 💬 The difference is intentional because this CI image must also run under an arbitrary caller UID. The prior installer placed `uv` in `/root/.local/bin`, which non-root users cannot access. This uses the digest-pinned official uv image to copy `uv`/`uvx` into `/usr/local/bin`, making them globally executable while avoiding a mutable downloaded install script. The existing cache and managed Python are likewise moved from `/root` to `/opt/uv`. No code change is needed for this comment.

chtruong814: @svcnemo-autobot @ko3n1g this doesn't answer my question.  You are making changes across other projects to allow non-root running but are not copying over a binary from a base uv image. Why make this change here only it seems whereas continue to install via curl.  Unless there is a good reason, keep the install via curl as before and just install in a location that non-root can be used.

svcnemo-autobot: 💬 You’re right. I had misread the original question. There was no repository-specific reason to switch RL to copying binaries from the uv image. I have now restored the existing versioned curl installer and only changed its destination to the non-root-accessible `/usr/local/bin`; `/opt/uv` remains the shared cache and managed-Python location. The follow-up fix also preserves the pre-`FROM` `UV_VERSION` contract required by RL’s meta test.

PR #3539, opened 2026-08-07
fix: preserve fp32 router correction bias during refit

fujial-code:
# What does this PR do?

Preserve the Nemotron-Nano V3 MoE `e_score_correction_bias` buffer in FP32 during DTensor policy refit instead of unconditionally casting it to the policy dtype (typically BF16).

The change keeps the refit payload and its metadata consistent:

- `dtensor_params_generator` sends the router correction bias in its original FP32 dtype.
- `prepare_refit_info` advertises FP32 for the same tensor.
- Other refit tensors continue to use the configured policy dtype.

# Why?

`e_score_correction_bias` is maintained in FP32, but the current refit path casts every tensor to `self.dtype`. For BF16 policies, that downcast can introduce coarse rounding in the router correction values and change expert routing between the trainer and generation model.

This is a focused mitigation for the router-state precision issue. It does not include the native LoRA prototype and does not claim to address the separate finite-precision loss from materializing merged LoRA weights in BF16.

# Issues

No linked issue.

# Usage

No configuration change is required. FP32 preservation is selected only for tensors whose final fully qualified name is `e_score_correction_bias` and whose source dtype is FP32.

# Testing

- Added unit coverage for the transmitted tensor dtype.
- Added unit coverage for matching refit metadata.
- Verified that ordinary FP32 buffers are still converted to BF16.
- `git diff --check` passes.
- Python syntax compilation passes for the implementation and test files.
- Targeted pytest was not run on the HSG login node because that environment does not provide `torch` or `pytest`; full Nano V3 regression remains pending.

- [x] Followed the contribution conventions and signed off the commit.
- [x] Added focused unit tests.
- [ ] Run the targeted unit tests in the NeMo-RL test environment.
- [ ] Run the Nano V3 trainer-generation logprob/KL regression.

# Additional information

This draft intentionally contains only the two-file router FP32 fix. Native LoRA support, LoRA merge-precision investigation code, experiment recipes, and stream-fence changes are out of scope.

### DIFF
--- a/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
+++ b/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
@@ -87,6 +87,18 @@ from nemo_rl.utils.packed_tensor import packed_broadcast_producer
 from nemo_rl.utils.timer import Timer
 
 
+def _refit_tensor_dtype(
+    fqn: str, tensor: torch.Tensor, default_dtype: torch.dtype
+) -> torch.dtype:
+    """Preserve the FP32 dtype used by inference-critical MoE router state."""
+    is_router_correction_bias = fqn.rsplit(".", maxsplit=1)[-1] == (
+        "e_score_correction_bias"
+    )
+    if is_router_correction_bias and tensor.dtype == torch.float32:
+        return tensor.dtype
+    return default_dtype
+
+
 def dtensor_params_generator(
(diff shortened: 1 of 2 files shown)


fujial-code:
Superseded by #3540 after moving the unchanged commit to jia/preserve-fp32-router-bias. Commit SHA remains 7a849d756e1facc705072af7f52e16a4391defb4.

PR #3540, opened 2026-08-07
fix: preserve fp32 router correction bias during refit

fujial-code:
# What does this PR do?

Preserve the Nemotron-Nano V3 MoE `e_score_correction_bias` buffer in FP32 during DTensor policy refit instead of unconditionally casting it to the policy dtype (typically BF16).

The change keeps the refit payload and its metadata consistent:

- `dtensor_params_generator` sends the router correction bias in its original FP32 dtype.
- `prepare_refit_info` advertises FP32 for the same tensor.
- Other refit tensors continue to use the configured policy dtype.

# Why?

`e_score_correction_bias` is maintained in FP32, but the current refit path casts every tensor to `self.dtype`. For BF16 policies, that downcast can introduce coarse rounding in the router correction values and change expert routing between the trainer and generation model.

This prevents an avoidable FP32-to-BF16 precision loss for the router correction bias in the full-weight refit path.

# Experimental context

The plot below shows the 50-step merged-LoRA run with the FP32 router-bias preservation enabled together with a stream fence.

train/gen_kl_error for the 50-step merged-LoRA run

`train/gen_kl_error` starts near the `0.002` scale, grows after approximately step 18, and reaches a transient peak of approximately `0.102` around step 26 before returning near its initial scale. The fix prevents the FP32 router correction state from being quantized during each refit, removing that source of precision loss from the merged full-weight sync path.

# Issues

No linked issue.

# Usage

No configuration change is required. FP32 preservation is selected only for tensors whose final fully qualified name is `e_score_correction_bias` and whose source dtype is FP32.

# Testing

- Added unit coverage for the transmitted tensor dtype.
- Added unit coverage for matching refit metadata.
- Verified that ordinary FP32 buffers are still converted to BF16.
- `git diff --check` passes.
- Python syntax compilation passes for the implementation and test files.
- Targeted pytest was not run on the HSG login node because that environment does not provide `torch` or `pytest`; full Nano V3 regression remains pending.

- [x] Followed the contribution conventions and signed off the commit.
- [x] Added focused unit tests.
- [ ] Run the targeted unit tests in the NeMo-RL test environment.
- [ ] Run the Nano V3 trainer-generation logprob/KL regression.


fujial-code (commented):

Reviewed the rebased two-file change at `f5984ab954d97fc6ea460221e55bf68442ab0e31` with parallel RL/design, bug/test, comment/process, and independent devil's-advocate passes. No remaining correctness, mixed-dtype transport, or design findings were identified.

### What was verified

- The pinned Automodel implementation creates `e_score_correction_bias` in FP32 and documents that BF16 quantization can change expert routing.
- The refit payload and `state_dict_info` metadata use the same dtype-selection helper after HF state-dict adaptation.
- Static inspection found per-tensor dtype support in the IPC, collective, HTTP, and checkpoint-engine refit paths.

### Findings resolved before staging this review

- Updated the generator docstring to describe `target_dtype` as the default and document the FP32 router-bias exception.
- Applied the Ruff formatting change reported by the previous CI run.
- Rebasing onto current `main` leaves this PR one commit ahead with the same two-file scope.
- Added the required `CI:L2` label and reauthorized CI for the new SHA.

### Verification status

- `git diff --check`: passed.
- DCO, title, copyright, and secrets checks: passed or running cleanly on the new SHA.
- The full `CI:L2` test matrix is currently queued/running. Local runtime tests were not claimed because the login node has neither `uv` nor a working GPU environment.
- The existing 50-step plot combines router-FP32 preservation with a stream fence, so it should be treated as preliminary rather than an isolated causal A/B result.

No inline findings remain on the current head. The PR should stay in Draft until the targeted/runtime CI results are available.

_Generated by Claude Code_

tests/unit/models/policy/test_dtensor_worker_v2.py
```diff
@@ -686,6 +686,24 @@ def test_dtype_conversion(self):
                 f"Tensor {name} should be converted to {target_dtype}"
             )
 
+    def test_preserves_fp32_router_correction_bias(self):
+        """FP32 MoE router state must not be downcast during refit."""
+
+        class RouterModel(nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.register_buffer(
```
yuki-97: Both `RouterModel`s register the buffer at the top level, so the state-dict key is `e_score_correction_bias` with no dots — `rsplit(".", maxsplit=1)[-1]` returns the whole string and the suffix match is never exercised. Real FQNs are always nested (`model.layers.N.mlp.gate.e_score_correction_bias`).

Narrowing the helper to `fqn == "e_score_correction_bias"` would pass both tests and preserve nothing at runtime.

**Suggest fix:** nest the buffer under a child module here and at the sibling test below. `_maybe_merge_lora_weight` early-returns on non-`.weight` names, so nesting is safe.

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -87,18 +87,31 @@
 from nemo_rl.utils.timer import Timer
 
 
+def _refit_tensor_dtype(
```
yuki-97: The v1 sibling has the identical unconditional cast and isn't covered here: `dtensor_policy_worker.py:112` (`full_tensor.to(target_dtype)`) and `:1849` (`(tensor.shape, self.dtype)`).

v1 is still selectable via `dtensor_cfg._v2: false` (`lm_policy.py:163-188` gates out only LoRA and HSDP), and transformers keeps this buffer strictly FP32 on that path too — `_keep_in_fp32_modules_strict = ["e_score_correction_bias"]` in `deepseek_v3/modeling_deepseek_v3.py:551` — so a MoE model on v1 hits exactly this bug.

**Suggest fix:** lift `_refit_tensor_dtype` into a shared helper both workers call, or note in the description why v1 is out of scope.

yuki-97 (approved):
LGTM! thanks for the fix! @fujial-code

PR #3545, opened 2026-08-07
fix(vllm): support native BF16 FlashInfer TRTLLM refit

seonjinn:
## TL;DR for reviewers

FlashInfer TRTLLM repacks BF16 MoE expert weights into a backend-private layout at load time, so NeMo-RL's generic refit (write canonical tensors into live params, run generic post-processing) silently corrupts the runtime weights on this backend. This PR makes the affected refit paths follow vLLM's native layerwise reload lifecycle instead. **The three things to review:** (1) the lifecycle context manager and its fail-closed semantics in `vllm_backend.py`; (2) the transport-buffer detach that keeps deferred reload tensors alive after NeMo-RL reuses NCCL/IPC buffers; (3) the single-buffer collective path that removes a cross-stream read-after-write hazard.

## Problem

FlashInfer TRTLLM does not execute incoming BF16 expert weights in their original layout — vLLM repacks them into backend-specific runtime storage at model load. Two failure modes follow from the generic refit path:

1. **Layout bypass:** loading new weights without the native reload protocol never rebuilds the repacked runtime storage, so the model keeps executing stale expert weights even though the transfer succeeded.
2. **Buffer aliasing:** the native protocol defers some `weight_loader` calls until finalization, and those deferred calls hold references into NCCL/IPC transport buffers that NeMo-RL reuses for the next transfer — so by the time they replay, the data may have been overwritten.

## How the fix works

For this backend only, each refit follows vLLM's native lifecycle:

1. Initialize vLLM's layerwise reload state.
2. Load the new BF16 weights through the existing collective or IPC/ZMQ transport.
3. **Detach** any deferred reload tensors from reusable transport buffers (clone-on-alias, keyed by storage pointer).
4. **Finalize** so vLLM rebuilds the FlashInfer TRTLLM runtime layout (plus the HpcModule refresh that vLLM's own finalizer omits — upstream gap, to be filed).
5. Only then reuse the transport buffers.

Backend detection inspects the *realized* model, so `auto` recipes that resolve to FlashInfer TRTLLM at runtime are covered.

**Stream safety:** with the native lifecycle active, vLLM replays buffered `weight_loader`s eagerly when a layer's last shard arrives — potentially on a different CUDA stream than the one that produced the clones. The collective consumer therefore runs single-buffer on this path (trading broadcast/load overlap for a closed race window, this backend only).

**Fail-closed:** any failure after reload initialization marks the worker permanently unusable (`_refit_unusable_reason`); subsequent refit attempts return False instead of running against a partially updated model.

## Scope

| Path | Behavior |
|---|---|
| collective, IPC/ZMQ (unquantized FlashInfer TRTLLM) | native lifecycle (this PR) |
| `nccl_reshard` | keeps its direct-buffer protocol (setup-time mapping needs stable live storage) |
| checkpoint-engine, sparse-delta (unquantized FlashInfer TRTLLM) | **rejected loudly** — they bypass the native lifecycle and would corrupt silently |
| FP8 / ModelOpt-quantized / Triton / other backends | unchanged |
| static MTP draft weights | same lifecycle during their one-time disk load |
| co-trained MTP refits | unsupported, fail early with an explicit error |

Runtime support only; recipe changes follow separately. Independent of #3478 (MXFP8 refit performance). Sibling ordering: this PR → #3659 (extends the lifecycle to `nccl_reshard`) → #3669 (perf); #3651 (opt-in `reload_weights` API) is complementary and rebases on top.

## Validation

- Unit tests: repeated refits, runtime `auto` backend selection, transport-buffer ownership, failure handling (including public-method fatal-raise and worker poisoning), unaffected backend paths, static MTP draft loading.
- 3-step Qwen3-30B-A3B smoke on 4 nodes x 4 GB200: all 48 worker refits completed without fatal error (job `5950903`). Note: the smoke predates the single-buffer stream fix and validates the lifecycle, not stream safety — the race window was narrow enough that a clean smoke was consistent with it.
- CI: `/ok to test` issued on the current head; earlier `CI:Lfast` runs targeted a pre-review SHA.


yuki-97 (commented):
hi @youngeunkwon0405 , could you help review this PR? thanks!

seonjinn:
## `/review-pr-team` self-review

Ran the upstream team review with RL, test, bug-finding, design, comment-review, and devil-advocate agents. There were no existing inline review threads requiring action.

### Outcome

- No PR-introduced blocking issue remains after the follow-ups in `95452e910` and `8881a6df2`.
- Documented that a failed native layerwise reload permanently makes the worker unsafe to reuse.
- Added direct coverage for the fatal-error policy and for a detach failure after a successful load.
- Added a real vLLM layerwise-reload regression test that overwrites a reused transport buffer between deferred weight batches and verifies the final materialized parameters retain both original values.
- Clarified the transport boundary in code and in the PR description: collective and IPC/ZMQ refits use the native lifecycle; `nccl_reshard` intentionally retains its direct-buffer protocol because its setup-time mapping depends on stable live parameter storage. Rebuilding parameter storage would invalidate that mapping.
- Confirmed checkpoint-engine transports are unchanged and outside this PR.
- Rejected an FP8 KV-cache concern after checking the pinned vLLM layerwise finalization source; the native finalizer restores attention scales.

### Verification

- Merged current `main` and updated recursive submodules without conflicts.
- `git diff --check` and Python AST parsing passed locally. The repository lockfile does not support local macOS arm64 test execution.
- The post-fix independent review found no blocking correctness issue.
- Latest-head `CI:Lfast` is running.
- The existing 4-node GB200 smoke completed 48 repeated worker refits without a fatal error (job `5950903`).

seonjinn:
Contributor self-review status: this implementation is superseded by #3659, which provides the current BF16 TRTLLM NCCL-reshard lifecycle and compatibility path on latest main. Merging main here would preserve a duplicate implementation and increase maintenance risk. Recommendation: close this PR in favor of #3659; do not retrigger CI.

seonjinn:
Self-review completed with `/review-pr-team` on head `10e40a591`.

- Reviewed the native layerwise reload lifecycle, repeated refits, MTP handling, collective and IPC transports, failure handling, ModelOpt gating, and compatibility with the changes now on `main`.
- Found and fixed one correctness issue: the vLLM layerwise finalizer does not run the separate `HpcModule` post-load pass. The updated code now refreshes derived HPC state for both the policy model and a disk-loaded MTP drafter without re-running quantization or attention post-processing.
- Added tests for the HPC-only pass and both native layerwise reload call sites.
- No other blocking correctness, compatibility, security, or deadlock issues were found.

Local validation completed:
- `git diff --check`
- Python syntax compilation for the changed implementation and test files

The full vLLM unit suite is left to CI because the local host cannot create the repository Linux-only locked vLLM environment. Follow-up PRs #3659 and #3669 should be restacked after this PR lands.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -196,7 +249,31 @@ def _get_named_parameters(self) -> dict[str, torch.nn.Parameter]:
...
+        }
+        load_error: Exception | None = None
+        try:
+            self.model_runner.model.load_weights(weights=policy_weights)
+        except Exception as error:
+            load_error = error
+            raise
+        finally:
+            try:
+                _detach_pending_layerwise_weights(
```
youngeunkwon0405, line 289: [`nemo_rl/models/generation/vllm/vllm_backend.py:268`](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/vllm_backend.py#L268)

**[BUG] Cross-stream read-after-write race on the collective refit path.**

`packed_broadcast_consumer` round-robins chunks across `NRL_REFIT_NUM_BUFFERS=2` CUDA streams and [only syncs a stream before reusing that same stream](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/utils/packed_tensor.py#L202); `post_unpack_func` (= `_load_weights`) runs inside `torch.cuda.stream(streams[buffer_idx])`. With the layerwise reload active:

1. Chunk k's detach pass enqueues `loaded_weight.clone()` on stream A.
2. vLLM replays a layer **eagerly on the chunk that delivers its last shard** — [`if info.load_numel >= info.load_numel_total: _layerwise_process(layer, info)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/reload/layerwise.py#L217) — which [re-runs every buffered `weight_loader(*args)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/reload/layerwise.py#L354), i.e. on chunk k+1's stream B, reading stream A's clones with no event/sync edge (the global [`for s in streams: s.synchronize()`](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/utils/packed_tensor.py#L255) only runs after the loop). The clones also aren't `record_stream`-protected, so the allocator can reuse their blocks prematurely.

FusedMoE layers on the target models routinely exceed the chunk size, so layers straddling chunks (and thus streams) is the common case. Failure mode is intermittent silent garbage in expert weights — narrow window, so the 48-refit smoke passing is consistent with this. Pre-PR there was no cross-chunk data dependency, which is why this surfaces now.

Suggested fix: record a `torch.cuda.Event` after `post_unpack_func` on each stream and `wait_event` it on the next stream before its `post_unpack_func` (keeps broadcast/load overlap), or force a single buffer/stream when the native reload is active. The IPC/ZMQ path looks safe — it stays on one stream and fences before each ACK.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -606,12 +703,80 @@ def _get_sparse_delta_applier(self) -> Any:
...
+        model_runner = getattr(self, "model_runner", None)
+        vllm_config = getattr(model_runner, "vllm_config", None)
+        if vllm_config is None:
+            return False
+        if getattr(vllm_config, "quant_config", None) is not None:
+            return False
+
+        return _model_uses_unquantized_flashinfer_trtllm(self.model_runner.model)
+
+    def _validate_weight_update_compatibility(self) -> None:
```
youngeunkwon0405: [`nemo_rl/models/generation/vllm/vllm_backend.py:723`](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/vllm_backend.py#L723)

**[Fail-loud] Checkpoint-engine and sparse-delta refits still bypass the native lifecycle silently.**

[`_update_weights_from_checkpoint_engine_async` calls `self._load_weights(...)`](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/checkpoint_engine.py#L152) with `_nrl_layerwise_reload_active=False`, and [`vllm_sparse_delta.py` calls `model.load_weights(...)` directly](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/vllm_sparse_delta.py#L372) — so an unquantized FlashInfer TRTLLM MoE model refit via those transports hits exactly the pre-PR silent-corruption path (live `w13`/`w2` are in the [4-D block layout](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/fused_moe/flashinfer_utils.py#L93) after startup, so canonical-layout writes corrupt silently). Neither combo is rejected anywhere today.

Since this validator is already called from `prepare_refit_info`, consider extending it (or the transport entry points) to raise for these transport+backend combinations until they're supported — mirroring the co-trained-MTP rejection just below.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -72,6 +72,55 @@ class IPCWeightManifestError(RuntimeError):
...
+        info = get_layerwise_info(module)
+        for _, arguments in info.loaded_weights:
+            loaded_weight = arguments.arguments.get("loaded_weight")
+            if not isinstance(loaded_weight, torch.Tensor):
+                continue
+            if loaded_weight.untyped_storage().data_ptr() in source_storage_ptrs:
+                arguments.arguments["loaded_weight"] = loaded_weight.clone()
+
+
+def _process_hpc_modules_after_loading(model: torch.nn.Module) -> None:
```
youngeunkwon0405: [`nemo_rl/models/generation/vllm/vllm_backend.py:95`](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/vllm_backend.py#L95)

**[UPSTREAM] Workaround is correct — consider filing the gap against vLLM.**

This helper correctly mirrors the HpcModule pass in [`model_loader/utils.process_weights_after_loading`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/utils.py#L129-L137), which [`finalize_layerwise_processing`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/reload/layerwise.py#L226) omits (it only re-runs quant methods and attention finalization). vLLM's own reload engines ([ipc_engine](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/reload/ipc_engine.py#L210), [nccl_engine](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/reload/nccl_engine.py#L138)) share the same gap. Worth filing at https://github.com/vllm-project/vllm/issues so the workaround can eventually be dropped: "finalize_layerwise_processing should run the HpcModule pass that model_loader/utils.process_weights_after_loading runs".

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -631,7 +796,7 @@ def finalize() -> None:
 
     def _weight_update_errors_are_fatal(self) -> bool:
```
youngeunkwon0405, line 844: [`nemo_rl/models/generation/vllm/vllm_backend.py:797`](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/vllm_backend.py#L797)

**[TEST] The fatal-raise behavior is untested through the public update methods.**

When this predicate returns True, `update_weights_via_ipc_zmq` ([L905](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/vllm_backend.py#L905)) and `update_weights_from_collective` ([L934](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/vllm_backend.py#L934)) re-raise instead of returning False. The new tests pin [the predicate](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/tests/unit/models/generation/test_vllm_backend.py#L514) and [the lifecycle in isolation](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/tests/unit/models/generation/test_vllm_backend.py#L553), but no test drives either public method with a fatal-native config and a failing load — so reverting the `raise` to `return False` would keep the whole unit suite green while failed refits get silently swallowed again (the exact regression this guards against). A test that calls the public method and asserts `pytest.raises` would pin it.

tests/unit/models/generation/test_vllm_backend.py
```diff
@@ -119,6 +121,470 @@ def _make_mtp_refit_extension(
...
+    class FakeHpcModule:
+        def __init__(self):
+            self.process_weights_after_loading = MagicMock()
+
+    hpc_module = FakeHpcModule()
+    other_module = object()
+    model = SimpleNamespace(
+        named_modules=lambda: [("", other_module), ("rope_norm", hpc_module)]
+    )
+    monkeypatch.setattr(
```
youngeunkwon0405: [`tests/unit/models/generation/test_vllm_backend.py:151`](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/tests/unit/models/generation/test_vllm_backend.py#L151)

**[LINT] `ruff format --check` fails here with the repo-pinned ruff 0.9.9** — this call fits on one line:

```suggestion
    monkeypatch.setattr("vllm.model_executor.layers.hpc.HpcModule", FakeHpcModule)
```

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -321,6 +398,7 @@ def prepare_refit_info(self, state_dict_info: dict[str, Any]) -> None:
             state_dict_info (dict): A dictionary containing the info for refit.
                 e.g. {tensor_name: (shape, dtype)}
         """
+        self._validate_weight_update_compatibility()
```
youngeunkwon0405: [`nemo_rl/models/generation/vllm/vllm_backend.py:401`](https://github.com/NVIDIA-NeMo/RL/blob/10e40a5914a30f16b5bcfe9b8b226143a20f77c0/nemo_rl/models/generation/vllm/vllm_backend.py#L401)

**[DOCSTRING] `prepare_refit_info` can now raise, but the docstring has no Raises section.** Consider adding:

```
Raises:
    RuntimeError: If the model realizes the unquantized FlashInfer TRTLLM MoE
        backend while a co-trained MTP drafter is enabled (unsupported by the
        native layerwise refit lifecycle).
```

youngeunkwon0405 (commented):
Team review (5 agents: RL-code, vLLM-API, tests, bug-finder, devil's advocate; every vLLM claim verified against the pinned v0.25.1 wheel source).

This is a well-built change — the lifecycle design, the realized-backend gating (including `auto`), the failure latching, and especially the buffer-overwrite regression test all held up under adversarial verification. The detach-necessity claim, the FlashInfer TRTLLM finalize semantics, and the HPC-pass gap were each confirmed in vLLM source. The earlier FP8-KV-cache concern was re-checked and is indeed covered: the layerwise finalizer [re-runs `layer.process_weights_after_loading(model_config.dtype)` and reloads streamed k/v scales](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/reload/layerwise.py#L289-L303), equivalent to the skipped generic pass.

**6 inline comments**: 1 confirmed concurrency bug on the collective path (cross-stream RAW race introduced by the deferral+detach), 1 fail-loud gap (checkpoint-engine/sparse-delta transports), 1 upstream-issue suggestion, 1 test gap, 1 lint failure (repo-pinned ruff 0.9.9), 1 docstring nit.

Lower-confidence notes, not requesting changes:
- The detach helper relies on private `reload.layerwise` internals (`LayerReloadingInfo.loaded_weights` shape, the `loaded_weight` arg name); imports are function-local so older vLLM fails loudly only when the feature runs. A version-pinning comment would help future upgrades.
- `except Exception` in `_weight_update_lifecycle` misses `BaseException` (e.g. a cancelled refit): the worker isn't marked unusable and the next refit would silently merge stale buffered clones (vLLM [skips still-initialized layers](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/reload/layerwise.py#L107)). Low reachability; `except BaseException` closes it if you think it's worth it.
- Question: does BF16-TRTLLM + `nccl_reshard` fail loudly today, or run the generic finalize silently until #3659 lands? Our read is the reshard mapping most plausibly errors at prepare time on the 4-D block layout, but a confirmation would help.
- Given the per-refit init/detach/finalize overhead, one before/after refit-latency number in the description would be a nice-to-have (perf work itself is understood to be #3669).

_Generated by Claude Code_

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -606,12 +703,80 @@ def _get_sparse_delta_applier(self) -> Any:
...
+            )
+
+            model = self.model_runner.model
+
+            def finalize() -> None:
+                with torch.device(self.device):
+                    finalize_layerwise_reload(model, self.model_config)
+                    _process_hpc_modules_after_loading(model)
+                    self._maybe_process_mtp_drafter_after_loading()
+                torch.accelerator.synchronize()
```
youngeunkwon0405: Why are we using `torch.accelerator.synchronize()`, I think we can just use `torch.cuda.synchronize()` do avoid giving a wrong signal to others.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -606,12 +703,80 @@ def _get_sparse_delta_applier(self) -> Any:
...
+                finalize_layerwise_reload,
+                initialize_layerwise_reload,
+            )
+
+            model = self.model_runner.model
+
+            def finalize() -> None:
+                with torch.device(self.device):
+                    finalize_layerwise_reload(model, self.model_config)
+                    _process_hpc_modules_after_loading(model)
```
youngeunkwon0405: Can I ask what is the `hpc_modules`? and why this is only needed in the flashinfer_trtllm refit path?

nemo_rl/models/generation/vllm/vllm_backend.py
youngeunkwon0405: I think resolving this issue is too complex, you can just set NRL_REFIT_NUM_BUFFERS=1 when the refit goes this path, or add assertion to gate NRL_REFIT_NUM_BUFFERS>1 case

youngeunkwon0405: Also, above this line, it is nicer to add a comment of why we are doing this detach. Even for me it took time to reason about this.

youngeunkwon0405 (commented):
I did a thorough review of this PR. The logic is hard to understand, but I think the problem itself is just complex. I think the implementation functionally makes sense and is reasonable. Please address the comments. Thank you!

youngeunkwon0405:
@guyueh1 I studied this PR today and learned that this is also using the layer-wise reload feature of vLLM, which is I didn't realize before. I am adding you because you have been working on the reload PR. Maybe you might also want to check an overlap with your PR? I will also review your PR soon.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Addressed in e81aba7b0 the way you suggested: `packed_broadcast_consumer` now takes an optional `num_buffers` override and the collective path passes `num_buffers=1` whenever the native layerwise reload is active, removing the cross-stream RAW hazard while leaving the default two-buffer overlap untouched for every other backend (chunk boundaries only depend on the packed-buffer target size, so the producer keeping 2 buffers is fine). Also added the comment above the detach explaining why the clone is needed.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Fixed in e81aba7b0: added `_reject_unsupported_native_trtllm_transport()` and wired it into `prepare_sparse_delta_refit_info`, `update_weights_from_decoded_sparse_payload`, and `_update_weights_from_checkpoint_engine_async` — both transports now raise for an unquantized FlashInfer TRTLLM model instead of silently writing canonical layouts into repacked runtime storage. Covered by new unit tests.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Agreed — will file the gap upstream ("finalize_layerwise_processing should run the HpcModule pass that model_loader/utils.process_weights_after_loading runs") and link it here so the workaround can be dropped once fixed. Also added a code comment in e81aba7b0 explaining the gap so the workaround's purpose is discoverable without the PR context.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Added in e81aba7b0: `test_update_weights_from_collective_reraises_on_fatal_native_refit` drives the public `update_weights_from_collective` with a fatal-native config and a failing transport, asserts the exception propagates (not `return False`), and additionally pins that the worker is marked unusable for the next attempt.

tests/unit/models/generation/test_vllm_backend.py
seonjinn: Applied your suggestion verbatim in e81aba7b0.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Added the Raises section in e81aba7b0.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: You're right — switched to `torch.cuda.synchronize()` in e81aba7b0.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: `HpcModule` is vLLM's fused high-performance module wrapper; after a weight load it recomputes derived state from the raw weights via `process_weights_after_loading`. The normal (non-layerwise) refit path runs vLLM's `model_loader/utils.process_weights_after_loading`, which includes an HpcModule pass — but `finalize_layerwise_reload` (the native lifecycle this PR adopts for FlashInfer TRTLLM) only re-runs quant methods and attention finalization and skips that pass, so HPC modules would keep stale derived state. That's why the refresh is needed only on this path. Added a code comment explaining this in e81aba7b0, and we'll file the gap upstream (see the other thread).

seonjinn:
> I did a thorough review of this PR. The logic is hard to understand, but I think the problem itself is just complex. I think the implementation functionally makes sense and is reasonable. Please address the comments. Thank you!

Thank you so much for your thorugh review Youngeun, I'll try to fix codes for clear understanding.

seonjinn:
Superseding my 08-18 comment ("close in favor of #3659"): that had the dependency direction backwards — #3659 explicitly extends this PR's native layerwise refit to the `nccl_reshard` transport, and #3669 is a perf follow-up on top of that. The intended order is **#3545 → #3659 → #3669**.

On the overlap question with @guyueh1's #3651 (native `reload_weights` API): the two are complementary layers, not duplicates. This PR fixes the default NeMo-RL loader paths (IPC + collective) by wrapping them in vLLM's layerwise reload lifecycle, streaming chunk-by-chunk; #3651 adds an opt-in `refit_with_reload_api` flag that bypasses the NeMo-RL loader and hands a whole weight iterator to `model_runner.reload_weights` (collective-only, non-colocated, also covers MXFP8), while keeping this PR's path as the default else-branch. They touch the same regions of `update_weights_from_collective` and `packed_tensor.py`, so whichever lands second needs a rebase — proposal: land this first (default-path correctness + fail-closed lifecycle + IPC coverage), rebase #3651 on top for the opt-in native API.

seonjinn (commented):
Self-review summary (supplemental to youngeunkwon0405's review, whose 8 comments were addressed in e81aba7b0 with per-thread replies).

Method: five review agents over the post-fix diff, then an adversarial verification pass that re-checked every citation, verified vLLM internals against a real vLLM tree, and re-ran CI queries live. No inline code findings survived adversarial verification -- the lifecycle state machine (fail-closed on every failure window including finalize), the transport-buffer detach, the single-buffer race fix, and the mock fidelity of the new tests were all confirmed sound. Two follow-up fixes were pushed as 86c94f16: a version-guard on the fused_moe.oracle import (on vLLM builds without that module layout, every unquantized-model refit would have raised ImportError instead of skipping the TRTLLM check) and a missing num_buffers kwarg on a pre-existing test fake that would have failed the vllm-gated test lane.

**Merge gates (process, not code):**

1. **CI has not run on any recent head.** The last /ok to test targeted f0c23ae96 (08-18), fourteen commits and a 169-file upstream merge ago. A fresh /ok to test on the final head is required; posted above.
2. **PR body needs a refresh.** Three statements are stale relative to the code: "Checkpoint-engine transports are unchanged" (they are now rejected for the unquantized TRTLLM backend, by design); the CI claim references a stale SHA; and the 48-refit smoke predates the cross-stream race fix -- a clean smoke was consistent with the race being present (narrow-window silent corruption), so it should be cited as evidence for the lifecycle, not for stream-safety. Worth adding while editing: the native path now runs the collective consumer single-buffer, trading broadcast/load overlap for stream-safety on this backend only.
3. **Sibling-PR order** is clarified in the comment above: #3545 -> #3659 (nccl_reshard extension) -> #3669 (perf), with #3651 (opt-in reload_weights API) complementary and rebased on top.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -921,12 +964,15 @@ def update_weights_from_collective(self) -> bool:
...
 
         try:
+            native_layerwise_refit = self._uses_native_layerwise_refit("collective")
             with self._weight_update_lifecycle("collective") as finalize:
                 packed_broadcast_consumer(
                     iterator=iter(self.state_dict_info.items()),
                     group=self.model_update_group,
                     src=0,
                     post_unpack_func=self._load_weights,
+                    # One stream orders deferred clone/replay dependencies.
```
youngeunkwon0405: `# Double buffering (num_buffer>1) causes a race condition when using native_layerwise_refit.`

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -265,6 +281,7 @@ def _load_full_hf_weights(
             raise
         finally:
             try:
+                # Preserve deferred replay inputs before this buffer is reused.
```
youngeunkwon0405:
```
# Native layerwise reload may defer weight_loader calls until all shards for
# a layer have arrived. NCCL/IPC weights are views into reusable receive buffers
# that subsequent transfers may overwrite, so clone any pending tensors that
# alias the current buffer and preserve them until deferred replay consumes them.
```

youngeunkwon0405 (changes requested):

seonjinn:
/ok to test [e450256](https://github.com/NVIDIA-NeMo/RL/pull/3545/commits/e450256eb21c317c6872ff95064a99607514bb69)

youngeunkwon0405 (approved):

terrykong (approved):

PR #3547, opened 2026-08-07
fix(docker): add --package nemo-rl to uv sync calls to fix workspace extra resolution

kajalj22:
## What does this PR do?

Fixes `uv sync --frozen --extra X` failing with:
```
error: Extra 'vllm' is not defined in the project's optional-dependencies table
```

## Root cause

uv 0.11.x generates **revision 4** lockfiles when packages actually change (e.g. during an automated MBridge bump). In revision 4, `uv sync --frozen --extra X` without an explicit `--package` flag resolves the sync target against the **first workspace member** (`nemo-gym`) rather than the CWD package (`nemo-rl`). Since `nemo-gym` has no extras, any `--extra X` call fails.

The committed `uv.lock` on `main` is **revision 3** (generated when nothing changes) and works fine. But as soon as any dependency changes and `uv lock` runs, the lockfile becomes revision 4 and all `--extra X` syncs break.

## Fix

Add `--package nemo-rl` to all `uv sync` calls in the hermetic build stage. This makes workspace package selection explicit regardless of lockfile revision or member ordering.

## Testing

Verified via the nemo-ci mbridge bump automation (`build-NeMo-RL-mbridge-bump` job) which exercises exactly this code path on a freshly-relocked bump branch.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -220,17 +220,19 @@ fi
 # Do everything in one layer to prevent large layers.
 
 # The venv is symlinked to avoid bloating the layer size
-UV_LINK_MODE=hardlink uv sync --frozen --no-install-project
+# --package nemo-rl pins the sync target explicitly so uv resolves extras against
+# nemo-rl's pyproject.toml regardless of workspace member ordering in the lockfile.
+UV_LINK_MODE=hardlink uv sync --frozen --package nemo-rl --no-install-project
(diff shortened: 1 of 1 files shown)


kajalj22:
Closing — needs further review on whether --package nemo-rl skips nemo-gym base dependency installation in the container. Root cause and fix documented in research and investigation notes.

PR #3550, opened 2026-08-08
feat: add GDPO research project

kashif:
## What

Adds a self-contained `research/gdpo` project for training masked diffusion language models such as LLaDA with the GDPO method (https://arxiv.org/abs/2510.08554).

GDPO estimates sequence likelihood with Semi-deterministic Monte Carlo over the masked-diffusion ELBO, then uses a sequence-level policy ratio. It is unrelated to the existing multi-reward `gdpo` advantage estimator.

## Design

- `research/gdpo/gdpo/` owns the SDMC estimator, block denoiser, validation, Automodel generation adapter, GDPO training loop, and `DTensorGDPOPolicyWorker` extension.
- `research/gdpo/gdpo.py` is the dedicated entrypoint.
- Core GRPO only gains generic policy/generation extension hooks; it no longer imports diffusion-specific code.
- Likelihood settings live under `policy.masked_diffusion.likelihood`.
- Rollout settings live under `policy.generation.backend: automodel` and `policy.generation.denoise_cfg`.
- Rollouts run on the live training workers, so generation is colocated and weight synchronization is a no-op.

The exemplar follows the published GDPO setup: LoRA r=128/alpha=64, gauss-3 SDMC, block length 32, 128 diffusion steps, epsilon 0.5, beta 0, and learning rate 3e-7. Full fine-tuning has not been qualified in this PR; it is not known to be required for stability and needs substantially more activation memory.

## Testing

- 147 GDPO unit tests pass.
- 190 affected core tests pass; 79 GPU-only tests skip on the CPU host.
- 16 recipe/config consistency tests pass.
- Full repository pre-commit passes, including Ruff, formatting, Pyrefly, and recipe minimization.
- `uv lock --check` passes and the GDPO entrypoint imports and exposes `--help`.

The earlier 2xH100 LLaDA-8B-Instruct smoke run completed three GSM8K steps. The 1n8g research recipe remains to be qualified end to end.


kashif:
cc @kevinrojas1499

research/gdpo/configs/gdpo_llada_8b.yaml
```diff
@@ -0,0 +1,113 @@
+# GRPO for masked diffusion language models (dLLMs).
+#
```
sajadn, line 2: i don't see the Lora configs that are mentioned in the PR description, have you ran experiments using full fine-tuning or do we need to update the recipe? i think quadrature is another config with inconsistencies

kashif: ok fixing!

kashif:
@sajadn if you can kindly check now appreciate it!

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -606,6 +641,220 @@ def on_microbatch_start(mb_idx):
             return metrics
 
     @wrap_with_nvtx_name("dtensor_policy_worker_v2/get_logprobs")
+    def _dllm_elbo_logprobs(
```
sajadn: can we introduce a new `DTensorGDPOPolicyWorker` and inherits `DTensorPolicyWorkerV2Impl` and introduces the additional methods of `_dllm_elbo_logprobs`, `generate` and `_dllm_train_elbo_scorer`?

nemo_rl/algorithms/grpo.py
```diff
@@ -1411,6 +1422,26 @@ def init_nemo_gym():
             flush=True,
         )
 
+    elif backend == "dllm":
```
sajadn: having `dllm` as an additional generation backend at the same level to `megatron`, `vllm` and `sglang` can be a bit confusing. The ideal solution will be using either `vllm ` or `sglang` given that they added support for dllm. But still we can use this version with some code changes. I suggest you introduce a new algorithm file something like `algorithm/diffusions/gdpo.py` and move your code there. I think you can get rid of the parts that are not necessary for you. Also rename the backend to something like `automodel` rather than `dllm` although at this point it only has support for dllm.

nemo_rl/models/policy/dllm/config.py
```diff
@@ -0,0 +1,124 @@
...
+
+from pydantic import BaseModel, Field
+
+_NonNegativeInt = Annotated[int, Field(ge=0)]
+_PositiveInt = Annotated[int, Field(ge=1)]
+_Probability = Annotated[float, Field(ge=0.0, le=1.0)]
+_NonNegativeFloat = Annotated[float, Field(ge=0.0)]
+
+
+class DllmConfig(BaseModel, extra="allow"):
```
sajadn: this is shared between log probability estimation and generation configs, right?

nemo_rl/models/policy/__init__.py
```diff
@@ -582,6 +583,11 @@ class PolicyConfig(TypedDict):
     draft: NotRequired[DraftConfig | DraftConfigDisabled]
     pretrained_checkpoint: NotRequired[PretrainedCheckpointConfig]
     router_replay: NotRequired[RouterReplayConfig | RouterReplayConfigDisabled]
+    # Masked diffusion LM policy. When enabled, sequence likelihoods are ELBOs
+    # estimated by SDMC quadrature instead of autoregressive token logprobs.
+    # Loaded from YAML as a plain dict; coerce it with
+    # nemo_rl.models.policy.dllm.dllm_config_from_policy.
+    dllm: NotRequired[DllmConfig]
```
sajadn: i think we need the separation of the generation configs form policy configs. Also we might need to add additional approaches for computing the log likelihood other than `GDPO`. A config like below might be a good setup:

```
policy:
    masked_diffusion:
      enabled: true
      mask_id: null
      shift_targets: false
      likelihood:
        type: sdmc
        quadrature: gauss-2
        mc_samples: 1
  
    generation:
      backend: automodel          # which runtime
      denoise_cfg:
        type: block               
        block_length: 32
        diffusion_steps: 128
        temperature: 1
```

nemo_rl/models/automodel/train.py
```diff
@@ -355,6 +390,26 @@ def forward_with_post_processing_fn(
     # (NVIDIA-NeMo/Automodel#1705).
     use_cache = _needs_kv_cache_for_shared_layers(model)
 
+    # Masked diffusion policies are scored by an ELBO accumulated over several
+    # masked views of the sequence, so there is no single logits tensor to hand
+    # the loss. The accumulation stays differentiable through every forward, so
+    # backward flows through all quadrature points.
+    if elbo_scorer is not None:
```
sajadn: how about creating a `train_gdpo.py` to keep things a bit separated?

sajadn:
Hey, 
Thanks for the PR. I put some comments that I appreciate you take a look. 
This is among the first dLLM PRs in the repo so we might need to do a few iterations until it lands. 

As you mentioned yourself, if we move the generation to either vLLM/Sglang things are gonna get much cleaner but it can be the follow up PR. Overall we start to merge research focused PRs into a newly introduced `/research` folder that we appreciate if you move your contents there. 

I'll keep looking at the PR and let you know in case I come up with more comments/thoughts. 
Cheers!

research/gdpo/gdpo/elbo.py
```diff
@@ -0,0 +1,420 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
```
sajadn, line 2: i think this whole file is better to be moved to the `algorithm` folder

sajadn:
And out of curiosity, have you tried recipes without Lora or is Lora necessary for stable training.

research/gdpo/configs/gdpo_llada_8b.yaml
kashif: The exemplar now matches the published setup: LoRA r=128/alpha=64 and gauss-3. I have not qualified full fine-tuning, and there is no evidence here that it is required for stability; it also needs substantially more memory because SDMC retains several forward graphs.

nemo_rl/models/policy/__init__.py
kashif: Done in `ee9296fdc`. Likelihood configuration is under `policy.masked_diffusion.likelihood`; denoising is under `policy.generation.denoise_cfg`; the generation backend is `automodel`.

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
kashif: Done in `ee9296fdc`. GDPO now uses a research-owned `DTensorGDPOPolicyWorker` subclass; ELBO scoring, the differentiable training scorer, and generation are no longer implemented in the core DTensor worker.

nemo_rl/algorithms/grpo.py
kashif: Done in `ee9296fdc`. The implementation is now the standalone `research/gdpo` project with its own entrypoint, and the runtime is named `automodel`. Core GRPO only exposes generic policy/generation factory hooks.

research/gdpo/gdpo/elbo.py
kashif: Done in `ee9296fdc`. The estimator moved out of policy code into the standalone `research/gdpo` algorithm package.

nemo_rl/models/automodel/train.py
kashif: Done in `ee9296fdc`. The differentiable SDMC forward/backward path now lives in `research/gdpo/gdpo/train_gdpo.py`.

kashif:
Addressed the review in `ee9296fdc`: GDPO is now a standalone `research/gdpo` project with a dedicated worker, training loop, Automodel generation adapter, split likelihood/denoising config, moved tests, and research docs. The published LoRA r=128/alpha=64 setup is restored; full fine-tuning has not been qualified and is not known to be necessary for stability. All local unit, config/recipe, lock, and pre-commit checks pass.

research/gdpo/configs/gdpo_llada_8b.yaml
kashif: Follow-up in `b74de6d7b`: I also aligned the less-visible LoRA details with upstream GDPO—dropout 0.05, pre-dropout semantics, Kaiming-uniform A initialization, and exactly q/k/v/o/up/down/gate projections rather than NeMo RL’s inherited all-linear defaults.

sajadn:
Hi Kashif,

I still have two concerns: 
1. Previous and current log probabilities use different scoring paths. The GDPO worker overrides _logprobs_for_microbatch() to compute SDMC/ELBO scores, but the base worker’s get_logprobs() loop does not call that hook. Previous-policy scores therefore come from ordinary scoring on clean inputs, while current-policy scores use masked ELBO estimation. This makes the importance ratio compare different quantities. can you tell if this is intentional or if it's a bug?

2. Generation ignores the remaining sequence budget. I ran the unchanged gdpo-llada-8b-instruct-1n8g-fsdp2tp1.yaml recipe on eight H100s. It reached initial validation, then failed with:

     AssertionError: tokens_left_for_obs=-131 should not be negative.
     This should not happen if the inference engine respects the max sequence length.
The sampler reserves 256 generation slots without accounting for prompt length,

btw, I started refactoring your code a bit and I plan to open a PR from my branch (https://github.com/sajadn/RL/tree/refactor/gdpo-research) to yours so you can review, but I'll hold until we resolve the two issues mentioned above.

kashif:
@sajadn feel free to also commit to my PR directly as you have permission. The two issues are fixed

sajadn:
@kashif thanks, do you have any experiment results using this branch? something that I can rerun after the refactor to make sure nothing is broken

kashif:
let me ask the other co-authors @sajadn

sajadn:
NemoRL is going to make a release this week so in case you like to be included in that release please let me know about your experimental results so we can verify it.

kashif:
Did a quick sanity run on a single H100 after merging latest main: LLaDA-8B-Instruct, LoRA r128, GSM8K, 2 prompts × 6 generations, 12 updates per rollout batch, lr 3e-7, clip 0.5 (i.e. the official per-device GDPO settings), 10 steps.

It runs end to end, ~95s/step (~30s denoising, ~65s for the 12 updates):

| step | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
|---|---|---|---|---|---|---|---|---|---|---|
| reward | 1.00 | 0.83 | 0.42 | 0.92 | 0.33 | 0.67 | 1.00 | 0.92 | 0.92 | 1.00 |
| grad norm | 0 | 0.037 | 0.040 | 0.074 | 0.076 | 0.062 | 0 | 0.023 | 0.014 | 0 |

Grad norm is exactly 0 on the all-correct steps (zero advantage) and non-zero otherwise, and the ratios stay within ~0.996–1.007. Far too short to say anything about learning, so treat it as a plumbing check.

I also compared against the reference repo on the same inputs: greedy generation is token-identical on 4 prompts, and the gauss-3 ELBO agrees to 0.1–0.3% given the same masks.

The merge with main broke a few upstream calls (`LogprobsPostProcessor` ctor, `train_context_fn`, `cp_sharder`), and step 1 was denoising on CPU-offloaded weights. Fixed in 7760ab9af5945acb9d2f25be4e9968936a2571ec, which also adds a CPU test that pins those signatures so this kind of drift fails in L0 next time.

PR #3552, opened 2026-08-08
perf(imports): stop loading optional integrations to import grpo

tianyi-zhang-02:
# What does this PR do?

Keeps optional integrations out of `import nemo_rl.algorithms.grpo` until their feature is used.

- `wandb`, `mlflow`, `swanlab`, and `matplotlib.pyplot` use patch-compatible lazy module proxies in `logger.py`.
- W&B tables, Ray's CLI memory helper, FastAPI, and Uvicorn move to their call sites.
- FastAPI handler imports remain runtime-visible so request annotations resolve correctly.
- `should_log_nemo_gym_full_result_tables` is forwarded through both legacy and SingleController rollout paths. When disabled, neither path imports W&B nor constructs a `wandb.Table`.

No config or public API changes. Supersedes #3437.

# Validation

Current head: `2f520276e84c92f3b9e4372885b939ff8480cdd1`, including `main` at `fd7112c374af23b3bcc392637fb737495f8a3f31`.

On 2026-09-21, the four changed import-isolation, rollout, and setup test files passed: **258 passed**, 0 failed/skipped. Environment: Runpod Secure Cloud, 1× H100 80 GB, `nvcr.io/nvidia/nemo-rl:v0.7.0`, driver 580.126.09, CUDA 13.0, PyTorch 2.11.0+cu130, with `HF_HOME=/workspace/huggingface` for recipe interpolation. The current-main test fixture required `pytest==9.0.3` and pinned `nemo-lens` commit `b85578f`, installed only in the temporary Pod.

Earlier, at head `99e4bb91983c059c7f4f4c39bd5e8263d986d574`, seven isolated subprocess imports on the same warm filesystem cache gave:

| | `main` median | This PR median | Change |
| --- | ---: | ---: | ---: |
| `import nemo_rl.algorithms.grpo` | 7.939 s | 6.237 s | -21.4% |
| watched optional modules loaded | 7 | 0 | all deferred |

The watched set is `wandb`, `mlflow`, `swanlab`, `matplotlib.pyplot`, `fastapi`, `uvicorn`, and `ray.scripts.scripts`. This import microbenchmark was not rerun after the main-only refresh.


tianyi-zhang-02:
Force-pushed after rebuilding this on current main. Two things were wrong with the previous version, both worth recording:

**It silently reverted ~115 lines of the vLLM 0.25 migration.** I had taken `vllm_worker_async.py` wholesale from an older branch, so the diff carried that branch's version of the whole file rather than just the import change. Against current main that reads as a revert: `OnlineRenderer` back to `OpenAIServingRender`, the `parser=` argument back to the `tool_parser=`/`reasoning_parser=` pair, the `spec_lookahead` handling deleted, and — worst — the `val_temperature`/`val_top_p` validation-sampling branch replaced by a bare `assert request.temperature == generation_config["temperature"]`, which would have failed every validation rollout. It merged cleanly because main had not touched those exact lines, so nothing would have flagged it.

**`Request` was left as an unquoted annotation** on two nested `async def`s in `vllm_sparse_refit.py` after moving the import under `TYPE_CHECKING`. Annotations on a nested `def` evaluate when the enclosing function runs, so that is a `NameError` at request-handling time, not a typing nit. `ruff check --select TC` catches it; the repo's ruff config selects only `D,F`, so CI would not have.

The branch now differs from main by 11 lines in `vllm_worker_async.py` — only the import deferral and two quoted annotations.

Re-measured on the rebuilt branch, warm cache, median of 7:

| | before | after |
|---|---|---|
| `import nemo_rl.algorithms.grpo` | 7.71 s | 3.42 s (−56%) |
| modules | 7098 | 5203 (−27%) |

`wandb`, `mlflow`, `fastapi` and `uvicorn` are absent from `sys.modules` afterwards. 196 tests across `tests/unit/utils/test_logger.py` and `tests/unit/experience/` pass identically before and after, no test modified. `ruff check`, `ruff format --check` and `ruff check --select TC` all clean.

yuki-97 (commented):
@RayenTian to review

RayenTian:
/ok to test [b677b51](https://github.com/NVIDIA-NeMo/RL/pull/3552/commits/b677b51e1fb62dffb66c61f1f0c1328d2c2d2b8c)

nemo_rl/utils/logger.py
```diff
@@ -43,6 +47,41 @@
...
+
+    def __getattr__(self, attr: str) -> Any:
+        if self._module is None:
+            import importlib
+
+            self._module = importlib.import_module(self._module_name)
+        return getattr(self._module, attr)
+
+
+mlflow = _LazyModule("mlflow")  # type: ignore[assignment]
```
RayenTian, line 81: [`logger.py:79`](https://github.com/NVIDIA-NeMo/RL/blob/0742bc5a9e31dff7910ac4687904101ad23f4958/nemo_rl/utils/logger.py#L79)

The lazy-import win isn't guarded by a test — a future top-level `import wandb` would silently undo it. Consider adding:

```python
# tests/unit/algorithms/test_grpo_import_isolation.py (new file)
import subprocess, sys
import pytest

@pytest.mark.parametrize("mod", ["wandb", "mlflow", "fastapi", "uvicorn"])
def test_grpo_import_stays_lazy(mod):
    r = subprocess.run([sys.executable, "-c",
        f"import sys, nemo_rl.algorithms.grpo; assert {mod!r} not in sys.modules"],
        capture_output=True, text=True)
    assert r.returncode == 0, r.stderr
```

nemo_rl/experience/metric_utils.py
```diff
@@ -34,6 +32,8 @@ def calculate_single_metric(
     Returns:
         Dict mapping "{key_name}/{stat}" to its value for stat in mean, max, min, median, stddev (nan for a single value), and histogram (a wandb.Histogram).
     """
+    from wandb import Histogram
```
RayenTian: [`metric_utils.py:35`](https://github.com/NVIDIA-NeMo/RL/blob/0742bc5a9e31dff7910ac4687904101ad23f4958/nemo_rl/experience/metric_utils.py#L35)

Deferred imports should note *why* (like [`memory_tracker.py:42`](https://github.com/NVIDIA-NeMo/RL/blob/0742bc5a9e31dff7910ac4687904101ad23f4958/nemo_rl/utils/memory_tracker.py#L42) does in this PR). Same applies at `rollout_manager.py:598`, `rollouts.py:2564`, `vllm_sparse_refit.py:420` & `:535`.

```suggestion
    # defer optional wandb import so importing grpo doesn't load it
    from wandb import Histogram
```

RayenTian (commented):
Import-hygiene refactor to keep optional integrations (wandb/mlflow/swanlab/matplotlib/fastapi/uvicorn/ray.scripts) out of the `import nemo_rl.algorithms.grpo` path. Reviewed by a 3-agent team + devil's-advocate pass: **no correctness issues** — every deferred name is covered by an in-scope import and all annotations referencing deferred modules are quoted. A repo-wide scan confirms no other module-top eager imports of these deps remain. Two low-severity suggestions inline; neither blocks merge.

_Generated by Claude Code_

RayenTian:
Thanks for adding this @tianyi-zhang-02! Left some comments and triggered CI.

tianyi-zhang-02:
Both taken, thanks.

Added the guard test — used your shape, with a subprocess for the same reason you did, since another test may already have the module in `sys.modules`. Extended the list to `swanlab` and `matplotlib.pyplot` as well, since those were part of the win. All six fail on main and pass here.

Also added the "why" comment at each deferred site: `metric_utils`, `rollout_manager`, `rollouts`, `vllm_sparse_refit` (both) and `vllm_worker_async` (both — you did not name those, but they are the same pattern).

One thing I put in the `fastapi` ones beyond "it is optional", because it cost me a debugging round. Those annotations have to stay unquoted at runtime. FastAPI resolves handler annotations against the endpoint's `__globals__`, so a quoted annotation whose name only exists under `TYPE_CHECKING` leaves an unresolved `ForwardRef`. It does not raise — the parameter is silently demoted to a query param and every request 422s. That is why these are local runtime imports rather than `TYPE_CHECKING` plus quotes :)

RayenTian:
/ok to test [1c9ff15](https://github.com/NVIDIA-NeMo/RL/pull/3552/commits/1c9ff15b8fabe953882cd50f3d3ec1cb9f295990)

tianyi-zhang-02:
Merged current main in. Should be a clean merge w/ no conflicts. 
Needs another `/ok to test` whenever you get a chance :)

tianyi-zhang-02:
Merged current main in — #3748 landed and conflicted with this branch in `metric_utils.py`. Resolved and pushed as `b48c850`.

The conflict itself was trivial (#3748 added `is_histogram_metric` where this branch had moved the wandb import out of module scope; both kept). But #3748 also made part of this PR obsolete in a good way, which is worth flagging since it changes the diff:

`calculate_single_metric` used to return `Histogram(values)` for the histogram key, which is why the wandb import needed deferring rather than removing. #3748 moved that serialization into the logger and left `metric_utils` returning `list(values)`, so the import has no remaining use at all. `ruff` caught it as F401. I dropped it rather than keeping it behind a `noqa` — **`wandb` no longer appears in `metric_utils.py` in any form**, which is a slightly stronger result than what this PR originally did there.

`test_grpo_import_isolation.py` + `test_logger.py` + `tests/unit/experience/`: 340 passed, ruff clean. The six-module guard still passes.

Needs a fresh `/ok to test b48c850bbb34de66aa660151212aabbdd607a5c7` when you have a moment @RayenTian — the last one pointed at `1c9ff15`, which two merges have since moved past.

tianyi-zhang-02:
Head is now `85a69e708` — main merged in since your `/ok to test 1c9ff15`, so that pin is stale. Needs a fresh one whenever you get a chance.

nemo_rl/utils/logger.py
tianyi-zhang-02: The test is already in the PR, `tests/unit/algorithms/test_grpo_import_isolation.py`, subprocess-isolated like your sketch.

Slightly wider list than you suggested: wandb, mlflow, swanlab, matplotlib.pyplot, fastapi, uvicorn. 85a69e708 adds a second one covering `_LazyModule` itself, since the first would still pass if `_LazyModule` quietly stopped resolving the real module on attribute access.

tianyi-zhang-02:
Final follow-up on SHA `99e4bb91983c059c7f4f4c39bd5e8263d986d574`:

- kept the subprocess guard and deferred-import comments from your review;
- closed the remaining behavior gap by forwarding `should_log_nemo_gym_full_result_tables` through both legacy and SingleController rollouts, so the disabled path does zero W&B imports and zero `Table` construction;
- on the Runpod image, seven subprocesses measured `import grpo` at `7.939 s → 6.237 s` median, with all seven watched optional modules loaded on `main` and none on this branch;
- 235 focused tests pass, including the real FastAPI `TestClient` path; Ruff check/format are clean.

cc @RayenTian — this is the final branch after the import-isolation review.

nemo_rl/experience/metric_utils.py
tianyi-zhang-02: Done in 1c9ff15, and still true at the current head.

`rollout_manager.py` and `rollouts.py` say W&B is optional and keep the import behind the same gate as table construction. In `vllm_sparse_refit.py`, the `uvicorn`/`FastAPI` one says they're only needed when the worker actually serves refit over HTTP. The `Request` one has a different reason, since that deferral isn't about import cost: FastAPI resolves endpoint annotations against module globals, and a `TYPE_CHECKING`-only name becomes an unresolved `ForwardRef` that turns the parameter into a query param and 422s every request.

`metric_utils.py` is no longer in this diff: #3748 removed the wandb import there on main, and this branch took main's version when I merged it in.

PR #3564, opened 2026-08-10
perf(xtoken): project only the top-k teacher columns in P-KL

tianyi-zhang-02:
# What does this PR do ?

`_compute_p_kl` projected the student distribution onto the **full** teacher vocabulary and then kept `vocab_topk` of those columns:

```python
projected_full = project_student_to_teacher_vocab(...)     # [B, T_s_local, V_t]
projected_topk = projected_full[..., global_top_indices]   # [B, T_s, k]
```

`projected_full` had exactly two consumers: a `.shape[-1]` read and that gather. At the shipped `vocab_topk: 8192` against a ~152k teacher vocab, ~95% of the sparse matmul's output columns were computed, materialized in fp32, all-reduced across the TP group, and thrown away — and it stays live through the KL at `:2401-2408`, so it is peak memory, not transient. This folds the column slice into the projection matrix so only the k columns are produced.

**Why the reorder is legal.** `project_student_to_teacher_vocab` is `torch.sparse.mm(M.t(), dense)`, so teacher column *j* is `Σ_s M[s,j]·p_s` — an independent contraction over the student axis. Restricting `M`'s columns first is value-preserving, and the backward `grad_dense = torch.sparse.mm(M, grad_out)` is structurally zero outside those columns. A renormalization over `V_t` would break it; there is none — the only one is *within* the top-k subset (`:2363`). Chunk averaging in between is a `bmm` over the token axis, which commutes with a vocab slice. Hoisting the top-k selection is safe because `select_teacher_topk_indices` reads only `teacher_full_logits` under `no_grad`, and the vocab width is available as `sparse_projection.size(1)`.

**Numerics.** Both sides land at fp32 round-off: forward differs by 1.1e-07 relative, gradient by at most 4.8e-07. At the shipped config (4 nonzeros/row, k=8192) the gradient is bitwise identical, but that is not universal — it degrades to fp32 ulp as the matrix densifies or k grows.

# Issues

No existing issue.

# Usage

No config change. A `k == V_t` fast path skips the slice entirely, so small-vocab tests are unaffected — verified: the slice preserves `_nnz()` and the projection is bitwise equal to using the unsliced matrix.

RTX 4090, V_s=151936, V_t=128256, B×T = 1×512, 4 nonzeros per student row. Note this sweep has the vocab sizes the other way round from the shipped recipe, where the student is Llama-3.2-1B (V_s≈128k) and the teacher is Qwen3-4B (V_t≈152k). The reduction scales with `V_t / vocab_topk`, so the shipped direction is 18.5× rather than the 15.7× measured here (`build_projection_matrix.sh` default), forward + backward, marginal peak:

| `vocab_topk` | peak before | peak after | ms before | ms after |
|---|---:|---:|---:|---:|
| 1024 | 1.409 GiB | 1.164 GiB | 53.7 | 9.8 |
| 4096 | 1.415 GiB | 1.176 GiB | 54.5 | 12.8 |
| **8192** (shipped) | 1.423 GiB | 1.192 GiB | 55.4 | 14.8 |
| 32768 | 1.469 GiB | 1.288 GiB | 58.8 | 22.8 |

`V_t / vocab_topk` = 15.7× in this sweep (18.5× in the shipped direction) is the reduction **on the projected tensor** and on the all-reduce, not end-to-end: marginal peak drops 1.14–1.21× across the sweep because `student_probs` dominates. Speed is the larger effect at TP=1 — 3.7× at the shipped k, 5.5× at k=1024, 2.6× at k=32768.

The win grows with TP, since the student probs shrink to `V_s/TP` per rank while the projected tensor stays full `V_t`. Per-rank shapes on one GPU, same methodology (all-reduce column is payload size, not measured throughput):

| TP | probs per rank | peak before | peak after | | all-reduce per rank |
|---:|---|---:|---:|---:|---|
| 1 | [1, 512, 151936] | 1.423 GiB | 1.191 GiB | 1.19× | 250 → 16 MiB |
| 2 | [1, 512, 75968] | 0.893 GiB | 0.611 GiB | 1.46× | 250 → 16 MiB |
| 4 | [1, 512, 37984] | 0.813 GiB | 0.321 GiB | 2.53× | 250 → 16 MiB |
| 8 | [1, 512, 18992] | 0.775 GiB | 0.180 GiB | 4.31× | 250 → 16 MiB |

**TP composition, measured.** Two real processes on a gloo `ProcessGroup` with the student probs sharded across them (V_s=4096, V_t=2048, k=256):

| | rank 0 | rank 1 |
|---|---:|---:|
| pre-PR TP=2 vs TP=1 | 4.77e-07 | 4.77e-07 |
| post-PR TP=2 vs TP=1 | 4.77e-07 | 4.77e-07 |
| post-PR vs pre-PR, both TP=2 | **0.00e+00** | **0.00e+00** |

At the same TP degree the two are bitwise identical, so the column slice composes exactly with the per-rank row slice. The 4.77e-07 against TP=1 is TP's own partial-sum all-reduce, unchanged by this PR. Measured on gloo. At TP=2 each output element is a two-rank sum, so the result is backend- and algorithm-independent; at higher TP degrees NCCL may select different reduction algorithms for the two payload sizes (250 MiB vs 16 MiB), so I would expect fp32-ulp agreement there rather than bitwise.

- [x] Did you run the unit tests and functional tests locally? (unit only)
- [x] Did you add or update any necessary documentation? (n/a — no user-facing surface changes)

# Additional Information

`grep -rn "_compute_p_kl" tests/` returns nothing, so P-KL has no unit coverage today. Added four CPU tests for `slice_sparse_projection_cols` (dense-slice equality, project/slice commutation, empty columns surviving as zeros, all-columns identity). `tests/unit/algorithms/x_token/` plus `test_loss_functions.py` go 91 → 95 passed, no new failures. `ruff` clean.

Two things I could not check: whether this is the nightly x-token path rather than opt-in (I read it as the former — `distillation-xtoken-off-policy-qwen3-4b-to-llama3.2-1b-1n8g-dtensor-tp4cp2.sh` passes `teachers.0.projection_matrix_path`, and `gold_loss: false` is the default), and whether the gold path at `:2496` wants the same treatment.

cc @RayenTian @yuki-97


yuki-97 (commented):
@RayenTian to review, also cc @avenkateshha

RayenTian (approved):

RayenTian:
/ok to test [853f3c0](https://github.com/NVIDIA-NeMo/RL/pull/3564/commits/853f3c0295e500fe534657b59e45efeccb95202b)

RayenTian:
Thanks for making NeMoRL better. LGTM! Approved!

PR #3569, opened 2026-08-10
feat: overlap NeMo-Gym spinup with MInf load

### LINKED ISSUE #3775: mcore inference HTTP server port pick is TOCTOU-prone; mirror vLLM's held-socket reservation
The mcore inference HTTP server picks its port by scanning, releasing, and later re-binding. Another process on the node can take the port in that gap. vLLM already solved this; mcore should do the same.

Pre-existing — not introduced by any particular PR. Surfaced while reviewing #3731.

## The gap

Permalinks pinned at `dfdf6894d5380f13f8b9c285315aa0f2ee99dafb`.

1. [`megatron_worker.py:317`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/generation/megatron/megatron_worker.py#L317) calls `_get_free_port_local()`.
2. That helper binds, listens, then **closes** the socket and returns only the integer — the `with` block at [`virtual_cluster.py:235-245`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/distributed/virtual_cluster.py#L235-L245) exits before `return port`.
3. The integer is passed as [`server_port=free_port`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/generation/megatron/megatron_worker.py#L323).
4. Upstream re-binds it at [`text_generation_server.py:184`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_server/dynamic_text_gen_server/text_generation_server.py#L184).

Between 2 and 4 the port is free for anyone to take.

The pool is `[1400, 1999)` (`DEFAULT_MASTER_PORT_RANGE_*`, [`virtual_cluster.py:128-129`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/distributed/virtual_cluster.py#L128-L129)) — shared with Ray master ports, so it is not private to the HTTP server.

## Failure modes

- **Usual:** an uncaught `OSError: [Errno 98] Address already in use` from the re-bind at step 4.
- **Quieter and worse:** upstream sets `SO_REUSEPORT` when available ([`text_generation_server.py:177-181`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_server/dynamic_text_gen_server/text_generation_server.py#L177-L181)). If the process that took the port also set `SO_REUSEPORT` under the same UID, **both sockets bind and the kernel splits incoming connections between them** — no error, just some requests going to the wrong server.

Only global rank 0 starts a server ([`megatron_worker.py:362-366`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/generation/megatron/megatron_worker.py#L362-L366)), so there is no contention between mcore servers inside one job. The exposure is co-tenants on the same node — which is exactly the contention vLLM hit.

## How vLLM solved it

vLLM holds the socket open from reservation through serving, so there is no gap:

- [`vllm_worker_async.py:132-153`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/generation/vllm/vllm_worker_async.py#L132-L153) `_reserve_port()` binds and keeps a listening socket on the worker.
- [`vllm_worker_async.py:921-928`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/generation/vllm/vllm_worker_async.py#L921-L928) hands that same socket to uvicorn via `sockets=`.

Its docstring states the invariant directly: the socket *"is never closed and re-opened, so there is zero gap where another process could steal the port."*

Related prior work on vLLM port contention: #2380, #3103.

## Suggested fix

Upstream already supports the handoff — no Megatron-LM change needed. [`start_text_gen_server`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_server/dynamic_text_gen_server/text_generation_server.py#L144-L153) accepts `sock: Optional[socket.socket]`, documented at [L163](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_server/dynamic_text_gen_server/text_generation_server.py#L163) as *"The caller may pass in a socket it has already bound ahead of time."* It validates the socket is bound to a real port and reuses it.

So in `_setup_openai_api_server`: bind and hold the socket (as `_reserve_port` does), then pass it through as `sock=` instead of `server_port=`.

There is a second benefit. Doing the reservation at construction time also makes the URL available before the engine is loaded, which is how vLLM populates `dp_openai_server_base_urls` under `defer_model_load` ([`vllm_generation.py:276-279`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/generation/vllm/vllm_generation.py#L276-L279)). That is the same change #3731 needs for its deferred-init path, so the two can land together.

tdene:
See #3499

**Baseline (`36b5999a7`):**
```
▶ Worker Initialization Timing:
  Generation init: 134.8s
  Policy init: 38.1s
  NeMo-Gym init: 32.5s
  Other setup: 2.7s
  Total setup: 226.2s
```

**Branch (`tde/megatron_gym_overlap`):**
```
▶ Worker Initialization Timing:
  Generation init: 132.6s (+ reserve 1.0s)
  Policy init: 39.4s
  NeMo-Gym init: 64.1s
  Other setup: 3.6s
  Total setup: 194.8s
```

Total wall-time is 226.2s → 194.8s (−31.4s, −14%).

# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1289,11 +1289,15 @@ def setup(
                 "  ⚡ Reserving the Megatron server address for overlapped NeMo Gym init",
                 flush=True,
             )
+            reserve_t0 = time.perf_counter()
             reserved_url, reserved_http_server_port, port_holder = (
                 MegatronGeneration.reserve_http_server_address(
                     train_cluster if colocated_inference else inference_cluster,
                     policy_config,
                 )
+            )
+            setup_timing_metrics.generation_init_reserve_time_s = (
+                time.perf_counter() - reserve_t0
             )
             print(f"  ✓ Reserved Megatron server URL: {reserved_url}", flush=True)
 
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1334,8 +1334,7 @@ def setup(
             ]
             nemo_gym_actor, nemo_gym_time = results["nemo_gym"]
             setup_timing_metrics.policy_init_time_s = policy_time
-            setup_timing_metrics.megatron_generation_init_time_s = megatron_gen_time
-            setup_timing_metrics.generation_init_time_s = (
+            setup_timing_metrics.megatron_generation_init_time_s = (
                 reserve_time + megatron_gen_time
             )
             setup_timing_metrics.nemo_gym_init_time_s = nemo_gym_time
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1283,28 +1283,13 @@ def setup(
     # Handle generation-specific setup
     if backend == "megatron":
         mcore_generation_config = generation_config["mcore_generation_config"]
-        http_port_range_low = mcore_generation_config.get("http_server_port_range_low")
-        http_port_range_high = mcore_generation_config.get(
-            "http_server_port_range_high"
-        )
-        if (http_port_range_low is None) != (http_port_range_high is None):
-            raise ValueError(
-                "Set both policy.generation.mcore_generation_config."
-                "http_server_port_range_low and http_server_port_range_high "
-                f"(or neither); got low={http_port_range_low}, "
-                f"high={http_port_range_high}."
-            )
-        overlap_nemo_gym_spinup = enable_nemo_gym and http_port_range_low is not None
 
-        if overlap_nemo_gym_spinup:
-            # ---- NeMo Gym: probe the rank-0 node for a free server port so the
-            # URL is known before any worker exists, then spin up NeMo Gym while
-            # the policy loads and the inference engine warms up.
+        if enable_nemo_gym and mcore_generation_config["expose_http_server"]:
             print(
                 "  ⚡ Reserving the Megatron server address for overlapped NeMo Gym init",
                 flush=True,
             )
-            reserved_url, reserved_http_server_port = (
+            reserved_url, reserved_http_server_port, port_holder = (
                 MegatronGeneration.reserve_http_server_address(
                     train_cluster if colocated_inference else inference_cluster,
                     policy_config,
@@ -1336,9 +1321,12 @@ def setup(
                 "nemo_gym": init_nemo_gym,
             }
             print(f"  ⚡ Init tasks: {', '.join(init_tasks.keys())}", flush=True)
-            with ThreadPoolExecutor(max_workers=len(init_tasks)) as executor:
-                submitted = {k: executor.submit(fn) for k, fn in init_tasks.items()}
-                results = {k: f.result() for k, f in submitted.items()}
+            try:
+                with ThreadPoolExecutor(max_workers=len(init_tasks)) as executor:
+                    submitted = {k: executor.submit(fn) for k, fn in init_tasks.items()}
+                    results = {k: f.result() for k, f in submitted.items()}
+            finally:
+                ray.kill(port_holder)
 
             policy, policy_time, policy_generation, megatron_gen_time = results[
                 "megatron"
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1102,11 +1102,6 @@ def setup(
 
     # vllm model loading prefers clean environment, initialize policy_generation before policy in colocated mode
     backend = generation_config["backend"]
-    gen_init_time_key = (
-        "megatron_generation_init_time_s"
-        if backend == "megatron"
-        else f"{backend}_init_time_s"
-    )
     generation_config["model_name"] = policy_config["model_name"]  # Needed for vLLM
     generation_config["_debug_payload_metrics"] = grpo_config.debug_payload_metrics
     remote_transport = None
@@ -1258,7 +1253,7 @@ def setup(
             parallel_wall_time = time.perf_counter() - parallel_start_time
 
             # Store timing metrics
-            setattr(setup_timing_metrics, gen_init_time_key, generation_time)
+            setup_timing_metrics.generation_init_time_s = generation_time
             setup_timing_metrics.policy_init_time_s = policy_time
             setup_timing_metrics.parallel_wall_time_s = parallel_wall_time
             setup_timing_metrics.parallel_init_enabled = 1.0
@@ -1272,7 +1267,7 @@ def setup(
 
             # Initialize generation engine first (clean GPU memory), then policy
             policy_generation, generation_time = init_generation_fn()
-            setattr(setup_timing_metrics, gen_init_time_key, generation_time)
+            setup_timing_metrics.generation_init_time_s = generation_time
 
             policy, policy_time = init_policy()
             setup_timing_metrics.policy_init_time_s = policy_time
@@ -1334,9 +1329,10 @@ def setup(
             ]
             nemo_gym_actor, nemo_gym_time = results["nemo_gym"]
             setup_timing_metrics.policy_init_time_s = policy_time
-            setup_timing_metrics.megatron_generation_init_time_s = (
+            setup_timing_metrics.generation_init_time_s = (
                 reserve_time + megatron_gen_time
             )
+            setup_timing_metrics.generation_init_load_time_s = megatron_gen_time
             setup_timing_metrics.nemo_gym_init_time_s = nemo_gym_time
 
             served_urls = policy_generation.dp_openai_server_base_urls
@@ -1352,7 +1348,7 @@ def setup(
 
             # Colocated wraps the training policy; non-colocated builds a dedicated inference policy.
             policy_generation, megatron_gen_time = init_megatron_generation(policy)
-            setup_timing_metrics.megatron_generation_init_time_s = megatron_gen_time
+            setup_timing_metrics.generation_init_time_s = megatron_gen_time
 
         print(
             f"  ✓ Using {backend} backend for generation with {policy_config['model_name']}",
@@ -1475,7 +1471,11 @@ def setup(
                 policy_generation, vllm_load_time = results["vllm"]
                 policy, policy_time = results["policy"]
             nemo_gym_actor, nemo_gym_time = results["nemo_gym"]
-            setup_timing_metrics.vllm_init_time_s = vllm_reserve_time + vllm_load_time
+            setup_timing_metrics.generation_init_time_s = (
+                vllm_reserve_time + vllm_load_time
+            )
+            setup_timing_metrics.generation_init_reserve_time_s = vllm_reserve_time
+            setup_timing_metrics.generation_init_load_time_s = vllm_load_time
             setup_timing_metrics.policy_init_time_s = policy_time
             setup_timing_metrics.nemo_gym_init_time_s = nemo_gym_time
         else:
@@ -1611,7 +1611,7 @@ def setup(
             # Load the model weights now.
             t0 = time.perf_counter()
             policy_generation.weight_synchronizer.sync_weights()
-            setup_timing_metrics.generation_init_load_time_s = time.perf_counter() - t0
+            setup_timing_metrics.weight_sync_time_s = time.perf_counter() - t0
     # if it is not colocated inference, initialize collective communication for update weights
     elif (
         not colocated_inference
@@ -1729,7 +1729,7 @@ def setup(
     )
 
     # Log worker initialization timing metrics to logger
-    print_setup_timing_summary(setup_timing_metrics, gen_init_time_key)
+    print_setup_timing_summary(setup_timing_metrics)
     logger.log_metrics(
         setup_timing_metrics.to_metrics_dict(), step=0, prefix="timing/setup"
     )
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1282,9 +1282,7 @@ def setup(
 
     # Handle generation-specific setup
     if backend == "megatron":
-        mcore_generation_config = generation_config["mcore_generation_config"]
-
-        if enable_nemo_gym and mcore_generation_config["expose_http_server"]:
+        if enable_nemo_gym:
             print(
                 "  ⚡ Reserving the Megatron server address for overlapped NeMo Gym init",
                 flush=True,
@@ -1296,9 +1294,8 @@ def setup(
                     policy_config,
                 )
             )
-            setup_timing_metrics.generation_init_reserve_time_s = (
-                time.perf_counter() - reserve_t0
-            )
+            reserve_time = time.perf_counter() - reserve_t0
+            setup_timing_metrics.generation_init_reserve_time_s = reserve_time
             print(f"  ✓ Reserved Megatron server URL: {reserved_url}", flush=True)
 
             def init_megatron_stack():
@@ -1338,6 +1335,9 @@ def setup(
             nemo_gym_actor, nemo_gym_time = results["nemo_gym"]
             setup_timing_metrics.policy_init_time_s = policy_time
             setup_timing_metrics.megatron_generation_init_time_s = megatron_gen_time
+            setup_timing_metrics.generation_init_time_s = (
+                reserve_time + megatron_gen_time
+            )
             setup_timing_metrics.nemo_gym_init_time_s = nemo_gym_time
 
             served_urls = policy_generation.dp_openai_server_base_urls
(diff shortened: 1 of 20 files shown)


shanmugamr1992 (dismissed):

nemo_rl/algorithms/grpo.py
```diff
@@ -1272,22 +1280,88 @@ def initialize_generation_with_policy(
...
-        policy_generation, megatron_gen_time = init_megatron_generation(policy)
-        setup_timing_metrics.megatron_generation_init_time_s = megatron_gen_time
+        mcore_generation_config = generation_config["mcore_generation_config"]
 
-        if enable_nemo_gym:
-            # The Megatron inference engine must be up before its server URLs exist.
-            nemo_gym_actor, nemo_gym_time = _spinup_nemo_gym(
-                policy_generation.dp_openai_server_base_urls,
-                generation_config["model_name"],
+        if enable_nemo_gym and mcore_generation_config["expose_http_server"]:
```
yuki-97: `expose_http_server` is guaranteed true here, so the fallback below is dead code.

`_should_use_nemo_gym` already asserts it for the megatron backend ([grpo.py:2317](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/nemo_rl/algorithms/grpo.py#L2317)), so `enable_nemo_gym` implies it — which makes the `if enable_nemo_gym:` block at [grpo.py:1356](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/nemo_rl/algorithms/grpo.py#L1356) unreachable.

**Suggest fix:** drop the conjunct and the now-unused `mcore_generation_config` local above it, and delete the dead block.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -90,6 +94,52 @@ def init_cluster_placement_groups(
...
+            Tuple of (server base URL, reserved port, port-holder actor handle).
+            The caller must keep the handle referenced until rank 0 has adopted
+            the socket (worker init complete), then `ray.kill` it.
+        """
+        if config["generation"]["colocated"]["enabled"]:
+            # Colocated generation shares the training policy's cluster; trigger
+            # the same (idempotent) default placement-group init Policy would.
+            cluster.get_placement_groups()
+        else:
+            cls.init_cluster_placement_groups(cluster, config)
```
yuki-97: This call is a no-op — [line 124](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/nemo_rl/models/generation/megatron/megatron_generation.py#L124) makes the same call three lines down.

`get_placement_groups` triggers the lazy `_init_placement_groups()` itself ([virtual_cluster.py:1033](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/nemo_rl/distributed/virtual_cluster.py#L1033)), so the colocated branch adds nothing the unconditional read below wouldn't already do.

```suggestion
        # Colocated generation shares the training policy's cluster and uses the
        # default placement-group init, triggered lazily by the read below.
        if not config["generation"]["colocated"]["enabled"]:
            cls.init_cluster_placement_groups(cluster, config)
```

tests/unit/distributed/test_virtual_cluster.py
```diff
@@ -602,3 +605,110 @@ def test_default_port_ranges_ordered_and_below_ephemeral_floor():
...
+    copies cannot silently drift -- no GPU or mcore extra needed
+    (MegatronGeneration imports without megatron.core).
+    """
+    placement_groups = ["PG0", "PG1"]
+    cluster = SimpleNamespace(
+        num_gpus_per_node=8,
+        _sorted_bundle_indices=sorted_bundle_indices,
+        get_placement_groups=lambda: placement_groups,
+    )
+    config = {"generation": {"colocated": {"enabled": True}}}
```
yuki-97: This pins the one branch production never takes — every shipped megatron+gym recipe is non-colocated.

[grpo-nanov3-30BA3B-2n8g-megatron_generation-async-gym.yaml:20-21](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/examples/configs/recipes/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation-async-gym.yaml#L20-L21) sets `colocated.enabled: false`, so the reservation always goes through `init_cluster_placement_groups` at [megatron_generation.py:119](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/nemo_rl/models/generation/megatron/megatron_generation.py#L119) — which no test executes.

**Suggest fix:** parametrize `colocated` alongside `sorted_bundle_indices`; the fake cluster needs an `_init_placement_groups` attribute so the test can assert it got `strategy="PACK"` and the expected `use_unified_pg`.

nemo_rl/algorithms/metric_utils.py
```diff
@@ -77,21 +78,26 @@ def print_setup_timing_summary(
     """
     print("\n▶ Worker Initialization Timing:")
 
-    if metrics.generation_init_reserve_time_s:
+    if gen_init_time_key is not None:
```
yuki-97: Two formats for the same line: SC + gym-on **decomposes** (`15.0s (reserve 3.0s + load 12.0s)`), this branch **adds on** (`132.6s (+ reserve 1.0s)`) — a reader has to know which driver emitted the line to know whether the parenthetical is inside or outside the leading number.

And the divergence spreads rather than stays contained: the moment grpo's vllm path also reserves, `generation_init_reserve_time_s` means "already inside the total" on SC and "add this on top" on grpo, with nothing in the field to tell them apart. It's a logged metric, so that breaks consumers, not just readers.

**Suggest fix:** set `generation_init_time_s = reserve + megatron_gen_time` in grpo's megatron path, matching [setup.py:747](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/nemo_rl/algorithms/single_controller_utils/setup.py#L747), and align the rendering here with the SC branch.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -90,6 +94,52 @@ def init_cluster_placement_groups(
...
+        # Zero-gap reservation: a holder actor on the rank-0 node binds and
+        # HOLDS the socket (num_cpus=0, so it schedules even on a full bundle);
+        # rank 0 later adopts the live fd via receive_held_socket, so the port
+        # can never be stolen in between and any free port is safe.
+        holder = PortHolder.options(
+            scheduling_strategy=PlacementGroupSchedulingStrategy(
+                placement_group=placement_groups[0],
+                placement_group_bundle_index=rank0_bundle_index,
+            ),
+        ).remote()
```
yuki-97: `num_cpus=0` belongs in `.options()` rather than baked into `ray.remote(num_cpus=0)(...)` at [virtual_cluster.py:246](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/nemo_rl/distributed/virtual_cluster.py#L246) — that is the shape `_get_node_ip_and_free_port` uses at [:1091-1097](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/nemo_rl/distributed/virtual_cluster.py#L1091-L1097), where the reason sits next to the argument instead of in a comment three lines up. After applying, drop it from the `ray.remote` call.

Note `_CapturingPortHolder.options` ([test_virtual_cluster.py:620](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/tests/unit/distributed/test_virtual_cluster.py#L620)) is keyword-only on `scheduling_strategy` and needs `num_cpus` (or `**kwargs`) added, otherwise the fixture raises.

```suggestion
        holder = PortHolder.options(
            scheduling_strategy=PlacementGroupSchedulingStrategy(
                placement_group=placement_groups[0],
                placement_group_bundle_index=rank0_bundle_index,
            ),
            # Explicitly 0: otherwise this is unschedulable once the bundle's CPUs are in use.
            num_cpus=0,
        ).remote()
```

Separately, consider moving `_held_port_uds_name` / `receive_held_socket` / `_HeldPortReservation` / `PortHolder` into `nemo_rl/distributed/held_port.py` and dropping the underscore from the class, since it becomes that module's main export. `virtual_cluster.py` is 1227 lines with no other Ray actor, and the fd-handoff protocol has invariants worth isolating. The tests need to move with it, but not symmetrically. `test_reserve_http_server_address_pins_rank0_bundle` ([test_virtual_cluster.py:677](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/tests/unit/distributed/test_virtual_cluster.py#L677)), plus its `_CapturingPortHolder` fake and `patched_holder` fixture, exercises `MegatronGeneration.reserve_http_server_address` and belongs next to the other megatron generation tests — that is already true today, independent of the move. `test_http_server_port_reservation` ([test_megatron_generation_parse.py:171](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/tests/unit/models/generation/test_megatron_generation_parse.py#L171)) legitimately spans both sides, since it ends in `MegatronGenerationMixin._setup_openai_api_server`, so it stays put and just imports from `held_port` instead.

nemo_rl/distributed/virtual_cluster.py
```diff
@@ -178,6 +179,73 @@ def _get_node_ip_local() -> str:
...
+            "the port holder on this node is gone, so the pre-published URL would be unreachable."
+        ) from e
+    finally:
+        client.close()
+    if not fds:
+        raise RuntimeError(f"Port holder for port {port} sent no file descriptor.")
+    return socket.socket(fileno=fds[0])
+
+
+class _HeldPortReservation:
```
yuki-97: Consider moving `_held_port_uds_name`, `receive_held_socket`, `_HeldPortReservation` and `PortHolder` into `nemo_rl/distributed/held_port.py`, and dropping the underscore from the class since it becomes that module's main export. `virtual_cluster.py` is 1227 lines and this is its only Ray actor; the fd-handoff protocol has invariants of its own — who closes which fd, exactly one handoff, what happens if the holder dies first — that are easier to see in a file that is only about them.

The tests need to move with it, but not symmetrically. `test_reserve_http_server_address_pins_rank0_bundle` ([test_virtual_cluster.py:677](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/tests/unit/distributed/test_virtual_cluster.py#L677)), plus its `_CapturingPortHolder` fake and `patched_holder` fixture, exercises `MegatronGeneration.reserve_http_server_address` and belongs next to the other megatron generation tests — that is already true today, independent of the move. `test_http_server_port_reservation` ([test_megatron_generation_parse.py:171](https://github.com/NVIDIA-NeMo/RL/blob/9b616b5be278128ab2d4e98851ccecc7361ee15e/tests/unit/models/generation/test_megatron_generation_parse.py#L171)) legitimately spans both sides, since it ends in `MegatronGenerationMixin._setup_openai_api_server`, so it stays put and just imports from `held_port` instead.

nemo_rl/distributed/virtual_cluster.py
```diff
@@ -178,6 +179,73 @@ def _get_node_ip_local() -> str:
...
+            socket.send_fds(conn, [b"s"], [self._sock.fileno()])
+        finally:
+            conn.close()
+            self._uds.close()
+            # The receiver holds a duplicate fd; the local one is done.
+            self._sock.close()
+
+
+# Ray shell around the reservation; pinned to the rank-0 bundle by the caller.
+PortHolder = ray.remote(num_cpus=0)(_HeldPortReservation)  # pragma: no cover
```
yuki-97: wdyt about this? which is the way we always use in the repo.
```suggestion
@ray.remote(num_cpus=0)  # pragma: no cover
class RemoteHeldPortReservation(HeldPortReservation):
    pass
```

yuki-97 (commented):
Nice speedup, and the held-socket handoff is a clean way to get a real URL before the engine exists.

One thing worth writing down in the code: why megatron can't just mirror vLLM's `defer_model_load`. `MegatronPolicyWorker.__init__` builds the model, so there is no `load_model()` to defer to, and the colocated path has no generation-owned worker to bind a port on — a deferred-object mirror would cost splitting that `__init__`, which is also the training worker. The held-socket reservation is the cheap way to get the same "URL before weights" contract, but nothing in the diff says so, and it is the first question a reader has.

Five comments: two bits of dead code introduced by the restructure, a reserve-time metric that renders differently on the SC and grpo drivers, the actor-declaration shape, and a suggested home for the held-port machinery and its tests.

nemo_rl/distributed/virtual_cluster.py
tdene: done

nemo_rl/distributed/virtual_cluster.py
tdene: done

nemo_rl/algorithms/metric_utils.py
tdene: done

nemo_rl/models/generation/megatron/megatron_generation.py
tdene: done

nemo_rl/algorithms/grpo.py
tdene: done

tdene:
> Nice speedup, and the held-socket handoff is a clean way to get a real URL before the engine exists.
> 
> One thing worth writing down in the code: why megatron can't just mirror vLLM's `defer_model_load`. `MegatronPolicyWorker.__init__` builds the model, so there is no `load_model()` to defer to, and the colocated path has no generation-owned worker to bind a port on — a deferred-object mirror would cost splitting that `__init__`, which is also the training worker. The held-socket reservation is the cheap way to get the same "URL before weights" contract, but nothing in the diff says so, and it is the first question a reader has.
> 
> Five comments: two bits of dead code introduced by the restructure, a reserve-time metric that renders differently on the SC and grpo drivers, the actor-declaration shape, and a suggested home for the held-port machinery and its tests.

I raised an issue at #3752 to track this discrepancy, and mentioned it in the code. Thanks for all the other suggested fixes!

nemo_rl/algorithms/metric_utils.py
yuki-97: sorry didn't say clearly last time.

vllm is `vllm_init_time_s = vllm_reserve_time + vllm_load_time` which means the init time in legacy path also contains both reserve + load.
so what I meant is `megatron_generation_init_time_s = reserve_time + megatron_gen_time` (just change `megatron_generation_init_time_s` to include `reserve_time`) and could use `megatron_generation_init_time_s - reserve_time` to get the `load_time` when print the summary in `nemo_rl/algorithms/metric_utils.py`.

it should be great to unify all rollout backends in legacy path to also use `generation_init_time_s`, `generation_init_reserve_time_s`, `generation_init_load_time_s` like SC, but it's out of the scope of this PR, so I don't recommend to do so in this PR.

tdene: Ah, my bad, I understand now

terrykong:
Context from reviewing #3731, no action asked for here — filed as #3777 so it is not lost.

This PR removes the Gym serialization for megatron, which is the big one. What it does not change is that megatron still initializes **policy then generation** one after the other, inside `init_megatron_stack`:

```python
def init_megatron_stack():
    """Init policy then generation; rank 0 holds the reserved port."""
    p, policy_t = init_policy(...)
    pg, gen_t = init_megatron_generation(p, ...)
```

vLLM does not have this — it submits `init_vllm_deferred` and `init_policy` as separate tasks alongside `init_nemo_gym`, so for non-colocated it gets a three-way overlap. Every non-megatron backend gets the same via `initialize_generation_with_policy` (`grpo.py:1217`), which megatron is the only backend not to call.

So after this PR:

The open question in #3777 is whether the serialization is still required. `grpo.py:1276` says it is so "checkpoint conversion completes before inference starts", which is clearly true colocated — but non-colocated passes `policy=None` and `skip_weight_load=True`, and `weights_path` comes from the checkpointer rather than from `init_policy()`, so the dependency may not bind there. That needs someone who knows the conversion path, which is why it is a separate issue rather than a suggestion on this PR.

Flagging it here mainly so it is on your radar while this area is being reworked — if it turns out to be free to fix, this PR is the natural place; if not, #3777 can carry it as a follow-up.

nemo_rl/algorithms/metric_utils.py
tdene: Done; I also did the whole unification, please let me know if you would rather I roll it back.

yuki-97: tysm for helping do the unification! LGTM!

yuki-97 (approved):

PR #3574, opened 2026-08-10
build(container): support non-root runtime execution

ko3n1g:
## Background / motivation

- ARC and sandboxed CI mount an immutable source commit at a custom path and run under a non-root UID.
- The release image currently anchors `uv`, Python, and build caches below `/root`, which is not traversable by that runtime identity.
- OCI security review also requires downloaded build tools to be verified before execution.

## What changed

- Move Python, uv, TRT-LLM wheel, and ccache state to shared `/opt` paths with ownership applied in the layers that create the data.
- Add UID/GID 65532, writable source/venv/cache targets, explicit runtime cache variables, and build-time non-root compatibility probes while retaining root as the default runtime identity.
- Keep `HOME` absent from the final image config so OCI runtimes derive the correct home from `/etc/passwd` after a user override.
- Update CI cache export paths and the `ray.sub` cache override to the relocated uv cache.
- Replace the uv installer script with a pinned release archive and verify per-architecture SHA-256 checksums for uv and CMake.

## Details

- Heavy cache ownership is set during creation to avoid duplicating large layers with a later recursive snapshot.
- Build and cache-export stages continue to run as root; the release image supports `--user 65532:65532` and Kubernetes `runAsUser: 65532` without changing its default root identity.
- Both the baked checkout and the custom-source mount path are checked for non-root writes.
- The final release-stage sync installs the NeMo RL project into `/opt/nemo_rl_venv`.

## Tested

- `docker buildx build --check -f docker/Dockerfile .` — completed with no warnings.
- `bash -n ray.sub` — passed.
- `git diff --check` — passed.
- `pre-commit run --files docker/Dockerfile ray.sub docs/cluster.md` — applicable repository check passed.
- DL/JOC NeMo-CI pipeline [`63363213`](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63363213) — running from `main-mirror` with `RL_REF=f0ad0fd15886eb4b7ff3718963f70f06b4e812e4`; limited to `llm_performance_grpo_llama3_1_8b_instruct_2n8g` in the `performance` suite on `cw_dfw`.

## Pre-checks

- [x] Selected `CI:L0` because the container image changes.
- [ ] Full container/GPU proof is running in DL/JOC pipeline `63363213`.
- [x] Dockerfile comments document the runtime ownership and cache contract.


docker/Dockerfile
```diff
@@ -409,18 +522,125 @@ if [[ -n "${NEMO_GYM_PREFETCH_CONFIGS:-}" ]]; then
...
+for interpreter in sorted(set(interpreters)):
+    subprocess.run(
+        [str(interpreter), "-c", "import sys; assert sys.executable"],
+        check=True,
+    )
+PY
+
+# Root remains the default runtime identity and keeps the conventional home.
+USER root
+ENV HOME=/root
```
thomasdhc: [P1] Findings: runc only derives `HOME` from `/etc/passwd` when `HOME` is absent from the process environment and `cuda-dl-base` sets no `HOME` of its own, so baking `ENV HOME=/root` here makes `--user 65532:65532` and Kubernetes `runAsUser: 65532` land on a root-owned `0700` home instead of `/home/nemo-runtime`, breaking every `$HOME`-relative write this image needs (HF cache, `$HOME/.triton` JIT cache, ray, uv's `~/.local`). Proposal: delete `ENV HOME=/root` and its `test "${HOME}" = /root` assertion so runc resolves `/root` for root and `/home/nemo-runtime` for 65532 from `/etc/passwd`.

docker/Dockerfile
```diff
@@ -94,23 +95,97 @@ rm -rf /var/lib/apt/lists/*
...
+    runtime_user="$(getent passwd "${RUNTIME_UID}" | cut -d: -f1)" && \
+    runuser --user "${runtime_user}" -- \
+        env HOME=/home/nemo-runtime \
+        UV_PYTHON_INSTALL_DIR="${UV_PYTHON_INSTALL_DIR}" \
+        /usr/local/bin/uv python install "${PYTHON_VERSION}"
+
+# Environments baked with uv's symlink mode must not point through /root: a
+# non-root runtime cannot traverse that directory even when the target files
+# themselves are world-readable.
+ENV UV_CACHE_DIR=/opt/nemo_rl_cache/uv
```
thomasdhc, line 190: [P1] Findings: moving the image cache to `/opt/nemo_rl_cache/uv` leaves `ray.sub:118` and `ray.sub:120` mounting `UV_CACHE_DIR_OVERRIDE` at the now-unread `/root/.cache/uv`, so the documented override silently becomes a no-op and venv rebuilds re-download every wheel per node per job instead of reusing the shared cache. Proposal: point the `ray.sub` mount target at `/opt/nemo_rl_cache/uv` and update the stale default documented at `docs/cluster.md:136`.

thomasdhc (changes requested):
Changes requested.

### Review notes
- The build-time non-root probes each export `HOME=/home/nemo-runtime` themselves, so they do not exercise the home directory a plain `--user 65532:65532` invocation actually receives.
- The release probe now ends with `uv sync --frozen --all-groups --no-install-project`, so the shipped `/opt/nemo_rl_venv` no longer carries the project install that the previous `uv run` prefetch left in place.

docker/Dockerfile
ko3n1g: Fixed in f0ad0fd1: ray.sub now mounts UV_CACHE_DIR_OVERRIDE at /opt/nemo_rl_cache/uv, and docs/cluster.md documents that same container default.

docker/Dockerfile
ko3n1g: Fixed in f0ad0fd1: removed both probe-level HOME overrides and the final ENV HOME=/root, and the non-root probe now asserts the /etc/passwd-derived /home/nemo-runtime value.

docker/Dockerfile
```diff
@@ -381,17 +477,34 @@ if [[ -n "${SKIP_TRTLLM_BUILD:-}" ]]; then
...
+install -d -o "${RUNTIME_UID}" -g "${RUNTIME_GID}" /opt/ray_venvs
 if [[ -n "$NEGATIVE_FILTERS" ]]; then
-    UV_LINK_MODE=symlink uv run nemo_rl/utils/prefetch_venvs.py --negative-filters $NEGATIVE_FILTERS
+    UV_LINK_MODE=symlink uv run --locked --extra nemo_gym nemo_rl/utils/prefetch_venvs.py --negative-filters $NEGATIVE_FILTERS
 else
-    UV_LINK_MODE=symlink uv run nemo_rl/utils/prefetch_venvs.py
+    UV_LINK_MODE=symlink uv run --locked --extra nemo_gym nemo_rl/utils/prefetch_venvs.py
 fi
+git config --system --unset-all safe.directory /opt/nemo-rl/3rdparty/Gym-workspace/Gym
+git config --system --unset-all safe.directory /opt/nemo-rl
```
thomasdhc: [P2] Findings: The checkout is transferred to UID 65532 while root remains the image's default runtime identity, and this line drops the last `safe.directory` entry before the layer commits, so a default container run hits Git's dubious-ownership refusal on any repository-discovering call under /opt/nemo-rl, silently stripping the commit suffix from `nemo_rl.__version__` and disabling the source-code and diff artifacts that WandbLogger attaches. Proposal: Retain a persistent system `safe.directory` entry for /opt/nemo-rl and the Gym workspace in the final image instead of unsetting it.

thomasdhc (commented):
Review completed.

### Review notes
- The container fingerprint check is unaffected by the ownership move, because `tools/generate_fingerprint.py` passes an explicit `--git-dir` and `--work-tree` and therefore bypasses Git's ownership validation.
- The pinned uv and CMake archives verify against the upstream published SHA-256 checksums for both x86_64 and aarch64.

docker/Dockerfile
ko3n1g: Fixed in 4835e3f7e. The final image now retains narrowly scoped system `safe.directory` entries for `/opt/nemo-rl` and the Gym workspace, so root runtime Git discovery continues to provide version and W&B source metadata while the checkouts remain owned by UID 65532.

docker/Dockerfile
```diff
@@ -302,15 +389,17 @@ fi
...
+# Restore the upstream all-groups environment in the same layer so the
+# transient TRT-LLM installation does not add a large intermediate venv layer.
+# The preceding dependency layer already warmed the NeMo Gym artifacts.
 uv sync --link-mode symlink --locked --all-groups --no-install-project
 
 # The final sync can repopulate the shared uv cache, so repeat the security
 # cleanup performed in the preceding dependency layer.
-find /root/.cache/uv -type d -path "*ray/_private/runtime_env/agent/thirdparty_files/aiohttp*" -exec rm -rf {} +
+find "${UV_CACHE_DIR}" -type d -path "*ray/_private/runtime_env/agent/thirdparty_files/aiohttp*" -exec rm -rf {} +
+chown -R "${RUNTIME_UID}:${RUNTIME_GID}" /opt/nemo_rl_venv "${UV_CACHE_DIR}"
```
thomasdhc: [P2] Findings: `chown -R` without an ownership filter issues a chown syscall on every entry regardless of its current owner, so this layer and the two release prefetch layers each re-chown the entire 43-54 GB uv cache that the creating layer already owns, costing roughly 160-200 s per layer against 4.6 s for the same tree inside its creating layer and forcing an equivalent overlayfs copy-up on the builder. Proposal: add `--from=0` to the post-creation recursive chowns, or reuse the `find … -uid 0 -exec chown --no-dereference` pattern already applied to `/opt/nemo-rl`, so already-owned files are skipped.

thomasdhc (dismissed):
LGTM, please address below P2 in a follow-up.

### Review notes
- The relocated uv cache, Python installation directory, and prefetched environments no longer resolve through `/root`, and the release-stage probes exercise both the baked checkout and the custom-source mount target as the unprivileged UID.
- Root remains the default runtime identity, and every shipped Git worktree is registered as a system safe directory, so root-run Git commands still work after the source tree moves to the runtime UID.

docker/Dockerfile
```diff
@@ -232,13 +312,19 @@ fi
 uv sync --link-mode symlink --frozen --extra mcore --no-install-project
 uv sync --link-mode symlink --frozen --extra automodel --no-install-project
 uv sync --link-mode symlink --frozen --extra modelopt --no-install-project
+# L0_Unit_Tests_Nemo_Gym.sh selects this extra at runtime. Warm its artifacts
+# into the image cache, then restore the upstream all-groups environment.
+uv sync --link-mode symlink --frozen --all-groups --extra nemo_gym --no-install-project
```
kajalj22: Do we need to add this here? Gym was excluded from the Dockerfile initially intentionally due to CVE impacts

docker/Dockerfile
```diff
@@ -384,9 +485,12 @@ EOF
 # Unshallow the repo to get the full history (in the case it was from the scratch layer).
 # Potentially not necessary if the repo is passed in as a complete repository (w/ full git history),
 # so do a quick check before trying to unshallow.
-RUN git rev-parse --is-shallow-repository | grep -q true && git fetch --unshallow || true
+RUN if git -C /opt/nemo-rl rev-parse --is-shallow-repository | grep -q true; then \
+        git -C /opt/nemo-rl -c fetch.recurseSubmodules=false fetch --unshallow; \
+    fi && \
+    chown -R "${RUNTIME_UID}:${RUNTIME_GID}" /opt/nemo-rl/.git
```
kajalj22: The previous version of this command ended in || true, so a failed git fetch --unshallow was tolerated and the build continued. This rewrite drops that fallback — if the fetch fails (no network, restricted/offline runner), the if block fails, && short-circuits, and the whole RUN now hard-fails the build.

Given this PR's own motivation is supporting sandboxed/restricted-network CI, could this bite exactly that scenario? Was making this fail-loud intentional, or should the || true tolerance be preserved (with the chown moved to run unconditionally)?

kajalj22:
Two good to haves, not blockers:

1. Since docs/docker.md is the repo's main "how to build the image" doc but wasn't touched here: could you add a short section on the new RUNTIME_UID/RUNTIME_GID build args and how to actually exercise non-root execution (e.g. docker run --user 65532:65532 <image> / Kubernetes runAsUser: 65532) if you think that is appropriate here? Right now the image still defaults to root, so without docs there's no discoverable way for a user to learn this capability exists.

2. Nice set of build-time non-root probes. One gap: nothing in cicd-main.yml runs docker run --user 65532:65532 <published image> ... against the actual pulled/tagged artifact post-build — the build-time USER switch is a real permission check via BuildKit, but it doesn't cover the final published image or production-like constraints (e.g. readOnlyRootFilesystem, dropped capabilities). Not a blocker, but could be a good CI follow-up.

docker/Dockerfile
ko3n1g: Fixed in 70be0dc8b. Fail-loud unshallowing was not intentional: `fetch --unshallow` is best-effort again, while the `.git` ownership repair now runs unconditionally. I also simulated a failed fetch and confirmed that `chown` is still reached.

docker/Dockerfile
ko3n1g: Fixed in 9afded236. The post-creation recursive ownership repairs now use `chown --from=0`, including the TRT-LLM skip/final paths and both release prefetch layers, so already-owned cache entries are inspected but do not receive redundant chown syscalls or overlayfs copy-ups.

docker/Dockerfile
ko3n1g: Agreed; fixed in 9afded236. The default build no longer warms the `nemo_gym` extra or verifies it in the offline runtime probe, and the normal Ray environment prefetch no longer selects that extra. The existing `NEMO_GYM_PREFETCH_CONFIGS` path remains the explicit opt-in for images that intentionally prefetch Gym environments.

PR #2514, opened 2026-05-17
fix(nemo-gym): clamp max_new_tokens to prompt + output <= max_model_len

yuki-97:
## Summary

- Revert the hardcoded `max_new_tokens` introduced in #2365.
- Match pre-2365 / native async vLLM behavior: clamp `max_output_tokens` so that `prompt_tokens + max_output_tokens <= max_model_len` (analogous to `allowed_new_tokens` in vLLM's native async path).

## Changes

- `vllm_worker_async.py`: add `_set_max_tokens` / `_clamp_max_tokens` helpers; apply clamping in `_preprocess_chat` for both the normal path and the prefix-replacement path.
- `rollouts.py`: always write `generation_config["max_new_tokens"]` into `responses_create_params["max_output_tokens"]` (remove the `is not None` guard since it's always `not None`).
- Revert config/test changes from #2365: `grpo_nanov3.yaml`, `grpo_workplace_assistant_nemotron_nano_v2_9b.yaml`, `tests/functional/grpo_async_gym.sh`.

## Test plan

- [x] `tests/functional/grpo_async_gym.sh` passes without overriding `max_new_tokens`.
- [x] Behavior is consistent with pre-2365.

cc @binhu-nv

PR #3581, opened 2026-08-11
perf(vllm): avoid repeated multi-turn tokenization

jthomson04:
# What does this PR do ?

Avoid repeated full-history tokenization for supported multi-turn vLLM chat requests. When an earlier assistant message includes exact token metadata, NeMo-RL reuses that model prefix and tokenizes only the contextual suffix for the next turn.

The guarded path keeps vLLM chat parsing, template rendering, tool and reasoning adjustment, validation, and engine preparation. First turns and unsupported requests continue to use native vLLM preprocessing. This PR does not change the public API or require a Gym change.

Text-only requests on multimodal-capable models are supported when every message has plain string content. Requests with OpenAI content parts or parsed image, audio, or video data continue to use native vLLM preprocessing.

## Performance

A matched end-to-end validation used 320 rollouts per arm and approximately 18,000 model calls:

The baseline used commit `0ded4a9f510f55efb154d8005519d2b7e4ed463e`. The optimized arm used commit `56bb18ece647029ec49ad477e2047ac934578cab`.

| Metric | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| Server preprocessing, mean | 0.938 s | 0.038 s | 96.0% |
| Server preprocessing, p50 | 0.700 s | 0.026 s | 96.3% |
| Server preprocessing, p95 | 2.497 s | 0.093 s | 96.3% |
| Observed per-turn response latency, mean | 3.984 s | 3.327 s | 16.5% |
| Observed per-turn response latency, p50 | 2.956 s | 2.047 s | 30.8% |

The response-latency measurements are end-to-end model-call latencies observed by the agent. They include inference and request-processing time.

The validation used a Nemotron Nano v3.5 SWE checkpoint with:

- `NemotronHForCausalLM`
- `PreTrainedTokenizerFast`
- `<|im_end|>` as the EOS token
- `enable_thinking=true`
- vLLM 0.25.1
- chat-template SHA-256 `82753bef5cedc4932c1ed509b5c9a12be680fd86d1adb65bc3f7398d11c8eebc`

# Issues

None.

# Usage

No configuration change is required. NeMo-RL selects the optimization only when complete token metadata is available and all safety guards pass.

**Pre checks**:

- [x] Added focused unit tests for prompt-token construction, guards, fallbacks, and HTTP behavior.
- [x] Ran the focused unit and end-to-end functional validation.
- [x] No documentation update is required because there is no public API or configuration change.

# Additional Information

Validation included:

- `pytest -q tests/unit/models/generation/test_openai_server_utils.py`
- `pytest -q -s tests/unit/models/generation/test_vllm_generation.py::test_vllm_http_server_correct_merged_tokens_matches_baseline`
- Exact native-versus-injected prompt comparison across 3,709 subsequent-turn requests, with zero mismatches.
- Full end-to-end validation with 320 valid rollouts per arm, zero injection failures, and zero unexpected fallbacks.
- Head-level shadow validation at commit `8790130218435a6d350d3c0b089f67660266e333` completed 64 valid rollouts. It compared 3,639 injected prompts with native vLLM preprocessing and found zero mismatches, zero injection failures, zero unexpected fallbacks, zero `/tokenize` calls, and zero HTTP errors.

### DIFF
--- a/nemo_rl/models/generation/openai_server_utils.py
+++ b/nemo_rl/models/generation/openai_server_utils.py
@@ -38,7 +38,15 @@ class CompleteTokenInjectionError(ValueError):
 def find_latest_tokenized_assistant(
     messages: list[Any],
 ) -> tuple[int, Mapping[str, Any]] | None:
-    """Return the latest fully tokenized assistant and its assistant ordinal."""
+    """Return the ordinal and message for the latest tokenized assistant.
+
+    Args:
+        messages: Validated chat messages in conversation order.
+
+    Returns:
+        The assistant ordinal and message when an assistant contains both prompt
+        and generation token IDs, or ``None`` when no complete metadata exists.
+    """
     assistant_ordinal = -1
     latest_tokenized_assistant = None
     for message in messages:
@@ -98,6 +106,28 @@ def build_complete_prompt_token_ids(
     assistant while making its assistant-close sequence the exact splice
     boundary. The close sequence can differ from the tokenizer EOS token, as it
     does for Gemma chat templates.
+
+    Args:
+        tokenizer: Tokenizer used by the serving renderer.
+        conversation: Parsed text-only chat messages.
+        assistant_ordinal: Ordinal of the assistant covered by the model prefix.
+        model_prefix_token_ids: Non-empty
+            ``prompt_token_ids + generation_token_ids`` from that assistant.
+        render_prompt_text: Callback that renders the complete marked
+            conversation as text.
+        render_assistant_close_text: Optional callback that renders the marked
+            conversation through the selected assistant. It must expose an
+            unconditional assistant-close sequence after the marker. Templates
+            that gate the close on ``add_generation_prompt`` fall back to native
+            preprocessing.
+
+    Returns:
+        Exact model-prefix tokens followed by the tokenized contextual suffix.
+
+    Raises:
+        CompleteTokenInjectionError: The splice contract could not be verified.
+            Callers must use native preprocessing instead of rejecting the
+            request.
     """
     try:
         marked_conversation = deepcopy(conversation)
@@ -139,12 +169,17 @@ def build_complete_prompt_token_ids(
     if not isinstance(marked_text, str):
         raise CompleteTokenInjectionError("The marked prompt must render as text.")
 
-    marker_pos = marked_text.find(_COMPLETE_TOKEN_BOUNDARY_MARKER)
-    if marker_pos < 0:
+    marker_count = marked_text.count(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+    if marker_count == 0:
         raise CompleteTokenInjectionError(
             "The chat template did not preserve the complete-token boundary marker."
         )
+    if marker_count > 1:
+        raise CompleteTokenInjectionError(
+            "The complete-token boundary marker collided with rendered request data."
+        )
 
+    marker_pos = marked_text.find(_COMPLETE_TOKEN_BOUNDARY_MARKER)
     marker_end = marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
 
     if render_assistant_close_text is None:
@@ -187,6 +222,9 @@ def build_complete_prompt_token_ids(
             )
         close_marker_end = close_marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
         raw_assistant_close_text = closed_assistant_text[close_marker_end:]
+        # Drop only leading whitespace before the close sequence. The guard below
+        # uses the raw text so the slice removes exactly that same whitespace;
+        # trailing whitespace remains part of the verified close sequence.
         assistant_close_text = raw_assistant_close_text.lstrip()
         if not assistant_close_text:
             raise CompleteTokenInjectionError(
--- a/nemo_rl/models/generation/openai_server_utils.py
+++ b/nemo_rl/models/generation/openai_server_utils.py
@@ -85,6 +85,7 @@ def build_complete_prompt_token_ids(
     assistant_ordinal: int,
     model_prefix_token_ids: list[int],
     render_prompt_text: Callable[[list[Any]], str],
+    render_assistant_close_text: Callable[[list[Any]], str] | None = None,
 ) -> list[int]:
     """Join exact model-prefix tokens with a verified contextual suffix.
 
@@ -94,7 +95,9 @@ def build_complete_prompt_token_ids(
 
     The complete conversation is rendered with the tokenized assistant replaced
     by a marker. This preserves context-sensitive template behavior after that
-    assistant while making its closing EOS the exact splice boundary.
+    assistant while making its assistant-close sequence the exact splice
+    boundary. The close sequence can differ from the tokenizer EOS token, as it
+    does for Gemma chat templates.
     """
     try:
         marked_conversation = deepcopy(conversation)
@@ -142,32 +145,69 @@ def build_complete_prompt_token_ids(
             "The chat template did not preserve the complete-token boundary marker."
         )
 
-    eos_token = getattr(tokenizer, "eos_token", None)
-    eos_token_id = getattr(tokenizer, "eos_token_id", None)
-    if not isinstance(eos_token, str) or eos_token_id is None:
-        raise CompleteTokenInjectionError(
-            "Complete-token injection requires tokenizer EOS text and ID."
-        )
-
-    suffix_pos = marked_text.find(
-        eos_token, marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
-    )
-    if suffix_pos < 0:
-        raise CompleteTokenInjectionError(
-            "The chat template did not close the tokenized assistant with EOS."
-        )
     marker_end = marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
-    # The legacy prefix replacement also cuts at EOS, so dropping whitespace
-    # between the marker and EOS keeps both paths token-identical.
-    if marked_text[marker_end:suffix_pos].strip():
-        raise CompleteTokenInjectionError(
-            "The chat template did not close the tokenized assistant immediately; "
-            "the EOS boundary would skip intervening messages."
-        )
+
+    if render_assistant_close_text is None:
+        eos_token = getattr(tokenizer, "eos_token", None)
+        if not isinstance(eos_token, str):
+            raise CompleteTokenInjectionError(
+                "Complete-token injection requires tokenizer EOS text."
+            )
+        suffix_pos = marked_text.find(eos_token, marker_end)
+        if suffix_pos < 0:
+            raise CompleteTokenInjectionError(
+                "The chat template did not close the tokenized assistant with EOS."
+            )
+        if marked_text[marker_end:suffix_pos].strip():
+            raise CompleteTokenInjectionError(
+                "The chat template did not close the tokenized assistant immediately; "
+                "the EOS boundary would skip intervening messages."
+            )
+        assistant_close_text = eos_token
+        contextual_suffix_text = marked_text[suffix_pos:]
+    else:
+        try:
+            closed_assistant_text = render_assistant_close_text(
+                marked_conversation[: assistant_index + 1]
+            )
+        except CompleteTokenInjectionError:
+            raise
+        except Exception as e:
+            raise CompleteTokenInjectionError(
+                "The assistant-close probe could not be rendered."
+            ) from e
+        if not isinstance(closed_assistant_text, str):
+            raise CompleteTokenInjectionError(
+                "The assistant-close probe must render as text."
+            )
+        close_marker_pos = closed_assistant_text.find(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+        if close_marker_pos < 0:
+            raise CompleteTokenInjectionError(
+                "The assistant-close probe did not preserve the boundary marker."
+            )
+        close_marker_end = close_marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+        raw_assistant_close_text = closed_assistant_text[close_marker_end:]
+        assistant_close_text = raw_assistant_close_text.lstrip()
+        if not assistant_close_text:
+            raise CompleteTokenInjectionError(
+                "The chat template did not emit an assistant-close sequence."
+            )
+        contextual_suffix_text = marked_text[marker_end:]
+        if not contextual_suffix_text.startswith(raw_assistant_close_text):
+            raise CompleteTokenInjectionError(
+                "The assistant-close sequence changed in the complete conversation."
+            )
+        contextual_suffix_text = contextual_suffix_text[
+            len(raw_assistant_close_text) - len(assistant_close_text) :
+        ]
 
     try:
+        assistant_close_token_ids = _coerce_token_id_list(
+            tokenizer.encode(assistant_close_text, add_special_tokens=False),
+            "Assistant-close token IDs",
+        )
         suffix_token_ids = _coerce_token_id_list(
-            tokenizer.encode(marked_text[suffix_pos:], add_special_tokens=False),
+            tokenizer.encode(contextual_suffix_text, add_special_tokens=False),
             "Contextual suffix token IDs",
         )
     except CompleteTokenInjectionError:
@@ -177,16 +217,30 @@ def build_complete_prompt_token_ids(
             "The contextual suffix could not be tokenized."
         ) from e
 
-    if not suffix_token_ids or suffix_token_ids[0] != eos_token_id:
+    if (
+        not assistant_close_token_ids
+        or suffix_token_ids[: len(assistant_close_token_ids)]
+        != assistant_close_token_ids
+    ):
         raise CompleteTokenInjectionError(
-            "The contextual suffix did not begin with EOS."
+            "The contextual suffix did not begin with the assistant-close sequence."
         )
+    model_prefix_token_ids = _coerce_token_id_list(
+        model_prefix_token_ids, "Model prefix token IDs"
+    )
     if not model_prefix_token_ids:
         raise CompleteTokenInjectionError("The model prefix token IDs were empty.")
-    model_cut_end = len(model_prefix_token_ids)
-    if model_prefix_token_ids[-1] == eos_token_id:
-        model_cut_end -= 1
-    return model_prefix_token_ids[:model_cut_end] + suffix_token_ids
+
+    max_overlap = min(len(model_prefix_token_ids), len(assistant_close_token_ids))
+    overlap = next(
+        (
+            overlap
+            for overlap in range(max_overlap, 0, -1)
+            if model_prefix_token_ids[-overlap:] == suffix_token_ids[:overlap]
+        ),
+        0,
+    )
+    return model_prefix_token_ids + suffix_token_ids[overlap:]
 
 
 def replace_prefix_tokens(
--- a/nemo_rl/models/generation/openai_server_utils.py
+++ b/nemo_rl/models/generation/openai_server_utils.py
@@ -24,6 +24,7 @@ retokenization drift to correct.
 """
 
 from collections.abc import Callable, Mapping
+from copy import Error as CopyError
 from copy import deepcopy
 from typing import Any
 
@@ -70,12 +71,11 @@ def _assistant_index_from_ordinal(
 def _coerce_token_id_list(value: Any, field_name: str) -> list[int]:
     if not isinstance(value, list):
         raise CompleteTokenInjectionError(f"{field_name} must be a list.")
-    try:
-        return [int(token_id) for token_id in value]
-    except (TypeError, ValueError) as e:
+    if any(type(token_id) is not int for token_id in value):
         raise CompleteTokenInjectionError(
             f"{field_name} must contain only integer token IDs."
-        ) from e
+        )
+    return list(value)
 
 
 def build_complete_prompt_token_ids(
@@ -96,7 +96,12 @@ def build_complete_prompt_token_ids(
     by a marker. This preserves context-sensitive template behavior after that
     assistant while making its closing EOS the exact splice boundary.
     """
-    marked_conversation = deepcopy(conversation)
+    try:
+        marked_conversation = deepcopy(conversation)
+    except (CopyError, TypeError) as e:
+        raise CompleteTokenInjectionError(
+            "The conversation could not be copied for complete-token injection."
+        ) from e
     if any(
         _COMPLETE_TOKEN_BOUNDARY_MARKER in str(message)
         for message in marked_conversation
@@ -152,6 +157,8 @@ def build_complete_prompt_token_ids(
             "The chat template did not close the tokenized assistant with EOS."
         )
     marker_end = marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+    # The legacy prefix replacement also cuts at EOS, so dropping whitespace
+    # between the marker and EOS keeps both paths token-identical.
     if marked_text[marker_end:suffix_pos].strip():
         raise CompleteTokenInjectionError(
             "The chat template did not close the tokenized assistant immediately; "
@@ -174,8 +181,10 @@ def build_complete_prompt_token_ids(
         raise CompleteTokenInjectionError(
             "The contextual suffix did not begin with EOS."
         )
+    if not model_prefix_token_ids:
+        raise CompleteTokenInjectionError("The model prefix token IDs were empty.")
     model_cut_end = len(model_prefix_token_ids)
-    if model_prefix_token_ids and model_prefix_token_ids[-1] == eos_token_id:
+    if model_prefix_token_ids[-1] == eos_token_id:
         model_cut_end -= 1
     return model_prefix_token_ids[:model_cut_end] + suffix_token_ids
 
--- a/nemo_rl/models/generation/openai_server_utils.py
+++ b/nemo_rl/models/generation/openai_server_utils.py
@@ -13,32 +13,33 @@
 # limitations under the License.
 """Shared helpers for the OpenAI-compatible HTTP generation servers.
 
-These utilities are backend-agnostic: they operate on token-ID lists plus a
-tokenizer, with no engine calls. They are shared by the vLLM async worker
-(``vllm_worker_async.py``) and the TRT-LLM HTTP server (``trtllm_http_server.py``),
-which both put a message-based ``/v1/chat/completions`` layer in front of a token
-engine for the agentic NeMo-Gym path. SGLang does not use these — it is driven
-token-in/token-out via ``generate(input_ids)`` and never re-templates messages,
-so it has no retokenization drift to correct.
+These backend-agnostic utilities operate on token-ID lists, parsed chat messages,
+and caller-supplied tokenizers or render functions. They make no engine calls.
+They are shared by the vLLM async worker (``vllm_worker_async.py``) and the
+TRT-LLM HTTP server (``trtllm_http_server.py``), which both put a message-based
+``/v1/chat/completions`` layer in front of a token engine for the agentic
+NeMo-Gym path. SGLang does not use these — it is driven token-in/token-out via
+``generate(input_ids)`` and never re-templates messages, so it has no
+retokenization drift to correct.
 """
 
-import json
 from collections.abc import Callable, Mapping
 from copy import deepcopy
 from typing import Any
-
 
 _COMPLETE_TOKEN_BOUNDARY_MARKER = "NEMO_RL_PREFIX_BOUNDARY_7F3A9C1E"
 
 
 class CompleteTokenInjectionError(ValueError):
-    """Raised when complete prompt tokens cannot be derived safely."""
+    """Signal that complete-token injection must use native preprocessing."""
 
 
-def find_latest_tokenized_assistant_ordinal(messages: list[Any]) -> int | None:
-    """Return the assistant ordinal carrying the latest Gym token metadata."""
+def find_latest_tokenized_assistant(
+    messages: list[Any],
+) -> tuple[int, Mapping[str, Any]] | None:
+    """Return the latest fully tokenized assistant and its assistant ordinal."""
     assistant_ordinal = -1
-    latest_tokenized_ordinal = None
+    latest_tokenized_assistant = None
     for message in messages:
         if not isinstance(message, Mapping) or message.get("role") != "assistant":
             continue
@@ -47,8 +48,8 @@ def find_latest_tokenized_assistant_ordinal(messages: list[Any]) -> int | None:
             message.get("prompt_token_ids") is not None
             and message.get("generation_token_ids") is not None
         ):
-            latest_tokenized_ordinal = assistant_ordinal
-    return latest_tokenized_ordinal
+            latest_tokenized_assistant = (assistant_ordinal, message)
+    return latest_tokenized_assistant
 
 
 def _assistant_index_from_ordinal(
@@ -64,33 +65,6 @@ def _assistant_index_from_ordinal(
     raise CompleteTokenInjectionError(
         "The tokenized assistant message was not present after chat parsing."
     )
-
-
-def _normalize_tool_arguments_for_template(
-    messages: list[Any], *, before_index: int
-) -> None:
-    for message in messages[:before_index]:
-        if not isinstance(message, dict) or message.get("role") != "assistant":
-            continue
-        tool_calls = message.get("tool_calls")
-        if not isinstance(tool_calls, list):
-            continue
-        for tool_call in tool_calls:
-            if not isinstance(tool_call, dict):
-                continue
-            function = tool_call.get("function", tool_call)
-            if not isinstance(function, dict):
-                continue
-            arguments = function.get("arguments")
-            if not isinstance(arguments, str):
-                continue
-            try:
-                parsed_arguments = json.loads(arguments)
-            except json.JSONDecodeError:
-                parsed_arguments = {}
-            function["arguments"] = (
-                parsed_arguments if isinstance(parsed_arguments, dict) else {}
-            )
 
 
 def _coerce_token_id_list(value: Any, field_name: str) -> list[int]:
@@ -114,6 +88,10 @@ def build_complete_prompt_token_ids(
 ) -> list[int]:
     """Join exact model-prefix tokens with a verified contextual suffix.
 
+    ``assistant_ordinal`` must identify the assistant turn covered by
+    ``model_prefix_token_ids``. A mismatch can produce a valid but incorrect
+    prompt, so callers must derive both values from the same message.
+
     The complete conversation is rendered with the tokenized assistant replaced
     by a marker. This preserves context-sensitive template behavior after that
     assistant while making its closing EOS the exact splice boundary.
@@ -129,9 +107,6 @@ def build_complete_prompt_token_ids(
 
     assistant_index = _assistant_index_from_ordinal(
         marked_conversation, assistant_ordinal
-    )
-    _normalize_tool_arguments_for_template(
-        marked_conversation, before_index=assistant_index
     )
 
     marked_assistant = marked_conversation[assistant_index]
@@ -175,6 +150,12 @@ def build_complete_prompt_token_ids(
     if suffix_pos < 0:
         raise CompleteTokenInjectionError(
             "The chat template did not close the tokenized assistant with EOS."
+        )
+    marker_end = marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+    if marked_text[marker_end:suffix_pos].strip():
+        raise CompleteTokenInjectionError(
+            "The chat template did not close the tokenized assistant immediately; "
+            "the EOS boundary would skip intervening messages."
         )
 
     try:
--- a/nemo_rl/models/generation/openai_server_utils.py
+++ b/nemo_rl/models/generation/openai_server_utils.py
@@ -22,7 +22,181 @@ token-in/token-out via ``generate(input_ids)`` and never re-templates messages,
 so it has no retokenization drift to correct.
 """
 
+import json
+from collections.abc import Callable, Mapping
+from copy import deepcopy
 from typing import Any
+
+
+_COMPLETE_TOKEN_BOUNDARY_MARKER = "NEMO_RL_PREFIX_BOUNDARY_7F3A9C1E"
+
+
+class CompleteTokenInjectionError(ValueError):
+    """Raised when complete prompt tokens cannot be derived safely."""
+
+
+def find_latest_tokenized_assistant_ordinal(messages: list[Any]) -> int | None:
+    """Return the assistant ordinal carrying the latest Gym token metadata."""
+    assistant_ordinal = -1
+    latest_tokenized_ordinal = None
+    for message in messages:
+        if not isinstance(message, Mapping) or message.get("role") != "assistant":
+            continue
+        assistant_ordinal += 1
+        if (
+            message.get("prompt_token_ids") is not None
+            and message.get("generation_token_ids") is not None
+        ):
+            latest_tokenized_ordinal = assistant_ordinal
+    return latest_tokenized_ordinal
+
+
+def _assistant_index_from_ordinal(
+    conversation: list[Any], assistant_ordinal: int
+) -> int:
+    current_ordinal = -1
+    for index, message in enumerate(conversation):
+        if not isinstance(message, Mapping) or message.get("role") != "assistant":
+            continue
+        current_ordinal += 1
+        if current_ordinal == assistant_ordinal:
+            return index
+    raise CompleteTokenInjectionError(
+        "The tokenized assistant message was not present after chat parsing."
+    )
+
+
+def _normalize_tool_arguments_for_template(
+    messages: list[Any], *, before_index: int
+) -> None:
+    for message in messages[:before_index]:
+        if not isinstance(message, dict) or message.get("role") != "assistant":
+            continue
+        tool_calls = message.get("tool_calls")
+        if not isinstance(tool_calls, list):
+            continue
+        for tool_call in tool_calls:
+            if not isinstance(tool_call, dict):
+                continue
+            function = tool_call.get("function", tool_call)
+            if not isinstance(function, dict):
+                continue
+            arguments = function.get("arguments")
+            if not isinstance(arguments, str):
+                continue
+            try:
+                parsed_arguments = json.loads(arguments)
+            except json.JSONDecodeError:
+                parsed_arguments = {}
+            function["arguments"] = (
+                parsed_arguments if isinstance(parsed_arguments, dict) else {}
+            )
+
+
+def _coerce_token_id_list(value: Any, field_name: str) -> list[int]:
+    if not isinstance(value, list):
+        raise CompleteTokenInjectionError(f"{field_name} must be a list.")
+    try:
+        return [int(token_id) for token_id in value]
+    except (TypeError, ValueError) as e:
+        raise CompleteTokenInjectionError(
+            f"{field_name} must contain only integer token IDs."
+        ) from e
+
+
+def build_complete_prompt_token_ids(
+    *,
+    tokenizer: Any,
+    conversation: list[Any],
+    assistant_ordinal: int,
+    model_prefix_token_ids: list[int],
+    render_prompt_text: Callable[[list[Any]], str],
+) -> list[int]:
+    """Join exact model-prefix tokens with a verified contextual suffix.
+
+    The complete conversation is rendered with the tokenized assistant replaced
+    by a marker. This preserves context-sensitive template behavior after that
+    assistant while making its closing EOS the exact splice boundary.
+    """
+    marked_conversation = deepcopy(conversation)
+    if any(
+        _COMPLETE_TOKEN_BOUNDARY_MARKER in str(message)
+        for message in marked_conversation
+    ):
+        raise CompleteTokenInjectionError(
+            "The complete-token boundary marker collided with request data."
+        )
+
+    assistant_index = _assistant_index_from_ordinal(
+        marked_conversation, assistant_ordinal
+    )
+    _normalize_tool_arguments_for_template(
+        marked_conversation, before_index=assistant_index
+    )
+
+    marked_assistant = marked_conversation[assistant_index]
+    if not isinstance(marked_assistant, dict):
+        raise CompleteTokenInjectionError(
+            "The tokenized assistant message was not a message object."
+        )
+    marked_assistant["content"] = _COMPLETE_TOKEN_BOUNDARY_MARKER
+    marked_assistant.pop("reasoning_content", None)
+    marked_assistant.pop("reasoning", None)
+    marked_assistant.pop("tool_calls", None)
+
+    try:
+        marked_text = render_prompt_text(marked_conversation)
+    except CompleteTokenInjectionError:
+        raise
+    except Exception as e:
+        raise CompleteTokenInjectionError(
+            "The marked chat template could not be rendered."
+        ) from e
+
+    if not isinstance(marked_text, str):
+        raise CompleteTokenInjectionError("The marked prompt must render as text.")
+
+    marker_pos = marked_text.find(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+    if marker_pos < 0:
+        raise CompleteTokenInjectionError(
+            "The chat template did not preserve the complete-token boundary marker."
+        )
+
+    eos_token = getattr(tokenizer, "eos_token", None)
+    eos_token_id = getattr(tokenizer, "eos_token_id", None)
+    if not isinstance(eos_token, str) or eos_token_id is None:
+        raise CompleteTokenInjectionError(
+            "Complete-token injection requires tokenizer EOS text and ID."
+        )
+
+    suffix_pos = marked_text.find(
+        eos_token, marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+    )
+    if suffix_pos < 0:
+        raise CompleteTokenInjectionError(
+            "The chat template did not close the tokenized assistant with EOS."
+        )
+
+    try:
+        suffix_token_ids = _coerce_token_id_list(
+            tokenizer.encode(marked_text[suffix_pos:], add_special_tokens=False),
+            "Contextual suffix token IDs",
+        )
+    except CompleteTokenInjectionError:
+        raise
+    except Exception as e:
+        raise CompleteTokenInjectionError(
+            "The contextual suffix could not be tokenized."
+        ) from e
+
+    if not suffix_token_ids or suffix_token_ids[0] != eos_token_id:
+        raise CompleteTokenInjectionError(
+            "The contextual suffix did not begin with EOS."
+        )
+    model_cut_end = len(model_prefix_token_ids)
+    if model_prefix_token_ids and model_prefix_token_ids[-1] == eos_token_id:
+        model_cut_end -= 1
+    return model_prefix_token_ids[:model_cut_end] + suffix_token_ids
 
 
 def replace_prefix_tokens(
--- a/nemo_rl/models/generation/vllm/vllm_worker_async.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker_async.py
@@ -61,6 +61,7 @@ from nemo_rl.models.generation.vllm.utils import (
 from nemo_rl.models.generation.vllm.vllm_worker import BaseVllmGenerationWorker
 
 LOGGER = logging.getLogger(__name__)
+COMPLETE_TOKEN_INJECTION_VLLM_VERSION = "0.25.1"
 
 
 def _prefix_matches_tokenized_assistant(
@@ -101,7 +102,7 @@ def _complete_token_injection_eligibility(
     has_multimodal_config: bool,
     prompt_embeds_enabled: bool,
     renderer_supported: bool,
-    vllm_version_supported: bool = True,
+    vllm_version_supported: bool,
 ) -> tuple[int | None, str | None]:
     """Return an assistant ordinal or a reason to use native preprocessing."""
     if request.stream:
@@ -450,7 +451,7 @@ class VllmAsyncGenerationWorkerImpl(
         from copy import deepcopy
         from logging import Filter as LoggingFilter
         from logging import LogRecord
-        from typing import List, Optional, Union
+        from typing import Any, List, Optional, Union
 
         from fastapi import Request
         from fastapi.responses import JSONResponse, StreamingResponse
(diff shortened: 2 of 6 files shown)


jthomson04 (commented):
Reviewed with a team of agents (RL codebase, vLLM API, bug scan, tests, plus an adversarial pass), each verifying claims against the pinned vLLM **v0.25.1** source rather than the local checkout. Nice result on the numbers — the 96% preprocessing reduction with an exact native-vs-injected comparison over 3,709 requests is the right correctness experiment, and the fail-safe posture is well built: nearly every hazard probed routes to `CompleteTokenInjectionError` and native preprocessing.

**Verdict: this is worth landing.** The win is measured, large, and on the right axis, and I went looking for a simpler design without finding one — rendering the truncated conversation to locate the boundary is what the fallback already does (two full renders, i.e. the cost being eliminated), and a prefix cache would need a conversation key Gym doesn't send. The marker trick buys one render plus a suffix-only encode while preserving context-sensitive template behavior after the spliced turn.

Nothing below is a blocker. The three I'd most want addressed are small and concrete: the missing empty-prefix guard, the untested load-bearing `deepcopy`, and exposing the injection counters. The rest is hardening, tests, and docs.

An adversarial pass **downgraded several of my own initial findings**, and I've rewritten those comments rather than leave them overstated:

- The **empty-prefix** and **prefix/ordinal** items are defensive hardening, not live bugs — I searched for a producer of empty token lists and found none, and the derived-prefix path is provably safe because upstream passes `request.messages` through untouched.
- My suggestion to wrap the fast path in a broad `except Exception` was **wrong** and is retracted in-thread; it conflicts with this repo's `error-handling` guidance and would make an upstream break undetectable. Two targeted fixes replace it.
- A **whitespace-splice** concern I raised was refuted: the behavior is byte-identical to `replace_prefix_tokens`, so it's consistent by design. Only a test rename and a clarifying comment remain.

Items recorded as **verified clean**, so they don't get re-litigated:

- The seven `getattr(request, X, False)` gates are all correct for v0.25.1. `return_prompt_text` is the only true tri-state (`bool | None = None`), but its only two consumers use plain truthiness with no server-level default, so `None` genuinely means off.
- `isinstance(self.renderer, HfRenderer)` is exact — v0.25.1 has zero `HfRenderer` subclasses, so `MistralRenderer` and friends are excluded and the hardcoded `tokenize=False` matches native. `kimi_audio` also maps to `HfRenderer` but is excluded by the multimodal gate.
- The `mm_data`/`mm_uuids` guard does not fire on text-only requests (`resolve_items()` returns `(None, None)`), so the optimization is live rather than dead code.
- The `assistant_ordinal` <-> parsed-conversation mapping holds: `conversation.extend(sub_messages)` looks like it could fan out, but `_parse_chat_message_content_parts` returns a single-element list on both branches, and `_postprocess_messages` never adds or removes messages.
- Max-model-len validation is **not** bypassed — `apply_post_tokenization` still runs `_token_len_check` for token-id prompts.

Two open questions rather than defects: whether the fast path engages for Gym's Responses-API agents given their list-shaped content, and the fast-vs-fallback boundary divergence. On the latter — the fallback selects the *last* assistant while the fast path selects the latest *tokenized* one, so for a conversation ending in an untokenized assistant they produce different prompts (the fallback drops an intervening user turn). That is **pre-existing and out of scope here**, and the new path is the correct one; I mention it only because it means a silent fallback can change the answer, which is why the unread counters matter.

Finally, the `L0_Unit_Tests_Automodel` failure is `test_value_worker_train_decreases_loss[2gpu_dp2]` — value-worker/DTensor, unrelated to this change.

_Generated by Claude Code_

nemo_rl/models/generation/openai_server_utils.py
```diff
@@ -13,17 +13,172 @@
...
+        ) from e
+
+    if not suffix_token_ids or suffix_token_ids[0] != eos_token_id:
+        raise CompleteTokenInjectionError(
+            "The contextual suffix did not begin with EOS."
+        )
+    model_cut_end = len(model_prefix_token_ids)
+    if model_prefix_token_ids and model_prefix_token_ids[-1] == eos_token_id:
+        model_cut_end -= 1
+    return model_prefix_token_ids[:model_cut_end] + suffix_token_ids
```
jthomson04: [`openai_server_utils.py:177-180`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L177-L180)

**Defensive hardening, not a live bug** — I originally flagged this harder than the evidence supports, so scoping it properly.

The sibling `replace_prefix_tokens` opens with [`if not model_prefix_token_ids: return template_token_ids`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L247-L248); this function has no equivalent, so an empty prefix returns the suffix alone. Executed against this branch:

```
build_complete_prompt_token_ids(prefix=[]) -> [2, 40, 99]        # bare EOS + tail
replace_prefix_tokens(prefix=[])           -> [31, 32, 2, 40, 99] # full conversation
```

On reachability: I went looking for a producer of empty token lists and **found none**. Gym gates on truthiness (`if self.return_token_id_information and m.get("prompt_token_ids")`), and on the chat-native side `prompt_token_ids` comes from `/tokenize`, which is never empty for a non-empty conversation. The remaining route is a client explicitly sending `required_prefix_token_ids: []`, which `[] is not None` admits at [:673-676](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L673-L676) — reachable at the API surface, but nothing does it today.

Still worth the two lines, since it's a behavior *regression* against the legacy path (which treats `[]` as "no replacement" and returns the correct prompt) and the failure is silent:

```suggestion
    if not model_prefix_token_ids:
        raise CompleteTokenInjectionError("The model prefix token IDs were empty.")
    model_cut_end = len(model_prefix_token_ids)
    if model_prefix_token_ids and model_prefix_token_ids[-1] == eos_token_id:
        model_cut_end -= 1
    return model_prefix_token_ids[:model_cut_end] + suffix_token_ids
```

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -483,6 +670,54 @@ async def preprocess_chat(
                     if actual_request_max_tokens is not None:
                         self._set_max_tokens(request, 1)
 
+                if (
+                    isinstance(request, NeMoRLChatCompletionRequest)
+                    and request.required_prefix_token_ids is not None
+                ):
+                    self._complete_token_injection_attempts += 1
+                    assistant_ordinal, fallback_reason = (
```
jthomson04, line 804: [`vllm_worker_async.py:677-688`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L677-L688)

**Latent contract hazard rather than a live bug** — scoping this more carefully than I first did, including a result that argues *for* the current code.

The **derived** case is provably safe, which is worth stating explicitly: upstream passes the request's own list through untouched — [`await self.preprocess_chat(request, request.messages, ...)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/online_renderer.py#L159-L167) — so [`model_post_init`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L468-L476) and the recomputation at [:103](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L103) always see the identical list and always agree on the ordinal. The only intervening mutation is [`message["tool_calls"] = list(...)`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L655-L657), which can't change roles or ordinals.

The gap is the **client-supplied** case: `required_prefix_token_ids` is a settable field on [`NeMoRLChatCompletionRequest`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L815-L818), and `model_post_init` only derives it when omitted — so a request supplying both an override prefix *and* token-annotated messages gets a prefix spliced at an independently-derived boundary. Nothing in NeMo-RL, Gym, or the tests does this, so it's not reachable today; the docstring's ["callers must derive both values from the same message"](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L91-L93) is currently the only enforcement.

A one-line guard would close the contract — either ignore a client-supplied prefix on the fast path, or fall back when it disagrees:

```python
derived = list(message["prompt_token_ids"]) + list(message["generation_token_ids"])
if list(request.required_prefix_token_ids) != derived:
    return None, "client-supplied prefix does not match message token metadata"
```

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -483,6 +670,54 @@ async def preprocess_chat(
...
+                                default_template_content_format=(
+                                    default_template_content_format
+                                ),
+                                default_template_kwargs=default_template_kwargs,
+                                tool_dicts=tool_dicts,
+                                parser=parser,
+                                assistant_ordinal=assistant_ordinal,
+                                skip_mm_cache=skip_mm_cache,
+                            )
+                        except CompleteTokenInjectionError as e:
```
jthomson04, line 840: [`vllm_worker_async.py:707-708`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L707-L708)

**Correcting my earlier framing here.** I originally suggested a defensive `except Exception`; that was the wrong ask and I'm retracting it. It runs against this repo's `error-handling` guidance ("limit the except to the smallest set of errors possible"), and because the injection counters are unread, a broad catch would convert a genuine upstream break into a permanent, invisible loss of the whole optimization. The failure you'd be swallowing is exactly the one you want to be told about.

The exposure is also narrower than a blanket catch implies: vLLM is pinned exactly (`vllm==0.25.1`), so a rename can only arrive with a reviewed bump, and the module-level imports in `_setup_vllm_openai_api_server` would fail loudly at server startup rather than per-request.

That leaves one genuinely new runtime coupling, and it's worth being precise about why it's different from the other calls here. [`renderer._apply_prompt_extras`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L601) is private, and native reaches it from *inside* `base.py` — so an upstream rename renames the internal call site too and native keeps working, while NeMo-RL's external call is the only thing that breaks. Every other call in this method fails identically under native preprocessing; this one doesn't. Best fixed by inlining it (see that comment); a startup-time `hasattr` assertion would also work and matches the skill's "prefer failing at `setup()` time over a deep-in-the-loop crash".

Minor, same principle: the eligibility gates use `getattr(request, X, <default>)`, which fails **open** — an upstream rename silently returns the default and the gate stops firing, admitting requests that should have fallen back. Direct attribute access would make that drift a loud `AttributeError` instead.

nemo_rl/models/generation/openai_server_utils.py
```diff
@@ -13,17 +13,172 @@
...
+
+    ``assistant_ordinal`` must identify the assistant turn covered by
+    ``model_prefix_token_ids``. A mismatch can produce a valid but incorrect
+    prompt, so callers must derive both values from the same message.
+
+    The complete conversation is rendered with the tokenized assistant replaced
+    by a marker. This preserves context-sensitive template behavior after that
+    assistant while making its closing EOS the exact splice boundary.
+    """
+    marked_conversation = deepcopy(conversation)
```
jthomson04: [`openai_server_utils.py:99`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L99)

Minor robustness note, not a live defect. `deepcopy` here sits outside the `try` blocks that wrap [rendering](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L122-L129) and [encoding](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L161-L171), so a non-deepcopyable message would raise a raw `TypeError` rather than `CompleteTokenInjectionError` and escape the caller's [fallback handler](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L707).

I should be straight that the only reproduction I have is synthetic (a `MappingProxyType` message → `cannot pickle 'mappingproxy' object`). vLLM's real pipeline can't produce that input: `conversation` is built fresh by `parse_chat_messages_async` from JSON-parsed request data as [`ConversationMessage(role=..., content=...)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/chat_utils.py#L1591-L1606) with `str` or plain-dict content, so there's no live 500 here.

Raising it only because it's a one-line inconsistency with the fallback contract the rest of this function maintains — wrapping it in the same pattern used at :122-129 would make the guarantee uniform.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -447,6 +504,138 @@ def _clamp_max_tokens(
...
+                        tokenize=False,
+                        **apply_chat_template_kwargs,
+                    )
+                    if not isinstance(rendered, str):
+                        raise CompleteTokenInjectionError(
+                            "The marked prompt did not render as text."
+                        )
+                    return rendered
+
+                prompt_token_ids = await asyncio.to_thread(
```
jthomson04, line 713: [`vllm_worker_async.py:589-596`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L589-L596)

**Worth an acknowledgement and a comment rather than a rewrite** — flagging so the trade-off is deliberate rather than incidental.

vLLM runs chat-template rendering on [`ThreadPoolExecutor(max_workers=renderer_num_workers)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/base.py#L86-L87) with [`renderer_num_workers: int = 1`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/config/model.py#L333), and sizes the tokenizer pool to match at [`renderer_num_workers + 1`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/hf.py#L928-L929). `asyncio.to_thread` uses the loop's default executor instead (CPython `min(32, cpu_count + 4)`), so rendering moves off a deliberately 1-wide executor onto a ~32-wide one — which guarantees the size-2 pool is exhausted. The overflow path then [deep-copies a fast tokenizer and retains the copy](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/tokenizers/hf.py#L44-L57):
```python
except queue.Empty:
    tok = copy.deepcopy(og_tokenizer)
finally:
    tokenizer_pool.put(tok)          # retained; the queue has no maxsize
```
Both calls route through it — `tokenizer.__class__ = TokenizerPool` is an [in-place class swap](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/tokenizers/hf.py#L101), so `encode` and `safe_apply_chat_template`'s internal `apply_chat_template` both borrow.

To be precise about size: each request holds at most one tokenizer at a time, so this is **bounded** — a one-time step from 2 to at most the executor width, then stable. Not a leak.

The fair counter, which is why I'm not asking for a rewrite: routing through `renderer._executor` would serialize the fast path at width 1 and might well erase the measured win, so `to_thread` may be a deliberate choice. If so, a comment saying that would be ideal. Worth knowing either way that `apply_chat_template` is GIL-bound Jinja, so the extra width buys less throughput than it looks like while adding event-loop contention — which is the thing `renderer_num_workers` exists to bound.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -447,6 +504,138 @@ def _clamp_max_tokens(
...
+                    conversation=conversation,
+                    assistant_ordinal=assistant_ordinal,
+                    model_prefix_token_ids=list(request.required_prefix_token_ids),
+                    render_prompt_text=render_prompt_text,
+                )
+                prompt = {"prompt_token_ids": prompt_token_ids}
+                tokenized_prompt = await renderer.tokenize_prompt_async(
+                    prompt, tok_params
+                )
+                renderer._apply_prompt_extras(
```
jthomson04: [`vllm_worker_async.py:601-608`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L601-L608)

`_apply_prompt_extras` is a private vLLM method, and `vllm/renderers/` was restructured in 0.25 (as your comment at [:639-642](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L639-L642) notes) — so this is the most version-fragile line in the change, and the one place that fails silently at request time rather than loudly at startup.

It can be dropped entirely with no behavior change. [The method](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/base.py#L650-L660) calls `extract_target_prompt`, which for a decoder-only model [reduces to `parse_dec_only_prompt`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/inputs/preprocess.py#L248-L253), which [returns the dict unchanged](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/inputs/preprocess.py#L145-L156). Since the prompt dict is built right here, the call is exactly `tokenized_prompt.update(extras)`:

```suggestion
                for key in ("mm_processor_kwargs", "cache_salt"):
                    if (value := getattr(request, key, None)) is not None:
                        tokenized_prompt[key] = value
```

Related, and the main structural risk in the PR: `_preprocess_chat_with_complete_token_ids` mirrors [`OnlineRenderer.preprocess_chat`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/online_renderer.py#L330-L417) plus parts of `render_chat_async`/`render_messages_async`, and the liability is drift on the next bump — which this file has already lived through once. A pointer comment on the method naming the upstream function and version it was derived from would give the next bump an anchor to re-diff against.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -51,13 +58,55 @@
...
+    if has_multimodal_config:
+        return None, "multimodal model"
+    if prompt_embeds_enabled:
+        return None, "prompt embeddings enabled"
+    if not renderer_supported:
+        return None, "unsupported renderer"
+    for message in messages:
+        if not isinstance(message, dict):
+            return None, "non-object message"
+        if message.get("role") not in {"system", "user", "assistant", "tool"}:
```
jthomson04, line 133: [`vllm_worker_async.py:97`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L97)

This role allowlist is doing more work than it appears to — it's load-bearing for **correctness**, not just conservatism, and that's worth a comment.

[`safe_apply_chat_template`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/hf.py#L730-L734) hoists and merges every system message to index 0:
```python
if any(msg["role"] == "developer" for msg in conversation) and not _detect_developer_role_support(chat_template):
    conversation = _convert_developer_to_system(conversation)
    conversation = _consolidate_system_messages(conversation)
```
That reordering would relocate the boundary marker relative to `assistant_ordinal` and silently produce a wrong prompt. It can only never fire because `developer` is rejected here. A future maintainer widening this set would break the splice with no test to catch it.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -51,13 +58,55 @@
...
+            return None, "unsupported message role"
+        content = message.get("content")
+        if content is not None and not isinstance(content, str):
```
jthomson04, line 138: [`vllm_worker_async.py:98-100`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L98-L100)

A coverage question, not a defect — and I want to be explicit that your numbers already establish the fast path fires.

Your "zero `/tokenize` calls" result is checkable from the Gym side and is the strong form of that evidence: Gym only calls `/tokenize` when a response message lacks token IDs (`responses_api_models/vllm_model/app.py:694`, `if self.config.return_token_id_information and "prompt_token_ids" not in choice_dict["message"]`), and NeMo-RL populates exactly that field from the engine's own output ([`choice.message.prompt_token_ids = list(prompt_token_ids)`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/utils.py#L409)). So zero `/tokenize` calls means assistant messages carried engine token IDs end-to-end and were echoed back on the next turn — precisely the input this fast path consumes. Combined with "real multi-turn injection use, zero unexpected fallbacks", the round-trip is demonstrably live.

The narrower point: this gate requires `content` to be a `str`, and Gym's **Responses→Chat converter** produces list content — `nemo_gym/responses_converter.py:265-281` rewrites `m["content"]` to `[{"type": "text", "text": ...}]`, `:302-308` passes that list into `NeMoGymChatCompletionUserMessageParam`, and `chat_completions()` forwards `body.model_dump()` verbatim. Note `:154-166` wraps even a bare-string `input` into a list. So for Responses-API-driven agents the first user message trips this check and the optimization silently never engages.

Would flattening a single `{"type": "text"}` part to a string here widen coverage to those agents? Pairs with the counters comment below — that's the only way anyone would notice either way.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -652,7 +887,12 @@ class NeMoRLOpenAIServingChat(NeMoRLOpenAIServingChatMixin, OpenAIServingChat):
             pass
 
         class NeMoRLOnlineRenderer(NeMoRLOpenAIServingMixin, OnlineRenderer):
-            pass
+            def __init__(self, *args, **kwargs):
+                super().__init__(*args, **kwargs)
+                self._complete_token_injection_attempts = 0
```
jthomson04, line 1071: [`vllm_worker_async.py:889-895`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L889-L895)

These three counters are incremented but never read — a repo-wide grep finds no reader, no export, and no test; only `_logged_complete_token_injection_fallbacks` is read, for warn-once dedup.

That matters more than usual here, for two reasons. It's the only way to answer the content-shape question above, and because [`_record_complete_token_injection_fallback`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L507-L515) logs each reason exactly once per renderer lifetime, an operator on a multi-hour run sees one WARNING at the start and nothing after. Given a silent fallback can change the prompt (see the comment on the e2e test), "did the fast path fire?" is currently unanswerable in production. Either exposing them or dropping them would be an improvement.

nemo_rl/models/generation/openai_server_utils.py
```diff
@@ -13,17 +13,172 @@
...
+
+    suffix_pos = marked_text.find(
+        eos_token, marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+    )
+    if suffix_pos < 0:
+        raise CompleteTokenInjectionError(
+            "The chat template did not close the tokenized assistant with EOS."
+        )
+    marker_end = marker_pos + len(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+    if marked_text[marker_end:suffix_pos].strip():
```
jthomson04: [`openai_server_utils.py:154-159`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L154-L159)

Worth a comment saying the whitespace drop is intentional. `.strip()` lets a pure-whitespace gap through, and those tokens don't reach the prompt — which reads like an oversight but is actually correct: `replace_prefix_tokens` [cuts at the EOS token](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L263-L283) and drops the same gap, so the two paths are byte-identical here.

For what it's worth the branch is reachable — of the 34 templates vLLM v0.25.1 ships, [`tool_chat_template_llama4_json.jinja:97`](https://github.com/vllm-project/vllm/blob/v0.25.1/examples/tool_chat_template_llama4_json.jinja#L97) emits `"\n<|eot|>"` and [`tool_chat_template_granite_20b_fc.jinja:121`](https://github.com/vllm-project/vllm/blob/v0.25.1/examples/tool_chat_template_granite_20b_fc.jinja#L121) emits `' <|endoftext|>'`. Tightening the guard would push those two families permanently onto the fallback while changing zero output tokens, so the current behavior is the right call — it just deserves a line explaining why.

tests/unit/models/generation/test_openai_server_utils.py
```diff
@@ -15,7 +15,221 @@
...
+    result = build_complete_prompt_token_ids(
+        tokenizer=tokenizer,
+        conversation=conversation,
+        assistant_ordinal=1,
+        model_prefix_token_ids=[10, 777, 2, 40, 31, 32, 2],
+        render_prompt_text=_render_marked_conversation,
+    )
+
+    assert result == [10, 777, 2, 40, 31, 32, 2, 40, 99]
+    assert conversation[1]["tool_calls"][0]["function"]["arguments"] == {
```
jthomson04: [`test_openai_server_utils.py:100`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/tests/unit/models/generation/test_openai_server_utils.py#L100)

This assertion targets the wrong message, so the no-mutation property isn't actually covered. `assistant_ordinal=1` marks the assistant at index **3**, but this checks index **1** — a message the function never touches.

Confirmed by mutation testing: replacing [`deepcopy(conversation)`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L99) with a shallow `list(conversation)` leaves all 14 tests in this file green. That matters because `_preprocess_chat_with_complete_token_ids` [returns `conversation` back to vLLM](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L637), so an in-place mark would corrupt it downstream. Asserting against a deep copy of the whole conversation, with `assistant_ordinal=0`, catches it.

tests/unit/models/generation/test_openai_server_utils.py
```diff
@@ -15,7 +15,221 @@
...
+        assert result is None
+    else:
+        assert result is not None
+        assistant_ordinal, message = result
+        assert assistant_ordinal == expected_ordinal
+        assert message is messages[expected_index]
+
+
+@pytest.mark.parametrize("gap", ["", " \n\t"])
+def test_complete_token_injection_requires_adjacent_assistant_eos(gap) -> None:
```
jthomson04: [`test_openai_server_utils.py:179-180`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/tests/unit/models/generation/test_openai_server_utils.py#L179-L180)

The name says "requires adjacent" but the body asserts the opposite — that a `" \n\t"` gap is *tolerated* and yields the same tokens as no gap. The behavior is correct (see the comment on `openai_server_utils.py:155`); it's the name that misleads.

```suggestion
@pytest.mark.parametrize("gap", ["", " \n\t"])
def test_complete_token_injection_tolerates_whitespace_before_assistant_eos(gap) -> None:
```

tests/unit/models/generation/test_openai_server_utils.py
```diff
@@ -15,7 +15,221 @@
...
+
+
+def test_complete_token_injection_requires_assistant_eos() -> None:
+    tokenizer = _CompleteTokenTokenizer()
+    conversation = [
+        {"role": "assistant", "content": "first"},
+        {"role": "user", "content": "next"},
+    ]
+
+    with pytest.raises(CompleteTokenInjectionError, match="did not close"):
```
jthomson04: [`test_openai_server_utils.py:223`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/tests/unit/models/generation/test_openai_server_utils.py#L223)

`match="did not close"` also matches the *other* error message, ["did not close the tokenized assistant immediately; ..."](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L156-L159), so this test would still pass if the wrong error were raised. The sibling at [:201](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/tests/unit/models/generation/test_openai_server_utils.py#L201) already uses a specific matcher.

```suggestion
    with pytest.raises(
        CompleteTokenInjectionError,
        match="did not close the tokenized assistant with EOS",
    ):
```

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -2130,6 +2324,44 @@ async def test_vllm_http_server_correct_merged_tokens_matches_baseline(
         vllm_http_server_generated_token["token"].removeprefix("token_id:")
     )
 
+    # Compare the injected continuation prompt with the native /tokenize path.
```
jthomson04: [`test_vllm_generation.py:2327`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/tests/unit/models/generation/test_vllm_generation.py#L2327)

Two notes on this comparison, which is a genuinely valuable addition.

First, the comment is slightly off: `/tokenize` binds `NeMoRLTokenizeChatRequest`, not [`NeMoRLChatCompletionRequest`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L673-L676), so it takes the legacy `replace_prefix_tokens` path rather than native preprocessing. This is a new-path-vs-legacy-path equivalence check — worth saying so, since it also explains why the whitespace behavior above is consistent rather than contradictory.

Second, because `preprocess_chat` falls back to that *same* legacy path on any error, the assertion degenerates to legacy == legacy if injection never runs — it passes either way. Combined with the patch coverage showing zero on [`_preprocess_chat_with_complete_token_ids`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/vllm/vllm_worker_async.py#L517-L637), there's currently no artifact that distinguishes "the fast path ran" from "the fast path silently never ran". Asserting `_complete_token_injection_successes == 1`, or `caplog`-asserting that no fallback warning fired, would make the claim falsifiable.

tests/unit/models/generation/test_openai_server_utils.py
```diff
@@ -15,7 +15,221 @@
...
+        build_complete_prompt_token_ids(
+            tokenizer=tokenizer,
+            conversation=conversation,
+            assistant_ordinal=0,
+            model_prefix_token_ids=[31, 32, 2],
+            render_prompt_text=lambda messages: (
+                "NEMO_RL_PREFIX_BOUNDARY_7F3A9C1E missing boundary close"
+            ),
+        )
```
jthomson04, line 301: [`test_openai_server_utils.py`](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/tests/unit/models/generation/test_openai_server_utils.py)

`codecov/patch` is currently failing at 58.98% against an 80% target. The gap is concentrated in the `CompleteTokenInjectionError` branches of `openai_server_utils.py` — 18 uncovered lines (65, 72, 75-76, 104, 114, 124-127, 132, 136, 143, 166-169, 174), all pure-Python and GPU-free.

These are cheap to reach with the existing `_CompleteTokenTokenizer` fixture: marker collision, ordinal absent after parsing, tokenizer missing `eos_token`/`eos_token_id`, render returning non-`str`, render raising, template dropping the marker, `_coerce_token_id_list` non-list/non-int, and suffix not starting with EOS. Covering them takes this file to ~100% and should clear the patch gate on its own. One case worth adding while you're here: every existing test passes a prefix ending in EOS, so the max-tokens-truncation branch at [:177-179](https://github.com/NVIDIA-NeMo/RL/blob/ace567be0b10b42be8bfc66048ca7096e4c28343/nemo_rl/models/generation/openai_server_utils.py#L177-L179) is never exercised.

Separately, `_coerce_token_id_list` accepts `["2", "40"]` and truncates `[2.9]` to `[2]`, which is looser than its "must contain only integer token IDs" message implies — harmless for `tokenizer.encode` callers, but this module is shared with the TRT-LLM server.

nemo_rl/models/generation/openai_server_utils.py
jthomson04: Implemented. Empty model prefixes now raise CompleteTokenInjectionError and use native preprocessing. The focused test covers this case.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Implemented. Eligibility now compares prefix length and both token segments against the selected assistant metadata without making another combined list. A mismatch has a stable fallback reason and test.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Implemented. Known eligibility fields now use direct access so vLLM API drift fails clearly. The private prompt-extras call was also removed.

nemo_rl/models/generation/openai_server_utils.py
jthomson04: Implemented. CopyError and TypeError from deepcopy are now wrapped as CompleteTokenInjectionError, with focused fallback coverage.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Documented. The role allowlist comment now explains that vLLM can consolidate developer and system messages and invalidate the assistant ordinal.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Implemented. Prompt extras are now applied directly, and the method docstring identifies vLLM 0.25.1 OnlineRenderer.preprocess_chat as the source behavior to re-diff on upgrades.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Kept on the safe fallback as planned. Responses-style content-part lists now have an explicit comment and a production-shaped test with a stable non-text message content reason.

nemo_rl/models/generation/openai_server_utils.py
jthomson04: Documented. The comment now explains that whitespace before EOS is dropped intentionally to match legacy prefix replacement.

tests/unit/models/generation/test_vllm_generation.py
jthomson04: Implemented. The HTTP test snapshots the private worker counters and proves exactly one injection success with no new fallback, while still checking equality with the legacy tokenize result.

tests/unit/models/generation/test_openai_server_utils.py
jthomson04: Fixed. The test name now states that whitespace before assistant EOS is tolerated.

tests/unit/models/generation/test_openai_server_utils.py
jthomson04: Implemented. Added focused cases for every listed error branch, strict integer token IDs, and both prefix-ending branches. The helper module now reports 100 percent line coverage locally.

tests/unit/models/generation/test_openai_server_utils.py
jthomson04: Fixed. The test now matches the exact missing-EOS error text.

tests/unit/models/generation/test_openai_server_utils.py
jthomson04: Fixed. The test now compares the full conversation against a deep copy, so mutation of the selected assistant is detected.

nemo_rl/models/generation/openai_server_utils.py
```diff
@@ -13,17 +13,181 @@
...
+    if not isinstance(marked_text, str):
+        raise CompleteTokenInjectionError("The marked prompt must render as text.")
+
+    marker_pos = marked_text.find(_COMPLETE_TOKEN_BOUNDARY_MARKER)
+    if marker_pos < 0:
+        raise CompleteTokenInjectionError(
+            "The chat template did not preserve the complete-token boundary marker."
+        )
+
+    eos_token = getattr(tokenizer, "eos_token", None)
```
yuki-97: Injection is silently dead for the entire Gemma family: `tokenizer.eos_token` diverges from the template's assistant-close token.

`google/gemma-3-{1b,27b}-it` and `gemma-2-9b-it` all have `eos_token = "<eos>"` (id 1), but the chat template closes assistant turns with `<end_of_turn>\n` (id 106). `marked_text.find("<eos>", ...)` at line 152 returns -1 every request → `CompleteTokenInjectionError` → silent fallback; only the fallback counter shows it. The PR's 96% preprocessing speedup is unavailable for any Gemma-on-Gym combo, and the same class extends to any tokenizer where `eos_token` isn't the template's assistant-close literal.

**Suggest fix:** probe the real assistant-close token at renderer startup (render an empty assistant turn, capture the first token after the marker), or expose an `assistant_close_token` config knob. At minimum, emit a one-shot warning when this fallback reason fires so the perf regression isn't silent.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -483,6 +725,57 @@ async def preprocess_chat(
...
+                            renderer_supported=isinstance(self.renderer, HfRenderer),
+                        )
+                    )
+                    if fallback_reason is not None:
+                        self._record_complete_token_injection_fallback(fallback_reason)
+                        self._maybe_log_complete_token_injection_stats()
+                    else:
+                        assert assistant_ordinal is not None
+                        try:
+                            res = await self._preprocess_chat_with_complete_token_ids(
```
yuki-97, line 827: Sentinel `max_tokens=1` (introduced in #2514 to sidestep vLLM's own length check while NeMo-RL computes its own clamp) leaks into `VLLMValidationError` messages on any non-`CompleteTokenInjectionError` exception.

`_set_max_tokens(request, 1)` at :726 mutates before both branches; injection `try` at :749 catches only `CompleteTokenInjectionError`, native `try` at :790 catches `(ValueError, VLLMValidationError)` and re-raises — both leave `max_tokens=1` in place. `_token_len_check` in vLLM `renderers/params.py:441` embeds it in the error text, so the client sees "requested 1 output tokens" instead of what they actually asked for. Realistic when Gym's multi-turn context grows past `max_model_len`.

**Suggest fix:** wrap both sites in `try/finally` restoring `actual_request_max_tokens` on non-success exit. Also add a test where Gym submits `max_new_tokens` such that `prompt_len + max_new_tokens > max_model_len`, and asserts the raised error's requested-tokens field matches the caller's value (not `1`). No existing test in `tests/unit/models/generation/test_vllm_generation.py` covers this path today.

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -143,6 +145,200 @@
...
+def _complete_token_request(**updates):
+    values = {
+        "stream": False,
+        "n": 1,
+        "continue_final_message": False,
+        "echo": False,
+        "return_assistant_tokens_mask": False,
+        "add_special_tokens": False,
+        "return_prompt_text": False,
+        "required_prefix_token_ids": [3, 4],
```
yuki-97: Trivially-zero fixture: a length-only or slot-swap regression in `_prefix_matches_tokenized_assistant` would still pass here.

`prompt=[3]`, `gen=[4]`, `required_prefix=[3, 4]` — all slot lengths collapse to 1, so any 2-token list satisfying `[:1]==[3]` and `[1:]==[4]` is the only such list. The mismatch test at :329 uses `[1, 2]`, which catches gross substitutions but not slot-swap.

**Suggest fix:** make the fixture use distinct lengths — e.g. `prompt=[10, 11]`, `gen=[12, 13, 14]`, `required_prefix=[10, 11, 12, 13, 14]`. A slot-swapped candidate `[12, 13, 14, 10, 11]` would then fail the byte-equality checks.

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -143,6 +145,200 @@
...
+                    "content": "partial",
+                    "prompt_token_ids": [1],
+                }
+            ],
+            "no fully tokenized assistant",
+            id="partial_metadata",
+        ),
+    ],
+)
+def test_complete_token_injection_eligibility_rejects_messages(
```
yuki-97, line 334: No positive test locks the role allowlist `{"system", "user", "assistant", "tool"}` at `vllm_worker_async.py:119`.

The parametrized tests only assert rejection of `"developer"`, and `_tokenized_assistant_messages()` uses only `user`/`assistant` — so accidentally narrowing the allowlist (dropping `"system"` or `"tool"`) would not fail any test.

**Suggest fix:**
```python
@pytest.mark.parametrize("role", ["system", "user", "assistant", "tool"])
def test_eligibility_accepts_supported_roles(role):
    messages = [{"role": role, "content": "x"}, *_tokenized_assistant_messages()]
    _, reason = _complete_token_injection_eligibility(
        _complete_token_request(), messages,
        has_multimodal_config=False, prompt_embeds_enabled=False, renderer_supported=True,
    )
    assert reason is None
```

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -51,13 +59,82 @@
...
+        return None, "multimodal model"
+    if prompt_embeds_enabled:
+        return None, "prompt embeddings enabled"
+    if not renderer_supported:
+        return None, "unsupported renderer"
+    for message in messages:
+        if not isinstance(message, dict):
+            return None, "non-object message"
+        # vLLM can consolidate developer and system messages, which can change
+        # the assistant ordinal and invalidate the splice boundary.
```
yuki-97: Comment reads as if it's justifying tolerance for system consolidation, but the branch below rejects `developer` entirely — so consolidation never fires on the eligible path.

vLLM v0.25.1 `renderers/hf.py:730-738` gates both `_convert_developer_to_system` and `_consolidate_system_messages` on `any(role == "developer") AND not _detect_developer_role_support(...)`. For messages that pass this allowlist, neither helper runs and system messages go through untouched.

```suggestion
        # vLLM rewrites messages only when a developer role is present (converts to system,
        # then merges systems to index 0), which would shift the assistant ordinal.
```

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -447,6 +539,158 @@ def _clamp_max_tokens(
...
+                default_template_kwargs,
+                tool_dicts,
+                parser,
+                assistant_ordinal: int,
+                *,
+                skip_mm_cache: bool,
+            ):
+                """Preserve vLLM preprocessing while skipping prompt re-encoding.
+
+                This mirrors vLLM 0.25.1 ``OnlineRenderer.preprocess_chat``.
```
yuki-97, line 633: Injection depends on vLLM 0.25.1 semantics but has no runtime pin — the docstring says "mirrors vLLM 0.25.1" and nothing checks it.

If `pyproject.toml` bumps vLLM and `safe_apply_chat_template` / `_convert_developer_to_system` / `_consolidate_system_messages` / `HfRenderer` public API changes, injection silently produces wrong prompts. The `_prefix_matches_tokenized_assistant` byte check only guards the prefix, not the suffix rendering — silent off-policy training is the failure mode. Same fragility underlies [the allowlist comment at :117-118](https://github.com/NVIDIA-NeMo/RL/pull/3581#discussion_r3772415660).

**Suggest fix:** at `NeMoRLOnlineRenderer.__init__` (or module load), `assert vllm.__version__ == "0.25.1"` (or a small whitelist of verified versions) so a version bump forces a re-diff before injection silently drifts. A startup probe that renders a fixed conversation and checks the transformation is more robust but heavier.

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -2130,6 +2363,50 @@ async def test_vllm_http_server_correct_merged_tokens_matches_baseline(
...
+    native_response = requests.post(url=f"{base_urls[0]}/../tokenize", json=native_body)
+    native_response.raise_for_status()
+    stats_before_injection = _get_complete_token_injection_stats(vllm_generation)
+    injected_response = requests.post(
+        url=f"{base_urls[0]}/chat/completions", json=continuation_body
+    )
+    injected_response.raise_for_status()
+    assert (
+        injected_response.json()["choices"][0]["message"]["prompt_token_ids"]
+        == native_response.json()["tokens"]
```
yuki-97, line 2816: Parity test only covers Qwen3-0.6B / TP1 / greedy / single-continuation.

The `model_name == "Qwen/Qwen3-0.6B"` gate at :2268 means Gemma / Llama / Nemotron never exercise this parity. If `eos_token` diverges from the template's assistant-close ([see the Gemma comment on `openai_server_utils.py:145`](https://github.com/NVIDIA-NeMo/RL/pull/3581#discussion_r3772369521)), injection silently 100%-fallbacks and `successes` stays at 0 — no test catches that. Also unrepresented: multi-turn accumulation, prompt near `max_model_len` ([see the `max_tokens=1` comment](https://github.com/NVIDIA-NeMo/RL/pull/3581#discussion_r3772369528)).

**Suggest fix:** parametrize over a small model matrix (`Qwen3-0.6B`, `gemma-3-1b-it`, `Llama-3.1-8B-Instruct`) and assert `successes == successes_before + 1` per turn — a family where injection silently fallbacks would flip that to 0 and be caught. Optional second test: N-turn loop with final prompt crossing `max_model_len`, asserting the raised error's requested-tokens field equals the caller's value.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -652,7 +945,12 @@ class NeMoRLOpenAIServingChat(NeMoRLOpenAIServingChatMixin, OpenAIServingChat):
             pass
 
         class NeMoRLOnlineRenderer(NeMoRLOpenAIServingMixin, OnlineRenderer):
-            pass
+            def __init__(self, *args, **kwargs):
```
yuki-97, line 1069: Design suggestion: make injection a config-toggleable feature, default-on.

Given the edge-case surface area in this review ([Gemma `eos_token`](https://github.com/NVIDIA-NeMo/RL/pull/3581#discussion_r3772369521), [`max_tokens=1` sentinel](https://github.com/NVIDIA-NeMo/RL/pull/3581#discussion_r3772369528), [misleading allowlist comment](https://github.com/NVIDIA-NeMo/RL/pull/3581#discussion_r3772415660), [missing vLLM version pin](https://github.com/NVIDIA-NeMo/RL/pull/3581#discussion_r3772454060)): if any situation isn't caught and causes a mismatch, users could just flip a config to use the unoptimized path immediately, rather than waiting for a fix. A `vllm_cfg.enable_complete_token_injection: true` flag (default-on to preserve the 96% speedup, gated at the `isinstance(request, NeMoRLChatCompletionRequest)` branch on :728) would do it. Also useful as an A/B debugging tool when training drift is suspected.

Do you think this is worth adding? @jthomson04 @terrykong

yuki-97 (commented):
thanks @jthomson04 , nice optimization!

tests/unit/models/generation/test_vllm_generation.py
jthomson04: Fixed in 71cbbccdf62d2d57edfba3b7c4edb9b7aca11cd4. The test is parameterized across system, user, assistant, and tool so narrowing the supported role set now fails.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Fixed in 71cbbccdf62d2d57edfba3b7c4edb9b7aca11cd4. NeMo-RL already pins vLLM 0.25.1 in pyproject.toml and uv.lock. The fast path now also checks the loaded vLLM version and uses the safe native fallback for any other version. The fallback reason is covered by the eligibility test.

nemo_rl/models/generation/openai_server_utils.py
jthomson04: Fixed in 71cbbccdf62d2d57edfba3b7c4edb9b7aca11cd4. The splice now probes the actual assistant-close sequence by rendering the selected assistant as the final message with add_generation_prompt=false. It verifies that the full render starts with the same close sequence and removes only the token overlap already present in the model prefix. A Gemma-style test covers eos_token=<eos> with <end_of_turn> as the template close, with and without the close token already in the prefix.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Fixed in 71cbbccdf62d2d57edfba3b7c4edb9b7aca11cd4. The comment now states that vLLM rewrites messages only when a developer role is present and explains how that rewrite shifts the assistant ordinal.

tests/unit/models/generation/test_vllm_generation.py
jthomson04: Fixed in 71cbbccdf62d2d57edfba3b7c4edb9b7aca11cd4. The fixture now uses prompt length 2 and generation length 3. A separate slot-swapped prefix assertion confirms that reversed segments fall back.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: I kept this internal and configuration-free as requested for this PR. Unsupported requests, renderers, vLLM versions, and any splice validation failure automatically use native preprocessing. The one-shot fallback warnings and counters provide the debugging signal without adding a new public configuration surface.

tests/unit/models/generation/test_vllm_generation.py
jthomson04: Addressed in 71cbbccdf62d2d57edfba3b7c4edb9b7aca11cd4 with a model-independent close-token test for the reported Gemma failure mode, including both prefix-ending branches. The existing Qwen HTTP test still proves exact injected-versus-native token equality and a real success counter increment. I did not add a large downloaded-model matrix because these focused tests cover the splice contract without adding heavyweight CI setup.

nemo_rl/models/generation/vllm/vllm_worker_async.py
jthomson04: Fixed in 71cbbccdf62d2d57edfba3b7c4edb9b7aca11cd4. Every error path after the temporary max-token override now restores the caller value. Context-length validation errors are rebuilt with the real requested output-token count. The focused server test covers the native context error and an unexpected injection error, and checks that max_completion_tokens is restored.

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -143,6 +150,266 @@
...
+
+    assert ordinal == 1
+    assert reason is None
+
+
+@pytest.mark.parametrize("role", ["system", "user", "assistant", "tool"])
+def test_complete_token_injection_eligibility_accepts_supported_roles(role) -> None:
+    messages = [{"role": role, "content": "text"}, *_tokenized_assistant_messages()]
+
+    _, reason = _complete_token_injection_eligibility(
```
yuki-97: **Nothing pins the ordinal-counting convention that `find_latest_tokenized_assistant` and `_assistant_index_from_ordinal` must share.** Both count *every* assistant (openai_server_utils.py:50 and :66). If either drifted to counting only tokenized ones the marker would land on an earlier turn, and the marker-count, close-probe and `startswith` guards would all still pass — a valid-but-wrong prompt with nothing firing.

Moving `assistant_ordinal += 1` inside the `if` at openai_server_utils.py:51-56 leaves every test in both files green. This case is the only discriminating shape in the suite — `role="assistant"` prepends a metadata-free assistant, so the correct ordinal is 2 — but the ordinal is discarded here.

**Suggest fix:** capture and assert it; checked against the suite, nothing else moves.

```python
    ordinal, reason = _complete_token_injection_eligibility(
        ...
    )

    assert reason is None
    assert ordinal == (2 if role == "assistant" else 1)
```

nemo_rl/models/generation/openai_server_utils.py
```diff
@@ -13,17 +13,273 @@
...
+        raise CompleteTokenInjectionError(
+            "The contextual suffix did not begin with the assistant-close sequence."
+        )
+    model_prefix_token_ids = _coerce_token_id_list(
+        model_prefix_token_ids, "Model prefix token IDs"
+    )
+    if not model_prefix_token_ids:
+        raise CompleteTokenInjectionError("The model prefix token IDs were empty.")
+
+    max_overlap = min(len(model_prefix_token_ids), len(assistant_close_token_ids))
```
yuki-97: **The splice's correctness rests on an unstated precondition: the assistant-close sequence begins at an added-token boundary.** Under it, :234's `startswith` anchors the text boundary, :258-265 anchors the token boundary, the added token is a hard tokenizer split so the `model_prefix | suffix` seam cannot merge, and a stop token cannot produce a false `overlap` here. Outside it none of the four hold, and no guard between :243 and :281 constrains the close sequence's token type.

Not a live bug — every template that reaches this path satisfies it (`<|im_end|>`, `<|eot_id|>`, `<end_of_turn>`; `passthrough_prompt_response` emits no close and is already rejected at :229). But the docstring states the `add_generation_prompt` contract at :117-123 and not this one, so the invariant currently lives only in the implementation.

**Suggest fix:** document it. If you want it enforced, `tokenizer.convert_ids_to_tokens(assistant_close_token_ids[0]) in tokenizer.get_added_vocab()` after :246 — not `all_special_ids`, since Gemma-3's `<end_of_turn>` lives only in `added_tokens_decoder`, so that predicate would disable the feature for Gemma and redden the `hf_gated` test at tests/unit/models/generation/test_openai_server_utils.py:510.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -51,11 +59,97 @@
...
+    if request.echo:
+        return None, "echo request"
+    if request.return_assistant_tokens_mask:
+        return None, "assistant token mask request"
+    if request.add_special_tokens:
+        return None, "add_special_tokens request"
+    if request.return_prompt_text:
+        return None, "return_prompt_text request"
+    if has_multimodal_config:
+        return None, "multimodal model"
```
nanz-nv: Currently this check rejects multimodel model. Do you think it is possible to loose this check and allow text only input for models with multimodal capability?

jthomson04: Addressed in 8203799818569dc6baec063082a2d8555cc21967. The model-capability guard is removed, so a VLM-capable model can use complete-token injection for plain-string chat requests. Content-part requests remain on native preprocessing, and parsed `mm_data` or `mm_uuids` still causes a native fallback. Focused injection and HTTP-server tests pass.

PR #3599, opened 2026-08-12
fix(async-grpo): eliminate prompt skips across checkpoint save/restore

yfw:
# What does this PR do ?

Makes async-GRPO checkpoint resume lossless: no prompt is skipped and none is duplicated across a save/restore cycle, for both `load_replay_buffer=true` and `false`.

## Problem

On a SWE E2E run, 675 of 736 skipped prompts (~30% of the reached dataset) were traced to two mechanisms:

1. **Non-atomic checkpoint.** The saved dataloader cursor runs ahead of what training consumed: the cursor advances on every yield, the buffer holds only *completed* groups, and in-flight work is serialized nowhere. Each restore lost a ~96-prompt window. The loss is also length-biased — a group completes only when its longest rollout finishes, so the lost in-flight set is exactly the long-rollout prompts.
2. **Gap-fill tail discard.** When a target needed fewer prompts than a dataloader batch holds, `_process_batch` sliced off what it needed and dropped the rest (61 of the 736 skips).

## Fix (one commit each)

1. **Carry over gap-fill tails** (`b1b196a1e`) — `_process_batch` returns the unconsumed remainder; the loop consumes it before the next pull, and it is serialized in `rollouts.pt` so a checkpoint can't strand it.
2. **Frontier-aligned checkpointing** (`4edd6976`) — every yielded prompt is stamped with a monotonic stream ordinal; the collector keeps a small ring of pre-pull dataloader snapshots keyed by ordinal. Checkpoints persist the snapshot at the trained frontier instead of the live cursor, plus frontier metadata. On resume the dataloader re-yields the covered window; rows already trained or retained in the restored buffer are dropped at yield, and the existing gap-fill path regenerates exactly the remainder. Un-stampable batches and legacy (pre-frontier) checkpoints fall back to today's live-cursor behavior.
3. **`checkpointing.load_replay_buffer`** (`a72ebf48`) — set `false` to skip the buffer restore and regenerate the whole buffered window fresh, removing the short-rollout bias of retained groups. With frontier alignment this is just the empty-retained-set case of the same resume path — no second code path, and no prompts are skipped either way. Default remains `true`. Supersedes #3583.
4. **A/B regression test** (`aee3c801`) — drives the legacy and frontier resume paths through the same scenario: legacy loses prompts `[5, 8, 9, 11]`, frontier regenerates all of them with zero duplicates.
5. **Self-review fixes** (`a482c6c5`, `8c1e170d`) — frontier derived from the sampled groups' own ordinals instead of `consumed_samples`; dataloader snapshot and pending batch read in one actor call under the pending lock; native group ordinals captured from the rollout's *input* rows; shared rollout-state builder; fallback warnings, scoped docs, and `rollouts.pt` contract tests.
6. **Conservative checkpoint cut** (`3b32a8b6`) — thanks to @terrykong for the failure scenario: a target refilled from later prompts (tolerated failure, or a resume gap-filling an incomplete restored target) can train past another target's in-flight groups, and a frontier-aligned save would strand them. Checkpoints now cut at `min(lowest in-flight ordinal, trained frontier)` and persist the ordinals already trained above the cut, so a resume regenerates exactly what was lost — no skip, no re-training. Healthy path unchanged (cut == frontier, empty trained list); every step mutation-tested.

## Guarantees and caveats

- This buys **data determinism** (exact prompt coverage and order), not bitwise reproducibility: regenerated trajectories are fresh samples under post-resume weights (in-flight vLLM state is not serializable). Retained groups are reused bitwise.
- With `async_grpo.max_generation_failures > 0`, prompts dropped by a tolerated failure are lost during **normal operation** — that loss happens at failure time, not at resume, and a resume neither recovers them nor (with the ordinal-derived frontier) re-trains or double-counts anything around them. The default (`0`) fails fast and is unaffected.
- Resuming from a pre-upgrade checkpoint disables frontier alignment **permanently for that checkpoint lineage** — the run and every run descended from its checkpoints stay on live-cursor semantics (a warning is logged at resume). Fresh runs are protected from their first checkpoint.

## Validation (E2E kill/resume on a production workload)

Nemotron nano-v3.5 RLVR via NeMo-Gym (SWE/agentic + multilingual chat blend), 512-prompt slice, `shuffle: false`, 32 prompts/step, checkpoint every step. Jobs were `scancel`ed mid-generation and relaunched under the same lineage; the `train_data_step*.jsonl` dumps were then verified against the dataset's expected stream order — every step must train exactly its 32-prompt window, each prompt exactly `num_generations` times.

- **Run A**: 14 steps across 3 job segments with 2 mid-generation kills (6 steps re-executed after kills landed before checkpoint finalization) → **448/448 prompts, no skip, no duplicate**, and every re-executed step trained the identical prompt set. The resume logs show the mechanism working: `Frontier-aligned resume: dataloader rewound to ordinal 64, trained frontier 64, 13 retained prompt groups` → `dropping 13 already-covered prompts (ordinals 64–95), regenerating 19` — those 19 in-flight prompts are exactly what the pre-PR code silently skipped.
- **Run D** (with the conservative cut, `max_generation_failures=3` as in the SWE recipes): 6 steps, 2 kills → **192/192**, zero `Checkpoint cut lowered` warnings and zero worker failures across all segments — confirming the cut is inert on healthy runs.

Baseline for the same workload class before this PR: ~30% of reached prompts skipped across restarts (675/736, from the original findings).

# Usage

```python
```

**Pre checks**:

# Additional Information
* Credit to @pjin-nvidia who identified this issue and provided reference implementation. And @snowmanwwg for running related test showing impact of saving partial replay buffer.


nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -360,27 +449,152 @@ def _collection_loop(self):
...
+        rows = batch.get("extra_env_info") if hasattr(batch, "get") else None
+        ordinals = [
+            row.get(NEMO_GYM_TASK_INDEX_KEY) if isinstance(row, dict) else None
+            for row in (rows if isinstance(rows, list) else [])
+        ]
+        if not ordinals or any(ordinal is None for ordinal in ordinals):
+            raise ValueError(
+                "Frontier-aligned resume requires task indices on every row, "
+                "but a re-yielded batch is missing them."
+            )
```
yfw: **1 action item.** PR-introduced; blocks CI.

`uv run --group dev pre-commit run --all-files` fails here with 2 pyrefly errors:

```
ERROR `>` is not supported between `None` and `int`   nemo_rl/algorithms/async_utils/trajectory_collector.py:519
ERROR `>=` is not supported between `None` and `int`  nemo_rl/algorithms/async_utils/trajectory_collector.py:529
```

`ordinals` infers as `list[int | None]`, and the `any(... is None)` raise on the next line doesn't narrow the element type for pyrefly.

### AI-1
Bind the narrowed list to a new name. I verified this exact edit locally: pyrefly goes **2 errors -> 0**, and `pytest tests/unit/algorithms/test_async_utils.py::TestAsyncTrajectoryCollector` stays green (38 passed).

I also dropped the `if hasattr(batch, "get")` guard: `BatchedDataDict` subclasses `collections.UserDict`, so `.get` always exists. (Same dead guard at [`:69`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L69) and [`:465`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L465) if you want them too.)

```suggestion
        rows = batch.get("extra_env_info")
        raw_ordinals = [
            row.get(NEMO_GYM_TASK_INDEX_KEY) if isinstance(row, dict) else None
            for row in (rows if isinstance(rows, list) else [])
        ]
        if not raw_ordinals or any(ordinal is None for ordinal in raw_ordinals):
            raise ValueError(
                "Frontier-aligned resume requires task indices on every row, "
                "but a re-yielded batch is missing them."
            )
        ordinals = cast(list[int], raw_ordinals)
```

Note `[int(o) for o in raw_ordinals]` does **not** work — I tried it, and pyrefly rejects `int(None)` for the same non-narrowing reason (1 error). `cast` is already imported in this file.

nemo_rl/algorithms/grpo.py
```diff
@@ -5149,12 +5229,21 @@ def async_grpo_train(
...
-                            trajectory_collector.get_dataloader_state.remote()
+                        # Save the dataloader state aligned with the trained
+                        # frontier (consumed_samples) rather than the live
+                        # cursor, which has already advanced past prompts that
+                        # are still in flight or buffered. A resume re-yields
+                        # the covered window and regenerates whatever the
+                        # restored buffer does not retain, so no prompt is
+                        # skipped. Falls back to the live cursor for runs
+                        # whose ordinals are not frontier-aligned.
+                        dataloader_snapshot = ray.get(
```

**TL;DR — `consumed_samples` is a counter, but `base_ordinal` is a stream position; with `max_generation_failures > 0` (13 shipped configs, including the SWE recipes this PR targets) a resume both re-trains already-trained prompts and permanently skips others.**

PR-introduced. Some context, since this spans three files:

1. Every yielded prompt gets an ordinal equal to its position in the dataloader stream ([`_stamp_task_indices`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L473)).
2. The checkpoint saves the snapshot at the "trained frontier", and the frontier passed in is `grpo_save_state.consumed_samples` (this line).
3. But `consumed_samples` is a pure counter: [`grpo.py:5153`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5153) does `consumed_samples += num_prompts_per_step` once per step.
4. A *tolerated* generation failure discards the prompts dispatched for that target and releases it — [`trajectory_collector.py:1169-1170`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1169) calls `_release_target` in `finally`, and those prompts are never re-queued. A later batch refills the target with **higher** ordinals. `_failure_count` resets on success ([`:1127`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1127)); the holes in the stream never do.

So after `H` cumulative tolerated failures, the true trained stream position is `consumed_samples + H`, while the saved frontier is `consumed_samples`. On resume:

- Ordinals in `[F, F+H)` were **already trained**, but they pass the `ordinal >= frontier` keep-test at [`_filter_covered_rows:529`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L529) and aren't retained (`_prepare_for_training_step` dropped their target), so they are **regenerated and trained a second time**.
- Ordinals below `F` that were dropped-not-trained are filtered out as "covered" and **never regenerated**.

Both halves of the guarantee in the PR title fail, in opposite directions.

This isn't an exotic config. `max_generation_failures: 3` is set in 13 shipped configs — `examples/nemo_gym/grpo_qwen3_30ba3b_thinking_swe{1,2}.yaml` (the same SWE workload the 675/736 measurement came from), all 4 under `examples/nemo_gym/nemotron-3-super/`, and all 7 under `examples/nemo_gym/nemotron-3-ultra/`. At the default of `0` the first failure is fatal, so no further checkpoint is written and the run is unaffected.

Damage is bounded: once the lag exceeds the snapshot ring, `best is None` and you fall back to the live cursor.

### AI-1
Don't derive the frontier from a counter. The trained rows already carry their true ordinals in `extra_env_info`, so `max(trained ordinal) + 1` is available at the same point and is failure-hole-proof. That's the same data you already have, and it makes the frontier mean what `base_ordinal` means.

nemo_rl/algorithms/grpo.py
```diff
@@ -5164,6 +5253,13 @@ def async_grpo_train(
                         rollouts_state = ray.get(
                             trajectory_collector.get_rollouts_state.remote()
                         )
+                        if dataloader_snapshot["frontier_aligned"]:
```
yfw, line 5349: **2 action items.** PR-introduced. Applies to the **fallback** path only — on the frontier path the saved state is an immutable ring snapshot and [`:4321`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L4321) forces `pending_batch: None`, so the tear can't bite there.

**TL;DR — the dataloader cursor and `_pending_batch` are read in two separate actor calls with a full replay-buffer `torch.save` between them, so a fallback checkpoint can persist a cursor older than its pending batch and duplicate every prompt in it.**

The save block reads three things at three different times while the collection thread keeps running (it is *not* paused here — pause/resume close at [`:5010`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5010)/[`:5073`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5073), before checkpointing):

1. [`:5240`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5240) cursor (T1)
2. [`:5249`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5249) full buffer `torch.save` — seconds to minutes
3. [`:5253`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5253) `_pending_batch` (T2)

If the collector consumes the pending batch and pulls the next one during step 2, the checkpoint pairs a cursor at batch K with a pending suffix of batch K+1. On resume the pending batch is dispatched *and* batch K+1 is re-yielded — every prompt in it is generated and trained twice. Pre-PR the tail was discarded, so this path could skip but never duplicate.

Two things make this wider than it looks: `_pending_batch` is non-`None` in the steady state (`_process_batch` returns the whole batch when no target needs generation, [`:355`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L355)), and it is `None` for the *entire* duration of `_process_batch` ([`:404-427`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L404)), which itself does blocking `ray.get`s and can park on `_refit_pause_cleared.wait()` for a whole refit. So a checkpoint in that window loses the pending batch outright.

### AI-1
Return the cursor and the pending batch from a **single** actor call under `_pending_lock`. Cheap strict improvement if you don't want the refactor: read `get_rollouts_state()` *before* `get_checkpoint_dataloader_state()`, which degrades the failure from a duplicate back to a pre-PR-style skip.

### AI-2
`PENDING_PROMPTS_KEY` is written unconditionally by [`get_rollouts_state`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L895) but hard-ignored on frontier resume, so it's dead weight in every frontier checkpoint — unnecessary for VLM runs whose `message_log` carries image tensors. Pop it when `dataloader_snapshot["frontier_aligned"]` is true.

nemo_rl/algorithms/grpo.py
```diff
@@ -4256,6 +4301,41 @@ def async_grpo_train(
...
+            f"{len(retained_task_indices)} retained prompt groups"
+        )
+    else:
+        collector_start_kwargs = {
+            "next_nemo_gym_task_index": next_nemo_gym_task_index,
+            "pending_batch": (rollouts_state or {}).get(PENDING_PROMPTS_KEY),
+            # Ordinal == stream position only holds for runs that have used
+            # frontier-aligned checkpoints from the start; a legacy resume
+            # keeps live-cursor checkpoints.
+            "ordinals_frontier_aligned": last_checkpoint_path is None,
```
yfw, line 4350: **1 action item.** PR-introduced.

`ordinals_frontier_aligned=last_checkpoint_path is None` is the only place this is set outside the frontier branch, and nothing ever sets it back to `True`. So a single legacy resume latches it `False` -> that run writes only fallback checkpoints -> those carry no frontier metadata -> the *next* resume takes the legacy branch again -> `False` again.

That means the degradation is not scoped to one process. It is permanent for the entire descendant checkpoint chain, across arbitrarily many future restarts. The PR description says "Runs resuming from a pre-upgrade checkpoint keep the old cursor semantics **for that run**" — that understates it, and every existing run's first resume after this lands enters this state.

It also isn't merely "you keep the old skips": combined with the torn-checkpoint issue on the fallback path, a legacy-descended lineage now *duplicates* prompts, which it never did before this PR.

### AI-1
Log it once, loudly, in this branch — right now nothing in the log says the headline feature is inactive:

```python
print(
    "⚠️ Legacy checkpoint resume: frontier-aligned checkpointing is disabled "
    "for this run and all checkpoints descended from it; checkpoints will save "
    "the live dataloader cursor and may skip in-flight prompts."
)
```

and correct the PR description's "for that run".

**Context — no action.** Re-establishing alignment via a recorded `ordinal - consumed_samples` offset would fix this properly, but that's a design change, not a review ask.

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -337,10 +397,39 @@ def _collection_loop(self):
...
+                    except StopIteration:
+                        dataloader_exhausted = True
+                        break
+                    first_ordinal = self._next_nemo_gym_task_index
+                    if self._stamp_task_indices(batch) and pre_pull_state is not None:
+                        with self._snapshot_lock:
+                            self._dataloader_snapshots.append(
+                                (first_ordinal, pre_pull_state)
+                            )
+                    batch = self._filter_covered_rows(batch)
```
yfw, line 512: **1 action item.** PR-introduced. Low severity — see the correction at the end, I overstated the reachability of this in the original version of this comment.

`_filter_covered_rows` runs unconditionally right after `_stamp_task_indices`, including when stamping just **failed**. Those two siblings then disagree about what an unstampable batch means:

- [`_stamp_task_indices:465-470`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L465) — warns, sets `_ordinals_frontier_aligned = False`, degrades gracefully.
- [`_filter_covered_rows:514-518`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L514) — raises `ValueError` on the same condition.

If the raise does fire, it propagates to the collection loop's blanket `except Exception` at [`:434`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L434), sets `collection_failed = True`, and the driver turns it into a `RuntimeError` whose message tells the user the **dataloader is exhausted and to increase `data.train.max_num_epochs`** — a wrong diagnosis that would send someone down the wrong path for a while.

### AI-1
Make the two sibling paths agree: skip filtering when stamping returned `False` (and set `_ordinals_frontier_aligned = False`) rather than raising. If you'd rather keep the raise, make it explain itself — "this indicates the dataset or processor changed since the checkpoint was written" — so the misleading exhaustion message doesn't stand alone. Prose rather than a suggestion block because I haven't run either variant.

**Correction — no action.** My original wording said this is "reachable if the dataset or processor changes between the save and the resume." I'm withdrawing that as a reachability claim. The two in-tree processors that emit `extra_env_info: None` (`nemo_rl/data/processors.py:183`, `:826`) do so for *every* sample, so such a run is unstampable from its first batch, never writes frontier metadata, leaves `_resume_frontier_ordinal` as `None`, and returns early at [`:505`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L505) — the raise can't fire. A mixed-processor dataset that reaches it is constructible in principle but I could not demonstrate one in-tree. The finding stands on the sibling inconsistency and the wrong error message, not on a demonstrated failure.

docs/guides/async-grpo.md
```diff
@@ -170,13 +172,17 @@ On resume, the buffer is restored before the trajectory collector starts, then c
 3. **Incomplete targets kept** — target steps that still lack a full batch are kept in the buffer. The collector will *gap-fill* only the missing trajectories for those targets before moving on.
 4. **Buffer truncated** — if the restored count exceeds `max_size`, the buffer is truncated, prioritising entries closest to the resume step.
 
+The dataloader then re-yields the window between the trained frontier and the old cursor. Prompts already trained or retained in the restored buffer are dropped; everything else — in-flight work lost at the save, groups evicted during cleanup — is regenerated in order, so **no prompt is skipped and none is duplicated** across a save/restore cycle. (Resumes from checkpoints written before frontier alignment keep the previous live-cursor behaviour.)
```
yfw: **1 action item.** PR-introduced.

This states the guarantee unconditionally and parenthesises only the legacy-checkpoint case, but the code has three more paths that silently restore lossy behaviour — and one that adds duplication the sentence explicitly denies:

- `max_generation_failures > 0` — see my comment on [`grpo.py:5240`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5240); prompts are both skipped **and** duplicated.
- Unstampable `extra_env_info` rows — [`_stamp_task_indices`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L463) latches `_ordinals_frontier_aligned = False`.
- An empty or evicted snapshot ring — [`get_checkpoint_dataloader_state`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L573) falls back to the live cursor.

Separately, groups that fail generation or age out of the buffer are still lost, as they are during normal operation — the current wording reads as if resume is lossless in an absolute sense.

### AI-1
Scope the claim. Something like: no prompt is skipped **as a result of the save/restore itself**, then list the conditions under which the guarantee does not hold, and state that once a run falls back it stays on live-cursor semantics for all descendant checkpoints.

tests/unit/algorithms/test_async_utils.py
```diff
@@ -1176,6 +1189,7 @@ def create_local_collector(
             replay_buffer=replay_buffer,
             start_step=0,
             next_nemo_gym_task_index=next_nemo_gym_task_index,
+            pending_batch=pending_batch,
```
yfw, line 1323: **1 action item.** PR-introduced coverage gap.

`create_local_collector` forwards `pending_batch=` but not `resume_frontier_ordinal=` / `resume_covered_task_indices=`, so the kwargs [`grpo.py:4316`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L4316) actually builds never reach `__init__` in any test. Every frontier test hand-pokes the private attrs instead (`:1478-1481`, `:1548-1553`), including `_skip_horizon`, which means the tests encode the *intended* semantics while production computes something else — see my comment on [`trajectory_collector.py:186`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L186).

It also leaves `load_replay_buffer: false` (the empty-retained-set case this PR advertises) with no coverage of its constructor path.

### AI-1
Forward the resume kwargs, then drop the attribute pokes:

```suggestion
        pending_batch=None,
        resume_frontier_ordinal=None,
        resume_covered_task_indices=None,
    ):
```

and pass them through to `collector_cls(...)` alongside `pending_batch=pending_batch`.

tests/unit/algorithms/test_grpo.py
```diff
@@ -1037,6 +1101,23 @@ def get_rollouts_state(self):
         mock.remote = MagicMock(return_value={NEXT_NEMO_GYM_TASK_INDEX_KEY: 0})
         return mock
 
+    @property
+    def get_checkpoint_dataloader_state(self):
+        """Return a remote-callable mock yielding a frontier snapshot result.
+
+        Mirrors the fallback shape (not frontier-aligned) so the checkpoint
+        path persists the plain dataloader state without frontier metadata.
+        """
```
yfw, line 1161: **1 action item.** PR-introduced coverage gap.

This stub hardcodes `"frontier_aligned": False`, so the checkpoint path only ever exercises the fallback branch. `grep -rn "FRONTIER_ORDINAL_KEY\|RESUME_BASE_ORDINAL_KEY" tests/` returns **zero matches** (I verified), which means both halves of the checkpoint-file contract are dead in tests:

- the write at [`grpo.py:5257-5262`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5257)
- the read-back and `collector_start_kwargs` build at [`grpo.py:4309-4331`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L4309), including plumbing `RETAINED_TASK_INDICES_KEY` into `resume_covered_task_indices`

The collector-level mechanics are well covered; the file contract that connects them is not. A typo in either key name, or forgetting to plumb `base_ordinal` into `next_nemo_gym_task_index`, ships green today.

(`RETAINED_TASK_INDICES_KEY` *is* referenced in tests — it's specifically the two frontier keys that aren't.)

### AI-1
Parametrise the stub over both shapes and assert `torch.load(rollouts.pt)` carries the two keys with `consumed_samples` / `base_ordinal`.

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -501,6 +719,7 @@ def _run_rollout_batch() -> None:
             import traceback
 
             traceback.print_exc()
+            return leftover
```
yfw: **1 action item.** PR-introduced as a gap in this PR's own contract — the docstring at [`:588-593`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L588) says "The caller re-queues it so no yielded prompt is ever discarded." Not a regression vs `main`, which dropped the whole batch anyway.

`leftover` is `None` in two opposite situations and the caller can't tell them apart:

- **Raise after the gap-fill slice** at [`:639`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L639) but before `worker.start()` at [`:698`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L698) — `leftover` holds only the suffix, and the prefix already sliced into `rollout_batch` is dropped with no record. Live raisers in that window: `_should_use_nemo_gym` ([`:649`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L649)), `attach_initial_nemo_gym_image_payloads` ([`:668`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L668)), `repeat_interleave` ([`:669`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L669)), and the thread start itself.
- **Raise before anything is consumed** (e.g. the `ray.get` at [`:623`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L623) on a transient actor failure) — `leftover` is still `None`, so [`:425`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L425) reads "fully consumed" and the whole batch is dropped.

### AI-1
`worker_started` already distinguishes exactly these cases:

```suggestion
            # No worker was started, so nothing in `batch` was consumed: hand
            # the whole batch back rather than only the gap-fill tail.
            return batch if not worker_started else leftover
```

Bonus: returning `batch` makes `leftover is batch` true at [`:428`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L428), so the 0.05s yield applies and a deterministic failure stops spinning at full speed.

nemo_rl/algorithms/grpo.py
```diff
@@ -4226,23 +4280,14 @@ def async_grpo_train(
...
-                    max_age_steps=max_trajectory_age_steps,
-                )
-            )
-            print("✅ Replay buffer restored from checkpoint")
-        else:
-            print(
-                f"⚠️ No replay buffer checkpoint found at {replay_buffer_path}. "
-                "Starting with an empty replay buffer."
-            )
+        replay_buffer_restore_metadata = _maybe_restore_async_replay_buffer_checkpoint(
```
yfw, line 4287: **1 action item.** PR-introduced.

`replay_buffer_restore_metadata` is annotated `dict[str, int] | None` at [`:4280`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L4280), but this PR widened `load_from_path` to return `dict[str, Any]` and it now carries `RETAINED_TASK_INDICES_KEY -> list[int]`.

pyrefly can't catch this — `grpo.py` isn't in `pyrefly.toml` `project-includes` — so it will rot silently.

### AI-1
Change the annotation at `:4280` to `dict[str, Any] | None`.

nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -361,8 +362,17 @@ def load_from_path(
...
         max_age_steps: int | None = None,
-    ) -> dict[str, int]:
-        """Restore inside the actor and return only compact coordination metadata."""
+    ) -> dict[str, Any]:
+        """Restore inside the actor and return only compact coordination metadata.
+
+        Returns:
+            Mapping with ``num_trajectories`` (pre-filter count),
+            ``NEXT_NEMO_GYM_TASK_INDEX_KEY`` (one past the highest saved task
+            index, computed before age/step filtering so a used index is never
```
yfw: **1 action item.** PR-introduced.

This new docstring says the next index is computed "before age/step filtering so a used index is never re-issued" — but the frontier path in this same PR deliberately re-issues them: [`grpo.py:4317`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L4317) sets `next_nemo_gym_task_index = resume_base_ordinal`, rewinding the counter below `max(saved) + 1` so the covered window can be re-yielded.

That's correct behaviour (re-issuing the *same* ordinal for the *same* prompt is the point, and the only uniqueness check is per-repeated-batch in [`_build_task_index_map`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1177)), but the stated invariant now only holds on the legacy path.

### AI-1
Scope this docstring to the legacy path.

### Follow-up
`docs/design-docs/nemo-gym-integration.md:237` is made wrong by the same change — it says "**Gym** prompts receive a monotonic `_ng_task_index`. The counter is checkpointed, restored, and cross-checked ... so task identities are not reused after restart." Both halves are now false: stamping moved out of the `use_nemo_gym` guard into `_collection_loop`, so *every* prompt is stamped, and a frontier resume re-issues. That file isn't in this diff, but this PR is what invalidated it.

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -131,6 +158,36 @@ def __init__(
...
+        # False when ordinals cannot be trusted to equal the trained-prompt
+        # position (legacy-checkpoint resume, or unstampable batches); the
+        # checkpoint then falls back to saving the live dataloader cursor.
+        self._ordinals_frontier_aligned = ordinals_frontier_aligned
+        self._resume_frontier_ordinal = resume_frontier_ordinal
+        self._covered_task_indices: set[int] = set(resume_covered_task_indices or [])
+        self._skip_horizon = max(
+            self._covered_task_indices,
+            default=(resume_frontier_ordinal or 0) - 1,
+        )
```
yfw, line 244: **1 action item.** PR-introduced. Low severity — this is a latent defect, **not** a live bug.

`max(iterable, default=X)` uses `X` only when the iterable is *empty*, so with a non-empty retained set the frontier term is discarded entirely and `_skip_horizon == max(retained)`. The expression reads as "at least `frontier - 1`", but that isn't what it computes.

Three of us independently tried to reach a state where `max(retained) < frontier - 1` and **could not**: `_prepare_for_training_step` keeps only `target_weight_version >= current_step` ([`replay_buffer.py:497-501`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/replay_buffer.py#L497)) and each target owns a contiguous ordinal block, so retained ordinals are always `>= frontier`. The gap-fill accounting cancels exactly — `num_prompts_to_generate = min(batch, needed)` against `get_trajectories_needed = max(0, P - count)` means every prompt gained above `max(R)` is paid for by one lost below it, forcing `F <= max(R)` whenever `R` is non-empty. So I'm explicitly **not** claiming this duplicates training data today.

What makes it worth fixing anyway: the invariant that saves it is unwritten, cross-module, and enforced by nothing — and both new frontier tests hand-set `_skip_horizon` rather than exercising this expression, so nothing would catch it changing.

### AI-1
Take the max of both terms. Only ever raises the horizon, and the keep-predicate at [`:528-531`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L528) is unchanged, so it cannot over-drop:

```suggestion
        self._skip_horizon = max(
            [*self._covered_task_indices, (resume_frontier_ordinal or 0) - 1]
        )
```

(This form needs no `-1` sentinel and the list is never empty, so it can't regress into the same class of bug as the `default=` version. `max(max(covered, default=-1), (frontier or 0) - 1)` is equivalent if you prefer it.)

**Context — no action.** Two things this fix does *not* do:

1. It does **not** close the `consumed_samples` bug I flagged on [`grpo.py:5240`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5240). In that walkthrough the corrected expression evaluates to `max(15, 7) = 15` — unchanged — and the already-trained ordinal still passes the filter. These are separate defects that happen to live in the same function; fixing this one will not make that one go away.
2. `resume_frontier_ordinal or 0` treating `0` as falsy is harmless: `(0 or 0) - 1 == -1`, which is the correct horizon when nothing has been trained.

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -360,27 +449,152 @@ def _collection_loop(self):
...
+                        best is None or base_ordinal > best[0]
+                    ):
+                        best = (base_ordinal, state)
+            if best is not None:
+                return {
+                    "dataloader_state": best[1],
+                    "base_ordinal": best[0],
+                    "frontier_aligned": True,
+                }
+        return {
```
yfw, line 677: **1 action item.** PR-introduced.

This fallback returns `frontier_aligned: False` with **no log at all** when the ring no longer holds a snapshot at or below the frontier. Its sibling degradation path at [`_stamp_task_indices:465-470`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/async_utils/trajectory_collector.py#L465) does print a warning, so the two are inconsistent — and this one silently restores exactly the prompt-skipping behaviour the PR exists to remove.

It's also the safety valve that bounds the `max_generation_failures` bug I flagged on [`grpo.py:5240`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/algorithms/grpo.py#L5240), which makes it worth being able to see in a log.

### AI-1
Warn on the `best is None` branch, naming `frontier_ordinal` and the ring's current `[min_base, max_base]` so eviction is diagnosable.

### Follow-up
This branch is untested. `test_get_checkpoint_dataloader_state_selects_frontier_snapshot` only covers the `_ordinals_frontier_aligned = False` fallback, not the aligned-but-evicted one. A test that seeds `_dataloader_snapshots` with bases all *above* the requested frontier and asserts `frontier_aligned is False` passes on HEAD and pins this.

yfw (commented):
Multi-agent review of the frontier-aligned checkpointing work. 5 agents (RL/codebase, bug hunt, tests, design, adversarial verification); every finding below survived an adversarial pass, and 10 were cut as unfounded or disproportionate.

**Overall:** the diagnosis is careful and well evidenced, and the `_process_batch` return-the-remainder contract is the highest-value, lowest-risk change here — it turned "does the gap-fill tail survive?" into a pure return-value assertion. The frontier-alignment half is where the findings cluster.

**Two things worth deciding before this leaves draft:**

1. **`consumed_samples` is a counter being used as a stream position.** With `max_generation_failures > 0` — 13 shipped configs, including the SWE recipes that motivated this PR — a resume both re-trains already-trained prompts and permanently skips others. Details on `grpo.py:5240`.
2. **This may be two PRs.** The `_pending_batch` carry-over is small, safe, well-tested, and plausibly accounts for most of the 675 skips on its own (pre-PR, `_process_batch` discarded an entire dataloader batch whenever no target needed generation). Frontier alignment is the complex half and carries every bug below. Landing the first now and iterating on the second separately is a real option.

**CI:** `pre-commit` currently fails — 2 pyrefly errors, both PR-introduced. Verified fix in the first comment.

Also verified and explicitly **not** flagged, so you don't re-derive them: the `iter()` hoist is behaviour-preserving (and the pause reorder is an improvement); `dataloader_exhausted` is correct in both directions; the snapshot ring doesn't alias mutating state; `repeat_interleave` preserves stamped ordinals; `_ng_task_index` on native runs is inert across every consumer; `load_replay_buffer: false` is correct end-to-end and backward compatible; discarding the serialized pending batch on a frontier resume is correct; and the config/exemplar plumbing is complete (all 12 GRPO exemplars inherit via `defaults:`, and the reference config was updated in lockstep).

_Generated by Claude Code_

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -1019,8 +1254,14 @@ async def _enqueue_rollout_group(
...
             "timestamp": time.time(),
         }
-        if rollout_result.task_index is not None:
-            trajectory_group[NEMO_GYM_TASK_INDEX_KEY] = rollout_result.task_index
+        group_task_index = rollout_result.task_index
+        if group_task_index is None:
+            # Native groups carry their ordinal in the stamped extra_env_info
+            # rows; recording it on the group lets a frontier restore report
+            # which prompts the retained buffer already covers.
+            group_task_index = _group_task_index_from_batch(final_batch_cpu)
```
yfw: **1 action item.** PR-introduced. Low severity — this does **not** fire in any shipped configuration today; it's a defensive ask.

This recovers a native group's ordinal from `final_batch_cpu["extra_env_info"]` — the **post-rollout** rows. For multi-turn environments the rollout replaces those rows wholesale rather than updating them:

- [`rollouts.py:1315-1316`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/experience/rollouts.py#L1315) — `if not terminated and not truncated: ... current_extra_env_info = env_output.metadata[0]`
- [`rollouts.py:1013-1014`](https://github.com/NVIDIA-NeMo/RL/blob/d436cc7cecb1d629374fd2a2c7e758d1643f9007/nemo_rl/experience/rollouts.py#L1013) — the batched variant

An env that returns freshly-built state dicts (game state, tool scratchpads) drops `_ng_task_index`, `_group_task_index_from_batch` returns `None`, and the group is buffered without it — so it appears in neither `RETAINED_TASK_INDICES_KEY` nor `NEXT_NEMO_GYM_TASK_INDEX_KEY`. The retained set then under-reports and the whole buffered window is regenerated on resume. Wasted compute and step-composition skew, not corruption.

I enumerated every env in `nemo_rl/environments/` that passes `metadata=` and **none of them trigger this**: `math_environment` (`:519`, `:634`), `code_jaccard_environment` (`:246`), `vlm_environment` (`:208`), and `reward_model_environment` (`:271`) all return `terminateds` that are unconditionally true, so the guard above protects them. The one genuinely multi-turn env, `code_environment` (`:251`), is safe for a different reason: its worker returns the *same* dict objects it was passed, mutated in place (`:108`, `:129-135`, `:149`).

That last point is why I'm still raising it. The invariant holds **by accident** in the only env that exercises the path — nothing in the code says an environment must preserve keys it didn't create, and the next multi-turn env that builds its metadata fresh will silently regress the retained set with no test catching it.

### AI-1
Capture the ordinal from the **input** batch rather than re-deriving it from the rollout output. `RolloutGroupResult.task_index` already exists and is populated on the gym path — setting it on the native path too (read from `input_batch["extra_env_info"][i]` before the rollout) removes the dependency on env behaviour entirely. Failing that, stash the per-group ordinal in `_run_rollout_batch_worker` alongside `group_index`.

nemo_rl/algorithms/async_utils/trajectory_collector.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/grpo.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/grpo.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/async_utils/trajectory_collector.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

docs/guides/async-grpo.md
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

tests/unit/algorithms/test_async_utils.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

tests/unit/algorithms/test_grpo.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/async_utils/trajectory_collector.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/grpo.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/async_utils/replay_buffer.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/async_utils/trajectory_collector.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/async_utils/trajectory_collector.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/async_utils/trajectory_collector.py
yfw: fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813

nemo_rl/algorithms/grpo.py
yfw: partially fixed in a482c6c56dfdf11c8a57ee165f072a48c5bdf813 — the legacy-resume warning landed and correctly states the degradation applies to every descendant checkpoint.

Still remaining: the PR description itself. It reads "Runs resuming from a pre-upgrade checkpoint keep the old cursor semantics for that run", which is the wording this warning contradicts. The neighbouring bullet — "a permanently failed group shifts the stream, so one resume may re-train a substitute prompt once" — is also superseded now that the frontier is derived from the trained groups' own ordinals rather than `consumed_samples`.

yfw: Remaining item is now done — the PR description drops the "for that run" wording for "permanently for that checkpoint lineage", and the logging parenthetical is scoped to "a warning is logged at resume", which matches the code: the save-time warning at trajectory_collector.py:576 sits inside `if self._ordinals_frontier_aligned:`, so it covers ring eviction rather than the legacy-lineage case. Resolving.

terrykong (approved):

PR #3490, opened 2026-08-04
feat: colocated Megatron reshard

tdene:
# What does this PR do ?

This PR adds the ability to refit during colocated RL.

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #3605, opened 2026-08-12
fix: model save directly after validation

tdene:
# What does this PR do ?

`save_checkpoint` assumes the training model is on GPU. But when it is called during colocated, right after a validation run, the model is not on GPU.

This is probably best fixed with a bigger refactor, but this PR prevents crashes. Optimizations beyond that are not a priority due to how rare this issue is.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/models/policy/workers/megatron_policy_worker.py
+++ b/nemo_rl/models/policy/workers/megatron_policy_worker.py
@@ -2982,6 +2982,19 @@ class MegatronPolicyWorkerImpl(
                 ckpt_cfg=self.mcore_state.cfg.checkpoint,
                 blocking=True,
             )
+
+            # Onload model before saving it.
+            self.model = self.move_model(
+                self.model, "cuda", move_params=True, move_grads=False
+            )
+            if (
+                optimizer_path is not None
+                and self.optimizer is not None
+                and not self.optimizer_cpu_offload
+            ):
+                self.move_optimizer("cuda")
+            torch.cuda.synchronize()
+
             self.mcore_state.cfg.checkpoint.save = weights_path
 
             optimizer_to_save = None
--- a/tests/unit/models/policy/test_megatron_worker.py
+++ b/tests/unit/models/policy/test_megatron_worker.py
@@ -241,6 +241,55 @@ def test_megatron_offload_after_refit_finalizes_before_model_move(monkeypatch):
     assert events.index("finalize_async_save") < events.index("move_model")
 
 
+def test_megatron_save_checkpoint_onloads_model_before_save(monkeypatch):
+    """Params offloaded by colocated generation must be onloaded before the save walks them."""
+    import nemo_rl.models.policy.workers.megatron_policy_worker as worker_module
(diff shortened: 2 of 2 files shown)


cspades (approved):
> when it is called during colocated, right after a validation run, the model is not on GPU, because ...

The inference model is used for validation?

shanmugamr1992 (approved):

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -2982,6 +2982,19 @@ def save_checkpoint(
...
+            # Onload model before saving it.
+            self.model = self.move_model(
+                self.model, "cuda", move_params=True, move_grads=False
+            )
+            if (
+                optimizer_path is not None
+                and self.optimizer is not None
+                and not self.optimizer_cpu_offload
+            ):
+                self.move_optimizer("cuda")
```
yuki-97, line 2995: Residency of params / grad buffers / optimizer state at the moment `save_checkpoint` returns, before and after this PR:

| Path | Before: params / grads / optimizer | After |
|---|---|---|
| All other call sites reaching `MegatronPolicyWorkerImpl.save_checkpoint` -- each preceded by `prepare_for_training()` | GPU / GPU / GPU | unchanged |
| async + mcore-inf, non-colocated | GPU / **CPU** / **CPU** | GPU / **CPU** / GPU |
| async + mcore-inf, colocated reshard (#3490 / #2884) | **CPU** / **CPU** / **CPU** | GPU / **CPU** / GPU |

yuki-97: since this change is a no-op for other path (sync grpo or async grpo w/ other rollout backend), I'm ok with it.

but I want to double check is it intended for the optimizer to be kept on GPU after save ckpt for mcore-inf path?
especially for the colocated case (#3490 / #2884) which may affect the GPU memory that rollout part could use?

tdene: > but I want to double check is it intended for the optimizer to be kept on GPU after save ckpt for mcore-inf path?

Hah, this type of conversation is one reason why #3490 ended up being so big.

In #3490, we have an explicit argument that controls whether the optimizer moves between CPU and GPU, or stays permanently on GPU. We would like it to move, but @cspades reported that this is incredibly slow. We need both #3602 and NVIDIA/Megatron-LM#4241 before it becomes efficient to move it between CPU and GPU.

So, the way I see it, the optimizer in mcore-inf reshard is a huge mess that can only be untangled by #3490. Since #3490 is not merged yet, I do not want to address it here. I do not think we should let this prevent us from merging this PR.

If this PR, #3605, gets merged before #3490, then we can let #3490 fix this issue correctly. If #3490 gets merged first, then I will edit this PR. What do you think?

yuki-97: yea make sense, let's merge this first.

PR #3608, opened 2026-08-12
feat: mopd teacher topk

macandro96:
# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


nemo_rl/algorithms/loss/loss_functions.py
```diff
@@ -549,6 +586,94 @@ def __call__(
...
+                dtype=support_indices.dtype,
+            )
+            target_in_support = support_indices.eq(target_indices.unsqueeze(-1)).any(
+                dim=-1
+            )
+            teacher_target_logprobs = data["teacher_reference_logprobs"][:, 1:].to(
+                device=curr_logprobs.device,
+                dtype=curr_logprobs.dtype,
+            )
+            clip_loss = topk_reverse_kl_loss(
```
macandro96, line 661: **1 action item.**

**TL;DR — when `opd_topk` is set, the full PPO ratio/clip pipeline runs and its metrics are reported alongside `opd_topk_*`, but `clip_loss` is unconditionally overwritten by `topk_reverse_kl_loss(...)` so all of that PG-side work is wasted and the emitted `probs_ratio*` numbers are meaningless.**

[`loss_functions.py:556-588`](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L556) computes `loss1`, `loss2`, `clip_loss = torch.max(loss1, loss2)` for the PG branch. Then [`loss_functions.py:641`](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L641) unconditionally overwrites: `clip_loss = topk_reverse_kl_loss(...)`.

Two consequences:

1. **Wasted compute.** The `__init__` guard at [`loss_functions.py:252`](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L252) forces `disable_ppo_ratio=True`, so the else branch at [L577-579](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L577) executes: `ratios = curr_logprobs; ratios_clamped = curr_logprobs`, and `loss1/loss2/clip_loss` do a full advantages*curr_logprobs multiplication. Cheap-ish, but pure waste.

2. **Misleading metrics.** [L861-887](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L861) then reports `probs_ratio`, `probs_ratio_min/max`, `probs_ratio_clamped*` — but `ratios` here equals `curr_logprobs` (raw logprobs, not exp'd ratios), so these numbers are log-probabilities being reported as if they were IS ratios. A user reading the metrics dashboard alongside `opd_topk_head_loss` etc. sees `probs_ratio ≈ -3.4` next to a healthy `opd_topk_student_mass` and has no way to know the ratio value is nonsense.

**AI-1:** short-circuit the PG branch when `self.opd_topk is not None` — skip the `if force_on_policy_ratio / elif not disable_ppo_ratio / else` block and the `use_cispo` / `torch.max` step entirely, jumping straight to the `opd_topk` computation. Then drop the `probs_ratio*` and `sampling_importance_ratio` keys (and their `metric_normalizations` entries added in `__init__`) from the returned dict on this path, so the metrics dictionary contains only `opd_topk_*` + `loss` + `kl_penalty` + valid samples.

One way to structure it (illustrative — several valid factorings):

```python
if self.opd_topk is not None:
    # ... existing validation / clip_loss = topk_reverse_kl_loss(...) block ...
    # skip PG ratios, dual clip, TIS, CISPO entirely below
else:
    if self.force_on_policy_ratio: ...
    elif not self.disable_ppo_ratio: ...
    else: ...
    # if use_cispo / else: torch.max ...
```

nemo_rl/algorithms/loss/loss_functions.py
```diff
@@ -225,13 +228,37 @@ class ClippedPGLossFn(LossFunction):
...
+            if not cfg.disable_ppo_ratio:
+                raise ValueError("Top-k OPD requires loss_fn.disable_ppo_ratio=true.")
+            if cfg.use_cispo:
+                raise ValueError("Top-k OPD is incompatible with CISPO.")
+            if cfg.ratio_clip_c is not None:
+                raise ValueError("Top-k OPD is incompatible with dual PPO clipping.")
+            if cfg.sequence_level_importance_ratios or not cfg.token_level_loss:
+                raise ValueError(
+                    "Top-k OPD requires token-level loss and importance ratios."
+                )
```
macandro96, line 263: **1 action item.**

**TL;DR — the `opd_topk` init guards forbid CISPO, dual-clip, sequence-level IS, fused-linear-logprobs, and non-token-level loss, but NOT `use_importance_sampling_correction`, `truncated_importance_sampling_type`, or `positive_example_nll_weight` — any of which silently modifies the reverse-KL loss without warning.**

At [`loss_functions.py:245-261`](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L245) the opd_topk validation block guards several incompatibilities. But downstream:

- [L797-800](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L797): if `use_importance_sampling_correction=True`, `importance_weights_to_use = actor_importance_weights = exp(prev - gen)`. Then [L803-807](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L803) multiplies token-level IS weights into `clip_loss` — which is now the top-k reverse-KL estimator. Multiplying IS weights into a KL distillation loss is not part of the OPD formulation, and there is no warning.
- [L720-793](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L720): if `truncated_importance_sampling_type` is set, `actor_importance_weights` gets tis / icepop / seq-mask filtering, and L318 asserts `use_importance_sampling_correction` must be True — so TIS forces the previous silent-multiply.
- [L848-857](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/loss/loss_functions.py#L848): if `positive_example_nll_weight > 0`, an NLL term is added to the loss, mixing PG-VAPO with reverse-KL distillation.

**AI-1:** extend the `if opd_topk is not None:` block in `__init__` to raise on these three settings. Suggested block:

```suggestion
        if opd_topk is not None:
            if opd_topk < 1:
                raise ValueError(f"opd_topk must be at least 1, got {opd_topk}.")
            if use_fused_linear_logprobs:
                raise ValueError(
                    "Top-k OPD is incompatible with fused linear logprobs."
                )
            if not cfg.disable_ppo_ratio:
                raise ValueError("Top-k OPD requires loss_fn.disable_ppo_ratio=true.")
            if cfg.use_cispo:
                raise ValueError("Top-k OPD is incompatible with CISPO.")
            if cfg.ratio_clip_c is not None:
                raise ValueError("Top-k OPD is incompatible with dual PPO clipping.")
            if cfg.sequence_level_importance_ratios or not cfg.token_level_loss:
                raise ValueError(
                    "Top-k OPD requires token-level loss and importance ratios."
                )
            if cfg.use_importance_sampling_correction:
                raise ValueError(
                    "Top-k OPD is incompatible with use_importance_sampling_correction=True "
                    "(would multiply actor IS weights into the reverse-KL estimator)."
                )
            if cfg.truncated_importance_sampling_type is not None:
                raise ValueError(
                    "Top-k OPD is incompatible with truncated_importance_sampling_type."
                )
            if cfg.positive_example_nll_weight != 0:
                raise ValueError(
                    "Top-k OPD is incompatible with a non-zero positive_example_nll_weight."
                )
```

Worth confirming which of these are `_validate_topk_opd_config` already catching in `grpo.py`; if all three are also gated there, this is defense-in-depth at the loss layer (still worth doing — anyone constructing `ClippedPGLossFn` outside grpo won't get the algo-level check).

nemo_rl/algorithms/opd.py
```diff
@@ -112,6 +159,107 @@ def is_non_colocated_teachers_enabled(master_config: Any) -> bool:
...
+
+    The exact reverse-KL contribution is used for a selected top-k support.
+    When the sampled target is outside that support, its score-function
+    contribution estimates the remaining vocabulary tail without transferring
+    full-vocabulary teacher logits.
+
+    All log-probabilities must be normalized against their model's full
+    vocabulary. Support selection and teacher values are treated as constants;
+    gradients flow only through the current student log-probabilities.
+    """
```
macandro96, line 244: **No action needed in code — 1 docstring update.**

**TL;DR — the docstring says "gradients flow only through the current student log-probabilities" but does not say the estimator is only unbiased under an on-policy assumption; under async replay with a stale sampling policy, the tail term inherits the same off-policy skew as async GRPO.**

At [`opd.py:256`](https://github.com/NVIDIA-NeMo/RL/blob/b692a79447b95df4e5d13803e90cbb4625ac0b09/nemo_rl/algorithms/opd.py#L256), `tail_advantage = (teacher_target_logprobs - student_target_logprobs).detach()` uses the *current-model* student logprob at the sampled token. The score-function identity `∇θ E_πs[log πs − log πt] = E_πs[(1 + log πs − log πt) · ∇θ log πs]` assumes the expectation is over the *current* πs. Under async / replay, tokens were sampled from a stale πs and no importance ratio corrects the gap — the code correctly requires `disable_ppo_ratio=true` to avoid double-counting a PG ratio, but the async off-policy skew is not something the estimator itself corrects. This is the same trade-off async GRPO already accepts, but the docstring reads as if the estimator were unconditionally safe.

**AI-1:** add a `Note` paragraph to the docstring making the on-policy assumption explicit and pointing to ICE-POP / small-lag as the safety condition:

```suggestion
    All log-probabilities must be normalized against their model's full
    vocabulary. Support selection and teacher values are treated as constants;
    gradients flow only through the current student log-probabilities.

    Note:
        The tail term is a score-function estimator built on the current-model
        student logprob at the sampled target token. Under async replay the token
        was sampled from a stale student policy and no importance ratio corrects
        the sampling-distribution mismatch — the estimator inherits the same
        off-policy bias as async GRPO, and correctness relies on either fully
        on-policy training or a small policy lag (e.g. bounded via ICE-POP).
        The loss layer enforces ``disable_ppo_ratio=True`` for this reason, but
        does not compensate for staleness beyond that.
    """
```

macandro96 (commented):
Review focused on core correctness in the new `opd_topk` loss path. Three findings below. Broader design/docstring/PR-body notes intentionally deferred to a follow-up review.

_Generated by Claude Code_

nemo_rl/algorithms/opd.py
macandro96: addressed in [1cb3955](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/1cb39558976ec285119bdb7a9d164fd811549de8)

nemo_rl/algorithms/loss/loss_functions.py
macandro96: addressed in [1cb3955](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/1cb39558976ec285119bdb7a9d164fd811549de8)

nemo_rl/algorithms/loss/loss_functions.py
macandro96: addressed in [1cb3955](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/1cb39558976ec285119bdb7a9d164fd811549de8)

docs/about/algorithms/mopd.md
```diff
@@ -141,6 +213,39 @@ The reference recipe self-distills `Qwen/Qwen3-1.7B` (student == teacher) across
...
 correctness smoke test, not a demonstration of distillation gains.
 
+To exercise the student-top-k path with that recipe, disable its inherited
+sequence packing and set the support size explicitly:
+
+```sh
+uv run examples/nemo_gym/run_grpo_nemo_gym.py \
+  --config examples/configs/recipes/llm/mopd-qwen3-1.7b-3n8g-megatron-pack.yaml \
+  policy.sequence_packing.enabled=false \
+  on_policy_distillation.student_topk=64 \
```
yfw, line 243: [`docs/about/algorithms/mopd.md:220-226`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/docs/about/algorithms/mopd.md#L220)

**1 action item.**

**TL;DR — both new top-k commands abort in `setup()` on the recipe they name, because that recipe enables the two ICE-POP settings this PR's own guards reject.**

PR-introduced. [`mopd-qwen3-1.7b-3n8g-megatron-pack.yaml:22`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/examples/configs/recipes/llm/mopd-qwen3-1.7b-3n8g-megatron-pack.yaml#L22) sets `use_importance_sampling_correction: true` and [`:25`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/examples/configs/recipes/llm/mopd-qwen3-1.7b-3n8g-megatron-pack.yaml#L25) sets `truncated_importance_sampling_type: icepop`. Neither command overrides them, so `ClippedPGLossFn.__init__` raises at [`loss_functions.py:263`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/loss/loss_functions.py#L263):

```
ValueError: Top-k OPD is incompatible with use_importance_sampling_correction=True.
```

and after overriding only that one, [`:267`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/loss/loss_functions.py#L267) raises `Top-k OPD is incompatible with truncated_importance_sampling_type.` Both overrides are required.

I verified they are also **sufficient** — every other condition in `ClippedPGLossFn.__init__` and [`_validate_topk_opd_config`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/grpo.py#L451) already resolves green through the `defaults: ../../grpo_math_1B.yaml` chain (`ratio_clip_c: null`, `token_level_loss: true`, `use_cispo: false`, `positive_example_nll_weight: 0.0`, `use_fused_linear_logprobs: false`, `top_p: 1.0`/`top_k: null`, plus the recipe's own `async_grpo.enabled`, `adv_estimator.name: opd`, `reference_policy_kl_penalty: 0.0`, `disable_ppo_ratio: true`, `megatron_cfg.enabled`). With these two added, both commands clear every guard.

**Action:** add the two overrides to both command blocks (the second is at `:232-236`, commented separately). It would also help to list these two guards in the constraints NOTE at [`:152-160`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/docs/about/algorithms/mopd.md#L152) — a reader following the page top-to-bottom hits the `loss_fn` block at `:68-77`, which correctly prescribes ICE-POP for base MOPD, and has no way to know top-k rejects it.

```suggestion
  policy.sequence_packing.enabled=false \
  on_policy_distillation.student_topk=64 \
  loss_fn.use_importance_sampling_correction=false \
  loss_fn.truncated_importance_sampling_type=null \
```

docs/about/algorithms/mopd.md
```diff
@@ -141,6 +213,39 @@ The reference recipe self-distills `Qwen/Qwen3-1.7B` (student == teacher) across
...
+  data.validation.data_path=/path/to/val.jsonl
+```
+
+To run the teacher-selected variant with the recipe's inherited sequence
+packing, keep packing enabled and select the teacher support explicitly:
+
+```sh
+uv run examples/nemo_gym/run_grpo_nemo_gym.py \
+  --config examples/configs/recipes/llm/mopd-qwen3-1.7b-3n8g-megatron-pack.yaml \
+  on_policy_distillation.teacher_topk=64 \
```
yfw, line 256: [`docs/about/algorithms/mopd.md:232-236`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/docs/about/algorithms/mopd.md#L232)

**1 action item.** Same two overrides are needed here — see the comment on the student-top-k command above for the reproduction and why both are required.

```suggestion
  on_policy_distillation.teacher_topk=64 \
  loss_fn.use_importance_sampling_correction=false \
  loss_fn.truncated_importance_sampling_type=null \
```

examples/nemo_gym/nemotron-3-ultra/mopd.yaml
```diff
@@ -838,12 +838,12 @@ on_policy_distillation:
     default_teacher_cfg:
       tensor_model_parallel_size: 8
       pipeline_model_parallel_size: 1
+      expert_tensor_parallel_size: 1
       expert_model_parallel_size: 16
       context_parallel_size: 2
       num_nodes: 4
```
yfw, line 843: [`examples/nemo_gym/nemotron-3-ultra/mopd.yaml:838-847`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/examples/nemo_gym/nemotron-3-ultra/mopd.yaml#L838)

**1 action item.**

**TL;DR — this teacher previously got single-NVLink-domain placement derived from `cluster.segment_size`; with segment constraints now opt-in per teacher, this recipe opts into nothing, so its 4 teacher nodes may straddle NVLink domains and the teacher's TP/CP collectives can cross the slower inter-domain fabric.**

PR-introduced, and the refactor itself is right — this is just a recipe that needs to opt back in.

The mechanism, briefly, because it moved:
1. Teachers are placed by [`reserve_teacher_clusters`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/opd.py#L423), which can pin a teacher inside one NVLink domain via `prepare_segment_topology`.
2. **Before:** that was derived from the cluster — `if segment_size is not None: teacher_segment_size = num_nodes`. With [`cluster.segment_size: 16`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/examples/nemo_gym/nemotron-3-ultra/mopd.yaml#L40) this called `prepare_segment_topology(4, 4, ...)` and pinned the teacher.
3. **Now:** [`opd.py:464`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/opd.py#L464) reads `teacher_config.segment_size`, which this recipe never sets, so `node_resource_constraints` stays `None`.
4. What is lost is specifically the placement-group constraint that *forces* the 4 bundles into one domain. Bundle *ordering* stays topology-aware via `_sort_bundle_indices_by_topology`, so ranks are still laid out sensibly — it is domain pinning that is gone, not topology awareness.

This is not a broken config now failing earlier. Training allocates with `prepare_segment_topology(segment_size=16, ...)` and `select_segment_nodes` takes only **complete** 16-node segments, so the remainder is whole untouched domains — a 4-node single-domain request against that could not fail. The recipe worked; the constraint is simply absent now.

That the constraint was load-bearing is visible in the recipe's own comments: [`:186`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/examples/nemo_gym/nemotron-3-ultra/mopd.yaml#L186) *"TP=8 spans 2 GB200 nodes (4 GPUs each) via intra-rack NVLink"* and [`:190`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/examples/nemo_gym/nemotron-3-ultra/mopd.yaml#L190) *"All-to-all spans 64 GPUs (16 nodes), fits within one NVLink domain"* — with `gpus_per_node: 4`, `segment_size: 16` is 16 nodes = 64 GPUs = one domain, and the teacher's 4 nodes are 16 GPUs inside it.

**Action:** set the new opt-in key on `default_teacher_cfg`. `segment_size: 4` reproduces the previous call exactly (`prepare_segment_topology(4, 4, ...)`), and `4 % 4 == 0` satisfies `select_segment_nodes`. I checked the alternative of clamping (`min(num_nodes, cluster_segment_size)`) inside `reserve_teacher_clusters` and it is worse — `select_segment_nodes` requires `num_nodes % segment_size == 0`, so a 24-node teacher on a 16-node domain would raise. Per-recipe opt-in is the safer design; this recipe just needs to use it.

```suggestion
      num_nodes: 4
      segment_size: 4
```

nemo_rl/models/megatron/train.py
```diff
@@ -816,15 +905,106 @@ def processor_fn_inner(output_tensor):
...
+                temperature=1.0,
+                tp_group=active_tp_grp,
+            )
+            support_logprobs = gather_logits_at_global_indices(
+                local_logprobs,
+                support_indices.to(device=output_tensor.device, dtype=torch.long),
+                tp_group=active_tp_grp,
+                vocab_start_index=vocab_start_index,
+                vocab_end_index=vocab_start_index + vocab_shard_size,
+            )
```
yfw: [`nemo_rl/models/megatron/train.py:995-1006`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/models/megatron/train.py#L995)

**TL;DR — this PR adds `gather_vocab_parallel_logprobs_at_indices`, then uses it at one of the three sites that want it; the other two do an unchunked double-normalizing version by hand.**

PR-introduced; all three sites are `+` lines.

- **Site A, correct** — [`loss/utils.py:450`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/loss/utils.py#L450) concatenates the target index onto the support and makes **one** chunked call.
- **Site B** — [`loss/utils.py:156`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/loss/utils.py#L156), unpacked `OPD_TOPK`: `vocab_parallel_log_softmax` over the full `[B, S, V_local]`, then `gather_logits_at_global_indices`, then a **second** independent full normalization inside `get_next_token_logprobs_from_logits`. `chunk_size` is in scope (it is passed to that second call seven lines down) but unused for the support gather.
- **Site C, this block** — same unchunked pair, and `logprob_chunk_size` is read by every sibling post-processor in this file but not here.

Site C is the one with a measurable cost: the teacher is inference-only, which is exactly the case the helper's own docstring says chunking bounds ([`model_utils.py:1467`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/distributed/model_utils.py#L1467) distinguishes it from the autograd case). Reachable today via this PR's own documented command, which sets `student_topk=64` on a recipe with `logprob_chunk_size: 2048`; the teacher config is a `deepcopy` of the policy config, so it carries that value and this class ignores it. At that recipe's shape (S=32768, V=151936, TP=2) the fp32 buffer is ~9.27 GiB versus ~0.58 GiB chunked. The swap was verified bit-exact against the current body (max abs diff 0.0 at `chunk_size` in {None, 1, 2, 3, 7, 100}).

Site B is under autograd, so chunking does **not** bound peak memory there — the win is dropping one of two full-vocabulary log-softmaxes.

**Action:** at both sites replace the `vocab_parallel_log_softmax` + `gather_logits_at_global_indices` pair with a single call, concatenating the target index onto the support the way site A already does:

```python
support_logprobs = gather_vocab_parallel_logprobs_at_indices(
    output_tensor,
    support_indices.to(device=output_tensor.device, dtype=torch.long),
    tp_group=active_tp_grp,
    vocab_start_index=vocab_start_index,
    vocab_end_index=vocab_start_index + vocab_shard_size,
    chunk_size=self.cfg.get("logprob_chunk_size"),
)
```

Deliberately not a `suggestion` block: this leaves `gather_logits_at_global_indices` and `vocab_parallel_log_softmax` with no remaining uses, so please also drop the now-unused imports at `train.py:56,58` (and `loss/utils.py:32,36` for site B) — otherwise `ruff check` goes red on F401.

### Follow-up
The chunking contract these two call sites would then depend on is currently unpinned by any test — see the comment on `tests/unit/distributed/test_model_utils.py:46`.

tests/unit/distributed/test_model_utils.py
```diff
@@ -42,6 +43,75 @@
...
+        vocab_start_index=0,
+        vocab_end_index=4,
+        chunk_size=2,
+    )
+    expected = logits.log_softmax(dim=-1).gather(-1, indices)
+
+    torch.testing.assert_close(actual, expected)
+    actual.sum().backward()
+    assert logits.grad is not None
+    assert torch.isfinite(logits.grad).all()
```
yfw: [`tests/unit/distributed/test_model_utils.py:46-73`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/tests/unit/distributed/test_model_utils.py#L46)

**1 action item.**

PR-introduced. This test passes `chunk_size=2` and its value assertion does verify that the chunked path reassembles correctly — that part is worth keeping. What it cannot do is fail if chunking stops happening: [`vocab_parallel_log_softmax`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/distributed/model_utils.py#L1441) reduces over the vocabulary dim while chunking slices the sequence dim, so the result is chunk-invariant **by construction** and no value assertion anywhere can observe the difference.

Concretely: setting `effective_chunk_size = seq_len` unconditionally at [`model_utils.py:1509`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/distributed/model_utils.py#L1509) — i.e. silently ignoring the caller's `chunk_size` — leaves this file fully green (`2 passed, 34 skipped`). The docstring at [`:1467`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/distributed/model_utils.py#L1467) describes chunking as bounding the live fp32 vocabulary working set during inference, so that is a documented memory bound that a future refactor could delete with no test failure.

**Action:** spy on the log-softmax call and assert the per-chunk sequence lengths. Verified against this file: passes as written, and goes red under the mutation above.

```suggestion
def test_gather_vocab_parallel_logprobs_at_indices_chunks_and_keeps_gradients(
    monkeypatch,
):
    from nemo_rl.distributed.model_utils import vocab_parallel_log_softmax

    chunk_seq_lens = []

    def _recording_log_softmax(logits, **kwargs):
        chunk_seq_lens.append(logits.shape[1])
        return vocab_parallel_log_softmax(logits, **kwargs)

    monkeypatch.setattr(
        "nemo_rl.distributed.model_utils.vocab_parallel_log_softmax",
        _recording_log_softmax,
    )

    logits = torch.tensor(
        [
            [
                [1.0, 2.0, 3.0, 4.0],
                [4.0, 3.0, 2.0, 1.0],
                [0.0, 1.0, 0.0, 1.0],
                [2.0, -1.0, 0.0, 3.0],
                [-2.0, 0.0, 2.0, 1.0],
            ]
        ],
        requires_grad=True,
    )
    indices = torch.tensor([[[3, 1], [0, 2], [1, 3], [3, 0], [2, 1]]], dtype=torch.long)

    actual = gather_vocab_parallel_logprobs_at_indices(
        logits,
        indices,
        vocab_start_index=0,
        vocab_end_index=4,
        chunk_size=2,
    )
    expected = logits.log_softmax(dim=-1).gather(-1, indices)

    assert chunk_seq_lens == [2, 2, 1]
    torch.testing.assert_close(actual, expected)
    actual.sum().backward()
    assert logits.grad is not None
    assert torch.isfinite(logits.grad).all()
```

nemo_rl/algorithms/grpo.py
```diff
@@ -4234,10 +4352,13 @@ def async_grpo_train(
...
-    assert master_config.loss_fn.use_importance_sampling_correction, (
-        "Importance sampling correction must be enabled for async GRPO for good convergence due to off-policy samples!"
-    )
+    if not master_config.loss_fn.use_importance_sampling_correction:
+        warnings.warn(
+            "Importance sampling correction is disabled for async GRPO; "
+            "off-policy samples may affect convergence."
+        )
     max_generation_failures = master_config.grpo.async_grpo.max_generation_failures
+    opd_topk, topk_source = opd_module.get_topk_support_config(master_config)
```
yfw: [`nemo_rl/algorithms/grpo.py:4355-4359`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/grpo.py#L4355)

**1 action item.**

PR-introduced. The hard startup assert that required importance-sampling correction for **every** async GRPO run is now an unconditional `warnings.warn`. The relaxation is genuinely necessary — top-k OPD requires `async_grpo.enabled=true` and [`loss_functions.py:263`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/loss/loss_functions.py#L263) raises on `use_importance_sampling_correction=True`, so a top-k run could not otherwise start. It just is not scoped to that case.

I resolved the `defaults:` chain for every YAML under `examples/`: **52 async-enabled recipes, all 52 explicitly set `use_importance_sampling_correction: true`**, so scoping the exemption blocks nothing that ships. What it restores is the startup failure for a hand-written async config that omits IS correction, which today trains with uncorrected off-policy bias behind a warning in the Ray worker logs.

**Action:** gate on `opd_topk`, which is already computed two lines below — hence the suggestion spans through `:4361` to move that assignment above the guard.

```suggestion
    opd_topk, topk_source = opd_module.get_topk_support_config(master_config)
    if opd_topk is None:
        assert master_config.loss_fn.use_importance_sampling_correction, (
            "Importance sampling correction must be enabled for async GRPO for good convergence due to off-policy samples!"
        )
    elif not master_config.loss_fn.use_importance_sampling_correction:
        warnings.warn(
            "Importance sampling correction is disabled for async GRPO; "
            "off-policy samples may affect convergence."
        )
    max_generation_failures = master_config.grpo.async_grpo.max_generation_failures
```

yfw (commented):
Review of the top-k reverse-KL OPD estimator. The feature is well defended — `_validate_topk_opd_config` covers eight setup-time conditions, `ClippedPGLossFn.__init__` eight more, and the new estimator's tail term is pinned by tests that go red when the score-function `+1` or the in-support mask is flipped. Six inline comments below; the rest is context.

**Please rebase.** The PR currently conflicts with `main` — 7 conflicting hunks across `nemo_rl/algorithms/opd.py`, `nemo_rl/models/megatron/train.py`, `nemo_rl/models/policy/teacher_worker_group.py`, `tests/unit/algorithms/test_opd.py` and `tests/unit/models/megatron/test_train.py`.

**`pre-commit` is red on this branch.** `ruff-format` reformats `tests/unit/algorithms/test_grpo.py` (~L4176) and `tests/unit/algorithms/test_opd.py` (~L876) — running the hook fixes those automatically. `pyrefly` reports 4 `no-matching-overload` errors on `pydantic.fields.Field` at [`nemo_rl/algorithms/opd.py:67, 89, 111, 112`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/opd.py#L67); rewriting each as `Annotated[Optional[int], Field(ge=1)] = None` takes it from 4 errors to 0 with identical pydantic semantics (default `None`, `ge=1` still enforced), and matches the existing pattern at [`vllm/config.py:150`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/models/generation/vllm/config.py#L150).

**Two of the comments below are about non-top-k behavior**, which is worth calling out since the estimator itself is cleanly opt-in: the teacher segment-placement change affects the existing non-colocated-teacher path (including the shipped ultra MOPD recipe, which never enables top-k), and the async importance-sampling relaxation at `grpo.py:4355` applies to every async GRPO run. Both are small fixes; neither is about the new estimator.

**Doc terminology.** [`mopd.md:246`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/docs/about/algorithms/mopd.md#L246) refers to the baseline as "full-vocabulary MOPD", but this page states the opposite at [`:29-31`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/docs/about/algorithms/mopd.md#L29) — MOPD "needs only the teacher's log-probability for the *sampled* token rather than the full vocabulary distribution" — and calls it "the sampled-token MOPD loss" at [`:41`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/docs/about/algorithms/mopd.md#L41). Suggest reusing "sampled-token MOPD" there so the page has one name for the baseline. Same sentence has a second issue: it asks the reader to compare *loss* convergence, but the top-k `loss` is a gradient surrogate on a different scale and sign from the sampled-token loss, so the comparison isn't meaningful as written — reward and `opd_topk_head_loss` (which is exact) would be.

Separately, the estimator's `Note` at [`opd.py:226`](https://github.com/NVIDIA-NeMo/RL/blob/753e1d4aa40e36981626dbf9455290a5aac72bec/nemo_rl/algorithms/opd.py#L226) names on-policy training and an ICE-POP-bounded lag as its two bias mitigations, but top-k requires async GRPO and rejects `truncated_importance_sampling_type` — so both named mitigations are unavailable in the only supported mode, leaving `grpo.async_grpo.max_trajectory_age_steps` as the remaining lever.

**Minor.** All four changed lines in `examples/nemo_gym/nemotron-3-ultra/mopd.yaml` are no-ops — `expert_tensor_parallel_size: 1` matches both the new pydantic default and the policy's own value at `:188`; `precision: bf16` was already normalized to `bfloat16` by the new `BeforeValidator`; and `moe_flex_dispatcher_backend` is only consulted under `moe_token_dispatcher_type: flex` while that teacher uses `alltoall`. The one substantive change that file needs is the `segment_size` comment above.

Also: the five new `"<key>" in metrics` guards (`grpo.py:3896, 3932, 5705`, `grpo_sync.py:1259, 1293`) appear to protect against a state the PR no longer produces — the `opd_topk` return dict still contains `token_mult_prob_error` and `gen_kl_error`, both computed unconditionally, and the keys it does drop are consumed only by key-driven loops. Which `KeyError` motivated them? `grpo_train_sync` in particular cannot reach top-k at all, since it requires async.

Finally, the PR description is still the unfilled template. A sentence on what top-k OPD does and how it was validated would help reviewers a lot on a change this size.

_Generated by Claude Code_

docs/about/algorithms/mopd.md
macandro96: addressed in [193a9f4](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/193a9f4bf3f604a84242c67fb8154b5110116e4f)

docs/about/algorithms/mopd.md
macandro96: addressed in [193a9f4](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/193a9f4bf3f604a84242c67fb8154b5110116e4f)

examples/nemo_gym/nemotron-3-ultra/mopd.yaml
macandro96: addressed in [193a9f4](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/193a9f4bf3f604a84242c67fb8154b5110116e4f)

macandro96:
/ok to test [193a9f4](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/193a9f4bf3f604a84242c67fb8154b5110116e4f)

nemo_rl/models/megatron/train.py
macandro96: addressed in [193a9f4](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/193a9f4bf3f604a84242c67fb8154b5110116e4f)

tests/unit/distributed/test_model_utils.py
macandro96: addressed in [193a9f4](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/193a9f4bf3f604a84242c67fb8154b5110116e4f)

nemo_rl/algorithms/grpo.py
macandro96: addressed in [193a9f4](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/193a9f4bf3f604a84242c67fb8154b5110116e4f)

macandro96:
/ok to test [8b6bb82](https://github.com/NVIDIA-NeMo/RL/pull/3608/commits/8b6bb8260f53fcd900e66e3daf1400bfc5b25caa)

PR #3613, opened 2026-08-12
feat(sglang): engine fault tolerance for rollouts

Kh4L:
Original work: #3187 by @xiuhu17. The original commit and authorship are preserved, including its sign-off.

Adds opt-in SGLang engine fault tolerance. A serving health monitor detects rollout failures; before each refit, a separate actor/process liveness probe also detects deaths while the monitor is paused or the engine is offloaded. Recovery precedes the first weights-onload RPC, deregisters failed router workers, and replaces the whole logical engine group before reconnecting the weight-update transport. A per-engine restart budget bounds repeated failures. Child-process guards prevent an abruptly killed actor from leaving a GPU server behind.

Generation continues on healthy survivors within the existing nonstreaming retry budget. Verified `sglang-router==0.3.2` can repeatedly choose a dead cache-affine worker until its circuit opens; NeMo-RL's existing outer HTTP retries allow the single-worker failure case to reach a survivor. Started streams, retry exhaustion, trainer death, and death during an active weight transfer are not guaranteed recovery cases. Engine replacement takes place at the next refit so fresh policy weights are applied before rollout resumes.

Validation at final head `9399a3223`: HSG CPU job `7014630` passed 107 tests, including two real-Ray helper-serialization cases; pyrefly reported 0 errors. GPU job `7014647` passed all six real-engine chaos cases, including offloaded server/actor death and survival of an observed in-flight SIGKILL. All four interrupted-batch requests completed with 2,048 tokens. The full SGLang functional lane includes the new `EXPECT=survival` chaos harness; this local harness is generation/recovery, not a full GRPO training run. A maintainer must run `/ok to test 9399a32230cacb855de3e5cb6a8153ec3f963594` after the updated head is published.

PR #3612, opened 2026-08-12
feat(sglang): megatron backend weight refit for sglang rollouts

Kh4L:
Original work: #3190 by @xiuhu17. This is a re-cut opened at his request so the change can complete review. His commits retain their original authorship and `Signed-off-by` trailers.

## Summary

- Routes SGLang weight refit through the weight-synchronizer interface.
- Uses Ray CUDA IPC for colocated Megatron/FSDP2 policies.
- Uses a trainer-rank-0 NCCL broadcast for non-colocated Megatron policies.
- Rejects unsupported non-colocated FSDP2/SGLang during setup; support is tracked in #3745.
- Wraps bucket transfer in SGLang `begin_weight_update` / `end_weight_update` sessions so derived quantized layouts are finalized once after the last bucket.
- Removes the legacy HTTP transport. It was live on `main`; this PR supersedes it with the IPC/NCCL transports.

## Runtime dependencies

- SGLang: `0.5.17.dev30+g3003d70f6` from pinned commit `3003d70f680d41c59d1b7acbf65cb47795dfd19e`
- SGLang kernel: `0.4.5`
- FlashInfer Python/cubin/CUDA-13 JIT cache: `0.6.14`
- Transformers in the SGLang environment: `5.12.1`

The pinned SGLang revision contains [sglang#27948](https://github.com/sgl-project/sglang/pull/27948), replacing the local TMS CUDA-graph patch removed here. The matching FlashInfer cubin and JIT-cache wheels are explicit so engine startup does not fall back to compiling kernels.

Dependency versions are isolated per extra: only SGLang uses Transformers `5.12.1` and TileLang `0.1.11`; Megatron/vLLM stay on Transformers `5.8.1`, and base/non-SGLang extras stay on TileLang `0.1.9` where the package's x86_64 marker applies. This PR no longer depends on #3498.

## Review fixes included

- Lazily initializes backend-neutral refit transport state and reuses the shared Megatron HF iterator/bucketer.
- Guarantees generation/policy phase restoration when pause, onload, or transfer fails.
- Declares the four SGLang policy methods on `ColocatablePolicyInterface`.
- Preserves SGLang's engine-side `message` error text.
- Cleans up stale Gloo subgroups and the Ray lock actor, bounds lock acquisition, and routes pause/continue through the common HTTP helper.
- Centralizes `NCCL_CUMEM_ENABLE` selection before the first NCCL process group.
- Removes stale test deselections and moves the ordering AST check to the CPU lane.

## Validation

[...]

PR #3614, opened 2026-08-12
feat(sglang): mxfp8 quantized rollout and refit

Kh4L:
**Original work: #3188 by @xiuhu17.** Re-cut at his request so it can be carried through review on his behalf. His commit is preserved with its original authorship and `Signed-off-by`.

**2 of 3 remaining in the stack.** `#3613 (fault tolerance) → this → #3615 (nvfp4)`. #3612 has merged. GitHub PRs from a fork must target `main`, so this diff also contains #3613. Review the thirteen commits after #3613.

This PR cannot merge before #3613 because it contains #3613 commits.

## What it adds

MXFP8 quantized rollouts with Megatron weight refit: quantization setup and conversion, cache-aware checkpoint handling, and weight-group iteration that carries derived scales through each refit. MoE router weights remain in BF16.

Startup requires a local checkpoint produced by the explicit offline GPU converter, keeping FlashInfer conversion out of the GPU-less Ray driver while preserving deterministic cache reuse.

## Verification

The repository-pinned Ruff, Ruff-format, and TOML-format hooks pass across the refreshed branch. Focused shared/MXFP8 checkpoint tests pass, including all three SGLang startup paths, Nemotron tensor names, complete scale validation, and deterministic cache reuse.

HSG job `6588661` passed on 3 nodes / 12 GPUs using `batch`, `short`, and `coreai_dlalgo_nemorl`. It completed four real Nemotron-3-Nano Megatron optimizer steps and four broadcast SGLang MXFP8 refits with decode CUDA graphs enabled. The job hash-guarded and overlaid the sole `bfa0eaf3..54e9be45` source delta, observed the expected 2688→2816 hidden-size padding, and finished Slurm `COMPLETED` in 12m54s. Validation artifact: `/lustre/fsw/portfolios/coreai/users/spanev/nemorl-sglang-validation/20260827-pr3614-bfa0eaf3/runs/mxfp8-hpad-6588661/validation.json`.

Current head `f777ea03137ec1d2747392c9a54fe25c12249342` adds only the required BF16 declarations to four real SGLang test fixtures on top of the GPU-validated `54e9be45c46ed9d33d245d1c9032c6cec6cc8653`; runtime, recipe, and dependency files are unchanged.


nemo_rl/models/generation/sglang/mxfp8_quantization_core.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def quantize_mxfp8(weight: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    """Return ``(qweight, scale)`` in the SGLang MXFP8 layout.
+
+    Uses flashinfer's swizzle-free MXFP8 kernel
+    (``is_sf_swizzled_layout=False``).
+    """
+    try:
+        from flashinfer import mxfp8_quantize as flashinfer_mxfp8_quantize
```
Kh4L, line 119: **1 action item.**

**TL;DR — the auto-conversion path runs in the Ray *driver*, whose venv has no `flashinfer` and no assigned GPU, so the first `scheme: mxfp8` run cannot convert.**

Traced statically (no GPU here, so not observed). The chain: `grpo.setup()` / `ppo.setup()` call `ensure_sglang_quantized_checkpoint`, and `setup()` is a plain call from `main()` in `examples/run_grpo.py` — not a Ray actor. It reaches `convert_mxfp8` → `quantize_mxfp8` → this import.

Why the driver doesn't have it: `flashinfer-python` is in **no** `[project].dependencies`, only in the `vllm` / `sglang` / `mcore` extras. `docker/Dockerfile` sets a single `UV_PROJECT_ENVIRONMENT`, and its final syncs are `--all-groups` — dependency *groups*, not extras — with the comment at :217-219 saying the earlier per-extra syncs exist only to warm the cache. Worker tiers get extras via `PY_EXECUTABLES`; the driver does not. Every documented launch is a bare `uv run examples/run_grpo.py`. Same call also needs `torch.cuda.is_available()` at [`mxfp8_setup.py:135`](https://github.com/NVIDIA-NeMo/RL/blob/a30d34a0a1cd7e13133f6a9dd65fb9e4889c9c9a/nemo_rl/models/generation/sglang/mxfp8_setup.py#L135), on a process with no Ray-assigned GPU.

Scope, so this isn't overstated: a pre-converted `converted_model_path` returns before any flashinfer import, so the feature is usable — it's the headline auto-convert path that isn't.

The error text compounds it by blaming "the current **actor** environment ... normally provided by the pinned `mcore` or `sglang` extras", pointing at worker venvs when the failure is in the driver.

**Action:** pick one and enforce it in `ensure_sglang_quantized_checkpoint` — either run conversion in a one-shot `@ray.remote(num_gpus=1)` task with the mcore/sglang runtime env, or refuse to convert in-process and require `converted_model_path`, raising a message that says so. The second matches this module's own "Offline HF -> MXFP8 conversion" docstring and is the smaller change. Either way the error text here should name the driver.

nemo_rl/models/generation/sglang/mxfp8_setup.py
```diff
@@ -0,0 +1,507 @@
...
+
+    converted = quantization_cfg.get("converted_model_path")
+    if converted and is_existing_mxfp8_checkpoint(converted):
+        _merge_checkpoint_modules_to_not_convert(
+            checkpoint_path=converted,
+            quantization_cfg=quantization_cfg,
+        )
+        return converted
+
+    cache_root = (
```
Kh4L, line 493: **1 action item, 1 follow-up.**

**TL;DR — `model_path` is documented as "local folder or HF repo id" and the shipped exemplar sets it to a repo id, but the mxfp8 path only handles directories and fails blaming a file that was never going to exist.**

Observed executing. [`grpo_math_1B_sglang.yaml:11`](https://github.com/NVIDIA-NeMo/RL/blob/a30d34a0a1cd7e13133f6a9dd65fb9e4889c9c9a/examples/configs/grpo_math_1B_sglang.yaml#L11) sets `model_path: ${policy.model_name}` → `"Qwen/Qwen2.5-1.5B"`. Flip the `scheme` in that same shipped exemplar and you get:

```
ValueError: Qwen/Qwen2.5-1.5B/config.json must define a positive integer num_hidden_layers.
```

Nothing in the three new modules touches the HF hub (`grep snapshot_download|huggingface_hub|from_pretrained` → nothing), while [`config.py`](https://github.com/NVIDIA-NeMo/RL/blob/a30d34a0a1cd7e13133f6a9dd65fb9e4889c9c9a/nemo_rl/models/generation/sglang/config.py#L111) promises repo ids and SGLang itself resolves them.

**Action:** reject a non-directory `model_path` up front when the scheme is quantized, with a message that says "mxfp8 requires a local checkpoint directory" — or resolve it through `snapshot_download`. Right now the user is sent to debug a `config.json`.

**Follow-up (minor):** `_hash_qualified_save_dir` calls `os.path.abspath(model_dir)`, so a *relative* path produces a different cache directory per working directory — observed: two CWDs, two digests, two conversions. Wasted work, not incorrectness.

**Context — no action.** `cache_root`'s three-level fallback (config → `NRL_MXFP8_CACHE` → `~/.cache/nemo_rl/mxfp8`) is the only place `NRL_MXFP8_CACHE` appears in the repo, and neither `cache_root` nor `converted_model_path` appears in the exemplar YAML. Given they're the escape hatch for the action item above, they're the two keys most worth documenting.

nemo_rl/algorithms/grpo.py
```diff
@@ -1369,15 +1369,52 @@ def init_vllm_then_policy():
...
+            sglang_quantization_cfg
+        )
+        validate_sglang_quantized_refit_backend(
+            scheme=sglang_quantization_scheme,
+            use_megatron=bool(
+                policy_config.get("megatron_cfg", {}).get("enabled", False)
+            ),
+        )
+        generation_config["sglang_cfg"]["model_path"] = (
+            ensure_sglang_quantized_checkpoint(
```
Kh4L: **1 action item.**

**TL;DR — there is a third SGLang bring-up site that this PR doesn't guard, so `run_grpo_single_controller.py` with `scheme: mxfp8` boots a BF16 engine and then refits mxfp8 into it.**

Traced statically, and I verified the file myself. Grepping for the class rather than the helper gives exactly three bring-up sites: `grpo.py:1126`, `ppo.py:737`, and [`single_controller_utils/setup.py:204-215`](https://github.com/NVIDIA-NeMo/RL/blob/a30d34a0a1cd7e13133f6a9dd65fb9e4889c9c9a/nemo_rl/algorithms/single_controller_utils/setup.py#L204), which constructs `SGLangGeneration(...)` directly. This PR guarded the first two. That file contains zero occurrences of `quantization`, `ensure_sglang` or `validate_sglang`.

The consequence is the two halves of this block going missing:
1. `model_path` is never rewritten to the converted checkpoint → SGLang boots the **unconverted BF16** weights, while `_refit` reads `target_precision` from the same YAML block and sends mxfp8. What SGLang does with uint8 qweights plus unexpected `.weight_scale_inv` names I could not determine (sglang isn't installed) — hard failure at best, silent corruption at worst.
2. `validate_sglang_quantized_refit_backend` never runs. This half is milder: `dtensor_policy_worker_v2.py:1148-1151` still raises on non-bf16, so it degrades a setup-time failure into a first-refit failure.

Not a fringe path — it has `docs/guides/single-controller.md`, a unit test, and five test-suite scripts including Megatron ones.

**Action:** factor this block into one helper (it is already byte-identical between `grpo.setup` and `ppo.setup`) and call it from all three sites, so a fourth can't silently miss it. That also removes the duplication this PR introduces here and in `ppo.py`.

nemo_rl/models/generation/sglang/quantization_utils.py
```diff
@@ -0,0 +1,334 @@
...
+            )
+        if num_layers_at_start_in_bf16 + num_layers_at_end_in_bf16 > num_hidden_layers:
+            raise ValueError(
+                "The configured BF16 head/tail layer counts exceed the model's "
+                f"{num_hidden_layers} decoder layers."
+            )
+
+    tail_start_idx = num_hidden_layers - num_layers_at_end_in_bf16
+    dynamic_layer_prefixes = [
+        *(f"model.layers.{i}." for i in range(num_layers_at_start_in_bf16)),
```
Kh4L: **1 action item.**

**TL;DR — `num_layers_at_{start,end}_in_bf16` is a silent no-op on any model whose HF decoder prefix isn't `model.layers.`, including the one this work targets.**

Observed executing. The prefix is hardcoded here, but [`nccl_reshard_utils.py:462-472`](https://github.com/NVIDIA-NeMo/RL/blob/a30d34a0a1cd7e13133f6a9dd65fb9e4889c9c9a/nemo_rl/weight_sync/nccl_reshard_utils.py#L462) documents that this repo handles four layouts — `model.layers.N`, `model.language_model.layers.N`, `backbone.layers.N` (NemotronH), and bare `layers.N` (DeepSeek) — and solves exactly this with a layout-agnostic `_LAYER_RE`, one directory away.

Generated substrings `('model.layers.0.', 'model.layers.3.')` match: `model.layers.` ✅, `model.language_model.layers.` ✅ (by coincidence — `"language_model"` ends in `"model"`), `backbone.layers.` ❌, bare `layers.` ❌.

Reachability: of `NVIDIA-Nemotron-3-Nano-30B-A3B`'s 6243 tensors, `"model.layers.0."` matches **zero**. On that model both knobs do nothing at all, with no warning — a numerics knob that silently does nothing is worse than one that errors. No *shipped* SGLang recipe hits this today (all five are Qwen2.5-Math-1.5B, a `model.layers.` model), which is why this is inline rather than fixed.

**Action:** derive the prefix with the existing `_LAYER_RE` idiom instead of hardcoding, or — cheaper — assert that each generated prefix matches at least one weight name. Offline conversion already holds the full name list at the call site, so it can fail loudly listing the unmatched prefixes.

nemo_rl/models/generation/sglang/mxfp8_setup.py
```diff
@@ -0,0 +1,507 @@
...
+        substring
+        for substring in expanded_high_precision
+        if substring not in configured_high_precision
+    )
+    quantization_cfg["modules_to_not_convert"] = list(
+        dict.fromkeys((*user_modules, *checkpoint_modules, *concrete_atomic_modules))
+    )
+
+
+def _quantization_fingerprint(quantization_cfg: dict[str, Any]) -> str:
```
Kh4L: **No action needed in this PR — context for the `CONVERTER_VERSION` bump, and one follow-up.**

Observed executing: appending to `SKIP_WEIGHT_SUBSTRINGS` leaves this fingerprint bit-identical. It hashes six *config* keys; the module-level constants that equally determine the output — `SKIP_WEIGHT_SUBSTRINGS`, `HF_FUSED_LINEAR_WEIGHT_GROUPS`, `HF_MOE_EXPERT_NAME_MARKERS` — are outside it, and so is `should_quantize`'s own logic. The cache directory adds only the abs path, the source-config hash, and `CONVERTER_VERSION`.

So `CONVERTER_VERSION` is the **only** lever, and it's hand-maintained with no test binding it to the constants. That is why the router fix in this commit bumps it to `"5"`: without that, a directory converted under the old skip list is reused with its router already quantized, `is_existing_mxfp8_checkpoint(save_dir)` short-circuits before any reconversion, and the refit — now correctly skipping the router — sends BF16 into a slot the engine loaded as MXFP8.

**Follow-up:** folding `repr(SKIP_WEIGHT_SUBSTRINGS)` and the other selection constants into this fingerprint would make the invalidation automatic and remove the "remember to bump the version" footgun entirely.

nemo_rl/models/generation/sglang/quantization_utils.py
```diff
@@ -0,0 +1,334 @@
...
+) -> None:
+    """Reject quantized refits on backends that cannot emit quantized weights."""
+    if scheme != "bf16" and not use_megatron:
+        raise NotImplementedError(
+            f"SGLang {scheme} weight refit requires a Megatron policy because "
+            "the DTensor/FSDP refit path only supports BF16."
+        )
+
+
+def validate_checkpoint_high_precision_layout(
```
Kh4L, line 111: **1 action item.**

**TL;DR — this validator checks one direction only: it catches "config says BF16 but the checkpoint has scales", never "config says quantize but the checkpoint has none".**

Observed executing: given a checkpoint with a `.weight` and no `.weight_scale_inv` companion for a module the current config *would* quantize, this returns cleanly. The loop is driven by `high_precision_substrings`, so a module absent from that list is never examined.

The refit then emits `qweight` + `.weight_scale_inv` for a module SGLang loaded unquantized — wrong numerics with no trainer-side exception.

Reachability is the *same class* as the case you do check: both arrive via `converted_model_path` or a `model_path` that already points at an MXFP8 checkpoint, and both are triggered by editing exclusions between runs. `is_existing_mxfp8_checkpoint` also accepts any checkpoint whose `quant_method`/`weight_block_size`/`scale_fmt` match — including third-party ones this converter never produced.

**Action:** add the symmetric check — every `.weight` the current config would quantize must already carry its companion scale. The guard is half-built; this is the other half.

nemo_rl/models/policy/workers/megatron_sglang_weight_iterator.py
```diff
@@ -0,0 +1,147 @@
...
+        *,
+        target_precision: SglangQuantizationScheme = "bf16",
+        buffer_size_bytes: int,
+    ) -> Iterator[list[tuple[str, torch.Tensor]]]:
+        """Yield finalized HF tensor buckets sized by transmitted bytes."""
+        if buffer_size_bytes <= 0:
+            raise ValueError(
+                f"buffer_size_bytes must be positive, got {buffer_size_bytes}"
+            )
+        if target_precision == "mxfp8":
```
Kh4L: **No action needed in this PR — design note, deliberately not blocking.**

Scheme dispatch is hand-wired here and at four other shared sites, and the quantization decision is made *per tensor* inside the loop below. The next PR in this stack (#3615, nvfp4) is the measurement: it edits five shared production files, and because NVFP4 must quantize gate/up together under one shared global scale, it could not extend this loop at all — it bypasses it with an early `yield from ... ; return` and adds a ~75-line `@staticmethod` that touches no `self`, in this shared file.

That's the useful signal: the seam needed to be **stream-level**, not per-tensor, because one scheme is a stateful stream transform rather than a map. A small frozen record of `(ensure_checkpoint, refit_skip_substrings, finalize_hf_stream)` per scheme would let this class be the bucketizer its docstring already claims to be, and would let each scheme be unit-tested from a plain list of `(name, tensor)` instead of monkeypatching a symbol on this shared module.

Explicitly **not** suggesting the two `*_quantization_core.py` modules be merged — they change for genuinely different reasons (flashinfer vs Transformer Engine, block scale vs group + per-tensor global scale, different companion suffixes) and duplication there is correct. The seam belongs above them.

Raising it here only because #3615 already implements the stream-shaped version, so the cheapest moment to unify is while both are in flight. Not worth gating this PR on.

tests/unit/models/generation/sglang/test_mxfp8_setup_selection.py
```diff
@@ -0,0 +1,110 @@
...
+            },
+        },
+    )
+    _write_json(
+        checkpoint / "model.safetensors.index.json",
+        {"weight_map": {name: "model.safetensors" for name in weight_names}},
+    )
+
+
+def test_existing_mxfp8_checkpoint_preserves_atomic_qkv_skip(
```
Kh4L: **1 action item.**

**TL;DR — the cache-key functions decide whether an expensive GPU conversion is reused, and no test exercises them in either direction.**

`_quantization_fingerprint` and `_hash_qualified_save_dir` have no coverage: making the fingerprint a constant, or dropping it from the cache-dir payload, leaves the suite green. With that mutation two different skip policies collide on one directory — so a run silently loads a checkpoint quantized under a *different* policy. Also uncovered: the `converted_model_path` branch, `cache_root`/`NRL_MXFP8_CACHE` resolution, the cache-hit branch, and `_get_num_hidden_layers`' `text_config` (VLM) fallback.

This test itself is narrower than its name suggests: it writes `modules_to_not_convert=[q, k, v]` and asserts the same three come back, which they do via `checkpoint_modules` — `concrete_atomic_modules` contributes nothing, so dropping the atomic expansion from the merge leaves it passing. The sibling test at L81 *is* load-bearing (it kills both layout-validation mutants).

I deliberately did not bulk-add tests to someone else's PR, but they're written and verified: a reviewer-authored set covering these paths runs green against this code (50 tests). The highest-value ones are "two skip policies must not share a cache directory" and "an identical policy reuses it".

**Action:** cover the two cache-key functions in both directions, and rename this test to what it actually asserts (it honours a supplied `modules_to_not_convert`; it doesn't establish the QKV synchronization).

Kh4L (commented):
## Self-review (`/review-pr-team`), per the contributor self-review policy

Scoped to this PR's own increment (`be5015649..8f6003694` as reviewed — the mxfp8 commit plus its ruff-format fix). The two layers below have their own posted self-reviews and were not re-reviewed.

**The headline: the bf16 default path is a strict no-op and is safe, but the mxfp8 path could not complete `setup()` on any config in this repo, and on the model this work targets it quantized the MoE router.** Four findings are fixed in a30d34a0a; six are left inline for your judgement because each needs a design decision I did not want to make unilaterally on someone else's design.

### Fixed in a30d34a0a

1. **MoE routers were being quantized.** `SKIP_WEIGHT_SUBSTRINGS` held `"router"` and `"mlp.gate."`, which between them miss most router spellings. A top-k router is a discrete argmax, so quantization error flips expert selection — a rollout/trainer policy divergence, the one failure an RL run can't absorb. Widened to `.gate.` + `shared_expert_gate.`, covering `mlp.gate.` (Qwen3-MoE, DeepSeek), `block_sparse_moe.gate.` (Mixtral), `mixer.gate.` (NemotronH/nanov3) and Qwen2-MoE's separate gate, while leaving `gate_proj` / `gate_up_proj` / `shared_expert.gate_proj` quantized. The repo already draws this line in `DEFAULT_NVFP4_IGNORE` and the nanov3 recipe's `real_quant_ignore`.
2. **`CONVERTER_VERSION` → 5**, mandatory given (1). `_quantization_fingerprint` hashes the config knobs but not the module-level selection constants, so changing the skip list without bumping it silently reuses a cached checkpoint whose router is already quantized — while the refit, now skipping it, sends BF16 into a slot the engine loaded as MXFP8. See the inline comment on the fingerprint for the durable fix.
3. **`convert_mxfp8` required `model.safetensors.index.json`.** That's every shipped SGLang recipe: all five run Qwen2.5-Math-1.5B-Instruct, which ships a single `model.safetensors` and no index. `_checkpoint_weight_names` in the same module already handled both layouts; the converter now takes the same branch.
4. **`test_quantized_refit_requires_megatron_backend` asserted nothing** — only the two non-raising cases. Neutering the guard to a bare `return` left all 17 tests green. Plus `ruff check` failures (two unused imports, three import-ordering) that the earlier ruff-format commit didn't cover but `lint-check` would have caught, and a 2025 copyright header on two files added in a 2026 commit.

### Process disclosure

The adversarial pass ran but was unreliable to reach; several rounds produced nothing before it reported. It ultimately delivered, and it earned its place — **it overturned my own fix.** My first version of finding (1) enumerated `block_sparse_moe.gate.` and `shared_expert_gate.`, which still left `backbone.layers.N.mixer.gate.weight` quantized — the router of `NVIDIA-Nemotron-3-Nano-30B-A3B`, the model this work targets. It caught that by fetching the real checkpoint's safetensors header and running `should_quantize` against the actual tensor names and shapes. The committed fix is its `.gate.` widening, not mine.

It also corrected the review in three other directions worth recording: it disputed a confidence-90 finding of chained `.get()` at `grpo.py:1402` as a style note on code matching pre-existing house style exactly (dropped); it downgraded a coverage claim after showing three independent hand-rolled measurements disagreed on the denominator (so no percentage appears below); and it found one agent's cited test artifact did not exist, so that agent's numbers are not cited here.

### Environment caveats

No GPU and no `sglang`/`flashinfer` installed, so **nothing involving SGLang's or flashinfer's runtime behaviour was executed by anyone.** Findings are marked traced-vs-executed inline. `pytest` could not run the repo suite (`uv.lock` is Linux-only on this machine); the new test file's assertions were verified by executing the real `should_quantize` against them, and the file will collect normally in CI, where `transformers` is a core dependency.

### What would take this from "unexercised" to "shippable"

The PR body is candid that no mxfp8 run exists. In rough order of cost:

1. **One smoke run** — GRPO + Megatron + SGLang + `scheme: mxfp8`, 2–3 steps, on a sharded model. This alone exercises everything fixed above.
2. **Round-trip correctness** — per-token logprob agreement between the mxfp8 SGLang rollout and the Megatron policy on a fixed prompt batch. The existing sglang recipes already gate on `median(token_mult_prob_error) < 1.1`, so the same number under mxfp8 vs the bf16 baseline is the natural A/B with no new tooling.
3. **≥3 refits, not 1**, with the importance-sampling ratio logged per step — drift from quantizing *during* refit only shows after several updates.
4. **Throughput and memory A/B**, bf16 vs mxfp8, same model/TP/batch. Without this there is no stated benefit.
5. **A run with `num_layers_at_start_in_bf16 > 0`** where the emitted `config.json` is inspected — the cheapest possible check, and it would have caught the layer-band issue flagged inline.

No `docs/` page accompanies seven new user-facing config keys. `docs/fp8.md` is the natural home — it already documents the vLLM FP8 story including the equivalent layer-band knobs.

_Generated by Claude Code_

nemo_rl/models/generation/sglang/mxfp8_quantization_core.py
Kh4L: Fixed in 72feae3f. Startup no longer imports or runs FlashInfer conversion in the Ray driver: it accepts a local preconverted MXFP8 checkpoint (including the deterministic cache path) or fails with explicit offline-GPU conversion instructions. The commented import is now only reached by explicit offline or worker-side quantization, not driver startup.

tests/unit/models/generation/sglang/test_mxfp8_setup_selection.py
Kh4L: Fixed in 72feae3f. Tests now assert stable cache keys for identical policies, distinct keys for different skip policies, deterministic-cache reuse, and explicit preconverted and HF-ID paths. The original test was renamed to describe the supplied QKV skip behavior it actually verifies.

nemo_rl/models/policy/workers/megatron_sglang_weight_iterator.py
Kh4L: Acknowledged as nonblocking. This PR stays scoped to the MXFP8 implementation; the stream-level dispatch refactor can be handled separately alongside the NVFP4 extension.

nemo_rl/models/generation/sglang/quantization_utils.py
Kh4L: Fixed in 72feae3f. Reused checkpoints are now validated symmetrically: high-precision weights must not have a scale companion, and every weight selected for quantization must have one. A missing-scale test covers the latter.

nemo_rl/models/generation/sglang/quantization_utils.py
Kh4L: Fixed in 72feae3f. Head/tail rules now emit boundary-safe `layers.N.` substrings, which cover `model.layers`, `model.language_model.layers`, `backbone.layers`, and bare `layers` layouts; tests cover all four and prevent layer 1 from matching layer 10.

nemo_rl/algorithms/grpo.py
Kh4L: Fixed in 72feae3f. The duplicated setup was factored into `prepare_sglang_quantized_generation`, and all three SGLang bring-up sites (GRPO, PPO, and single-controller) call it before constructing SGLang.

nemo_rl/models/generation/sglang/mxfp8_setup.py
Kh4L: Fixed in 72feae3f. HF repo IDs now fail before config parsing with a direct instruction to download and convert offline and set `converted_model_path`; that path is validated as a local MXFP8 directory. Relative paths are canonicalized with `abspath`, so the cache key follows the actual source directory selected by the caller CWD.

nemo_rl/models/generation/sglang/mxfp8_setup.py
Kh4L: Acknowledged. This PR keeps `CONVERTER_VERSION` as the explicit invalidation mechanism and bumps it with selection-logic changes; automatically hashing the module-level selection constants is a useful nonblocking follow-up.

Kh4L:
Current head `f777ea03137ec1d2747392c9a54fe25c12249342` fixes all four stale plain-BF16 SGLang test fixtures. Compared with HSG-validated `54e9be45c46ed9d33d245d1c9032c6cec6cc8653`, only those unit-test fixtures changed; runtime, recipes, and dependencies are identical. Fresh CI is triggered on the exact head.

PR #3615, opened 2026-08-12
feat(sglang): nvfp4 quantized rollout and refit

Kh4L:
**Original work: #3426 by @xiuhu17.** Re-cut at his request so it can be carried through review on his behalf. His commits retain their original authorship and `Signed-off-by`.

**3 of 3 remaining in the stack.** `#3613 (fault tolerance) → #3614 (mxfp8) → this`. #3612 has merged. GitHub PRs from a fork must target `main`, so this diff contains the two PRs below it. Review the ten commits after #3614.

This PR cannot merge before the PRs below it because it contains their commits.

## What it adds

NVFP4 quantized rollouts with Megatron weight refit, built on the MXFP8 refit layer. The refreshed implementation uses the shared weight iterator introduced below this PR and keeps the obsolete standalone iterator deleted.

The SGLang config also exposes the MoE and FP4 GEMM runner backends. This lets NVFP4 validation explicitly exercise CuteDSL instead of accepting Nemotron-H's SM100 automatic MoE selection (`flashinfer_trtllm`).

Startup requires a local checkpoint produced by NeMo-RL's offline GPU converter. Converted outputs carry an explicit MoE-only layout marker, while incompatible external ModelOpt NVFP4 layouts are rejected before SGLang starts.

The latest runtime fix preserves SGLang's padded NVFP4 storage during online refits, refreshes CUDA-graph-captured scale tensors in place after each update, and treats a semantic `{success: false}` update response as a refit failure.

## Transformer Engine dependency

NVFP4 conversion uses Transformer Engine's `NVFP4Quantizer`, including the `nvfp4_use_4over6` and `nvfp4_4over6_err_mode` controls, so this layer moves Transformer Engine to v2.17. The regenerated lock changes only Transformer Engine (`2.15.0+42b8400 → 2.17.0+2e559f0`) within the current 560-package resolution.

## Verification

`uv lock --check` resolves 560 packages. The repository-pinned Ruff, Ruff-format, and TOML-format hooks pass across the refreshed branch. Focused shared MXFP8/NVFP4 checkpoint tests pass, including complete two-scale validation and rejection of standard external ModelOpt layouts. The pinned SGLang `0.5.17.dev30+g3003d70f6` source patches apply idempotently against a pristine install.

HSG job `6593575` completed successfully (`0:0`) on 3 nodes / 12 GB200 GPUs using `batch`, `short`, and `coreai_dlalgo_nemorl`. It exercised exact head `736472843ba46eca4c16f127009452bdb148f090` with Nemotron-H, Megatron training, SGLang NVFP4 rollout/refit, and the quantized validation gates. All 4/4 optimizer steps and 4/4 broadcast refits completed with positive gradient norms and clean teardown.

Current head `7d429c388bdaade042f6aa0e108abd289fddc6b1` is the same runtime tree restacked on #3614, plus required declarations in SGLang test fixtures. Compared with the GPU-validated head, only five files under `tests/unit/models/generation/sglang/` changed; runtime, recipe, and dependency files are unchanged.


nemo_rl/models/policy/workers/megatron_sglang_weight_iterator.py
```diff
@@ -0,0 +1,244 @@
...
+                ),
+                *nvfp4_quantized_entries(
+                    up_name,
+                    up_output,
+                    include_input_scale=False,
+                ),
+            ]
+            del pending_pairs[pair_base]
+
+        if pending_pairs:
```
Kh4L: **1 action item.**

**TL;DR — this completeness check fires only after the rest of the refit has already been transmitted, so a stranded gate/up half leaves the engines holding a mix of new and stale weights with no rollback.**

Observed executing. `pending_pairs` buffers a half until its partner arrives, but this generator is consumed lazily by [`iter_hf_weight_buckets`](https://github.com/NVIDIA-NeMo/RL/blob/6e65999985827e4c7c1b999a840ca4be77881918/nemo_rl/models/policy/workers/megatron_sglang_weight_iterator.py#L112) and the caller transmits each bucket as it is produced. So the `raise` here happens *after* earlier buckets are already in the engines. Feeding a rank-2 gate and a rank-3 up:

```
buckets already emitted BEFORE the raise:
  [['...up_proj.weight'],
   ['...down_proj.weight', '...down_proj.weight_scale', '...down_proj.weight_scale_2'],
   ['lm_head.weight']]
RAISED: ... incomplete pairs: {'model.layers.0.mlp.experts.0': ['gate']}
```

Scope, so this isn't overstated: the **name**-axis skip is correctly symmetric — `should_skip_nvfp4_gated_pair` keeps both halves in BF16, and your own test covers it. This is specifically the non-name rejections in `should_quantize_nvfp4` (dtype, `dim() != 2`), which have no pair-level counterpart, so one half can be quantized while the other silently isn't.

**Action:** decide pair completeness *before* emitting anything — the conversion-task list is known up front, so the offline side already does exactly this in `_build_conversion_policy`. Alternatively buffer the whole refit before transmitting. As written, the guard is real but fires too late to protect anything.

nemo_rl/models/generation/sglang/nvfp4_quantization_core.py
```diff
@@ -0,0 +1,429 @@
...
+    """Return whether an HF tensor should be emitted in NVFP4 format."""
+    if any(substring in name for substring in skip_weight_substrings):
+        return False
+    if not is_moe_expert_weight_name(name):
+        return False
+    if weight.dtype not in (torch.float16, torch.bfloat16, torch.float32):
+        return False
+    if weight.dim() != 2:
+        return False
+    if weight.shape[-1] % NVFP4_GROUP_SIZE != 0:
```
Kh4L, line 113: **1 action item.**

**TL;DR — this is the only rejection in the function that raises; the three above it return `False`. So the same tensor is either fatal or harmless depending purely on whether the user happened to list it in `modules_to_not_convert`.**

Observed executing:

```
_iter_nvfp4_hf_named_tensors([("...experts.0.down_proj.weight", bf16[16,72])], skip=())
  -> ValueError: Last dim 72 must be divisible by 16 for NVFP4 (...)   # the whole refit dies

same tensor, skip=("down_proj",)
  -> ['...experts.0.down_proj.weight']                                  # passes through in BF16
```

`is_moe_expert_weight_name`, the dtype check and the rank check all mean "not eligible, pass it through untouched". A K that isn't a multiple of 16 means the same thing, but kills the run instead. It also pre-empts the offline converter's richer error at [`_convert_shards`](https://github.com/NVIDIA-NeMo/RL/blob/6e65999985827e4c7c1b999a840ca4be77881918/nemo_rl/models/generation/sglang/nvfp4_setup.py#L318), which names both dtype *and* shape.

**Action:** return `False` here and let the caller decide. `_convert_shards` already raises deliberately for an ineligible tensor it *selected*, which is the right place for the loud failure — a live refit passing an unquantizable expert through in BF16 is strictly better than aborting the refit, and matches what every other rejection in this function does.

nemo_rl/models/generation/sglang/nvfp4_quantization_core.py
```diff
@@ -0,0 +1,429 @@
...
+
+FP4_E2M1_MAX = 6.0
+FP8_E4M3_MAX = 448
+NVFP4_GROUP_SIZE = 16
+TE_NVFP4_ROW_ALIGNMENT = 16
+
+NVFP4_QUANTIZATION_CONFIG: dict[str, Any] = {
+    "group_size": NVFP4_GROUP_SIZE,
+    "ignore": [],
+    "kv_cache_scheme": {"dynamic": False, "num_bits": 8, "type": "float"},
```
Kh4L: **1 action item.**

**TL;DR — every converted checkpoint advertises static FP8 KV-cache quantization that the converter never produces, and SGLang answers the missing scales by silently substituting 1.0.**

Traced against the pinned SGLang rev (sglang isn't installed here, so not observed). This dict is hardcoded, and [`_write_output_metadata`](https://github.com/NVIDIA-NeMo/RL/blob/6e65999985827e4c7c1b999a840ca4be77881918/nemo_rl/models/generation/sglang/nvfp4_setup.py#L482) additionally writes `kv_cache_quant_algo: "FP8"` into `hf_quant_config.json`. No `k_scale`/`v_scale` tensor is emitted anywhere in the diff, and a BF16 source has none.

Downstream, in order:
- [`ModelOptFp4Config.from_config`](https://github.com/sgl-project/sglang/blob/3003d70f680d41c59d1b7acbf65cb47795dfd19e/python/sglang/srt/layers/quantization/modelopt_quant.py#L1262) reads `kv_cache_scheme` and maps `{"type": "float", "num_bits": 8}` → `kv_cache_quant_algo = "FP8"`.
- [`_get_quant_method`](https://github.com/sgl-project/sglang/blob/3003d70f680d41c59d1b7acbf65cb47795dfd19e/python/sglang/srt/layers/quantization/modelopt_quant.py#L296) then attaches `ModelOptFp8KVCacheMethod` to **every** `RadixAttention` layer.
- [`BaseKVCacheMethod`](https://github.com/sgl-project/sglang/blob/3003d70f680d41c59d1b7acbf65cb47795dfd19e/python/sglang/srt/layers/quantization/kv_cache.py#L53) seeds both scales to a `-1.0` sentinel and, finding no scales in the checkpoint, takes the "both invalid" branch and sets them to `1.0` — **no warning**.

Impact is bounded but real: inert while `kv_cache_dtype` is `auto`, because the scales go unused. Set `kv_cache_dtype: fp8_e4m3` — which this repo exposes — and the KV cache is quantized with an uncalibrated unit scale and nothing says so. Note `MXFP8_QUANTIZATION_CONFIG` emits no KV-cache key at all, so this is new here rather than inherited.

The source-preserving override at [`_write_output_metadata`](https://github.com/NVIDIA-NeMo/RL/blob/6e65999985827e4c7c1b999a840ca4be77881918/nemo_rl/models/generation/sglang/nvfp4_setup.py#L456) doesn't rescue it: it fires only for a source whose `quantization_config` carries a `kv_cache_scheme` *and* an absent/`bf16` `quant_method`, which no plain BF16 HF checkpoint has.

**Action:** drop `kv_cache_scheme` from this dict and `kv_cache_quant_algo` from `hf_quant_config.json`, leaving the source-preserving branch as the only way either appears. A checkpoint shouldn't declare a scheme it doesn't carry.

nemo_rl/models/generation/sglang/nvfp4_setup.py
```diff
@@ -0,0 +1,761 @@
...
+    payload = (
+        f"{absolute_model}|{source_fingerprint}|{quantization_fingerprint}|"
+        f"v{CONVERTER_VERSION}"
+    )
+    digest = hashlib.sha1(payload.encode("utf-8")).hexdigest()[:16]
+    base = os.path.basename(os.path.normpath(absolute_model)) or "hf"
+    return os.path.join(os.path.abspath(cache_root), f"{base}-nvfp4-{digest}")
+
+
+def is_existing_nvfp4_checkpoint(path: str) -> bool:
```
Kh4L, line 674: **1 action item.**

**TL;DR — this never recognizes a real ModelOpt NVFP4 checkpoint, because released ones put the quantization block in `hf_quant_config.json`, not `config.json`.**

Observed over the network against two shipped NVIDIA releases — `nvidia/Qwen3-235B-A22B-NVFP4` and `nvidia/DeepSeek-R1-0528-NVFP4`. Both report **no** `quantization_config` in `config.json`, while their `hf_quant_config.json` carries `{"quant_algo": "NVFP4", "group_size": 16, ...}`. That is also the file SGLang itself declares via `get_config_filenames()`, and the file this module *writes* on the way out.

The consequence chain: `is_existing_nvfp4_checkpoint` → `False`, so `ensure_nvfp4_checkpoint` falls through to conversion; `is_bf16_source_checkpoint` then sees an absent `quantization_config` and returns `True`; and `convert_nvfp4` tries to re-quantize weights that are already NVFP4, dying in `_convert_shards` with `not a supported NVFP4 source (dtype=torch.uint8, ...)` — after having already copied the checkpoint's metadata into the cache directory.

Loud rather than silent, so this is usability rather than corruption. But it means the "point `model_path` at an existing NVFP4 checkpoint" path — the cheapest way to try this feature, and the one that avoids needing a GPU on the driver — doesn't work for any checkpoint the ecosystem actually publishes.

**Action:** read `hf_quant_config.json` as well as `config.json`, preferring whichever is present. `_read_source_config` is one call away and `_write_output_metadata` already treats both as the pair.

nemo_rl/models/generation/sglang/nvfp4_setup.py
```diff
@@ -0,0 +1,761 @@
...
+    _, configured, _, _ = _validated_conversion_options(
+        quantization_cfg,
+        num_hidden_layers=num_hidden_layers,
+    )
+    requested_high_precision = get_dynamic_high_precision_substrings(
+        quantization_config=quantization_cfg,
+        num_hidden_layers=num_hidden_layers,
+    )
+    checkpoint_weight_names = tuple(_source_weight_map(checkpoint_path))
+    configured_high_precision = tuple(
```
Kh4L, line 729: **1 action item.**

**TL;DR — the `ignore`/`exclude_modules` list ingested here contains glob patterns in real ModelOpt checkpoints, but it is matched with plain substring containment, so those entries are silent no-ops.**

Observed over the network: `nvidia/DeepSeek-R1-0528-NVFP4`'s `exclude_modules` contains entries like `"model.layers.0.self_attn*"`, `"model.layers.1.self_attn*"`, one per layer. Downstream, [`should_quantize_nvfp4`](https://github.com/NVIDIA-NeMo/RL/blob/6e65999985827e4c7c1b999a840ca4be77881918/nemo_rl/models/generation/sglang/nvfp4_quantization_core.py#L105) tests `any(substring in name ...)`, and `"model.layers.0.self_attn*" in "model.layers.0.self_attn.q_proj.weight"` is `False` — the trailing `*` guarantees it never matches.

Two consequences, and the second is the quieter one: those entries do nothing for skipping, *and* they make `validate_checkpoint_high_precision_layout` pass vacuously, because that guard is driven by the same substrings. So the check that exists to catch "the checkpoint disagrees with the config" cannot fire for any glob entry.

This compounds with the guard being one-directional — it only flags high-precision-selected names that carry quantized companions, never checkpoint tensors that are quantized but which the refit will emit in BF16. Since NVFP4 deliberately has no static skip list, no attention or dense module can ever be *selected* into it, so for exactly the modules these globs name, the guard is structurally incapable of firing.

Provenance: the glob format is observed; the control-flow consequence is traced.

**Action:** match these with `fnmatch` rather than `in`, or normalize the trailing `*` on ingest here. Worth doing together with the `hf_quant_config.json` item above — both only bite on the reuse-an-existing-checkpoint path, and together they make that path unusable.

nemo_rl/models/generation/sglang/nvfp4_quantization_core.py
```diff
@@ -0,0 +1,429 @@
...
+        nvfp4_e4m3_max=nvfp4_e4m3_max,
+    )
+    return (
+        qweight,
+        block_scale.view(torch.float8_e4m3fn),
+        weight_scale_2,
+    )
+
+
+def _contiguous_pair_view(
```
Kh4L, line 311: **1 action item (small).**

**TL;DR — every other precondition failure here falls back to `torch.cat`, but a non-canonical stride produces a wrong *result* instead.**

`is_contiguous()` ignores the stride of size-1 dimensions, so a tensor with a size-1 leading dim can carry an arbitrary `stride[0]` and still report contiguous. That defeats both the stride equality test (which is otherwise a sound proxy for matching column counts) and the storage-offset adjacency reasoning. Observed executing:

```
b = torch.arange(40.)
f = b.as_strided((1,4),(8,1),0)   # is_contiguous() -> True
s = b.as_strided((1,4),(8,1),4)   # is_contiguous() -> True, strides equal, "adjacent"
_contiguous_pair_view(f,s) -> [[0,1,2,3],[8,9,10,11]]     # row 1 read from offset 8
expected                   -> [[0,1,2,3],[4,5,6,7]]
```

Realistic exposure is low — it needs a 1-row expert weight with a synthetic stride — so this is a robustness hole, not a live bug. Recording it because the failure mode is silently-wrong-numbers rather than a fallback, which is the one outcome this helper is otherwise careful to avoid.

For the record, two things I checked here that are **fine**: the missing `first.shape[1] == second.shape[1]` test doesn't matter (the caller raises on K mismatch first, and for canonically-strided 2-D tensors the stride check is an exact proxy), and overlap is impossible given the offset-adjacency requirement.

**Action:** one line — `if first.stride() != (first.shape[1], 1): return None`.

nemo_rl/models/generation/sglang/nvfp4_quantization_core.py
```diff
@@ -0,0 +1,429 @@
...
+        raise RuntimeError(
+            "Transformer Engine returned an incomplete rowwise NVFP4 tensor."
+        )
+
+    qweight = rowwise_data[:num_rows, : num_cols // 2].contiguous()
+    block_scale = rowwise_scale_inv[
+        :num_rows,
+        : num_cols // NVFP4_GROUP_SIZE,
+    ].contiguous()
+    weight_scale_2 = nvfp4_global_decode_scale_te(
```
Kh4L, line 300: **No action needed in this PR — a coupling worth naming, plus the verification I did so nobody has to redo it.**

The block scales in the emitted checkpoint come from Transformer Engine, but `weight_scale_2` is re-derived here from our own reimplementation of TE's formula. The two must agree **exactly** — they are the two halves of one dequantization — and nothing enforces or tests that.

I verified the copy is faithful today, against TE's reference recipe:

```python
global_encode_scale = torch.div(GLOBAL_SCALE_E4M3_MAX * FLOAT4_E2M1_MAX, global_amax)
global_encode_scale = torch.min(global_encode_scale, finfo(float32).max)
...  # zero -> one
global_decode_scale = torch.div(1.0, global_encode_scale)
```

`nvfp4_global_encode_scale_te` matches term for term. Two related things I confirmed at the same time, both of which could have been quietly catastrophic and both of which are **correct**:

- **`_amax_rowwise` is a per-tensor scalar here**, so `.reshape(-1)[0]` is right rather than silently discarding all but row 0. TE only reshapes the global amax to a per-row `[m, 1, 1]` under `row_scaled_nvfp4`, and this passes `row_scaled_nvfp4=False`.
- **`_rowwise_scale_inv` is plain row-major, not swizzled** (`with_gemm_swizzled_scales` defaults `False`), so the `[:num_rows, :num_cols // NVFP4_GROUP_SIZE]` slice reads the region it intends to rather than a tiled layout.

The 448-vs-256 gating also mirrors TE, which likewise ignores `nvfp4_e4m3_max` unless `nvfp4_use_4over6` is set.

**Why it's still worth a note:** this is a hand-copied numerical contract against private attributes (`_rowwise_data`, `_rowwise_scale_inv`, `_amax_rowwise`) with a pinned-version comment as its only guard. A change to `pow_2_scales`, a different `row_scaled_nvfp4` default, or a tweak to the clamp desynchronizes the halves, and the symptom is every dequantized expert weight off by a constant factor — not a crash. Worth either a comment recording *which* TE function this mirrors, or an assertion in the smoke test that a round-tripped weight matches TE's own dequantize.

Kh4L (commented):
## Self-review (`/review-pr-team`), per the contributor self-review policy

Scoped to this PR's own increment (`35d910009..4f56b148a` — the nvfp4 feature, the TE 2.17 bump, the lock regen, and my two rebase fixes). The three layers below have their own posted self-reviews and were not re-reviewed.

**The headline: the lint gate is red and the two tests guarding the pair-completeness logic have never executed.** `pytest` is used at `test_megatron_sglang_weight_iterator.py:240,251` but never imported — `ruff` fails with `F821` twice, and both tests error with `NameError` instead of running. Fixed in `6e6599998`. Beyond that, the offline/online consistency I most expected to be wrong is actually **correct**, and the dependency bump is **justified and clean** — but there are five correctness issues in the new code and a checkpoint-compatibility gap that stops stock ModelOpt NVFP4 checkpoints from being recognized at all.

### Fixed in `6e6599998`

**The missing `import pytest`.** Verified rather than assumed: same harness, pre-fix file → `2 failed, 5 passed` (`NameError: name 'pytest' is not defined`); post-fix → `7 passed`. `ruff check` now clean across all 15 changed files. The two dead tests were the *only* coverage for `_iter_nvfp4_hf_named_tensors`' incomplete-pair and duplicate-role guards, so replacing either `raise` with a bare `return` was indistinguishable from the shipped state.

Provenance note, since I'd rather say it than have it found: this is my own miss. My commit `4f56b148a` says these fixes were "surfaced by running the repo's own `pre-commit` against this layer" — I ran it against the *production* files and not the tests, which is exactly how an `F821` in a test file survived a commit whose stated purpose was fixing an `F821`.

### Left inline for your judgement

Seven items, in rough severity order. All are marked traced-vs-executed inline.

1. **A stranded gate/up half is detected only *after* part of the refit has already shipped.** The completeness check sits at the end of the generator, but buckets are consumed lazily and transmitted as they're produced — so the engines are left holding a mix of new and stale weights with no rollback. The name-axis skip is correctly symmetric; this is the dtype/rank rejections, which have no pair-level counterpart.
2. **`should_quantize_nvfp4` raises where all three of its sibling checks return `False`.** The same expert tensor is fatal or harmless depending only on whether the user happened to list it in `modules_to_not_convert`.
3. **The converted checkpoint advertises FP8 KV-cache quantization it does not contain.** SGLang reads `kv_cache_scheme` → attaches `ModelOptFp8KVCacheMethod` to every `RadixAttention` → finds no `k_scale`/`v_scale` → *silently* substitutes 1.0. Inert at `kv_cache_dtype: auto`, uncalibrated and unwarned otherwise. mxfp8 emits no KV metadata at all, so this is new here.
4. **Stock ModelOpt NVFP4 checkpoints are never recognized.** `is_existing_nvfp4_checkpoint` reads only `config.json`, but real releases put the block in `hf_quant_config.json` — the file SGLang itself declares. Confirmed against `nvidia/Qwen3-235B-A22B-NVFP4` and `nvidia/DeepSeek-R1-0528-NVFP4`: both have no `quantization_config` in `config.json`. The converter then tries to re-quantize an already-NVFP4 checkpoint and dies, after copying its metadata into the cache dir.
5. **`exclude_modules` entries are globs, matched with `in`.** Real ModelOpt checkpoints ship `"model.layers.0.self_attn*"`; plain substring containment never matches, so those entries are silent no-ops for skipping *and* make `validate_checkpoint_high_precision_layout` pass vacuously.
6. **`_contiguous_pair_view` doesn't require canonical strides.** A size-1 leading dim can carry any `stride[0]` and still report contiguous, which defeats both the stride proxy and the adjacency check — and unlike every other precondition failure it returns a *wrong result* rather than falling back to `torch.cat`.
7. **`weight_scale_2` is re-derived from a local copy of TE's formula rather than read back from the quantizer.** The copy is faithful today (I diffed it against TE's reference recipe, see below). But the block scales come from TE while the global scale comes from our reimplementation, and the two must agree *exactly* or every dequantized weight is off by a constant factor. Any upstream change — `pow_2_scales`, a different `row_scaled_nvfp4` default, a tweak to the clamp — silently desynchronizes them, and no test would catch it.

### The thing a smoke test has to prove

SGLang re-runs `process_weights_after_loading` on every quantized module after each refit (`model_runner.end_weight_update` → `postprocess_weight`). Its counterpart `restore_weight` dispatches to `restore_weights_before_loading`, and **ModelOpt defines no such hook** — so NVFP4 MoE re-finalizes over already-finalized state. Some transforms guard for this deliberately (the W13 deinterleave carries a sticky `_w13_deinterleaved` flag); the CuteDSL-v2 branch applies `interleave_w13_halves` unconditionally. Whether that corrupts depends on whether the refit fully overwrites both params first, and nothing here establishes it. The previously validated MoE hot-reload work ran on `flashinfer_trtllm`, which takes the *other* branch — so the CuteDSL path is unexercised. **This needs ≥3 refits, not 1**; a single-refit run cannot tell "correct" from "corrupts on re-finalize".

### Verified good — recording these so nobody re-litigates them

- **Offline and online quantization selection agree.** This is the property most likely to silently break an RL run, so I executed it rather than reasoning. The *raw* predicates do diverge — with `extra_high_precision_layers_hf: ["down_proj"]` the offline converter quantizes **0** tensors while the online iterator quantizes **8**, because offline skips at MoE-container granularity and online at pair granularity. But `_sync_checkpoint_ignore` merges the checkpoint's container-level `ignore` list back into `modules_to_not_convert`, and the synchronizer re-reads that config fresh from the driver-side generation object every refit, so the mutation propagates. Replaying the merge with the real functions: **4/4 cases agree.** No test pins this reconciliation, though — the two shipped tests each cover one side in isolation.
- **The TE bump is necessary, not incidental.** `nvfp4_use_4over6`, `nvfp4_e4m3_max` and `nvfp4_4over6_err_mode` are all new in v2.17 (v2.16 adds only `row_scaled_nvfp4`), so the `>=2.17` requirement in the docstring is exactly right. Tag `v2.17` resolves to `2e559f06…`, matching the lock; `VERSION.txt` says `2.17.0`, matching the hand-declared metadata; `v2.17.0` does not exist, so `@v2.17` is the correct spelling. TE's `requires-dist` is unchanged between 2.15 and 2.17, so the 12-line lock diff is correct rather than suspicious. No second, now-stale TE pin exists anywhere in the repo. v2.16 is the only breaking release crossed and the repo uses none of what it removed. Moving from the `release_v2.15` *branch* to the `v2.17` *tag* is a real improvement — `release_v2.17` has already drifted past its own release.
- **The Transformer Engine contract holds.** Two things here could have been quietly catastrophic, so I checked both against upstream rather than trusting the shape asserts. `_rowwise_scale_inv` is a plain row-major block-scale matrix, not a swizzled/tiled one (`with_gemm_swizzled_scales` defaults `False`), so `[:num_rows, :num_cols // 16]` reads the region it intends to. And `_amax_rowwise` is a **per-tensor scalar** here, so `amax_rowwise.reshape(-1)[0]` is right: TE's reference recipe only reshapes the global amax to a per-row `[m, 1, 1]` under `row_scaled_nvfp4`, and this PR passes `row_scaled_nvfp4=False`. `nvfp4_global_encode_scale_te` also reproduces TE's formula exactly — same `e4m3_max * 6.0 / amax`, same `finfo.max` clamp, same zero→one guard, same reciprocal for the decode scale.
- Cross-shard deferred-pair handling is correct and genuinely tested, and the emitted companion tensors have the right ranks and dtypes.

### Two build notes worth a line in the PR description

- **TE v2.17.1 was available before this lock was regenerated.** Published 2026-08-07; it fixes checkpoint "extra state" being unpickled for stateless recipes and adds an explicit env-var guard. A framework that loads checkpoints routinely probably wants it. Caveat: the fix introduces an opt-in env var, so it may not be purely drop-in.
- **NeMo-Automodel declares `transformer-engine[pytorch]<=2.11.0`**, and NeMo-RL requests `nemo-automodel[moe]`. The global `override-dependencies` already overshot that ceiling at 2.15; this widens it from +4 to +6 minors. Pre-existing, not caused here, but crossing 2.16 makes it worth stating.

### Tests and docs

No test exercises any real quantization math — every one monkeypatches the leaf quantizer. Three mutations survive the full suite undetected: `nvfp4_global_decode_scale_te` → constant `1.0`; `nvfp4_weight_e4m3_max` → hardcoded `448` (ignoring `NVTE_NVFP4_4OVER6*` entirely); `_contiguous_pair_view` → always `None`. All three are pure CPU functions that need neither a GPU nor Transformer Engine, so this gap is closable today.

Relatedly, `NVTE_NVFP4_4OVER6*` is undeclared cross-process configuration that scales every emitted `weight_scale_2`: offline conversion runs on the driver, online refit in the Megatron worker, and nothing checks the two saw the same environment.

There is no documentation for SGLang rollout weight precision anywhere — not for `nvfp4`, and not for `mxfp8` below it. `docs/fp8.md` is *not* the natural home (different backend, different mechanism); `docs/guides/quantization-aware-rl.md` is, and it already tabulates verified configurations. Which raises the point worth making explicitly: **NeMo-RL already has a real-quant NVFP4 rollout lane** — ModelOpt + vLLM, `policy.generation.real_quant: true` — with published convergence results on the very models this targets, including a 300-step Qwen3-30B-A3B W4A4 run and a Nano3 W4A16 run. That lane is vLLM-only so an SGLang implementation is legitimately needed, but this PR adds a second, independent NVFP4 packing implementation without saying how the two relate or why the existing one couldn't be reused. Its `real_quant_ignore` list is also the existing precedent for the skip-list design here. Separately, `docs/guides/refit.md:26` now describes constraints that quantized SGLang violates.

### Environment caveats

No GPU, and `sglang`/`flashinfer`/`transformer_engine` are not installed, so **nothing involving their runtime behaviour was executed by anyone** — SGLang and TE findings are traced against fetched upstream sources at the pinned revisions, and quoted inline. `pytest` cannot run the repo suite here (`tests/unit/conftest.py` imports `ray`), so test executions used a stubbed module loader; the counts above are real runs of the real test bodies. Nothing in this PR has run on hardware, which the PR body already says.

_Generated by Claude Code_

nemo_rl/models/policy/workers/megatron_sglang_weight_iterator.py
Kh4L: Kept the guard with the supported-model invariant: Megatron Bridge emits gate/up consecutively from the same fused tensor, and startup conversion validates name completeness. Guaranteeing zero prior sends would require a full model prepass/materialization or transactional rollback; all current refit transports are non-transactional on late failures. We will cover that as a transport-level follow-up rather than materialize a full Nemotron export here.

nemo_rl/models/generation/sglang/nvfp4_quantization_core.py
Kh4L: Keeping the ValueError intentionally. The explicit checkpoint ignore path returns before this check; an unskipped unaligned expert means the NVFP4 startup layout and live-refit layout disagree. Passing that tensor through in BF16 would send BF16 into an NVFP4-loaded module, so failing loudly is safer.

nemo_rl/models/generation/sglang/nvfp4_quantization_core.py
Kh4L: Fixed on the latest head. Plain BF16 sources no longer gain KV quantization metadata. Source KV metadata is preserved, and hf_quant_config advertises FP8 only for the exact float/8-bit scheme recognized by the pinned SGLang revision. Added focused regressions.

nemo_rl/models/generation/sglang/nvfp4_setup.py
Kh4L: Fixed safely on the latest head. hf-only external NVFP4 metadata is detected before conversion/cache creation and gets a clear unsupported-layout error. Reusing stock ModelOpt checkpoints is unsafe here because they can quantize dense/attention tensors while this online refit quantizes MoE experts only; NeMo-RL-converted flat-config checkpoints remain reusable.

nemo_rl/models/generation/sglang/nvfp4_setup.py
Kh4L: External ModelOpt exclude_modules are no longer ingested: hf-only stock layouts are rejected before reuse, while NeMo-RL-generated reusable checkpoints carry the flat config ignore list used by this path. Glob matching is therefore outside the supported reuse path.

nemo_rl/models/generation/sglang/nvfp4_quantization_core.py
Kh4L: Fixed on the latest head with the canonical-stride guard and a regression reproducing the singleton-row padded-stride case. Unsafe layouts now take the torch.cat fallback.

nemo_rl/models/generation/sglang/nvfp4_quantization_core.py
Kh4L: Thanks for verifying the numerical contract against pinned Transformer Engine. The formula remains unchanged; the pinned-version contract and the real NVFP4 hardware smoke will cover this path.

Kh4L:
Exact-head hardware validation passed on HSG.

- Commit: `736472843ba46eca4c16f127009452bdb148f090`
- Exact image digest: `sha256:c8c54cfe061113a08e546876e353309271fad8e8c4060328c2758fe5cb46df55`
- Job `6593575`: `COMPLETED 0:0` in 11m46s, 3 nodes / 12 GB200 GPUs
- Real Nemotron-H Megatron training + TP4 SGLang NVFP4 rollout/refit
- 4/4 optimizer steps and 4/4 broadcast refits passed
- Positive tokens/gradients; quantized-checkpoint structure passed (25,089 weights and 6,192 complete NVFP4 companion groups)
- Result marker: `PR3615_NEMOTRON_NVFP4_4STEP_OK`

This directly covers the repeated CuteDSL re-finalization concern: all four consecutive online refits completed successfully on the exact PR image.

Kh4L:
Current head `7d429c388bdaade042f6aa0e108abd289fddc6b1` is restacked on refreshed #3614 and adds the NVFP4 declaration to the failing runner-backend fixture. Compared with HSG-validated `736472843ba46eca4c16f127009452bdb148f090`, only five SGLang unit-test fixtures changed, so job `6593575` remains applicable to the runtime tree. Fresh CI is triggered on the exact head.

PR #3618, opened 2026-08-13
fix: make nccl_reshard_refit_info wire-safe for the generation-worker.

youngeunkwon0405:
I am separating PR #3532 into two. Wire-safe part and the MTP gating part.

# What does this PR do ?

Make the nccl_reshard_refit_info object wire-safe before sending it to the generation worker during initialization.
nccl_reshard_refit_info contains torch tensors, created by Megatron. The problem is that Megatron replaces the
torch.storage._load_from_bytes function that is required for unpickling to a Megatron-specific function.
So, during unpickling from the generation worker, it will call import megatron which will cause an import error.
This PR fixes that issue.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -236,9 +236,13 @@ class SingleControllerActor:
             rollout_task.cancel()
             train_task.cancel()
             await asyncio.gather(rollout_task, train_task, return_exceptions=True)
-            self._logger.finish()
-            # Flush the last checkpoint's background finalization before exit.
-            await asyncio.to_thread(self._checkpointer.shutdown)
+            try:
+                self._weight_synchronizer.shutdown()
+            except Exception as e:  # teardown must not mask the original failure
+                print(f"Error during weight-synchronizer shutdown: {e}", flush=True)
+            finally:
+                self._logger.finish()
+                await asyncio.to_thread(self._checkpointer.shutdown)
 
         return {
             "train_steps": self._train_steps,
--- a/nemo_rl/weight_sync/nccl_reshard_utils.py
+++ b/nemo_rl/weight_sync/nccl_reshard_utils.py
@@ -19,6 +19,8 @@ This module provides:
   shared torch.distributed process group (needed for cross-world transfers)
 - Placement rules: mapping param names to TP/EP sharding strategies
 - build_nccl_reshard_refit_info: compute per-layer param metadata for refit
+- make_nccl_reshard_refit_info_wire_safe: convert placements and meshes into
+  plain dicts/lists before the refit metadata crosses a process boundary
 - restore_refit_info_placements: undo msgspec dict-flattening of placements
   and meshes on the receiving side
 
@@ -245,6 +247,45 @@ def _restore_placement(p):
             return Shard(p["dim"])
         return Replicate()
     return Replicate()
+
+
+def make_nccl_reshard_refit_info_wire_safe(refit_info: dict) -> dict:
+    """Copy refit metadata into types safe for vLLM's subprocess RPC.
+
+    Importing ``megatron.core`` replaces ``torch.storage._load_from_bytes`` with
+    a Megatron function, so Tensor pickles require Megatron at unpickle time.
+    vLLM subprocesses may not have it importable; convert the metadata to the
+    plain representation accepted by ``restore_refit_info_placements``.
+    """
+
+    def _wire_mesh(mesh):
+        if isinstance(mesh, MeshInfo):
+            return {"mesh": mesh.mesh.tolist()}
+        return mesh
+
+    def _wire_placement(placement):
+        if isinstance(placement, Shard):
+            return {"dim": placement.dim}
+        if isinstance(placement, Replicate):
+            return {}
+        return placement
+
+    wire_info = dict(refit_info)
+    wire_layers = {}
+    for layer_name, params in refit_info.get("per_layer_params", {}).items():
+        wire_params = []
+        for param_info in params:
+            wire_param = dict(param_info)
+            for key in ("src_mesh_info", "dst_mesh_info"):
+                if key in wire_param:
+                    wire_param[key] = _wire_mesh(wire_param[key])
+            for key in ("src_placements", "dst_placements"):
+                if key in wire_param:
+                    wire_param[key] = [_wire_placement(p) for p in wire_param[key]]
+            wire_params.append(wire_param)
+        wire_layers[layer_name] = wire_params
+    wire_info["per_layer_params"] = wire_layers
+    return wire_info
 
 
 def restore_refit_info_placements(refit_info: dict) -> dict:
--- a/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
+++ b/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
@@ -43,6 +43,9 @@ import ray
 
 from nemo_rl.utils.timer import Timer
 from nemo_rl.weight_sync.interfaces import WeightSynchronizer
+from nemo_rl.weight_sync.nccl_reshard_utils import (
+    make_nccl_reshard_refit_info_wire_safe,
+)
 
 
 class NcclReshardWeightSynchronizer(WeightSynchronizer):
@@ -205,9 +208,22 @@ class NcclReshardWeightSynchronizer(WeightSynchronizer):
             train_world_size,
             inference_world_size,
         )
-        self._generation.prepare_nccl_reshard_refit_info(nccl_reshard_refit_info)
+
+        # nccl_reshard_refit_info holds MeshInfo rank tensors created under
+        # Megatron, whose pickles resolve a Megatron-patched storage loader and
+        # therefore need `import megatron` on unpickle. Convert them to plain
+        # lists here; the vLLM worker rebuilds them in
+        # `restore_refit_info_placements()`.
+        wire_refit_info = make_nccl_reshard_refit_info_wire_safe(
+            nccl_reshard_refit_info
+        )
+        self._generation.prepare_nccl_reshard_refit_info(wire_refit_info)
 
     def shutdown(self) -> None:
         # The NCCL process groups' lifecycle is managed by Ray actor teardown;
         # the workers that own the groups are destroyed with the cluster.
-        pass
+        # Break the VllmGeneration <-> synchronizer reference cycle so the
+        # generation wrapper is garbage-collectable after teardown. The
+        # synchronizer is never used again after shutdown(), so losing the
+        # handle is safe.
+        self._generation = None
--- a/tests/unit/weight_sync/test_nccl_reshard_utils.py
+++ b/tests/unit/weight_sync/test_nccl_reshard_utils.py
@@ -21,6 +21,7 @@ mesh construction, placement rules, expert grouping, and the top-level
 torch.distributed, no model object — so this module runs on CPU with no extras.
 """
 
+import pickle
 from types import SimpleNamespace
 
 import pytest
@@ -38,6 +39,8 @@ from nemo_rl.weight_sync.nccl_reshard_utils import (
     group_expert_params_in_metadata,
     is_expert_param,
     is_nccl_reshard_param,
+    make_nccl_reshard_refit_info_wire_safe,
+    restore_refit_info_placements,
 )
 
 
@@ -476,3 +479,81 @@ def test_build_refit_info_groups_experts_and_tags_them():
         for layer in info["layer_names"]
         for p in info["per_layer_params"][layer]
     )
+
+
+# --------------------------------------------------------------------------
+# make_nccl_reshard_refit_info_wire_safe / restore_refit_info_placements
+# --------------------------------------------------------------------------
+def _contains_tensor(obj) -> bool:
+    if isinstance(obj, torch.Tensor):
+        return True
+    if isinstance(obj, dict):
+        return any(_contains_tensor(v) for v in obj.values())
+    if isinstance(obj, (list, tuple)):
+        return any(_contains_tensor(v) for v in obj)
+    if hasattr(obj, "__dict__"):  # e.g. MeshInfo wrapping its rank tensor
+        return any(_contains_tensor(v) for v in vars(obj).values())
+    return False
+
+
+def _refit_info_for_wire() -> dict:
+    return build_nccl_reshard_refit_info(
+        _dense_metadata(),
+        train_parallelism={"tp_size": 2, "ep_size": 1, "pp_size": 1},
+        gen_parallelism={"tp_size": 4, "ep_size": 1, "pp_size": 1},
+        train_world_size=2,
+        gen_world_size=4,
+    )
+
+
+def test_wire_safe_refit_info_carries_no_tensors_and_pickles_plainly():
+    # MeshInfo holds a rank tensor; a tensor pickled in a megatron-importing
+    # process needs `import megatron` at unpickle time, so the wire form must
+    # be tensor-free end to end (the invariant this function exists for).
+    info = _refit_info_for_wire()
+    assert _contains_tensor(info)
+
+    wire = make_nccl_reshard_refit_info_wire_safe(info)
+
+    assert not _contains_tensor(wire)
+    # Placements are encoded exactly as restore expects: Shard -> {"dim": d},
+    # Replicate -> {}.
+    for params in wire["per_layer_params"].values():
+        for p in params:
+            assert isinstance(p["src_mesh_info"], dict)
+            assert isinstance(p["dst_mesh_info"], dict)
+            for placement in p["src_placements"] + p["dst_placements"]:
+                assert isinstance(placement, dict)
+    pickle.loads(pickle.dumps(wire))
+
+
+def test_wire_safe_does_not_mutate_the_train_side_refit_info():
+    info = _refit_info_for_wire()
+
+    make_nccl_reshard_refit_info_wire_safe(info)
+
+    for layer in info["layer_names"]:
+        for p in info["per_layer_params"][layer]:
+            assert isinstance(p["src_mesh_info"], MeshInfo)
+            assert isinstance(p["dst_mesh_info"], MeshInfo)
+            for placement in p["src_placements"] + p["dst_placements"]:
+                assert isinstance(placement, (Shard, Replicate))
+
+
+def test_wire_safe_then_restore_reproduces_placements_and_meshes():
+    info = _refit_info_for_wire()
+
+    wire = pickle.loads(pickle.dumps(make_nccl_reshard_refit_info_wire_safe(info)))
+    restored = restore_refit_info_placements(wire)
+
+    assert restored["layer_names"] == info["layer_names"]
+    for layer in info["layer_names"]:
+        original = {p["name"]: p for p in info["per_layer_params"][layer]}
+        assert len(restored["per_layer_params"][layer]) == len(original)
+        for p in restored["per_layer_params"][layer]:
+            o = original[p["name"]]
+            for key in ("src_mesh_info", "dst_mesh_info"):
+                assert isinstance(p[key], MeshInfo)
+                assert torch.equal(p[key].mesh, o[key].mesh)
+            for key in ("src_placements", "dst_placements"):
+                assert p[key] == o[key]
--- a/tests/unit/weight_sync/test_nccl_reshard_utils.py
+++ b/tests/unit/weight_sync/test_nccl_reshard_utils.py
@@ -557,3 +557,31 @@ def test_wire_safe_then_restore_reproduces_placements_and_meshes():
                 assert torch.equal(p[key].mesh, o[key].mesh)
             for key in ("src_placements", "dst_placements"):
                 assert p[key] == o[key]
+
+
+def test_wire_safe_pickle_is_independent_of_a_patched_storage_loader(monkeypatch):
+    # megatron.core replaces torch.storage._load_from_bytes at import time; a
+    # tensor pickled under the patch records the patcher's module path and
+    # cannot be unpickled without it. Mimic the patch with a dummy module and
+    # assert only the raw form references it — the invariant the wire-safe
+    # conversion exists to guarantee.
+    import io
+    import sys
+    import types
+
+    dummy = types.ModuleType("dummy_unpickler_module")
+
+    def _dummy_load_from_bytes(b):
+        return torch.load(io.BytesIO(b), weights_only=True)
+
+    _dummy_load_from_bytes.__module__ = "dummy_unpickler_module"
+    _dummy_load_from_bytes.__qualname__ = "_dummy_load_from_bytes"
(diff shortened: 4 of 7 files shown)


youngeunkwon0405:
/ok to test [e0c8961](https://github.com/NVIDIA-NeMo/RL/pull/3618/commits/e0c89613aabc0baf20a401bea072f80e47041725)

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -247,6 +249,45 @@ def _restore_placement(p):
...
+        if isinstance(mesh, MeshInfo):
+            return {"mesh": mesh.mesh.tolist()}
+        return mesh
+
+    def _wire_placement(placement):
+        if isinstance(placement, Shard):
+            return {"dim": placement.dim}
+        if isinstance(placement, Replicate):
+            return {}
+        return placement
```
youngeunkwon0405, line 271: [`nemo_rl/weight_sync/nccl_reshard_utils.py:261-271`](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/nemo_rl/weight_sync/nccl_reshard_utils.py#L261-L271)

**[GUIDELINE]** `_wire_mesh` / `_wire_placement` silently pass through unknown types. The function's contract is "nothing tensor-bearing survives", so a future placement type (e.g. `_StridedShard`, `Partial`) would ship its tensors and resurface as exactly the obscure vLLM-worker unpickle error this PR exists to prevent — fail-loud would catch it at the source. One subtlety when tightening this: plain `dict` inputs should still pass through unchanged so the function stays idempotent (double-convert remains a fixed point, mirroring `restore_refit_info_placements`' documented idempotency) — i.e. return dicts as-is, `raise TypeError(f"unexpected mesh type {type(mesh)} in refit info")` for anything else. A one-line `make_nccl_reshard_refit_info_wire_safe(wire) == wire` assert in the existing tests would pin the idempotency contract alongside it.

nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
```diff
@@ -205,9 +208,22 @@ def init_communicator(self) -> None:
...
 
     def shutdown(self) -> None:
         # The NCCL process groups' lifecycle is managed by Ray actor teardown;
         # the workers that own the groups are destroyed with the cluster.
-        pass
+        # Break the VllmGeneration <-> synchronizer reference cycle so the
+        # generation wrapper is garbage-collectable after teardown. The
+        # synchronizer is never used again after shutdown(), so losing the
+        # handle is safe.
+        self._generation = None
```
youngeunkwon0405, line 229: [`nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py:222-229`](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py#L222-L229)

**[BUG, low]** Narrow teardown race with an in-flight sync: [`_sync_weights` runs the synchronizer in a non-cancellable `asyncio.to_thread`](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/nemo_rl/algorithms/single_controller.py#L692-L695), and cancelling the awaiting task lets `run()`'s finally proceed while the worker thread is still between [lines 118-119 here](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py#L118-L119). `shutdown()` nulling `_generation` in that window makes the live thread raise `AttributeError`, skip `ray.get(futures_train)`, and leave train workers blocked in the NCCL collective — masking the original failure. The line itself is worth keeping (on the sync-GRPO path [`VllmGeneration.shutdown()` calls this](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/nemo_rl/models/generation/vllm/vllm_generation.py#L940-L941) and a real reference cycle exists there); a one-line hardening is snapshotting `generation = self._generation` at the top of `sync_weights()` and using the local.

Related test note: `NcclReshardWeightSynchronizer` has no test file, while both sibling synchronizers cover `shutdown()` ([test_checkpoint_engine_weight_synchronizer.py:249](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/tests/unit/weight_sync/test_checkpoint_engine_weight_synchronizer.py#L249), [test_vllm_remote_sparse_weight_synchronizer.py:191](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/tests/unit/weight_sync/test_vllm_remote_sparse_weight_synchronizer.py#L191)) — a minimal shutdown-idempotency test (the module imports fine on CPU) was verified passing during review.

nemo_rl/algorithms/single_controller.py
```diff
@@ -215,7 +215,12 @@ async def run(self) -> dict[str, Any]:
             rollout_task.cancel()
             train_task.cancel()
             await asyncio.gather(rollout_task, train_task, return_exceptions=True)
-            self._logger.finish()
+            try:
+                self._weight_synchronizer.shutdown()
+            except Exception as e:  # teardown must not mask the original failure
+                print(f"Error during weight-synchronizer shutdown: {e}", flush=True)
+            finally:
+                self._logger.finish()
```
youngeunkwon0405, line 244: [`nemo_rl/algorithms/single_controller.py:218-223`](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/nemo_rl/algorithms/single_controller.py#L218-L223)

**[BUG, low]** This call closes a real gap — the single-controller path previously never invoked the checkpoint-engine/sparse synchronizers' `shutdown()` at all — but it can also hang: `CheckpointEngineWeightSynchronizer.shutdown()` does an unbounded [`ray.get` against policy + generation workers](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/nemo_rl/weight_sync/checkpoint_engine_weight_synchronizer.py#L250-L257), and on the failure path those workers may be wedged rather than dead. A hang isn't covered by `except Exception`, so `run()` never returns and the driver's own cleanup never executes. Consider a bounded `ray.get(..., timeout=...)` in that synchronizer (or a timeout around this call). Two small things to fold in while here: `await asyncio.to_thread(self._weight_synchronizer.shutdown)` would match this file's convention for blocking synchronizer calls (L685, L692), and `{e!r}` or `traceback.print_exc()` would preserve the traceback of a teardown error that, by construction, nobody else reports.

Test note: no unit test calls `SingleControllerActor.run()`, so this teardown contract is uncovered. Three tests using the existing `object.__new__(controller_cls)` pattern ([test_single_controller.py:186](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/tests/unit/single_controller/test_single_controller.py#L186)) were drafted and verified during review — including the key one: a `_train_pump` failure still propagates when `shutdown()` also raises.

tests/unit/weight_sync/test_nccl_reshard_utils.py
```diff
@@ -476,3 +479,81 @@ def test_build_refit_info_groups_experts_and_tags_them():
...
+
+
+def _refit_info_for_wire() -> dict:
+    return build_nccl_reshard_refit_info(
+        _dense_metadata(),
+        train_parallelism={"tp_size": 2, "ep_size": 1, "pp_size": 1},
+        gen_parallelism={"tp_size": 4, "ep_size": 1, "pp_size": 1},
+        train_world_size=2,
+        gen_world_size=4,
+    )
```
youngeunkwon0405, line 506: [`tests/unit/weight_sync/test_nccl_reshard_utils.py:499-506`](https://github.com/NVIDIA-NeMo/RL/blob/e0c89613aabc0baf20a401bea072f80e47041725/tests/unit/weight_sync/test_nccl_reshard_utils.py#L499-L506)

**[TEST]** Two ways these (good) tests could pin the invariant harder:

1. **The fixture never produces a `Replicate`.** Both meshes here are 1-D pure-TP, so every placement is a `Shard` — the `Replicate -> {}` encoding (a *falsy* empty dict, exactly what a future truthiness guard would silently drop) and multi-dim `mesh.tolist()` nesting are never exercised, nor are `pp_stage` / `grouped_expert_proj` pass-through. One consolidated MoE+EP+PP case (e.g. tp 2→4 with ep_size=2, pp_size=2 and a `layer_to_pp_stage` map) covers all of it — verified passing during review: the 2-D mesh yields `[Replicate(), Shard(0)]`.

2. **`pickle.loads(pickle.dumps(wire))` doesn't actually prove megatron-independence** — this test process has no patched unpickler, so that line passes even for the raw tensor-bearing info. The invariant the PR exists for can be pinned directly (~10 lines): monkeypatch `torch.storage._load_from_bytes` with a function from a dummy module (mimicking [megatron-core's patch](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/__init__.py)), then assert the dummy module name appears in `pickle.dumps(raw_info)` but NOT in `pickle.dumps(wire_info)`. Verified to pass/fail correctly against the real code during review.

youngeunkwon0405 (commented):
Team review: 4 agents (RL expert, test reviewer, bug finder, devil's advocate) + lead; findings independently re-verified before posting.

**The fix is correct and necessary — verified empirically.** The devil's advocate reproduced the failure end-to-end on CPU: with `torch.storage._load_from_bytes` patched the way the pinned megatron-core does ([megatron/core/__init__.py @ d12f6c8c](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/__init__.py), `torch.storage._load_from_bytes = safe_load_from_bytes`), the raw refit-info pickle references `megatron.core.safe_globals.safe_load_from_bytes` and cannot unpickle without megatron, while the wire-safe form unpickles and restores exactly. The converter was further verified for MoE+EP+PP round-trip fidelity, idempotency in both directions, and non-mutation of the caller's dict. All 50 unit tests in the touched file pass on a CPU-only environment.

4 inline comments, none above low severity: a fail-loud suggestion in the wire helpers, a narrow teardown race + a hang risk around the new `shutdown()` wiring (which otherwise closes a real gap), and two test-strengthening suggestions. Linter (ruff check + format): clean.

_Generated by Claude Code_

nemo_rl/weight_sync/nccl_reshard_utils.py
youngeunkwon0405: This is fine

nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
youngeunkwon0405: Not a realistic case.

nemo_rl/weight_sync/nccl_reshard_utils.py
youngeunkwon0405: Agreed — the guarded types are the only ones `get_placements` can emit today, so the fail-loud variant guards an unreachable case. Declining as too defensive; resolving.

nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
youngeunkwon0405: Agreed — the window requires a rollout failure, an in-flight sync, and shutdown landing between two specific lines, during a teardown where the actors are killed moments later. Declining as not a realistic case; resolving.

nemo_rl/algorithms/single_controller.py
youngeunkwon0405: Declining this one as too defensive.

tests/unit/weight_sync/test_nccl_reshard_utils.py
youngeunkwon0405: Adopted the patched-unpickler suggestion in 234e68b633c4b9acb85d889eb70339d756b06ff3 — `test_wire_safe_pickle_is_independent_of_a_patched_storage_loader` mimics megatron-core's `torch.storage._load_from_bytes` patch with a dummy module and asserts it appears in the raw pickle but not the wire-safe one (verified passing, 51/51 locally). Declining the MoE/PP fixture extension: the converter treats `Replicate` identically to `Shard` (one `isinstance` branch), and the new test pins the actual tensor-independence invariant directly, so the dense case gives sufficient signal.

youngeunkwon0405: Not a realistic case. Added a test case for the #2.

nemo_rl/algorithms/single_controller.py
```diff
@@ -236,9 +236,13 @@ async def run(self) -> dict[str, Any]:
             rollout_task.cancel()
             train_task.cancel()
             await asyncio.gather(rollout_task, train_task, return_exceptions=True)
-            self._logger.finish()
-            # Flush the last checkpoint's background finalization before exit.
-            await asyncio.to_thread(self._checkpointer.shutdown)
+            try:
+                self._weight_synchronizer.shutdown()
```
terrykong, line 240: [`nemo_rl/algorithms/single_controller.py:240`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/nemo_rl/algorithms/single_controller.py#L240)

**1 action item.**

**TL;DR — this new line raises `AttributeError` on every existing test that drives `run()`; the broad `except` swallows it and the tests still pass, so CI is green while permanently exercising the error path.**

PR-introduced. The test double has no `shutdown()`:

1. [`_FakeWeightSynchronizer`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/single_controller/test_sc_checkpointing.py#L223) defines only `__init__` and `sync_weights` — no `shutdown`, no base class.
2. It is injected as the synchronizer at [line 384](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/single_controller/test_sc_checkpointing.py#L384): `weight_synchronizer=_FakeWeightSynchronizer(),  # type: ignore[arg-type]`. The `type: ignore` is the tell that it has drifted from the ABC, where [`shutdown` is `@abstractmethod`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/nemo_rl/weight_sync/interfaces.py#L131).
3. [`_run_actor_run`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/single_controller/test_sc_checkpointing.py#L423) drives `await actor.run()`, which reaches this line in the `finally`.
4. `AttributeError` is a subclass of `Exception`, so [line 241](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/nemo_rl/algorithms/single_controller.py#L241) catches it and prints `Error during weight-synchronizer shutdown: '_FakeWeightSynchronizer' object has no attribute 'shutdown'`.

The three tests at [1115](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/single_controller/test_sc_checkpointing.py#L1115), [1177](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/single_controller/test_sc_checkpointing.py#L1177) and [1203](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/single_controller/test_sc_checkpointing.py#L1203) therefore pass without ever exercising a successful shutdown.

### AI-1
**Action:** add `shutdown()` to `_FakeWeightSynchronizer` and assert it in one test. Not a `suggestion` block because the edit is in a file outside this diff.

```python
class _FakeWeightSynchronizer:
    def __init__(self) -> None:
        self.sync_count = 0
        self.shutdown_count = 0

    def sync_weights(self, *, kv_scales: Any = None) -> None:
        self.sync_count += 1

    def shutdown(self) -> None:
        self.shutdown_count += 1
```

then `assert actor_args.weight_synchronizer.shutdown_count == 1` in one of the `_run_actor_run` tests.

_Not verified — we could not run pytest (macOS; `uv.lock` supports linux only)._

nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
```diff
@@ -205,9 +208,22 @@ def init_communicator(self) -> None:
...
+
+        # nccl_reshard_refit_info holds MeshInfo rank tensors created under
+        # Megatron, whose pickles resolve a Megatron-patched storage loader and
+        # therefore need `import megatron` on unpickle. Convert them to plain
+        # lists here; the vLLM worker rebuilds them in
+        # `restore_refit_info_placements()`.
+        wire_refit_info = make_nccl_reshard_refit_info_wire_safe(
+            nccl_reshard_refit_info
+        )
+        self._generation.prepare_nccl_reshard_refit_info(wire_refit_info)
```
terrykong, line 220: [`nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py:217-220`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py#L217-L220)

**1 action item.**

**TL;DR — delete these three lines and all four new tests still pass, so nothing guards the fix itself.**

PR-introduced gap. The four new tests all call [`make_nccl_reshard_refit_info_wire_safe`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/nemo_rl/weight_sync/nccl_reshard_utils.py#L252) directly; none covers this call site. `grep -rn NcclReshardWeightSynchronizer tests/` returns nothing, so the class has no coverage at all — reverting the conversion here brings the original vLLM-worker unpickle failure back with CI green.

There is direct precedent to copy: [`tests/unit/weight_sync/test_weight_synchronizer.py`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/weight_sync/test_weight_synchronizer.py) already has a mock-based `test_init_communicator*` for IPC (L183), HTTP (L293) and [Collective (L397)](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/weight_sync/test_weight_synchronizer.py#L397), with reusable `_mock_policy` / `_mock_generation` / `_mock_cluster` fixtures at L43/58/73.

### AI-1
**Action:** add a `TestNcclReshardWeightSynchronizer` to that file — a `@patch(...ray)` test asserting the object handed to `generation.prepare_nccl_reshard_refit_info` is tensor-free, plus a two-line `shutdown()` test asserting `sync._generation is None`. Not a `suggestion` block because the file is outside this diff. Sketch — one way of several:

```python
NcclReshardWeightSynchronizer(policy, gen, train_cluster, inference_cluster).init_communicator()

(shipped,), _ = gen.prepare_nccl_reshard_refit_info.call_args
for params in shipped["per_layer_params"].values():
    for p in params:
        assert isinstance(p["src_mesh_info"], dict)
        assert isinstance(p["dst_mesh_info"], dict)
```

_Not verified — we could not run pytest (macOS; `uv.lock` supports linux only), so the mock wiring may need a nudge._

terrykong (commented):
Team review: 5 agents + 3 independent adversarial passes.

**The fix is correct and complete.** Re-verified independently: megatron-core does patch the storage unpickler at import ([`megatron/core/__init__.py:58`](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/__init__.py#L58)), so the premise is real; the only tensor-bearing values are `MeshInfo.mesh` under `per_layer_params` (`misc_meta` is `shape`/`dtype` scalars), so the shallow copy misses nothing; the encodings match `restore_refit_info_placements` exactly and int64 round-trips through `tolist()`. Reusing the encoding vLLM's msgspec already emits, so the decoder needed zero changes, is a good call.

**2 inline comments, both low severity** — test-coverage gaps only. Nothing blocking.

Eight other candidates were dropped after adversarial challenge, including two that looked strong at first: a claimed double-`shutdown()` (refuted — `VllmGeneration.weight_synchronizer` is `None` on this path, so the two call sites are mutually exclusive) and a claimed stale-msgspec-docstring contradiction (refuted — the sync and async engines take different transports, so the existing msgspec docstrings and this PR's pickle premise are both true).

One low-severity follow-up, not worth reopening a resolved thread: [`CheckpointEngineWeightSynchronizer.shutdown()`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/nemo_rl/weight_sync/checkpoint_engine_weight_synchronizer.py#L250) does an untimed `ray.get`, and this PR is what first puts it on the single-controller failure path where workers may be wedged; a hang there is not caught by `except Exception`. A bounded `ray.get(..., timeout=...)` would be worth a separate issue.

_Caveat: no tests or linters were run for this review — the environment is macOS and `uv.lock` supports linux only. Both test suggestions are unverified._

_Generated by Claude Code_

tests/unit/weight_sync/test_nccl_reshard_utils.py
terrykong: Confirmed at head — [`test_wire_safe_pickle_is_independent_of_a_patched_storage_loader`](https://github.com/NVIDIA-NeMo/RL/blob/a103e75522fe44dc7bba8e04edfa966443c1292c/tests/unit/weight_sync/test_nccl_reshard_utils.py#L562) pins the invariant exactly: the storage `__reduce__` path records the patched loader's module, so the positive assertion makes the test fail the moment the converter leaks a tensor rather than rotting into a vacuous pass. Agreed on skipping the MoE/PP fixture.

nemo_rl/algorithms/single_controller.py
youngeunkwon0405: Got it; this will be addressed in the next commit.

nemo_rl/weight_sync/nccl_reshard_weight_synchronizer.py
youngeunkwon0405: added a new test case to address this.

youngeunkwon0405:
/ok to test [ddab4e8](https://github.com/NVIDIA-NeMo/RL/pull/3618/commits/ddab4e86fed2bd78769867c0ec5fd4cefefe8250)

terrykong (approved):

PR #3619, opened 2026-08-13
fix: gate MTP layers from the nccl-reshard bulk refit path

youngeunkwon0405:
I am separating PR https://github.com/NVIDIA-NeMo/RL/pull/3532 into two. Wire-safe part and the MTP gating part.

# What does this PR do ?

Gate the MTP layers from the nccl-reshard path. It will force the use of the conventional refit path for the MTP layer.
The PR also contains the correct MTP gating for the DeepSeek MTP and the refactor in the misc path call to correctly update the MTP drafter in the vLLM engine.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/docs/design-docs/nccl-reshard-refit.md
+++ b/docs/design-docs/nccl-reshard-refit.md
@@ -58,6 +58,15 @@ nccl-reshard-refit implementation:
   `.weight`, dense MLP and MoE experts alike; see `is_nccl_reshard_param()`). These are
   resharded shard-to-shard with `xferdtensor` over dedicated NCCL communicators. For
   large models this covers the vast majority of the refit bytes. For the current version of implementation, it only detects `(experts).N.{gate_proj|up_proj|down_proj}` as the subject of this performant transportation path. The coverage will be expanded via future updates.
+  Two FFN-named groups are explicitly excluded and ride the misc path instead:
+  shared-expert weights (`*.shared_expert.*`, which fuse differently on the vLLM
+  side) and co-trained MTP drafter weights (which vLLM keeps in a separate
+  drafter module updated through `load_weights`). MTP weights are recognized two
+  ways: bare-`mtp.`-prefix HF names (NemotronH, Qwen3.5) via
+  `is_nccl_reshard_param()`, and DeepSeek-style MTP exported as trailing
+  `model.layers.N` indices via provenance — the Megatron-side name on the
+  Bridge conversion tasks is uniformly `mtp.*`, so the worker excludes those HF
+  layers when building the metadata (`_collect_mtp_hf_layer_names()`).
 * **Misc path** — everything else (embeddings, attention projections, layernorms, the
   MoE router, `lm_head`, FP8 `_scale_inv` siblings, FP8 KV-cache scales, …). These ride
   a packed broadcast (conventional `packed_tensor.py` implementation) over the shared
--- a/docs/design-docs/nccl-reshard-refit.md
+++ b/docs/design-docs/nccl-reshard-refit.md
@@ -39,6 +39,9 @@ single `ValueError` listing every violation. The current requirements are:
 * vLLM expert parallelism is supported with the NeMo RL convention
   `expert_parallel_size == tensor_parallel_size`. 
 * Generation-side, PP > 1 is not supported. 
+* **No ModelOpt real quantization** — `policy.generation.real_quant=false`. Real-quant
+  rollouts refit through vLLM's layerwise-reload weight loaders, which the bulk
+  `xferdtensor` writes bypass.
 
 Operational knobs:
 
@@ -64,9 +67,10 @@ nccl-reshard-refit implementation:
   drafter module updated through `load_weights`). MTP weights are recognized two
   ways: bare-`mtp.`-prefix HF names (NemotronH, Qwen3.5) via
   `is_nccl_reshard_param()`, and DeepSeek-style MTP exported as trailing
-  `model.layers.N` indices via provenance — the Megatron-side name on the
-  Bridge conversion tasks is uniformly `mtp.*`, so the worker excludes those HF
-  layers when building the metadata (`_collect_mtp_hf_layer_names()`).
+  `model.layers.N` indices via provenance — the Megatron-side name carries an
+  `mtp.` module segment (bare for LM bridges, `language_model.mtp.*` for the VL
+  and EXAONE bridges), so the worker excludes those HF layers when building the
+  metadata (`_collect_mtp_hf_layer_names()`).
 * **Misc path** — everything else (embeddings, attention projections, layernorms, the
   MoE router, `lm_head`, FP8 `_scale_inv` siblings, FP8 KV-cache scales, …). These ride
   a packed broadcast (conventional `packed_tensor.py` implementation) over the shared
--- a/nemo_rl/models/generation/vllm/vllm_backend.py
+++ b/nemo_rl/models/generation/vllm/vllm_backend.py
@@ -56,7 +56,7 @@ except ImportError:
     )
 
 
-WeightUpdateTransport = Literal["ipc", "collective"]
+WeightUpdateTransport = Literal["ipc", "collective", "nccl_reshard"]
 WeightUpdateFinalizer = Callable[[], None]
 
 
@@ -1064,33 +1064,26 @@ class VllmInternalWorkerExtension:
 
         import time
 
-        misc_t0 = time.perf_counter()
-        self._receive_and_load_misc_params()
-        torch.cuda.synchronize()
-        if torch.distributed.get_rank() == 0:
-            print(
-                f"[nccl_reshard_refit] misc recv+load (gen side): "
-                f"{time.perf_counter() - misc_t0:.2f}s",
-                flush=True,
-            )
-        torch.cuda.empty_cache()
-        from vllm.config import set_current_vllm_config
-        from vllm.model_executor.model_loader.utils import (
-            process_weights_after_loading,
-        )
+        with self._weight_update_lifecycle("nccl_reshard") as finalize:
+            misc_t0 = time.perf_counter()
+            self._receive_and_load_misc_params()
+            torch.cuda.synchronize()
+            if torch.distributed.get_rank() == 0:
+                print(
+                    f"[nccl_reshard_refit] misc recv+load (gen side): "
+                    f"{time.perf_counter() - misc_t0:.2f}s",
+                    flush=True,
+                )
+            torch.cuda.empty_cache()
 
-        # Finalize post-load weight processing: dense Linear + attention/MLA, and
-        # crucially the per-MoE-backend w13 layout (FlashInfer CUTLASS/TRTLLM) that
-        # the canonical [gate; up] bulk write above defers to here.
-        with set_current_vllm_config(self.model_runner.vllm_config):
-            process_weights_after_loading(
-                self.model_runner.model, self.model_config, self.device
-            )
+            # Finalize post-load weight processing: dense Linear + attention/MLA,
+            # the per-MoE-backend w13 layout (FlashInfer CUTLASS/TRTLLM) that the
+            # canonical [gate; up] bulk write above defers to here, and the MTP
+            # drafter's mirror of the same. The FP8 KV-cache per-layer k/v scales
+            # are finalized by the lifecycle on exit.
+            finalize()
 
-        torch.cuda.empty_cache()
-
-        # Finalize FP8 KV-cache per-layer k/v scales after the misc broadcast.
-        self._maybe_process_fp8_kv_cache()
+            torch.cuda.empty_cache()
         return True
 
     def _receive_and_load_misc_params(self) -> None:
--- a/nemo_rl/models/policy/workers/megatron_policy_worker.py
+++ b/nemo_rl/models/policy/workers/megatron_policy_worker.py
@@ -201,6 +201,39 @@ def _estimate_refit_tensor_size_in_bytes(
     return param.numel() * tp_size * ep_size * element_size
 
 
+def _collect_mtp_hf_layer_names(conversion_tasks: Optional[list]) -> set[str]:
+    """Return HF layer names whose weights originate from Megatron's MTP module.
+
+    The Megatron-side parameter name is uniformly ``mtp.*`` for every model
+    family, but the HF-side name is not: DeepSeek-style bridges export MTP as
+    trailing main-model layer indices (``model.layers.{N + num_layers}.*``),
+    which ``is_nccl_reshard_param`` cannot recognize from the name shape alone.
+    Provenance — the Megatron name carried by the Bridge conversion task — is
+    the reliable signal. Results are keyed by ``_extract_layer_name`` so a
+    whole MTP layer is routed to the misc path together.
+
+    Args:
+        conversion_tasks: Megatron-Bridge ``WeightConversionTask`` list; only
+            ``global_param_name`` (resolved Megatron-side name) and
+            ``mapping.hf_param`` (str, or dict of str for compound mappings)
+            are consulted. ``None`` entries are tolerated.
+
+    Returns:
+        Set of HF layer names, e.g. ``{"model.layers.61", "mtp.layers.0"}``.
+    """
+    from nemo_rl.weight_sync.nccl_reshard_utils import _extract_layer_name
+
+    mtp_layers: set[str] = set()
+    for task in conversion_tasks or []:
+        # FP8 export can leave None placeholders in the task list (BF16 export
+        # pre-filters them; see _build_refit_conversion_tasks).
+        if task is not None and task.global_param_name.startswith("mtp."):
+            hf = task.mapping.hf_param
+            for hf_name in hf.values() if isinstance(hf, dict) else [str(hf)]:
+                mtp_layers.add(_extract_layer_name(hf_name))
+    return mtp_layers
+
+
 @contextmanager
 def _meta_tensor_alloc_context():
     """Skip real GPU work during metadata enumeration.
@@ -2402,7 +2435,18 @@ class MegatronPolicyWorkerImpl(
         # state_dict_metadata[hf_name] -> [shape, dtype]
         # At the same time, filter the params to the misc subset (packed_broadcast path).
         # misc_meta[hf_name] -> [shape, dtype]
-        from nemo_rl.weight_sync.nccl_reshard_utils import _extract_layer_prefix
+        from nemo_rl.weight_sync.nccl_reshard_utils import (
+            _extract_layer_name,
+            _extract_layer_prefix,
+        )
+
+        # HF layers whose weights come from Megatron's MTP module. The prefix
+        # gate inside is_nccl_reshard_param only catches families whose HF
+        # names keep the bare ``mtp.`` prefix (NemotronH, Qwen3.5); DeepSeek
+        # exports MTP as trailing ``model.layers.N`` indices, so provenance is
+        # the only reliable signal. vLLM keeps the MTP drafter separate from
+        # the main model and updates it through load_weights -> misc path.
+        mtp_hf_layers_names = _collect_mtp_hf_layer_names(self.refit_conversion_tasks)
 
         layer_prefix = None
         with _meta_tensor_alloc_context():
@@ -2414,7 +2458,10 @@ class MegatronPolicyWorkerImpl(
                 _nbytes = tensor.numel() * tensor.element_size()
                 # Downsized whitelist: only FFN gate/up/down weights take the bulk
                 # nccl-reshard path; everything else -> misc (packed_broadcast).
-                if is_nccl_reshard_param(name):
+                if (
+                    is_nccl_reshard_param(name)
+                    and _extract_layer_name(name) not in mtp_hf_layers_names
+                ):
                     state_dict_metadata[name] = meta
                     _xfer_bytes += _nbytes
                     if layer_prefix is not None:
--- a/nemo_rl/models/policy/workers/megatron_policy_worker.py
+++ b/nemo_rl/models/policy/workers/megatron_policy_worker.py
@@ -203,20 +203,11 @@ def _estimate_refit_tensor_size_in_bytes(
 
 def _collect_mtp_hf_layer_names(conversion_tasks: Optional[list]) -> set[str]:
     """Return HF layer names whose weights originate from Megatron's MTP module.
-
-    The Megatron-side parameter name is uniformly ``mtp.*`` for every model
-    family, but the HF-side name is not: DeepSeek-style bridges export MTP as
-    trailing main-model layer indices (``model.layers.{N + num_layers}.*``),
-    which ``is_nccl_reshard_param`` cannot recognize from the name shape alone.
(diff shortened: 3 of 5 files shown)


youngeunkwon0405:
/ok to test [31ff9af](https://github.com/NVIDIA-NeMo/RL/pull/3619/commits/31ff9af53b3ae2e81715158f789b2007d03d4cc8)

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -191,9 +191,13 @@ def is_nccl_reshard_param(param_name: str) -> bool:
     ``load_weights`` path.
 
     Shared-expert FFN weights (``*.shared_expert.*``) are routed to misc path.
+    Co-trained MTP weights are also routed to misc because vLLM keeps the MTP
+    drafter separate from the main model and updates it through ``load_weights``.
     """
     if "shared_expert" in param_name:
         return False
+    if param_name.startswith("mtp."):
+        return False
```
youngeunkwon0405, line 204: [`nemo_rl/weight_sync/nccl_reshard_utils.py:197-200`](https://github.com/NVIDIA-NeMo/RL/blob/31ff9af53b3ae2e81715158f789b2007d03d4cc8/nemo_rl/weight_sync/nccl_reshard_utils.py#L197-L200)

**[TEST]** This new branch has zero coverage — deleting these two lines still leaves the whole suite green. The two tests added in this PR cover only the provenance helper and are mcore-gated; this prefix gate is the CPU-testable half of the fix, and the natural home is the existing parametrize table in [`test_is_nccl_reshard_param`](https://github.com/NVIDIA-NeMo/RL/blob/31ff9af53b3ae2e81715158f789b2007d03d4cc8/tests/unit/weight_sync/test_nccl_reshard_utils.py#L192), which already covers the sibling `shared_expert` exclusion. These rows use real bridge-exported names ([NemotronH](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py#L479-L495), Qwen3.5 MoE) and were verified passing in a CPU-only venv:

```python
        # MTP drafter weights keep the bare ``mtp.`` prefix (NemotronH, Qwen3.5)
        # and ride the misc path -> vLLM's separate drafter load_weights.
        ("mtp.layers.0.mixer.up_proj.weight", False),
        ("mtp.layers.0.mixer.down_proj.weight", False),
        ("mtp.layers.0.mixer.experts.3.up_proj.weight", False),
        ("mtp.layers.0.mlp.gate_proj.weight", False),
        ("mtp.layers.0.mlp.experts.gate_up_proj", False),
```

A companion `("mtp.layers.0.mixer.up_proj.weight", "mtp.layers.0")` row in the `_extract_layer_name` table would also pin the key the provenance set is compared against.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -201,6 +201,39 @@ def _estimate_refit_tensor_size_in_bytes(
...
+    mtp_layers: set[str] = set()
+    for task in conversion_tasks or []:
+        # FP8 export can leave None placeholders in the task list (BF16 export
+        # pre-filters them; see _build_refit_conversion_tasks).
+        if task is not None and task.global_param_name.startswith("mtp."):
+            hf = task.mapping.hf_param
+            for hf_name in hf.values() if isinstance(hf, dict) else [str(hf)]:
+                mtp_layers.add(_extract_layer_name(hf_name))
+    return mtp_layers
+
```
youngeunkwon0405, line 229: [`nemo_rl/models/policy/workers/megatron_policy_worker.py:204-235`](https://github.com/NVIDIA-NeMo/RL/blob/31ff9af53b3ae2e81715158f789b2007d03d4cc8/nemo_rl/models/policy/workers/megatron_policy_worker.py#L204-L235)

**[DOCSTRING, low]** The docstring's claim that the Megatron-side name is "uniformly ``mtp.*`` for every model family" (echoed in the design-doc bullet) doesn't hold at the pinned Bridge: the Qwen3.5-VL bridge builds its MTP mappings with [`megatron_prefix="language_model."`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/qwen_vl/qwen35_vl_bridge.py#L276-L278), and EXAONE-4.5 hardcodes [`language_model.mtp.layers.*`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/exaone/exaone45/exaone45_bridge.py#L311-L318) — so `startswith("mtp.")` collects nothing for those families.

Not a live bug: both families keep the bare `mtp.` prefix on the **HF** side, so the `is_nccl_reshard_param` gate already excludes them, and DeepSeek (the only family exporting MTP as trailing `model.layers.N` indices) uses the bare Megatron prefix. But the stated invariant is the wrong one, and the gate is one bridge away from a silent gap. Cheap hardening + accurate docs:

```python
_MTP_MEGATRON_RE = re.compile(r"(?:^|\.)mtp\.")
...
        if task is not None and _MTP_MEGATRON_RE.search(task.global_param_name):
```

and reword to "the Megatron-side name always contains an ``mtp.`` module segment (bare for LM bridges, ``language_model.mtp.*`` for VL/EXAONE bridges)".

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1064,33 +1064,26 @@ def _recv_one_param(param_info, group, stream):
...
-        )
-
-        # Finalize post-load weight processing: dense Linear + attention/MLA, and
-        # crucially the per-MoE-backend w13 layout (FlashInfer CUTLASS/TRTLLM) that
-        # the canonical [gate; up] bulk write above defers to here.
-        with set_current_vllm_config(self.model_runner.vllm_config):
-            process_weights_after_loading(
-                self.model_runner.model, self.model_config, self.device
-            )
+        with self._weight_update_lifecycle("nccl_reshard") as finalize:
```
youngeunkwon0405, line 1067: [`nemo_rl/models/generation/vllm/vllm_backend.py:1067`](https://github.com/NVIDIA-NeMo/RL/blob/31ff9af53b3ae2e81715158f789b2007d03d4cc8/nemo_rl/models/generation/vllm/vllm_backend.py#L1067)

**[GUIDELINE, low]** Routing nccl-reshard through `_weight_update_lifecycle` newly engages the [ModelOpt real-quant override](https://github.com/NVIDIA-NeMo/RL/blob/31ff9af53b3ae2e81715158f789b2007d03d4cc8/nemo_rl/modelopt/models/generation/vllm_quant_backend.py#L466-L519), whose pre-`yield` `initialize_layerwise_reload` assumes all weights flow inside the reload window — which the bulk xferdtensor phase ([before the `with`](https://github.com/NVIDIA-NeMo/RL/blob/31ff9af53b3ae2e81715158f789b2007d03d4cc8/nemo_rl/models/generation/vllm/vllm_backend.py#L1037-L1063)) violates by construction, and real-quant errors are [fatal](https://github.com/NVIDIA-NeMo/RL/blob/31ff9af53b3ae2e81715158f789b2007d03d4cc8/nemo_rl/modelopt/models/generation/vllm_quant_backend.py#L521-L522). [`check_nccl_reshard_refit_support`](https://github.com/NVIDIA-NeMo/RL/blob/31ff9af53b3ae2e81715158f789b2007d03d4cc8/nemo_rl/weight_sync/nccl_reshard_utils.py#L517) currently has no guard rejecting `policy.generation.real_quant`, so the combination is reachable by config. It's almost certainly already unsupported pre-PR (NVFP4-packed vLLM params vs bf16 bulk writes), so the ask is only to fail loudly at setup:

```python
    if generation.get("real_quant"):
        violations.append(
            "policy.generation.real_quant must be False "
            "(nccl_reshard_refit's bulk xferdtensor writes bypass the "
            "layerwise-reload weight loaders that ModelOpt real quant requires)."
        )
```

(The non-real-quant/fakequant branch delegates to `super()` and is unaffected.)

youngeunkwon0405 (commented):
Team review: 4 agents (RL expert, test reviewer, bug finder, devil's advocate) + lead; every finding independently re-verified against the pinned Megatron-Bridge SHA (0c565c9a) before posting.

**The fix is correct and necessary.** All three pre-PR failure modes were verified real: the `layer_prefix` assert fires for bare-`mtp.` families (NemotronH/Qwen3.5), DeepSeek's trailing-index MTP hits the per-stage KeyError, and the lifecycle switch newly runs the MTP drafter post-load processing that the old inline finalization never did. Verified clean: finalization ordering is exactly preserved vs main (independently diffed by two agents); the provenance gate operates on wildcard-resolved mappings with literal MTP layer indices, so no false or missed keys are possible for the pinned bridges; GLM-4.5's grouped MTP experts are correctly excluded by the layer-key design (a per-param gate would have missed them); and the two added tests use exact bridge-exported names.

3 inline comments, none blocking: a free CPU test-coverage gap for the new prefix gate (verified rows provided), a docstring/design-doc accuracy fix for the "uniformly mtp.*" claim (two counterexample families at the pin), and an optional setup-time guard for the real-quant combination the lifecycle change newly touches. Linter (ruff check + format): clean.

_Generated by Claude Code_

nemo_rl/weight_sync/nccl_reshard_utils.py
youngeunkwon0405: This is fine

nemo_rl/models/generation/vllm/vllm_backend.py
youngeunkwon0405: resolved

docs/design-docs/nccl-reshard-refit.md
```diff
@@ -58,6 +58,15 @@ nccl-reshard-refit implementation:
...
   large models this covers the vast majority of the refit bytes. For the current version of implementation, it only detects `(experts).N.{gate_proj|up_proj|down_proj}` as the subject of this performant transportation path. The coverage will be expanded via future updates.
+  Two FFN-named groups are explicitly excluded and ride the misc path instead:
+  shared-expert weights (`*.shared_expert.*`, which fuse differently on the vLLM
+  side) and co-trained MTP drafter weights (which vLLM keeps in a separate
+  drafter module updated through `load_weights`). MTP weights are recognized two
+  ways: bare-`mtp.`-prefix HF names (NemotronH, Qwen3.5) via
+  `is_nccl_reshard_param()`, and DeepSeek-style MTP exported as trailing
+  `model.layers.N` indices via provenance — the Megatron-side name on the
+  Bridge conversion tasks is uniformly `mtp.*`, so the worker excludes those HF
+  layers when building the metadata (`_collect_mtp_hf_layer_names()`).
```
terrykong: [`docs/design-docs/nccl-reshard-refit.md:67-69`](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/docs/design-docs/nccl-reshard-refit.md#L67-L69)

**1 action item.**

This is the other half of [#3773326791](https://github.com/NVIDIA-NeMo/RL/pull/3619#discussion_r3773326791) — the docstring side landed in 926e5a5ee, this bullet didn't.

PR-introduced. The text still says the Megatron-side name is "uniformly `mtp.*`", but your own gate stopped assuming that: 432ee4865 widened it to `".mtp." in name` because three families prefix the segment with `language_model.` —

- [`qwen35_vl_bridge.py`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/qwen_vl/qwen35_vl_bridge.py#L275-L279) — `_get_moe_mtp_mappings(megatron_prefix="language_model.")`
- [`exaone45_bridge.py`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/exaone/exaone45/exaone45_bridge.py#L311-L318) — hardcoded `megatron_param="language_model.mtp.layers.*..."`
- [`nemotron_omni_bridge.py`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/nemotron_omni/nemotron_omni_bridge.py#L275-L285) — `megatron_prefix="language_model."` applied unconditionally, MTP included

The HF-side claim just above it (bare `mtp.` prefix) is correct — only the Megatron-side clause needs the reword.

### AI-1

```suggestion
  `model.layers.N` indices via provenance — the Megatron-side name carries an
  `mtp.` module segment (bare for LM bridges, `language_model.mtp.*` for the VL
  and EXAONE bridges), so the worker excludes those HF layers when building the
  metadata (`_collect_mtp_hf_layer_names()`).
```

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -191,9 +191,13 @@ def is_nccl_reshard_param(param_name: str) -> bool:
     ``load_weights`` path.
 
     Shared-expert FFN weights (``*.shared_expert.*``) are routed to misc path.
+    Co-trained MTP weights are also routed to misc because vLLM keeps the MTP
+    drafter separate from the main model and updates it through ``load_weights``.
```
terrykong: [`nemo_rl/weight_sync/nccl_reshard_utils.py:194-195`](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/weight_sync/nccl_reshard_utils.py#L194-L195)

**1 action item — docstring wording only. The routing itself is correct; nothing is broken.**

These lines are on `is_nccl_reshard_param`, so they read as "this function routes MTP weights to misc". It only does that for half of them:

| HF name | this function returns | what sends it to misc |
|---|---|---|
| `mtp.layers.0.mlp.up_proj.weight` (NemotronH) | `False` | this function, via [L199](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/weight_sync/nccl_reshard_utils.py#L199) |
| `model.layers.61.mlp.gate_proj.weight` (DeepSeek) | **`True`** | the caller's provenance check |

The split is HF vs Megatron naming. This function only ever sees **HF** names — [L2456](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2456) iterates [`_iter_params_with_optional_kv_scales`](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2124-L2129) ("Yield exported HF parameters") and [L2267](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2267) passes `str(hf_param)`. DeepSeek's mapping ([`deepseek/common.py`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/0c565c9a063bc8578ec094b4eb488594ed8b033a/src/megatron/bridge/models/deepseek/common.py#L136-L139)) puts `mtp.` on only one side:

```python
megatron_param = f"mtp.layers.{mtp_layer}.mtp_model_layer.mlp.linear_fc1.weight"   # has mtp.
gate           = f"model.layers.{mtp_layer + num_transformer_layers}.mlp.gate_proj.weight"  # does not
```

So DeepSeek MTP is caught by the second conjunct at [L2456-2457](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2456-L2457), fed by [`_collect_mtp_hf_layer_names`](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/models/policy/workers/megatron_policy_worker.py#L204), which matches on the Megatron name at [L222](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/models/policy/workers/megatron_policy_worker.py#L222-L224).

Worth a line because the next person to read this function will take it as the full MTP rule — which is the assumption this PR is fixing.

### AI-1

```suggestion
    Bare ``mtp.``-prefixed HF names are routed to misc too, because vLLM keeps
    the MTP drafter separate and updates it through ``load_weights``. That only
    covers families whose HF names keep the ``mtp.`` prefix. DeepSeek exports
    MTP under ``model.layers.N`` HF names (the ``mtp.`` appears only on the
    Megatron side), so those return True here; the caller drops them instead,
    using the layer set from ``_collect_mtp_hf_layer_names``.
```

While you're in there, the summary line at [L185](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/weight_sync/nccl_reshard_utils.py#L185) says "iff", which had been true — before this PR the call site was a bare `if is_nccl_reshard_param(name):`. It's outside this diff hunk so I can't suggest it inline, but "Return True if" would now be accurate.

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -561,6 +565,16 @@ def check_nccl_reshard_refit_support(master_config: dict) -> None:
             "dynamic expert load balancing can change ownership afterwards)."
         )
 
+    # ModelOpt real-quant rollout holds NVFP4-packed vLLM params and refits
+    # through vLLM's layerwise-reload weight loaders; the bulk xferdtensor
+    # path writes directly into param storage, bypassing both.
+    if generation.get("real_quant"):
```
terrykong, line 575: [`nemo_rl/weight_sync/nccl_reshard_utils.py:571`](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/weight_sync/nccl_reshard_utils.py#L571)

**1 action item.**

PR-introduced. This adds a new user-facing config requirement, but the requirements list in the design doc — [`nccl-reshard-refit.md:24-41`](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/docs/design-docs/nccl-reshard-refit.md#L24-L41), which this PR also edits — still has its original 7 bullets. Someone who hits `policy.generation.real_quant must be False` finds nothing about it in the docs.

That list is an enumeration, not a sampling — every other check in this function has a bullet (bullet 4 covers the ETP/PP-layout family, bullet 5 the precision family). `real_quant` and `enable_eplb` are the only two that don't.

### AI-1
Add a bullet after [L41](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/docs/design-docs/nccl-reshard-refit.md#L41). Not a `suggestion` block — L41 is outside this PR's diff hunk, so it isn't committable from here:

```markdown
* **No ModelOpt real quantization** — `policy.generation.real_quant=false`. Real-quant
  rollouts refit through vLLM's layerwise-reload weight loaders, which the bulk
  `xferdtensor` writes bypass.
```

`enable_eplb` is missing from that list too — pre-existing, not yours, mentioning only in case you want to fold it in while you're there.

terrykong (commented):
Re-review at `926e5a5`. **The fix is correct and needed** — I could not construct a case where it breaks, and both pre-PR failure modes are real, not hypothetical:

- Bare-`mtp.` families (NemotronH/Qwen3.5): `mtp.layers.0.mixer.up_proj.weight` ends in `up_proj.weight`, so it took the bulk path, and `_extract_layer_prefix` then returned `mtp` against a `backbone` incumbent → the [`layer_prefix mismatch` assert](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2461).
- DeepSeek-style: MTP exports at `model.layers.61`, so the assert passes but `_build_layer_to_pp_stage` never emits that index → the [`KeyError`](https://github.com/NVIDIA-NeMo/RL/blob/926e5a5ee58b5a5f061cc78f09772136ef44cb2d/nemo_rl/weight_sync/nccl_reshard_utils.py#L827) whose message already names MTP as the expected cause.

Verified clean, for the record:

- **Nothing is dropped.** The gate sits on the `if` of an `if/else`; the `else` writes `misc_meta`, and because `_task_is_misc` filters against `set(misc_meta.keys())`, the MTP conversion tasks land in `_misc_conversion_tasks` too. Both halves of the misc path carry MTP, so there's no silent stale-drafter case.
- **Keying the exclusion by HF layer rather than by param was the right call.** It keeps `_task_is_misc`'s "first HF name is representative" assumption valid for compound mappings — a whole MTP layer moves to misc together, so a `GatedMLPMapping` can never end up half-bulk/half-misc and desync the packed-broadcast order.
- **The provenance gate can't go stale per-family.** The key is derived with `_extract_layer_name` from the same task's `hf_param` that produces the name being tested, so it matches by construction. Checked at the pinned Bridge: 17 families define MTP mappings, all name the Megatron module `mtp.layers.N`, and a grep for `nextn|eagle|draft|speculat` in any `megatron_param=` returns nothing.
- **GLM-4.5's grouped-MoE MTP is handled** — its HF name ends in `experts.gate_up_proj`, so `is_nccl_reshard_param` returns True and only the provenance set excludes it. A per-param gate would have missed this.
- **No over-broad exclusion is possible.** MTP names without a `layers.N` produce coarse keys (`mtp`, `model.embed_tokens`). Those can never collide with an FFN param: every FFN name sits under `layers.N` and keys as `<prefix>.layers.N`, while the fallback branches structurally cannot emit that shape.
- **The lifecycle refactor preserves ordering exactly** and adds the one thing that was missing — the deleted inline block was a hand-rolled copy of the base `finalize()` that omitted `_maybe_process_mtp_drafter_after_loading()`. Adopting the shared hook matches how `ipc` and `collective` already call it.
- **The `real_quant` guard closes the loop.** The bulk xferdtensor phase sits outside the `with`, while ModelOpt's override runs `initialize_layerwise_reload` on entry — the new violation forbids exactly that pairing. Guard and narrow `with` are a matched pair.

Reviewed by a 7-agent team; every finding was re-verified against the pinned Megatron-Bridge SHA `0c565c9a` before posting. Two candidate findings were dropped under challenge: a missing `real_quant` row in the config-validation test table (that table spot-checks 2 of 17 violations, so there's no convention being broken), and a GLM-4.5 test fixture (the existing DeepSeek fixture already exercises the provenance-only path). Linters pass; `pyrefly` couldn't run in my environment, so it's unchecked rather than failing.

Three comments below, all low severity, none blocking — and all the same shape: the routing rule changed, and one of the four places describing it got updated. No code changes requested.

_Generated by Claude Code_

terrykong (approved):

PR #3626, opened 2026-08-13
feat(logger): WanDB now shows core metrics in a separate tab for convenience

dkmisra:
## Add core metrics tab in WandB for key training signals

# What does this PR do?

Adds a `_CORE_METRIC_KEYS` frozenset in `nemo_rl/utils/logger.py` listing the key training signals: `total_reward/mean`, `mean_total_reward`, `mean_generation_length`, `gen_tokens_per_sample/mean`, `mean_gen_tokens_per_sample`, `approx_entropy`, `kl_penalty`, and `gen_kl_error`.

Extends `Logger.log_metrics` to re-log any of these keys under a `core/<prefix>` namespace (e.g. `core/train`, `core/validation`) after each normal dispatch, so they appear in a dedicated WandB "core" section without being removed from their original `train/*` / `validation/*` location. No changes to any algorithm code — the interception is at the logger level and applies to all callers (`grpo.py`, `grpo_sync.py`, etc.).

### Notes

- `kl_penalty` is always emitted by the loss function (`0` when `reference_policy_kl_penalty` is disabled), so `core/<prefix>/kl_penalty` renders as a flat-zero series on runs without a KL penalty rather than being absent.
- `mean_total_reward` / `mean_gen_tokens_per_sample` cover the classic (non-NeMo-Gym) multi-turn and single-turn rollout paths; `total_reward/mean` / `gen_tokens_per_sample/mean` cover the NeMo-Gym path. Both spellings are tracked so the core tab isn't empty on either path.

### Fixes since the initial version

- The `core/` re-log previously ignored the caller's `prefix` and always wrote to a flat `"core"` key, so train and validation values for the same metric could silently collide into one WandB series. Fixed by namespacing as `core/<prefix>` (e.g. `core/train`, `core/validation`) so they stay distinct.
- `_CORE_METRIC_KEYS` was missing the classic multi-turn/single-turn path's key names (`mean_total_reward`, `mean_gen_tokens_per_sample`) — added, so standard (non-NeMo-Gym) GRPO runs also populate the core tab's reward/length signal.

### Tests

Added to `tests/unit/utils/test_logger.py::TestLogger`:

- `test_log_metrics_relogs_core_metrics` — a metrics dict containing a core key (`kl_penalty`, `total_reward/mean`) triggers a second `log_metrics` dispatch to every backend under `core/<prefix>`, alongside the original prefixed call.
- `test_log_metrics_core_namespace_separates_train_and_validation` — logging `total_reward/mean` once with `prefix="train"` and once with `prefix="validation"` lands under distinct `core/train` and `core/validation` series rather than colliding into one.
- `test_log_metrics_without_core_metrics_logs_once` — a metrics dict with no core keys triggers exactly one dispatch (no accidental extra `core/` call), which is also a regression guard for the pre-existing `assert_called_once_with`-based tests in this file.

All 85 tests in `tests/unit/utils/test_logger.py` pass locally (`uv run pytest unit/utils/test_logger.py`), including the 3 new ones, with no regressions to the existing suite.

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/utils/logger.py
+++ b/nemo_rl/utils/logger.py
@@ -46,13 +46,15 @@ from nemo_rl.distributed.batched_data_dict import BatchedDataDict
 # Flag to track if rich logging has been configured
 _rich_logging_configured = False
 
-# Metrics that are re-logged under the "core/" prefix so they appear in a
-# dedicated WandB section alongside their normal train/* location.
+# Metrics that are re-logged under the "core/<prefix>" namespace so they appear in a
+# dedicated WandB section alongside their normal train/* or validation/* location.
 _CORE_METRIC_KEYS: frozenset[str] = frozenset(
     {
         "total_reward/mean",
+        "mean_total_reward",
         "mean_generation_length",
         "gen_tokens_per_sample/mean",
+        "mean_gen_tokens_per_sample",
         "approx_entropy",
         "kl_penalty",
         "gen_kl_error",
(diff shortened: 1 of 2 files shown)


dkmisra:
Wanted to quickly that I did verify it and it works. I can see the core tab get populated with important metrics. Used this example:

`uv run examples/run_grpo.py   --config examples/configs/grpo_smoke.yaml   logger.wandb_enabled=True   logger.wandb.name="test-wandb-core-tab"`

PR #3628, opened 2026-08-13
feat(megatron): support fractional optimizer CPU offload

seonjinn:
# What does this PR do?

Adds fractional optimizer CPU offload to the Megatron policy backend.

- Supports `optimizer_offload_fraction` values in `(0, 1]`.
- Exposes Megatron Core's `overlap_cpu_optimizer_d2h_h2d` option.
- Keeps optimizer placement owned by Megatron Core's
  `HybridDeviceOptimizer`.
- Preserves existing behavior when CPU offload is disabled.

Optimizer CPU offload requires `use_distributed_optimizer=true`. This PR does
not change activation offload or NeMo RL's generic optimizer movement path.

# Usage

```yaml
policy:
  megatron_cfg:
    optimizer:
      use_distributed_optimizer: true
      optimizer_cpu_offload: true
      optimizer_offload_fraction: 0.5
      overlap_cpu_optimizer_d2h_h2d: true
```

# Validation

- Focused tests cover config validation, Megatron Core forwarding, and worker
  lifecycle behavior.
- A CW H100 test completed 100 steps for disabled, 50%, and 100% offload
  configurations. All 400 recorded losses were finite.
- The single-GPU Qwen3-0.6B microbenchmark confirms the expected memory versus
  throughput trade-off:

| Case | Mean step time | Max sampled GPU memory |
| --- | ---: | ---: |
| Disabled | `0.752s` | `15.24 GiB` |
| 50%, no overlap | `1.187s` | `14.13 GiB` |
| 50%, overlap | `1.216s` | `14.65 GiB` |
| 100%, overlap | `1.678s` | `10.53 GiB` |

Timings use steps 11-100. This is a functional microbenchmark, not a
production-scale throughput claim.

- [x] Added focused unit/config tests.
- [x] Added user documentation and exemplar configuration.
- [x] Completed GPU functional validation.
- [x] Completed `/review-pr-team` self-review and resolved findings.


seonjinn:
Self-review (`/review-pr-team`) completed on `60b5c936d38c997a1c654e3d0c70ef3ee7205fb1`.

Resolved findings:
- Reject `overlap_cpu_optimizer_d2h_h2d=true` when optimizer CPU offload is disabled, avoiding a silent no-op.
- Added the new overlap field to in-repo config fixtures and covered both policy and value-worker optimizer-placement lifecycles.
- Documented the CPU-Adam bottleneck at high offload fractions and the transfer-overlap mitigation.

Direct Ruff, formatting, YAML parsing, Python compilation, and config validation checks pass. The remaining readiness gate is the Linux/GPU functional and off/50%/100% performance validation; the PR remains draft until that evidence is attached.

seonjinn:
# Optimizer CPU Offload A/B Results

## Result

CW job `15725145` completed four 100-step cases on one H100 GPU at the final
PR head. The primary window excludes the first ten startup steps and contains
90 samples per case.

| Case | Mean | p50 | p90 | p99 | Delta vs disabled | Policy training | Max sampled GPU memory | GPU memory saved |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Disabled | `0.752s` | `0.749s` | `0.774s` | `1.034s` | baseline | `0.752s` | `15.24 GiB` | baseline |
| 50%, no transfer overlap | `1.187s` | `1.172s` | `1.206s` | `1.656s` | `+57.75%` | `1.186s` | `14.13 GiB` | `1.11 GiB` |
| 50%, transfer overlap | `1.216s` | `1.195s` | `1.242s` | `1.614s` | `+61.61%` | `1.215s` | `14.65 GiB` | `0.59 GiB` |
| 100%, transfer overlap | `1.678s` | `1.656s` | `1.708s` | `2.124s` | `+122.94%` | `1.677s` | `10.53 GiB` | `4.71 GiB` |

The slowdown remains stable when the analysis begins at steps 21, 41, or 61.
At 50% offload, transfer overlap was `2.45%` slower than no overlap over steps
11-100. This supersedes the eight-step pilot, where overlap appeared `7.51%`
faster; the pilot window was too short for that conclusion.

## Interpretation

- Fractional and full optimizer CPU offload execute successfully through
  Megatron Core's `HybridDeviceOptimizer`.
- Greater offload fractions reduce sampled GPU memory, with full offload
  showing the clearest reduction.
- The performance cost is almost entirely inside policy training. NeMo RL
  dispatch overhead is negligible in this microbenchmark.
- Transfer overlap is not beneficial for this small, single-GPU workload. It
  may behave differently when transfer size and available compute overlap are
  larger, so this is not a multi-GPU production-throughput claim.
- All 400 recorded losses are finite. The largest absolute difference from
  the disabled trajectory is `4.221e-3`; this is a numerical sanity check, not
  a convergence claim.

## Configuration

- Model: `Qwen/Qwen3-0.6B`
- Sequence length: `1024`
- GBS/MBS: `8/1`
- Parallelism: TP1, PP1, CP1
- Sequence packing: disabled
- Distributed and precision-aware optimizer: enabled
- Validation, checkpointing, and W&B: disabled
- Measurement window: steps 11-100

## Provenance

- Cluster: `cw-dfw-cs-001`
- SLURM job: `15725145`
- Allocation: one node, one H100 GPU
- Result: `COMPLETED`, exit code `0:0`, elapsed `00:47:34`
- PR commit: `ee766da42275892d4c0df4be51cb3596b3ffe73d`
- Container:
  `/lustre/fs1/portfolios/coreai/projects/coreai_dlalgo_nemorl/users/sna/containers/nemo_rl_nightly_optimizer_offload_60b5c936_20260813_15710559.sqsh`
- Raw results:
  `/lustre/fs1/portfolios/coreai/projects/coreai_dlalgo_nemorl/users/sna/experiments/optimizer-cpu-offload-20260813/results_100step`
- Driver log:
  `/lustre/fs1/portfolios/coreai/projects/coreai_dlalgo_nemorl/users/sna/experiments/optimizer-cpu-offload-20260813/direct1_100step_15725145.log`

seonjinn:
Contributor self-review refresh for `62a0356547febb9c3b1906db9d0ed627b391bec3` after the latest main merge.

An independent source audit found no actionable correctness issues in optimizer offload validation, policy/value lifecycle handling, defaults, or docs. Ruff/format/config checks pass locally. Residual validation: rerun the 0%/50%/100% GPU matrix on the current Bridge/MCore pins and add checkpoint save/resume coverage; the earlier performance run used older dependency pins.

nemo_rl/models/megatron/setup.py
```diff
@@ -1125,50 +1103,26 @@ def _apply_performance_config(model_cfg: Any, config: PolicyConfig) -> None:
...
+        raise ValueError(
+            "optimizer_cpu_offload=True requires 0 < optimizer_offload_fraction <= 1"
+        )
+    if optimizer_cpu_offload and not optimizer_config["use_distributed_optimizer"]:
+        raise ValueError(
+            "optimizer_cpu_offload=True requires use_distributed_optimizer=True"
+        )
+    if not optimizer_cpu_offload and optimizer_config.get(
+        "overlap_cpu_optimizer_d2h_h2d", False
+    ):
```
terrykong: [`nemo_rl/models/megatron/setup.py:1121`](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/models/megatron/setup.py#L1121)

**1 action item.**

PR-introduced. `.get(key, False)` gives a non-`None` default to a `NotRequired` field, which `config-conventions` rules out by name: *"Never provide a non-`None` default — that hides behavior and defeats the purpose of making the field optional."* The skill's "Don't" block shows this exact shape (`cfg.get("disable_ppo_ratio", False)`); its "Do" form for optional booleans is bare truthiness. Dropping `, False` is behaviour-identical, since `None` is falsy.

**Action:** drop the `, False`.

```suggestion
    if not optimizer_cpu_offload and optimizer_config.get(
        "overlap_cpu_optimizer_d2h_h2d"
    ):
```

One alternative — of two: declare the field as a required `bool` in [`MegatronOptimizerConfig`](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/models/policy/__init__.py#L268) like every sibling (`optimizer_cpu_offload`, `optimizer_offload_fraction`) and subscript it. That removes this default *and* the `setdefault` at [L1279](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/models/megatron/setup.py#L1279) in one change. You already add the key to all 28 exemplar/reference YAMLs plus `tools/refit_verifier.py` and the two generation fixtures, so it is required in practice.

nemo_rl/models/megatron/setup.py
```diff
@@ -1322,6 +1276,7 @@ def _create_megatron_config(
         "overlap_param_gather": overlap_param_gather,
         "reuse_grad_buf_for_mxfp8_param_ag": reuse_grad_buf_for_mxfp8_param_ag,
     }
+    optimizer_kwargs.setdefault("overlap_cpu_optimizer_d2h_h2d", False)
```
terrykong: [`nemo_rl/models/megatron/setup.py:1279`](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/models/megatron/setup.py#L1279)

**1 action item.**

PR-introduced. This line changes nothing on either path:

- **Key present** (all 28 YAMLs now set it): [L1274](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/models/megatron/setup.py#L1274) already splats the whole optimizer dict into `optimizer_kwargs`, so `setdefault` is a no-op.
- **Key absent**: Megatron Core already defaults it to the same value — [`overlap_cpu_optimizer_d2h_h2d: bool = False`](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/optimizer/optimizer_config.py#L353) — and the kwargs go straight into `OptimizerConfig(**optimizer_kwargs)` at [L1313](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/models/megatron/setup.py#L1313).

It also runs against the convention this file states ~100 lines above, at [L1175](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/models/megatron/setup.py#L1175): *"Forward checkpoint knobs only when explicitly set in YAML; otherwise Megatron Bridge's own `CheckpointConfig` defaults apply (the exemplar configs own the values) … no call-site default."*

**Action:** delete the line.

```suggestion
```

docs/design-docs/training-backends.md
```diff
@@ -57,50 +57,36 @@ dictionaries follow the hierarchy of nested model-config objects. NeMo
...
+transfers with CPU optimizer updates. At high offload fractions, CPU Adam can
+dominate iteration time; enable this overlap to hide D2H and H2D transfers
+behind the CPU optimizer step.
+
+Megatron Core owns optimizer state placement while this mode is enabled, so
+NeMo RL does not run its generic per-phase optimizer move during colocated
+training and generation. With fractional offload, the non-offloaded optimizer
+state remains GPU resident; choose a fraction that leaves enough GPU memory for
+the colocated generation backend. Optimizer CPU offload is independent of
+activation CPU offload.
```
terrykong: [`docs/design-docs/training-backends.md:84`](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/docs/design-docs/training-backends.md#L84)

**1 action item.**

The paragraph says the non-offloaded state "remains GPU resident", but not which way the comparison runs — and that is the surprising half.

With `optimizer_cpu_offload: false`, NeMo RL moves the **entire** optimizer state to CPU before colocated generation: [`grpo.py:2924`](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/algorithms/grpo.py#L2924) calls `offload_after_refit()` inside `if colocated_inference:` ("unload optimizer to make space for generation"), and [`offload_before_refit`](https://github.com/NVIDIA-NeMo/RL/blob/ee766da42275892d4c0df4be51cb3596b3ffe73d/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2834) does `move_optimizer("cpu")`. Both are gated on `not self.optimizer_cpu_offload`, so with offload **on** the move is skipped for *any* fraction. A fraction below `1.0` can therefore leave a **higher** generation-phase peak than disabling offload altogether. Your own numbers show the resident remainder: 14.13 GiB at f=0.5 vs 10.53 GiB at f=1.0.

Adjacent rather than something this PR broke — the worker gating is unchanged on `main` — but this PR is what makes `f<1` reachable, so your new paragraph is the natural place to say it.

Not asking for a code change: making `move_optimizer` fraction-aware would be unsafe, because `HybridDeviceOptimizer` shares `state` dict objects with its GPU sub-optimizer ([`_sync_sub_optimizers_state_to_hdo`](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py#L302)), so moving them would hand CPU-resident moments to CUDA params. The existing guard is right as written.

**Action:** add one sentence to the paragraph.

```suggestion
Megatron Core owns optimizer state placement while this mode is enabled, so
NeMo RL does not run its generic per-phase optimizer move during colocated
training and generation. With fractional offload, the non-offloaded optimizer
state remains GPU resident; choose a fraction that leaves enough GPU memory for
the colocated generation backend. Note that with `optimizer_cpu_offload: false`
NeMo RL moves the whole optimizer state to CPU before colocated generation, so a
fraction below `1.0` can leave a higher generation-phase peak than disabling
optimizer CPU offload entirely. Optimizer CPU offload is independent of
activation CPU offload.
```

terrykong (commented):
Reviewed by an agent team (6 reviewers plus an adversarial pass), checking every Megatron Core claim against the pinned SHA `d12f6c8c9aff51e166d872fd70151687a8e3f375`.

No action items — the two things we would have raised you already fixed yourself in 788fcf090f5db4462ed1cb4977a9dbd5fac4b6c0 (dropping the `, False` call-site default from the `.get()`, and deleting the redundant `optimizer_kwargs.setdefault(...)`), which is also what `config-conventions` asks for. The added `adam`/`sgd` guard there is right too: Megatron Core only wires `CPUAdam`/`CPUSGD` into the hybrid optimizer, so anything else would have failed obscurely.

Thanks for the validation write-up — the 100-step A/B with an explicit measurement window, the note superseding the 8-step pilot, and the finite-loss check made this much easier to review.

**Context — no action.** Things we verified and deliberately are not raising, in case it saves a round trip:

- The `optimizer_cpu_offload` → `use_distributed_optimizer` guard is correct and worth keeping. The combination it rejects fails with a `KeyError` on the first optimizer step: `Float16OptimizerWithFloat16Params.__init__` overwrites the `HybridDeviceOptimizer`'s own `param_groups` with fresh fp32 clones ([optimizer.py#L1018](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/optimizer/optimizer.py#L1018)), while HDO keyed `param_to_inner_param` on the original tensors ([hybrid_optimizer.py#L189](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py#L189)) and `step()` looks them up unguarded ([#L339](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py#L339)). Your docs sentence about the "parameter mapping" is accurate.
- Removing `assert optimizer_offload_fraction == 1.0` is right: Megatron Core has always accepted arbitrary fractions (HDO's own default is `0.5`) and range-checks the field nowhere, so your `0 < f <= 1` check is the only validation that exists.
- `_force_sync_optimizer_fp32_from_model` stays correct at `f<1` — it iterates the HDO maps rather than assuming full coverage, and the precision-aware on/off branches turn out to be covered by complementary levels (level 3 when it is on, level 1 when it is off).
- The greedy per-parameter partition rounds the realized offload fraction *up*, never down, so it can never leave more on the GPU than requested.
- Config coverage is complete: every YAML declaring `optimizer_cpu_offload` got the new key except two `defaults:`-based override files, where omitting it is correct.

One thing you may want to note in the docs paragraph at some point (not blocking): with `optimizer_cpu_offload: false` NeMo RL moves the *whole* optimizer state to CPU before colocated generation ([`offload_before_refit`](https://github.com/NVIDIA-NeMo/RL/blob/c599b3199a5744618e4b11d6f2b7ac92573ba7d2/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2834), and `grpo.py`'s `offload_after_refit()` under `if colocated_inference:`), and both are gated on `not self.optimizer_cpu_offload` — so a fraction below `1.0` skips that eviction entirely and can leave a higher generation-phase peak than turning offload off. The worker gating is unchanged on `main`; this PR is just what makes `f<1` reachable.

Two pre-existing items worth a follow-up issue rather than this PR: the `_force_sync_optimizer_fp32_from_model` docstring describes the upstream gap backwards (`_copy_model_params_to_main_params` does level 3 only and returns early, [distrib_optimizer.py#L3067](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/optimizer/distrib_optimizer.py#L3067)) — the same error is in Megatron-Bridge's copy; and that helper now duplicates Bridge's public `sync_hybrid_device_optimizer_fp32_master_copies`.

_Generated by Claude Code_

terrykong (approved):

PR #3630, opened 2026-08-13
fix(vllm): support misaligned MXFP8 MoE dimensions

seonjinn:
## Summary

Enable FlashInfer TRTLLM MXFP8 MoE for models whose dimensions are valid for MXFP8 but do not meet the kernel's stricter alignment requirements.

The key idea is simple: **keep checkpoint/refit weights unchanged, and pad only the tensors used by the MoE kernel.**

## Why this is needed

FlashInfer TRTLLM requires:

- hidden size divisible by 512
- TP-local intermediate size divisible by 128

Nemotron Nano 3 does not meet either requirement:

| Dimension | TP1 | TP2 | Kernel requirement |
|---|---:|---:|---:|
| Hidden size | 2688 | 2688 | multiple of 512 |
| Local intermediate size | 1856 | 928 | multiple of 128 |

Without padding, MXFP8 rollout fails while initializing the MoE kernel.

## How it works

For a TP1 Nano expert:

```text
Canonical checkpoint/refit tensors       FlashInfer execution tensors
hidden:       2688  ------------------>   3072
intermediate: 1856  ------------------>   1920
                    zero-padding only
```

The forward path follows the same rule:

```text
activation [..., 2688]
  -> pad to [..., 3072]
  -> run FlashInfer MXFP8 MoE with padded W13/W2
  -> crop output back to [..., 2688]
```

Checkpoint loading and repeated refits still read and write the original 2688/1856 tensors. The padded regions contain zero-valued weights with neutral E8M0 scales, so they do not change the original output.

## Implementation

- derive aligned runtime dimensions from the model and tensor-parallel partition
- preserve canonical W13/W2 values and scales for checkpoint loading and refit
- create separate padded runtime weights and scales for FlashInfer
- pad gated W13 gate/up projections independently
- reuse padded runtime buffers across refits
- reject the unsupported padded modular-kernel path with an explicit error
- install the patch in both legacy and vLLM 0.25.1 Ray executor workers

## Validation

- Lyris GB200, NeMo-RL nightly, vLLM 0.25.1
- 33 targeted vLLM unit tests passed; Pyrefly reported 0 errors
- Nemotron Nano 3, BF16 training + all-MXFP8 MoE rollout
- TP1 and TP2 each completed 2 GRPO steps
- TP1 refit: 2.68 s / 2.25 s; TP2 refit: 2.10 s / 1.62 s
- coverage includes padding contracts, gated/non-gated numerical parity, repeated refit, grouped loading, kernel reuse, and Ray worker serialization

## Scope

This PR only adds FlashInfer TRTLLM MXFP8 MoE dimension padding. Adaptive 8x4/128x4 scale-layout selection, QKVO quantization, and mixed BF16/MXFP8 layerwise reload are outside this PR.


seonjinn:
## /review-pr-team self-review

Reviewed tensor/layout correctness, vLLM worker lifecycle, refit behavior, error boundaries, tests, and CI compatibility.

Resolved findings:

- moved the new tests into the L0 vLLM unit-test collection
- prevented checkpoint-engine refit from running the whole-model post-load hook twice
- preserved valid E8M0 scale byte zero instead of rewriting it
- installed the MXFP8 patch before model loading in vLLM 0.25.1 RayExecutorV2 workers
- isolated per-executor FP8 configs and added a cloudpickle serialization regression test
- added gated and non-gated numerical parity coverage for the padded MoE path
- verified TP1 and TP2 with real Nemotron Nano 3 GRPO runs

Validation at `aabb197611026ef7d93c1985c4d609012ed7bdd6`:

- Lyris GB200, vLLM 0.25.1: 33 targeted tests passed
- Pyrefly: 0 errors
- Nano TP1: 2 GRPO steps completed
- Nano TP2: 2 GRPO steps completed

No unresolved high-confidence findings remain from the self-review.

seonjinn:
Contributor self-review refresh for `0440e4f29e8a7bf3b201b85a8408b7eb75d1bc98` after the latest main merge.

The logical-shape preservation, aligned fast path, misaligned MoE padding, checkpoint-engine behavior, and NCCL lifecycle were rechecked. No new source-level findings were identified. Ruff/format/config checks pass locally; Linux vLLM coverage remains delegated to CI.

nemo_rl/models/generation/vllm/quantization/mxfp8_utils.py
```diff
@@ -30,3 +34,24 @@ def pad_flashinfer_scale_k(input_tensor: Any) -> Any:
     padded = input_tensor.new_zeros(padded_shape)
     padded[..., : input_tensor.shape[-1]] = input_tensor
     return padded
+
+
+def flashinfer_mxfp8_moe_padding_plan(
+    hidden_size: int, intermediate_size: int
+) -> tuple[int, int]:
```
guyueh1, line 43: [`nemo_rl/models/generation/vllm/quantization/mxfp8_utils.py:39`](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/mxfp8_utils.py#L39-L41)

**No action needed** — well, 1 action item: this is a new public, cross-file-used utility (imported by `fp8.py`) with no docstring.

### AI-1
Add a one-line Google-style docstring.

```suggestion
def flashinfer_mxfp8_moe_padding_plan(
    hidden_size: int, intermediate_size: int
) -> tuple[int, int]:
    """Compute FlashInfer TRTLLM MXFP8 MoE execution dimensions.

    Args:
        hidden_size: Unpadded model hidden size.
        intermediate_size: Unpadded TP-local MoE intermediate size.

    Returns:
        A ``(padded_hidden_size, padded_intermediate_size)`` tuple, rounded up to the
        kernel's required alignment (512 / 128 respectively).
    """
```

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -1295,25 +1303,129 @@ def process_weights_after_loading_mxfp8_moe(self, layer) -> None:
...
+            return pad_tensor_dim(tensor, 1, padded_intermediate_size, pad_value)
+
+        intermediate_size = tensor.shape[1] // 2
+        sharded = tensor.reshape(
+            tensor.shape[0], 2, intermediate_size, *tensor.shape[2:]
+        )
+        padded_shape = list(sharded.shape)
+        padded_shape[2] = padded_intermediate_size
+        padded = tensor.new_full(padded_shape, pad_value)
+        padded[:, :, :intermediate_size].copy_(sharded)
```
guyueh1: [`nemo_rl/models/generation/vllm/quantization/fp8.py:1306-1350`](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L1306-L1350)

**1 action item.** `pad_tensor_dim` and `pad_w13_intermediate` are nested inside `process_weights_after_loading_mxfp8_moe`, but neither needs the closure: `pad_tensor_dim` is already a pure function of its parameters, and `pad_w13_intermediate` only uses one free variable (`is_gated`), which is trivially passable as a parameter. No other patch function in this file nests helpers this way — the established convention is module-level helpers in `mxfp8_utils.py` (e.g. the sibling `flashinfer_mxfp8_moe_padding_plan`). Nesting them here makes them untestable in isolation from the whole weight-loading patch. `set_parameter` genuinely needs `layer`, so it can stay nested.

### AI-1
Move `pad_tensor_dim` and `pad_w13_intermediate` to `mxfp8_utils.py`, passing `is_gated: bool` explicitly to the latter; leave `set_parameter` nested.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -1362,34 +1468,100 @@ def process_weights_after_loading_mxfp8_moe(self, layer) -> None:
...
+            kernel_moe_config.intermediate_size_per_partition = padded_intermediate_size
+            kernel_moe_config.intermediate_size_per_partition_unpadded = (
+                unpadded_intermediate_size
+            )
+            kernel_moe_config.intermediate_size = (
+                padded_intermediate_size * kernel_moe_config.moe_parallel_config.tp_size
+            )
         self.moe_quant_config = self.get_fused_moe_quant_config(layer)
         assert self.moe_quant_config is not None
         assert self.experts_cls is not None
```
guyueh1: [`nemo_rl/models/generation/vllm/quantization/fp8.py:1498-1510`](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L1498-L1510)

**1 action item.** `kernel_moe_config = copy(self.moe)` (shallow `copy.copy`) followed by 5 manual field assignments. `FusedMoEConfig` is a dataclass whose [`__post_init__`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/fused_moe/config.py) asserts `intermediate_size % tp_size == 0` and *derives* `intermediate_size_per_partition` / defaults the `_unpadded` fields — `copy.copy()` skips all of that validation/derivation. Today's 5 manually-set values happen to already satisfy the derivation, so this is harmless now, but it's a latent trap if that validation logic is extended later.

### AI-1
Use `dataclasses.replace(self.moe, hidden_dim=padded_hidden_size, hidden_dim_unpadded=unpadded_hidden_size, intermediate_size_per_partition=padded_intermediate_size, intermediate_size_per_partition_unpadded=unpadded_intermediate_size, intermediate_size=padded_intermediate_size * kernel_moe_config.moe_parallel_config.tp_size)` instead, so `__post_init__` re-runs.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -1362,34 +1468,100 @@ def process_weights_after_loading_mxfp8_moe(self, layer) -> None:
...
         from vllm.model_executor.layers.quantization.fp8 import make_fp8_moe_kernel
 
+        kernel_moe_config = self.moe
+        if requires_padding:
+            kernel_moe_config = copy(self.moe)
+            kernel_moe_config.hidden_dim = padded_hidden_size
+            kernel_moe_config.hidden_dim_unpadded = unpadded_hidden_size
+            kernel_moe_config.intermediate_size_per_partition = padded_intermediate_size
+            kernel_moe_config.intermediate_size_per_partition_unpadded = (
+                unpadded_intermediate_size
```
guyueh1: [`nemo_rl/models/generation/vllm/quantization/fp8.py:1500-1503`](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L1500-L1503)

**1 action item.** `kernel_moe_config.hidden_dim_unpadded` / `intermediate_size_per_partition_unpadded` are set here but are dead for this kernel path: [`TrtLlmFp8ExpertsBase.__init__`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/fused_moe/experts/trtllm_fp8_moe.py) (the class constructed for `Fp8MoeBackend.FLASHINFER_TRTLLM`) only reads `hidden_dim` / `intermediate_size_per_partition` — never the `_unpadded` variants.

### AI-1
Drop these two assignments, or add a one-line comment noting they're set only for other `FusedMoEConfig` consumers, not this kernel.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -1295,25 +1303,129 @@ def process_weights_after_loading_mxfp8_moe(self, layer) -> None:
...
+            "Padded FlashInfer TRTLLM MXFP8 MoE requires a monolithic kernel."
+        )
+
+    if requires_padding:
+        from vllm.model_executor.layers.quantization.utils.mxfp8_utils import (
+            MXFP8_BLOCK_SIZE,
+        )
+
+        layer.mxfp8_unpadded_hidden_size = unpadded_hidden_size
+        layer.mxfp8_padded_hidden_size = padded_hidden_size
```
guyueh1: [`nemo_rl/models/generation/vllm/quantization/fp8.py:1389-1390`](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L1389-L1390)

**1 action item.** `layer.mxfp8_unpadded_hidden_size` / `layer.mxfp8_padded_hidden_size` (read back in `apply_monolithic_mxfp8_moe` at [nemo_rl/models/generation/vllm/quantization/fp8.py:1535-1536](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L1535-L1536)) duplicate the same two numbers already stored on `kernel_moe_config.hidden_dim` / `hidden_dim_unpadded` two lines later at [nemo_rl/models/generation/vllm/quantization/fp8.py:1499-1500](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L1499-L1500) — two sources of truth for the same value, one on `layer`, one on the kernel config.

### AI-1
Consider having `apply_monolithic_mxfp8_moe` read the width off `self.moe` / the stored kernel config instead of this second `layer.*` side-channel, if that's accessible from the patched `apply_monolithic` call site.

tests/unit/models/generation/test_vllm_fp8_quantization.py
```diff
@@ -848,6 +849,307 @@ def make_kernel(**kwargs):
...
+        "_shuffle_mxfp8_moe_batched",
+        lambda _layer, w13, w2, s13, s2, _gated, _tile: (w13, w2, s13, s2),
+    )
+
+    with pytest.raises(NotImplementedError, match="requires a monolithic kernel"):
+        fp8.process_weights_after_loading_mxfp8_moe(method, layer)
+
+
+@pytest.mark.parametrize("requires_padding", [False, True])
+def test_apply_monolithic_mxfp8_moe_uses_padded_apply_weights(
```
guyueh1, line 1439: [`tests/unit/models/generation/test_vllm_fp8_quantization.py:1063`](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/tests/unit/models/generation/test_vllm_fp8_quantization.py#L1063)

**1 action item.** `test_apply_monolithic_mxfp8_moe_uses_padded_apply_weights` only exercises the matching-shape success path. The hidden-size-mismatch `ValueError` in `apply_monolithic_mxfp8_moe` at [nemo_rl/models/generation/vllm/quantization/fp8.py:1537-1540](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L1537-L1540) (`x.shape[-1] != unpadded_hidden_size`) has no covering test.

### AI-1
Add a case with a mismatched `x.shape[-1]` asserting the `ValueError` and its message. (NOT verified — no runnable environment on this host.)

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -1295,25 +1303,129 @@ def process_weights_after_loading_mxfp8_moe(self, layer) -> None:
...
+        padded.narrow(dim, 0, current_size).copy_(tensor)
+        return padded
+
+    def set_parameter(name: str, value: torch.Tensor) -> None:
+        value = value.contiguous()
+        parameter = getattr(layer, name, None)
+        if parameter is not None and tuple(parameter.shape) == tuple(value.shape):
+            parameter.data.copy_(value)
+            return
+        setattr(layer, name, torch.nn.Parameter(value, requires_grad=False))
```
guyueh1: [`nemo_rl/models/generation/vllm/quantization/fp8.py:1327-1333`](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L1327-L1333)

**1 action item.** `set_parameter` generalizes a pattern that already exists elsewhere in this file outside the padding path: `process_weights_after_loading_mxfp8_linear` ([nemo_rl/models/generation/vllm/quantization/fp8.py:961-985](https://github.com/NVIDIA-NeMo/RL/blob/5648979a4310789b2c3d173a8a09b351f31d4c31/nemo_rl/models/generation/vllm/quantization/fp8.py#L961-L985)) creates a new `weight_scale` `Parameter` on first load and `.copy_()`s in place on refit — the same "first-load creates, refit copies in place" shape this function encodes, just written out inline and keyed on `hasattr` rather than a shape check.

### AI-1
Extract `set_parameter` as a module-level utility (e.g. in `mxfp8_utils.py`, alongside the other shared helpers) so `process_weights_after_loading_mxfp8_linear` and any other `process_weights_after_loading_*` function with this create-or-copy shape can reuse it instead of re-deriving it. Also rename it — `set_parameter` reads as a generic setter, but its actual contract is "copy in place when the shape already matches (preserving `weight_loader` and other `Parameter` subclass metadata), otherwise replace with a new `Parameter`" — a name like `assign_or_replace_parameter` would make that contract legible at call sites.

guyueh1 (commented):
Agent-team review of PR #3630. 5 specialist agents (rl-expert, bug-finder, test-agent, design-reviewer, comment-reviewer) reviewed independently; devil's-advocate stress-tested every finding before this review was staged.

Overall: this is a well-verified, well-tested change. rl-expert independently confirmed the core padding/kernel-integration logic against the actual pinned vLLM v0.25.1 source (not a stale local checkout) and found no correctness bugs. The 6 comments below are low-severity design/doc/test polish suggestions, none blocking.

Also worth noting: the accompanying test-file rename (`tests/test_mxfp8_flashinfer_padding.py` → `tests/unit/models/generation/test_vllm_mxfp8_utils.py`) fixes a pre-existing CI gap — the old path was never wired into any L0 shard, so this PR adds coverage that didn't previously run in CI.

_Generated by Claude Code_

guyueh1:
@seonjinn i am ok with the current change, but because we added assign_or_replace_parameter and refactors not just mxfp8 but also some other refit patches for block-wise fp8 etc, can you run those recipe tests to make sure they are not broken?

guyueh1 (dismissed):

guyueh1:
/ok to test https://github.com/NVIDIA-NeMo/RL/commit/6a6839873e79b206d6b225e9c2c142080c41d97d

seonjinn:
I expanded the regression check around the shared helper. On the current head, regular block-wise FP8 code is unchanged from the PR base, and `assign_or_replace_parameter` is only used by the MXFP8 linear/MoE paths. Local validation completed with 48 pytest cases passing (MXFP8 recipe/config, helper, layout, repeated/padded refit), all 11 non-performance FP8/MXFP8 recipe drivers passing dry-run resolution, and all 17 relevant functional/recipe scripts passing shell validation. I also identified the existing repeated in-place block-wise FP8, checkpoint/sharded refit, layerwise reload, and checkpoint-engine tests for the contributor CI run. The remaining gap is real Linux/CUDA recipe execution: the Nano nightly MXFP8 recipes use Megatron generation and do not exercise this vLLM helper, so the current-head CI plus a vLLM MXFP8 GB200 smoke are still needed for end-to-end coverage.

seonjinn:
Fixed a runtime-storage ownership bug in 30ca6e4f. When batched shuffle returns shared scratch, installing a new runtime Parameter must clone that storage; contiguous() alone does not copy it. Compatible subsequent refits still use copy_ and preserve CUDA Graph addresses. Added two-layer tests for initial allocation, resizing, forced replacement, and repeated updates on CPU/CUDA. Integration GPU red/green job 7094400 is submitted on GB200; results and Nano 20-step KL validation are pending. This is a concrete aliasing fix, but recovery of the observed high KL has not yet been demonstrated.

guyueh1 (dismissed):

guyueh1 (dismissed):

guyueh1:
@seonjinn can you fix the unit test

guyueh1 (dismissed):

seonjinn:
/ok to test [6ecfdf7](https://github.com/NVIDIA-NeMo/RL/pull/3630/commits/6ecfdf76090d39cfda0c21a459950acdd4d2578e)

guyueh1 (approved):

seonjinn:
@terrykong Could you review this PR ?

seonjinn:
/ok to test [f3abda6](https://github.com/NVIDIA-NeMo/RL/pull/3630/commits/f3abda6b8ce0fcd16ae3bb25aef70b23b13cdacb)

seonjinn:
/ok to test [0bd1dc8](https://github.com/NVIDIA-NeMo/RL/pull/3630/commits/0bd1dc8d3296d0a28a3fcb8c0a7c06316efeeebb)

terrykong (approved):

seonjinn:
/ok to test [3240185](https://github.com/NVIDIA-NeMo/RL/pull/3630/commits/3240185f86ede628735ba29f8c04710c461812ff)

seonjinn:
/ok to test [7f1e1de](https://github.com/NVIDIA-NeMo/RL/pull/3630/commits/7f1e1de58f9fa4e17fdfb4a49e342fadfe20ff6e)

PR #3631, opened 2026-08-13
feat(vlm): carry a media-token validity mask through sequence packing

DanialTaheri:
## What

Lets a row whose text legitimately contains the media placeholder train, instead
of being dropped at data-prep time.

Stacked on #3494 — based on `smohsenitahe/super-omni-mtp-cp`, so this diff is
only the two commits. GitHub will retarget the base to `main` once #3494 merges.

## Why this is separate

#3494 removes placeholder sanitization, and the VLM blend trains fine without
this: every row there either carries no literal placeholder, or carries exactly
as many as it has images. A 50-step Super Omni run with sanitization off and the
mask entirely unwired completed 50/50 with zero media-alignment failures.

The mask matters only for rows with `placeholders > 0, images == 0` — text that
spells the placeholder with no image attached. There are **zero** such rows in
the filtered VLM blend, so #3494 does not need this. Text blends do: in a 100k
sample of a Nemotron post-training blend, 2,539 rows (2.5%) contain a literal
`<image>` with no image — competitive-programming statements where it replaced
inline math.

## Why the model cannot answer this itself

`_merge_projected_media` enforces one projected feature per valid placeholder.
The mask it derives comes from the padding or attention mask, and both answer
*"is this a real token"*, which is a different question from *"is this a media
anchor"*. They coincide only while every media token in a valid position anchors
an image.

The caller knows how many media items each row carries, so it can answer. This
carries that answer to the model.

## The packing constraint

The mask is built in sample space, where a row is a sample. Sequence packing
then concatenates samples into one THD sequence and CP-shards it, after which
no per-row question can be asked — so the mask travels through the *same*
packing transform as `input_ids` rather than being derived downstream, exactly
as `mtp_loss_mask` already does.

A mask that is merely misaligned does not raise; it attaches image features to
the wrong positions. The packing tests are the load-bearing ones.

## Dependency

Needs NVIDIA-NeMo/Megatron-Bridge#5573, which adds the `media_token_validity_mask`
argument to `NemotronOmniModel.forward`. Until that merges and the pin is bumped,
`_model_accepts_media_token_validity_mask` finds no such parameter and no mask is
ever attached — inert, not broken.

## Tests

- 6 unit tests for mask construction
- 2 mcore-marked tests pinning the post-packing layout
- Cluster: 256-row mixed batch (128 real-image rows + 128 text-only rows carrying
  real `<image>` statements) reached Step 1/1, rollouts 100%, zero alignment failures

### DIFF
--- a/docs/design-docs/media-token-validity-mask.md
+++ b/docs/design-docs/media-token-validity-mask.md
@@ -1,0 +1,214 @@
+# Training on text that contains `<image>`
+
+## Summary
+
+A media placeholder is an ordinary vocabulary entry. Text can legitimately
+contain it, and until now that text could not be trained on: the model read
+every occurrence as an anchor for a projected image feature and failed when no
+feature existed. The workaround was a sanitizer that rewrote the offending
+tokens before they reached the model, which changed the data to fit the model.
+
+This design removes the sanitizer and replaces it with a **media-token validity
+mask**: the caller, which knows how many media items each row carries, marks
+which placeholder positions are real anchors. Positions it does not mark keep
+their ordinary token embedding.
+
+## The problem
+
+`NemotronOmniModel._merge_projected_media` enforces a strict 1:1 contract — one
+projected feature per valid placeholder:
+
+```python
+media_mask = input_ids == media_token_id
+if attention_mask is not None:
+    media_mask = media_mask & attention_mask.bool()
+
+expected_features = int(media_mask.sum().item())
+actual_features = media_embeddings.shape[0]
+if expected_features != actual_features:
+    raise ValueError("Expanded-sequence media alignment failed: ...")
+```
+
+The contract is right. The question is where `media_mask` comes from. Before
+this change the model derived it from whichever mask it had:
+
+```python
+media_token_validity_mask = None
+if padding_mask is not None:
+    media_token_validity_mask = ~padding_mask
+elif attention_mask is not None and attention_mask.dim() == input_ids.dim():
+    media_token_validity_mask = attention_mask
+```
+
+Both of those answer **"is this a real token?"**. The merge needs **"is this a
+media anchor?"**. Those coincide only while every media token in a non-padding
+position anchors an image. A text row that spells the placeholder breaks the
+equivalence: the position is a real token, so the derived mask marks it valid,
+so the merge demands a feature that was never meant to exist.
+
+This is not hypothetical. In a 100k-row sample of a Nemotron post-training
+blend, 2,539 rows (2.5%) contain a literal `<image>` with **zero** attached
+images — competitive-programming statements where `<image>` replaced inline
+math, e.g. `"for any character <image> there is exactly one character"`. They
+carry between 1 and 12 placeholders each.
+
+There is a second, subtler source. Chat templates that treat a literal
+`<image>` in the prose as the placement anchor suppress their own generated
+image block. A row with two literal tokens and one attached image therefore
+renders two placeholders for one feature.
+
+## Why not sanitize
+
+The previous approach rewrote the data at rollout time: drop the literal token
+when images were attached, or replace it with the word `image` when not.
+
+It worked, but it had three problems:
+
+1. **It edits the user's prose.** `"for any character <image> there is"` became
+   `"for any character image there is"`. The model trains on text the author
+   did not write.
+2. **It hides malformed data.** A row with two placeholders and one image is a
+   defect; sanitizing silently patched it. Filtering the blend surfaced 392 such
+   rows, which turned out to be unanswerable questions.
+3. **It cannot express the legitimate case.** There is no rewrite that both
+   preserves the prose and tells the model "this one is not an anchor".
+
+## Design
+
+Give the caller a way to state the answer directly, since the caller is the only
+party that knows it. The model keeps its strict contract; it just stops guessing
+the input to that contract.
+
+### Model change
+
+`NemotronOmniModel.forward` takes a keyword-only argument that takes precedence
+over the derived masks:
+
+```python
+if media_token_validity_mask is None:
+    if padding_mask is not None:
+        media_token_validity_mask = ~padding_mask
+    elif attention_mask is not None and attention_mask.dim() == input_ids.dim():
+        media_token_validity_mask = attention_mask
+```
+
+Behavior is unchanged when the argument is omitted. This is the only change to
+the model, and it is additive.
+
+### Building the mask
+
+`build_media_token_validity_mask(input_ids, media_token_id, media_counts_by_row)`
+marks media-token positions in rows whose media count is zero:
+
+- Every row carries media → returns `None`; the model derives its own mask.
+- Text rows exist but none spell the token → returns `None` for the same reason.
+- Otherwise → a `[B, S]` bool mask, `False` at media-token positions of
+  media-less rows.
+
+Rows that **do** carry media keep every position valid. A genuine
+placeholder/feature disagreement there is still reported rather than masked
+away, so the mask cannot be used to silence real misalignment.
+
+### Carrying it through sequence packing
+
+This is the part that determines whether the mask means anything.
+
+The mask is built in **sample space**, where row `i` of `input_ids` pairs with
+`media_counts_by_row[i]`. Sequence packing then concatenates many samples into
+one THD sequence and context-parallel-shards it. After that, one "row" holds
+many samples and each rank holds a slice — the per-row question the mask answers
+can no longer be asked.
+
+So the mask must travel through the *same* transform as `input_ids` rather than
+be derived downstream. It is packed alongside them, exactly as `mtp_loss_mask`
+already is:
+
+```python
+if "media_token_validity_mask" in data_dict:
+    packed_media_mask, local_media_mask, _, _, _ = _pack_sequences_for_megatron(
+        data_dict["media_token_validity_mask"].to(data_dict["input_ids"].dtype),
+        seq_lengths,
+        pad_individual_seqs_to_multiple_of,
+        pad_packed_seq_to_multiple_of,
+        pad_full_seq_to,
+        cp_rank=get_context_parallel_rank(),
+        cp_size=get_context_parallel_world_size(),
+    )
+    media_token_validity_mask = (
+        packed_media_mask
+        if model_slices_context_parallel_inputs
+        else local_media_mask
+    ).bool()
+```
+
+Two details matter:
+
+- **Packed in token dtype.** Packing pads with `value=0`, which is a valid token
+  id but not a valid bool. It is converted back to bool after packing. Padding
+  positions become `False`, which is harmless because padding holds no media
+  token.
+- **Packed vs CP-local.** A model that slices context parallelism itself
+  receives the full THD row so it can insert media before selecting its
+  CP-owned embeddings. Its mask must stay unsharded to line up. Every other
+  model consumes the CP-local shard. This mirrors the `mtp_loss_mask` choice
+  for the same reason. `NemotronOmniModel` sets
+  `model_slices_context_parallel_inputs = True`, so it gets the full row.
+
+### Capability detection
+
+The mask is only sent to models whose `forward` actually declares it:
+
+```python
+def _model_accepts_media_token_validity_mask(model) -> bool:
+    ...
+    if "media_token_validity_mask" in inspect.signature(chunk.forward).parameters:
+        return True
+```
+
+The check is on the signature rather than a class flag because a model that does
+not know about the mask would absorb it into `**kwargs` and ignore it — which
+looks identical to the mask having been applied. Failing to send it is visible;
+sending it into a void is not.
+
+### Where it is attached
+
+At all three forward sites: training and **both** logprob paths. Logprobs run
+the same forward as training, so a batch that needs the mask needs it there too
+— otherwise logprobs would be computed against a different media alignment than
+the one trained on.
+
+### Self-packing models
+
+Models that pack internally (`delegate_pack_to_model`) raise
+`NotImplementedError` rather than silently dropping the mask. A mask built
+against caller-side rows would reach the merge in a layout that no longer
+matches its tokens, and a misaligned media mask attaches features to the wrong
+positions without erroring.
+
+## Consequences
+
+**Removed.** `sanitize_nemo_gym_example_image_placeholders`,
+`_normalize_image_placeholders`, `_count_image_payloads`, the
+`sanitize_image_placeholders` config flag, and their tests — 212 lines.
+
+**Data prep is now required, not optional.** Without the sanitizer there is no
+safety net: a blend whose placeholder and image counts disagree fails loudly at
+the media merge. That is the intended behavior — the failure is a real defect —
+but it means malformed rows must be filtered before training rather than being
+absorbed at rollout time.
+
+**Text rows containing `<image>` can be trained on** instead of dropped, which
+is what the mask exists for.
+
+## Validation
+
+| Test | Result |
+|---|---|
+| 50-step Super Omni run, sanitization removed | 50/50 steps, zero media-alignment / device / IMA errors; reward 0.458 (steps 1–8) → 0.520 (steps 26–49), peak 0.5879 vs. a 0.4932 baseline best |
+| Mixed batch: 128 rows with real images + 128 text-only rows carrying real `<image>` statements | Reached Step 1/1, rollouts 100%, zero alignment failures |
+| Unit: mask construction | 6 tests |
+| Unit: mask survives packing, lands on the same tokens | 2 tests (require mcore) |
+
+The packing tests are the load-bearing ones: they assert the mask still marks
+the intended tokens after `_pack_sequences_for_megatron`, because a mask that is
+merely misaligned does not raise — it attaches features to the wrong positions.
--- a/docs/design-docs/media-token-validity-mask.md
+++ b/docs/design-docs/media-token-validity-mask.md
@@ -2,16 +2,16 @@
 
 ## Summary
 
-A media placeholder is an ordinary vocabulary entry. Text can legitimately
-contain it, and until now that text could not be trained on: the model read
-every occurrence as an anchor for a projected image feature and failed when no
-feature existed. The workaround was a sanitizer that rewrote the offending
-tokens before they reached the model, which changed the data to fit the model.
+A media placeholder is an ordinary vocabulary entry, so text can legitimately
+contain it — a competitive-programming statement that spells `<image>` in its
+prose, for instance. The model reads every occurrence as an anchor for a
+projected image feature and fails when no feature exists, so those rows cannot
+be trained on.
 
-This design removes the sanitizer and replaces it with a **media-token validity
-mask**: the caller, which knows how many media items each row carries, marks
-which placeholder positions are real anchors. Positions it does not mark keep
-their ordinary token embedding.
+This design adds a **media-token validity mask**: the caller, which knows how
+many media items each row carries, marks which placeholder positions are real
+anchors. Positions it does not mark are skipped by the media merge and keep
+whatever language embedding the forward gave them.
 
 ## The problem
 
--- a/nemo_rl/data/multimodal_utils.py
+++ b/nemo_rl/data/multimodal_utils.py
@@ -18,6 +18,7 @@ import logging
 import re
 import uuid
 from collections import defaultdict
+from collections.abc import Sequence
 from copy import deepcopy
 from io import BytesIO
 from typing import Any, Optional, Union
@@ -1151,3 +1152,60 @@ def load_media_from_message(
                 loaded_media["video"].append(vid)
 
     return loaded_media
+
+
+def build_media_token_validity_mask(
+    input_ids: torch.Tensor,
+    media_token_id: int,
+    media_counts_by_row: Sequence[int],
+    base_mask: Optional[torch.Tensor] = None,
+) -> Optional[torch.Tensor]:
+    """Mark media-token positions in rows that carry no media of that modality.
+
+    A media token is an ordinary vocabulary entry with its own embedding row.
+    It only means "a projected feature belongs here" when media is attached;
+    in a text-only row the same id is whatever the author wrote, and counting
+    it as a placeholder makes the model demand a feature that does not exist.
+
+    Rows that do carry media keep every position valid, so a real
+    placeholder/feature disagreement there is still reported rather than
+    silently masked away.
+
+    Args:
+        input_ids: ``[B, S]`` token ids, one row per sample.
+        media_token_id: Vocabulary id the model treats as a media placeholder.
+        media_counts_by_row: Attached media items per row, e.g. from
+            :meth:`PackedTensor.logical_segment_counts_by_row`.
+        base_mask: Optional ``[B, S]`` validity mask to refine, so masks for
+            several modalities can be combined.
+
+    Returns:
+        A ``[B, S]`` bool mask, or ``None`` when nothing needs masking and the
+        caller should leave the model's own derivation untouched.
+    """
+    if input_ids.ndim != 2:
+        raise ValueError(f"input_ids must be [B, S], got {tuple(input_ids.shape)}")
+    if len(media_counts_by_row) != input_ids.shape[0]:
+        raise ValueError(
+            "media_counts_by_row must have one entry per row: got "
+            f"{len(media_counts_by_row)} for {input_ids.shape[0]} rows"
+        )
+
+    empty_rows = [row for row, count in enumerate(media_counts_by_row) if count == 0]
+    if not empty_rows:
+        return base_mask
+
+    is_media_token = input_ids == media_token_id
+    if not bool(is_media_token[empty_rows].any()):
+        # Text-only rows exist but none of them spell the token, so the
+        # model's own derivation is already correct.
+        return base_mask
+
+    mask = (
+        torch.ones_like(input_ids, dtype=torch.bool)
+        if base_mask is None
+        else base_mask.clone()
+    )
+    for row in empty_rows:
+        mask[row] &= ~is_media_token[row]
+    return mask
--- a/nemo_rl/data/multimodal_utils.py
+++ b/nemo_rl/data/multimodal_utils.py
@@ -1209,3 +1209,73 @@ def build_media_token_validity_mask(
     for row in empty_rows:
         mask[row] &= ~is_media_token[row]
     return mask
+
+
+def media_placeholder_token_id_from_chunks(chunks: Sequence[Any]) -> Optional[int]:
+    """The vocabulary id these model chunks treat as a media placeholder, if any."""
+    for chunk in chunks:
+        token_id = getattr(chunk, "image_token_index", None)
+        if token_id is not None:
+            return int(token_id)
+    return None
+
+
+def chunks_accept_media_token_validity_mask(chunks: Sequence[Any]) -> bool:
+    """Whether a model chunk's forward takes an explicit media-token validity mask.
+
+    Checked against the signature rather than a class flag so a model that does
+    not know about the mask never receives it: such a forward would absorb it
+    into ``**kwargs`` and silently ignore it, which looks identical to the mask
+    having been applied. Failing to send it is visible; sending it into a void
+    is not.
+    """
+    for chunk in chunks:
+        try:
+            parameters = inspect.signature(chunk.forward).parameters
+        except (TypeError, ValueError):
+            continue
+        if "media_token_validity_mask" in parameters:
+            return True
+    return False
+
+
+def image_counts_by_row(batch: Any, num_rows: int) -> Optional[list[int]]:
+    """How many images each row of the batch actually carries.
+
+    Returns None when the batch describes its images in a way this cannot read,
+    so the caller leaves the model's own derivation alone rather than guessing a
+    count and masking against it.
+    """
+    pixel_values = batch.get("pixel_values", None)
+    if pixel_values is None:
+        # A text-only batch genuinely has no images anywhere, which is exactly
+        # the case the mask exists for -- not a missing-data case.
+        return [0] * num_rows
+    if not isinstance(pixel_values, PackedTensor):
+        return None
+    counts = pixel_values.logical_segment_counts_by_row()
+    return counts if len(counts) == num_rows else None
+
+
+def attach_media_token_validity_mask(batch: Any, media_token_id: Optional[int]) -> None:
+    """Mark media tokens that anchor nothing, so the model keeps their embedding.
+
+    Builds the mask while rows still are samples. Sequence packing later
+    concatenates those rows into one THD sequence, after which no per-row
+    question can be asked, so the packing step carries this through the same
+    transform as ``input_ids`` rather than deriving it downstream.
+
+    The batch is duck-typed rather than annotated as ``BatchedDataDict``:
+    that module imports this one, so naming it here would be circular.
+    """
+    if media_token_id is None:
+        return
+    input_ids = batch.get("input_ids", None)
+    if not isinstance(input_ids, torch.Tensor) or input_ids.ndim != 2:
+        return
+    counts = image_counts_by_row(batch, input_ids.shape[0])
+    if counts is None:
+        return
+    mask = build_media_token_validity_mask(input_ids, media_token_id, counts)
+    if mask is not None:
+        batch["media_token_validity_mask"] = mask
--- a/nemo_rl/models/megatron/data.py
+++ b/nemo_rl/models/megatron/data.py
@@ -50,6 +50,7 @@ class ProcessedInputs:
     mtp_loss_mask: Optional[torch.Tensor] = None
     routed_experts: Optional[torch.Tensor] = None
     routed_experts_cp_sharded: Optional[torch.Tensor] = None
+    media_token_validity_mask: Optional[torch.Tensor] = None
 
 
 @dataclass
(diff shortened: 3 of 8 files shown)


nemo_rl/models/megatron/data.py
```diff
@@ -501,6 +519,43 @@ def process_microbatch(
                         else local_mtp_loss_mask
                     )
 
+                # Pack the media-token validity mask the same way as input_ids.
+                # The mask answers a per-token question, so it only means
+                # anything while it sits in the same layout as the tokens the
+                # model will compare it against. Packing is what destroys the
+                # per-sample rows it was built from, so it has to travel through
+                # the identical transform rather than be rebuilt afterwards.
+                if "media_token_validity_mask" in data_dict:
```
yfw: [`nemo_rl/models/megatron/data.py:528`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L528)

**4 action items, all already written and passing on our side.**

**TL;DR — neither new test file exercises a line this PR adds to `process_microbatch`; the `mtp_loss_mask` path it mirrors has six such tests and the mask has zero.**

Both new tests call [`_pack_sequences_for_megatron`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L960) directly, which is pre-existing and untouched by this PR. `grep -rn "media_token_validity_mask" tests/` outside the two new files returns nothing, so the four blocks that *are* this PR's contribution to `data.py` have no coverage.

`mtp_loss_mask` — the path this PR mirrors line-for-line — has one test per branch:

| existing `mtp_loss_mask` test | mirrors this uncovered block |
|---|---|
| [`test_megatron_data.py:572`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L572) | [`data.py:528-549`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L528-L549) packing |
| [`:435`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L435) | [`:553-557`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L553-L557) CP-slicing ternary |
| [`:312`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L312) | [`:610-613`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L610-L613) unpacked branch |
| [`:848`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L848) | [`:315-325`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L315-L325) `NotImplementedError` |

Those siblings also already defend against the exact false-pass a mirrored feature invites — [`:453-463`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L453-L463) deliberately uses distinct tensors so a wrong one cannot "pass unnoticed, since both sides would then be identical".

### AI-1 — four `process_microbatch` tests
Mirror the four above. Happy to hand over the source we wrote and ran.

### AI-2 — two capability-probe tests
[`_model_media_placeholder_token_id`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/policy/workers/megatron_policy_worker.py#L187) and [`_model_accepts_media_token_validity_mask`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/policy/workers/megatron_policy_worker.py#L200) have none, while all three sibling probes in the same file do — [`test_megatron_worker.py:50-57`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/policy/test_megatron_worker.py#L50-L57), [`:60-69`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/policy/test_megatron_worker.py#L60-L69), [`:140-149`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/policy/test_megatron_worker.py#L140-L149). ~12 lines each with a duck-typed stub. The highest-value case is a `forward(self, x, **kwargs)` stub asserting `False`, since that is exactly the silent-swallow your docstring justifies the signature check by — and [`NemotronOmniModel.forward`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L647) really does have `**kwargs` followed by [`del kwargs`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L656). *Our sketch here is unverified against the real module: `transformer_engine` is unavailable in our environment, as it is for the pre-existing tests in that file.*

### AI-3 — assert the property the padding test is named for
[`test_media_mask_packing.py:74`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_media_mask_packing.py#L74) is `assert IMG != 0`, comparing two module-level constants, so it can never fail. Meanwhile no assertion touches `packed_mask` at the padded positions the test is named for. `assert not bool(packed_mask[0, 3:].any())` closes that gap — we ran it against the real packing and it passes.

### AI-4 — parametrise `cp_size` while you are there
Both tests run at `cp_size=1`, where [`_get_tokens_on_this_cp_rank`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/distributed/model_utils.py#L1199) is never reached — so the packed-vs-CP-local ternary, which exists *only* for CP>1, is untested. [`test_shard_routed_experts_for_cp_matches_input_ids_zigzag`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L1597) is the in-file pattern to copy. We mutation-checked ours: packing the mask at `cp_rank=0` while `input_ids` uses `cp_rank` makes it fail at `[rank1-cp2]`, so it discriminates rather than passing on a no-op.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -180,6 +184,86 @@ def _model_slices_context_parallel_inputs(model: Any) -> bool:
...
+            continue
+        if "media_token_validity_mask" in parameters:
+            return True
+    return False
+
+
+def _attach_media_token_validity_mask(
+    batch: BatchedDataDict[Any], media_token_id: Optional[int]
+) -> None:
+    """Mark media tokens that anchor nothing, so the model keeps their embedding.
```
yfw: [`nemo_rl/models/policy/workers/megatron_policy_worker.py:227`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/policy/workers/megatron_policy_worker.py#L227)

**1 action item.**

**TL;DR — a spared placeholder does not keep its own embedding; it gets token id 0's, because the model zeroes every image-token position before the embedding lookup and before the mask is consulted.**

[`modeling_nemotron_omni.py:687`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L687), the file's only `masked_fill`, is unconditional and runs *before* the mask is read at [`:706`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L706):

```python
input_ids_text = input_ids.masked_fill(input_ids == self.image_token_index, 0)
combined_embeddings = self.language_model.embedding(input_ids=input_ids_text, ...)
```

[`_merge_projected_media`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L325) then writes only at mask-True positions, and nothing restores. Your own upstream test records this — [`test_nemotron_omni_model.py:584`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/tests/unit_tests/models/nemotron_omni/test_nemotron_omni_model.py#L584) asserts `[[0.0] * 3]` at the spared position, with the comment *"That embedding is of token id 0"*.

This is adjacent rather than something this PR broke: the [upstream comment at `:693-699`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L693-L699) makes the same claim, so the sentence propagated from there. The behaviour is also defensible — deterministic, and better than crashing or rewriting the author's prose — and `input_ids`, hence the labels, are untouched, so the substitution is input-side only. **No code change requested.**

**Action:** correct this summary line.

```suggestion
    """Mark media tokens that anchor nothing, so the merge skips them.
```

The same mechanism softens [`Why not sanitize` bullet 1](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L67-L70) ("the model trains on text the author did not write"): the sanitizer substituted the embedding of the word `image`, and this substitutes token id 0's. Both substitute — this one at least preserves the prose and the labels.

One argument for fixing it upstream rather than here: `sound_token_index` is *not* `masked_fill`'d, so the two modalities are asymmetric, which is what makes it look like an oversight. But the obvious fix (`&= media_token_validity_mask`) is **not** trivially safe — the `masked_fill` doubles as an out-of-range guard, and [`nemotron_omni_provider.py:184-187`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/nemotron_omni_provider.py#L184-L187) asserts only `image_token_index > 0`, never an upper bound. Possibly worth an issue at https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/new.

docs/design-docs/media-token-validity-mask.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Removed.** `sanitize_nemo_gym_example_image_placeholders`,
+`_normalize_image_placeholders`, `_count_image_payloads`, the
+`sanitize_image_placeholders` config flag, and their tests — 212 lines.
+
+**Data prep is now required, not optional.** Without the sanitizer there is no
+safety net: a blend whose placeholder and image counts disagree fails loudly at
+the media merge. That is the intended behavior — the failure is a real defect —
+but it means malformed rows must be filtered before training rather than being
+absorbed at rollout time.
```
yfw: [`docs/design-docs/media-token-validity-mask.md:190-198`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L190-L198)

**3 action items**, all in this one file, grouped so they can be fixed in one pass.

**TL;DR — the doc describes a sanitizer removal that no merged commit performed, and credits the mask with a run the PR body says had the mask unwired.**

### AI-1 — the "Removed" paragraph names symbols that exist in no commit
`git log --oneline origin/main -S sanitize_image_placeholders` → empty. `git merge-base --is-ancestor a4ac69648 origin/main` → no; against this branch → also no. A repo-wide grep finds all four symbols only inside this doc. #3494 was squash-merged, and the sanitizer was added *and* removed inside that branch, so it netted to zero and never reached `main`. The knock-on at [`:194`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L194) then asserts a regression from a state that never shipped.

```suggestion
**No sanitizer ships.** An earlier iteration of #3494 prototyped one
(`sanitize_nemo_gym_example_image_placeholders` and friends, plus a
`sanitize_image_placeholders` config flag). It was dropped inside that branch
before merge, so it never reached `main` and this change removes nothing.

**Data prep is required.** There is no safety net at rollout time: a blend whose
placeholder and image counts disagree fails loudly at the media merge. That is
the intended behavior — the failure is a real defect — but it means malformed
rows must be filtered before training.
```

Two other spots carry the same tense: [`:11`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L11) ("This design **removes the sanitizer**") and the [`Why not sanitize`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L66) section, whose reasoning is worth keeping — just as "the approach we rejected" rather than as shipped behaviour.

### AI-2 — Validation row 1 measures a different change
[`:207`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L207) credits `reward 0.458 → 0.520, peak 0.5879` to the mask, but the PR body says of that same run: *"with sanitization off and **the mask entirely unwired**"*. It is also absent from the body's `## Tests` list, so the doc makes a claim the PR description itself does not. Suggest dropping the row or relabelling it as #3494 evidence — the 256-row mixed-batch run and the unit tests are the right support for what remains. Worth noting row 2 stops at one optimizer step, and this doc argues at [`:212-214`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L212-L214) that a misaligned mask does not raise, so one non-crashing step cannot distinguish correct from misaligned.

### AI-3 — the "second, subtler source" is not handled, and is disowned later
[`:55-58`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L55-L58) motivates the design with a row carrying two literal tokens and one attached image. That row has `count > 0`, so it never enters [`empty_rows`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/data/multimodal_utils.py#L1194) and the merge still raises — the design does not solve it. Two sections later, [`Why not sanitize` bullet 2](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L70-L72) calls the same row shape a defect to be filtered out. One sentence saying which it is would resolve the tension.

tests/unit/data/test_media_token_validity_mask.py
```diff
@@ -0,0 +1,71 @@
...
+from nemo_rl.data.multimodal_utils import build_media_token_validity_mask
+
+IMG = 7  # stand-in media token id
+TXT = 1
+
+
+def test_returns_base_mask_when_every_row_has_media():
+    # Nothing to correct: each row's media token really does mark a feature.
+    input_ids = torch.tensor([[TXT, IMG, TXT], [IMG, TXT, TXT]])
+    assert build_media_token_validity_mask(input_ids, IMG, [1, 1]) is None
```
yfw: [`tests/unit/data/test_media_token_validity_mask.py:27`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/data/test_media_token_validity_mask.py#L27)

**1 action item.** Low severity — `base_mask` has no production caller yet, so this is a latent-seam gap rather than a live bug.

Both `test_returns_base_mask_when_*` tests are named for returning the base mask, but pass `base_mask=None` and assert `is None` — so [`return base_mask`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/data/multimodal_utils.py#L1196) and `return None` are indistinguishable across the entire suite.

Measured rather than asserted: replacing `build_media_token_validity_mask` with `def f(...): return None` leaves **4 failed, 2 passed**, and the 2 survivors are exactly these two tests. The other four are load-bearing.

**Action:** exercise the passthrough. Verified before posting — `ruff format --check` and `ruff check` clean, 6 passed.

```suggestion
    assert build_media_token_validity_mask(input_ids, IMG, [1, 1]) is None
    # A caller-supplied base mask must come back untouched, not be dropped.
    base = torch.tensor([[True, True, False], [True, True, True]])
    result = build_media_token_validity_mask(input_ids, IMG, [1, 1], base_mask=base)
    assert result is base
```

The same two lines apply to `test_returns_base_mask_when_text_rows_do_not_spell_the_token` at [`:33`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/data/test_media_token_validity_mask.py#L33) with counts `[1, 0]`. Please keep the call unwrapped — at 84 chars it is under ruff's default 88, and a hand-wrapped version gets collapsed back and fails `--check`.

nemo_rl/data/multimodal_utils.py
```diff
@@ -1151,3 +1152,60 @@ def load_media_from_message(
...
+        input_ids: ``[B, S]`` token ids, one row per sample.
+        media_token_id: Vocabulary id the model treats as a media placeholder.
+        media_counts_by_row: Attached media items per row, e.g. from
+            :meth:`PackedTensor.logical_segment_counts_by_row`.
+        base_mask: Optional ``[B, S]`` validity mask to refine, so masks for
+            several modalities can be combined.
+
+    Returns:
+        A ``[B, S]`` bool mask, or ``None`` when nothing needs masking and the
+        caller should leave the model's own derivation untouched.
```
yfw: [`nemo_rl/data/multimodal_utils.py:1182-1184`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/data/multimodal_utils.py#L1182-L1184)

**1 action item.**

`Returns:` says `None` when nothing needs masking, but both no-op paths [`return base_mask`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/data/multimodal_utils.py#L1196), which is `None` only when the caller passed none. A caller that supplies a base mask and branches on `is None` would silently drop its own mask.

Folded in two clauses our reviewers had to reverse-engineer: `image_token_index` covers video too (Omni carries video as N frames sharing that id, [`:655`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L655)), and the model consumes this mask *instead of* its own derived one ([`:706`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L706)) rather than intersecting with it.

**Action:** verified before posting — `ruff format --check` clean, `py_compile` clean.

```suggestion
    Returns:
        A ``[B, S]`` bool mask, or ``base_mask`` unchanged (``None`` when no
        base mask was given) when nothing needs masking and the caller should
        leave the model's own derivation untouched.
```

nemo_rl/models/megatron/data.py
```diff
@@ -501,6 +519,43 @@ def process_microbatch(
...
+                        packed_media_mask,
+                        local_media_mask,
+                        _,
+                        _,
+                        _,
+                    ) = _pack_sequences_for_megatron(
+                        # Pack in the token dtype: padding is filled with 0,
+                        # which is a valid token id but not a valid bool. Read
+                        # the dtype off the unpacked ids, since the local
+                        # input_ids is already the packed tensor here.
```
yfw: [`nemo_rl/models/megatron/data.py:536-539`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L536-L539)

**1 action item.** Comment-only — the code is correct.

The stated premise is false: `0` **is** a valid bool fill. Measured:

```
>>> torch.nn.functional.pad(torch.tensor([True, False, True]), (0, 2), value=0)
tensor([True, False, True, False, False])   # dtype=torch.bool
```

`torch.cat` and `index_select` preserve `torch.bool` as well, so packing the mask as a bool would give an identical result — the cast is defensive, not load-bearing. That matters because this comment is the only explanation in the codebase for why the cast exists, so as written it will convince the next maintainer the cast is required.

**Action:** keep the cast, fix the reason.

```suggestion
                        # Pack in the token dtype so this call mirrors the
                        # input_ids one exactly; packing the bool directly
                        # would also work, since F.pad fills with False. Read
                        # the dtype off the unpacked ids, since the local
                        # input_ids is already the packed tensor here.
```

nemo_rl/models/megatron/data.py
yfw: [`nemo_rl/models/megatron/data.py:528`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L528)

**4 action items, all already written and passing on our side.**

**TL;DR — neither new test file exercises a line this PR adds to `process_microbatch`; the `mtp_loss_mask` path it mirrors has six such tests and the mask has zero.**

Both new tests call [`_pack_sequences_for_megatron`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L960) directly, which is pre-existing and untouched by this PR. `grep -rn "media_token_validity_mask" tests/` outside the two new files returns nothing, so the four blocks that *are* this PR's contribution to `data.py` have no coverage.

`mtp_loss_mask` — the path this PR mirrors line-for-line — has one test per branch:

| existing `mtp_loss_mask` test | mirrors this uncovered block |
|---|---|
| [`test_megatron_data.py:572`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L572) | [`data.py:528-549`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L528-L549) packing |
| [`:435`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L435) | [`:553-557`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L553-L557) CP-slicing ternary |
| [`:312`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L312) | [`:610-613`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L610-L613) unpacked branch |
| [`:848`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L848) | [`:315-325`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/megatron/data.py#L315-L325) `NotImplementedError` |

Those siblings also already defend against the exact false-pass a mirrored feature invites — [`:453-463`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L453-L463) deliberately uses distinct tensors so a wrong one cannot "pass unnoticed, since both sides would then be identical".

### AI-1 — four `process_microbatch` tests
Mirror the four above. Happy to hand over the source we wrote and ran.

### AI-2 — two capability-probe tests
[`_model_media_placeholder_token_id`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/policy/workers/megatron_policy_worker.py#L187) and [`_model_accepts_media_token_validity_mask`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/models/policy/workers/megatron_policy_worker.py#L200) have none, while all three sibling probes in the same file do — [`test_megatron_worker.py:50-57`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/policy/test_megatron_worker.py#L50-L57), [`:60-69`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/policy/test_megatron_worker.py#L60-L69), [`:140-149`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/policy/test_megatron_worker.py#L140-L149). ~12 lines each with a duck-typed stub. The highest-value case is a `forward(self, x, **kwargs)` stub asserting `False`, since that is exactly the silent-swallow your docstring justifies the signature check by — and [`NemotronOmniModel.forward`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L647) really does have `**kwargs` followed by [`del kwargs`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py#L656). *Our sketch here is unverified against the real module: `transformer_engine` is unavailable in our environment, as it is for the pre-existing tests in that file.*

### AI-3 — assert the property the padding test is named for
[`test_media_mask_packing.py:74`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_media_mask_packing.py#L74) is `assert IMG != 0`, comparing two module-level constants, so it can never fail. Meanwhile no assertion touches `packed_mask` at the padded positions the test is named for. `assert not bool(packed_mask[0, 3:].any())` closes that gap — we ran it against the real packing and it passes.

### AI-4 — parametrise `cp_size` while you are there
Both tests run at `cp_size=1`, where [`_get_tokens_on_this_cp_rank`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/nemo_rl/distributed/model_utils.py#L1199) is never reached — so the packed-vs-CP-local ternary, which exists *only* for CP>1, is untested. [`test_shard_routed_experts_for_cp_matches_input_ids_zigzag`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/tests/unit/models/megatron/test_megatron_data.py#L1597) is the in-file pattern to copy. We mutation-checked ours: packing the mask at `cp_rank=0` while `input_ids` uses `cp_rank` makes it fail at `[rank1-cp2]`, so it discriminates rather than passing on a no-op.

docs/design-docs/media-token-validity-mask.md
```diff
@@ -0,0 +1,214 @@
...
+A media placeholder is an ordinary vocabulary entry. Text can legitimately
+contain it, and until now that text could not be trained on: the model read
+every occurrence as an anchor for a projected image feature and failed when no
+feature existed. The workaround was a sanitizer that rewrote the offending
+tokens before they reached the model, which changed the data to fit the model.
+
+This design removes the sanitizer and replaces it with a **media-token validity
+mask**: the caller, which knows how many media items each row carries, marks
+which placeholder positions are real anchors. Positions it does not mark keep
+their ordinary token embedding.
```
yfw: [`docs/design-docs/media-token-validity-mask.md:5-14`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L5-L14)

**1 action item.**

**TL;DR — the doc is written as a history of replacing a sanitizer, but that sanitizer never existed on `main`, so the reader has no prior state to compare against; suggest restating it as what the design does rather than how it was arrived at.**

### The sanitizer is not part of this repo's history

```
$ git log --oneline origin/main -S sanitize_nemo_gym_example_image_placeholders   ->  0 commits
$ git log --oneline origin/main -S sanitize_image_placeholders                    ->  0 commits
$ git merge-base --is-ancestor a4ac69648 origin/main                              ->  not an ancestor
```

A repo-wide grep for `sanitize_nemo_gym_example_image_placeholders`, `_normalize_image_placeholders`, `_count_image_payloads` and `sanitize_image_placeholders` matches only this document. #3494 was squash-merged, and the sanitizer was both added and removed inside that branch, so it netted to zero and reached neither `main` nor this branch.

So sections written against it — [`## Why not sanitize`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L66), the [`**Removed.** ... 212 lines`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L190-L192) paragraph, and [`Without the sanitizer there is no safety net`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L194) — ask the reader to evaluate a comparison whose other half they cannot check out, and the "Removed" line claims a deletion no merged commit performed.

### Action: restate the design rather than the path to it

To be clear, recording a rejected alternative is a legitimate thing for a design doc to do — the issue is only that this particular alternative never shipped, so it reads as one branch's development history rather than the repo's. Keeping a short "alternatives considered" paragraph would be entirely reasonable; three sections plus a past-tense `## Consequences` is what makes it hard to follow.

| Section | Suggested |
|---|---|
| [`## Summary`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L5-L14) | rewrite — see block below |
| [`## The problem`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L16) | **keep** — the 2,539-rows-in-100k motivation is concrete and checkable |
| [`## Why not sanitize`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L66) | cut, or compress to one "alternatives considered" paragraph |
| [`## Design`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L82) / [packing](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L118) / [capability detection](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L164) | **keep** — the load-bearing part, and the packing rationale is the most valuable thing here |
| [`## Consequences`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L188-L198) | rewrite — nothing is removed; "data prep is required" stands on its own |
| [`## Validation`](https://github.com/NVIDIA-NeMo/RL/blob/93f16afb1502a763ab91a88c354cbd55c09de92b/docs/design-docs/media-token-validity-mask.md#L203-L208) | drop row 1 — the PR body says that run had "the mask entirely unwired", so it measures a different change |

A starting point for the Summary:

```suggestion
A media placeholder is an ordinary vocabulary entry, so text can legitimately
contain it — a competitive-programming statement that spells `<image>` in its
prose, for instance. The model reads every occurrence as an anchor for a
projected image feature and fails when no feature exists, so those rows cannot
be trained on.

This design adds a **media-token validity mask**: the caller, which knows how
many media items each row carries, marks which placeholder positions are real
anchors. Positions it does not mark are skipped by the media merge and keep
whatever language embedding the forward gave them.
```

yfw:
I think we don't need the Megatron-Bridge bump in this PR anymore because of https://github.com/NVIDIA-NeMo/RL/commit/04a687d7290ae212ccbc457830fadea24402cd14. Also I don't think anything needs to change in uv.lock anymore?

yfw (approved):

PR #3633, opened 2026-08-13
feat(nemo-gym): expose resolved config entries

ananthsub:
## Summary

Adds `NemoGym.list_entries()` so NeMo RL can inspect the runnable entries that Gym resolved during startup.

The input configuration contains `config_paths`, but those paths do not directly describe the agents, models, and resource servers that Gym loaded. Reading the resolved configuration gives shard discovery an accurate inventory without reproducing Gym's configuration merge in NeMo RL.

`list_entries()` returns each resolved entry name and the server-type groups whose nested configuration has an `entrypoint`. Gym does not start a process for a leaf without an `entrypoint`, so those names are excluded from the routing inventory. Calling `list_entries()` before `_spinup()` fails because the actor does not yet have a resolved Gym configuration.

## Stack

This PR depends on #3368. The shard configuration schema follows in #3369.

### DIFF
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -62,6 +62,15 @@ from nemo_rl.utils.venvs import make_actor_runtime_env
 NEMO_GYM_ACTOR_FQN = "nemo_rl.environments.nemo_gym.NemoGym"
 NEMO_GYM_GRACEFUL_SHUTDOWN_TIMEOUT_S = 120
 
+# The three server-type keys Gym nests under a top-level config entry. Gym's
+# constant is private (nemo_gym.discovery._SERVER_GROUP_KEYS), and the literal
+# list also appears in global_config.py, config_types.py, and cli/env.py.
+GYM_SERVER_TYPE_KEYS = (
+    "responses_api_agents",
+    "responses_api_models",
+    "resources_servers",
+)
+
 # Kept local so the Gym actor does not depend on model-config dtype resolution.
 # Must cover every name resolve_routed_experts_dtype can produce.
 _ROUTED_EXPERTS_DTYPES = {
@@ -589,6 +598,51 @@ Depending on your data shape, you may want to change these values."""
                 f"HTTP {response.status} {await response.text()}"
             )
         return await response.json()
+
+    def list_entries(self) -> Dict[str, List[str]]:
+        """Report which config entries this actor actually spawned.
+
+        Returns ``{entry_name: [server_type_keys]}`` read from Gym's *resolved*
+        config, so entries that arrived via ``config_paths`` are included. The
+        config NeMo RL passed in is not a substitute: it still holds
+        ``config_paths`` as file paths and none of the entries they expand
+        into, so reading it would miss every agent and judge loaded from a
+        path.
+
+        Entries whose server config has no ``entrypoint`` are omitted because
+        Gym does not start a process for them.
+
+        Callers compare these names across actors to build the agent->shard map
+        and to catch an entry duplicated across shards. Names are all that is
+        interpreted; what an entry *means* is Gym's business.
+        """
+        if self.rh is None:
+            raise RuntimeError(
+                "list_entries() needs a running Gym stack; call _spinup() first."
+            )
+
+        from nemo_gym.global_config import get_global_config_dict
+        from omegaconf import DictConfig
+
+        resolved = get_global_config_dict()
+        entries: Dict[str, List[str]] = {}
+        for name, entry in resolved.items():
+            if not isinstance(entry, (dict, DictConfig)):
+                continue
+            # Fixed key order so the map is stable across actors and runs.
+            types = []
+            for key in GYM_SERVER_TYPE_KEYS:
+                server_group = entry.get(key)
+                if not isinstance(server_group, (dict, DictConfig)):
+                    continue
+                if any(
+                    isinstance(server, (dict, DictConfig)) and "entrypoint" in server
+                    for server in server_group.values()
+                ):
+                    types.append(key)
+            if types:
+                entries[str(name)] = types
+        return entries
 
     async def run_rollouts(
         self,
--- a/tests/unit/environments/test_nemo_gym_utils.py
+++ b/tests/unit/environments/test_nemo_gym_utils.py
@@ -18,9 +18,13 @@ These run in the default L0 suite. Keep this module free of heavy imports
 """
 
 import copy
+import sys
+import types
+from contextlib import contextmanager
 from unittest.mock import MagicMock, call, patch
 
 import pytest
+from omegaconf import DictConfig
 
 from nemo_rl.environments import nemo_gym as nemo_gym_mod
 from nemo_rl.environments.nemo_gym import (
@@ -403,3 +407,90 @@ def test_nemo_gym_shutdown_before_spinup_is_a_noop():
     actor.__init__({})
 
     actor.shutdown()  # must not raise
+
+
+@contextmanager
+def _stub_gym_resolved_config(resolved):
+    """Stand in for nemo_gym.global_config, which lives in the actor's venv."""
+    package = types.ModuleType("nemo_gym")
+    module = types.ModuleType("nemo_gym.global_config")
+    module.get_global_config_dict = lambda: resolved
+    with patch.dict(
+        sys.modules, {"nemo_gym": package, "nemo_gym.global_config": module}
+    ):
+        yield
+
+
+def _spun_up_actor():
+    cls = nemo_gym_mod.NemoGym.__ray_metadata__.modified_class
+    actor = cls.__new__(cls)
+    actor.__init__({})
+    actor.rh = MagicMock()
+    return actor
+
+
+def test_list_entries_reports_entry_names_and_server_types():
+    resolved = DictConfig(
+        {
+            "math_agent": {
+                "responses_api_agents": {"simple_agent": {"entrypoint": "app.py"}}
+            },
+            "math_env": {"resources_servers": {"math": {"entrypoint": "app.py"}}},
+            # An entry can carry more than one server type.
+            "judge": {
+                "responses_api_models": {"local_vllm_model": {"entrypoint": "app.py"}},
+                "resources_servers": {"judge_tools": {"entrypoint": "app.py"}},
+            },
+            # Plain Gym settings are not entries.
+            "port_range_low": 5000,
+            "default_host": "10.0.0.1",
+            "config_paths": ["a.yaml"],
+        }
+    )
+
+    with _stub_gym_resolved_config(resolved):
+        entries = _spun_up_actor().list_entries()
+
+    assert entries == {
+        "math_agent": ["responses_api_agents"],
+        "math_env": ["resources_servers"],
+        "judge": ["responses_api_models", "resources_servers"],
+    }
+
+
+def test_list_entries_skips_dicts_that_hold_no_server_type():
+    """A dict-shaped setting is not an entry unless it nests a server type."""
+    resolved = DictConfig(
+        {
+            "real_entry": {"resources_servers": {"env": {"entrypoint": "app.py"}}},
+            "some_setting": {"nested": "value"},
+        }
+    )
+
+    with _stub_gym_resolved_config(resolved):
+        entries = _spun_up_actor().list_entries()
+
+    assert entries == {"real_entry": ["resources_servers"]}
+
+
+def test_list_entries_skips_an_entry_that_starts_no_server():
+    resolved = DictConfig(
+        {
+            "math_env": {"resources_servers": {"math": {"entrypoint": "app.py"}}},
+            "code_gen": {"resources_servers": {"code": {"host": "10.0.0.1"}}},
+        }
+    )
+
+    with _stub_gym_resolved_config(resolved):
+        entries = _spun_up_actor().list_entries()
+
+    assert entries == {"math_env": ["resources_servers"]}
+
+
+def test_list_entries_before_spinup_raises():
+    cls = nemo_gym_mod.NemoGym.__ray_metadata__.modified_class
+    actor = cls.__new__(cls)
+    actor.__init__({})
+
+    with pytest.raises(RuntimeError, match="call _spinup"):
+        actor.list_entries()


nemo_rl/environments/nemo_gym.py
```diff
@@ -490,6 +666,39 @@ def set_tokenizer(self, tokenizer: PreTrainedTokenizerBase) -> None:
         """
         self._tokenizer = tokenizer
 
+    def list_entries(self) -> Dict[str, List[str]]:
+        """Report which config entries this actor actually spawned.
```
terrykong, line 503: **1 action item, 1 follow-up.**

**TL;DR — the docstring says this reports entries the actor "actually spawned", but the loop reports every resolved entry that has a server-type key, including entries Gym never started anything for.**

Gym starts a server only when the leaf config carries an `entrypoint` — [`RunHelper.start`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L407) walks two levels down and skips anything without one. Nothing warns: [`is_almost_server`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L708) returns True only when the nested config *has* an entrypoint, so `error_on_almost_servers` (default True) never fires for this shape. And [the loop here](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/nemo_rl/environments/nemo_gym.py#L697) checks only that a server-type key is present.

A live one: `math_formal_lean_refinement_agent` at [`ifbench_teacher.yaml#L644`](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/examples/nemo_gym/nemotron-3-ultra/ifbench_teacher.yaml#L644) is a `responses_api_agents` entry with no nested entrypoint, and no config path supplies one — [that file's `config_paths` block](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/examples/nemo_gym/nemotron-3-ultra/ifbench_teacher.yaml#L415-L421) lists six unrelated files. The name resolves and Gym starts nothing for it, so the agent->shard map [`build_agent_shard_map`](https://github.com/NVIDIA-NeMo/RL/pull/3371) builds can contain an agent that was never started. Nothing reads that map to route work yet, but its "an agent must live in exactly one shard" check runs on these names too — if one collides with a real agent of the same name in another shard, the resulting `ShardSetupError` names two shards when only one entry was ever started.

Fixing the recipes is not this PR's job — the finding is that the docstring promises something the method does not deliver.

### AI-1 — say what the method reports, not what ran

```suggestion
        """Report the config entries in this actor's resolved Gym config.
```

Please also add one sentence in the docstring body: an entry can show up here without Gym having started anything for it, because Gym starts a server only when the leaf has an `entrypoint`. [`math_formal_lean_refinement_agent`](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/examples/nemo_gym/nemotron-3-ultra/ifbench_teacher.yaml#L644) is the live example to name.

### Follow-up — filter on a nested entrypoint (your call)

If you want the stronger version, match [Gym's own rule](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L407) and only report an entry whose leaf has an `entrypoint`. Then the docstring's original wording becomes true. The cost is real and worth weighing: the three existing tests break. Entirely optional — AI-1 alone is enough.

<details>
<summary>The entrypoint filter, and what it costs the tests</summary>

Two of the three tests fail under the filter — [`test_list_entries_reports_entry_names_and_server_types`](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/tests/unit/environments/test_nemo_gym_utils.py#L312) and [`test_list_entries_skips_dicts_that_hold_no_server_type`](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/tests/unit/environments/test_nemo_gym_utils.py#L339) — both returning `{}`, because no fixture leaf has an entrypoint. No assertions need to change; only the fixture leaves do (`{}` -> `{"entrypoint": "app.py"}`).

This test fails on today's code (`1 failed, 3 passed`) and passes under the filter (`4 passed`):

```python
def test_list_entries_skips_an_entry_that_starts_no_server():
    """Gym starts a server only when its leaf carries an entrypoint.

    A recipe can leave an entry's config path commented out, so the entry name
    still resolves but nothing runs behind it. Reporting it would put a name in
    the agent->shard map for something that was never started.
    """
    resolved = DictConfig(
        {
            "math_env": {"resources_servers": {"math": {"entrypoint": "app.py"}}},
            "code_gen": {"resources_servers": {"code": {"host": "10.0.0.1"}}},
        }
    )

    with _stub_gym_resolved_config(resolved):
        entries = _spun_up_actor().list_entries()

    assert entries == {"math_env": ["resources_servers"]}
```

Scale: 7 such entries across 2 shipped recipes. The second one is `code_gen` at [`grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L331`](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L331), whose config path is [commented out at L319](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L319).
</details>

nemo_rl/environments/nemo_gym.py
```diff
@@ -38,28 +41,28 @@
...
 from nemo_rl.models.policy import PolicyConfig, TokenizerConfig
 from nemo_rl.utils.routed_experts_codec import decode_routed_experts
 from nemo_rl.utils.timer import Timer
 from nemo_rl.utils.venvs import make_actor_runtime_env
 
 NEMO_GYM_ACTOR_FQN = "nemo_rl.environments.nemo_gym.NemoGym"
 
+# The three server-type keys Gym nests under a top-level config entry. Gym has
+# no exported constant for these -- it repeats the same literal list internally
+# (global_config.py, config_types.py, cli/eval.py) -- so it is restated here.
```

**TL;DR — one of the three files this comment cites does not hold the three-key list, so a maintainer who follows the pointer finds nothing and cannot tell whether the duplicated constant is still needed.**

[`cli/eval.py#L216`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/eval.py#L216) checks the single key `"responses_api_agents"`, not the list. The other two are right: [`global_config.py#L1091`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/global_config.py#L1091) and [`config_types.py#L716`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/config_types.py#L716). Two places are missed, including the closest thing Gym has to a real constant: [`discovery._SERVER_GROUP_KEYS`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/discovery.py#L71) (private, so "no exported constant" still holds) and [`cli/env.py#L985`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L985).

### AI-1 — point at the places that hold the list

```suggestion
# The three server-type keys Gym nests under a top-level config entry. Gym's own
# constant for these is private (`nemo_gym.discovery._SERVER_GROUP_KEYS`) and it
# repeats the literal list elsewhere (global_config.py, config_types.py,
# cli/env.py), so it is restated here.
```

tests/unit/environments/test_nemo_gym_utils.py
```diff
@@ -300,3 +287,74 @@ def test_nemo_gym_shutdown_before_spinup_is_a_noop():
...
+def test_list_entries_reports_entry_names_and_server_types():
+    resolved = DictConfig(
+        {
+            "math_agent": {"responses_api_agents": {"simple_agent": {}}},
+            "math_env": {"resources_servers": {"math": {}}},
+            # An entry can carry more than one server type.
+            "judge": {
+                "responses_api_models": {"local_vllm_model": {}},
+                "resources_servers": {"judge_tools": {}},
+            },
```
terrykong: **1 action item.**

**TL;DR — the `judge` fixture declares its two server-type keys in the same order `GYM_SERVER_TYPE_KEYS` lists them, so the one test that could check the fixed key order cannot fail if that order is lost.**

[`list_entries()`](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/nemo_rl/environments/nemo_gym.py#L669) iterates [`GYM_SERVER_TYPE_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/nemo_rl/environments/nemo_gym.py#L60) instead of the entry's own keys on purpose — [the comment says "Fixed key order so the map is stable across actors and runs"](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/nemo_rl/environments/nemo_gym.py#L696). With the fixture as written, declaration order and constant order give the same answer, so the assertion cannot tell them apart. Swapping the implementation to declaration order (`types = [key for key in entry if key in GYM_SERVER_TYPE_KEYS]`) still gives `3 passed`. With the two fixture lines swapped, that same break gives `1 failed`.

### AI-1 — declare the two keys back-to-front

The existing assertion stays as is; `3 passed` with this applied, on both current and fixed code.

```suggestion
            # Legal but rare: the schema allows a second server-type key, though
            # Gym starts only the first. Declared back-to-front here, so the
            # report can only come out in GYM_SERVER_TYPE_KEYS order.
            "judge": {
                "resources_servers": {"judge_tools": {}},
                "responses_api_models": {"local_vllm_model": {}},
            },
```

This also replaces the current comment, "An entry can carry more than one server type", which reads as if Gym runs both. It does not — [`RunHelper.start`](https://github.com/NVIDIA-NeMo/Gym/blob/c3bac96314a59f28b896f597eb9845d175bb0252/nemo_gym/cli/env.py#L398) takes `list(server_config_dict)[0]` and starts one. The shape is still legal: Gym's pydantic models accept a two-server-type entry (`extra="allow"`, undiscriminated union).

terrykong (commented):
Nice, small addition — thanks for splitting it out of #3371. The method is reachable and used: #3371's `_discover_agent_shard_map` calls it, and both halves of the return shape do work — the values pick out agents, the keys drive the cross-shard duplicate check. Nothing here blocks the merge; both comments below are wording fixes.

Two things worth calling out as good. The [docstring's note that the passed-in config still holds `config_paths` as unexpanded file paths](https://github.com/NVIDIA-NeMo/RL/blob/7477f97ce6ee270c1efb595ce2340dc6c4958a6c/nemo_rl/environments/nemo_gym.py#L672) heads off the obvious wrong turn — reading the config NeMo RL passed in instead of Gym's resolved one. That argument is the clearest case for why this method has to exist at all, and it is worth keeping prominent when you reword the first line. And the actor/driver split is clean: the actor reports facts, #3371 does the interpreting.

Three things we checked and are not asking you to change, noted so a later reviewer does not re-open them:

- **`Dict[str, List[str]]` is the right return shape.** Narrowing to one type per entry would make NeMo RL hard-code a winner, and Gym's own pick is content-dependent — it flips on whether `domain` is present.
- **The allow-list of three keys beats Gym's `NEMO_GYM_RESERVED_TOP_LEVEL_KEYS` deny-list.** `_spinup` injects top-level keys that are not reserved (`policy_model_name`, `policy_api_key`, `policy_base_url`, the aiohttp limits), and a deny-list would report those as entries.
- **We did not ask for a test guarding `GYM_SERVER_TYPE_KEYS` against Gym adding a fourth server type.** Nobody has named a planned fourth type, and the test would have to live in a file this PR does not touch. The corrected citation list in the comment on the constant gives the accurate set of places to check on the next submodule bump.

One note rather than an ask: iterating `GYM_SERVER_TYPE_KEYS` already makes each entry's type list independent of the order the keys appear in the YAML, which is the useful part. We looked for something that depends on that order and did not find one — in #3371 the list is only ever tested with `in`, and the map it prints is agent name to shard name. So the "stable across actors and runs" comment is doing less work than it sounds like. Fine to leave.

_Generated by Claude Code_

nemo_rl/environments/nemo_gym.py
ananthsub: Fixed in 88d9d83a. `list_entries()` now reports only server-type groups with a nested `entrypoint`, matching the condition Gym uses before starting a process. The fixtures now model runnable entries, and a regression test covers a resolved entry with no process behind it.

nemo_rl/environments/nemo_gym.py
ananthsub: Fixed in 88d9d83a. The comment now points to the private `nemo_gym.discovery._SERVER_GROUP_KEYS` constant and the three files that repeat the full list; the incorrect `cli/eval.py` pointer was removed.

terrykong (dismissed):
Both comments addressed — thanks.

You took the stronger option on the first one: `list_entries()` now requires a nested `entrypoint`, which matches the condition Gym itself uses before starting a process, so the docstring's original wording is true rather than just softened. The new `test_list_entries_skips_an_entry_that_starts_no_server` is the case that was missing.

One heads-up, not a blocker: `fast_L1_Functional_Tests_PPO` is failing at this head (and `CI quality check` just reports that same failure). It looks unrelated to this change — `list_entries()` has no callers yet, and that job is a DAPO-math run that never builds a Gym actor. Worth a glance before merge in case it came in with the rebase.

Also minor: your two "Fixed in 88d9d83a" replies point at a commit the later rebase erased, so those links 404. The fix itself is present at 93d556c2cdffd715ab8a94987efdb1a8f11c8774.

terrykong (approved):

PR #3635, opened 2026-08-14
fix(math-env): one score per sample when answer extraction fails

tianyi-zhang-02:
# What does this PR do ?

`HFVerifyWorker.verify` appends a sample's score before it tries to extract that sample's answer:

```python
results.append(float(ret_score))

if return_extracted_answer:
    assert extracted_answer is not None
    assert len(extracted_answer) == 2
    ...
    extracted_answers.append(extracted_prediction[0][0])

except (Exception, TimeoutException):
    results.append(0.0)
    extracted_answers.append(None)
```

math-verify returns an **empty** prediction list whenever it parses nothing — `"I don't know."`, an empty or truncated generation — so `[0][0]` raises `IndexError`, control lands in the `except`, and that sample contributes a **second** score. Every later reward is then one slot out of position relative to the batch.

Nothing downstream notices. `eval_pass_k` splits the rewards positionally against `num_tests_per_prompt`, so a longer tensor shifts every group boundary and it reports a **wrong number rather than failing**.

Second bug in the same line: `extracted_prediction` is a flat `list[str]`, so `[0][0]` is the first *character*. A wrong answer of `137` is recorded as `"1"` — which then feeds cons@k majority voting, where distinct wrong answers collide.

Reproduced against the installed math-verify, batch of five with two unparseable generations:

```
before:  len(results) = 7 for a batch of 5    answers = ['42', None, '1',   None, '42']
after:   len(results) = 5                     answers = ['42', None, '137', None, '42']
```

The fix computes the score and the answer into locals and appends exactly once per sample. Extraction moves into a helper that returns `None` instead of raising, and sits in an `else` branch so that failing to record an answer can never retract a score the verifier already returned.

This is the shape `HFMultiRewardVerifyWorker` in the same file already uses: compute every reward first, commit them together, and let the `except` be the only other writer. `HFVerifyWorker` was the one that committed mid-way.

# Issues

No existing issue.

# Usage

No config change. `nemo_rl/evals/eval.py:387` passes `return_extracted_answer` positionally as `True`, unconditionally, for every eval — which is why grepping the flag name doesn't find the caller.

The affected set is the eval configs that leave `verifier_type` at its default `"math"`, i.e. the ones routed to `HFVerifyWorker`: `eval.yaml` (AIME 2024), `math_eval.yaml`, `local_eval.yaml`. Unparseable generations are common on those, so runs have been reporting a slightly wrong pass@k.

`mmlu.yaml`, `mmlu_zh_cn.yaml`, `gpqa_eval.yaml` and `mmlu_pro.yaml` set `verifier_type` to one of the multichoice values, and `verifier_type` selects the worker class outright (`MathEnvironment.WORKER_CLASS_DICT`), so they never reach this code. `MultilingualMultichoiceVerifyWorker` and `EnglishMultichoiceVerifyWorker` append once per sample with no `try`/`except`, so they are correct as they stand.

- [x] Did you run the unit tests and functional tests locally? (unit only)
- [x] Did you add or update any necessary documentation? (n/a — no user-facing surface changes)

# Additional Information

`grep -rn "return_extracted_answer" tests/` previously matched only `test_code_jaccard_environment.py`, so this path had no coverage — which is how it survived. Two tests added: one drives a batch containing an unparseable generation and a parseable wrong answer, and one pins the helper's never-raises contract over the shapes math-verify and the dapo branch actually return. The first fails on `main` with *"one score per sample; an unparseable generation used to add a second"*. The file goes 14 → 22 passed. `ruff` clean.

Two things I would rather you decided than assumed:

1. **The `dapo_math_verify` branch** sets `extracted_answer` to a plain string, so the old `len(...) == 2` assert only held for two-character answers. The helper now returns that string as-is. No shipped eval config selects this branch, so I could not exercise it end to end.
2. **Whether to record the first prediction at all** when nothing matched. I kept the existing intent and only fixed it to be the prediction rather than its first character; recording `None` would arguably be cleaner for cons@k.

I have not run the eval functional tests — those need a served model. The reproduction above uses the real `math_metric` with the same arguments the worker passes.

cc @yuki-97


tianyi-zhang-02:
Correction to the description, since it was wrong in a way a reviewer would have had to catch: I listed MMLU and GPQA as affected. They aren't.

`verifier_type` picks the worker class outright via `MathEnvironment.WORKER_CLASS_DICT`, and those configs set it to a multichoice value, so they never reach `HFVerifyWorker`. Both multichoice workers append once per sample with no `try`/`except` — nothing to fix there.

The real affected set is narrower: the configs that leave `verifier_type` at its default `"math"` — `eval.yaml` (AIME 2024), `math_eval.yaml`, `local_eval.yaml`.

Description updated. The bug itself and the reproduction are unchanged.

yuki-97 (commented):
@jinglinglingling to review

jinglinglingling (approved):

tests/unit/environments/test_math_environment.py
```diff
@@ -507,3 +507,63 @@ def test_math_timeout_handling(math_env):
...
+    metadata = [{"ground_truth": "42"} for _ in responses]
+
+    result = ray.get(math_env.step.remote(message_log_batch, metadata, True))
+
+    assert result.rewards.shape == (3,), (
+        "one score per sample; an unparseable generation used to add a second"
+    )
+    assert result.rewards[0] == 1.0
+    assert result.rewards[2] == 0.0, "the wrong answer keeps its own slot"
+    assert len(result.answers) == 3
```
yuki-97: This pins two of the three paths — extraction succeeded (samples 0/2) and extraction returned nothing (sample 1). The third, **an exception during scoring with `return_extracted_answer=True`**, has no coverage.

That path is the one relying on the `extracted: str | None = None` init at the top of the loop (`nemo_rl/environments/math_environment.py:124`). If that init is ever moved or made conditional, `extracted_answers` loses a slot again — silently, exactly like before this PR.

Two exception cases already exist, but both call `step` with the default `return_extracted_answer=False`, so neither enters the block:
- `test_math_exception_handling` (`tests/unit/environments/test_math_environment.py:446`)
- `test_math_timeout_handling` (`:463`)

Passing `True` there plus one pairing assert would close it:

```python
result = ray.get(math_env.step.remote(message_log_batch, metadata, True))
...
assert len(result.answers) == result.rewards.shape[0] == 1
```

Asserting the pairing rather than `result.answers == [None]` on purpose: I haven't run these, and whether each lands in `except` or `else` depends on which exceptions math-verify swallows internally. If both turn out to take `else`, a `ground_truth` whose gold fails to parse would force `except` (`math_verify/metric.py` raises `ValueError("No gold targets found ...")`).

yuki-97 (approved):
LGTM. Thanks for digging this out — good catch on both the first-character truncation and the double-append.

Checked:

- **The logic changes are correct.** `math_metric`'s `str_preds` is a flat `list[str]`, so the old `[0][0]` took the first *character*; and since the asserts fired after `results.append(float(ret_score))`, every failure appended a second score. Both trigger conditions are reachable, not defensive-only. The `dapo_math_verify` branch is a real fix too — `compute_score` returns `pred` as a plain string, so `assert len(...) == 2` was a length check on a string.
- **Outside the `try` is safe.** `grader.verify` (math-verify 0.9.0) swallows everything and returns `False`; its only unconditional re-raise is the "threaded environment" `ValueError`, unreachable here because `HFVerifyWorker` is a default `@ray.remote` actor running on the worker process's main thread — and off it, `self.verify_func(...)` inside the `try` would fail first anyway.

On why this stayed hidden: training never reaches the block — `nemo_rl/experience/rollouts.py:712` uses the default `return_extracted_answer=False`. Only `nemo_rl/evals/eval.py:387` passes `True`, and the only metric any shipped config enables is `pass@k`, which never reads `answers`, so the `[0][0]` truncation was invisible. The double-append did shift `pass@k` rewards, but silently — `rewards.split(num_tests_per_prompt)` doesn't require divisibility. `cons@k` would have caught it loudly (`eval.py:251` asserts `len(extracted_answers) == len(rewards)`), but no config turns it on.

PR #3636, opened 2026-08-14
feat(modelopt): generalize real-quant vLLM refit

mxinO:
## Summary

- replace the W4A4/W4A16-specific packing and vLLM reload implementation with ModelOpt deployment tensors produced through Megatron-Bridge
- configure vLLM from the initialized policy descriptor and use its public start/load/finish weight-update transaction for colocated IPC and non-colocated collective refit
- support CPU-offloaded export tensors without increasing persistent GPU memory
- make refit failures atomic and terminal so partially updated rollout engines are never reused
- add Qwen3/Nemotron-Super W4A4 and Nano3 W4A16 real-quant recipes, documentation, and focused functional coverage
- require eager execution for CUTLASS routed-MoE real quant because CUDA graphs otherwise retain pre-refit activation-scale tensors

Design and implementation-plan documents are included. No compatibility layer for the previous internal real-quant exporter is retained.

Depends on:
- NVIDIA/Model-Optimizer#2190
- NVIDIA-NeMo/Megatron-Bridge#5575

## Testing

- Ruff plus 71 focused unit tests; MCore-only policy test passed
- non-colocated collective functional test passed W4A16 real quant (KL 0.000896, probability ratio 1.0230) and W4A8 fake quant (KL 0.00514, ratio 1.0629)
- Qwen3-30B-A3B W4A4 completed two refits and two training steps with losses 0.0003/0.0002 and KL 0.0250/0.0217
- Nano3-30B-A3B W4A16 completed two refits across hybrid Mamba/MoE layers and two training steps with losses 0.0001/0.0001 and KL 0.0013/0.0008
- fixed-weight export and all 16 post-refit vLLM workers matched packed weights/scales bitwise in the Qwen3 diagnostic

The two large-model Slurm jobs completed all application steps and wrote artifacts, then exited nonzero during Python finalization when Ray attempted to initialize a second core worker. That cleanup-only Ray issue is separate from the refit path and is called out here rather than hidden.


mxinO:
Closing this draft while the real-quant refit design is reworked and validated locally. The branch is retained for reference.

PR #3638, opened 2026-08-14
feat(vllm): support refit-safe MXFP8 linear backends

seonjinn:
# What does this PR do?

Allows NeMo-RL MXFP8 rollout refit to preserve an explicitly selected vLLM CuTeDSL or TRTLLM dense linear backend.

These native kernels replace checkpoint-layout parameters with backend-specific runtime layouts. Re-running their post-load hook directly during refit would replace CUDA-Graph-visible storage. This change uses vLLM's layerwise reload lifecycle for native MXFP8 linear layers, so each transformed weight is copied into the original runtime storage. Failed or incomplete reloads restore saved runtime parameters without processing partial checkpoint tensors.

The current `auto` behavior remains unchanged: NeMo-RL keeps FlashInfer CUTLASS as its default refit backend. Existing CUTLASS and MoE refit paths are unchanged.

TRTLLM selection depends on [vLLM PR #52204](https://github.com/vllm-project/vllm/pull/52204). NeMo-RL's vLLM pin must include that change before this PR can be marked ready.

# Issues

Depends on vLLM PR #52204.

# Usage

```yaml
policy:
  generation:
    vllm_cfg:
      precision: fp8
      is_mx: true
    vllm_kwargs:
      linear_backend: flashinfer_cutedsl
      # Use flashinfer_trtllm after vLLM PR #52204 is available.
```

- [x] Added repeated-refit, layout-change, failure-cleanup, backend-default, and scale-routing tests.
- [x] Ran the affected vLLM unit suites: `50 passed`.
- [x] Ran Ruff check and format check on all changed files.
- [x] Ran Pyrefly: `0 errors`.
- [ ] Run a real GPU refit with the TRTLLM kernel after vLLM PR #52204 is included in the dependency pin.

# Additional Information

The repeated-refit tests cover CuTeDSL-style transpose and TRTLLM-style padding/layout changes while asserting stable `Parameter` identity and `data_ptr()` across two refits. Failure-path tests cover partial transfer abort, initialization failure, and one-shot finalization cleanup.


seonjinn:
Self-review completed with independent correctness, test, and upstream-compatibility passes.

Resolved findings:
- Added an abort path that restores saved runtime tensors instead of processing incomplete checkpoint tensors.
- Protected partial initialization and made finalization one-shot while preserving the first error.
- Added CuTeDSL transpose and TRTLLM padding/layout regression cases over two refits.
- Added direct coverage for the existing CUTLASS default and backend-specific scale routing.
- Replaced class-name-only detection with dynamic `isinstance` checks against registered vLLM kernel types.

Validation:
- Affected vLLM unit suites: 50 passed.
- Ruff check and format check passed.
- Pyrefly: 0 errors.

Remaining dependency: the real TRTLLM GPU integration requires vLLM PR #52204 in NeMo-RL's dependency pin. This PR remains draft until that test can run.

PR #3640, opened 2026-08-14
ci: AUT-1111 support arbitrary Megatron-LM commits

chtruong814:
# What does this PR do?

Adds workflow-dispatch support for testing NeMo-RL against an arbitrary full Megatron-LM commit reachable from an NVIDIA/Megatron-LM branch.

It also adds optional test selection without changing the default CI label behavior:

- `test_script_pattern` filters unit and functional wrapper names using a case-insensitive substring.
- `test_platform` selects `all`, `h100`, or `gb200` test job matrices. Unit and Lfast tests remain H100-only.
- `CI:megatron` remains a functional-only label and applies the `megatron` script pattern internally. `megatron` is not a separate `test_to_run` dispatch tier.
- Functional wrappers with `gb200` in their name are excluded from the H100 matrix and retained in the GB200 matrix.
- Selections that do not produce any runnable tests for the requested level and platform fail during preflight instead of finishing green without tests.

With an empty pattern and the default `test_platform=all`, the existing CI matrices are preserved: 23 H100 unit wrappers, 17 H100 functional/Lfast wrappers, and 18 GB200 functional wrappers.

When `megatron_lm_commit` is supplied, one CPU preparation job checks out the requested commit, runs `uv lock` once, and publishes a run-scoped lockfile artifact. The H100 and GB200 container builds and downstream unit and functional test checkouts consume the same verified lockfile instead of resolving dependencies independently.

The artifact includes the NeMo-RL commit, Megatron-LM commit, and `uv.lock` checksum. Consumers verify all three before installing it. Artifact names include the workflow run attempt so reruns do not collide with immutable artifacts.

Example Run with Megatron-LM commit and filtered to a targeted test
https://github.com/NVIDIA-NeMo/RL/actions/runs/31825921864

# Usage

Run one functional wrapper on H100 against an arbitrary Megatron-LM commit:

```shell
gh workflow run cicd-main.yml \
  --repo NVIDIA-NeMo/RL \
  --ref <branch> \
  -f test_to_run=L1 \
  -f test_script_pattern=L1_Functional_Tests_Megatron_4 \
  -f test_platform=h100 \
  -f megatron_lm_commit=<full-40-character-sha>
```

Omit `megatron_lm_commit` to use the commit pinned by the repository. Omit the filtering inputs to retain the normal test matrix for the selected level.

**Pre checks**:

- [x] No additional unit-test fixture is needed for this workflow-only change; the selector and artifact installer were exercised locally.
- [ ] GPU unit and functional suites were not run locally; they require the GitHub Actions runner/container path.
- [x] No separate documentation update is required; the workflow inputs and this usage example document the operator interface.

# Validation

- Verified default selector output: 23 unit, 17 H100 functional/Lfast, and 18 GB200 functional wrappers.
- Verified filtered selector output for `Megatron_4`, `Mcore`, and `GB200_MXFP8`, including effective-zero validation.
- Parsed the modified workflow and composite-action YAML files.
- Passed Bash syntax checks for the Megatron helper and selector scripts.
- Exercised manifest, commit, checksum, and lockfile-install verification with a local fixture.
- Passed `git diff --check`.
- Actionlint reports only the pre-existing `matrix.is_optional` warning.

# Additional Information

The branch contains signed-off commits. To authorize PR CI for the current head:

```text
/ok to test e2e8d83bcd808b3a24eaaf690ba3eb0df1ccfdc6
```


balasaajay (dismissed):

kajalj22 (dismissed):

.github/workflows/cicd-main.yml
```diff
@@ -312,13 +416,86 @@ jobs:
...
+        env:
+          MEGATRON_LM_COMMIT: ${{ needs.pre-flight.outputs.megatron_lm_commit }}
+        run: |
+          megatron_lm_dir="3rdparty/Megatron-Bridge-workspace/Megatron-Bridge/3rdparty/Megatron-LM"
+          bash .github/scripts/checkout-megatron-lm-commit.sh "$megatron_lm_dir" "$MEGATRON_LM_COMMIT"
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@d4b2f3b6ecc6e67c4457f6d3e41ec42d3d0fcb86 # v5
+        with:
+          version: "0.11.28"
```
kajalj22: Could we use the same version as the dockerfile dynamically /reuse var instead of hardcoding here? It's frequently updated

kajalj22 (approved):

balasaajay (approved):

PR #3652, opened 2026-08-14
fix(vllm): correct MXFP8 rollout quantization scope

seonjinn:
## Summary

- add direct ModelOpt ignore patterns for MXFP8 rollout quantization
- keep `lm_head` and required recipe exclusions in BF16, including when users provide custom quantization overrides
- keep non-expert linear layers in BF16 while quantizing routed expert FC1/FC2
- add sync and async MXFP8 rollout performance recipes for DeepSeek V3 and Nemotron3 Super
- pin Qwen3-30B-A3B MXFP8 rollout to the `flashinfer_trtllm` MoE backend

## Quantization scope

The generated quantization config always adds `lm_head` to `ignored_layers` and ModelOpt `ignore`. User-provided ignore entries are combined with these required exclusions and deduplicated.

The shipped recipes apply model-specific scope:

- Qwen3-30B-A3B and Qwen3-235B: routed experts only
- DeepSeek V3: routed experts in main layers 3-60; attention, router, shared experts, dense prefix layers, and MTP remain in BF16
- Nemotron3 Super: `mixer.experts` only; Mamba, attention, router, shared experts, and latent projections remain in BF16
- Dense Qwen3-32B: MLP remains quantized while attention and `lm_head` remain in BF16

The new DeepSeek V3 64n4g and Nemotron3 Super 32n4g recipes inherit their sync or async BF16 baseline topology.

## Validation

Validated at head `d78681749476a7cac62e03923ca697544f4a7182` on GB200 with vLLM 0.25.1:

- vLLM FP8/ModelOpt MXFP8 tests: 23 passed
- FP8 override merge tests: 7 passed
- MXFP8 recipe and launcher tests: 14 passed
- Ruff, Python compile, Bash syntax, and diff checks: passed

The Qwen3-30B-A3B end-to-end smoke is pending. DeepSeek V3 and Nemotron3 Super currently have config-level validation only; end-to-end runs remain a merge gate.


seonjinn:
**Self-review completed at d78681749476a7cac62e03923ca697544f4a7182.**

Reviewed the complete ModelOpt/vLLM quantization path, model-specific scope, recipe inheritance, launcher topology, custom override behavior, and existing PR comments.

Issues found and resolved during self-review:

- aligned the Qwen3-30B-A3B MXFP8 global batch size with the baseline workload
- preserved the intended GB200 segment topology in all Qwen MXFP8 launchers
- corrected DeepSeek and Nemotron shared-expert test probes to target quantizable child linears
- normalized whitespace in quantization_ignore_patterns
- preserved generated lm_head and recipe exclusions when users provide a custom hf_overrides.quantization_config
- treated a null custom quantization config as absent and rejected other non-mapping values

Final-head validation on GB200 with vLLM 0.25.1:

- vLLM FP8/ModelOpt MXFP8 tests: **23 passed**
- FP8 override merge tests: **7 passed**
- MXFP8 recipe and launcher tests: **14 passed**
- Ruff, Python compile, Bash syntax, and diff checks: passed

No unresolved blocking code issue was found in the final diff.

Open merge gates:

- Qwen3-30B-A3B end-to-end smoke is pending
- DeepSeek V3 and Nemotron3 Super have config-level validation but no end-to-end run yet

seonjinn:
/ok to test [24539c6](https://github.com/NVIDIA-NeMo/RL/pull/3652/commits/24539c6af4f02790441149dac30fbdf208a5b145)

tests/test_suites/llm/performance/grpo-deepseek-v3-64n4g-mxfp8-rollout.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
+source $SCRIPT_DIR/common.env
+# disable NVLS to avoid OOM issue
+export NCCL_NVLS_ENABLE=0
+
+# Use the DeepSeek-V3 checkpoint converted to BF16.
+if [[ -z "$NRL_DEEPSEEK_V3_BF16_CKPT" ]]; then
```
guyueh1: **2 action items, please fix in this PR.**

This is PR-introduced in this new DeepSeek MXFP8 launcher. Since [`common.env` enables `set -u`](https://github.com/NVIDIA-NeMo/RL/blob/854fc7ed018214fd554d43efb9c232cdd9c8a1d9/tests/test_suites/llm/performance/common.env#L5), this guard exits with `NRL_DEEPSEEK_V3_BF16_CKPT: unbound variable` when the checkpoint path is unset, so users never see the intended help text.

### AI-1
Make this guard nounset-safe:

```suggestion
if [[ -z "${NRL_DEEPSEEK_V3_BF16_CKPT:-}" ]]; then
```

### AI-2
Quote the later checkpoint-path overrides in this script: `policy.model_name="$NRL_DEEPSEEK_V3_BF16_CKPT"` and `policy.tokenizer.name="$NRL_DEEPSEEK_V3_BF16_CKPT"`.

guyueh1 (commented):
Found one small launcher issue in the new DeepSeek MXFP8 scripts.

tests/test_suites/llm/performance/grpo-deepseek-v3-64n4g-async-1off-mxfp8-rollout.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
+source $SCRIPT_DIR/common.env
+# disable NVLS to avoid OOM issue
+export NCCL_NVLS_ENABLE=0
+
+# Use the DeepSeek-V3 checkpoint converted to BF16.
+if [[ -z "$NRL_DEEPSEEK_V3_BF16_CKPT" ]]; then
```
guyueh1: **2 action items, please fix in this PR.**

This is PR-introduced in this new DeepSeek MXFP8 launcher. Since [`common.env` enables `set -u`](https://github.com/NVIDIA-NeMo/RL/blob/854fc7ed018214fd554d43efb9c232cdd9c8a1d9/tests/test_suites/llm/performance/common.env#L5), this guard exits with `NRL_DEEPSEEK_V3_BF16_CKPT: unbound variable` when the checkpoint path is unset, so users never see the intended help text.

### AI-1
Make this guard nounset-safe:

```suggestion
if [[ -z "${NRL_DEEPSEEK_V3_BF16_CKPT:-}" ]]; then
```

### AI-2
Quote the later checkpoint-path overrides in this script: `policy.model_name="$NRL_DEEPSEEK_V3_BF16_CKPT"` and `policy.tokenizer.name="$NRL_DEEPSEEK_V3_BF16_CKPT"`.

tests/test_suites/llm/performance/grpo-deepseek-v3-64n4g-mxfp8-rollout.sh
seonjinn: Fixed in 19d91a1730940f9a1d5537348c80344315e6636e. The guard is nounset-safe, both checkpoint overrides are quoted, and the launcher regression test covers an unset variable and a path containing spaces.

tests/test_suites/llm/performance/grpo-deepseek-v3-64n4g-async-1off-mxfp8-rollout.sh
seonjinn: Fixed in 19d91a1730940f9a1d5537348c80344315e6636e. The guard is nounset-safe, both checkpoint overrides are quoted, and the launcher regression test covers an unset variable and a path containing spaces.

seonjinn:
Contributor self-review refresh for `19d91a1730940f9a1d5537348c80344315e6636e` on current main.

The expert-only MXFP8 quantization scope and ignore-pattern behavior were rechecked, with no new source-level findings. Ruff/format/config checks pass locally; GPU quantization behavior remains covered by the requested CI tier.

nemo_rl/models/generation/vllm/config.py
```diff
@@ -45,6 +45,11 @@ class VllmSpecificArgs(TypedDict):
     cap_max_tokens_to_context: NotRequired[bool]
     # Use ModelOpt MXFP8 quantization when precision is fp8.
     is_mx: NotRequired[bool]
+    # Legacy substring keywords expanded through Hugging Face module names.
+    quantization_ignored_layer_kws: NotRequired[list[str]]
```
guyueh1, line 70: it would be good to mark this flag as deprecated for this release (0.8) and officially deprecate it after 0.8 is out, since it is overlapped with the new quantization_ignore_patterns flag

nemo_rl/models/generation/vllm/patches.py
```diff
@@ -582,6 +588,69 @@ def ensure_vllm_source_compat() -> None:
     _patch_vllm_radio_layerscale_loader(patch_logger)
 
 
+def _mxfp8_speculative_post_init(self: Any) -> Any:
```
guyueh1: i am not sure if it's ok to globally patch it; we can add "mtp.*","language_model.mtp.*"  to the ignore list and achieve the same effect right?

guyueh1: i think let user adding mtp to the ignore list is a better way

seonjinn: Oh, okay. It's okay to let user add mtp in ignored list.

nemo_rl/models/generation/vllm/config.py
seonjinn: Okay, sounds good.

seonjinn: Addressed in f477201c8. The legacy key remains supported in 0.8, now emits a deprecation warning, and the FP8 documentation directs new configurations to quantization_ignore_patterns.

nemo_rl/models/generation/vllm/patches.py
seonjinn: Addressed in f477201c8. Removed the global SpeculativeConfig patch and its patch-specific tests. DeepSeek and Nemotron MTP recipes now explicitly exclude mtp.* and language_model.mtp.*, and the documentation states that external draft precision is configured separately.

seonjinn:
Contributor self-review refresh for `d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7` after addressing the latest review.

Reviewed the changed MXFP8 config, runtime warning, recipe scope, vLLM patch removal, tests, documentation, and the merge with current main.

- The legacy `quantization_ignored_layer_kws` key remains compatible in 0.8 and now emits a deprecation warning.
- The global `SpeculativeConfig` patch and its patch-specific tests were removed.
- DeepSeek and Nemotron MTP recipes explicitly exclude `mtp.*` and `language_model.mtp.*`.
- External draft precision remains a separate vLLM configuration concern.
- No stale speculative-patch references remain.

Validation:

- MXFP8 recipe and launcher tests: **15 passed**
- Ruff check and format check: passed
- Diff whitespace check: passed
- vLLM-dependent unit coverage is queued in the PR CI environment because vLLM is not installed in the local macOS worktree.

No new blocking issue was found in this review delta.

guyueh1:
LGTM

guyueh1 (dismissed):

seonjinn:
@terrykong could you do final review for this PR?

docs/fp8.md
```diff
@@ -57,6 +57,32 @@ FP8 generations are recommended to be configured with the following settings:
...
+
+`lm_head` is always excluded from FP8 and MXFP8 quantization, even when it is
+not listed in `quantization_ignore_patterns` in the YAML configuration.
+Models with MTP layers must list their MTP module names explicitly, for example
+`mtp.*` and `language_model.mtp.*`. External speculative-decoding drafts use a
+separate vLLM model configuration and must configure their precision separately.
+
+`quantization_ignored_layer_kws` is deprecated in NeMo RL 0.8. Existing
+configurations continue to work in this release, but new configurations should
+use `quantization_ignore_patterns`.
```
terrykong: [`docs/fp8.md:82-84`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/docs/fp8.md#L82-L84)

**1 action item.**

**TL;DR — the deprecation tells users to move to `quantization_ignore_patterns`, but that field raises for non-MX FP8, and both remaining configs using the old key are non-MX — so nobody who sees this notice can act on it.**

PR-introduced (the deprecation text is new here).

The notice is unqualified. But [`fp8.py:218-220`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/nemo_rl/models/generation/vllm/quantization/fp8.py#L218-L220) rejects the replacement whenever `is_mx` is falsy:

```python
if not is_mx:
    raise ValueError("quantization_ignore_patterns requires is_mx=True")
```

Both in-repo users of the deprecated key are plain FP8 with no `is_mx` anywhere in their `defaults:` chain:
- [`grpo-deepseek-v3-64n8g-fp8-async-1off.yaml:19`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/examples/configs/recipes/llm/performance/grpo-deepseek-v3-64n8g-fp8-async-1off.yaml#L19)
- [`grpo-moonlight-16ba3b-4n8g-megatron-fp8-e2e.yaml:43`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/examples/configs/recipes/llm/grpo-moonlight-16ba3b-4n8g-megatron-fp8-e2e.yaml#L43)

Following the doc on either one:

```diff
-      quantization_ignored_layer_kws: [q_proj, k_proj, v_proj, o_proj]
+      quantization_ignore_patterns: [model.layers.*.self_attn.*]
```

gives `ValueError: quantization_ignore_patterns requires is_mx=True`, raised in the rollout worker after the allocation is up.

To be clear: **the `is_mx` restriction itself is right** — non-MX `Fp8Config` reads only `ignored_layers`/`modules_to_not_convert` and matches exactly, never by wildcard ([vllm fp8.py#L156-L173](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/quantization/fp8.py#L156-L173)), so patterns there would silently do nothing. Only the docs need to say so.

(Worth noting the runtime `DeprecationWarning` at [`fp8.py:303`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/nemo_rl/models/generation/vllm/quantization/fp8.py#L303) won't reach users either — Python ignores that category outside `__main__`. That matches the existing repo idiom at [`vllm_worker_async.py:187`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/nemo_rl/models/generation/vllm/vllm_worker_async.py#L187) so it's not something to change here, but it does mean this doc paragraph is the only notice users actually get.)

### AI-1

```suggestion
`quantization_ignored_layer_kws` is deprecated for MXFP8 in NeMo RL 0.8; new
MXFP8 configurations should use `quantization_ignore_patterns` instead.
`quantization_ignore_patterns` requires `is_mx: true` — non-MX FP8 (`precision:
"fp8"` without `is_mx`) has no pattern-based equivalent yet and must keep using
`quantization_ignored_layer_kws`.
```

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -277,8 +298,15 @@ def init_fp8(vllm_cfg, model_name, model_parallel_size):
             bf16_params.extend(_get_params_in_layers(param_names, layers))
 
         fp8_block_quant_kwargs["ignored_layers"] = bf16_params
-    quantization_ignored_layer_kws = vllm_cfg.get("quantization_ignored_layer_kws", [])
-    if len(quantization_ignored_layer_kws):
+    quantization_ignored_layer_kws = vllm_cfg.get("quantization_ignored_layer_kws")
+    if "quantization_ignored_layer_kws" in vllm_cfg:
```
terrykong, line 304: [`nemo_rl/models/generation/vllm/quantization/fp8.py:302`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/nemo_rl/models/generation/vllm/quantization/fp8.py#L302)

**1 action item.**

**TL;DR — the repo's only end-to-end MXFP8 test still passes the deprecated key, so `quantization_ignore_patterns` has no real-hardware coverage at all and every L1 GB200 run trips this new warning.**

The target file is not in this diff, so this is pre-existing — but it's in scope because this PR is what deprecates the key.

[`tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh:48`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh#L48) still passes:

```
'++policy.generation.vllm_cfg.quantization_ignored_layer_kws=[q_proj,k_proj,v_proj,o_proj]'
```

It's wired in at [`L1_Functional_Tests_GB200_MXFP8.sh:37`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/functional/L1_Functional_Tests_GB200_MXFP8.sh#L37). Two consequences:

1. Every other test of the new field mocks `AutoConfig`/`AutoModel`, and the four new perf recipes are `performance_gb200.txt` entries that don't run in PR CI — so `quantization_ignore_patterns` currently has **zero** coverage on real hardware.
2. That run now emits this deprecation warning every time.

The good news: that script already sets `++policy.generation.vllm_cfg.is_mx=true` at [L42](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh#L42), so it passes the `is_mx` guard and can migrate — unlike the two YAML recipes in the `docs/fp8.md` comment.

### AI-1

Not a `suggestion` block: the edit is in a file outside this diff, and it needs a second coordinated change.

```diff
-    '++policy.generation.vllm_cfg.quantization_ignored_layer_kws=[q_proj,k_proj,v_proj,o_proj]' \
+    '++policy.generation.vllm_cfg.quantization_ignore_patterns=[model.layers.*.self_attn.*]' \
```

Please also re-anchor the `assert_grep 'ignored_layers'` at [L67](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh#L67) — it currently depends on the `print()` at [`fp8.py:327`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/nemo_rl/models/generation/vllm/quantization/fp8.py#L327), which only runs on the legacy branch. Grepping for `ignore` or a pattern substring would be more robust. I could not verify from source whether vLLM echoes `hf_overrides` into the run log such that `ignored_layers` would still appear, so please confirm that on hardware rather than assuming it survives the flip.

tests/test_mxfp8_rollout_recipes.py
```diff
@@ -194,3 +320,15 @@ def test_qwen3_235b_mxfp8_recipes_keep_baseline_runtime_knobs() -> None:
         assert "max_num_steps" not in grpo_config
         assert "val_batch_size" not in grpo_config
         assert "max_val_samples" not in grpo_config
+
+
+def test_deepseek_mxfp8_launchers_handle_unset_and_spaced_checkpoint_paths() -> None:
```
terrykong, line 336: [`tests/test_mxfp8_rollout_recipes.py:325`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/test_mxfp8_rollout_recipes.py#L325)

**1 action item.**

**TL;DR — nothing in CI ever runs this file, so the ~180 lines of recipe checks added here (including this regression test for @guyueh1's `set -u` thread) never execute.**

Pre-existing placement — the file was already at `tests/` root — but raised here because this PR adds the only automated check that the MXFP8 recipes carry the right patterns, and that check is inert.

Why it never runs:
1. CI discovers unit tests only via `find tests/unit -maxdepth 1 -name 'L0_Unit*.sh'` ([cicd-main.yml#L792](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/.github/workflows/cicd-main.yml#L792)).
2. Every one of those shards invokes [`tests/run_unit.sh`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/run_unit.sh#L36-L42) with a `unit/`-prefixed path — [`L0_Unit_Tests_Other.sh:30`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/unit/L0_Unit_Tests_Other.sh#L30) passes `"unit/"`, `Vllm_*.sh` passes `"unit/models/generation/test_vllm*.py"`, `Models_*.sh` passes `"unit/models/"` — so `run_unit.sh` forwards a `unit/`-scoped path in every case. Nothing under `tests/*.py` is ever handed to pytest.
3. `testpaths = ["tests"]` in [`pyproject.toml`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/pyproject.toml#L651) doesn't rescue it: `testpaths` applies only when pytest gets no path argument, and no lane invokes bare `pytest`. There is also no `tests/conftest.py`.

Independent confirmation: on `main` this same file contains a hard failure — `test_qwen3_235b_async_tp4_baseline_overlay` opens `grpo-qwen3-235b-32n4g-async-1off-tp4.yaml`, which has never existed on `main` — yet `main` is green. A file with a failing test cannot be running. (This PR deletes that test, which is correct on its own terms; mentioning it only as evidence.)

Note the same applies to the sibling [`tests/test_mxfp8_flashinfer_padding.py`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/test_mxfp8_flashinfer_padding.py).

### AI-1

Moving the file under `tests/unit/` is dependency-safe — it imports only `pathlib`, `pytest`, and `yaml`, no torch/vLLM/ray, and passes 15/15 when run directly. Multi-file/rename, so not a `suggestion` block. One option — of several — if you'd rather not move it, add it to [`L0_Unit_Tests_Other.sh:30`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/unit/L0_Unit_Tests_Other.sh#L30); since `$1` is already `unit/`, `run_unit.sh` forwards the remaining args verbatim and a bare filename resolves against cwd `tests/`:

```
uv run --no-sync bash -x ./tests/run_unit.sh "unit/" "test_mxfp8_rollout_recipes.py" "${IGNORE[@]}" "${EXCLUDED_UNIT_TESTS[@]}" --cov=nemo_rl --cov-report=term-missing --cov-report=json --hf-gated
```

terrykong (commented):
Review by a team of 6 agents (RL/codebase, bug, test, design, comment, devil's advocate). 3 inline comments below, all actionable; a few optional notes here.

**The change is worth making.** The old `quantization_ignored_layer_kws` could only express HF *checkpoint* names by substring, and needed `AutoModel.from_config` to enumerate them. The new patterns are model-load-free and can name vLLM-side **fused** modules that no checkpoint name contains — Nemotron's `mixer.qkv_proj` ([nemotron_h.py#L454](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/models/nemotron_h.py#L454)) is fused in vLLM but separate `q/k/v_proj` on disk, so the old mechanism could not have expressed the correct scope for the two new models at all.

**We independently verified the risky part.** The Nemotron-3-Super and DeepSeek-V3 pattern lists were checked against the real checkpoint index and vLLM v0.25.1 module construction, not just against each other. Every pattern maps to a real construction-time prefix, and the lists are complete — no quantizable linear outside `mixer.experts` is missed. Worth noting the non-obvious part: the checkpoint uses a `backbone.` prefix, and vLLM remaps it via `WeightsMapper(orig_to_new_prefix={"backbone": "model"})` ([nemotron_h.py#L791](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/models/nemotron_h.py#L791)), so `model.layers.*.mixer.*` is correct even though the checkpoint says `backbone.*`. Also confirmed `NemotronHMLP` is unfused with no `gate_proj`, so `up_proj`/`down_proj` is right.

Optional, take or leave:

- **Numerics worth one datapoint.** The Qwen3 MoE router moves from quantized to BF16 on the 30B-A3B and 235B recipes — the old keyword list didn't match `mlp.gate`, the new `model.layers.*.mlp.gate` does. That's a real numerics change rather than a rename, so a `train/token_mult_prob_error` (median + step 10) and throughput datapoint alongside the pending GB200 runs would be good to have. The two dense Qwen3-32B recipes are unaffected. (Separately: adding `train_global_batch_size: 2048` is a genuine fix — that recipe inherits `grpo_math_1B.yaml` at 512 while its BF16 baseline is 2048, so the comparison was previously apples-to-oranges.)
- **Rebase.** The branch conflicts with `main` in `fp8.py` and `test_vllm_fp8_quantization.py`. `main` picked up #3477, which moved the `AutoConfig.from_pretrained` call and inserted an `is_mx requires precision='fp8'` check at exactly the line where this PR adds `trust_remote_code=True`. That argument is required by Nemotron-3-Super's `auto_map`, so please make sure it survives the resolution — `test_init_fp8_loads_remote_model_config` will catch it if not.
- **One-line comment worth adding.** The generated config emits both `ignore` and `ignored_layers`, and exactly one is live per backend: ModelOpt/MXFP8 reads `ignore` ([modelopt.py#L316](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/quantization/modelopt.py#L316)), non-MX `Fp8Config` reads `ignored_layers`. Three reviewers had to reverse-engineer that dispatch; a one-line comment near [`fp8.py:337`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/nemo_rl/models/generation/vllm/quantization/fp8.py#L337) would save the next reader.
- **`moe_backend` on dense models.** `Qwen/Qwen3-32B` has no MoE layers, so `moe_backend: flashinfer_trtllm` on the two 32B recipes is never consulted. Harmless, but it's forced by the now-unconditional assertion at [`test_mxfp8_rollout_recipes.py:231-234`](https://github.com/NVIDIA-NeMo/RL/blob/d0c9e14a9ab0195cbc2a7caae7d3248ec32ea2f7/tests/test_mxfp8_rollout_recipes.py#L231-L234) — restoring the `if expected["moe_backend"] is not None:` guard would let those two recipes drop it.

_Generated by Claude Code_

docs/fp8.md
seonjinn: Addressed in bdf686c81: the notice is now qualified — "deprecated **for MXFP8**" — and explicitly states that non-MX FP8 (`precision: "fp8"` without `is_mx`) has no pattern-based replacement yet and must keep `quantization_ignored_layer_kws`, so the two non-MX recipes you listed are no longer told to migrate to a field that would reject them.

nemo_rl/models/generation/vllm/quantization/fp8.py
seonjinn: Addressed in bdf686c81: `grpo_vllm_mxfp8_rollout_gb200.sh` now passes `quantization_ignore_patterns=[model.layers.*.self_attn.*]` instead of the deprecated key, so the GB200 L1 run exercises the new field on real hardware and no longer trips the deprecation warning. A regression test (`test_mxfp8_functional_test_uses_ignore_patterns`) pins that the script never reverts to the old key.

tests/test_mxfp8_rollout_recipes.py
seonjinn: Addressed in bdf686c81: `L0_Unit_Tests_Other.sh` now passes `test_mxfp8_rollout_recipes.py` explicitly alongside `unit/`. Verified against `run_unit.sh`'s dispatch: since $1 starts with `unit/`, the remaining args are forwarded verbatim to pytest, which resolves them relative to `tests/` — so the recipe checks (including the `set -u` regression test from the earlier thread) now execute in the Other shard.

guyueh1 (dismissed):

seonjinn:
/ok to test [033a926](https://github.com/NVIDIA-NeMo/RL/pull/3652/commits/033a9262fd79b46b02cd36509635bbf7e90a99b2)

tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh
```diff
@@ -64,6 +64,6 @@ uv run tests/check_metrics.py "$JSON_METRICS" \
     'max(data["train/token_mult_prob_error"]) < 2.0'
 
 assert_grep 'quantization=modelopt_mxfp8|quant_algo.*MXFP8' "$RUN_LOG"
-assert_grep 'ignored_layers' "$RUN_LOG"
+assert_grep 'model\.layers\.\*\.self_attn\.\*' "$RUN_LOG"
```
terrykong: [`tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh:67`](https://github.com/NVIDIA-NeMo/RL/blob/033a9262fd79b46b02cd36509635bbf7e90a99b2/tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh#L67)

**1 action item.** Thanks for the fixes — the docs change and the CI lane fix both check out, and picking the lane option over `git mv` was the right call (a bare move would have broken `PROJECT_ROOT = parents[1]`). This is the one loose end.

**TL;DR — this grep passes because the script echoes its own CLI argument into the log, not because the patterns reached ModelOpt, so it would stay green even if `init_fp8` dropped the field entirely.**

`RUN_LOG` is `2>&1 | tee` ([L57](https://github.com/NVIDIA-NeMo/RL/blob/033a9262fd79b46b02cd36509635bbf7e90a99b2/tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh#L57)), and `run_grpo.py` prints the overrides and the resolved config at startup — [`Overrides: {...}`](https://github.com/NVIDIA-NeMo/RL/blob/033a9262fd79b46b02cd36509635bbf7e90a99b2/examples/run_grpo.py#L92) and [`pprint.pprint(config)`](https://github.com/NVIDIA-NeMo/RL/blob/033a9262fd79b46b02cd36509635bbf7e90a99b2/examples/run_grpo.py#L101). Both contain `model.layers.*.self_attn.*` verbatim, so the assert is satisfied before vLLM is even constructed.

The old `assert_grep 'ignored_layers'` was stronger: it matched [`print("ignored_layers", ...)`](https://github.com/NVIDIA-NeMo/RL/blob/033a9262fd79b46b02cd36509635bbf7e90a99b2/nemo_rl/models/generation/vllm/quantization/fp8.py#L329), which runs *inside* `init_fp8` after the list is computed. That print only executes on the legacy branch, so flipping the key removed the only outcome signal.

I checked whether vLLM would supply one instead, and it won't: `VllmConfig.__str__` prints `quantization_config=` from [`ModelConfig.quantization_config`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/config/model.py#L208), the user-facing field that defaults to `None` and is never assigned from `hf_config` — `hf_overrides` lands on `model_arch_config.quantization_config` instead. And `modelopt.py` never logs `exclude_modules`.

To be clear about scope: the run itself is genuine coverage now, and [L66](https://github.com/NVIDIA-NeMo/RL/blob/033a9262fd79b46b02cd36509635bbf7e90a99b2/tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh#L66) still confirms MXFP8 is active. Only this one assert is weaker than what it replaced.

### AI-1

Not a `suggestion` block — the fix spans two files. Mirror the legacy print onto the patterns path at [`fp8.py:334`](https://github.com/NVIDIA-NeMo/RL/blob/033a9262fd79b46b02cd36509635bbf7e90a99b2/nemo_rl/models/generation/vllm/quantization/fp8.py#L334):

```python
    if quantization_ignore_patterns:
        fp8_block_quant_kwargs.setdefault("ignore", []).extend(
            quantization_ignore_patterns
        )
        print("ignore", fp8_block_quant_kwargs["ignore"])
```

then anchor the assert on that emitted line rather than on the echoed config, e.g. `assert_grep '^ignore .*self_attn'`.

Also worth noting before merge: the GB200 L1 job hasn't actually run on this PR — only `Build GB200/GCP container` and `gb200-config` show up in the checks — so the flip is verified by reading, not by execution. Worth one real run given this is now the only end-to-end MXFP8 coverage.

terrykong (commented):
Follow-up on my earlier review at `d0c9e14` — re-verified all three comments against `033a9262f`.

Two are fully resolved:
- **`docs/fp8.md`** — the notice is correctly scoped to MXFP8 and states non-MX FP8 must keep `quantization_ignored_layer_kws`. Matches the guard at `fp8.py:220-222`.
- **CI collection** — adding `test_mxfp8_rollout_recipes.py` to `L0_Unit_Tests_Other.sh` works; I ran it as the lane would and got 16 passed. Choosing this over `git mv` also avoided a `PROJECT_ROOT` breakage the move would have caused.

The rebase also resolved cleanly: both `trust_remote_code=True` and #3477's `is_mx=True requires precision='fp8'` guard survived, which was the specific collision I flagged.

One comment inline on the remaining item — the functional-test assert is satisfied by the script's own echoed argument, so it no longer proves the patterns reach ModelOpt. Not a blocker.

Still open from the optional notes, take or leave: no throughput/convergence datapoint for the MoE-router QUANTIZED→BF16 change, and `moe_backend` remains on the two dense Qwen3-32B recipes.

_Generated by Claude Code_

tests/functional/grpo_vllm_mxfp8_rollout_gb200.sh
seonjinn: Addressed in a9ad4cb1c. Rather than mirroring the legacy print inside `init_fp8`, the marker is emitted one level up in `vllm_worker._log_effective_quantization_ignore_patterns` (called right after the hf_overrides merge), so it reports the *effective* ignore list — generated patterns plus any user-supplied `quantization_config.ignore` entries — not just the raw CLI argument. `grpo_vllm_mxfp8_rollout_gb200.sh` now anchors on `assert_grep 'NRL_MXFP8_EFFECTIVE_IGNORE=.*self_attn'`, which only appears after the merged config is computed, so it can no longer be satisfied by the startup config echo. `test_vllm_fp8_hf_overrides.py` covers the emission and `test_mxfp8_rollout_recipes.py` pins the script-side assert. On the real-run note: the branch is now merged with latest main (ef773ad8c) and CI re-triggered; the GB200 L1 lane still needs the run label if you want the end-to-end pass before merge.

terrykong (approved):

PR #3653, opened 2026-08-14
fix(vllm): detect if quantize is mxfp8 from worker config

guyueh1:
# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


seonjinn:
Self-review completed on head `cb633584`.

- Reviewed the changes in `nemo_rl/models/generation/vllm/quantization/fp8.py` and `tests/unit/models/generation/test_vllm_fp8_quantization.py`.
- Confirmed that MXFP8 refit detection now uses the resolved vLLM quantization config from the worker instead of relying on the checkpoint config.
- Checked the vLLM 0.25.1 API contract: `VllmConfig.quant_config` and `ModelOptMxFp8Config` are the correct interfaces for this detection.
- The added tests cover both MXFP8 and non-MXFP8 worker configurations.
- No blocking correctness, compatibility, security, or deadlock issues were found.
- CI, including lint and the vLLM unit-test shards, is passing.

The remaining non-code item is to replace the default PR description template with a short root-cause, change, and test summary.

seonjinn:
@guyueh1  LGTM!

seonjinn (dismissed):

guyueh1:
@arnavk-nvidia can you review?

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -336,6 +342,72 @@ def is_fp8_model(vllm_config):
...
 
+def is_mxfp8_model(vllm_config):
+    try:
+        from vllm.model_executor.layers.quantization.modelopt import (
+            ModelOptMxFp8Config,
+        )
+    except ImportError:
+        return False
+
+    return isinstance(getattr(vllm_config, "quant_config", None), ModelOptMxFp8Config)
```
seonjinn: **Medium — dead code**: `is_mxfp8_model` has zero callers repo-wide (only the definition matches a full grep) and no test exercises it. Given the PR title, this reads like the intended detection entry point whose call site got dropped from the diff. Suggest either wiring it into a real caller (e.g. as a cross-check of `vllm_config.quant_config` against the declared `is_mx` inside `_get_refit_fp8_config`) or removing it, so future contributors don't adopt an unvalidated helper.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -336,6 +342,72 @@ def is_fp8_model(vllm_config):
...
+        "The process-local FP8 configuration is unavailable in this vLLM "
+        "refit worker, and the worker's vLLM additional_config does not carry "
+        f"{NEMO_RL_FP8_CONFIG_KEY}."
+    )
+
+
+def _get_refit_is_mx(model_runner):
+    if global_fp8_config_is_mx_checked and global_fp8_config is not None:
+        return global_fp8_config.is_mx
+    return _get_refit_fp8_config(model_runner).is_mx
```
seonjinn: **Medium — cache never takes effect in production**: `_get_refit_is_mx` is the only reader of `global_fp8_config_is_mx_checked`, but its sole caller is a unit test (`test_refit_prefers_matching_process_local_fp8_config`). The production path, `load_weights` (line 540), calls `_get_refit_fp8_config` directly, which re-derives and re-validates the worker config on every refit call regardless of the flag. So the "cache validated refit MXFP8 mode" commit writes the flag (lines 390/395) but nothing on the real path ever reads it.

Not a correctness bug (`_get_refit_fp8_config` stays correct on every call), but the caching described by the commit is inert. Either have `load_weights` use `_get_refit_is_mx` for the branch decision, or drop this helper (and `is_mxfp8_model`) if the fast path isn't needed — the blockwise branch needs the full config anyway.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -336,6 +342,72 @@ def is_fp8_model(vllm_config):
...
+            ModelOptMxFp8Config,
+        )
+    except ImportError:
+        return False
+
+    return isinstance(getattr(vllm_config, "quant_config", None), ModelOptMxFp8Config)
+
+
+def fp8_config_from_vllm_config(vllm_config) -> FP8Config | None:
+    additional_config = getattr(vllm_config, "additional_config", None)
```
seonjinn, line 389: **Low — untested branch**: the `getattr(vllm_config, "additional_config", None)` default path (a `vllm_config` object with no `additional_config` attribute at all) has no direct unit test; the existing parametrized cases cover `FP8Config`, dict, non-dict, and bad-type values. Trivial behavior, but this function is central to the new detection path — a one-line parametrize addition would pin it.

seonjinn (commented):
Agent-team self-review (2 domain reviewers + adversarial verification pass; every cited line re-verified against the PR head afdca8ac4).

**Summary**: the core detection logic is sound. `_get_refit_fp8_config`'s three-way resolution (process-local config, worker `additional_config` bootstrap, `is_mx` mismatch rejection) is internally consistent and matches all 5 new unit tests, and the `additional_config` transport reuses the already-proven `nemo_rl_checkpoint_engine` pattern. The new tests are hermetic and follow the file's existing mocking conventions. The byte-wise tensor comparison (`.view(torch.uint8)` + `torch.equal`) was adversarially checked and is a correct workaround for the missing CPU `float8_e4m3fn` comparison kernel, not test-weakening.

**Main issue (Medium)**: two helpers introduced by this PR are never called from production code, and as a result the mode caching this PR adds is behaviorally inert on the real refit path (details inline).

No Critical or High issues found.

Non-blocking heads-up: #3294 also rewrites the `load_weights` region of `fp8.py`; whichever PR lands second will need a semantic rebase around the config-resolution plumbing.

seonjinn (dismissed):
Approving — detailed review in my earlier comment: core worker-config detection logic and tests are sound, no correctness issues. The two Medium items (unused is_mxfp8_model / _get_refit_is_mx helpers) are cleanup, non-blocking.

arnavk-nvidia (approved):

guyueh1:
i think this PR and https://github.com/NVIDIA-NeMo/RL/pull/3754 fixed the same thing (but in different ways), i'm ok with closing this if https://github.com/NVIDIA-NeMo/RL/pull/3754 merges

guyueh1:
i think this fix won't be necessary because https://github.com/NVIDIA-NeMo/RL/pull/3754 fixes the root cause, closing

PR #3654, opened 2026-08-14
feat(megatron): support te precision config recipes

guyueh1:
# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


guyueh1:
@arnavk-nvidia can you approve?

arnavk-nvidia (dismissed):
LGTM

guyueh1:
@terrykong for review, CI is passing and got 1 approval from @arnavk-nvidia

guyueh1:
@terrykong please final review

seonjinn (commented):
Reviewed the delta (3 files, +42) after the latest main merge. Verified against the pinned Megatron-LM (14346b65a via Megatron-Bridge 8c46dc42): `load_quantization_recipe` exists there with an stdlib-only import chain, `TransformerConfig.quant_recipe` is a real field at that pin, and the recipe is attached before `finalize()`/model construction, so ordering is correct. Both new unit tests pin real behavior (the `SimpleNamespace` negative test correctly avoids MagicMock attribute auto-vivification). No correctness issues found.

Two suggestions, both non-blocking:

1. **Document (or validate) the `fp8_cfg` + `te_precision_config_file` interaction.** `_apply_precision_config` sets `quant_recipe` while `_apply_performance_config` later sets `model_cfg.fp8`/`fp8_recipe`/`fp8_param` from `fp8_cfg`. Upstream mcore resolves these per-module via `get_quant_config_or_none(name, config.quant_recipe)`, so for recipe-matched modules the recipe silently wins over `fp8_cfg` intent. Other dual-knob cases in `setup.py` get explicit conflict validation (`_validate_model_override_conflicts`); a one-line warning or docstring note when both are set would save users from thinking they compose additively.

2. **Add a minimal example recipe or wrap the file-missing error.** A missing/bad path currently surfaces from `RecipeConfig.from_yaml_file` deep inside megatron-core with no NeMo-RL-level context, and the repo ships no example YAML documenting the format. A short docstring example or an existence check with a NeMo-RL-specific error would help discoverability.

nemo_rl/models/megatron/setup.py
```diff
@@ -1074,6 +1074,19 @@ def _apply_precision_config(
 
     te_precision_config_file = config["megatron_cfg"].get("te_precision_config_file")
     if te_precision_config_file is not None:
+        if not os.path.isfile(te_precision_config_file):
```
seonjinn: I've included warning message to let user know the issue. @guyueh1 does this make sense to you ?

seonjinn (dismissed):
LGTM

nemo_rl/models/megatron/setup.py
guyueh1: looks good

nemo_rl/models/megatron/setup.py
```diff
@@ -1071,6 +1072,25 @@ def _apply_precision_config(
...
+        if fp8_cfg is not None and fp8_cfg.get("enabled", False):
+            warnings.warn(
+                "Both megatron_cfg.fp8_cfg and megatron_cfg.te_precision_config_file "
+                "are set; modules matched by the precision recipe use the recipe's "
+                "per-module quantization config instead of fp8_cfg.",
+                stacklevel=2,
+            )
+        # NeMo-RL constructs TransformerConfig directly and therefore bypasses
+        # Megatron-LM's CLI path that normally turns this file into quant_recipe.
+        model_cfg.quant_recipe = load_quantization_recipe(te_precision_config_file)
```
terrykong: **1 action item.**

**TL;DR — a recipe file with `fp8_param: true` gives the model real FP8 weights, but every FP8 check in NeMo-RL reads `fp8_cfg` only, so the refit ships FP8 bytes into BF16 vLLM params with no error.**

PR-introduced. [`setup.py:1092`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/models/megatron/setup.py#L1092)

**Background.** The TE precision recipe has its **own** [`fp8_param`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/extensions/transformer_engine.py#L167) field. It is not the same field as `megatron_cfg.fp8_cfg.fp8_param`. Upstream describes it as: *"If cast the initialized parameters to fp8 precision and all-gather weights in FP8."*

**How it reaches the parameters** — this happens at module construction, not in `forward`:
1. `TELinear.__init__` resolves the recipe for this module: [`get_quant_config_or_none(name, config.quant_recipe)` then `finish_init(...)`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/extensions/transformer_engine.py#L1241-L1242)
2. it builds a context from the result: [`_get_fp8_model_init_for_quant_params(...)`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/extensions/transformer_engine.py#L1243)
3. which is [`fp8_model_init(enabled=qrecipe.fp8_param, ...)`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/extensions/transformer_engine.py#L262)
4. and [`super().__init__(...)` runs **inside** that context](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/extensions/transformer_engine.py#L1254) — so the weights are allocated as FP8.

**What breaks.** Every FP8 branch in NeMo-RL keys off `fp8_cfg`, which is unset in a recipe-only run:
- [`_is_fp8_export()`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2285) returns `False`, so [`_build_refit_conversion_tasks`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2294) uses `get_conversion_tasks` instead of `build_export_fp8_tasks` — no `*_scale_inv` companion tensors.
- [`check_nccl_reshard_refit_support`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/weight_sync/nccl_reshard_utils.py#L674) reads `fp8_cfg.get("fp8_param")`. Its `elif fp8_param:` guard exists precisely to reject FP8 training storage that has no FP8 generation consumer — and it never fires.

**Breaking case.** `te_precision.yaml`:
```yaml
configs:
  default:
    transformer_engine_config_type: TEQuantizationParams
    training_recipe: {fp8_quantization_recipe: blockwise, fp8_param: true}
matchers:
  all: {config: default, type: glob, pattern: "*", enabled: true}
```
Set `policy.megatron_cfg.te_precision_config_file` to it and leave `fp8_cfg` out. Training allocates FP8 weights, and the refit then treats them as BF16. I could not run this — no GPU on my side — so I can't tell you whether it fails loudly in the copy or silently produces wrong weights. Either way the guard meant to catch exactly this never fires, which is the part worth fixing.

Megatron-LM carries the same open gap right next to the line this mirrors: [`# TODO(kwyss): Prohibit fp8_params or fp4_params with this flexibility`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/training/argument_utils.py#L367). NeMo-RL has more at stake than upstream, because the refit byte-copies training weights straight into the inference engine.

One more thing worth knowing: even if NeMo-RL learned to read the recipe's `fp8_param`, [`_is_fp8_export()`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2285) only handles `blockwise`. A recipe selecting `tensorwise`, `mxfp8`, or `nvfp4` has no export path at all.

**Action:** reject `fp8_param` / `fp4_param` in the loaded recipe. Only configs reachable from an enabled matcher are checked, since Megatron already drops `enabled: false` matchers at parse time. One way to do it — of several:

```suggestion
        # NeMo-RL constructs TransformerConfig directly and therefore bypasses
        # Megatron-LM's CLI path that normally turns this file into quant_recipe.
        quant_recipe = load_quantization_recipe(te_precision_config_file)
        # A recipe can store primary weights in FP8/FP4 through its own
        # fp8_param/fp4_param fields, which are separate from fp8_cfg.fp8_param.
        # NeMo-RL derives sequence padding, refit export, and reshard validation
        # from fp8_cfg alone, so such weights would reach the inference engine as
        # if they were BF16. Reject until the refit path understands them.
        for config_key in sorted({m.config_key for m in quant_recipe.matchers}):
            payload = quant_recipe.configs.get(config_key) or {}
            for block in ("training_recipe", "evaluation_recipe"):
                block_cfg = payload.get(block) or {}
                if block_cfg.get("fp8_param") or block_cfg.get("fp4_param"):
                    raise ValueError(
                        "megatron_cfg.te_precision_config_file sets fp8_param or "
                        f"fp4_param in '{config_key}.{block}'. NeMo-RL reads "
                        "megatron_cfg.fp8_cfg for all FP8 behavior, so these "
                        "weights would be sent to the inference engine as BF16. "
                        "Use megatron_cfg.fp8_cfg for FP8 parameter storage."
                    )
        model_cfg.quant_recipe = quant_recipe
```

Worth a test each way: a recipe with `fp8_param: true` under an enabled matcher raises; the same config behind a disabled matcher does not.

nemo_rl/models/megatron/setup.py
```diff
@@ -1071,6 +1072,25 @@ def _apply_precision_config(
     }
     model_cfg.pipeline_dtype = dtype_map[config["megatron_cfg"]["pipeline_dtype"]]
 
+    te_precision_config_file = config["megatron_cfg"].get("te_precision_config_file")
```
terrykong, line 1162: **1 action item.**

**TL;DR — sequence padding is computed from `fp8_cfg.fp8_recipe`, but matched modules run the *recipe's* recipe, so a mixed setup can pad to 32 where blockwise FP8 needs 128.**

Adjacent rather than something this PR broke — but this PR is what makes the mismatch reachable. [`setup.py:1075`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/models/megatron/setup.py#L1075)

[`_get_fp8_token_alignment`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/models/megatron/data.py#L162) returns the padding multiple from `fp8_cfg["fp8_recipe"]` alone: 128 for `blockwise`, 32 for `mxfp8`, 16 otherwise. The comment at [data.py:1367-1374](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/models/megatron/data.py#L1367) calls 128 a hard requirement — *"blockwise FP8 : 128 (cublas block size)"*.

A precision recipe overrides the recipe **per module**, so the two can disagree:

```yaml
policy:
  megatron_cfg:
    fp8_cfg: {enabled: true, fp8: hybrid, fp8_recipe: mxfp8, fp8_param: false}
    te_precision_config_file: /path/te_precision.yaml   # matcher selects blockwise
```

1. `_get_fp8_token_alignment` sees `fp8_recipe: mxfp8` and returns **32**.
2. The matched layers run `Float8BlockScaling` from the recipe, which wants **128**.
3. Sequences get padded to a multiple of 32 and reach a blockwise GEMM unaligned.

This is not an exotic setup — a global recipe plus per-module overrides is the pattern Megatron-Bridge's own DeepSeek-V4 recipes use.

**Action:** not a suggestion block — the fix belongs in `nemo_rl/models/megatron/data.py`, which is outside this diff. Either make `_get_fp8_token_alignment` take the max over the recipe's matched configs, or (cheaper, and fine for this PR) write the limitation down in the docs section requested on the config key: token alignment comes from `fp8_cfg` only, so if the recipe selects a stricter one, set `fp8_cfg.fp8_recipe` to match or raise `make_sequence_length_divisible_by` yourself.

nemo_rl/models/policy/__init__.py
```diff
@@ -481,6 +481,8 @@ class MegatronConfig(TypedDict):
     clear_memory_caches_before_refit: NotRequired[bool]
     # FP8 quantization settings for the Megatron training backend.
     fp8_cfg: NotRequired[Fp8Config]
+    # Per-module Transformer Engine precision recipe loaded into Megatron quant_recipe.
+    te_precision_config_file: NotRequired[str]
```
terrykong, line 484: **1 action item.**

Picking up @seonjinn's earlier suggestion — the error-wrapping half landed, the discoverability half didn't. [`policy/__init__.py:485`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/nemo_rl/models/policy/__init__.py#L485)

`te_precision_config_file` appears in exactly three places in the repo: this TypedDict, `setup.py`, and the new tests. No example recipe file, and nothing in `docs/`. A user can't discover the option, and — more to the point — can't learn what the YAML it points at should contain.

Two behaviors are worth writing down because neither is guessable and both fail quietly:

1. **Every matcher needs `enabled: true`.** [`_build_matchers`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/quantization/quant_config.py#L148) does `enabled = matcher.get("enabled", False)` and skips the rest. Omit it and you get a recipe that parses fine and matches nothing.
2. **A recipe on its own does not change the forward pass.** [`_get_fp8_autocast_for_quant_recipe`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/extensions/transformer_engine.py#L312-L318) returns `nullcontext()` when the global FP8 autocast is off and [`override_nonquantized_autocast`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/extensions/transformer_engine.py#L157) is `False` — which is its default. So `te_precision_config_file` without `fp8_cfg.enabled: true` is a no-op for compute. (This does *not* apply to `fp8_param`, which still quantizes the weights — that's the separate issue above.)

For a working shape to copy, Megatron-Bridge's [`_deepseek_v4_mxfp8_quant_recipe()`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/recipes/deepseek/h100/deepseek_v4.py#L38-L58) is a good reference: MXFP8 for training, an empty `evaluation_recipe` to force BF16 in eval, one `*` matcher.

**Action:** not a suggestion block — the change belongs in [`docs/fp8.md`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/docs/fp8.md), which isn't in this diff. Add a short section after the existing `fp8_cfg` block covering: the config key, one example recipe YAML, the two gotchas above, and the note that token alignment still comes from `fp8_cfg`. Filling the same example into the PR description would help too — it's still the untouched template.

tests/unit/models/megatron/test_megatron_setup.py
```diff
@@ -1003,6 +1003,84 @@ def test_pipeline_dtype_mapping(self):
...
+                "pipeline_dtype": "bfloat16",
+                "te_precision_config_file": str(recipe_file),
+                "fp8_cfg": {"enabled": True},
+            }
+        }
+
+        with pytest.warns(UserWarning, match="fp8_cfg"):
+            _apply_precision_config(model_cfg, config, torch.bfloat16)
+
+        mock_load_recipe.assert_called_once_with(str(recipe_file))
```
terrykong, line 1197: **1 action item.**

All four new tests patch `load_quantization_recipe`, so nothing checks that a real recipe file actually parses. That's mostly fine — the parser is upstream's — but one cheap real-file test would pin the `enabled: true` gotcha, which is the easiest way to write a recipe that silently does nothing.

[`test_megatron_setup.py:1082`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/tests/unit/models/megatron/test_megatron_setup.py#L1082)

**Action:** add one unmocked test. Checked against the parser at [`quant_config.py`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/quantization/quant_config.py#L140-L151) — a matcher without `enabled: true` is dropped, so `matchers` comes back empty.

```suggestion
        mock_load_recipe.assert_called_once_with(str(recipe_file))

    def test_te_precision_config_parses_real_recipe_file(self, tmp_path):
        """A real recipe file parses, and matchers need an explicit enabled flag."""
        from megatron.core.quantization.utils import load_quantization_recipe

        recipe_file = tmp_path / "te_precision.yaml"
        recipe_file.write_text(
            "configs:\n"
            "  mxfp8:\n"
            "    transformer_engine_config_type: TEQuantizationParams\n"
            "    training_recipe: {fp8_quantization_recipe: mxfp8}\n"
            "    evaluation_recipe: {}\n"
            "matchers:\n"
            "  all: {config: mxfp8, type: glob, pattern: '*', enabled: true}\n"
        )
        recipe = load_quantization_recipe(str(recipe_file))
        assert len(recipe.matchers) == 1
        assert "mxfp8" in recipe.configs

        # A matcher without `enabled: true` is dropped, so nothing matches.
        disabled_file = tmp_path / "disabled.yaml"
        disabled_file.write_text(
            "configs:\n"
            "  mxfp8:\n"
            "    transformer_engine_config_type: TEQuantizationParams\n"
            "    training_recipe: {fp8_quantization_recipe: mxfp8}\n"
            "matchers:\n"
            "  all: {config: mxfp8, type: glob, pattern: '*'}\n"
        )
        assert load_quantization_recipe(str(disabled_file)).matchers == []
```

Heads up: I could not run this locally — `uv` can't build the project on my machine because the 3rdparty submodules aren't checked out. Please confirm it passes in the mcore lane before taking it.

tests/unit/models/megatron/test_megatron_setup.py
```diff
@@ -1003,6 +1003,84 @@ def test_pipeline_dtype_mapping(self):
...
+        }
+
+        _apply_precision_config(model_cfg, config, torch.bfloat16)
+
+        mock_load_recipe.assert_called_once_with(str(recipe_file))
+        assert model_cfg.quant_recipe is recipe
+
+    @patch("nemo_rl.models.megatron.setup.load_quantization_recipe")
+    def test_skips_te_precision_config_when_not_configured(self, mock_load_recipe):
+        """Missing recipe path leaves quant_recipe unset."""
```
terrykong: **No action needed if you disagree — this is a one-word nit.**

The docstring says "Missing recipe path", which reads like the missing-*file* case tested a few lines below. This test is about the config key being absent entirely.

[`test_megatron_setup.py:1032`](https://github.com/NVIDIA-NeMo/RL/blob/d2ba7c03dff311e9043c5d6a7ec4caa3309f5ac3/tests/unit/models/megatron/test_megatron_setup.py#L1032)

```suggestion
        """An unset config key leaves quant_recipe untouched."""
```

seonjinn (dismissed):

terrykong (commented):
Deep review by a team of 7 agents (RL expert, Megatron/Bridge expert, bug finder, test reviewer, design reviewer, comment reviewer, devil's advocate).

**Overall: the core change is right.** It mirrors Megatron-LM's own `--te-precision-config-file` wiring one-for-one, including the easy-to-get-wrong detail of setting `quant_recipe` **without** `use_kitchen` (unlike the kitchen branch beside it). Ordering is correct: the recipe is attached before `finalize()` / `__post_init__()` and before the model is built. Failing fast on a missing file with a NeMo-RL-level error naming the config key is a nice touch, and the comment explaining *why* the block exists will save the next reader real time. Verified against Megatron-LM `14346b65a2d0790e451919858f7771078105c5f0` (reached via the Megatron-Bridge pin `8c46dc42`).

One finding is worth acting on before merge (the `fp8_param` / refit gap); the rest are docs and tests.

Things the team checked and found clean, so nobody needs to re-litigate them: the new import pulls in no optional dependency (stdlib plus a guarded `yaml`), `os` and `warnings` are already imported, `quant_recipe` is a real `TransformerConfig` dataclass field, the tests carry the right `mcore` mark and land in the right L0 lane, and `_validate_model_override_conflicts` picks up the new key for free.

Note the repo linter did not run on my machine (`uv` can't build the project without the 3rdparty submodules checked out); `ruff check` and `ruff format --check` were run directly and are clean on all three files.

_Generated by Claude Code_

nemo_rl/models/megatron/setup.py
guyueh1: this is a corner case that even Megatron-LM itself does not guard; MLM doesn't intend to support when global fp8 recipe is A but locally some layers are recipe B, it only supports mixing different precisions (NVFP4, MXFP8, BF16); 

instead of the suggested change, I am just going to add a check to ban this case (global fp8 recipe != layerwise fp8 recipe)

guyueh1: added the check in [a87a274](https://github.com/NVIDIA-NeMo/RL/pull/3654/commits/a87a2748b4d999169cd00a9fd51634cb8e39ad66)

nemo_rl/models/policy/__init__.py
guyueh1: i added explanation in the doc and also added an L1 test in [a87a274](https://github.com/NVIDIA-NeMo/RL/pull/3654/commits/a87a2748b4d999169cd00a9fd51634cb8e39ad66)

guyueh1: but i think maybe doc is enough

terrykong (approved):

PR #3656, opened 2026-08-14
docs: document token_level_loss in the GRPO guide

### LINKED ISSUE #1274: Add token_level_loss in grpo guide
**Is your feature request related to a problem? Please describe.**
Request for documentation in grpo guide.

**Describe the solution you'd like**
Add a section on token_level_loss in grpo guide. 

**Additional context**
When customer asks about this parameter, it would be useful to have docs.

terrykong:
## What

Closes #1274.

`token_level_loss` had no explanation anywhere in the docs — it appeared only as a bare line in config blocks (`dapo.md`, `cispo.md`, `ppo.md`, `prorlv2.md`) and in ~30 example YAMLs. This adds a **Loss Normalization** section to `docs/guides/grpo.md`, placed under `## Loss` before the "Improvements to the GRPO Loss Formulation" subsections — it is not an optional stability add-on like those, it is a property of the baseline loss that the existing $E_t[\cdot]$ notation leaves unspecified.

## Contents of the new section

- **What it controls.** The reduction applied to the per-token loss, not the per-token quantity itself. The same switch also normalizes the KL penalty, and both denominators are global-batch counts rather than per-microbatch.
- **The two reductions**, written out once, plus a worked example: a 10-token and a 1000-token response give a 99%/1% split under `true` and 50%/50% under `false`, so short-response tokens carry ~100× the per-token weight under sequence-level.
- **Objectives table.** Full objective for GRPO/PPO/DAPO, CISPO, GSPO and REINFORCE/RLOO under both settings. The guide never wrote these objectives out anywhere, so the table does double duty. CISPO and GSPO each have one cell marked unavailable — those combinations are rejected by assertions in `ClippedPGLossFn.__init__`.
- **Losses without this parameter.** `token_level_loss` is a field of `ClippedPGLossConfig` only; SFT, DPO/RM, distillation, the PPO critic, cross-tokenizer distillation and Eagle3 draft training all fix `loss_type` as a class attribute. Setting it in an SFT config does nothing, which is worth saying out loud.
- **Metric normalization.** Which reported metrics follow the flag (`loss`, `kl_penalty`), which are always token-normalized, which key off a *different* parameter (`sampling_importance_ratio`, `is_oob_ratio`), and which are not normalized at all. Relevant when comparing runs across the setting.
- **Choosing a setting.** DAPO §3.3's argument for token-level in long-CoT, quoted rather than paraphrased, and explicitly framed as one reported result rather than a rule — the original GRPO paper formulates the loss at the sequence level, and this repo has no benchmark comparing the two.

## Also in this PR

- **Docstring fix** (`loss_functions.py:206`): the REINFORCE/RLOO objective was written as $E_t[\pi_\theta(a_t|s_t) \cdot A_t]$, but `:541` sets `ratios = curr_logprobs` — the raw log-probability, not `exp()`. The new table is derived from the code, so leaving the docstring disagreeing with it would have been confusing. Corrected to $\log \pi_\theta$.
- **Cross-links** from `dapo.md:86` and `cispo.md:56` to the new section. Those are the two pages where a reader meets this parameter first and currently gets no explanation — the same gap the issue reports.

No changes to `ppo.md` (it already defers wholesale to the GRPO guide) or `docs/index.md` (no new file; `grpo.md` is already in the toctree).

## Verification

The docs build could not be run locally — `uv.lock` only supports Linux and this is macOS — so the MyST-level checks were done directly:

- Heading anchors confirmed against `mdit_py_plugins.anchors.slugify`: `Loss Normalization (\`token_level_loss\`)` → `loss-normalization-token_level_loss`, matching the links from `dapo.md` and `cispo.md`. `myst_heading_anchors = 5` in `conf.py:68` covers the `###`/`####` levels used.
- Parsed the file with `markdown-it-py` + `dollarmath`: all 3 tables parse with the intended cell counts, and all inline/display math tokenizes. Table cells use `\lvert`/`\rvert` rather than bare `|`, which would otherwise split the rows.
- Every formula in the objectives table was read back off the implementation rather than from the papers — GSPO's $s_i$ against `:528-534`, CISPO against `:545`, the clipped surrogate against `:547-551`, REINFORCE against `:541-542`.
- Every row of the "losses without this parameter" table was checked against its actual consumer (`sft.py:229`, `dpo.py:284`, `rm.py:263`, `distillation.py:641`, `ppo.py:377`, `xtoken_off_policy_distillation.py:479`, `wrapper.py:251`).
- The DAPO quotes were pulled from the paper, not recalled.
- `ruff format --check` and `ruff check` pass on the modified Python file.

The permalink to `metric_normalizations` is SHA-pinned rather than `blob/main`, since a line-anchored link to `main` breaks silently as code shifts.

### DIFF
--- a/docs/guides/grpo.md
+++ b/docs/guides/grpo.md
@@ -504,7 +504,7 @@ The table below covers the metrics reported by `ClippedPGLossFn` (the GRPO/PPO f
 | **Keyed on a different parameter** | `sampling_importance_ratio` follows `sequence_level_importance_ratios`; `is_oob_ratio` is sequence-normalized only for `truncated_importance_sampling_type: seq-mask-tis` |
 | **Not normalized** — raw counts, per-microbatch means, or extrema combined with min/max | `num_valid_samples`, `positive_nll_loss`, `probs_ratio_min`, `probs_ratio_max`, `probs_ratio_clamped_min`, `probs_ratio_clamped_max` |
 
-The authoritative list is the [`metric_normalizations` table in `ClippedPGLossFn`](https://github.com/NVIDIA-NeMo/RL/blob/e98357616c0bb8fb13140841e225583b5cc36b0e/nemo_rl/algorithms/loss/loss_functions.py#L335-L371), which is kept in sync with the metrics the loss returns.
+The authoritative list is the [`metric_normalizations` table in `ClippedPGLossFn`](https://github.com/NVIDIA-NeMo/RL/blob/4a1454bf430624786251d14ba0197169c8e68a5c/nemo_rl/algorithms/loss/loss_functions.py#L335-L371), which is kept in sync with the metrics the loss returns.
 
 #### Choosing a setting
 
--- a/docs/guides/grpo.md
+++ b/docs/guides/grpo.md
@@ -432,19 +432,19 @@ where:
 
 ### Loss Normalization (`token_level_loss`)
 
-The formulas above are written with a bare expectation $E_t[\cdot]$, which leaves one detail unspecified: the loss is computed per token, but the optimizer needs a single number, so those per-token values have to be averaged. The `token_level_loss` parameter chooses how that average is taken. It does not change the per-token quantity itself — only how the per-token values are combined into the scalar that is differentiated.
+The formulas above are written with a bare expectation $E_t[\cdot]$, which leaves one detail unspecified: the loss is computed per token, but the optimizer needs a single number, so those per-token values have to be averaged. The `token_level_loss` parameter chooses how that average is taken. It does not change the per-token objective itself — only how the per-token values are combined into the scalar that is differentiated.
 
 There are two choices:
 
 - **`token_level_loss: true`** — average over every token in the batch. Every token counts the same, so a response contributes to the update in proportion to how many tokens it has.
 - **`token_level_loss: false`** — average within each response first, then average those per-response values. Every *response* counts the same regardless of length.
 
-Writing $\ell_{i,t}$ for the per-token quantity of token $t$ in response $o_i$, $|o_i|$ for the number of unmasked tokens in that response, and $N$ for the number of valid responses:
+Writing $\ell_{i,t}$ for the per-token objective of token $t$ in response $o_i$ — the quantity inside the expectation in the formula above — $|o_i|$ for the number of unmasked tokens in that response, and $N$ for the number of valid responses:
 
 $$
-L_{\text{token}}(\theta) = -\frac{1}{\sum_{i=1}^{N} |o_i|} \sum_{i=1}^{N} \sum_{t=1}^{|o_i|} \ell_{i,t}
+L_{\text{token}}(\theta) = \frac{1}{\sum_{i=1}^{N} |o_i|} \sum_{i=1}^{N} \sum_{t=1}^{|o_i|} \ell_{i,t}
 \qquad\qquad
-L_{\text{seq}}(\theta) = -\frac{1}{N} \sum_{i=1}^{N} \frac{1}{|o_i|} \sum_{t=1}^{|o_i|} \ell_{i,t}
+L_{\text{seq}}(\theta) = \frac{1}{N} \sum_{i=1}^{N} \frac{1}{|o_i|} \sum_{t=1}^{|o_i|} \ell_{i,t}
 $$
 
 Both denominators — $\sum_i |o_i|$ and $N$ — are counts over the whole global batch, not over a single microbatch. Each microbatch divides by the global count and contributes a partial sum, so the microbatch losses can simply be summed to recover the global-batch loss. See [Loss Functions](../design-docs/loss-functions.md) for how that normalization factor is computed and passed in.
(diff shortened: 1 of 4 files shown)


docs/guides/grpo.md
```diff
@@ -430,6 +429,90 @@ where:
...
+### Loss Normalization (`token_level_loss`)
+
+The formulas above are written with a bare expectation $E_t[\cdot]$, which leaves one detail unspecified: the loss is computed per token, but the optimizer needs a single number, so those per-token values have to be averaged. The `token_level_loss` parameter chooses how that average is taken. It does not change the per-token quantity itself — only how the per-token values are combined into the scalar that is differentiated.
+
+There are two choices:
+
+- **`token_level_loss: true`** — average over every token in the batch. Every token counts the same, so a response contributes to the update in proportion to how many tokens it has.
+- **`token_level_loss: false`** — average within each response first, then average those per-response values. Every *response* counts the same regardless of length.
+
+Writing $\ell_{i,t}$ for the per-token quantity of token $t$ in response $o_i$, $|o_i|$ for the number of unmasked tokens in that response, and $N$ for the number of valid responses:
```
yfw:
```suggestion
Writing $\ell_{i,t}$ for the per-token loss of token $t$ in response $o_i$, $|o_i|$ for the number of unmasked tokens in that response, and $N$ for the number of valid responses:
```

docs/guides/grpo.md
```diff
@@ -430,6 +429,90 @@ where:
...
+
+That is arithmetic, not a recommendation — which weighting is preferable depends on the algorithm and the workload. See [Choosing a setting](#choosing-a-setting) below.
+
+#### Objectives under each setting
+
+`token_level_loss` lives on `ClippedPGLossFn`, the loss shared by the policy-gradient algorithms. The table below writes out the full objective for each one under both settings. To keep the expressions readable they show the base objective only; optional terms — the KL penalty, dual clipping (`ratio_clip_c`), and the importance-sampling correction (`use_importance_sampling_correction`) — are applied on top of it and are documented in the sections that follow.
+
+Throughout, $r_{i,t} = \frac{\pi_\theta(o_{i,t})}{\pi_{\theta_{\text{old}}}(o_{i,t})}$ is the per-token probability ratio, $A_i$ is the advantage for response $i$, $\text{sg}(\cdot)$ is the stop-gradient operator, and $\varepsilon_{\text{low}}$/$\varepsilon_{\text{high}}$ are `ratio_clip_min`/`ratio_clip_max`.
```
yfw: That second \varepsilon doesn't seem to be rendering correctly, maybe:

```suggestion
Throughout, $r_{i,t} = \frac{\pi_\theta(o_{i,t})}{\pi_{\theta_{\text{old}}}(o_{i,t})}$ is the per-token probability ratio, $A_i$ is the advantage for response $i$, $\text{sg}(\cdot)$ is the stop-gradient operator, and $\varepsilon_{\text{low}}$, $\varepsilon_{\text{high}}$ are `ratio_clip_min`/`ratio_clip_max`.
```

terrykong: Confirmed and fixed in 8a6f95024 — good catch, and the cause is worth recording.

It is not a LaTeX problem; it is GitHub's inline-math parser. `$a$/$b$` closes the first span but then treats `/$b$` as literal text. Rendering the section through GitHub's own markdown API before and after:

```
# before:  and <math-renderer …>$\varepsilon_{\text{low}}$</math-renderer>/$\varepsilon_{\text{high}}$ are …
# after:   and <math-renderer …>$\varepsilon_{\text{low}}$</math-renderer>, <math-renderer …>$\varepsilon_{\text{high}}$</math-renderer> are …
```

Notably this renders **fine** under MyST/`dollarmath`, which is what actually builds the docs site — both forms parse to two math spans there. So the published page was never broken; only the GitHub view of the file was. Since these guides do get read directly on github.com, the comma is the right call regardless.

I swept the whole section for the same adjacency hazard: GitHub now emits 25 inline + 1 display math spans, exactly matching the MyST parse, with zero leftover `$` characters. This was the only instance.

docs/guides/grpo.md
terrykong: Applied the intent in 8a6f95024, with one deviation worth explaining — your suggestion surfaced a real bug.

I took $\ell_{i,t}$ as the **per-token objective** rather than the per-token loss, because as written the reductions had a leading minus:

$$L_{\text{token}}(\theta) = -\frac{1}{\sum_i |o_i|} \sum_i \sum_t \ell_{i,t}$$

so calling $\ell_{i,t}$ a loss would have read as "the loss is the negative sum of the losses."

Chasing that down, the minus was mine and it was wrong: the $L(\theta)$ formula ~25 lines earlier in this same section is written **without** one (`L = E[min(...)] - beta*D_KL`, the quantity being maximized). My new formulas silently flipped the sign convention against the ones directly above them. I have dropped the minus from both reductions and from all six formula cells in the objectives table, so the whole section now uses one convention.

Net effect of your comment: "quantity" -> "objective", plus a sign fix I would not have caught otherwise. If you would rather the section define $\ell_{i,t}$ as the per-token loss and carry the minus explicitly — which would match `clip_loss` in the code more literally — say the word and I will flip it that way instead.

yfw (approved):

PR #3659, opened 2026-08-15
fix(vllm): support BF16 TRTLLM NCCL reshard refit

seonjinn:
## What this PR does

Makes weight refit work for **BF16 MoE models running on the FlashInfer TRTLLM backend** when using `refit_transport=nccl_reshard`.

Without this fix, a refit silently corrupts the MoE expert weights and the model generates garbage. With this fix, expert weights load correctly, and any configuration we cannot handle is rejected with a clear error instead of corrupting anything.

#3545 already fixed this for the `ipc` and `collective` transports. This PR extends the fix to `nccl_reshard`, which needs a different approach.

## The problem in one picture

```
What the trainer sends:            What vLLM actually stores (TRTLLM):
┌──────────────────────┐           ┌──────────────────────────────┐
│ expert weights in    │  ──X──>   │ same weights, repacked into  │
│ checkpoint layout    │           │ a kernel-private layout      │
└──────────────────────┘           └──────────────────────────────┘
```

`nccl_reshard` streams every received tensor **directly into vLLM's live parameter buffers**. That is fast, and it is correct as long as the buffer stores the weight in the same layout as the checkpoint.

The FlashInfer TRTLLM MoE backend breaks that assumption: at model load time, vLLM **repacks** the grouped expert weights (`gate`/`up`/`down` projections) into a private layout its kernels need. Streaming checkpoint-layout bytes into a repacked buffer writes the right numbers into the wrong places — no error, just a corrupted model.

## Why we can't just reuse the #3545 fix

#3545 fixes `ipc`/`collective` by reloading the **whole model** through vLLM's normal weight loader (which knows how to repack). But `nccl_reshard` receives most weights directly into live buffers. A whole-model reload would reset *all* buffers back to checkpoint layout — including the dense/attention buffers that direct receive depends on and that were never broken. The fix must be surgical: touch only the expert weights.

## How the fix works

Split the weights into two groups and treat them differently:

| Weight kind | Path | Why |
|---|---|---|
| Dense, attention, embeddings | Direct receive into live buffers (unchanged) | Their layout matches the checkpoint — always was correct |
| Grouped MoE expert weights | Staging tensor → vLLM's own weight loader | Only vLLM knows how to repack them |

For the expert weights, concretely:

1. **Receive into a staging tensor, not the live buffer.** Each rank allocates a fresh BF16 tensor shaped as its expert-parallel slice of the global `[num_experts, ...]` weight.
2. **Replay through vLLM's loader.** After the transfer, each expert in the staging tensor is handed to vLLM's normal weight loader under its global name (`experts.<expert_start + local_idx>.<proj>.weight`). vLLM repacks it exactly like it does at model load.
3. **Reload lifecycle covers only the MoE modules.** Every other buffer stays live, so the direct-receive fast path keeps working for the rest of the model. One `finalize()` at the end rebuilds the kernel layout.

## When we refuse to refit (on purpose)

Two layouts cannot be mapped to "one named load per expert", so we reject them loudly at setup instead of guessing:

- **Expert weights sharded on a non-expert dim** (e.g. a TP slice through the middle of one expert's matrix) → `ValueError` when the refit map is built.
- **Non-linear expert placement** (`expert_placement_strategy != "linear"`, e.g. round-robin) → `RuntimeError`, because "global expert ID = `expert_start + local_idx`" would be wrong.

And if anything fails *after* the expert buffers have been invalidated mid-refit, the worker is marked permanently unusable rather than serving a half-updated model.

## What is NOT affected

Everything else is untouched: other MoE backends, other transports, quantized (FP8/MXFP8) paths, dense models.

## vLLM version compatibility

The backend detection touches three vLLM internals, verified byte-identical at v0.25.1 (our current pin) and v0.28.0: `UnquantizedMoeBackend.FLASHINFER_TRTLLM`, `UnquantizedFusedMoEMethod.unquantized_backend`, and `ExpertMapManager.placement_strategy`. If a future vLLM moves them, the guarded imports degrade to "backend not detected" and the placement check raises — there is no silent-corruption path.

## How it was validated

- **Unit tests (CPU):** staging shape/dtype, global expert-ID naming, reload lifecycle order, and both rejected layouts.
- **End to end:** 1-step Qwen3-30B-A3B BF16 GRPO on 16 GB200 — 54.00 GiB FFN reshard completed, `Policy generation refit completed successfully`, `moe_backend=flashinfer_trtllm`.
- `pre-commit` passes.

## Dependency

#3545 has merged and this branch is synced with latest `main`, so the diff here is only the `nccl_reshard` extension.


seonjinn:
Self-review completed against the latest `main` branch.

- Checked backend gating, expert-local shard selection, global expert IDs, staging dtype, native reload lifecycle, transport failure behavior, and unchanged dense/attention paths.
- Verified the vLLM native reload API through targeted unit tests and an end-to-end BF16 TRTLLM refit run.
- Found one missing negative test during review: tensor-sharded expert destinations were rejected in code but not covered. Added `test_build_hf_to_local_param_map_rejects_trtllm_tensor_sharding`.
- Validation: 52 targeted vLLM tests passed, 8 xferdtensor tests passed, and pre-commit passed on all changed files.

No unresolved correctness findings remain in the NCCL reshard extension. The PR remains draft because it includes and depends on #3545.

seonjinn:
Latest-main end-to-end validation completed at `28445421d502a31cc31975affafbc629be0745a1`.

- Hardware: 16 GB200 GPUs (4 nodes x 4 GPUs)
- Model: Qwen3-30B-A3B
- Configuration: BF16 training, BF16 rollout, `moe_backend=flashinfer_trtllm`, `refit_transport=nccl_reshard`
- Scope: one full async GRPO step
- Result: completed successfully with exit code 0
- Transfer split: 54.00 GiB through NCCL reshard and 2.87 GiB through the misc broadcast path (95.0% through reshard)
- Runtime confirmation: `Policy generation refit completed successfully` on all generation ranks

This verifies the implementation on the current `main`-based branch, including generation startup, refit, rollout collection, logprob/advantage processing, and the training step.

seonjinn:
Contributor self-review refresh for `c119bdb7ab1c7a34c5bea950c44548dae94d7038` after the latest main merge.

The BF16 TRTLLM NCCL-reshard lifecycle, dtype handling, and worker placement paths were rechecked against current main. No new source-level findings were identified. Ruff/format/config checks pass locally; Linux vLLM and distributed coverage remains delegated to CI.

nemo_rl/weight_sync/xferdtensor.py
youngeunkwon0405: Changes in xferdtensor.py might not be allowed. If this is for the case of unclean sharding, I recommend you to just assert.

tests/unit/models/generation/test_nccl_reshard_backend.py
```diff
@@ -324,6 +337,310 @@ def test_build_hf_to_local_param_map_specs_and_roundtrip():
...
+
+    ext = vllm_backend.VllmInternalWorkerExtension.__new__(
+        vllm_backend.VllmInternalWorkerExtension
+    )
+    ext.nccl_reshard_refit_info = {
+        "layer_names": [],
+        "per_layer_params": {},
+        "misc_meta": {},
+    }
+    ext._receive_and_load_misc_params = MagicMock()
```
youngeunkwon0405, line 915: [`tests/unit/models/generation/test_nccl_reshard_backend.py:619`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/tests/unit/models/generation/test_nccl_reshard_backend.py#L619)

**[TEST]** `test_nccl_reshard_refit_runs_transport_lifecycle` fails deterministically as committed: the extension is built via `__new__` and `pp_comm_groups` is never set, so `_nccl_reshard_refit_impl` hits the unconditional narrowing guard at [`vllm_backend.py:1668-1673`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L1668) (class default is `None`) and raises `RuntimeError` before reaching any assertion. Verified by running with vLLM 0.25.1 on CPU (order-independent); passes with the fix below.

```suggestion
    ext.pp_comm_groups = {}
    ext._receive_and_load_misc_params = MagicMock()
```

tests/unit/models/generation/test_nccl_reshard_backend.py
```diff
@@ -324,6 +337,310 @@ def test_build_hf_to_local_param_map_specs_and_roundtrip():
...
+            self.rank = rank
+
+        def init_nccl_communicator(self, *, device):
+            assert device == torch.device("cpu")
+
+    monkeypatch.setattr(
+        "nemo_rl.distributed.stateless_process_group.StatelessProcessGroup",
+        _FakeGroup,
+    )
+    monkeypatch.setattr(torch.distributed, "get_rank", lambda: 0)
```
youngeunkwon0405, line 816: [`tests/unit/models/generation/test_nccl_reshard_backend.py:526`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/tests/unit/models/generation/test_nccl_reshard_backend.py#L526)

**[TEST]** `test_legacy_refit_map_is_built_after_comm_groups_exist` fails when run standalone: it patches `torch.distributed.get_rank`, but `init_nccl_reshard_comm_group` → `resolve_rollout_rank` ([`checkpoint_engine.py:75-80`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/checkpoint_engine.py#L75)) also calls `torch.distributed.get_world_size()`, which raises `ValueError: Default process group has not been initialized` unless an earlier in-process test happens to leave a default PG initialized (order-dependent at best). Verified failing standalone with vLLM 0.25.1 on CPU; passes with the fix below.

```suggestion
    monkeypatch.setattr(torch.distributed, "get_rank", lambda: 0)
    monkeypatch.setattr(torch.distributed, "get_world_size", lambda: 2)
```

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1137,6 +1203,107 @@ def post(ctx: RefitCtx) -> None:
...
+            )
+            expert_start = 0 if local_slices[0].start is None else local_slices[0].start
+            grouped_proj = param_info["grouped_expert_proj"]
+            expert_prefix = param_info["name"].rsplit(f".{grouped_proj}.weight", 1)[0]
+            fused_param = (
+                "w13_weight"
+                if grouped_proj in ("gate_proj", "up_proj")
+                else "w2_weight"
+            )
+            expected_loaded_name = f"{expert_prefix}.{fused_param}"
```
youngeunkwon0405: [`nemo_rl/models/generation/vllm/vllm_backend.py:1268`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L1268)

**[BUG]** The `expected_loaded_name` check would brick DeepSeek-family models on every refit. DeepSeek V2/V3 hand-roll `load_weights` with the default `routed_experts_prefix="routed_experts"` ([`deepseek_v2.py:1533`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/models/deepseek_v2.py#L1533), mapping built in [`routed_experts.py` `build_expert_params_mapping`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/fused_moe/routed_experts.py#L1026)), so `loaded_params` contains `...experts.routed_experts.w13_weight` — the expected spelling without `.routed_experts.` is absent even when the load fully succeeds. `post` then raises on the first grouped param and the lifecycle marks the worker permanently unusable. (Qwen3-MoE reports the unqualified spelling via `AutoWeightsLoader` → `MoERunner`, which is why the e2e run and the mocked unit test pass.)

Suggested fix: accept either spelling, e.g.
```python
expected_loaded_names = {
    f"{expert_prefix}.{fused_param}",
    f"{expert_prefix}.routed_experts.{fused_param}",
}
# ... in post():
if loaded_names is not None and not (expected_loaded_names & loaded_names):
    raise RuntimeError(...)
```

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -868,6 +920,14 @@ def _weight_update_lifecycle(
             )
 
             model = self.model_runner.model
+            # NCCL reshard receives most weights directly into live parameter
+            # storage; only the TRTLLM MoE modules need the layerwise reload
+            # lifecycle to rebuild the kernel's private repacked layout.
+            reload_targets = (
```
youngeunkwon0405: [`nemo_rl/models/generation/vllm/vllm_backend.py:926`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L926)

**[BUG]** On this new native nccl_reshard branch only the TRTLLM MoE modules are reload-initialized, and vLLM's `finalize_layerwise_reload` skips non-initialized modules ([`reload/layerwise.py`, `not info.can_load()` → reset+continue](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/model_loader/reload/layerwise.py#L238)), while `_maybe_process_fp8_kv_cache()` runs only on the non-layerwise branch. So with `kv_cache_dtype=fp8`, the per-layer k/v scales are never reprocessed after a refit on this path — a silent-staleness gap that is new to nccl_reshard (ipc/collective initialize the whole model, and the pre-PR non-native branch ran model-wide `process_weights_after_loading` plus the fp8-kv hook). The retained comment at [`vllm_backend.py:1706-1710`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L1706) ("dense Linear + attention/MLA ... FP8 KV-cache per-layer k/v scales are finalized by the lifecycle on exit") is no longer accurate on this branch. The targeted Qwen3-30B config is unaffected.

Suggested fix: reject `self._uses_fp8_kv_cache()` in `_validate_native_layerwise_refit("nccl_reshard")` (fail loud like the other unsupported layouts) and update the comment at 1706-1710.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1137,6 +1203,107 @@ def post(ctx: RefitCtx) -> None:
...
+            ep_size = 1
+            for mesh_dim, placement in enumerate(param_info["dst_placements"]):
+                if isinstance(placement, Shard) and placement.dim == 0:
+                    ep_size *= int(mesh_tensor.shape[mesh_dim])
+            num_global_experts = int(param_info["global_shape"][0])
+            # The staged expert range uses torch.chunk semantics while vLLM
+            # places experts with a balanced split; the two agree only when
+            # the expert count divides evenly, so reject the uneven case
+            # instead of silently loading experts onto the wrong ranks.
+            if ep_size > 0 and num_global_experts % ep_size != 0:
```
youngeunkwon0405: [`nemo_rl/models/generation/vllm/vllm_backend.py:1241`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L1241)

**[TEST]** This uneven-expert-split rejection is the only new guard without a unit test — the sibling rejections (tensor-shard dims, missing fused destination, round-robin placement) are all covered. Also, `ep_size > 0` here is dead code: `ep_size` is initialized to 1 and only multiplied by mesh dim sizes ≥ 1.

The following test passes locally (CPU, vLLM 0.25.1) and matches the neighboring conventions in `test_nccl_reshard_backend.py`:

```python
def test_build_hf_to_local_param_map_rejects_uneven_trtllm_expert_split():
    """Expert counts that do not divide evenly across EP ranks are rejected."""
    expert_name = "model.layers.0.mlp.experts.gate_proj.weight"
    refit_info = {
        "gen_tp_size": 2,
        "layer_names": ["model.layers.0"],
        "per_layer_params": {
            "model.layers.0": [
                {
                    "name": expert_name,
                    "global_shape": [3, 32, 16],
                    "dtype": "torch.bfloat16",
                    "grouped_expert_proj": "gate_proj",
                    "dst_mesh_info": MeshInfo(torch.tensor([8, 9])),
                    "dst_placements": [Shard(0)],
                }
            ]
        },
    }
    ext = _make_ext(
        {
            "model.layers.0.mlp.experts.routed_experts.w13_weight": torch.empty(
                128, 16, 24, 64
            ),
        }
    )
    ext._uses_unquantized_flashinfer_trtllm = lambda: True

    with pytest.raises(ValueError, match="divide evenly"):
        ext.build_hf_to_local_param_map(refit_info)
```

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1137,6 +1203,107 @@ def post(ctx: RefitCtx) -> None:
...
+                raise ValueError(
+                    "BF16 FlashInfer TRTLLM nccl_reshard refit requires "
+                    "the global expert count to divide evenly across EP ranks; "
+                    f"got {num_global_experts} experts over {ep_size} ranks for "
+                    f"{param_info['name']!r}"
+                )
+
+            pp_stage = param_info.get("pp_stage", 0)
+            rank = self.pp_comm_groups[pp_stage].rank
```
youngeunkwon0405: [`nemo_rl/models/generation/vllm/vllm_backend.py:1250`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L1250)

**[GUIDELINE]** If `prepare_nccl_reshard_refit_info` ever runs before `init_nccl_reshard_comm_group` on a TRTLLM model, this subscript raises a bare `TypeError: 'NoneType' object is not subscriptable` (`pp_comm_groups` defaults to `None` at the class level). The comment at [`vllm_backend.py:1174-1175`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L1174) documents this ordering dependency, and this same file already narrows the identical condition with a named RuntimeError at [`vllm_backend.py:1668-1673`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L1668) — consider mirroring that guard here so a misordering fails with a named error.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1137,6 +1203,107 @@ def post(ctx: RefitCtx) -> None:
...
+            def post(ctx: RefitCtx) -> None:
+                weights = [
+                    (
+                        f"{expert_prefix}.{expert_start + local_idx}."
+                        f"{grouped_proj}.weight",
+                        expert_weight,
+                    )
+                    for local_idx, expert_weight in enumerate(ctx.buf.unbind(0))
+                ]
+                loaded_names = self._load_full_hf_weights(weights)
```
youngeunkwon0405, line 1453: [`nemo_rl/models/generation/vllm/vllm_backend.py:1292`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/models/generation/vllm/vllm_backend.py#L1292)

**[SUGGESTION, low severity]** The `loaded_names` check only catches the all-experts-fail case: vLLM's per-expert loader returns `False` silently for a non-local expert ([`routed_experts.py` expert weight_loader](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/fused_moe/routed_experts.py#L615)) and the fused name is reported on any single success, so a partial load would pass this check. Every realistic trigger is already fenced (linear placement + even split validated here; EPLB rejected at [`nccl_reshard_utils.py:611-616`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/nemo_rl/weight_sync/nccl_reshard_utils.py#L611); vLLM rejects redundant experts without EPLB), so this is defense-in-depth only: loading per-expert and checking the fused destination after each call would turn any future geometry regression into a loud failure instead of silent staleness.

docs/design-docs/nccl-reshard-refit.md
```diff
@@ -39,6 +39,10 @@ single `ValueError` listing every violation. The current requirements are:
   `vllm_cfg.precision=fp8` and `vllm_cfg.is_mx=true`; the generation ranks
   quantize each received BF16 shard before installing it. Blockwise-FP8 train →
   MXFP8 gen is not supported.
+* BF16 FlashInfer TRTLLM MoE is supported through vLLM's native
```
youngeunkwon0405: [`docs/design-docs/nccl-reshard-refit.md:42`](https://github.com/NVIDIA-NeMo/RL/blob/8a0f5737f44ea60dd64d66ba5fc607fed837ab89/docs/design-docs/nccl-reshard-refit.md#L42)

**[PERF-EVIDENCE, optional]** The validation evidence is solid for a correctness fix (CPU unit tests + 1-step Qwen3-30B-A3B e2e with a 54 GiB reshard — thanks for the detailed writeup). Two optional asks for the record: (a) the nccl_reshard refit wall-time on the MoE model, so the staging+replay overhead of the expert path vs the dense direct-receive path is documented; (b) whether post-refit weights/generations were checked for equivalence against the ipc/collective transport — that cross-transport equivalence is the strongest correctness signal for this class of fix.

nemo_rl/weight_sync/xferdtensor.py
```diff
@@ -210,51 +213,61 @@ def _get_mesh_coords(mesh, rank):
 
 
 def _compute_shard_slices(global_shape, mesh_shape, mesh_coords, placements):
-    """Return the slice of the global tensor this rank owns."""
+    """Return the slice of the global tensor this rank owns.
+
+    Match DTensor's sequential ``torch.chunk`` semantics, including uneven
```
youngeunkwon0405: Changes in xferdtensor.py might not be allowed. If this is for the case of unclean sharding, I recommend you to just assert.

nemo_rl/models/generation/vllm/vllm_backend.py
youngeunkwon0405: Seems like this is a real bug

youngeunkwon0405 (commented):
Could you please also validate this with qwen 3.5 35B-A3B model? Qwen3.5 is using grouped expert params from the hf format. I think it might be safer to validate in both cases.
For the `_trtllm_grouped_expert_spec` implementation, I will just trust you. For the other parts, it looks good to me. Please rollback nemo_rl/weight_sync/xferdtensor.py changes if there are other strong reasons other than the comment. Since the xferdtensor only handles the FFN large weights, it might not hit the uneven sharding scenarios.

seonjinn:
Addressed the review feedback in commit `ddf7495d4`.

- Removed the `xferdtensor.py` shard-calculation API and its duplicate tests.
- Reused the existing exact-transfer shard calculation from `xferdtensor_python` for the TRTLLM refit path, including uneven and repeated shard dimensions.
- Kept the change scoped to the vLLM backend; the general xferdtensor implementation is unchanged.

Validation:
- Python syntax check passes.
- Qwen3.5-35B-A3B BF16 and MXFP8 `flashinfer_trtllm` + `nccl_reshard` smoke jobs were submitted on OCI-HSG; results will be added after the jobs complete.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -868,6 +930,14 @@ def _weight_update_lifecycle(
             )
 
             model = self.model_runner.model
+            # NCCL reshard receives most weights directly into live parameter
+            # storage; only the TRTLLM MoE modules need the layerwise reload
+            # lifecycle to rebuild the kernel's private repacked layout.
+            reload_targets = (
```
sharonyu-115: [`vllm_backend.py:936`](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L936) **[BUG, confidence 90]**

On the `nccl_reshard` + BF16 TRTLLM native-layerwise-reload branch, `_maybe_process_fp8_kv_cache()` is only invoked on the *other* (non-native) branch of `_weight_update_lifecycle` — the native branch returns before reaching it (see [line 962](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L962) vs. the call site at [line 979](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L979)), and `_validate_native_layerwise_refit` ([lines 856-893](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L856)) never rejects an fp8 KV-cache config for this transport.

With `kv_cache_dtype=fp8`, per-layer k/v scales silently go stale after every `nccl_reshard` refit of a TRTLLM model — the same class of silent-corruption failure mode this PR is trying to close, just for KV-cache scales instead of expert weights. Verified independently against current HEAD (not just the diff).

Suggested fix: either call `_maybe_process_fp8_kv_cache()` after the TRTLLM reload finalizes on the native branch too, or reject `_uses_fp8_kv_cache()` explicitly in `_validate_native_layerwise_refit("nccl_reshard")` so misconfiguration fails loudly instead of silently.

(This corresponds to the still-open thread at https://github.com/NVIDIA-NeMo/RL/pull/3659#discussion_r3911120845, which you already called "a real bug".)

tests/unit/models/generation/test_nccl_reshard_backend.py
sharonyu-115: [`test_nccl_reshard_backend.py:619`](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/tests/unit/models/generation/test_nccl_reshard_backend.py#L619) **[TEST, confidence 90]**

`test_nccl_reshard_refit_runs_transport_lifecycle` builds `ext` via `__new__` and never sets `ext.pp_comm_groups`. `_nccl_reshard_refit_impl` unconditionally raises `RuntimeError` when `pp_comm_groups is None` before any assertion runs, so this test errors out as committed rather than passing.

```suggestion
    ext.pp_comm_groups = {}
```

(Corresponds to still-open thread https://github.com/NVIDIA-NeMo/RL/pull/3659#discussion_r3911120811.)

tests/unit/models/generation/test_nccl_reshard_backend.py
sharonyu-115: [`test_nccl_reshard_backend.py:526`](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/tests/unit/models/generation/test_nccl_reshard_backend.py#L526) **[TEST, confidence 88]**

`test_legacy_refit_map_is_built_after_comm_groups_exist` monkeypatches `torch.distributed.get_rank` but not `get_world_size`. `resolve_rollout_rank` ([checkpoint_engine.py:75-80](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/checkpoint_engine.py#L75)) calls the real `get_world_size()`, which raises `ValueError` unless a default process group was already initialized by an earlier test in the same pytest process. `tests/unit/conftest.py` has no autouse fixture that guarantees this — genuine order-dependence.

```suggestion
    monkeypatch.setattr(torch.distributed, "get_world_size", lambda: 2)
```

(Corresponds to still-open thread https://github.com/NVIDIA-NeMo/RL/pull/3659#discussion_r3911120834.)

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1137,6 +1213,107 @@ def post(ctx: RefitCtx) -> None:
...
+            )
+            expert_start = 0 if local_slices[0].start is None else local_slices[0].start
+            grouped_proj = param_info["grouped_expert_proj"]
+            expert_prefix = param_info["name"].rsplit(f".{grouped_proj}.weight", 1)[0]
+            fused_param = (
+                "w13_weight"
+                if grouped_proj in ("gate_proj", "up_proj")
+                else "w2_weight"
+            )
+            expected_loaded_name = f"{expert_prefix}.{fused_param}"
```
sharonyu-115: [`vllm_backend.py:1278`](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L1278) **Replying to the open thread here — resolving as not-a-bug, with evidence.**

The concern was that DeepSeek's default `routed_experts_prefix` would make `expected_loaded_name` never match `loaded_names`, bricking DeepSeek-family refits. Traced the actual pinned vLLM v0.25.1 source: `RoutedExperts.load_weights` builds its match mapping via `get_expert_mapping(include_fused=True)`, which **hardcodes `routed_experts_prefix=""`** for the fused-load path used here (https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/fused_moe/routed_experts.py#L944), overriding the `"routed_experts"` default that only applies to the legacy/hand-rolled `make_expert_params_mapping` entry point (unused on this path). So `AutoWeightsLoader` reports the short fused name (e.g. `...experts.w13_weight`), exactly what this code expects. The `.routed_experts.` segment is real in the physical `nn.Module` tree (hence it appears in this PR's own test fixture's raw *parameter* name), but it is never part of the reported *loaded-weight* name.

Recommend closing this thread — no code change needed here.

(Original thread: https://github.com/NVIDIA-NeMo/RL/pull/3659#discussion_r3911120840.)

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1137,6 +1213,107 @@ def post(ctx: RefitCtx) -> None:
...
+            if ep_size > 0 and num_global_experts % ep_size != 0:
+                raise ValueError(
+                    "BF16 FlashInfer TRTLLM nccl_reshard refit requires "
+                    "the global expert count to divide evenly across EP ranks; "
+                    f"got {num_global_experts} experts over {ep_size} ranks for "
+                    f"{param_info['name']!r}"
+                )
+
+            pp_stage = param_info.get("pp_stage", 0)
+            rank = self.pp_comm_groups[pp_stage].rank
```
sharonyu-115: [`vllm_backend.py:1260`](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L1260) **[GUIDELINE, low severity, confidence 75]**

`rank = self.pp_comm_groups[pp_stage].rank` is a bare subscript with no guard, while `_nccl_reshard_refit_impl` guards the identical hazard with a named `RuntimeError` ([lines 1676-1683](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L1676)). Currently unreachable in practice — `prepare_nccl_reshard_refit_info`'s docstring ([lines 1187-1188](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L1187)) documents the ordering contract that keeps this safe today — but a future refactor that violates that order would hit an opaque `TypeError`/`KeyError` here instead of a clear error. Non-blocking; consider mirroring the existing guard for consistency.

(Corresponds to still-open thread https://github.com/NVIDIA-NeMo/RL/pull/3659#discussion_r3911120857.)

nemo_rl/weight_sync/xferdtensor.py
```diff
@@ -239,22 +239,18 @@ def _compute_shard_slices(global_shape, mesh_shape, mesh_coords, placements):
...
-        # This rank owns chunk `linear_index`; its slice starts past every
-        # earlier chunk. (example: start = 64+64 = 128, end = 192 -> [128:192].)
         start = sum(sizes[:linear_index])
         end = start + sizes[linear_index]
         slices[tensor_dim] = slice(start, end)
 
     return slices
 
-
 def xferdtensor_golden(
```
sharonyu-115: **Resolving this thread — verified fixed.**

`git diff 8a0f5737f44ea60dd64d66ba5fc607fed837ab89 ddf7495d48715205cb14b1ab0d7854618fe2cbf8 -- nemo_rl/weight_sync/xferdtensor.py` shows the new `get_local_shard_slices()` API and the rewritten balanced-split `_compute_shard_slices` were fully removed in commit `ddf7495d4`, along with their duplicate tests. `_compute_shard_slices` is back to the original `torch.chunk`-based algorithm (matching pre-PR baseline). The TRTLLM refit path now sources its shard slices from the existing `xferdtensor_python._compute_shard_slices` via the new `_local_shard_slices` helper in [`vllm_backend.py:122-142`](https://github.com/NVIDIA-NeMo/RL/blob/ddf7495d48715205cb14b1ab0d7854618fe2cbf8/nemo_rl/models/generation/vllm/vllm_backend.py#L122), reusing established logic rather than introducing a new public API in the shared module. The only remaining diff in this file is comment/whitespace removal. Thanks for pushing on scope here.

(Original thread: https://github.com/NVIDIA-NeMo/RL/pull/3659#discussion_r3911120872.)

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1178,9 +1378,15 @@ def post(ctx: RefitCtx) -> None:
         }
         vllm_params = dict(self.model_runner.model.named_parameters())
         vllm_names_by_id = {id(param): name for name, param in vllm_params.items()}
+        use_trtllm_staging = self._uses_unquantized_flashinfer_trtllm()
```
sharonyu-115: Hi @seonjinn, I wonder if this staging decision needs to be made per realized MoE module rather than model-wide. A global FP8/MXFP8 `quant_config` can still contain ignored first/last `RoutedExperts` modules that vLLM constructs with `UnquantizedFusedMoEMethod` ([standard FP8 path](https://github.com/vllm-project/vllm/blob/752a3a504485790a2e8491cacbb35c137339ad34/vllm/model_executor/layers/quantization/fp8.py#L197-L217), [generic fallback](https://github.com/vllm-project/vllm/blob/752a3a504485790a2e8491cacbb35c137339ad34/vllm/model_executor/layers/fused_moe/routed_experts.py#L186-L202)); on Blackwell those BF16 modules can realize FlashInfer TRTLLM. Today `_uses_unquantized_flashinfer_trtllm()` short-circuits whenever the global `quant_config` is non-`None`, making `use_trtllm_staging=False` for the whole model. It seems like the ignored BF16 experts will then fall through to direct/merged NCCL writes into their TRTLLM BlockMajorK runtime buffers, recreating the corruption this PR fixes. If mixed models are intentionally out of scope, please reject this configuration explicitly instead of silently falling through, and add coverage for FP8 and MXFP8 with BF16 first/last MoE layers.

examples/configs/recipes/llm/grpo-qwen3.5-35ba3b-6n4g-async-1off-bf16-trtllm.yaml
```diff
@@ -0,0 +1,44 @@
...
+    vllm_cfg:
+      async_engine: true
+      precision: bfloat16
+      tensor_parallel_size: 4
+      pipeline_parallel_size: 1
+      expert_parallel_size: 4
+      gpu_memory_utilization: 0.8
+      enforce_eager: false
+    vllm_kwargs:
+      moe_backend: flashinfer_trtllm
```
sharonyu-115, line 27: Question: because the regression here is silent expert-weight corruption rather than a refit exception, what test demonstrates that the final TRTLLM-packed weights or logits are numerically correct after refitting this recipe? The current unit tests verify staging shape, values passed to the loader, naming, and lifecycle order, but they mock the actual vLLM load/finalize path; the exact Qwen3.5 end-to-end script is also still disabled pending established metric bounds. Could we add or run a Blackwell integration that refits distinctive weights A to B (ideally then B to C) and compares fixed-input logits with a freshly initialized model using B/C? At minimum, I think we need to have one nightly test with a train-versus-generation multi logprob error/KL parity threshold.

sharonyu-115 (changes requested):

seonjinn:
Mixed-scope validation update on `5b14cad6`:

- Qwen and Nemotron-style module names were tested with `(first N, last M)` values `(0,0)`, `(5,0)`, `(0,4)`, `(1,1)`, `(2,6)`, `(3,5)`, `(1,3)`, `(7,3)`, `(26,26)`, `(30,30)`, `(40,0)`, and `(0,40)`.
- Boundary-layer QKVO and routed experts stay BF16; middle-layer QKVO and routed experts use MXFP8; gates and shared experts stay BF16.
- Invalid negative, non-integer, and out-of-range boundaries fail before vLLM patching.
- Focused scope result: **35 passed, 42 deselected**. GitHub checks pass.

The combined mixed-refit stack also completed two colocated sync CUDA IPC runs with CUDA Graph on:

| Model | BF16 boundary | Result | `gen_kl_error` mean | Run |
|---|---:|---:|---:|---|
| Nemotron 3.5 Lightning 30B-A3B | first 1 / last 3 | 20/20 steps | 0.004403 | [W&B](https://wandb.ai/nvidia/nemo-rl-pr3294-mixed-sync/runs/zyu9yud0) |
| Qwen3.5 35B-A3B | first 3 / last 5 | 20/20 steps | 0.001732 | [W&B](https://wandb.ai/nvidia/nemo-rl-pr3294-mixed-sync/runs/66ue66te) |

These end-to-end runs include the dependent sync-refit changes and are compatibility evidence, not PR #3659-only performance results.

Nano caveat: its forward `gen_kl_error` meets the existing MXFP8 functional threshold, but reverse `policy_kl_error` has large tail outliers. A matched first-1/last-3 sync [MoE-only control](https://wandb.ai/nvidia/nemo-rl-pr3294-mixed-sync/runs/zc1jta1h) also completed 20/20 steps. Compared with MoE-only, QKVO+MoE reduced E2E throughput by 4.67%, reduced generation throughput by 4.28%, doubled mean generation KL, and increased reverse-KL median from 35.86 to 180.89. The refit mechanism works, but Nano should keep QKVO in BF16 and use MoE-only MXFP8 until the remaining reverse-KL tail is understood.

seonjinn:
/ok to test [96101df](https://github.com/NVIDIA-NeMo/RL/pull/3659/commits/96101df6434b6e5b7f4de684f6679557b99ed601)

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Fixed in `748fc71f` and `48301b6a`.

The decision is now made from each realized MoE module, not from the model-wide `quant_config`. `_unquantized_flashinfer_trtllm_modules()` finds modules whose realized method is `UnquantizedFusedMoEMethod` with the FlashInfer TRTLLM backend, and `build_hf_to_local_param_map()` stages only grouped-expert parameters owned by those modules. The partial reload lifecycle initializes/finalizes only those BF16 TRTLLM modules; MXFP8 MoE and linear modules remain on their quantized path and rebuild their layouts separately.

Coverage includes:
- `test_quantized_model_uses_native_refit_for_realized_bf16_trtllm_modules`
- `test_unquantized_trtllm_param_ids_are_scoped_to_realized_modules`
- `test_build_hf_to_local_param_map_stages_only_bf16_trtllm_experts`
- `test_mixed_mxfp8_native_refit_processes_each_module_once` for IPC, collective, and NCCL reshard

The vLLM unit-test shards pass on the current head.

examples/configs/recipes/llm/grpo-qwen3.5-35ba3b-6n4g-async-1off-bf16-trtllm.yaml
seonjinn: The current end-to-end evidence is stronger than the mocked unit tests, but it does not fully replace the A-to-B-to-C comparison you proposed.

On GB200, the combined mixed-refit stack completed 20/20 colocated sync steps for both models:
- Nemotron 3.5 Lightning 30B-A3B, first 1 / last 3 BF16: mean `gen_kl_error=0.004403` ([run](https://wandb.ai/nvidia/nemo-rl-pr3294-mixed-sync/runs/zyu9yud0))
- Qwen3.5 35B-A3B, first 3 / last 5 BF16: mean `gen_kl_error=0.001732` ([run](https://wandb.ai/nvidia/nemo-rl-pr3294-mixed-sync/runs/66ue66te))

Those runs exercise the final TRTLLM-packed weights through generation and compare generation logprobs with policy logprobs. They are combined-stack sync evidence, not the exact async NCCL-reshard recipe and not a fresh-model B/C logits comparison. I agree that the disabled recipe should not be treated as complete coverage. I am keeping this thread open until we post an async NCCL-reshard Blackwell run with an explicit train-versus-generation KL threshold; a stable nightly should then encode that threshold.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -607,15 +626,17 @@ def get_quantized_weight_iterator(
...
+    model_runner: Any,
+    *,
+    model_load_weights: Callable[..., object] | None = None,
 ) -> None:
     """Quantize weights for the legacy direct model-loading path."""
-    # Finally load the weights into vllm
-    model_runner.model.load_weights(
+    if model_load_weights is None:
+        model_load_weights = model_runner.model.load_weights
+    model_load_weights(
```
sharonyu-115, line 637: There seems to be a potential issue that needs to be verified: 
`get_quantized_weight_iterator(...)` returns a single-use generator. When the mixed BF16-TRTLLM layerwise lifecycle is active, the supplied `_load_full_hf_weights` callback consumes that generator once to collect storage pointers, then passes the exhausted iterator to vLLM. The load can therefore succeed with zero entries: IPC/legacy-collective batches are skipped, while NCCL reshard can update the bulk FFN path but leave misc weights stale.

Probably need to preserve the iterator contents (for example, materialize it once or record storage pointers as vLLM consumes it). Also suggest to add a unit test that invokes this real callback with `_nrl_layerwise_reload_active=True` and verifies the underlying model loader receives every generated weight.

tests/unit/models/generation/test_qwen35_bf16_trtllm_recipe.py
```diff
@@ -0,0 +1,71 @@
...
+    register_omegaconf_resolvers()
+    recipe_path = PROJECT_ROOT / "examples/configs/recipes/llm" / RECIPE_NAME
+    recipe = OmegaConf.to_container(
+        load_config_with_inheritance(recipe_path), resolve=True
+    )
+    assert isinstance(recipe, dict)
+    return recipe
+
+
+def test_qwen35_bf16_trtllm_recipe_uses_nccl_reshard() -> None:
```
sharonyu-115, line 39: It seems like this new recipe-test file is not selected by any L0 unit shard. The Models and Other shards ignore this directory, the vLLM base shard only collects `test_vllm*.py` (plus a small allowlist), and its `--vllm-only` catch-all retains only `pytest.mark.vllm` tests. Because this file is named `test_qwen35...` and is unmarked, both tests are skipped in CI.

Could you please verify if this test is really executed and if not, rename it to `test_vllm_qwen35_bf16_trtllm_recipe.py` so the lightweight tests enter the vLLM base shard, or explicitly add the file to an L0 shard.

docs/design-docs/nccl-reshard-refit.md
sharonyu-115, line 42: This support statement omits two hard restrictions enforced by `_validate_native_layerwise_refit`: BF16 FlashInfer TRTLLM refit rejects an FP8 KV cache and any co-trained MTP drafter. The later generic misc-path description mentions both MTP weights and FP8 KV-cache scales without qualifying that this backend combination cannot use them, so readers can reasonably configure a path that fails during initialization.

Please list both unsupported combinations here and qualify the later MTP/KV-cache wording accordingly.

sharonyu-115 (changes requested):

tests/unit/models/generation/test_vllm_qwen35_bf16_trtllm_recipe.py
```diff
@@ -0,0 +1,71 @@
...
+    register_omegaconf_resolvers()
+    recipe_path = PROJECT_ROOT / "examples/configs/recipes/llm" / RECIPE_NAME
+    recipe = OmegaConf.to_container(
+        load_config_with_inheritance(recipe_path), resolve=True
+    )
+    assert isinstance(recipe, dict)
+    return recipe
+
+
+def test_qwen35_bf16_trtllm_recipe_uses_nccl_reshard() -> None:
```
seonjinn: Renamed the file to `test_vllm_qwen35_bf16_trtllm_recipe.py`, so it matches the L0 `test_vllm_*.py` selection. Both recipe tests pass.

docs/design-docs/nccl-reshard-refit.md
seonjinn: Updated the limitations section to state explicitly that this BF16 FlashInfer TRTLLM path rejects FP8 KV cache and a co-trained MTP drafter. The later MTP/KV-cache wording is now qualified as generic refit behavior rather than support provided by this path.

nemo_rl/models/generation/vllm/quantization/fp8.py
seonjinn: Fixed in 38deaed6. `_load_full_hf_weights()` now wraps the iterable in a one-pass generator that records each tensor storage pointer as vLLM consumes it, so the original generator is neither pre-consumed nor materialized. The regression test failed on the old code with an empty received list and passes after the fix; the full backend test file also passes (79 tests).

seonjinn:
/ok to test [eb7529f](https://github.com/NVIDIA-NeMo/RL/pull/3659/commits/eb7529fc99abdd5ea9fceec1b98433a26989d1b4)

tests/unit/models/generation/test_nccl_reshard_backend.py
seonjinn: Fixed in `d0bcb961` and retained on the current head. The fixture now initializes `pp_comm_groups = {}` before calling the implementation. The backend suite passes on GB200.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Fixed on the current head. Native BF16 TRTLLM layerwise refit now rejects FP8 KV cache during setup, before any transfer can leave stale scales. The design document lists this restriction.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Fixed on the current head. The validation accepts both the direct fused name and the DeepSeek `.routed_experts.` spelling, and rejects the load only when neither destination is reported.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Fixed on the current head. Preparing the TRTLLM refit map before communication groups exist now raises a named lifecycle-order `RuntimeError` instead of a `NoneType` subscript error.

nemo_rl/weight_sync/xferdtensor.py
seonjinn: Applied. The `xferdtensor.py` behavior change was removed. Unsupported sharding is rejected explicitly when the refit plan is built.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: I kept the batched load here because splitting it into one vLLM reload call per expert would directly regress the refit path this PR optimizes. The geometry is validated before transfer: linear placement, even expert ownership, no EPLB/redundant experts, and no non-expert-dimension shard. The fused destination-name check remains an additional failure check. I am leaving this thread open in case you prefer the per-expert defense despite that cost.

tests/unit/models/generation/test_nccl_reshard_backend.py
seonjinn: Fixed in `d0bcb961` and retained on the current head. The standalone test now patches both `get_rank()` and `get_world_size()`, so it no longer depends on another test leaving a process group initialized.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -125,15 +127,72 @@ def _model_uses_unquantized_flashinfer_trtllm(model: torch.nn.Module) -> bool:
             UnquantizedFusedMoEMethod,
         )
     except ImportError:
-        return False
+        return []
```
sharonyu-115: [`vllm_backend.py:129-130`](https://github.com/NVIDIA-NeMo/RL/blob/3d2d7696749af4a350967154837823f06ce3c9a7/nemo_rl/models/generation/vllm/vllm_backend.py#L129-L130)

This is an adjacent fail-open guard, but its empty result now controls the new NCCL staging map. If either version-sensitive import moves, [`_uses_unquantized_flashinfer_trtllm()`](https://github.com/NVIDIA-NeMo/RL/blob/3d2d7696749af4a350967154837823f06ce3c9a7/nemo_rl/models/generation/vllm/vllm_backend.py#L910-L919) reports false, [`_validate_native_layerwise_refit()`](https://github.com/NVIDIA-NeMo/RL/blob/3d2d7696749af4a350967154837823f06ce3c9a7/nemo_rl/models/generation/vllm/vllm_backend.py#L942-L947) returns early, and the [NCCL map](https://github.com/NVIDIA-NeMo/RL/blob/3d2d7696749af4a350967154837823f06ce3c9a7/nemo_rl/models/generation/vllm/vllm_backend.py#L1506-L1516) leaves grouped experts on direct/merged receive. That reopens the silent packed-layout corruption this PR is intended to prevent, rather than making the placement check raise as the PR body states.

### AI-1

Please fail closed on this `ImportError` (for example, wrap and re-raise it with a supported-vLLM-version message) and add a failure-path unit test. The pinned supported versions provide both types, so this rejects no supported configuration.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -125,15 +127,72 @@ def _model_uses_unquantized_flashinfer_trtllm(model: torch.nn.Module) -> bool:
...
+
+
+def _reload_target_module_ids(
+    reload_targets: Sequence[torch.nn.Module],
+) -> set[int]:
+    """Return every module covered by the native reload targets."""
+    return {id(module) for target in reload_targets for module in target.modules()}
+
+
+def _process_mxfp8_modules_after_native_reload(
```
sharonyu-115: [`vllm_backend.py:150-168`](https://github.com/NVIDIA-NeMo/RL/blob/3d2d7696749af4a350967154837823f06ce3c9a7/nemo_rl/models/generation/vllm/vllm_backend.py#L150-L168)

**1 action item.**

The PR body says quantized FP8/MXFP8 paths are untouched, but this PR adds mixed-model MXFP8 post-processing and changes realized BF16 TRTLLM module selection inside globally quantized models.

### AI-1

Please update “What is NOT affected” to distinguish pure quantized models, which remain unchanged, from mixed FP8/MXFP8 plus BF16 TRTLLM models, which this PR intentionally changes.

sharonyu-115 (dismissed):

seonjinn:
/ok to test [fecc707](https://github.com/NVIDIA-NeMo/RL/pull/3659/commits/fecc7075ec371d4fc306f0dd474b67a3bf53c85f)

seonjinn:
/ok to test [cecdaf8](https://github.com/NVIDIA-NeMo/RL/pull/3659/commits/cecdaf83a127e73055592ae576bd1dede3fb8079)

seonjinn:
/ok to test [8bb95b3](https://github.com/NVIDIA-NeMo/RL/pull/3659/commits/8bb95b3e6fca19a80ef5272092b3e5463f59c751)

seonjinn:
/ok to test [7126eed](https://github.com/NVIDIA-NeMo/RL/pull/3659/commits/7126eedc7ad1c0e13736c6ea2244f166d0a60160)

terrykong (approved):

PR #3660, opened 2026-08-16
fix(sc): reduce and finalize gradients once per step, not per chunk

saumishr:
The SingleController streaming train path feeds N chunks into a single optimizer step, but two paths still behaved as though each chunk were a step of its own.

mcore calls config.finalize_model_grads_func at the end of every forward_backward_func invocation, so under streaming it fired once per chunk. Nothing already overridden gated it: it reaches finish_grad_sync directly, and model.no_sync() only clears is_last_microbatch, which finish_grad_sync ignores. With a distributed optimizer the reduce-scatter also writes into the buffer it reads, leaving this rank's shard DP-summed and the remainder local for the next chunk to accumulate on top of. Null the hook for the step's duration and invoke the saved callable once from _finish_train_step_body with num_tokens=None, since this path applies its own 1/N. Relocating rather than dropping it matters because it also owns the non-tensor-parallel/layernorm all-reduce across TP, the tied word- and position-embedding all-reduces across PP, the MoE router expert-bias update, and reset_model_temporary_tensors.

prepare_for_lp_inference offloaded the grad buffers on every logprob detour. mcore's offload_to_cpu(move_grads=True) does not copy the gradients anywhere, it resizes their storage to 0, and the matching reload zeroes it; param.main_grad stays a valid view throughout, so nothing raises and every chunk but the last is silently discarded against a 1/N normalizer that still counts them all. Add keep_train_buffers, which the controller passes as step_open.

Record the chunk count per step alongside grad_norm and sample the CUDA allocator at the phase boundaries under debug, so the accumulation is greppable from a run rather than inferred from timer ratios.

Cover both bugs in tests: the finalize hook is saved, nulled, invoked exactly once with num_tokens=None whether the step arrived as one chunk or five, and restored on finish and abort; the grad buffers survive a logprob detour mid-step; and the controller passes keep_train_buffers=step_open across a multi-chunk step.

# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/__init__.py
+++ b/nemo_rl/__init__.py
@@ -21,6 +21,24 @@ logging.basicConfig(
     format="%(levelname)s:%(name)s:%(filename)s:%(lineno)d: %(message)s",
 )
 
+# The call above installs the root handler but sets no level, so the root logger
+# stays at WARNING -- and because basicConfig returns early once handlers exist,
+# no later call can raise it (virtual_cluster.py's level=INFO is already a
+# no-op). Every nemo_rl logger inherited that, so until this line no log.info or
+# log.debug record anywhere in the package could reach a run's output, and the
+# instrumentation that emits them was unreachable rather than merely quiet.
+#
+# Set the level on the `nemo_rl` logger rather than on root: records still reach
+# the root handler, but raising the level does not also turn on debug output for
+# torch, ray, httpx and every other third-party logger.
+#
+# INFO by default. That is a change -- it surfaces the package's existing info
+# records, which are ~50 call sites concentrated in venv, cluster and vLLM-patch
+# setup, so they land at startup rather than per step. DEBUG additionally
+# enables the per-chunk and CUDA-allocator records on the training path, which
+# are per chunk per rank and are not meant for a normal run.
+logging.getLogger("nemo_rl").setLevel(os.environ.get("NRL_LOG_LEVEL", "INFO").upper())
+
 """
 This is a work around to ensure whenever NeMo RL is imported, that we
 add Megatron-LM to the python path. The editable install of megatron-core
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -35,6 +35,7 @@ Data flow:
 from __future__ import annotations
 
 import asyncio
+import logging
 import os
 import time
 from functools import partial
@@ -74,6 +75,10 @@ from nemo_rl.utils.logger import Logger
 from nemo_rl.utils.timer import TimeoutChecker, Timer
 
 Generation = Union[VllmGeneration, SGLangGeneration]
+
+# Named `log` rather than `logger` to keep it distinct from the experiment
+# Logger this module also uses as `self._logger`.
+log = logging.getLogger(__name__)
 
 
 @ray.remote(num_cpus=1, num_gpus=0)  # pragma: no cover
@@ -487,6 +492,7 @@ class SingleControllerActor:
             evicted_stale_prompt_groups = 0
             min_sample_version = None
             step_open = False
+            chunks_dispatched = 0
             calibration_batches: list[BatchedDataDict[Any]] = []
 
             with self._timer.time("total_step_time"):
@@ -554,8 +560,15 @@ class SingleControllerActor:
                         or self._reference_logprobs_required
                     ):
                         with self._timer.time("logprob_inference_prep"):
+                            # Once the step is open, gradients are accumulating
+                            # in the trainer's grad buffers across chunks. The
+                            # Megatron buffer offload frees that storage outright
+                            # and its reload zeroes it, so offloading here would
+                            # discard every chunk but the last while the 1/N
+                            # normalizer still counts all of them.
                             await asyncio.to_thread(
-                                self._trainer.prepare_for_lp_inference
+                                self._trainer.prepare_for_lp_inference,
+                                keep_train_buffers=step_open,
                             )
                         with self._timer.time("policy_and_reference_logprobs"):
                             if self._policy_logprobs_required:
@@ -626,7 +639,33 @@ class SingleControllerActor:
                     )
 
                     groups_dispatched += num_groups
+                    chunks_dispatched += 1
+                    # How many chunks a step is split into decides how many times
+                    # gradients accumulate before the single reduce, so record it
+                    # rather than leaving it to be inferred from phase timings.
+                    #
+                    # These reach a run's output only because nemo_rl/__init__.py
+                    # sets the `nemo_rl` logger to NRL_LOG_LEVEL (INFO by
+                    # default); the bare basicConfig() there pins the root logger
+                    # at WARNING and no later basicConfig can raise it. Note that
+                    # the handler writes to stderr while the progress prints
+                    # around this write to stdout, so the two are not guaranteed
+                    # to interleave in order in a Ray driver log.
+                    log.info(
+                        "train_pump: step %d chunk %d: %d group(s), %d/%d dispatched",
+                        version_during_step,
+                        chunks_dispatched,
+                        num_groups,
+                        groups_dispatched,
+                        grpo_cfg.num_prompts_per_step,
+                    )
 
+                log.info(
+                    "train_pump: step %d closing on %d chunk(s), %d group(s)",
+                    version_during_step,
+                    chunks_dispatched,
+                    groups_dispatched,
+                )
                 with self._timer.time("policy_training"):
                     result = await asyncio.to_thread(self._trainer.finish_train_step)
 
--- a/nemo_rl/models/policy/workers/dtensor_policy_worker.py
+++ b/nemo_rl/models/policy/workers/dtensor_policy_worker.py
@@ -1938,7 +1938,16 @@ class DTensorPolicyWorkerImpl(
             self.model = self.move_to_cpu(self.model)
 
     @wrap_with_nvtx_name("dtensor_policy_worker/prepare_for_lp_inference")
-    def prepare_for_lp_inference(self) -> None:
+    def prepare_for_lp_inference(self, keep_train_buffers: bool = False) -> None:
+        """Put the model in eval mode for logprob inference.
+
+        Args:
+            keep_train_buffers: Leave the optimizer state on CUDA because a train
+                step is already open. This backend accumulates gradients in
+                ``param.grad`` and never offloads them, so unlike the Megatron
+                backend there is nothing here that could discard them; the flag
+                only suppresses the per-chunk optimizer round trip.
+        """
         # onload model to cuda
         if not self.cpu_offload:
             self.move_to_cuda(self.model)
@@ -1949,7 +1958,11 @@ class DTensorPolicyWorkerImpl(
 
         # offload optimizer to cpu
         torch.randn(1).cuda()  # wake up torch allocator
-        if self.optimizer is not None and self.offload_optimizer_for_logprob:
+        if (
+            not keep_train_buffers
+            and self.optimizer is not None
+            and self.offload_optimizer_for_logprob
+        ):
             self.move_optimizer_to_device("cpu")
 
         gc.collect()
--- a/nemo_rl/models/policy/workers/megatron_policy_worker.py
+++ b/nemo_rl/models/policy/workers/megatron_policy_worker.py
@@ -1133,12 +1133,17 @@ class MegatronPolicyWorkerImpl(
     #    ``forward_backward_func`` calls between ``optimizer.step()`` would
     #    over-count: each call's terminal reduce sums an already-reduced
     #    bucket again. We wrap every call in ``self.model.no_sync()`` so
-    #    hooks accumulate locally only; one explicit ``start_grad_sync`` +
-    #    ``finish_grad_sync`` at finish does the single true reduce.
-    # 2. PP>1: the pipeline scheduler invokes ``config.grad_sync_func``
-    #    directly on last-microbatch boundaries — this bypasses the
-    #    ``no_sync`` gate. We null it for the duration of the step and
-    #    restore at finish/abort.
+    #    hooks accumulate locally only; the single true reduce happens once at
+    #    finish, inside the relocated ``finalize_model_grads_func`` call
+    #    (preceded by ``start_grad_sync`` when ``overlap_grad_reduce`` is set).
+    # 2. Three ``model.config`` hooks would otherwise fire a reduce mid-step,
+    #    none of them gated by the ``no_sync`` above: ``grad_sync_func`` (the
+    #    PP>1 scheduler calls it directly on last-microbatch boundaries),
+    #    ``no_sync_func`` (the PP=1 schedule runs its last microbatch outside
+    #    it) and ``finalize_model_grads_func`` (end of every
+    #    ``forward_backward_func``, i.e. once per chunk). All three are nulled
+    #    for the duration of the step and restored at finish/abort; see
+    #    ``begin_train_step`` for what each one does.
     # 3. Grad clip is bundled inside ``MegatronOptimizer.step()``; the 1/N
     #    rescale via ``self.model.scale_gradients(1/N)`` must run before
     #    ``optimizer.step()`` so the clip operates on the rescaled grad.
@@ -1286,21 +1291,17 @@ class MegatronPolicyWorkerImpl(
         #                    pair (typically step 2).
         #   finalize_model_grads_func
         #                  — every mcore schedule calls this at the END of each
-        #                    ``forward_backward_func`` invocation (schedules.py,
-        #                    ``if config.finalize_model_grads_func is not None
-        #                    and not forward_only``), so under streaming it fires
-        #                    once per chunk rather than once per step. Nothing we
+        #                    ``forward_backward_func`` invocation, so under
+        #                    streaming it fires once per chunk. Nothing we
         #                    already override gates it: it reaches
-        #                    ``finish_grad_sync()`` directly, and
-        #                    ``model.no_sync()`` only clears
-        #                    ``is_last_microbatch``, which is consulted by
-        #                    ``register_grad_ready`` on the overlap path and
-        #                    ignored by ``finish_grad_sync``. With a distributed
-        #                    optimizer the reduce-scatter also writes into the
-        #                    buffer it reads, leaving this rank's shard DP-summed
-        #                    and the remainder local for the next chunk to
-        #                    accumulate on top of. The real callable runs once,
-        #                    in ``_finish_train_step_body``.
+        #                    ``finish_grad_sync()`` directly, which ignores the
+        #                    ``is_last_microbatch`` flag that ``model.no_sync()``
+        #                    clears. With a distributed optimizer the
+        #                    reduce-scatter also writes into the buffer it reads,
+        #                    leaving this rank's shard DP-summed and the
+        #                    remainder local for the next chunk to accumulate on
+        #                    top of. The real callable runs once, in
+        #                    ``_finish_train_step_body``.
         # Save all three so finish/abort restores them.
         # Read "config" via getattr-by-string so the token stays out of
         # begin_train_step.__code__.co_names; otherwise cloudpickle matches
@@ -1550,55 +1551,44 @@ class MegatronPolicyWorkerImpl(
         # owns more than the cross-DP reduce: the non-tensor-parallel / layernorm
         # all-reduce across TP (sequence parallelism), the tied word-embedding
         # and position-embedding all-reduces across PP, the MoE router
-        # expert-bias update, and reset_model_temporary_tensors. Before this
-        # change those were reached ONLY via the per-chunk call, so dropping the
-        # hook outright would have silently lost all of them.
+        # expert-bias update, and reset_model_temporary_tensors.
         #
-        # ``num_tokens=None`` is deliberate: the 1/N normalization is applied
-        # just above from this path's own accumulated valid-token count, and
-        # mcore rescales only when num_tokens is not None, so passing None is
-        # what avoids double-normalizing.
+        # Both arguments are load-bearing by omission:
+        #   num_tokens=None — mcore rescales only when it is not None, and the
+        #     1/N normalization is already applied just above from this path's
+        #     own accumulated valid-token count.
+        #   no pg_collection — the saved callable is not bare
+        #     ``finalize_model_grads`` but Megatron-Bridge's
+        #     ``partial(finalize_model_grads, pg_collection=...)``
+        #     (bridge/training/setup.py, ``_update_model_config_funcs``), so
+        #     omitting the keyword lets that binding through where supplying one
+        #     would override it. It is group-equivalent to the
+        #     ``_build_default_pg_collection()`` the schedule passed on the
+        #     per-chunk path, down to the ``tp_dp_cp`` field that is the one
+        #     thing finalize_model_grads asserts on, for the expert-bias update.
         #
-        # No ``pg_collection`` keyword, which is not the same as leaving the
-        # parameter at its default: the saved callable is not bare
-        # ``finalize_model_grads`` but Megatron-Bridge's
-        # ``partial(finalize_model_grads, pg_collection=...)``
-        # (bridge/training/setup.py, ``_update_model_config_funcs``), built in
-        # nemo_rl/models/megatron/setup.py from
-        # ``ProcessGroupCollection.use_mpu_process_groups()``. Omitting the
-        # keyword lets that binding through; supplying one would override it,
-        # since an explicit keyword beats a partial's.
-        #
-        # The per-chunk path ran with a different collection object but the same
-        # groups, which is what makes relocating the call safe:
-        # megatron_forward_backward calls forward_backward_func without one, so
-        # the schedule built ``_build_default_pg_collection()`` from
-        # ``parallel_state`` and passed it explicitly, overriding the binding.
-        # Both collections carry the same tp / pp / embd / pos_embd / dp_cp
-        # groups, and both populate ``tp_dp_cp`` — the single field
-        # finalize_model_grads asserts on, for the MoE router expert-bias
-        # update — so recipes that set ``moe_router_enable_expert_bias`` satisfy
-        # it either way.
-        #
-        # ``force_all_reduce`` is genuinely left at its default. The schedule
-        # passes it explicitly, but its default is False on both sides and
-        # nothing in NeMo-RL sets it, so the reduce is dispatched the same way.
-        #
-        # Megatron-core's BucketGroup.finish_grad_sync, when
-        # overlap_grad_reduce=False, internally dispatches the synchronous
-        # collective via start_grad_sync(force_all_reduce=...). On the overlap
-        # path ``model.no_sync()`` stopped register_grad_ready from dispatching,
-        # so there is no outstanding handle for the finalize's finish_grad_sync
-        # to wait on — fire start_grad_sync ourselves in that case only.
+        # Checked before anything is dispatched, and asserted rather than
+        # falling back to a bare ``finish_grad_sync()``: that reduces across DP
+        # and silently skips every other collective listed above, which is wrong
+        # gradients with no error. Bridge binds the hook whenever an optimizer
+        # exists, and an open train step implies one because this method steps it
+        # below, so None here means the model config never went through setup.
+        finalize_model_grads_func = state["saved_finalize_model_grads_func"]
+        assert finalize_model_grads_func is not None, (
+            "finalize_model_grads_func was None at finish_train_step; expected "
+            "Megatron-Bridge's _update_model_config_funcs to have bound it on "
+            "the model config during setup"
+        )
+
+        # With overlap_grad_reduce, ``model.no_sync()`` stopped
+        # register_grad_ready from dispatching, so the finalize's internal
+        # finish_grad_sync has no outstanding handle to wait on — start the
+        # reduce ourselves in that case only.
         if self.cfg["megatron_cfg"]["distributed_data_parallel_config"][
             "overlap_grad_reduce"
         ]:
             self.model.start_grad_sync()
-        finalize_model_grads_func = state.get("saved_finalize_model_grads_func")
-        if finalize_model_grads_func is not None:
-            finalize_model_grads_func([self.model], None)
-        else:
-            self.model.finish_grad_sync()
+        finalize_model_grads_func([self.model], None)
 
         # Wait for the comm-stream reduce dispatched above before opt.step
         # reads main_grad. Without this, grad_norm collapses to ~1/2.
@@ -2891,6 +2881,17 @@ class MegatronPolicyWorkerImpl(
                 only the last chunk's contribution against a 1/N normalizer
                 computed over all of them. Keeping the buffers resident also
                 avoids round-tripping tens of GiB per chunk.
+
+                Suppressing the offload here is sufficient only because of an
+                mcore invariant on the other side of the detour:
+                ``prepare_for_training`` runs before every chunk and reloads with
+                ``move_grads=True``, and ``reload_from_cpu`` resizes and
+                ``zero_()``s the grad buffer only ``if grad_data_size > 0``, a
+                counter set only by a matching ``offload_to_cpu``. With the
+                offload suppressed it stays 0, so the reload is a no-op and
+                the accumulated gradients survive. An mcore change that dropped
+                that guard, or that zeroed unconditionally, would silently
+                reinstate this bug.
         """
         # First worker call after the controller's select() wait returns, so this
         # is also the "idle wait end" marker. Whether the offload was suppressed
--- a/nemo_rl/models/policy/workers/megatron_policy_worker.py
+++ b/nemo_rl/models/policy/workers/megatron_policy_worker.py
@@ -1174,9 +1174,14 @@ class MegatronPolicyWorkerImpl(
             "all_mb_metrics": [],
             "mb_losses": [],
             "total_num_microbatches": 0,
+            # One increment per train_microbatch call, i.e. the number of
+            # streaming chunks the controller has fed into this optimizer step
+            # so far.
+            "num_chunks": 0,
             # Saved across the step so we can restore at finish/abort.
             "saved_grad_sync_func": None,
             "saved_no_sync_func": None,
+            "saved_finalize_model_grads_func": None,
         }
 
     def _assert_step_open(self) -> dict[str, Any]:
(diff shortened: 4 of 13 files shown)


nemo_rl/algorithms/single_controller.py
```diff
@@ -626,7 +639,25 @@ async def _train_pump(self) -> None:
                     )
 
                     groups_dispatched += num_groups
+                    chunks_dispatched += 1
+                    # How many chunks a step is split into decides how many times
+                    # gradients accumulate before the single reduce, so record it
+                    # rather than leaving it to be inferred from phase timings.
+                    log.info(
```
saumishr, line 654: **[BUG] None of the telemetry this PR adds is emitted in a default run** (confidence: 88)

`nemo_rl/__init__.py:20` calls `logging.basicConfig(format=...)` with no `level`, which installs a root handler and leaves the root logger at `WARNING`. Every later `basicConfig` — including `virtual_cluster.py:30`'s `level=logging.INFO` — then sees `root.handlers` non-empty and returns early, so it cannot raise the level. Ray does not rescue this either: its worker entrypoint's `setup_logger` only does `logging.getLogger("ray").setLevel(...)` and never touches the root logger. Nothing anywhere calls `setLevel` on the `nemo_rl` logger.

So both of these `log.info` lines, every new `log.debug` line, and `_log_gpu_mem` in its entirety are dropped — while the progress `print(..., flush=True)` at line 536 of this same function does appear. The chunk-count and `grad_norm` record this PR exists to make greppable produces nothing.

Confirmed in production since this review was drafted: a 15-step 68-node run on a branch containing this PR has **zero** `train_pump` matches across its entire 12 MB driver log.

**Action:** either use `print(..., flush=True)` here to match the rest of `_train_pump`, or raise the level explicitly once in the actor (`logging.getLogger("nemo_rl").setLevel(os.environ.get("NRL_LOG_LEVEL", "WARNING"))`) and document the knob. Same applies to `megatron_policy_worker.py:1612`.

While you are there: the four `.item()` calls in the per-chunk `log.debug` at `megatron_policy_worker.py:1394` are evaluated unconditionally, so today they only cost host-device syncs and print nothing — worth an `log.isEnabledFor(logging.DEBUG)` guard, matching the one `_log_gpu_mem` already has.

The `nemo_rl/__init__.py` root cause is adjacent, not introduced here.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1453,17 +1541,47 @@ def _finish_train_step_body(self, state: dict[str, Any]) -> dict[str, Any]:
...
+        # expert-bias update, and reset_model_temporary_tensors. Before this
+        # change those were reached ONLY via the per-chunk call, so dropping the
+        # hook outright would have silently lost all of them.
+        #
+        # ``num_tokens=None`` is deliberate: the 1/N normalization is applied
+        # just above from this path's own accumulated valid-token count, and
+        # mcore rescales only when num_tokens is not None, so passing None is
+        # what avoids double-normalizing.
+        #
+        # ``pg_collection`` is likewise left at its default. megatron_forward_backward
```
saumishr: **[GUIDELINE] This paragraph is factually wrong, and it is the kind of comment someone will lean on when touching MoE or CP** (confidence: 92)

The saved callable is not bare `finalize_model_grads`. Megatron-Bridge's `_update_model_config_funcs` (`src/megatron/bridge/training/setup.py`) sets `config.finalize_model_grads_func = partial(finalize_model_grads, pg_collection=pg_collection)`, wired from `nemo_rl/models/megatron/setup.py` via `ProcessGroupCollection.use_mpu_process_groups()`. So the relocated call does **not** leave `pg_collection` at its default — the keyword is already bound, and the not-`None` branch runs.

The closing sentence is wrong in the other direction too: when a caller omits `pg_collection`, `forward_backward_func` builds `_build_default_pg_collection()` and passes it, so the per-chunk path took that same branch and always had to satisfy the `tp_dp_cp` assert.

This is a comment defect only, not a behavior change: the bound MPU collection is group-equivalent field by field to the schedule's default, and `use_mpu_process_groups(required_pgs=None)` does populate `tp_dp_cp`, so the assert still passes on `grpo-nemotron3-super-120BA12B`, which sets `moe_router_enable_expert_bias: true`. (This was nearly filed as a crash before that was checked.)

**Action:** rewrite the paragraph to say the hook is a Bridge-bound `partial` already carrying the MPU collection, that omitting the keyword deliberately lets that binding through, and that it is group-equivalent to what the schedule passed. Drop the "assert the per-chunk path never had to satisfy" claim.

tests/unit/models/policy/test_megatron_split_state.py
```diff
@@ -727,3 +786,292 @@ def test_handles_originally_none_grad_sync_func(self, mock_module_symbols):
...
+
+        w = _make_worker(LossType.TOKEN_LEVEL)
+        sentinel = w.model.config.finalize_model_grads_func
+        w.begin_train_step(loss_fn=w._test_loss_fn)
+        w.train_microbatch(_fake_batch())
+        w.finish_train_step()
+        args, kwargs = sentinel.call_args
+        assert args[0] == [w.model]
+        assert args[1] is None
+        # pg_collection is deliberately left at its default; passing the model's
```
saumishr: Same correction as `megatron_policy_worker.py:1560`: `pg_collection` is not left at a default — Megatron-Bridge binds it into the hook via `functools.partial`. Which means `kwargs == {}` here actually pins something useful (this call site adds no keyword of its own, so it cannot silently override the bound collection). Keep the assertion, fix the comment above it.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1187,20 +1191,64 @@ def _assert_step_open(self) -> dict[str, Any]:
...
     def _restore_saved_grad_sync_func(self, state: dict[str, Any]) -> None:
         """Restore the mcore hooks nulled in ``begin_train_step``.
 
-        Restores both ``grad_sync_func`` and ``no_sync_func`` from the
-        saved values on the open-step state. Idempotent on those values;
-        safe to call from the happy-path finish/abort or from a try/except
-        cleanup in train_microbatch / finish_train_step when those raise
-        mid-body. See begin_train_step for why ``.config`` is read via
-        getattr-by-string.
+        Restores ``grad_sync_func``, ``no_sync_func`` and
```
saumishr, line 1238: **[GUIDELINE] The method name says one hook; this docstring says three** (confidence: 92)

`_restore_saved_grad_sync_func` restores `grad_sync_func`, `no_sync_func` and `finalize_model_grads_func` now. Five nearby mentions went stale with it, and two are operator-facing: the warnings at lines 1359 and 1514 read "failed to restore grad_sync_func after …", which will misdirect anyone debugging a mid-step failure that actually lost the finalize hook. The except-block comments at 1350 and 1506 and the inline note at 1626 have the same problem.

**Action:** rename to `_restore_saved_mcore_hooks` — four call sites, one file, and no test references the private name — and update those five strings.

Minor, while you are in here: `chunk_index` is incremented before use, so it holds a count rather than an index; `num_chunks` would read better next to `total_num_microbatches`.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1187,20 +1191,64 @@ def _assert_step_open(self) -> dict[str, Any]:
             )
         return state
 
+    def _current_rank(self) -> int:
```
saumishr: **[GUIDELINE] This duplicates `get_rank_safe`, which the file already imports** (confidence: 82)

`get_rank_safe()` is imported at line 38, and its value is already cached as `self.rank`. All four call sites of this helper (1225, 1398, 1617, 2877) run after `setup_distributed()`, so this is redundancy rather than a latent bug — the missing `RANK`-env fallback is unobservable here.

**Action:** delete the helper and use `self.rank`. Note the test fixture then needs `w.rank = 0` in `_make_worker`, since those log calls evaluate the rank eagerly even with `_log_gpu_mem` stubbed out.

nemo_rl/models/policy/workers/dtensor_policy_worker.py
```diff
@@ -1949,7 +1958,11 @@ def prepare_for_lp_inference(self) -> None:
 
         # offload optimizer to cpu
         torch.randn(1).cuda()  # wake up torch allocator
-        if self.optimizer is not None and self.offload_optimizer_for_logprob:
+        if (
```
saumishr, line 1961: **[TEST] This branch has no coverage** (confidence: 85, low priority)

The Megatron equivalent got a whole test class (`TestPrepareForLpInference`); the `keep_train_buffers` branch added to both dtensor workers has none.

Worth stating why it is low priority: `begin_train_step` / `train_microbatch` / `finish_train_step` exist only on `MegatronPolicyWorkerImpl`, so a dtensor worker can never have an open step and `True` is unreachable here today. Inverting this condition would leave only a stale optimizer offload that the next `prepare_for_training` repairs.

**Action:** a ~15-line CPU-mockable test per file mirroring `TestPrepareForLpInference`, so the branch cannot be inverted unnoticed once dtensor gains the split API. Templates exist a few tests up in each file.

saumishr (commented):
## Review summary

Both bugs are real and were verified independently against the pinned Megatron-Core: `finalize_model_grads_func` is invoked at the end of every `forward_backward_func` call and reaches `finish_grad_sync` directly, and `offload_to_cpu(move_grads=True)` resizes grad storage to 0 rather than copying it. Seven of the eight upstream claims in the new comments check out line by line. The eighth is wrong and is the `pg_collection` comment flagged inline.

Two findings that could not be anchored to a diff line:

### [PERF-EVIDENCE] The PR ships no numbers, and this is not a small perturbation

Pre-fix, the per-chunk `finish_grad_sync` all-reduced a growing buffer at every chunk boundary and the finish path reduced it again, and the logprob detour freed all but the last chunk's contribution against an N-chunk normalizer. Gradient magnitude on the streaming path therefore moves substantially, which moves `clip_grad` behavior and the effective learning rate with it — meaning any hyperparameters tuned on that path were tuned against a broken normalizer.

Could we see before/after on `grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml`, same seed? It is the only nightly recipe with `min_groups_for_streaming_train` below `num_prompts_per_step` (16 vs 64), so the only one that produces multiple chunks:

1. `grad_norm` vs step
2. loss and task reward over >= 100 steps
3. peak HBM at the `lp_prep_exit` / `train_prep_exit` boundaries — that is what `keep_train_buffers` gives up, and what the new instrumentation was added to measure
4. a one-chunk control showing the change is inert when chunks == 1

### [TEST] Nothing that can run drives more than one chunk per optimizer step

Both bugs are chunk-count-dependent, but the GPU parity A/B in `tests/unit/models/policy/test_megatron_split_parity.py` — which `test_megatron_split_state.py`'s own docstring defers to for numerical equivalence — issues exactly one `train_microbatch` fan-out per step, and `tests/unit/single_controller/test_train_pump.py:317`, the only end-to-end pump test against real mcore, pins `min_groups_for_streaming_train=num_prompts`. Every new assertion in this PR sits on mocks where `megatron_forward_backward`, `scale_gradients`, `finish_grad_sync` and `move_model` are all fakes, so none of them can observe a wrong gradient.

The pump test is the cheaper and better-targeted place to add a chunked variant — single GPU, and mcore's grad-buffer offload/reload is not DP-dependent, so a numeric assertion there exercises the logprob-detour bug for real. A parity variant at `chunks=2, tp=1` (DP=2) is what would cover the reduce. Both files are outside this diff, so a follow-up is fine.

One thing worth watching when CI lands: since the hook was non-`None` before this PR (Bridge binds it), the single-chunk split path may have been reducing twice, in which case the existing parity baseline shifts. I could not settle that without running it.

---

**Also noted, below the bar for individual comments:** a `.v2` bump for the streaming2 recipe and its GB200 twin — `tests/test_suites/README.md` lists "convergence bug fix" as an explicit `.vN` trigger, though there are no stored numeric baselines to invalidate here, only loose threshold gates, so leaving this to a maintainer; a docstring sentence recording that the fix depends on mcore gating `reload_from_cpu` on `grad_data_size > 0`, since `prepare_for_training` runs on every chunk with `move_grads=True`; tests for the two restore-on-exception paths; a `TestLogGpuMem` pinning the "do not reset peak counters when DEBUG is off" contract.

**Checked and dropped:** keyword-only `*` on `keep_train_buffers`; `raise NotImplementedError` in the interface (every abstract sibling uses `pass`); the dtensor docstring reword (they already say what would have been asked); a `_MCORE_STEP_OVERRIDES` table; and a suspected OOM from `keep_train_buffers` overriding `offload_optimizer_for_logprob` — that knob is `false` throughout the streaming recipe's config chain, `prepare_for_training` re-onloads on the next call, and the grad buffer that must stay resident is larger than the DP-sharded optimizer state anyway.

---

*Reviewed by 5 specialist agents plus 2 adversarial passes: 13 findings confirmed, 9 disputed, 12 downgraded. Two self-overturns, both in the PR's favor — the `pg_collection` finding was nearly escalated to "crashes MoE recipes" before `use_mpu_process_groups(required_pgs=None)` proved it populates `tp_dp_cp`, and a `calculate_per_token_loss` compounding theory died on the loss wrapper's return arity. Repo linters: NOT RUN (no `uv`/`pre-commit` on the review host). No GPU, so no runtime verification of the new tests.*

nemo_rl/algorithms/single_controller.py
saumishr: Fixed in ab35ed039.

Both `_train_pump` lines are now `print(..., flush=True)`, matching the 23 other progress prints in that pump, and the module-level `log`/`import logging` this PR added are gone with them. The worker's `[step]` summary prints too, and is now rank-0 only: every field on it is already DP- or MP-reduced by that point, so the other 223 ranks of a 56-node run were only ever going to contribute duplicate lines and two host-device syncs each.

The per-chunk `log.debug` stays a debug record but is now wrapped in `log.isEnabledFor(logging.DEBUG)`, so its four `.item()` syncs are no longer paid on the critical path to build a message that is then dropped. `_log_gpu_mem` already had that guard.

New `TestStepSummaryIsEmitted` pins both halves — that `[step] chunks=2 ... grad_norm=0.5` reaches stdout, and that an off-rank worker stays silent — so this cannot regress to an invisible logger again.

I left `nemo_rl/__init__.py` alone. Raising the level there (or on the `nemo_rl` logger) would change what every module in the repo emits, which is a bigger blast radius than this PR should carry; happy to do it as a follow-up with an `NRL_LOG_LEVEL` knob if you'd prefer that over prints.

nemo_rl/models/policy/workers/megatron_policy_worker.py
saumishr: Correct, and thanks for chasing it down — rewritten in ab35ed039.

The paragraph now says the hook is Bridge's `partial(finalize_model_grads, pg_collection=...)` (`bridge/training/setup.py::_update_model_config_funcs`, fed from `nemo_rl/models/megatron/setup.py` via `use_mpu_process_groups()`), that omitting the keyword is what lets that binding through rather than leaving a default in force, and that the per-chunk path ran with a *different collection object but the same groups* — which is the actual reason relocating the call is safe. The "assert the per-chunk path never had to satisfy" claim is gone; `use_mpu_process_groups(required_pgs=None)` populates `tp_dp_cp`, so `moe_router_enable_expert_bias` recipes satisfy it either way.

One thing this turned up that neither of us had checked: `schedules.py` passes `force_all_reduce=force_all_reduce` alongside `pg_collection`, and the per-step call omits it. It defaults to `False` in both the schedule parameter and `finalize_model_grads`, and nothing in NeMo-RL sets it, so the two paths do dispatch the same reduce — but it is now stated in the comment rather than left to be re-derived.

tests/unit/models/policy/test_megatron_split_state.py
saumishr: Updated in ab35ed039. The comment now says what the assertion actually pins: no keywords at all, which is what keeps Bridge's `partial` binding intact, since an explicit keyword here would override it. Points at the call-site note for the group-equivalence argument rather than restating it.

nemo_rl/models/policy/workers/megatron_policy_worker.py
saumishr: Both done in ab35ed039.

Renamed to `_restore_saved_mcore_hooks` — four call sites, and no test referenced the private name. The two operator-facing warnings now read "failed to restore mcore hooks after ...", and the three comments that named only `grad_sync_func` (the `train_microbatch` and `finish_train_step` except blocks, and the notes at the finish and abort call sites) say "the mcore hooks".

`chunk_index` is now `num_chunks` in the step state, the test class, and the debug record's field label, which was `idx=` for a value that has always been a count.

nemo_rl/models/policy/workers/megatron_policy_worker.py
saumishr: Deleted in ab35ed039. All four sites use `self.rank` now, which the file already caches from `get_rank_safe()` at `__init__` and prints elsewhere (`_clear_fp8_caches`).

You were right about the fixture: `_make_worker` bypasses `__init__`, so `w.rank = 0` is set there explicitly. It is load-bearing beyond the stubbed `_log_gpu_mem` now, since the step summary reads the rank to decide whether to print — and `TestStepSummaryIsEmitted.test_only_rank_zero_prints` sets it to 3 to check the other side.

nemo_rl/models/policy/workers/dtensor_policy_worker.py
saumishr: Added in ab35ed039, for both dtensor workers.

`test_dtensor_prepare_for_lp_inference_keep_train_buffers` and its v2 twin are parametrized over both values and assert on the offload: `["cpu"]` when the flag is off, nothing when it is on. Both mirror the `prepare_for_training` test a few lines up, plus a `torch.randn` patch so the allocator wake-up's `.cuda()` does not require a GPU. The docstrings record why the `True` case is unreachable today, so nobody deletes them as dead coverage.

saumishr:
Pushed ab35ed039 with fixes for all six inline comments; replies on each thread. The two findings in the review body that were not anchored to a diff line:

**[TEST] the multi-chunk case** — not in this commit. Every new assertion still sits on mocks, and the two places that could drive a real multi-chunk step are both outside this diff and both need hardware I do not have in front of me: a `chunks=2, tp=1` variant of the parity A/B (needs DP=2, so 2 GPUs) and a chunked variant of `test_train_pump.py` against real mcore (1 GPU). The pump one is the better target — mcore's grad-buffer offload/reload is not DP-dependent, so a numeric assertion there exercises the logprob-detour bug for real on a single GPU. Happy to take it as a follow-up PR, or to fold it in here if you would rather not merge without it.

**[PERF-EVIDENCE]** — still owed. `grpo-llama3.1-8b-instruct-2n8g-async-1off-single-controller-streaming2.yaml` is the only nightly recipe that produces multiple chunks, and I have not run the before/after on it yet. Worth flagging that the `[step]` line now actually prints, so `grad_norm` vs chunk count is greppable straight out of the driver log, which is what that comparison needs.

One caveat on this commit: I could not execute the unit tests locally. The host has no `uv`, and its venv is missing `transformer_engine` (so `test_megatron_split_state.py` skips at the `megatron.bridge` import) and enough of the CPU-only chain that the dtensor files will not collect either. Everything here is `py_compile`-clean and reviewed by hand, but CI is the first real run — flag anything it turns up and I will fix it.

nemo_rl/algorithms/single_controller.py
saumishr: Correcting my reply above — it is stale in every particular. The branch has since been squashed and rebased onto current main, so `ab35ed0` no longer exists; the fix is in **e07e788ae**, and it went the other way.

**These are `log.info` again, not prints, and the level now comes from `NRL_LOG_LEVEL` (default INFO), set in `nemo_rl/__init__.py`.** I had argued that touching that file carried too much blast radius for this PR. That was based on a wrong assumption, and the history says the opposite:

- `virtual_cluster.py` has asked for `logging.basicConfig(level=logging.INFO)` since the initial commit (b1ba7144b, Mar 2025). It was the first `basicConfig` call then, so it worked.
- ed9cab7c1 ("feat: allow uv-less execution and fingerprint the environment", #1491, Dec 2025) added a **format-only** `basicConfig` to `nemo_rl/__init__.py` so its fingerprint warnings would carry file locations. Since that module runs before any submodule import, it became the first call, installed the root handler, and left the level at Python's default of WARNING — silently turning the project's original INFO request into a no-op.
- `git log -S basicConfig -- nemo_rl/__init__.py` returns exactly those two commits. No level was ever set and then removed to quiet something down.

So WARNING was never a decision, and restoring INFO restores the stated intent rather than changing policy. The knob sets the level on the `nemo_rl` logger rather than root, so raising it does not also enable debug output for torch, ray and httpx. Blast radius at INFO is ~50 call sites package-wide, concentrated in venv, cluster and vLLM-patch setup, so they land at startup rather than per step.

What is unchanged from the reply above: the worker's `[step]` summary is rank-0 only, and the per-chunk record stays at DEBUG behind `log.isEnabledFor`, since its arguments are evaluated at the call site whether or not the record is emitted and four of them are `.item()` host-device syncs.

Two things I checked against a real 68-node driver log rather than assuming:

1. **Worker logging records do reach `ray-driver.log`.** `(pid=2882750, ip=10.109.30.19) WARNING:root:__init__.py:195:` appears there — a record in this exact `basicConfig` format, emitted inside a Ray worker and forwarded with the pid prefix. That also confirms `__init__.py` executes in workers, which is what makes the new `setLevel` apply to them.
2. **`RAY_DEDUP_LOGS=1` does not eat per-step lines from a single actor.** In a 27-step job, the SC occupancy telemetry is present for every step 0–26, 403 lines, 12–17 per step. Dedup is active in that log (918 collapse markers) but is collapsing lines emitted identically across ranks — the `repeated 198x across cluster` counts track rank count.

One caveat now that these are records rather than prints: the handler writes to stderr while the surrounding pump progress lines print to stdout, and Ray captures those as separate streams, so their relative order in a driver log is not guaranteed. Noted in the code. If that proves annoying the clean fix is converting the file's remaining 23 prints to the logger, which is a separate change.

Tests moved with the mechanism: `TestStepSummaryIsEmitted` now asserts via `caplog` that the summary is logged at INFO from rank 0 and absent off-rank, and `TestChunkRecordIsGuarded` pins the chunk record silent at INFO and present under DEBUG.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1174,9 +1174,14 @@ def _split_step_state_init(
...
             "mb_losses": [],
             "total_num_microbatches": 0,
+            # One increment per train_microbatch call, i.e. the number of
+            # streaming chunks the controller has fed into this optimizer step
+            # so far.
+            "num_chunks": 0,
             # Saved across the step so we can restore at finish/abort.
             "saved_grad_sync_func": None,
             "saved_no_sync_func": None,
+            "saved_finalize_model_grads_func": None,
```
yuki-97, line 1189: The split-step orientation header at [1130-1141](https://github.com/NVIDIA-NeMo/RL/blob/e07e788aed1de2e45c9110f2ac029a0e7c864cf8/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1130-L1141) went stale with this change.

Item 1 still names `start_grad_sync` + `finish_grad_sync` as the reduce site, but that is now the relocated [`finalize_model_grads_func` call at 1599](https://github.com/NVIDIA-NeMo/RL/blob/e07e788aed1de2e45c9110f2ac029a0e7c864cf8/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1599) -- [1601](https://github.com/NVIDIA-NeMo/RL/blob/e07e788aed1de2e45c9110f2ac029a0e7c864cf8/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1601) is only the `else` fallback, which Megatron-Bridge makes unreachable. Item 2 lists one nulled hook while [`begin_train_step` at 1276](https://github.com/NVIDIA-NeMo/RL/blob/e07e788aed1de2e45c9110f2ac029a0e7c864cf8/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1276) now says three.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -2715,19 +2875,47 @@ def _broadcast_misc_params_packed(self, kv_scales=None) -> None:
...
+        """Put the model in eval mode for logprob inference.
+
+        Args:
+            keep_train_buffers: Leave the grad buffers and the optimizer state on
+                CUDA. Set this when a train step is already open. mcore's
+                ``_ParamAndGradBuffer.offload_to_cpu(move_grads=True)`` does not
+                copy gradients anywhere — it calls
+                ``grad_data.storage().resize_(0)``, freeing them — and the
+                matching ``reload_from_cpu`` resizes the storage back and
+                ``zero_()``s it. ``param.main_grad`` stays a valid view of that
```
yuki-97, line 2878: The docstring covers the offload but not the reload, which is where this is actually load-bearing.

[`prepare_for_training`](https://github.com/NVIDIA-NeMo/RL/blob/e07e788aed1de2e45c9110f2ac029a0e7c864cf8/nemo_rl/algorithms/single_controller.py#L589-L590) runs before every chunk -> mcore `reload_from_cpu` -> `grad_data.zero_()`, a no-op only because of the [`grad_data_size > 0` guard](https://github.com/NVIDIA/Megatron-LM/blob/d12f6c8c9aff51e166d872fd70151687a8e3f375/megatron/core/distributed/param_and_grad_buffer.py#L1653-L1656). Nothing in the PR pins that, so an mcore bump dropping the guard would silently reinstate the bug.

**Suggest fix:** name that invariant here.

Separately, take it or leave it: the comments run dense ([line 1546](https://github.com/NVIDIA-NeMo/RL/blob/e07e788aed1de2e45c9110f2ac029a0e7c864cf8/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1546-L1602) is 47 comment lines for 9 lines of code). Most of it is the reasoning that established correctness rather than invariants like the one above -- claude does a decent trimming pass if you want one.

nemo_rl/__init__.py
```diff
@@ -21,6 +21,24 @@
...
+# Set the level on the `nemo_rl` logger rather than on root: records still reach
+# the root handler, but raising the level does not also turn on debug output for
+# torch, ray, httpx and every other third-party logger.
+#
+# INFO by default. That is a change -- it surfaces the package's existing info
+# records, which are ~50 call sites concentrated in venv, cluster and vLLM-patch
+# setup, so they land at startup rather than per step. DEBUG additionally
+# enables the per-chunk and CUDA-allocator records on the training path, which
+# are per chunk per rank and are not meant for a normal run.
+logging.getLogger("nemo_rl").setLevel(os.environ.get("NRL_LOG_LEVEL", "INFO").upper())
```
yuki-97, line 40: `NRL_LOG_LEVEL` defaults to `INFO`, which flips the whole package from effectively-WARNING to INFO. That is a user-visible output change on every entrypoint, not just the SC path -- `basicConfig()` never set a level, so until this line no `nemo_rl` info record reached a run's output. The knob itself is not mentioned anywhere in `docs/`.

**Suggest fix:** document it. [`docs/debugging.md`](https://github.com/NVIDIA-NeMo/RL/blob/e07e788aed1de2e45c9110f2ac029a0e7c864cf8/docs/debugging.md) looks like the right home -- it is the page you land on when a run misbehaves, and `DEBUG` is what turns on the new per-chunk and allocator records. Something like:

```markdown
## Control Log Verbosity

NeMo RL sets the `nemo_rl` logger level from `NRL_LOG_LEVEL` (default `INFO`).
Set it to `DEBUG` to enable the per-chunk training and CUDA-allocator records,
which are emitted per chunk per rank and are not meant for a normal run:

    NRL_LOG_LEVEL=DEBUG uv run examples/run_grpo.py ...

The value must be a standard `logging` level name (`DEBUG`, `INFO`, `WARNING`,
`ERROR`, `CRITICAL`).
```

There is no central env-var page today -- [`docs/design-docs/env-vars.md`](https://github.com/NVIDIA-NeMo/RL/blob/e07e788aed1de2e45c9110f2ac029a0e7c864cf8/docs/design-docs/env-vars.md) only covers precedence and lists no `NRL_*` knobs -- so anywhere you think fits is fine.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1453,17 +1543,62 @@ def _finish_train_step_body(self, state: dict[str, Any]) -> dict[str, Any]:
...
         if self.cfg["megatron_cfg"]["distributed_data_parallel_config"][
             "overlap_grad_reduce"
         ]:
             self.model.start_grad_sync()
-        self.model.finish_grad_sync()
+        finalize_model_grads_func = state.get("saved_finalize_model_grads_func")
+        if finalize_model_grads_func is not None:
+            finalize_model_grads_func([self.model], None)
+        else:
+            self.model.finish_grad_sync()
```
yuki-97: SC will never come into the else branch, do you think we should remove it or `assert finalize_model_grads_func is not None` instead?

yuki-97 (commented):
thanks @saumishr , fix LGTM, left some minor comments.

nemo_rl/models/policy/workers/megatron_policy_worker.py
saumishr: Fixed in 4447ddf.

Item 1 now names the relocated `finalize_model_grads_func` call as the step's single reduce site, with `start_grad_sync` demoted to what it actually is — a precondition that only applies when `overlap_grad_reduce` is set.

Item 2 lists all three hooks with a one-line role each (`grad_sync_func` PP>1 scheduler, `no_sync_func` PP=1 last microbatch, `finalize_model_grads_func` end of every `forward_backward_func`) and then points at `begin_train_step` for the detail rather than restating it. That is also what keeps the two from drifting apart again: the header is now an index, and there is one place that explains each hook.

nemo_rl/models/policy/workers/megatron_policy_worker.py
saumishr: Named in 4447ddf.

The `keep_train_buffers` docstring now records the reload half explicitly: `prepare_for_training` runs before every chunk and reloads with `move_grads=True`, `reload_from_cpu` resizes and `zero_()`s the grad buffer only `if grad_data_size > 0`, and that counter is set only by a matching `offload_to_cpu` — so suppressing the offload keeps it at 0 and the reload is a no-op. It closes by saying that an mcore change dropping the guard, or zeroing unconditionally, would silently reinstate the bug.

Checked against the pinned mcore rather than taken on faith (`param_and_grad_buffer.py`): `offload_to_cpu` records `grad_data_size` then `resize_(0)`s, and `reload_from_cpu` resizes back, `zero_()`s, and resets the counter to 0. Your read is exactly right, including that this is the load-bearing side — the offload being suppressed is only half of why the gradients survive.

On the density: took the pass. The block you flagged is down from 47 comment lines to ~25, and the `begin_train_step` hook list lost a few more. What I dropped is the reasoning that established correctness rather than constraining a future reader — the `force_all_reduce` equivalence check, the `BucketGroup.finish_grad_sync` internals, and the `schedules.py` source quote. What I deliberately kept is the two by-omission arguments (`num_tokens=None`, no `pg_collection`), since those are the two things a well-meaning future edit would break by passing something, and the code cannot show why the absence matters.

nemo_rl/__init__.py
saumishr: Documented in 4447ddf, in `docs/debugging.md` as you suggested, using your draft as the base. Put it first on the page since raising the log level is the step-zero move when a run misbehaves, and updated the intro line that promised two scenarios.

Two things I added on top of your text, because a reader would otherwise have to discover them:

- the level applies to the `nemo_rl` logger rather than root, so `DEBUG` does not also turn on debug output for `torch`, `ray` and friends;
- setting it in the driver environment is enough to cover workers — `nemo_rl/__init__.py` runs in Ray workers too, and their records reach the driver log with the `(pid=..., ip=...)` prefix. I confirmed that on a real 68-node driver log rather than assuming it.

Left `docs/design-docs/env-vars.md` alone: it is precedence-only, with no `NRL_*` listing to extend, and inventing one there for a single knob seemed worse than one section on the page people actually land on.

nemo_rl/models/policy/workers/megatron_policy_worker.py
saumishr: Good catch. Went with the assert, in 4447ddf, and moved it above the `start_grad_sync` block so the invariant is checked before anything is dispatched.

Assert rather than plain removal because the two failure modes are not equally legible. With the `else` deleted, a `None` hook still crashes, but as a `TypeError: 'NoneType' object is not callable` at the end of a step, which says nothing about what is actually wrong. The message now names `_update_model_config_funcs`, so the failure points at setup instead of at the step that noticed.

And the fallback was worse than either: `finish_grad_sync` covers the cross-DP reduce and silently skips everything else the finalize owns — the non-TP/layernorm all-reduce across TP, the tied word- and position-embedding all-reduces across PP, the MoE router expert-bias update, `reset_model_temporary_tensors` — so it would have trained on wrong gradients without erroring.

Confirmed unreachable rather than assumed: Megatron-Bridge's `_update_model_config_funcs` binds `partial(finalize_model_grads, pg_collection=...)` under `if optimizer is not None` ([setup.py](https://github.com/NVIDIA/Megatron-Bridge/blob/main/src/megatron/bridge/training/setup.py)), we reach it from `finalize_megatron_setup`, and an open train step implies an optimizer because `_finish_train_step_body` steps it a few lines below.

Both tests that pinned the fallback moved with it: `test_grad_sync_falls_back_to_finish_grad_sync_without_finalize_hook` is now `test_grad_sync_refuses_to_reduce_without_the_finalize_hook`, still parametrized over `overlap_grad_reduce` and now asserting that neither `start_grad_sync` nor `finish_grad_sync` fires and the optimizer does not step; `test_handles_originally_none_hook` is now `test_originally_none_hook_fails_the_step`, which additionally pins that `finish_train_step`'s restore path still puts `grad_sync_func` back on the way out.

saumishr:
Thanks @yuki-97 — all four addressed in [4447ddf](https://github.com/NVIDIA-NeMo/RL/commit/4447ddfd4e556e61ca4e31b51c87a71bb85eeba7), with a reply on each thread.

- **Stale orientation header.** Item 1 now names the relocated `finalize_model_grads_func` call as the step's reduce site, `start_grad_sync` only as the `overlap_grad_reduce` precondition. Item 2 lists all three nulled hooks and defers to `begin_train_step` for the detail instead of restating it, so the two cannot drift apart again.
- **The `grad_data_size > 0` reload invariant.** Named in the `keep_train_buffers` docstring, along with the consequence: an mcore change dropping the guard, or zeroing unconditionally, silently reinstates the bug. You were right that the reload is the load-bearing half — suppressing the offload only matters because the matching reload is a no-op when nothing was offloaded.
- **`NRL_LOG_LEVEL`.** Documented in `docs/debugging.md` under a new "Control Log Verbosity" section, first on the page. Added that the level applies to the `nemo_rl` logger rather than root, and that setting it in the driver environment covers Ray workers.
- **The unreachable `else`.** Replaced with an assert, hoisted above the `start_grad_sync` block so nothing is dispatched before the check. A bare `finish_grad_sync` fallback was the worst of the three options — it reduces across DP and silently skips the TP layernorm all-reduce, the tied-embedding all-reduces across PP, the MoE expert-bias update and `reset_model_temporary_tensors`. Bridge binds the hook under `if optimizer is not None` and an open step implies an optimizer, so the branch really was dead; the assert message points at `_update_model_config_funcs` so a failure blames setup rather than the step that noticed. The two tests that pinned the fallback now pin the failure.

Also took the trimming pass you offered: the block you flagged is down from 47 comment lines to ~25. I dropped the reasoning that established correctness (the `force_all_reduce` equivalence, the `BucketGroup.finish_grad_sync` internals, the `schedules.py` source quote) and kept the two by-omission arguments, since those are what a future edit would break by helpfully passing something.

Still owed from the review body, unchanged: a real multi-chunk test (needs 1–2 GPUs; `test_train_pump.py` is the better target) and the before/after on `...single-controller-streaming2.yaml`. Happy to take either as a follow-up or fold in here.

Same caveat on this commit as the last: I could not run the mcore-marked tests locally — `megatron` is not importable on this host and the repo pins Python 3.13.14 against a 3.13.13 interpreter, so `uv` cannot build the env. `ruff format` and `ruff check` are clean and both changed files are `py_compile`-clean; CI is the first real run of the two rewritten tests.

mehraakash (approved):
LGTM, Tested!

terrykong (approved):

PR #3391, opened 2026-07-29
feat(dynamo): add managed Slurm generation with W&B telemetry

jthomson04:
## Summary

- add a managed-only Dynamo generation backend for sync and async GRPO on Ray-managed Slurm
- launch and own etcd, NATS JetStream, the Dynamo frontend, and a fixed Ray-scheduled vLLM worker fleet
- support NeMo-Gym token translation, NCCL weight refits, cache invalidation, deterministic cleanup, and W&B generation telemetry
- add a two-GPU smoke recipe plus a parameterized six-node Nemotron Nano SWE recipe

This PR is independent of #2990 and targets current `main`. It excludes Kubernetes, DGD, external Dynamo deployments, and the NeMo-Gym submodule update.

## Container

- derive from the standard NeMo-RL release image
- skip the NeMo-RL vLLM, SGLang, and TRT-LLM extras and actor environments
- install only `ai-dynamo[vllm]==1.3.0.post1` as the inference backend in isolated Python 3.12 `/opt/dynamo_venv`
- install fixed etcd 3.5.21 and NATS Server 2.11.6 binaries
- assert vLLM 0.23.0 and apply the vLLM PR #44814 layerwise-reload backport with an applicability check and marker

## Validation

- 294 focused unit tests across all changed test files
- full pre-commit suite, including ruff, format, pyrefly, and recipe minimization
- root and derived lock validation
- clean vLLM 0.23.0 patch applicability/apply/reverse check
- successful standard base and derived Dynamo image builds
- inspected the loaded base image, main environment, and every Ray actor environment: none contains vLLM, SGLang, TRT-LLM, or ai-dynamo
- inspected `/opt/dynamo_venv`: ai-dynamo 1.3.0.post1, vLLM 0.23.0, no SGLang/TRT-LLM, and the #44814 patch present in installed source
- shell syntax checks and dry runs for both Slurm launchers
- `L1_Functional_Tests_Dynamo` green on f270519, asserting `max(train/token_mult_prob_error) < 1.05`
- four-step 3n8g Nemotron Nano SWE1 run on H100: `train/token_mult_prob_error` median 1.0376 / final 1.0477, mean generation KL 0.003654, four refits and four cache invalidations, no leaked managed processes — full report in https://github.com/NVIDIA-NeMo/RL/pull/3391#discussion_r3808363649

PR #3661, opened 2026-08-16
fix(grpo): start collection after initial refit

al-rigazzi:
# What does this PR do ?

Start asynchronous trajectory collection only after the initial policy-to-generation refit or generation preparation has completed.

Previously, `AsyncTrajectoryCollector.start_collection()` could run before the initial policy weights had been transferred to the generation backend. This made rollout requests eligible to begin before generation was ready with the trained policy weights.

This PR restores the required startup order:

1. initialize the generation backend;
2. complete its initial refit or preparation;
3. start background trajectory collection;
4. set the collector's initial weight version.

The change affects only startup ordering. It does not alter the existing steady-state coordination or in-flight weight-update behavior after collection has started.

# Issues

None.

# Usage

No user-facing API or configuration changes are required.

**Pre checks**:
- [x] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests. (All tests related to this change passed.)
- [x] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs. (N/A: no public API or configuration change.)

# Additional Information

## Observed hang

The failure we observed was a hang in the initial policy-to-generation refit. It occurred at GBS256 in configurations with high rollout concurrency relative to global batch size.

However, all four GBS512 and GBS1024 controls completed successfully with the same pre-fix startup ordering. In each control, the initial refit started and both startup rollout workers were launched before refit completed. No complete prompt group was buffered before refit success, although the available logs cannot establish whether individual generation requests were already in flight. One GBS512 configuration also had enough concurrency to admit the complete two-target startup workload. A high concurrency-to-batch-size ratio is therefore not sufficient to trigger the hang; whether it is necessary was not established.

With `max_trajectory_age_steps=1`, the collector can reserve the current training step and one-step lookahead at startup. At GBS256, both step-sized rollout batches fit within the configured admission capacity. Smaller rollout payloads and lower admission backpressure may change batching, serialization, and request-routing timing enough to make a vulnerable interleaving more likely, but that mechanism was not independently isolated. Because collection starts asynchronously, calling `start_collection()` establishes eligibility for rollout work but does not prove that every generation worker was executing a rollout at the exact vulnerable instant.

When the failure occurred, policy ranks stopped making consistent progress through the initial refit's expert-parallel all-gather and pipeline-parallel broadcast operations. The NCCL watchdog eventually reported 600-second collective timeouts.

Sanitized representative excerpts:

```text
Watchdog caught collective operation timeout:
WorkNCCL(SeqNum=2928, OpType=ALLGATHER, ..., Timeout(ms)=600000)

Watchdog caught collective operation timeout:
WorkNCCL(SeqNum=36292, OpType=BROADCAST, ..., Timeout(ms)=600000)
```

These messages identify the collectives in which progress stopped; they do not, by themselves, establish why ranks diverged in collective progress.

The correctness requirement does not depend on how frequently the race reproduces: generation must receive or prepare the initial policy weights before rollout collection can begin. This PR makes that startup barrier explicit while leaving steady-state asynchronous collection and in-flight update behavior unchanged.

## Testing

All tests related to this change passed, including the new ordering regression test.

Testing for the affected asynchronous GRPO path included:

- added `test_async_grpo_refits_before_starting_trajectory_collection`, which fails if collection starts before the initial refit;
- ran the new regression test;
- ran related asynchronous GRPO tests;
- ran the complete `tests/unit/algorithms/test_grpo.py` module;
- exercised broader non-gated unit-test configurations where the available environment allowed it.

## Async behavior

The initial transition must be strict: when generation still needs the policy's initial weights, no rollout should start before refit or preparation completes.

After startup, the existing asynchronous behavior is unchanged. In particular, this PR does not change how steady-state refits coordinate new dispatches, already in-flight requests, weight versions, or KV-cache updates.

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -4351,9 +4351,7 @@ def async_grpo_train(
     # Start trajectory collection only after generation setup completes.
     # Starting it earlier can overlap rollout work with the initial
     # policy-to-generation refit and stall the refit collectives.
-    collection_task = trajectory_collector.start_collection.remote(  # noqa: F841
-        dataloader
-    )
+    collection_task = trajectory_collector.start_collection.remote(dataloader)
 
     # Ensure collector knows initial weight version
     trajectory_collector.set_weight_version.remote(weight_version)
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -4348,10 +4348,9 @@ def async_grpo_train(
             traceback.print_exc()
             return
 
-    # Start trajectory collection only after generation holds real weights.
-    # The engines come up with load_format=dummy (weights arrive via the refit
-    # above); collecting before the refit can race weight loading and produce
-    # rollouts from randomly initialized weights.
+    # Start trajectory collection only after generation setup completes.
+    # Starting it earlier can overlap rollout work with the initial
+    # policy-to-generation refit and stall the refit collectives.
     collection_task = trajectory_collector.start_collection.remote(  # noqa: F841
         dataloader
     )
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -4313,14 +4313,6 @@ def async_grpo_train(
         processor=processor,
     )
 
-    # Start trajectory collection in background
-    collection_task = trajectory_collector.start_collection.remote(dataloader)
-
-    # Ensure collector knows initial weight version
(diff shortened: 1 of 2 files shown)


al-rigazzi:
Closing as superseded by #3391. That PR independently introduced the required ordering: complete the initial policy-to-generation refit, synchronously publish the weight version, and only then start trajectory collection. The fix is now present on `main`.

PR #3589, opened 2026-08-12
feat(sc): contain rollout failures in the SingleController path

terrykong:
> [!NOTE]
> **Mirror of #3470**, originally opened by @asolergi-nv from the `asolergi-nv/RL` fork.
> The branch was re-pushed to `NVIDIA-NeMo/RL` unchanged (same head SHA) so the four parts
> can be linked as a proper PR stack — `gh stack` cannot operate on fork branches.
> **Please review and comment here.** The original #3470 is left open for history.

Part 1/4 of #3454

# What this fixes

Two failure modes in the SingleController async-GRPO path, both **silent**. Neither produces an exception, a log line, or an unhealthy Ray actor — which is why they have not been noticed.

**1. The GRPO path silently corrupts training.** `AsyncRolloutImpl` caught every exception from generation and committed the rollout anyway, with empty text and zero reward. Those rows enter the GRPO leave-one-out baseline as legitimate zero-reward samples, so a transient generation failure does not fail the run — it quietly biases the advantage estimate. This is a live data-correctness bug independent of any resiliency feature.

**2. The NeMo-Gym path silently wedges.** Gym's HTTP client is constructed with `ClientTimeout()` — every timeout disabled — and retries `ClientOSError` / `ServerDisconnectedError` in an uncapped loop. A dead vLLM endpoint therefore parks the rollout forever. Each parked rollout holds a `max_inflight_prompts` permit, so the rollout pump eventually blocks entirely while the train pump spins on an empty buffer. Ray reports every actor healthy throughout.

# What it does

- **Stops swallowing generation errors** (`rollout_manager.py`). The single highest-value change here: failures are classified and raised instead of being committed as zero-reward rows.
- **A failure taxonomy** separating infrastructure failures (retry elsewhere) from data failures (retrying will not help). HTTP is classified **by status, not exception type** — a `ClientResponseError` carrying 503 is infrastructure, one carrying 400 is data.
- **Deadlines** on rollouts, generation turns and environment steps, all defaulting to`None` (disabled).
- **Re-dispatch with two separate budgets.** A prompt that fails on infrastructure is re-dispatched to a different engine rather than dropped; a prompt that fails on its own data is retried a smaller number of times and then either fails the run or is skipped under an explicit budget. Two budgets because "the engine died" and "this prompt is malformed" deserve opposite treatment.
- **A watchdog** that publishes rollout counters and reports stalls — the failures that otherwise produce no signal at all.
[...]

PR #3665, opened 2026-08-16
feat(sc): tolerate and replace dropped rollouts in the SingleController

saumishr:
# What does this PR do ?

Make the SingleController path survive rollout failures: drop a prompt within budget instead of ending the run, close the affected training step short so an `in_order` sampler cannot hang on a group nobody is generating, and optionally replace the lost prompt so the step keeps its configured batch size.

# Issues

Builds directly on #3589, which added the failure taxonomy and per-prompt retry budgets but always ended the run once a budget was exhausted.

# Usage

All knobs live under `async_rl.rollout_failure`. Defaults are inert — a run that does not set these behaves exactly as it does on `main`.

```yaml
async_rl:
  rollout_failure:
    # Ride out a shard outage instead of losing the job: CONSECUTIVE prompts that may
    # exhaust their infra budget and be dropped. Any committed rollout resets the count,
    # so this bounds an outage rather than the run's lifetime. 0 (default) fails on the
    # first one, which is today's behaviour and v1's default.
    max_consecutive_dropped_prompts: 8

    # Smallest batch a step may close on, as a fraction of num_prompts_per_step. Neither
    # drop budget can bound this on its own, since both are run-scoped.
    min_step_batch_fraction: 0.9

    # "shrink" (default) trains the step on fewer groups, bounded by the floor above.
    # "replace" substitutes a fresh prompt so the step keeps its configured batch size.
    on_dropped_prompt: replace
    max_replacement_attempts: 1     # fresh prompts tried per lost group
    replacement_reserve_prompts: 1  # low-water mark for the spare-prompt pool
```

# Additional Information
Three commits, each standalone:

1. feat(sc): tolerate dropped rollouts instead of failing the run: max_consecutive_dropped_prompts, plus the shortfall credit. InOrderSampler stamps each batch with a target step and selects on an exact match, so a group that never arrives leaves that step permanently short and the train pump waits forever, turning a crash into a hang. The pump now credits the shortfall against the stamped step and closes it early. This also fixes the pre-existing max_skipped_prompts path, which could already strand an in-order step.

2. fix(sc): bound how short a dropped-prompt step may close: min_step_batch_fraction. Both drop budgets are run-scoped (the consecutive counter is cleared by any commit, including commits for other steps), so drops concentrated on one step could shrink it without limit. A fraction rather than a count, so it holds across batch sizes and protects small batches more strictly.

3. feat(sc): replace a dropped prompt so its step keeps its batch size — on_dropped_prompt, the spare-prompt pool, and borrowing.

# Design notes worth reviewer attention:

Only stamping samplers are affected. With weight_fifo or windowed, admit returns None and a step fills from whatever is ready, so a drop costs throughput but strands nothing. Nothing is diverted into the spare pool until the sampler has actually stamped a batch, so an unstamped run does not lose prompts to a pool it can never draw on.
The spare pool is refilled from the pump loop, by diverting one whole dataloader batch before admit(). The pump is the sole owner of the StatefulDataLoader, and advancing that iterator from a dispatch task which is where a drop is discovered, arbitrarily later would race the ordering that checkpoint/resume accounting depends on. Diverting before admit() also means a diverted batch never burns a target step.

Leftover spares are trained on. Once the dataloader is exhausted the pool is dispatched as ordinary steps, so the diverted batch is neither discarded nor silently subtracted from max_num_steps (which is derived from len(dataloader)). Partial pools are held back rather than dispatched as a step that is short by construction.
Borrowing is an optimization inside replace, not a mode. A fresh rollout puts the wait on the critical path — the trainer has closed the previous step and idles until the substitute commits. When a later step already holds a finished group, TQReplayBuffer.promote_ready_group re-stamps it into the dropped step (which then closes immediately) and the fresh prompt repays the lender, which is not due for another step. A borrow is only taken with a spare in hand: an unrepaid loan is the same hole one step later, walking forward to the final step, which has nobody to borrow from. Both paths hold the batch size, so selecting between them is not worth exposing; the two counters report which one ran.

promote_ready_group is deliberately synchronous. TQReplayBuffer.remove deletes its indices before its first await, so as long as the re-stamp does not yield, the index it picks cannot be shifted out from under it by a concurrent selection.

The fallback chain always terminates: borrow → fresh replacement → shrink, bounded by max_replacement_attempts and the pool. Without the shrink fallback, a step whose replacements keep failing would never close at all.

**Pre checks**:

### DIFF
--- a/nemo_rl/algorithms/async_utils/replay_buffer.py
+++ b/nemo_rl/algorithms/async_utils/replay_buffer.py
@@ -1085,6 +1085,51 @@ class TQReplayBuffer:
         """Return how many slots are stamped with ``target_step``."""
         return sum(1 for target in self.target_step_list if target == target_step)
 
+    def promote_ready_group(self, *, to_target_step: int) -> Optional[int]:
+        """Re-stamp a finished group from a later step so it lands in this one.
+
+        Fills a hole left by a dropped prompt with generation that is already done,
+        which is the point: the step closes immediately instead of waiting out a fresh
+        rollout. The step it was borrowed from is returned so the caller can repay it,
+        and the caller must -- an unrepaid loan is the same hole one step later, carried
+        forward until it reaches the last step, which has nobody to borrow from.
+
+        The furthest future step is preferred because it is due last and so has the most
+        slack to absorb the repayment. Only ready slots qualify: an unready one is a
+        reservation whose rollout is still running, so moving its stamp would hand this
+        step the same wait it is trying to avoid.
+
+        Promotion can only make a step fresher, never staler. Slots are appended in
+        dispatch order and the trainer version never decreases, so a group stamped for a
+        later step was generated at a weight version at least as new as the ones already
+        in this step.
+
+        Synchronous on purpose. ``remove`` deletes its indices before its own first
+        await, so as long as nothing here yields, the index picked below cannot be
+        shifted out from under the write by a selection running concurrently.
+
+        Args:
+            to_target_step: Training step to re-stamp the borrowed group onto -- the
+                step that lost a prompt. Must be at or ahead of the trainer version:
+                a group re-stamped onto a step already trained is never selectable
+                again and would only be evicted.
+
+        Returns:
+            The target step the group was taken from, or None when no later step has a
+            ready group to lend.
+        """
+        lender_idx: Optional[int] = None
+        lender_target: Optional[int] = None
+        for i, target in enumerate(self.target_step_list):
+            if target is None or target <= to_target_step or not self.ready_list[i]:
+                continue
+            if lender_target is None or target > lender_target:
+                lender_idx, lender_target = i, target
+        if lender_idx is None:
+            return None
+        self.target_step_list[lender_idx] = to_target_step
+        return lender_target
+
     def size(self) -> int:
         """Return the number of prompt-group entries currently held."""
         return len(self.meta_list)
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -36,10 +36,12 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import math
 import os
 import time
+from collections import deque
 from functools import partial
-from typing import Any, Optional, Union, cast
+from typing import Any, Awaitable, Callable, Optional, Union, cast
 
 import ray
 import torch
@@ -211,6 +213,34 @@ class SingleControllerActor:
 
         self._inflight_by_group_id: dict[str, tuple[asyncio.Task[None], int]] = {}
 
+        # Groups that will never arrive, keyed by the training step they were stamped
+        # for. A sampler that matches batches to steps exactly (InOrderSampler) can only
+        # ever select num_prompts_per_step groups carrying that stamp, so a dropped
+        # prompt leaves that step permanently one short and the train pump waits on a
+        # group no one is generating. The pump subtracts these to close the step short.
+        # Only stamped prompts appear here: with an unstamped sampler the batch fills
+        # from whatever is ready, so a drop costs throughput but strands nothing.
+        self._batch_shortfall: dict[int, int] = {}
+
+        # Spare prompts that on_dropped_prompt="replace" substitutes for dropped ones,
+        # and the per-step counts of how each step got made whole (logged so a step's
+        # batch size stays explainable after the fact). The reserve is filled only by the
+        # rollout pump, which owns the dataloader iterator; see the config docstring for
+        # why a dispatch task cannot pull from it directly.
+        #
+        # The two counters read from opposite ends of a borrow: a step that filled a
+        # hole with a later step's finished group counts a promotion, and the step it
+        # borrowed from counts the replacement that repaid it.
+        self._replacement_reserve: deque[DatumSpec] = deque()
+        self._batch_replacements: dict[int, int] = {}
+        self._batch_promotions: dict[int, int] = {}
+        # Whether the sampler has ever handed back a target step. Only stamped prompts
+        # can strand a step, so this gates the pool: filling it for a sampler that never
+        # stamps would divert a batch of prompts that nothing is ever able to draw on.
+        # Learned from admit rather than the sampler's type, because a custom sampler's
+        # stamping is not knowable until it answers.
+        self._sampler_stamps_target_steps: bool = False
+
         # Backpressure valve: max unconsumed rollout groups allowed in DataPlane.
         # Acquired before each rollout dispatch; released when the buffer
         # drops a group (sampler.evict or post-train buffer.remove).
@@ -244,6 +274,7 @@ class SingleControllerActor:
         await self._sync_weights()
 
         await self._maybe_restore_replay_buffer()
+        await self._maybe_restore_replacement_reserve()
 
         # Start the rollout and train pumps, plus the watchdog
         rollout_task = asyncio.create_task(self._rollout_pump())
@@ -352,6 +383,42 @@ class SingleControllerActor:
         for _ in range(restored):
             await self._buffer_capacity.acquire()
 
+    async def _maybe_restore_replacement_reserve(self) -> None:
+        """Restore spare prompts diverted before the previous run's checkpoint.
+
+        These were pulled from the dataloader and never dispatched, so the restored
+        dataloader resumes past them. Nothing else in the checkpoint holds them, and
+        without this they are simply gone: one batch of the dataset per divert, plus
+        the training step ``_clamp_max_num_steps`` had budgeted for it.
+
+        No sampler-name guard, unlike the buffer restore. Spares carry no stamp -- they
+        are prompts that never reached ``admit`` -- so nothing about them depends on
+        which sampler wrote the checkpoint. They are restored even into a run that has
+        since switched to ``on_dropped_prompt="shrink"``, where the pool is never drawn
+        on but is still drained back into training at the end of the dataloader.
+        """
+        if self._last_checkpoint_path is None:
+            return
+        reserve_path = os.path.join(
+            self._last_checkpoint_path, "replacement_reserve.pt"
+        )
+        # Absent for every run that never diverted a batch, which is every run that
+        # does not use "replace" -- so silence here rather than the buffer restore's
+        # warning, since this is the ordinary case rather than a lost artifact.
+        if not os.path.exists(reserve_path):
+            return
+        # weights_only=False: spares are pickled DatumSpecs, and the checkpoint is a
+        # trusted same-job artifact (the replay buffer restore loads on the same terms).
+        reserve_state = await asyncio.to_thread(
+            torch.load, reserve_path, weights_only=False
+        )
+        self._replacement_reserve.extend(reserve_state)
+        print(
+            f"📦 Restored {len(reserve_state)} pooled spare prompt(s) from checkpoint: "
+            f"{reserve_path}",
+            flush=True,
+        )
+
     async def _ray_get(self, obj_ref: Any) -> Any:
         """Await a Ray ObjectRef without blocking the asyncio event loop."""
         return await obj_ref
@@ -373,8 +440,10 @@ class SingleControllerActor:
         """Continuously dispatch rollout tasks until cancellation.
 
         Per batch:
-          0. await sampler.admit(...) to wait until the batch may dispatch and
-             obtain its target_step stamp.
+          0. Under on_dropped_prompt="replace", divert the batch into the spare pool
+             if the pool is below its low-water mark, and skip admission entirely.
+             Otherwise await sampler.admit(...) to wait until the batch may dispatch
+             and obtain its target_step stamp.
 
         Per prompt:
           1. Acquire _buffer_capacity slot (backpressure)
@@ -383,7 +452,14 @@ class SingleControllerActor:
           4. Call rollout_manager.generate_and_push(prompt) — local async
              RolloutManager reserves a slot, runs the rollout, then commits the
              group via TQReplayBuffer (→ dp_client.put_samples + mark ready)
-          5. Decrement _inflight_rollouts
+          5. If the prompt was dropped, substitute a spare and repeat step 4 -- for this
+             step, or for whichever step lends this one a finished group in its place
+             (see _take_replacement, _promote_into_step) -- or credit the step short so
+             the train pump can close it
+          6. Decrement _inflight_rollouts
+
+        Once every epoch is done, whatever is left in the spare pool is dispatched as
+        ordinary steps rather than discarded (see _drain_reserve_into_steps).
         """
         sem = asyncio.Semaphore(self._async_cfg.max_inflight_prompts)
         self._rollout_exhausted.clear()
@@ -396,26 +472,69 @@ class SingleControllerActor:
         ) -> None:
             task_started_event.set()
             self._inflight_rollouts += 1
+            # This task owns one slot of a step, which can outlive both the prompt it
+            # started with and the step it started on: a dropped prompt is substituted in
+            # place and the loop runs again, and the slot is re-aimed at whichever step
+            # lends this one a finished group. Both permits are held across
+            # substitutions because the slot stays occupied either way -- they are
+            # released once something commits, or once a step is credited short.
+            replacements = 0
             try:
-                outcome = await self._rollout_manager.generate_and_push(
-                    prompt,
-                    target_step=target_step,
-                    inflight_registry=self._inflight_by_group_id,
-                )
-            except BaseException:
-                # On success ownership transfers to the train pump, which
-                # releases this permit after consuming the committed group.
-                self._buffer_capacity.release()
-                raise
+                while True:
+                    try:
+                        outcome = await self._rollout_manager.generate_and_push(
+                            prompt,
+                            target_step=target_step,
+                            inflight_registry=self._inflight_by_group_id,
+                        )
+                    except BaseException:
+                        # On success ownership transfers to the train pump, which
+                        # releases this permit after consuming the committed group.
+                        self._buffer_capacity.release()
+                        raise
+
+                    if outcome is not RolloutOutcome.SKIPPED:
+                        break
+
+                    replacement = self._take_replacement(target_step, replacements)
+                    if replacement is None:
+                        # Nothing was committed, so the train pump will never see this
+                        # group and never release its permit on our behalf.
+                        self._buffer_capacity.release()
+                        self._credit_shortfall(target_step)
+                        return
+
+                    replacements += 1
+                    prompt = replacement
+                    print(
+                        f"  target_step={target_step}: substituting a spare prompt for "
+                        f"the dropped group (replacement {replacements}/"
+                        f"{self._async_cfg.rollout_failure.max_replacement_attempts}, "
+                        f"{len(self._replacement_reserve)} spare(s) left)",
+                        flush=True,
+                    )
+                    # Attempted only now that a spare is in hand, because the borrow is a
+                    # debt and the spare is what repays it. Borrowing without one would
+                    # leave the lender short instead: the same hole, one step later.
+                    lender_step = self._promote_into_step(target_step)
+                    if lender_step is not None:
+                        target_step = lender_step
+                    # A substitution is a fresh rollout, not a continuation of the one
+                    # that failed, so it observes the same pause a first dispatch does
+                    # instead of pushing new generation into a weight-sync window.
+                    await self._rollout_permitted.wait()
             finally:
                 self._inflight_rollouts -= 1
                 sem.release()
 
-            if outcome is RolloutOutcome.SKIPPED:
-                # Nothing was committed, so the train pump will never see this group
-                # and never release its permit on our behalf.
-                self._buffer_capacity.release()
-                return
+            if replacements and target_step is not None:
+                # Counted per slot, not per attempt: a step got its group back, which is
+                # the fact that explains why its batch is full despite a drop. Recorded
+                # against the step the spare actually committed to, which after a borrow
+                # is the lender rather than the step that was dropped from.
+                self._batch_replacements[target_step] = (
+                    self._batch_replacements.get(target_step, 0) + 1
+                )
 
             if self._async_cfg.diagnostics:
                 content = ""
@@ -434,13 +553,40 @@ class SingleControllerActor:
                 self._buffer_capacity.release()
                 sem.release()
 
+        async def _launch(prompt: DatumSpec, target_step: Optional[int]) -> None:
+            # check if buffer is full
+            await self._buffer_capacity.acquire()
+            # check if inflight rollouts is full
+            await sem.acquire()
+            # wait for rollout to be permitted
+            await self._rollout_permitted.wait()
+
+            task_started_event = asyncio.Event()
+            # dispatch rollout
+            task = rollout_tasks.create_task(
+                _dispatch_one_prompt(prompt, target_step, task_started_event)
+            )
+            self._dispatched_rollouts.add(task)
+            task.add_done_callback(self._dispatched_rollouts.discard)
+            task.add_done_callback(
+                partial(
+                    _release_permits_if_task_not_started,
+                    task_started_event=task_started_event,
+                )
+            )
+
         max_epochs = self._master_config.grpo.max_num_epochs
         async with asyncio.TaskGroup() as rollout_tasks:
             while max_epochs is None or self._current_epoch < max_epochs:
                 for prompt_batch in self._dataloader:
+                    if self._divert_batch_to_reserve(prompt_batch):
+                        continue
+
                     target_step = await self._sampler.admit(
                         trainer_version_fn=lambda: self._trainer_version
                     )
+                    if target_step is not None:
+                        self._sampler_stamps_target_steps = True
 
                     num_prompts = prompt_batch.size
                     if target_step is not None:
@@ -458,31 +604,17 @@ class SingleControllerActor:
                         prompt: DatumSpec = {  # type: ignore
                             k: v[prompt_idx] for k, v in prompt_batch.items()
                         }
-
-                        # check if buffer is full
-                        await self._buffer_capacity.acquire()
-                        # check if inflight rollouts is full
-                        await sem.acquire()
-                        # wait for rollout to be permitted
-                        await self._rollout_permitted.wait()
-
-                        task_started_event = asyncio.Event()
-                        # dispatch rollout
-                        task = rollout_tasks.create_task(
-                            _dispatch_one_prompt(
-                                prompt, target_step, task_started_event
-                            )
-                        )
-                        self._dispatched_rollouts.add(task)
-                        task.add_done_callback(self._dispatched_rollouts.discard)
-                        task.add_done_callback(
-                            partial(
-                                _release_permits_if_task_not_started,
-                                task_started_event=task_started_event,
-                            )
-                        )
+                        await _launch(prompt, target_step)
 
                 self._current_epoch += 1
+
+        # Only now that every dispatched rollout has settled is the pool genuinely
+        # spare. Draining it inside the group above would race them for it, and a
+        # rollout that was about to be dropped has the better claim: it needs a spare to
+        # keep its step whole, whereas an extra step is only worth having if one is
+        # left over. A second group because the first is closed to new tasks.
+        async with asyncio.TaskGroup() as rollout_tasks:
+            await self._drain_reserve_into_steps(_launch)
 
         # Drain in-flight so return implies "all rollouts in TQ".
         inflight = list(self._dispatched_rollouts)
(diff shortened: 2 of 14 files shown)


rohitrango:
this feature can potentially be combined/merged with allowing partial rollout groups depending on an eviction countdown and `time(latest rollout)/time(p50)` factor

saumishr:
@rohitrango agreed the decision point is shared: a straggler that trips `rollout_timeout_s` already ends up in this PR's drop/replace path, and a p50 ratio is a better trigger than a fixed wall clock, since it retunes itself as generations get longer over a run.

The delta is granularity. Partial groups need a variable N, which the replay buffer forbids on restore today ("a changed group size silently breaks the group-relative baseline"), and `min_step_batch_fraction` bounds groups per step, not rows per group. Worth noting too that evicting the slowest rows preferentially discards the longest generations on *every* step, not just failing ones, so it carries a length bias this PR's failure-driven drops do not.

Both seem like follow-up work on top of this rather than part of it — the drop path here is the natural place to hang the eviction decision once groups can be short.

nemo_rl/algorithms/single_controller.py
```diff
@@ -510,7 +819,8 @@ async def _train_pump(self) -> None:
 
                         # Select a batch
                         max_prompt_groups = (
-                            grpo_cfg.num_prompts_per_step - groups_dispatched
+                            self._target_groups_for_step(version_during_step)
+                            - groups_dispatched
                         )
```
saumishr: [`single_controller.py:822`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L822)

**1 action item.**

**TL;DR — a shortfall credited between L800 and L822 drives `max_prompt_groups` to 0 and kills the train pump with `ValueError: min_prompt_groups must be >= 1, got 0`, turning a tolerated drop back into the run failure this PR exists to remove.**

PR-introduced. The target is read twice per iteration — the loop bound at [L800](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L800) and again here — with `await asyncio.sleep(0)` at [L805](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L805) and `await self._sampler.evict(...)` at [L808](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L808) in between, while `groups_dispatched` is not updated until [L938](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L938). On `main` both expressions were the constant `num_prompts_per_step`, so `max_prompt_groups >= 1` was an invariant. It no longer is.

With `in_order`, `num_prompts_per_step: 128`, `min_groups_for_streaming_train: 32`, `max_consecutive_dropped_prompts: 8`:

1. 127 groups have committed and been selected while one prompt burns its infra budget, so the pump sits at `groups_dispatched=127`, target 128, spinning on the retry sleep at [L854](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L854).
2. L800 reads 128, enters the loop, and L805 yields.
3. The drop resolves inside that yield — `_credit_shortfall(0)` sets `_batch_shortfall[0] = 1`.
4. L822 now returns 127, so `max_prompt_groups = 0` and `min_prompt_groups = min(32, 0) = 0`.
5. [`_validate_group_bounds`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/async_utils/staleness_sampler.py#L208) raises `ValueError: min_prompt_groups must be >= 1, got 0`, which propagates out of `_train_pump` and ends the run.

`groups_dispatched == target - 1` is the *typical* state when a drop resolves, since a drop only lands after the infra budget and its backoff are spent. The SC exemplar is immune because it sets `min_groups_for_streaming_train: ${grpo.num_prompts_per_step}`, but the code default is 32 and streaming on a partial batch is the feature's whole point.

**Action: clamp the second read.** Keep it inside the loop body — hoisting it above the `while` would reintroduce exactly the hang commit 1 fixes.

```suggestion
                        max_prompt_groups = (
                            self._target_groups_for_step(version_during_step)
                            - groups_dispatched
                        )
                        # A prompt stamped for this step can be dropped while the pump
                        # is already inside this iteration, so the target can fall to
                        # what has already been dispatched. Close the step rather than
                        # asking the sampler for a non-positive batch.
                        if max_prompt_groups <= 0:
                            break
```

While you are here: the same shrunk target does not reach [L850](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L850), where the rollout-exhausted error still divides by `num_prompts_per_step` and so reports `dispatched 29/32` for a step whose target was legitimately 29. Reuse the local rather than calling `_target_groups_for_step` again there — it can raise, and raising inside the construction of an error message would swap the real error for a different one.

nemo_rl/algorithms/single_controller.py
```diff
@@ -466,6 +561,215 @@ def _release_permits_if_task_not_started(
...
+    ) -> None:
+        """Train on the leftover spares once the dataloader has nothing more to give.
+
+        Spares were consumed from the dataset like any other prompt, so leaving them in
+        the pool at the end of the last epoch throws away data the run already paid for.
+
+        It also restores the step count. ``_clamp_max_num_steps`` derives
+        ``max_num_steps`` from ``len(dataloader)``, and every diverted batch is one
+        fewer batch the loop can admit -- so without this a replace-mode run quietly
+        finishes one step short of the budget it was configured with, per divert.
```
saumishr, line 680: [`single_controller.py:614`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L614)

**1 action item.**

**TL;DR — the spare pool is not checkpointed while the dataloader position is, so a resumed `replace` run silently loses one batch and this docstring's guarantee does not survive the resume.**

PR-introduced. `_save_checkpoint` writes `save_state`, `train_dataloader.pt` and `replay_buffer.pt`, and never `_replacement_reserve`. The divert has already advanced the iterator by the time [`self._dataloader.state_dict()`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L1246) is captured, so the saved position is *past* prompts that exist only in memory.

1. The pump diverts batch 1 into `_replacement_reserve`; the dataloader records it as consumed.
2. A checkpoint is taken. The pool is not in it.
3. The job is preempted before `_drain_reserve_into_steps` runs — it only runs once the dataloader is exhausted.
4. On resume `_replacement_reserve` is empty and the dataloader restarts at batch 2. Those `num_prompts_per_step` prompts are gone, and `max_num_steps` is still clamped to `max_num_epochs * len(dataloader)`.

The likelier of the two symptoms is the silent one: at the end `groups_dispatched == 0` with an empty buffer, so the pump returns cleanly and the run just finishes short of the configured step count. The `RuntimeError` only appears if partial groups happen to be buffered.

**Action: persist `_replacement_reserve` alongside `replay_buffer.pt` in `_save_checkpoint` and restore it in `__init__`** — draining it back before the snapshot is not available, since the prompts have already left the iterator. If that is out of scope, narrow this docstring's claim to a single process, because as written it promises something a resumed run does not deliver. Either way `tests/unit/single_controller/test_sc_checkpointing.py` is the natural home for a round-trip case once the semantics are picked.

Severity is bounded — it needs `replace` plus checkpointing plus an actual resume — but that is precisely the shape of a long convergence chain.

tests/unit/single_controller/test_single_controller.py
```diff
@@ -374,6 +377,9 @@ def _train_pump_controller(*, sampler) -> object:
     ctrl._timer = Timer()
     ctrl._trainer_version = 0
     ctrl._train_steps = 0
+    ctrl._batch_shortfall = {}
```
saumishr, line 418: [`test_single_controller.py:380`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/tests/unit/single_controller/test_single_controller.py#L380)

**1 action item.**

**TL;DR — no test drives `_train_pump` with a non-empty `_batch_shortfall`, so the hang this PR exists to prevent is covered only as two disconnected halves, and the three new metrics are unasserted.**

PR-introduced. `_batch_shortfall` appears in tests on the credit side (`test_rollout_pump.py`) and as the `{}` initialisation here, never populated. So one test proves the credit is written and another proves `_target_groups_for_step` subtracts it, but nothing proves the pump *consumes* it. Concretely: hoist [L800](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L800)'s call into a variable above the `while` and the entire suite stays green while the run hangs exactly as it does on `main` — which is the regression the comment at [L797](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L797) is warning against.

The same test closes a second gap. `dropped_prompt_groups`, `replaced_prompt_groups` and `promoted_prompt_groups` are asserted nowhere, and [`self._trainer_version += 1`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L949) runs *before* the three reads at [L951-959](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller.py#L951) — so reading `_trainer_version` instead of `version_during_step` there would zero all three counters for every step, silently and forever.

**Action: add a `_train_pump` test reusing the `_train_pump_controller` harness**, with a sampler fake that credits `_batch_shortfall[0] = 1` on its second `select` and `_rollout_exhausted` cleared, asserting the step closes on the groups it got and that the three metrics carry the expected counts. Two things the harness does not give you: `num_prompts_per_step=2` against the default fraction makes `floor = ceil(0.9 * 2) = 2`, so lower `min_step_batch_fraction` or no short step is legal; and a step that completes also needs `_sync_weights`, `_logger` and the `ray.cluster_resources` patch, as `test_train_pump_logs_nonzero_stale_group_metrics` already sets up.

A second case starting at `_trainer_version = 3` with `{2: 1, 3: 1, 5: 1}` pruning to `{5: 1}` would cover the older-than-current half of the prune, which a version-0 test cannot reach.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -368,6 +368,7 @@ def _build_retry_policy(master_config: MasterConfig) -> RolloutRetryPolicy:
         backoff_base_s=failure_config.backoff_base_s,
         max_backoff_s=failure_config.max_backoff_s,
         max_skipped_prompts=failure_config.max_skipped_prompts,
+        max_consecutive_dropped_prompts=failure_config.max_consecutive_dropped_prompts,
```
saumishr, line 501: [`setup.py:371`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller_utils/setup.py#L371)

**1 action item.**

PR-introduced. `_build_retry_policy` has no test at all, and this line is the single translation point for the headline knob — the other four new fields are read straight off `_async_cfg.rollout_failure` by the controller. The two sides are each covered in isolation and never joined: `test_resiliency_config.py` asserts the field parses, and every test in `TestInfraDropTolerance` constructs `RolloutRetryPolicy` by hand.

Delete this line and every test still passes, while a user who sets `max_consecutive_dropped_prompts: 8` silently gets `main`'s behaviour and loses the job to the first dropped prompt — the exact outcome the PR is titled after.

**Action: add a wiring test to `tests/unit/single_controller/test_resiliency_config.py`** asserting `_build_retry_policy(cfg).max_consecutive_dropped_prompts == 8` for a config that sets it, and `== 0` for the default. `MasterConfig.model_construct(async_rl=AsyncRLConfig(...))` is enough — the function reads nothing else — but pass a real `AsyncRLConfig` instance rather than a dict, since `model_construct` skips coercion.

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -123,6 +204,25 @@ def _check_consistent(self) -> "RolloutFailureConfig":
                 f"async_rl.rollout_failure.max_backoff_s ({self.max_backoff_s}) must be "
                 f">= backoff_base_s ({self.backoff_base_s})"
             )
+        # Both of these would leave on_dropped_prompt="replace" configured but unable to
+        # ever produce a replacement, so it would quietly behave as "shrink". Rejected
+        # rather than tolerated: the whole point of asking for "replace" is the batch
+        # size guarantee, and losing it silently is the failure mode worth preventing.
+        if self.on_dropped_prompt == "replace":
```
saumishr, line 214: [`config.py:211`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/single_controller_utils/config.py#L211)

**1 action item.**

**TL;DR — three more knob combinations are inert or self-defeating in exactly the way this validator was written to catch, and all three currently surface only at runtime.**

PR-introduced. This block rejects the two ways `replace` can be configured but unable to produce a replacement. Three siblings get through:

- **The floor can swallow both drop budgets.** Nothing compares `min_step_batch_fraction` against `grpo.num_prompts_per_step`, so `ceil(fraction * N) == N` makes every drop fatal while setup passes — true at `N=8` with the 0.9 default, and at *any* N with `1.0`, which `le=1.0` accepts. An operator who set `max_consecutive_dropped_prompts: 8` to ride out an outage gets a run-ending `RuntimeError` on the first drop, hours in.
- **`replace` with a non-stamping sampler never stamps**, so `_divert_batch_to_reserve` never fires, the pool stays empty, and `_take_replacement` returns `None` on its `target_step is None` branch every time. `replaced_prompt_groups` sits at 0 with nothing distinguishing "nothing failed" from "the mode you asked for was never live".
- **`replacement_reserve_prompts` is unbounded.** Above one batch the pump diverts consecutive batches, and diverting happens before `admit()` — outside the sampler gate, `_buffer_capacity` and `sem` — so nothing is dispatched until the mark is reached.

**Action: warn at setup rather than raise, in `validate_single_controller_config`** (`sampler` and `rollout_failure` are siblings under `AsyncRLConfig`, so one validator reaches both). Raising is wrong for at least the first two: `replace` + `min_step_batch_fraction: 1.0` + a drop budget is the most coherent strict configuration there is — tolerate infra drops, never train a short batch, substitute instead — and a hard raise would forbid it. Same for one base YAML carrying `replace` with `--override async_rl.sampler.name=windowed` for an experiment. A bound on `replacement_reserve_prompts` is the one of the three that could reasonably be a raise, since above `num_prompts_per_step` it contradicts the field's own docstring.

nemo_rl/experience/rollout_manager.py
```diff
@@ -1294,19 +1332,45 @@ async def generate_and_push(
...
+        self._stats.record_infra_drop(reason, self._consecutive_infra_drops)
+        print(
+            f"dropping prompt idx={input_sample['idx']} after {infra_attempts} "
+            f"infrastructure failure(s) ({reason}: {last_infra_error}) "
+            f"[consecutive drop {self._consecutive_infra_drops}/"
+            f"{policy.max_consecutive_dropped_prompts}]",
+            flush=True,
+        )
+        self._stats.skipped += 1
+        return RolloutOutcome.SKIPPED
```
saumishr, line 1378: [`rollout_manager.py:1376`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/experience/rollout_manager.py#L1376)

**1 action item.**

PR-introduced. This is the second cause of `SKIPPED`, but the member's defining comment at [L65-67](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/experience/rollout_manager.py#L65) still names only the first — "The prompt exhausted its **data**-failure budget within max_skipped_prompts". You updated the `Returns:` block in this same hunk set to cover both; the enum comment three lines from the member was left behind, and it is the only documentation this seam has.

**Action: update the `SKIPPED` comment in `RolloutOutcome`** to something like:

```python
    # The prompt was given up on within a budget: its data budget within
    # max_skipped_prompts, or its infra budget within
    # max_consecutive_dropped_prompts. No group was committed, so the caller owns
    # releasing its backpressure permit and crediting the step's shortfall.
    SKIPPED = "skipped"
```

tests/unit/single_controller/test_tq_replay_buffer.py
```diff
@@ -440,6 +440,71 @@ def test_count_for_target_step_includes_reserved_slots(self):
         assert buf.count_for_target_step(7) == 0
 
 
+class TestTQReplayBufferPromote:
+    """on_dropped_prompt="promote" fills a step from a later step's finished work."""
```
saumishr: [`test_tq_replay_buffer.py:444`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/tests/unit/single_controller/test_tq_replay_buffer.py#L444)

**1 action item.**

PR-introduced. This documents `on_dropped_prompt="promote"`, but the field is `Literal["shrink", "replace"]` and `test_resiliency_config.py` explicitly asserts that `"promote"` is rejected — the PR's own design note is that borrowing is an optimization inside `replace`, not a mode. A reader chasing the borrow path from here is pointed at a config that raises.

```suggestion
    """Borrowing inside on_dropped_prompt="replace": a step is filled from a later
    step's already-finished work."""
```

nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -1085,6 +1085,45 @@ def count_for_target_step(self, target_step: int) -> int:
...
+        later step was generated at a weight version at least as new as the ones already
+        in this step.
+
+        Synchronous on purpose. ``remove`` deletes its indices before its own first
+        await, so as long as nothing here yields, the index picked below cannot be
+        shifted out from under the write by a selection running concurrently.
+
+        Returns:
+            The target step the group was taken from, or None when no later step has a
+            ready group to lend.
```
saumishr, line 1119: [`replay_buffer.py:1111`](https://github.com/NVIDIA-NeMo/RL/blob/f0a059c14504fc4225bebb79f99c3019cc6c757c/nemo_rl/algorithms/async_utils/replay_buffer.py#L1111)

**1 action item.**

PR-introduced. New public method on a cross-file interface (called from `single_controller.py:703`), and the only one of its parameter-bearing siblings in this class without an `Args:` section — `reserve`, `commit`, `remove`, `remove_group`, `state_dict` and `load_state_dict` all have one.

```suggestion
        Args:
            to_target_step: Training step to re-stamp the borrowed group onto -- the
                step that lost a prompt. Must be at or ahead of the trainer version;
                a group re-stamped onto a step already trained would only be evicted.

        Returns:
            The target step the group was taken from, or None when no later step has a
            ready group to lend.
```

saumishr (commented):
Team review of this PR: four analysis agents (RL/guidelines, bug hunting, tests, design) followed by two independent adversarial passes over every finding. Rebased onto `main` at `42f449161` first; the diff applies clean with no file overlap.

**Two of the eight are worth acting on before merge**: a torn read of the step target that can crash the train pump with `ValueError: min_prompt_groups must be >= 1, got 0` — converting a tolerated drop back into a run failure — and a spare pool that is not checkpointed while the dataloader position is, so a resumed `replace` run silently loses a batch. The rest are test gaps and one-line text fixes.

**Verified rather than assumed**, since the PR makes these claims explicitly:

- `promote_ready_group` really does contain no yield point, and `remove` really does delete its indices before its first await. Every other mutator of those indices was checked too: `reserve`, `commit` and `load_state_dict` only append or re-resolve `idx` by `group_id` after their awaits, so the re-stamp cannot have its index shifted out from under it. The docstring's contract holds as written.
- The `borrow -> replace -> shrink` chain always terminates. Every exit either commits, raises, or credits a shortfall; no path leaves a step neither closing nor aborting.
- The default path is a genuine no-op. With both budgets at 0 neither `SKIPPED` return is reachable, so the floor, the shortfall credit and all three reserve helpers are unreachable, and the added cost is one dict lookup and a `ceil` per pump iteration. No convergence A/B is warranted for this PR, which is why there is no `[PERF-EVIDENCE]` ask.

**Findings that did not survive the adversarial pass**, noted because they look plausible and may come up again: that `_take_replacement` needs the `target_step < self._trainer_version` guard `_promote_into_step` has, and that promotion can target a step that has closed but not yet advanced the trainer version. Both die on the same invariant — at loop exit `groups_dispatched + shortfall == N` exactly, so no slot can still be in flight against a closed step, and no ready group remains stamped for one. Adding that guard would be dead code.

_Generated by Claude Code_

saumishr:
> this feature can potentially be combined/merged with allowing partial rollout groups depending on an eviction countdown and `time(latest rollout)/time(p50)` factor

@rohitrango agreed the decision point is shared: a straggler that trips rollout_timeout_s already ends up in this PR's drop/replace path, and a p50 ratio is a better trigger than a fixed wall clock, since it retunes itself as generations get longer over a run.

The delta is granularity. Partial groups need a variable N, which the replay buffer forbids on restore today ("a changed group size silently breaks the group-relative baseline"), and min_step_batch_fraction bounds groups per step, not rows per group. Worth noting too that evicting the slowest rows preferentially discards the longest generations on every step, not just failing ones, so it carries a length bias this PR's failure-driven drops do not.

Both seem like follow-up work on top of this rather than part of it — the drop path here is the natural place to hang the eviction decision once groups can be short.

nemo_rl/algorithms/single_controller.py
saumishr: Fixed in `2146b67c2`. The target is read once now, before the subtraction, and a non-positive remainder `break`s the step closed instead of asking the sampler for a batch of zero.

Took the denominator with it: the exhaustion `RuntimeError` beside this reported `num_prompts_per_step`, which accused a legitimately shrunk step of missing groups it had been excused from. It reports that same local target now.

`test_train_pump_closes_a_step_short_when_a_stamped_prompt_is_dropped` is parametrized over both interleavings -- the credit landing inside `evict`, and landing between the two reads -- so neither one can regress alone. Both fail against the pre-fix file; the `credit_in_evict=True` case fails with exactly the `min_prompt_groups must be >= 1, got 0` from the TL;DR.

nemo_rl/algorithms/single_controller.py
saumishr: Fixed in `2146b67c2`. `_replacement_reserve` is snapshotted in the same await-free window as `self._dataloader.state_dict()` -- so the pair cannot disagree -- written as `replacement_reserve.pt` beside `train_dataloader.pt`, and restored in `run()` next to the buffer restore.

Two deliberate asymmetries with the buffer restore, both from spares never having reached `admit` and so carrying no stamp: no sampler-name guard, and silence rather than a warning when the file is absent (which is every run that never diverted). For the same reason `_drain_reserve_into_steps` loses its `on_dropped_prompt` gate, so a pool restored into a run that has since switched to `shrink` is still trained on rather than stranded -- an empty pool makes it a no-op anyway.

Un-gating that drain turned up six hand-built controllers in `test_rollout_pump.py` that predate `_replacement_reserve` and only passed because the gate short-circuited before reading it; they now go through one shared `_init_pump_ledgers` helper.

Five tests: save, restore, restore-when-the-buffer-restore-is-skipped, empty-pool-writes-no-file, and absent file. Three of the five fail against the pre-fix file.

tests/unit/single_controller/test_single_controller.py
saumishr: Added in `2146b67c2`, and it earned its keep -- written against the fix, it fails against the pre-fix file on both parametrizations.

`test_train_pump_closes_a_step_short_when_a_stamped_prompt_is_dropped` drives the pump with a sampler that drops a stamped prompt and credits the shortfall, parametrized over crediting inside `evict` and crediting between the two target reads. It asserts the step closes short, that `_consumed_samples` counts the groups actually trained (which was also wrong -- it advanced by `num_prompts_per_step` regardless, overstating the dataset position written into the checkpoint), and the three new metric counters.

`test_train_pump_prunes_stamps_older_than_the_step_that_just_closed` covers the other half you flagged: entries for closed steps are dropped and later ones kept.

nemo_rl/algorithms/single_controller_utils/setup.py
saumishr: Added in `2146b67c2`: `TestTheDropBudgetsReachTheRolloutLayer` in `test_resiliency_config.py` calls `_build_retry_policy` on a real `MasterConfig` and asserts `max_consecutive_dropped_prompts` and `max_skipped_prompts` arrive on the `RolloutRetryPolicy`, plus a second test pinning that the default still fails on the first drop.

Your test for the finding is the honest one: delete the assignment and these two fail, where before every config test stayed green.

nemo_rl/algorithms/single_controller_utils/config.py
saumishr: Added in `2146b67c2` as `_warn_on_inert_failure_settings`, called from `validate_single_controller_config` since all three need `grpo.num_prompts_per_step` and cannot be seen from inside the block: the floor that rounds up to the full batch while a drop budget is set, `replace` under a sampler that never stamps, and a reserve mark above one batch.

Warnings rather than the hard rejection the two existing checks use. Each of these is coherent and defensible to have typed -- a strict never-train-a-short-batch run, one base YAML whose sampler is overridden per experiment, a deliberately deep pool -- so rejecting them would forbid configurations somebody wants; what is not acceptable is finding out hours in. The `custom` sampler is in neither the stamps nor the never-stamps set, since a warning nobody can act on is noise.

Seven tests, covering each warning and the neighbouring cases that must stay silent, including that an untouched config warns about nothing.

nemo_rl/experience/rollout_manager.py
saumishr: Fixed in `2146b67c2` -- the `RolloutOutcome.SKIPPED` comment now names both causes, so it matches the `Returns:` block this hunk updated.

tests/unit/single_controller/test_tq_replay_buffer.py
saumishr: Fixed in `2146b67c2`, using your suggestion.

nemo_rl/algorithms/async_utils/replay_buffer.py
saumishr: Fixed in `2146b67c2`: `promote_ready_group` has the `Args:` section now, in line with its siblings on that interface.

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -288,6 +390,81 @@ def validate_sampler_buffer_capacity(
...
+
+    if failure_config.on_dropped_prompt == "replace" and sampler_never_stamps:
+        warnings.warn(
+            f"async_rl.rollout_failure.on_dropped_prompt='replace' does nothing with "
+            f"the {sampler_name} sampler: it does not stamp a target step, so no step "
+            "is ever left short, the spare pool is never filled, and every drop "
+            "behaves as 'shrink'. replaced_prompt_groups will stay at 0 whether or "
+            "not anything failed.",
+            stacklevel=2,
+        )
```
terrykong, line 660: [`config.py:442`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller_utils/config.py#L442)

_📊 Visualised as **Finding 1** on the review page: [https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-what-the-review-found.html#finding-1](https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-what-the-review-found.html#finding-1)_

**1 action item.**

**TL;DR — with `weight_fifo`, turning on any drop budget stops training for good: no step is ever told it lost a prompt, so the trainer keeps asking for a full batch that can never arrive.**

**First, the one word this all turns on.** When the rollout pump sends a batch out, [`in_order`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/async_utils/staleness_sampler.py#L409) gives that batch **the number of the training step it belongs to** (`target_step`), and step N will only ever train on groups carrying the number N. [`weight_fifo and windowed`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/async_utils/staleness_sampler.py#L344) return `None` instead — their batches carry **no step number** at all.

Everything this PR adds hangs off that number. [`_credit_shortfall`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller.py#L757) and [`_take_replacement`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller.py#L706) both return immediately when it is `None`, so with `weight_fifo` or `windowed` **neither "shrink" nor "replace" ever runs** — the step's target stays at the full `num_prompts_per_step` forever and `min_step_batch_fraction` is never consulted.

Pre-existing rather than something this PR broke — `max_skipped_prompts` could already reach it on `main`. What this PR changes: `max_consecutive_dropped_prompts` routes the far more common *infra* drop into the same path, the YAML comment invites raising it, and this function is the natural home for the guard.

**Why `weight_fifo` never recovers.** Two ranges disagree at exactly one point:

1. A drop leaves the oldest in-window weight version holding `N-1` groups.
2. [`select`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/async_utils/staleness_sampler.py#L381) always takes the oldest version in its window — that one.
3. [`_finalize_selection`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/async_utils/staleness_sampler.py#L229) refuses it: `len(valid_idxs) < min_prompt_groups`.
4. [`evict`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/async_utils/staleness_sampler.py#L176) cannot clear it either: evict removes `weight < min_valid_version` (**strict**), while [`select's window`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/async_utils/staleness_sampler.py#L373) is `min_valid_version <= weight` (**inclusive**). A version sitting exactly on the window floor is selectable but never evictable.
5. The trainer cannot advance to age it out, because the trainer is what is stuck. `admit` then gates the rollout pump too.

No exception is raised. The train pump sleeps 5ms and re-asks forever, both pumps go idle, and the job holds its GPUs until the wall clock kills it — `stall_action` defaults to `warn`, so the watchdog only prints.

`windowed` degrades more gently: it is not pinned to the oldest version, so the shortage is carried forward and the **final** step fails with `rollout exhausted before a complete training step was assembled`.

**Which of the two failures you get is decided by an unrelated speed setting.** Because the step's target never moves here, [`min_prompt_groups`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller.py#L871) collapses to just `min_groups_for_streaming_train`, and that alone picks the outcome: at `== num_prompts_per_step` (what the SC exemplar sets, via `${grpo.num_prompts_per_step}`) `select` can never be satisfied and training stops; below it, the step tops itself up with a prompt from the next batch and the shortage walks forward to the final step. Under `in_order` that setting has no say in fault tolerance at all.

Nothing we ship is exposed today (no shipped config sets a non-zero budget), but both shipped SingleController recipes use `windowed`, so they are one override away.

Worth knowing why the suite stays green: `test_rollout_pump.py` asserts `_batch_shortfall == {}` for `weight_fifo`, pinning the no-op as intended, and no `_train_pump` test runs a no-step-number sampler with a drop in flight.

**Action:** add a fourth block to `_warn_on_inert_failure_settings`, beside the existing check (`drop_budget` and `sampler_never_stamps` are already computed above). This rejects nothing that works today — a config that reaches it either stops making progress or dies at the last step. One way to do it, of several:

```suggestion
    if drop_budget and sampler_never_stamps:
        # A drop is only ever credited against a step that carries a step number, so
        # with these samplers the shortfall is never subtracted and the train pump keeps
        # asking for the full batch. weight_fifo cannot recover from that at all;
        # windowed defers the shortage to the final step. Neither is a configuration
        # that works today, so this is rejected/warned rather than left to fail hours in.
        message = (
            f"async_rl.rollout_failure drop budgets (max_skipped_prompts="
            f"{failure_config.max_skipped_prompts}, max_consecutive_dropped_prompts="
            f"{failure_config.max_consecutive_dropped_prompts}) are not honoured by "
            f"the {sampler_name} sampler: it does not give a batch the number of the "
            "training step it belongs to, so a dropped prompt credits no shortfall and "
            "the train pump keeps asking for the full grpo.num_prompts_per_step."
        )
        if sampler_name == "weight_fifo":
            raise ValueError(
                message + " The oldest in-window weight version is then permanently "
                "unselectable -- select pins on min(in_window), evict only drops "
                "strictly older weights -- and training stops making progress. Use "
                "async_rl.sampler.name='in_order', or set both budgets to 0."
            )
        warnings.warn(
            message + " The shortage is carried to the final training step, which "
            "then fails with 'rollout exhausted before a complete training step was "
            "assembled'. Use async_rl.sampler.name='in_order' to drop prompts safely.",
            stacklevel=2,
        )

    if failure_config.on_dropped_prompt == "replace" and sampler_never_stamps:
        warnings.warn(
            f"async_rl.rollout_failure.on_dropped_prompt='replace' does nothing with "
            f"the {sampler_name} sampler: it does not stamp a target step, so no step "
            "is ever left short, the spare pool is never filled, and every drop "
            "behaves as 'shrink'. replaced_prompt_groups will stay at 0 whether or "
            "not anything failed.",
            stacklevel=2,
        )

```

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -112,6 +126,75 @@ class RolloutFailureConfig(BaseModel, extra="allow"):
...
+    # that has stopped answering stops the run instead of retrying into a stall.
+    #
+    # 0 (the default) fails on the first exhaustion, which is both the behaviour before
+    # this knob existed and the default the v1 stack ships for the same idea
+    # (``async_grpo.max_generation_failures``).
+    #
+    # Counted separately from max_skipped_prompts rather than sharing one budget: a
+    # shared counter would let a bad dataset consume the allowance that exists to ride
+    # out a shard outage, which is the one distinction the failure taxonomy draws.
+    max_consecutive_dropped_prompts: NonNegativeInt = 0
```
terrykong, line 142: [`config.py:141`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller_utils/config.py#L141)

_📊 Visualised as **Finding 2** on the review page: [https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-what-the-review-found.html#finding-2](https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-what-the-review-found.html#finding-2)_

**No action needed in this PR — 2 follow-up asks.**

**TL;DR — `max_consecutive_dropped_prompts: 0` does not mean "tolerate no drops and carry on", it means "end the whole run on the first one"; and it measures the same thing `watchdog.stall_timeout_s` measures, with the opposite default reaction.**

PR-introduced. Not asking for `max_replacement_attempts` or `replacement_reserve_prompts` to be removed — they are defaulted sensibly and exposing a knob someone may need is cheap.

### AI-1 — the `0` hides a mode inside a number

`0` reads as "tolerate zero", but it means "stop the run". And "what happens to a step that lost a prompt" already has its own field, [`on_dropped_prompt`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller_utils/config.py#L181), so the answer is split across two places with a magic number in one of them:

```yaml
# today
on_dropped_prompt: shrink            # or replace
max_consecutive_dropped_prompts: 0   # 0 = end the run. Not obvious.

# suggestion: the mode says WHAT happens, the budget says HOW MUCH is tolerated
on_dropped_prompt: fail              # fail | shrink | replace; "fail" is today's default
max_consecutive_dropped_prompts: 8   # only read when the mode is not "fail"; require >= 1 there
```

Same behaviour, no magic `0`, and the default describes itself.

### AI-2 — it overlaps the watchdog, and the defaults contradict

Both answer "has the fleet stopped answering?", and both read it off the same fact — whether anything has committed lately. The watchdog resets its clock on `stats.committed` ([`single_controller.py:1172`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller.py#L1172)); this budget resets its counter on commit ([`rollout_manager.py:1341`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/experience/rollout_manager.py#L1341)).

| | catches | speed | default today |
|---|---|---|---|
| `max_consecutive_dropped_prompts` | only given-up prompts | fast, deterministic | `0` — ends the run on the first one |
| `watchdog.stall_timeout_s` | **any** lack of progress — stalls, wedges, dead environments | slow (600s) | `warn` — prints and keeps running |

Out of the box the run is hair-trigger about the narrow case and permissive about the general one. They layer well with the defaults flipped: a small non-zero budget as the fast, specific trip, and `stall_action: abort` as the catch-all backstop. There is a payoff for the sampler comment on this review — with `stall_action: abort`, that case stops being an unbounded stall burning GPUs and becomes a bounded failure with a real error. It does not fix it, but it caps the damage.

For reference, the v1 stack answers this whole question with one setting, [`async_grpo.max_generation_failures`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/grpo.py#L191), with the same reset-on-success behaviour — which this field's own docstring cites as the precedent.

_Not a suggestion block: the change spans the schema, the validators, and the exemplar YAML._

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -112,6 +126,75 @@ class RolloutFailureConfig(BaseModel, extra="allow"):
...
+    # any commit at all, including commits belonging to other steps. So drops
+    # concentrated on one step, interleaved with unrelated successes, keep resetting the
+    # counter while that one step shrinks without limit -- and a step is what the
+    # gradient is actually computed from.
+    #
+    # A fraction rather than a count so it holds across batch sizes, and so small
+    # batches are protected more strictly: losing 1 group of 4 is a 25% batch reduction
+    # and should not be treated like losing 1 of 128. The floor is ceil(fraction * N),
+    # so at the default a batch of 8 or fewer tolerates no drops at all.
+    min_step_batch_fraction: Annotated[float, Field(gt=0.0, le=1.0)] = 0.9
```
terrykong, line 157: [`config.py:156`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller_utils/config.py#L156)

_📊 Visualised as **Finding 3** on the review page: [https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-what-the-review-found.html#finding-3](https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-what-the-review-found.html#finding-3)_

**No action needed in this PR — documentation only. No bug here; the code is right.**

**TL;DR — `min_step_batch_fraction` and `min_groups_for_streaming_train` read as two spellings of the same idea, and the natural guess (that you must lower the second before a step may shrink) is wrong.**

Four similar-sounding numbers now decide how big a step is. Two are settings; two are local variables recomputed every pass of the train loop:

| value | where it lives | what it is | when it changes |
|---|---|---|---|
| `num_prompts_per_step` | config | groups a full step wants | **never** |
| `min_step_batch_fraction` | config | how much of that is still acceptable | never |
| `min_groups_for_streaming_train` | config | smallest chunk worth starting a pass on | never |
| `floor = ceil(N x fraction)` | derived | below this the run raises | never |
| `target_groups` | runtime | what **this** step will actually train on | once per step, per drop |
| `max_prompt_groups` | **local var** | how many this step still needs right now | **every pass** |
| `min_prompt_groups` | **local var** | smallest chunk to ask for right now | **every pass** |

Only `min_prompt_groups` comes from `min_groups_for_streaming_train`. There is deliberately no `max_groups_for_streaming_train` — the upper bound is a fact (what the step still needs), not a tuning choice.

They cannot fight, because the chunk size is capped by what the step still needs:

```
target_groups     = num_prompts_per_step - dropped        # :769, the fraction only decides if this is legal
max_prompt_groups = target_groups - groups_dispatched     # :868, already reduced
min_prompt_groups = min(min_groups_for_streaming_train, max_prompt_groups)   # :871  <- the cap
```

At `num_prompts_per_step=32`, `min_groups_for_streaming_train=30`, `min_step_batch_fraction=0.5` (floor 16):

```
 0 dropped -> target 32, max 32, min = min(30,32) = 30   # cap idle
 3 dropped -> target 29, max 29, min = min(30,29) = 29   # cap starts doing the work
15 dropped -> target 17, max 17, min = min(30,17) = 17   # step trains on 17, well under the chunk size
17 dropped -> target 15 < floor 16 -> RuntimeError
```

**Action:** one sentence on each field saying which thing it bounds — `min_step_batch_fraction` bounds the **whole step**, `min_groups_for_streaming_train` bounds **one chunk inside** a step, and the second is capped by the first so it never needs adjusting to tolerate drops. Two traps worth naming while you are there: the floor is a **check, not a target** (losing 1 of 4 gives `target=3`, not `4 x fraction`), and [`the cap at :871`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller.py#L871) is what lets a step legitimately train on fewer groups than the chunk size.

_Not a suggestion block: the wording belongs on two fields in two different classes._

nemo_rl/algorithms/single_controller.py
```diff
@@ -508,10 +855,19 @@ async def _train_pump(self) -> None:
...
+                        # can also have fallen to what is already dispatched, which is
+                        # not a batch the sampler can be asked for -- select() rejects
+                        # a min below 1 -- so close the step instead.
+                        target_groups = self._target_groups_for_step(
+                            version_during_step
                         )
+                        max_prompt_groups = target_groups - groups_dispatched
+                        if max_prompt_groups <= 0:
+                            break
                         min_prompt_groups = min(
```
terrykong: [`single_controller.py:871`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/single_controller.py#L871)

**No action needed in this PR** — a test gap; a follow-up is fine.

**TL;DR — this `min(...)` is the only thing that makes a shrunk step trainable, and every new shrink test sets `min_groups_for_streaming_train=1`, where the `min()` is a no-op.**

PR-introduced coverage gap. With the SC exemplar's `min_groups_for_streaming_train: ${grpo.num_prompts_per_step}`, a step that lost one prompt has `max_prompt_groups = N-1` while the configured chunk size is still `N`. This line is what stops `select` being called with `min=N, max=N-1`, which [`_validate_group_bounds`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/nemo_rl/algorithms/async_utils/staleness_sampler.py#L210) rejects outright:

```
ValueError: max_prompt_groups (31) must be >= min_prompt_groups (32)
```

The new shrink tests cannot see that. [`_train_pump_controller`](https://github.com/NVIDIA-NeMo/RL/blob/ad048f995296473511641f8cb1b8418625fe61ea/tests/unit/single_controller/test_single_controller.py#L357) sets `min_groups_for_streaming_train=1` with `num_prompts_per_step=2`, so after the drop the line evaluates `min(1, 1)` — identical whether or not the cap is here. **Replace this line with a bare `min_groups_for_streaming_train` and the whole suite still passes**, while the shipped exemplar config raises on its first dropped prompt.

**Action:** add one parametrisation of `test_train_pump_closes_a_step_short_when_a_stamped_prompt_is_dropped` with `min_groups_for_streaming_train == num_prompts_per_step` (and a `min_step_batch_fraction` low enough to permit the drop, as the existing case already does at `0.5`). That pins the cap at the value the exemplar actually ships.

_Not a suggestion block: the fix is a new parametrisation in the test file, not an edit to this line._

terrykong (commented):
Reviewed by a team of three agents (bug hunt, async/concurrency + sampler-buffer interaction, config/tests). Focused on functionality per the review request — no style or prose comments.

**📊 The comments below are visualised here: [three problems this review found](https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-what-the-review-found.html)** — one diagram per finding, including the exact boundary mismatch that makes `weight_fifo` stop making progress. Two companion pages explain the PR itself, if useful: [what problem it solves](https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-why-a-lost-rollout-ends-the-run.html) and [how the fix works](https://terrykong.github.io/gh-pages-poc/terryk/pr-3665-replacing-a-lost-rollout.html), the latter with a table of which of the four batch-size numbers are settings and which are recomputed each pass.

**One finding is a defect** — the sampler one. The other two are follow-up asks, both opening with "no action needed in this PR":

| where | what |
|---|---|
| `config.py:442` | **[bug]** with `weight_fifo`, any non-zero drop budget stops training for good; `windowed` fails at its final step. Pre-existing, but this PR adds a second way in and documents it as safe. Suggestion block included. |
| `config.py:141` | `max_consecutive_dropped_prompts: 0` hides "end the run" inside a number, and duplicates what `watchdog.stall_timeout_s` measures — with the opposite default reaction. |
| `config.py:156` | how `min_step_batch_fraction` and `min_groups_for_streaming_train` relate is undocumented, and the natural guess is wrong. |

Candidates raised and then dropped as not meeting the bar: a checkpoint reserve/buffer snapshot skew (real but bounded to one duplicated prompt group on one resume); a suspected behaviour change from the `min_step_batch_fraction` default (refuted — the path it affects already failed on `main`, so this is a fix); and a missing test for the chunk-size cap at `single_controller.py:871` (the code is correct and the line is pre-existing, so it did not clear the bar for this review). Permit accounting across the new replacement loop, `promote_ready_group` re-stamp safety, the second `TaskGroup` closure rebind, `_consumed_samples` units, the consecutive-drop counter bounds, and the example YAML all checked out clean.

Note: this PR currently shows a merge conflict against `main` (`mergeable: CONFLICTING`) and will need a rebase.

Tests were reviewed statically only — they cannot run on this host (the lockfile is gated to linux/x86_64), so no local pass/fail is claimed, and no run of this feature exists yet. The single most useful thing before merge would be one short run with `sampler.name: weight_fifo` and `max_consecutive_dropped_prompts: 1`, forcing one prompt to fail, to confirm whether training keeps advancing.

_Generated by Claude Code_

terrykong (approved):

terrykong:
given urgency, fixing the in_order bug was priority, other two things can be left for later

PR #3668, opened 2026-08-16
fix(gym): hold the tokenizer in NemoGym instead of passing it per rollout

saumishr:
# Overview
run_rollouts took the tokenizer as an argument and the rollout pump calls it once per prompt, so Ray deserialized a 7.45 MB tokenizer on the actor's task-execution thread for every prompt. That thread holds the GIL, so it blocked the actor's event loop and no rollout could issue its first HTTP request until its own copy finished loading.

Measured: 286 ms to serialize, 1052 ms to deserialize. A SingleController recipe admitting ~1000 prompts per step spends ~18 minutes deserializing one admission burst.

Reproduced on one CPU node with 1024 concurrent calls against one actor: 153 of 1024 prompts finished in 300 s passing the tokenizer per call, versus all 1024 in 16 s installing it once. Passing an ObjectRef instead does not help; Ray caches the object buffer and not the deserialized value, so the actor still pays per task.

Install it through the spinup paths so no call site can forget: the shared spinup_nemo_gym_actor helper takes it as a keyword argument, and distillation, which builds its actor inline, sets it alongside its own _spinup call. On the SingleController path it threads through _spinup_gym, since the kwargs that used to sit at the partial now live inside that function.

run_rollouts fails loudly rather than silently if it is somehow reached without one -- postprocessing with no tokenizer is worse than a named error, and the attribute is initialized in __init__ so the check reads as state rather than as an absent attribute.

Test fakes that stood in for the actor method lose the argument with it, and the SingleController setup test now asserts the tokenizer reaches spinup, which is the property that keeps it off the per-call path.

# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/environments/nemo_gym.py
+++ b/nemo_rl/environments/nemo_gym.py
@@ -416,6 +416,10 @@ class NemoGym(EnvironmentInterface):
         self.head_server_config: Any = None
         self.node_ip: Optional[str] = None
         self.head_server_port: Optional[int] = None
+        # Installed by set_tokenizer at spinup, not passed per rollout call. Declared
+        # here rather than in _spinup so a second spinup cannot wipe an installed
+        # tokenizer and then report that set_tokenizer was never called.
+        self._tokenizer: Optional[PreTrainedTokenizerBase] = None
         self._pad_dynamic_image_shapes = bool(cfg.get("pad_dynamic_image_shapes"))
         # Reconstruct the processor inside the actor (rather than serializing it
         # per rollout call) for full-trajectory multimodal postprocessing.
@@ -553,10 +557,33 @@ Depending on your data shape, you may want to change these values."""
         )
         self.rch = RolloutCollectionHelper()
 
+    def set_tokenizer(self, tokenizer: PreTrainedTokenizerBase) -> None:
+        """Install the tokenizer run_rollouts postprocesses with.
+
+        Called once per actor at spinup. It used to be a run_rollouts argument,
+        which meant Ray deserialized a tokenizer per prompt on this actor's
+        task-execution thread. That thread holds the GIL, so it blocked the
+        actor's event loop and no rollout could issue its first HTTP request
+        until its own copy finished loading.
+
+        The cost is not marginal. A tokenizer of this shape measured 7.45 MB on
+        the wire, 286 ms to serialize and 1052 ms to deserialize, so a
+        SingleController recipe admitting ~1000 prompts per step spends roughly
+        18 minutes deserializing a single admission burst, against a step that
+        should take minutes. Runs on that shape stalled without completing a
+        rollout.
+
+        Measured on one CPU node with 1024 concurrent calls against one actor:
+        153 of 1024 prompts finished in 300 s passing the tokenizer per call,
+        versus all 1024 in 16 s holding it here. Passing an ObjectRef instead
+        does not help -- Ray caches the object buffer, not the deserialized
+        value, so it still pays per task.
+        """
+        self._tokenizer = tokenizer
+
     async def run_rollouts(
         self,
         nemo_gym_examples: list[dict],
-        tokenizer: PreTrainedTokenizerBase,
         timer_prefix: str,
         deduplicate_multimodal_data: bool = False,
     ) -> AsyncGenerator[tuple[int, dict, dict | None], None]:
@@ -564,6 +591,11 @@ Depending on your data shape, you may want to change these values."""
         self._require_spinup()
         if not nemo_gym_examples:
             raise ValueError("NeMo-Gym rollout batch must not be empty")
+        if self._tokenizer is None:
+            raise RuntimeError(
+                "NemoGym.set_tokenizer must be called before run_rollouts"
+            )
+        tokenizer = self._tokenizer
 
         from nemo_rl.utils.fastokens import maybe_patch_fastokens
 
@@ -1024,6 +1056,7 @@ def spinup_nemo_gym_actor(
     base_urls: list[str],
     model_name: str,
     *,
+    tokenizer: PreTrainedTokenizerBase,
     enable_router_replay: bool,
     routed_experts_dtype: str,
     use_fastokens: bool,
@@ -1040,6 +1073,9 @@ def spinup_nemo_gym_actor(
             thinking_tags, num_gpu_nodes).
         base_urls: Per-DP-rank OpenAI-compatible server base URLs from the generation backend.
         model_name: Served model name the Gym rollouts should target.
+        tokenizer: Installed on the actor once, here, rather than passed per
+            rollout call. See NemoGym.set_tokenizer for why that distinction is
+            the difference between a working run and a stalled one.
         enable_router_replay: Sets require_routed_experts on the NemoGymConfig.
         routed_experts_dtype: Dtype name for R3 routed_experts tensors ("int8"/"int16"/"int32"),
             resolved by the caller from the model's expert count.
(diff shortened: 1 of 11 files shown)


saumishr (commented):
Two findings, both low severity: the new `_tokenizer` attribute is declared in `_spinup()` rather than in `__init__` (a latent order dependency, and not what the PR description says the code does), and the new distillation assertion does not check which object gets installed.

nemo_rl/environments/nemo_gym.py
```diff
@@ -552,18 +552,47 @@ def _spinup(self) -> None:
             port=self.head_server_port,
         )
         self.rch = RolloutCollectionHelper()
+        # Installed by set_tokenizer at spinup, not passed per rollout call.
+        self._tokenizer: Optional[PreTrainedTokenizerBase] = None
```
saumishr: `self._tokenizer` is declared at the end of `_spinup()` rather than in `__init__`. The PR description says the attribute "is initialized in `__init__` so the check reads as state rather than as an absent attribute" — that is the right call, but it is not what the code does. `__init__` (L411-418) already declares every other `_spinup`-populated field (`rh`, `rch`, `head_server_config`, `node_ip`, `head_server_port`) with a comment saying exactly why: Ray recreates a restarted actor through `__init__` only, and this actor is `@ray.remote(max_restarts=-1, max_task_retries=-1)`.

Nothing breaks today — `run_rollouts` calls `_require_spinup()` (L588) before it reads `self._tokenizer` (L591), so a restarted actor still gets the named spinup error rather than an `AttributeError`. But two latent problems follow from the placement. `_spinup()` resets the tokenizer to `None`, so any future path that installs before spinning up, or re-runs `_spinup` for recovery, silently loses it and then reports "set_tokenizer must be called before run_rollouts" when it *was* called. And `test_run_rollouts_requires_an_installed_tokenizer` cannot catch that, because it builds the instance with `object.__new__` and assigns `_tokenizer` by hand.

Suggested fix: move L555-556 into `__init__`, next to the other "Populated by `_spinup`" declarations. That also satisfies "Initialize all externally visible members of a class in the constructor" from the code-style guideline.

tests/unit/algorithms/test_distillation.py
```diff
@@ -1281,7 +1281,13 @@ def prepare_refit_info(self, *args, **kwargs):
...
     assert master_config.env["nemo_gym"] == nemo_gym_env_before
     nemo_gym_actor._spinup.remote.assert_called_once_with()
-    mock_ray.get.assert_called_once_with("spinup-ref")
+    # Two waits, in order: the spinup, then the tokenizer install. Distillation
+    # builds its actor inline rather than through spinup_nemo_gym_actor, so it
+    # is the one call site that has to set the tokenizer itself -- passing it
+    # per rollout is what made the actor deserialize it once per prompt.
+    assert mock_ray.get.call_args_list[0] == call("spinup-ref")
+    assert mock_ray.get.call_count == 2
+    nemo_gym_actor.set_tokenizer.remote.assert_called_once()
```
saumishr: `assert_called_once()` checks that the install happened but not *what* was installed, so a regression that forwarded the wrong object — a processor, `None`, a re-derived tokenizer — would still pass, even though "the right tokenizer reaches the actor" is the entire point of this change.

Suggested fix: tighten to `nemo_gym_actor.set_tokenizer.remote.assert_called_once_with(tokenizer)`. The `tokenizer` MagicMock from L1171 is in scope and `distillation.py:567` forwards `setup`'s argument verbatim, so the stricter assertion passes as-is. Worth being precise here because, as the comment right above notes, distillation builds its actor inline instead of going through `spinup_nemo_gym_actor`, which makes this test the only guard on that install.

nemo_rl/environments/nemo_gym.py
saumishr: Fixed in 4b8b4853d. The declaration moved into `__init__` next to the other `_spinup`-populated fields, and `_spinup` no longer touches `_tokenizer`, so a second spinup can't wipe an installed tokenizer and then report a call that was in fact made.

The guard test moved with it. It built the instance with `object.__new__` and assigned `_tokenizer` by hand, so it passed no matter where the declaration lived; it now constructs through `__init__` and asserts the `None` comes from the declaration itself.

On the framing: you're right that nothing was reachable. `run_rollouts` calls `_require_spinup()` before it reads the tokenizer and that guard trips on `rh` first, so a restarted actor was already getting the named spinup error rather than an `AttributeError`.

tests/unit/algorithms/test_distillation.py
saumishr: Fixed in 4b8b4853d — now `assert_called_once_with(tokenizer)`. The `tokenizer` MagicMock is in scope from L1171 and `distillation.py:567` forwards `setup`'s argument verbatim, so the stricter assertion passes unchanged.

Agreed on why it matters here specifically: distillation builds its actor inline instead of going through `spinup_nemo_gym_actor`, so this is the only test that observes that install.

saumishr:
Force-pushed [1c6f073](https://github.com/NVIDIA-NeMo/RL/commit/1c6f0730c800053ee75fe34e30588f7fba631115) — two changes, plus a history rewrite to keep internal identifiers out of the record.

**Scrubbed the internal references.** The `set_tokenizer` docstring cited a Slurm job id and an internal recipe label to establish the cost; the PR description carried the recipe label too. Both now describe the shape generically ("a SingleController recipe admitting ~1000 prompts per step", "runs on that shape stalled without completing a rollout"), and the measurements that actually carry the argument are unchanged. I folded that edit into the original commit and reworded its message rather than stacking a fix on top, so no commit in the branch — message or tree — contains the job id.

**Fixed the `L0_Unit_Tests_Nemo_Gym` failure**, which was real and not flake: `test_async_nemo_gym_rollout_manager` died on `NemoGym.set_tokenizer must be called before run_rollouts`. The shared `nemo_gym` fixture yields an actor after `_spinup` only, so once `run_rollouts` stopped taking a tokenizer per call, every test reaching the actor through `RolloutManager` had no way to install one — `test_nemo_gym_sanity` was patched inline and was the only one that could pass. The fixture now installs it the way `spinup_nemo_gym_actor` does, and the inline call is gone. That closes the audit: all three actor construction sites (`spinup_nemo_gym_actor`, distillation's inline build, and the fixture) install a tokenizer, and there are no others.

The earlier `CI quality check` failure was just the missing CI label, which is on the PR now.

saumishr:
Rebased onto current main (`f9a2b30c0`) and force-pushed [e483970](https://github.com/NVIDIA-NeMo/RL/commit/e4839706b51a1d401af5e4e6782203098c4d4e2b); conflicts are resolved.

One conflict, in `_spinup_gym`'s `build_tasks` entry in `nemo_rl/algorithms/single_controller_utils/setup.py`. #3590 changed `base_urls` to prefer the generation router's single URL when the router is live, and this PR adds `tokenizer=tokenizer` to the same call — both sides kept, since they are independent arguments.

The patch is otherwise unchanged: same 11 files and same +106/-23 as before the rebase, with only hunk offsets moving. I re-ran the audit that the last round's CI failure came from, against the new base: #3590 adds no new `NemoGym` construction or `run_rollouts` call site, so the three construction sites (`spinup_nemo_gym_actor`, distillation's inline build, the `nemo_gym` fixture) still all install a tokenizer, both `spinup_nemo_gym_actor` callers still pass one, and both remote `run_rollouts` calls still omit it. `ruff format`, `ruff check` and `py_compile` are clean on all 11 files.

`test_single_controller_setup.py`'s `base_urls` assertion still matches, because the router is disabled in that test's config and the resolved expression falls through to `dp_openai_server_base_urls`.

terrykong (approved):

PR #3669, opened 2026-08-16
perf(vllm): batch BF16 TRTLLM expert layout conversion

seonjinn:
## Summary

- Convert all BF16 `W13` experts with one indexed gather and all `W2` experts with a second gather, instead of running two gathers per expert.
- Limit the batched converter to the vLLM reload lifecycle and keep the output layout and runtime parameter addresses unchanged.

## Validation

- GB200, Qwen3-30B-A3B, 20-step BF16 training and BF16 rollout: refit bubble decreased from 3.637 s to 3.124 s (-14.1%).
- Generation KL error remained equivalent: 0.001887 baseline vs. 0.001886 batched.
- End-to-end throughput stayed within the run-to-run difference and is not claimed as improved.
- 82 targeted GB200 BF16 TRTLLM refit tests pass.

Depends on #3659.


seonjinn:
Contributor self-review refresh for `1b5bec2a9d42b88ba4ea8e22667232f9a2ca9370` after restacking on the latest #3659 and main.

The round-robin placement hardening is preserved, the refit batching change remains scoped to BF16 TRTLLM layout conversion, and no new source-level findings were identified. Ruff/format/config checks pass locally; Linux vLLM coverage remains delegated to CI.

seonjinn:
/ok to test [d24868e](https://github.com/NVIDIA-NeMo/RL/pull/3669/commits/d24868ebad75a48f1bc49c17e32b15cfe28c1c24)

guyueh1:
@seonjinn this code applies a runtime patch to the vllm vllm.model_executor.layers.fused_moe.oracle module; there are multiple ways in nemo-rl to patch vllm (offline or runtime), is this context based way used anywhere else? should we follow the way in fp8.py, or patches.py?

I think this change is fine, but just wondering about if this is best practice for patching vllm

seonjinn:
The scoped context is intentional here. `fp8.py` installs process-lifetime model/loading patches, while `patches.py` rewrites or installs compatibility patches before worker initialization. This optimization is only valid during TRTLLM BF16 layerwise-reload initialization/finalization, so making it global would also affect normal model loading and unrelated calls. The context manager serializes mutation with an `RLock`, uses a `ContextVar` so another thread falls back to the original converter, and restores the exact original function on success or exception. I therefore kept it local to the reload lifecycle rather than adding another process-wide vLLM patch. The focused test covers activation, cross-thread fallback, and restoration.

PR #3670, opened 2026-08-16
feat: validate NCCL HCA selection before Ray startup

seonjinn:
## Problem

A valid but incorrect `NCCL_IB_HCA` selection can silently slow multi-node collectives.

## What changed

- Validate each node's configured HCAs before Ray starts.
- In the default `auto` mode, repair only from an explicitly trusted list whose devices and ports are active.
- Otherwise leave the environment unchanged and emit a sanitized per-node report.

### OCI-based H100 collective improvement

A matched 2-node × 8-GPU H100 collective test compared the affected default HCA selection with the topology-aligned selection that this guard can apply from a trusted cluster-provided list.

| Metric | Affected default | Topology-aligned HCA | Improvement |
| --- | ---: | ---: | ---: |
| All-reduce bandwidth | 34.54 GB/s | 44.60 GB/s | **+29.13%** |
| Worst all-reduce latency | 1.000× (normalized) | 0.492× | **−50.76%** |
| Reduce-scatter bandwidth | 35.69 GB/s | 44.62 GB/s | **+25.02%** |
| All-reduce rank CV | 29.78% | 0.26% | **−29.52 pp (−99.13%)** |

The corrected result was within 4.1% of a healthy H100 reference (46.52 GB/s all-reduce and 45.66 GB/s reduce-scatter). Selecting only the cross-NIC devices reached 36.44 GB/s but retained 28.36% rank CV, confirming that device presence alone did not remove the rail imbalance. Private infrastructure identifiers and scheduler metadata are omitted.

### End-to-end guard validation

A matched 4-node × 4-GPU GB200 Qwen3-30B-A3B run compared `off` with the default `auto` mode over steps 2–20. `auto` left the existing NCCL environment unchanged on all four nodes; both 20-step runs completed successfully.

| Metric | Guard off | Default `auto` | Delta |
| --- | ---: | ---: | ---: |
| E2E step time | 184.95 s | 185.36 s | +0.22% |
| Policy training time | 81.05 s | 81.40 s | +0.43% |
| Policy + reference logprob time | 20.38 s | 20.36 s | −0.11% |
| E2E throughput | 2,253.20 tok/s/GPU | 2,247.27 tok/s/GPU | −0.26% |
| Policy training throughput | 5,122.59 tok/s/GPU | 5,099.21 tok/s/GPU | −0.46% |
| Policy + reference logprob throughput | 20,370.28 tok/s/GPU | 20,386.94 tok/s/GPU | +0.08% |

All steady-state differences were below 0.5%. The warm-up step was excluded; periodic validation steps remained included.

## Validation

- 30 unit and CLI integration tests passed; targeted line coverage is 91% locally.
- Pyrefly, formatting, launcher shell syntax, and diff checks passed locally.
- Both 20-step GB200 end-to-end runs completed with exit code 0, no topology override, and no runtime OOM, NCCL, or worker failures.


nemo_rl/utils/nccl_topology.py
```diff
@@ -0,0 +1,462 @@
...
+                replacement=None,
+                nccl_only=(),
+                candidate_only=(),
+                candidate_source=candidate_source,
+                warnings=(
+                    "NCCL_IB_HCA is not an exact include selector; "
+                    "device sets were not compared",
+                ),
+            )
+        return TopologyDiagnostic(
```
seonjinn, line 371: This mismatch is marked as `status="warning"`, but `warnings` remains empty. `main()` only prints entries from `diagnostic.warnings`, so both explicit `warn` and default `auto` with an untrusted UCX list silently continue despite the docs promising a warning. Please populate a mismatch warning here and add a CLI assertion that the warning is emitted.

nemo_rl/utils/nccl_topology.py
```diff
@@ -0,0 +1,462 @@
...
+            candidate_only=(),
+            candidate_source=candidate_source,
+        )
+
+    effective_action = action
+    if action == "auto":
+        has_trusted_candidate = candidate_source == "expected" or repair_source == "ucx"
+        effective_action = "repair" if has_trusted_candidate else "warn"
+
+    if effective_action == "warn":
```
seonjinn, line 346: This branch compares the NCCL and candidate sets without validating the current exact NCCL selection. If both lists contain the same missing HCA or inactive port, `warn` and default `auto` return `status="ok"`, defeating the guard's validation guarantee. Please run `_device_validation_errors` for the exact current selection here, surface those diagnostics as warnings, and cover matching missing/inactive lists.

nemo_rl/utils/nccl_topology.py
```diff
@@ -0,0 +1,462 @@
...
+                status="ok",
+                replacement=None,
+                nccl_only=(),
+                candidate_only=(),
+                candidate_source=None,
+            )
+        raise ValueError("An expected or UCX HCA list is required for comparison")
+
+    candidate = _parse_device_list(candidate_text)
+    current_set = {entry.selection_key for entry in current}
```
seonjinn: The trusted expected-list comparison drops NCCL's rail and plane assignments. Swapping only `:rail:plane` values between otherwise identical HCA ports returns `ok` in both `auto` and `strict`, even though NCCL consumes those fields for topology selection. Please compare full normalized `(name, port, rail, plane)` entries for `candidate_source == "expected"` while retaining device+port semantics for UCX, and cover auto/strict mapping-only mismatches. NCCL reference: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-ib-hca

seonjinn (commented):
Three validated correctness findings in the NCCL topology guard.

nemo_rl/utils/nccl_topology.py
seonjinn: Fixed in e6adc687c: mismatch diagnostics now populate `warnings`, and the CLI regression test verifies that the launcher-visible warning is emitted.

nemo_rl/utils/nccl_topology.py
seonjinn: Fixed in e6adc687c: the warn/default-auto path now validates the current exact NCCL selection even when the UCX list matches, with missing and inactive-port regression coverage.

nemo_rl/utils/nccl_topology.py
seonjinn: Fixed in e6adc687c: trusted expected-list comparisons now include normalized rail/plane assignments, while UCX remains device+port-only. Auto, strict, UCX compatibility, and omitted-field normalization are covered.

seonjinn (commented):
## `/review-pr-team` self-review

I ran the required multi-agent self-review across correctness, design, test coverage, launcher integration, and existing review threads, followed by a devil-advocate pass.

The review found and fixed:

- CLI subprocess tests hid `main()` from Codecov; two behavior tests now call the entry point in-process while one subprocess/sourceability integration test remains.
- Trusted HCA candidates could contain NCCL-invalid plane IDs or too many user-defined planes.
- `strict` could accept a non-exact selector when no comparison source was present.
- Missing or unsupported UCX repair-source configuration could silently bypass the guard.
- Colliding topology ranks could share report paths and atomic writes reused one temporary filename.
- The subprocess test could inherit topology-control variables from the invoking shell.

Validation after the fixes:

- 30 targeted tests pass.
- Targeted module line coverage is 91% locally, up from the CI-observed 74.19% patch coverage.
- Ruff, Pyrefly for the production module, `bash -n ray.sub`, and `git diff --check` pass.

The final devil-advocate pass found no remaining high-confidence correctness blocker.

seonjinn:
Contributor self-review refresh for `1d72c8bebba626184d9fe30b0a8571f5aae7b6e9` after the latest main merge.

The review found and fixed three issues: invalid current HCA metadata could bypass validation when UCX metadata was trusted, sysfs inventory failures escaped instead of producing a structured diagnostic, and the documented UCX fail-loud contract was not fully enforced. Focused NCCL topology tests now pass (38 passed), and Ruff/format/diff checks pass. I found no remaining source-level findings in the reviewed scope.

yuki-97:
hi @youngeunkwon0405 , could you help to take a review?

youngeunkwon0405:
I’m not familiar with the details of the IB HCA configuration, so I may be misunderstanding the intent of this PR. However, if this PR is primarily addressing a performance issue caused by an incorrect cluster configuration, I would expect the cluster team to fix it at the source. Could we first confirm whether the issue has been reported to the cluster team and whether the underlying configuration can be corrected there?

There may still be value in having this as a defense-in-depth safeguard. My only concern is whether it adds to Ray initialization latency, which is already a significant bottleneck at scale and requires further optimization. I think @terrykong would be the right person to review this PR.

seonjinn:
Agreed that the cluster configuration remains the source of truth; this PR is defense-in-depth and is not intended to replace a cluster-side fix. I do not have a closed cluster ticket linked in this PR, so the default `auto` path is deliberately fail-safe: without an explicitly trusted HCA list it only reports or skips and does not infer or overwrite the selection. The check runs once per node before Ray starts, uses only local sysfs reads plus one small per-node JSON/env write, and makes no Ray or collective calls. In the matched 4-node run, `auto` versus `off` kept all steady-state deltas below 0.5%; the PR body now records those numbers. A direct startup-latency measurement would still be useful before treating this as more than a defense-in-depth guard.

seonjinn (commented):
Self-review via `/review-pr-team` (6 agents: rl-expert, bug-finder, test-agent, design-reviewer, comment-reviewer, devil's-advocate). Reviewed head `ffb7200`, merged clean with `main` (`2aafb9d8`, zero conflicts, diff unchanged at 5 files / +1430).

**Verification environment:** macOS laptop, no GPU, no container, no real sysfs, no cluster run. All 38 unit tests were executed for real (`38 passed in 0.37s`) and every behavioral claim below was reproduced in-process against `diagnose_topology` / `main()`. Blast-radius claims come from reading `ray.sub` control flow, not from executing it, and are marked as such.

An earlier draft of this review claimed that a single down IB port kills the job under the default `auto` mode. **That is wrong and I am recording it rather than quietly dropping it.** Under default `auto` with no admin opt-in, a down port and an entirely empty inventory both produce `status="warning"`, rc 0:

```
auto / exact NCCL_IB_HCA / down port / no expected / no ucx  -> warning  ("HCA 'mlx5_0' port 1 is not active",)
auto / exact NCCL_IB_HCA / empty inventory / nothing else    -> warning  ("HCA 'mlx5_0' does not exist",)
auto / NRL_NCCL_EXPECTED_HCAS set / down port                -> error
```

`status="error"` under `auto` is reachable only after an admin opts in via `NRL_NCCL_EXPECTED_HCAS` or `NRL_NCCL_HCA_REPAIR_SOURCE=ucx`, both documented as optional and admin-owned. The fail-closed behavior on those paths is the semantics an admin asks for by declaring an authoritative list, and I am not reporting it as a defect.

---

## 1. MEDIUM — the opt-out knob fails closed on whitespace

`NRL_NCCL_TOPOLOGY_ACTION` is lowercased at [L469](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L469); `NRL_NCCL_HCA_REPAIR_SOURCE` is not, at [L473](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L473). Neither is stripped. Verified end-to-end through `main()`:

```
NRL_NCCL_TOPOLOGY_ACTION='off'   -> rc 0
NRL_NCCL_TOPOLOGY_ACTION='OFF'   -> rc 0
NRL_NCCL_TOPOLOGY_ACTION='off '  -> rc 2   "Unsupported topology action: 'off '"
NRL_NCCL_HCA_REPAIR_SOURCE=ucx   -> proceeds
NRL_NCCL_HCA_REPAIR_SOURCE=UCX   -> rc 2   "Unsupported HCA repair source: 'UCX'"
```

A single trailing space out of a sourced site config makes the documented escape hatch abort the job it was invoked to disable. This is the one place where a fail-closed guard makes its own kill switch fail closed, and it is the highest-value fix in this review. `inventory_required` at [L481-L490](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L481-L490) also compares `repair_source == "ucx"` independently, so an uppercase value silently takes a different inventory branch as well.

Fix: `.strip().lower()` on both reads. Separately worth deciding whether an unparseable action string should abort an allocation at all rather than falling back to `auto` with a warning.

## 2. MEDIUM — `docs/cluster.md` never says `auto` can fail the job

[`docs/cluster.md:175-182`](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/docs/cluster.md#L175-L182) describes `auto` as repairing "only when the cluster provides a trusted device list; otherwise it warns or skips without changing the environment." The `NRL_NCCL_EXPECTED_HCAS` entry documents only the success path. Neither tells an operator that `auto` can hard-fail bringup.

Reproduced: `auto` + `NRL_NCCL_EXPECTED_HCAS='=mlx5_9:1'` -> `status="error"` -> rc 2 -> `ray.sub` aborts. Realistic trigger: an admin sets the allowlist once in a site profile; months later an HCA is renamed by a firmware update or a port goes link-down, and every job on the *default* action refuses to launch while the docs say `auto` does not change anything.

The behavior is defensible; the documentation is the actual defect. Add a sentence to the `NRL_NCCL_TOPOLOGY_ACTION` entry stating that `auto` fails the job before Ray starts when a trusted list names a device or port that does not exist or is not active.

## 3. MEDIUM — `ray.sub` conflates "could not run the check" with "topology is invalid"

[`ray.sub#L581-L588`](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/ray.sub#L581-L588) uses `if ! python -m nemo_rl.utils.nccl_topology ...`, so every nonzero status lands in the same branch: rc 127 (no interpreter), rc 1 (any import-time traceback), rc 2 from argparse on a malformed invocation, and rc 2 from a genuine diagnosis. The failure message then names a report file that was never written.

This is a diagnosability bug, not a new way to kill a job. The cited trigger — cwd not being the repo, so `nemo_rl` does not import — is already fatal upstream: `ray.sub:101` resolves `COMMAND` relative to `SLURM_SUBMIT_DIR` and `:203` defaults `BASE_LOG_DIR` there, so a wrong submit dir means the driver command is unresolvable anyway. What this block adds is that the job now dies earlier with `[ERROR] NCCL topology validation failed` for what is actually a `ModuleNotFoundError`.

Fix: capture the status; `-eq 3` (see below) aborts, any other nonzero warns and continues. Note `main()` returns 2 ([L547](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L547)) and argparse also exits 2, so branching on the code requires moving the diagnostic exit to 3.

Blast radius, from reading control flow (not executed): the worker probe is sourced at `:943`; `exit 1` kills that task, the `ray-workers` srun exits, and the submit-host loop at `:1117-1125` detects it, touches `ENDED` and breaks. The head probe at `:735` runs before `touch STARTED_RAY_HEAD` at `:802`. One node aborting does take the allocation down, so the branch is worth getting right.

## 4. SUGGESTION — missing module docstring and reciprocal sync comment

`nccl_topology.py` has no module docstring (`ast.get_docstring` returns `None`; the file goes from the license header straight to `import argparse`), yet `main()` passes `description=__doc__` to `argparse` at [L462](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L462) — so `--help`, the only in-tree documentation an operator debugging a node-startup abort will reach for, prints no description.

Separately, the repo has an explicit bidirectional convention for exactly this shell-to-Python coupling, used twice within 60 lines of the new block: `ray.sub:531-533` (`NRL_GPU_CPU_AFFINITY_FILE` <-> `numa_utils.py`) and `:596-598` (`nvlink_domain_`/`topo_rank` <-> `virtual_cluster.py`), each with a `# IMPORTANT: ... must stay in sync with ...` comment on both sides. The new block has neither direction. Because the module is invoked as `python -m` from a heredoc, nothing statically links the two: renaming a variable produces no CI failure, only a runtime failure on a compute node.

One commit fixes all of it: add the module docstring (naming `ray.sub` as the caller and `docs/cluster.md` as the documented surface), add a `Raises:` section to `diagnose_topology`'s docstring — it raises `ValueError` at L204/214/216/296/382 and its only caller must catch it at L513 — and add the reciprocal comment in `ray.sub`.

## 5. LOW-MEDIUM — report and env-file writes are unguarded

`main()` wraps the sysfs read in `try/except OSError` at [L491-L502](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L491-L502), but both `_write_text_atomically` calls at [L529-L541](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L529-L541) are not. Pointing `--report` at an unwritable directory produces an uncaught `PermissionError` traceback and rc 1.

Scoped honestly: `ray.sub` already validates `$LOG_DIR` with a `.shared_fs_canary` at `:216`, checked at `:645`/`:879`, so an unwritable log dir is a hard abort upstream of this code, and a `$LOG_DIR` that hits ENOSPC kills the job regardless since the Ray logs and the log-sync rsync all live there. So the marginal harm is an unhandled traceback instead of a clean message. Worth fixing anyway: warn-only for the report write, `return` non-zero only for the env-file write when a replacement exists.

## 6. LOW — one unreadable device directory discards the whole inventory

The per-port `state` read is guarded at [L167-L170](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L167-L170), but the `iterdir()` calls in `read_rdma_inventory` are not. A mode-000 device subdirectory raises `OSError`, which `main()`'s handler turns into `status="error"` for the entire node.

Noting this specifically because, unlike the opt-in paths above, this one *can* abort under default `auto` with only `NCCL_IB_HCA` set and no admin opt-in — it is the only such condition I found. The trigger is rare, hence LOW. Fix: wrap the per-device body, skip the offending device, record a warning.

## 7. SUGGESTION — `inventory_required` duplicates the skip predicate

[L481-L490](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L481-L490) is a clause-for-clause copy of the auto-skip at L217-L222 plus the precondition at L215-L216, straddling the pure/impure boundary. The copies currently agree, but finding 1's uppercase divergence shows how they can disagree in practice. Extracting a single `requires_inventory(...)` used by both `main()` and `diagnose_topology` is ~10 lines and does not disturb the branch structure.

I deliberately did **not** conclude that the ~220-line `diagnose_topology` should be split. The action x candidate-source x exactness matrix is the problem's real shape, each of its return sites is a distinct outcome, and a tri-split would move complexity into parameter threading without removing a branch.

---

## Tests — four worth adding

The suite is well constructed and mostly non-tautological: 38 tests, all passing, 90% branch coverage. Rather than trust coverage, I ran **18 targeted mutations of the production file against the unchanged suite; 10 survived**. The survivals cluster in one place — the trust/authority state machine and the exit-code contract `ray.sub` consumes.

**In every case below the production code is correct and I verified it directly. These are unpinned invariants, not shipped bugs.**

```
port mismatch (strict, =data0:1 vs =data0:2)      -> error   correct
rail mismatch (strict, =data0:1:0 vs =data0:1:1)  -> error   correct
repair + untrusted ucx, no repair_source          -> ValueError("Repair requires NRL_NCCL_HCA_REPAIR_SOURCE=ucx")   correct
```

1. **The untrusted-UCX repair guard at [L379-L382](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L379-L382) has zero coverage.** Replacing it with `if effective_action != "repair":` leaves the suite fully green. This is the invariant the whole trust model rests on — an undeclared `UCX_NET_DEVICES` must never rewrite `NCCL_IB_HCA`. `test_enabled_actions_require_declared_ucx_source_to_be_present` looks like it covers this but matches `"NRL_NCCL_HCA_REPAIR_SOURCE=ucx requires UCX_NET_DEVICES"`, which is raised at L215-216, ~165 lines earlier; execution never reaches L379.

2. **`topology_key`'s port and rail components are pinned by no test.** Mutating `port = ""` and `rail = -1` at [L54-L59](https://github.com/NVIDIA-NeMo/RL/blob/ffb7200962d687a89829cbefa4fafe3a01c08173/nemo_rl/utils/nccl_topology.py#L54-L59) both survive; only `plane = -1` is caught. `test_strict_rejects_trusted_expected_rail_and_plane_mismatch` compares `=data0:1:0:1,data1:1:0:0` against `=data0:1:0:0,data1:1:0:1` — rail is `0` in all four entries, so only plane varies. The name promises coverage the test does not deliver; same defect in the sibling at test L162. Worth fixing the assertion and the name together.

3. **No test proves `strict` can pass.** All five `strict` tests assert `status == "error"`; the `ok` return sites at L288-L295 and L333-L339 are uncovered. A regression making `strict` reject everything would kill every node with CI green. (`"HCA {name!r} has no active port"` at L117-L118 is likewise produced by no test — mutating it to `if False:` survives.)

4. **No error path is ever run as a subprocess.** `raise SystemExit(main())` at L551 is uncovered and both exit-2 tests call `main()` in-process, so nothing proves `python -m nemo_rl.utils.nccl_topology` actually *exits* 2 — which is exactly what `ray.sub` keys on. One subprocess test on the error path covers it; pinning `schema_version` in the same test is worth a sub-assertion, since it exists so consumers can pin it and nothing currently does.

Test hygiene is clean and I verified it fresh rather than assuming: running the suite with all five env vars exported to poison values gives 38 passed, `--reverse` gives 38, and three independent `pytest-randomly` seeds give 38 each. No test touches the real `/sys` or changes cwd.

PR #3671, opened 2026-08-16
fix(sc): mask sequence logprob outliers

mehraakash:
# What does this PR do?

Adds sequence-level generation/training log-probability mismatch filtering to the SingleController GRPO path, matching the protection already available in the legacy GRPO path.

- Fetches generation log-probabilities when `grpo.seq_logprob_error_threshold` is configured.
- Applies the shared sequence-level mismatch calculation before advantage estimation.
- Writes the filtered `sample_mask` back to the data plane so masked sequences do not contribute to training.
- Aggregates mismatch and masking metrics correctly across streaming chunks.
- Skips advantage normalization and F/B for a streaming chunk with no valid loss tokens, while allowing later valid chunks in the same logical step to train.
- Fails before optimizer/scheduler advancement if an entire logical step contains no valid loss tokens.

When `grpo.seq_logprob_error_threshold` is `null`, SC does not fetch generation log-probabilities or apply the sequence mismatch mask. The empty-loss-token guard applies regardless of the threshold: a pre-existing all-zero chunk is consumed without F/B, and an entirely empty logical step fails before optimizer/scheduler advancement.

# Issues

SingleController previously accepted `seq_logprob_error_threshold` in the shared GRPO configuration but did not apply the mask before streaming training. Large generation/training log-probability disagreements could therefore affect V2 training even when the protection was configured.

# Usage

```yaml
grpo:
  seq_logprob_error_threshold: 2.0
```

Sequences whose mean multiplicative probability error exceeds the threshold are removed from the loss, and the following metrics are emitted at the logical-step level:

- `max/mean/min_seq_mult_prob_error`
- `max/mean/min_seq_mult_prob_error_after_mask`
- `num_masked_seqs_by_logprob_error`
- `masked_correct_pct`

# Validation

- 32 focused SingleController and metric-reduction unit tests passed at the current PR head.
- Added regression coverage for partial masking, an entirely masked streaming chunk, a pre-existing all-zero mask with the threshold disabled, an entirely invalid logical step, and an empty chunk followed by a valid chunk.
- Ruff check and format passed.
- Pyrefly, recipe minimization, and all applicable pre-commit hooks passed.
- A 100-step Nano 3.5 SingleController in-order/lookahead-1 run completed and wrote checkpoint 100 with this masking path enabled.

## Matched Nano 3.5 comparison

The table compares independent runs of the same V2 recipe, node geometry, GBS 2048, in-order/lookahead-1 sampler, and steps 26–100. The baseline used the pre-fix SC path, where the configured threshold was accepted but not applied; the masked run used threshold `2.0`. Acceptance was no greater than a 5% regression in mean reward, entropy, step time, or valid-token throughput, together with removal of pathological policy-KL outliers.

| Metric (steps 26–100) | Pre-fix SC | Masked SC | Change |
|---|---:|---:|---:|
| `train/reward` mean | 0.96905 | 0.97078 | +0.18% |
| `train/approx_entropy` mean | 0.68068 | 0.67070 | -1.47% |
| `train/policy_kl_error` median / max | 0.00314 / 35.98457 | 0.00185 / 0.00644 | outlier removed |
| `train/gen_kl_error` mean | 0.001796 | 0.001618 | -9.94% |
| `train/is_oob_ratio` mean | 2.679e-5 | 2.478e-5 | -7.50% |
| `timing/train/total_step_time` mean | 702.77 s | 718.32 s | +2.21% |
| `timing/train/valid_tokens_per_sec_per_gpu` mean | 69.95 | 71.73 | +2.54% |
| sequences masked | not applied | 86 / 153,600 (0.056%) | — |

The matched window shows no reward or throughput regression outside the 5% bound, while the rare policy-KL failure is contained. This PR is marked `CI:L2` because the change affects which sequences contribute to gradients.

### DIFF
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -47,7 +47,11 @@ import ray
 import torch
 
 from nemo_rl.algorithms.async_utils.staleness_sampler import create_sampler
-from nemo_rl.algorithms.grpo import GRPOSaveState, _write_latest_checkpoint_status
+from nemo_rl.algorithms.grpo import (
+    GRPOSaveState,
+    _write_latest_checkpoint_status,
+    compute_and_apply_seq_logprob_error_masking,
+)
 from nemo_rl.algorithms.metric_utils import SetupTimingMetrics
 from nemo_rl.algorithms.single_controller_utils.config import (
     AdvantageConfig,
@@ -255,6 +259,7 @@ class SingleControllerActor:
             "rewards": [],
             "masked_advantages": [],
             "sequence_lengths": [],
+            "seq_logprob_error_metrics": [],
         }
 
         print(
@@ -963,22 +968,29 @@ class SingleControllerActor:
 
                     # Compute advantages
                     with self._timer.time("advantage_calculation"):
-                        train_meta = await self._advantage_stage(train_meta)
+                        (
+                            train_meta,
+                            has_valid_training_tokens,
+                        ) = await self._advantage_stage(train_meta)
 
-                    # Train
+                    # Filtering can leave a streaming chunk with no training tokens.
+                    # Consume that chunk without F/B, then continue the same optimizer
+                    # step with the next chunk. Always restore training mode because
+                    # log-prob inference may have switched the model to inference mode.
                     with self._timer.time("training_prep"):
                         await asyncio.to_thread(self._trainer.prepare_for_training)
-                    with self._timer.time("policy_training"):
-                        if not step_open:
+                    if has_valid_training_tokens:
+                        with self._timer.time("policy_training"):
+                            if not step_open:
+                                await asyncio.to_thread(
+                                    self._trainer.begin_train_step,
+                                    self._loss_fn,
+                                )
+                                step_open = True
                             await asyncio.to_thread(
-                                self._trainer.begin_train_step,
-                                self._loss_fn,
+                                self._trainer.train_microbatches_from_meta,
+                                train_meta,
                             )
-                            step_open = True
-                        await asyncio.to_thread(
-                            self._trainer.train_microbatches_from_meta,
-                            train_meta,
-                        )
 
                     if train_meta.sequence_lengths:
                         self._step_log_dict["sequence_lengths"].extend(
@@ -1046,6 +1058,13 @@ class SingleControllerActor:
                     chunks_dispatched,
                     groups_dispatched,
                 )
+                if not step_open:
+                    raise RuntimeError(
+                        "SingleController has no valid response tokens after "
+                        "filtering. Check grpo.seq_logprob_error_threshold to "
+                        "avoid an optimizer step with an empty batch."
+                    )
+
                 with self._timer.time("policy_training"):
                     result = await asyncio.to_thread(self._trainer.finish_train_step)
 
@@ -1175,8 +1194,8 @@ class SingleControllerActor:
                 print(f"  • {k}: {v:.2f}s ({percent:.1f}%)")
 
             # TODO: per-step train_data jsonl dump, vllm metrics logger,
-            #   histogram log, rollout_metrics, seq_logprob_error_metrics,
-            #   pretty-print "Training Results" block, print_performance_metrics.
+            #   histogram log, rollout_metrics, pretty-print "Training Results"
+            #   block, print_performance_metrics.
             print(f"step_metrics={step_metrics}", flush=True)
             self._logger.log_metrics(
                 step_metrics, step=self._train_steps, prefix="train"
@@ -1656,7 +1675,7 @@ class SingleControllerActor:
         self._rollout_permitted.set()
         return aborted_stale_inflight_groups
 
-    async def _advantage_stage(self, meta: KVBatchMeta) -> KVBatchMeta:
+    async def _advantage_stage(self, meta: KVBatchMeta) -> tuple[KVBatchMeta, bool]:
         """Fetch advantage inputs, compute advantages, and write them back.
 
         SC owns the prompt-group-scoped advantage stage because the selected
@@ -1664,9 +1683,13 @@ class SingleControllerActor:
         DP sharding. Tensor payloads still move through DataPlane: SC fetches
         only the configured advantage input columns and writes the computed
         ``advantages`` column back under the same ``sample_ids``.
+
+        Returns:
+            The updated batch metadata and whether the batch contains at least
+            one valid training token.
         """
         if self._advantage_estimator is None:
-            return meta
+            return meta, True
         adv_cfg = self._advantage_cfg
 
         data = await self._call_dp(
@@ -1684,6 +1707,51 @@ class SingleControllerActor:
         sample_mask = squeeze_trailing_unit_dim(
             tensor_field(data, adv_cfg.sample_mask_field)
         ).float()
+
+        seq_logprob_error_threshold = (
+            self._master_config.grpo.seq_logprob_error_threshold
+        )
+        # Match the legacy path: whenever real policy logprobs are available,
+        # report sequence-level generation/training mismatch. A threshold adds
+        # masking; leaving it unset keeps this metrics-only.
+        if self._policy_logprobs_required:
+            masking_data = BatchedDataDict(
+                {
+                    "token_mask": token_mask,
+                    "sample_mask": sample_mask,
+                    "prev_logprobs": tensor_field(
+                        data,
+                        adv_cfg.policy_logprobs_field,
+                    ),
+                    "generation_logprobs": tensor_field(
+                        data,
+                        adv_cfg.generation_logprobs_field,
+                    ),
+                }
+            )
+            num_valid_seqs_before = float(
+                ((token_mask[:, 1:] * sample_mask.unsqueeze(-1)).sum(dim=-1) > 0)
+                .sum()
+                .item()
+            )
+            seq_error_metrics = compute_and_apply_seq_logprob_error_masking(
+                train_data=masking_data,
+                rewards=rewards,
+                seq_logprob_error_threshold=seq_logprob_error_threshold,
+            )
+            sample_mask = masking_data["sample_mask"]
+            num_valid_seqs_after = float(
+                ((token_mask[:, 1:] * sample_mask.unsqueeze(-1)).sum(dim=-1) > 0)
+                .sum()
+                .item()
+            )
+            seq_error_metrics["num_masked_seqs_by_logprob_error"] = (
+                seq_error_metrics.pop("num_masked_seqs")
+            )
+            seq_error_metrics["_num_valid_seqs_before"] = num_valid_seqs_before
+            seq_error_metrics["_num_valid_seqs_after"] = num_valid_seqs_after
+            self._step_log_dict["seq_logprob_error_metrics"].append(seq_error_metrics)
+
         mask = token_mask * sample_mask.unsqueeze(-1)
 
         repeated_batch: dict[str, torch.Tensor] = {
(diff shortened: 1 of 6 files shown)


nemo_rl/algorithms/single_controller.py
```diff
@@ -1048,6 +1053,48 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> KVBatchMeta:
...
+                .item()
+            )
+            seq_error_metrics["num_masked_seqs_by_logprob_error"] = (
+                seq_error_metrics.pop("num_masked_seqs")
+            )
+            seq_error_metrics["_num_valid_seqs_before"] = num_valid_seqs_before
+            seq_error_metrics["_num_valid_seqs_after"] = num_valid_seqs_after
+            self._step_log_dict["seq_logprob_error_metrics"].append(seq_error_metrics)
+
         mask = token_mask * sample_mask.unsqueeze(-1)
```
mehraakash, line 1755: **[P1] Handle a streaming chunk whose post-filter mask is empty.** If `seq_logprob_error_threshold` masks every sequence in this chunk, `mask` contains no valid loss tokens but the code still calls the advantage estimator and later runs F/B. `ReinforcePlusPlusAdvantageEstimator` can normalize by a zero mask and produce NaNs; estimators that tolerate it can still allow `finish_train_step()` to advance optimizer/scheduler state with no valid data. Please skip advantage/F/B for an empty chunk while continuing later chunks in the same logical step, and fail before `finish_train_step()` if every chunk in the logical step was empty. Regression tests should cover both all-empty and empty-then-valid streaming chunks.

mehraakash (commented):
The review found one blocking correctness case in the new SingleController masking path. I am addressing it before asking CI to run.

nemo_rl/algorithms/single_controller.py
mehraakash: Fixed in b1578ede61276e8d3b950dc12c1e10b87ea95de7. `_advantage_stage` now checks the exact loss-token mask (`mask[:, 1:]`), bypasses advantage normalization for an empty chunk, and tells the train pump not to run F/B for it. Later valid chunks in the same logical step still train; if the entire step is empty, the controller raises before `finish_train_step()` so optimizer/scheduler state cannot advance. `prepare_for_training()` is still called to restore model mode after log-prob inference. Added regression coverage for both all-empty and empty-then-valid streaming cases. Validation: 31 focused tests passed; Ruff check/format, Pyrefly, config minimization, and all applicable pre-commit hooks passed (Taplo skipped because no TOML changed and the test image has no Rust toolchain).

nemo_rl/algorithms/single_controller_utils/utils.py
```diff
@@ -120,6 +123,83 @@ def reduce_advantage_pump_metrics(
...
+        )
+        after_count = sum(
+            record["_num_valid_seqs_after"] for record in seq_logprob_error_metrics
+        )
+        masked_count = sum(
+            record["num_masked_seqs_by_logprob_error"]
+            for record in seq_logprob_error_metrics
+        )
+
+        before_records = [
```
mehraakash: This seems like it could be abstracted away in a function and can be replaced with a function call. It makes the code verbose.

nemo_rl/algorithms/single_controller.py
```diff
@@ -591,22 +575,32 @@ async def _train_pump(self) -> None:
...
+                            has_valid_training_tokens,
+                        ) = await self._advantage_stage(train_meta)
+
+                    # A streaming chunk may be entirely removed by sequence-level
+                    # filtering. Consume it, but do not send an empty batch through
+                    # advantage normalization or F/B; later chunks in this logical
+                    # step may still contain valid training tokens.
+                    # Restore training mode even for an empty chunk: logprob
+                    # inference may have switched modes after an earlier valid
+                    # chunk opened this logical train step.
```
mehraakash: Can be worded better.

nemo_rl/algorithms/single_controller_utils/utils.py
mehraakash: Addressed in 5f3ab8ae2b2e5208c426583425d73ff3242c2176. The caller is now a single `out.update(_reduce_seq_logprob_error_metrics(...))`; the helper owns the before/after-mask min, weighted-mean, max, count, and masked-correct aggregation. The focused 31-test suite and all applicable pre-commit hooks pass.

nemo_rl/algorithms/single_controller.py
mehraakash: Addressed in 5f3ab8ae2b2e5208c426583425d73ff3242c2176. The comment now states the mechanism directly: filtering may leave one streaming chunk empty; consume it without F/B, continue the same optimizer step with the next chunk, and always restore training mode after log-prob inference.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1283,6 +1359,8 @@ def _advantage_input_fields(self) -> list[str]:
         ]
         if self._policy_logprobs_required:
             fields.append(adv_cfg.policy_logprobs_field)
+        if self._master_config.grpo.seq_logprob_error_threshold is not None:
+            fields.append(adv_cfg.generation_logprobs_field)
```
saumishr, line 1825: **[P2] No shipped SingleController config enables this, so `CI:L2` will not exercise the new path.**

`grpo.seq_logprob_error_threshold` is `null` in `examples/configs/grpo_math_1B.yaml:50`, and `grpo_math_1B_megatron_single_controller.yaml` inherits it through `defaults`. Grepping `examples/` for a non-null value returns only the nemo_gym configs, two VLM recipes and `mopd-qwen3-1.7b` — none of the single-controller recipes. In `tests/functional/`, the only script that sets the knob at all is `ppo_megatron_non_colocated.sh`. So the functional and convergence tests that this PR's `CI:L2` label triggers never take the `threshold is not None` branch; the coverage is the new unit tests plus your manual 100-step run.

There is a cheap precedent for closing that gap without perturbing convergence: `tests/functional/ppo_megatron_non_colocated.sh:30` sets `ppo.seq_logprob_error_threshold=1000`, high enough that nothing is ever masked, purely so the plumbing runs. The same no-op-high value on an SC functional test would exercise the extra `generation_logprobs` column fetch, the helper call, the `sample_mask` write-back and the per-chunk metric reduction here.

Worth being explicit that a high threshold still leaves the empty-chunk skip and the new `RuntimeError` unit-test-only, since neither can trigger when nothing is masked.

nemo_rl/algorithms/single_controller.py
```diff
@@ -648,6 +660,14 @@ async def _train_pump(self) -> None:
 
                     groups_dispatched += num_groups
 
+                if not step_open:
+                    raise RuntimeError(
+                        "SingleController has no valid response tokens after "
+                        "filtering. Check overlong filtering and "
+                        "grpo.seq_logprob_error_threshold to avoid an optimizer "
+                        "step with an empty batch."
+                    )
```
saumishr, line 1066: **[P2] This message points the operator at a knob the SingleController path never reads.**

`grpo.overlong_filtering` is read in `grpo.py` (3118, 4685), `grpo_sync.py:778` and `ppo.py:1375`, but there is no reference to it anywhere in `single_controller.py`, `single_controller_utils/`, or the SC payload builder, and `validate_single_controller_config` neither rejects nor warns on it. An operator who hits this error and follows the advice will find nothing to change.

On this path the threshold is in fact the only mechanism that can zero a `sample_mask`, because the SC payload builder hardcodes `sample_mask = torch.ones(n)` instead of deriving it from `loss_multiplier` the way `collate_fn.py:214` and `grpo.py:3153` do. Dropping the clause makes the message shorter and strictly more accurate:

```suggestion
                    raise RuntimeError(
                        "SingleController has no valid response tokens after "
                        "filtering. Check grpo.seq_logprob_error_threshold to "
                        "avoid an optimizer step with an empty batch."
                    )
```

Separately and out of scope here: that `grpo.overlong_filtering` is silently ignored on the SC path at all looks like a pre-existing gap worth a follow-up issue.

saumishr (commented):
Reviewed at `14aaeb6` with a bug-scan, v1-parity, test-coverage and metric-math pass, followed by an adversarial pass over the findings.

**This needs a rebase before it can go anywhere.** GitHub reports the PR as conflicting (`mergeable_state: dirty`), 9 commits behind `main` and diverged. #3665 landed on `main` in the meantime and touches both `single_controller.py` and `single_controller_utils/config.py`, so expect the conflict there.

Three findings below, all P2/P3 — I did not find a blocking correctness defect. The substance of the change looks right, and the A/B table in the description is the kind of evidence that makes this easy to review.

Things I checked and found sound, so you don't have to re-litigate them:

- **`prev_logprobs` is always available when the masking block runs.** `_policy_logprobs_required = not (force_on_policy_ratio and threshold is None)` (line 122) is `True` for *both* values of `force_on_policy_ratio` once the threshold is set, so the unconditional read at 1260 cannot miss the column. v1's `skip_prev_logprobs` placeholder path has no SC analog that needed porting.
- **The unguarded `.pop("num_masked_seqs")` at 1286 is safe.** The helper's return is a single unconditional dict literal (`grpo.py:2613`) and `num_masked_seqs` is initialised to `0` before the threshold branch, so v1's `if "num_masked_seqs" in ...` guard is redundant defensiveness rather than a contract you are breaking.
- **The chunked metric reduction is mathematically equivalent to v1's whole-step computation.** `_num_valid_seqs_before` counts exactly the helper's `valid_seq_mask` population and `_num_valid_seqs_after` exactly its `kept_mask`, so both weighted means recover the true step-level mean; weighting `masked_correct_pct` by `num_masked_seqs_by_logprob_error` recovers `total_correct / total_masked`; and filtering min/max on `count > 0` drops the helper's `0.0` sentinels without excluding a genuine `0.0`. The unit test pins the weighting with *unequal* weights (4/6 and 3/4), so an unweighted mean would fail it — that is the right test to have written.
- **The tensor shapes line up.** The helper slices `token_mask`, `prev_logprobs` and `generation_logprobs` identically at `[:, 1:]`, and the new `_advantage_stage` test drives the *real* helper rather than a mock and pins `max_seq_mult_prob_error == e`, which a misalignment would move.
- **`step_open` resets per logical step** (line 515), so the new guard is correctly scoped, and `_create_advantage_estimator` has no path returning `None`, so the `is None` early return cannot silently disable masking on a real run.
- The empty-step `RuntimeError` does newly apply to threshold-unset runs, but reaching it needs every row of every chunk to have zero trainable tokens, and the prior behaviour was a silent optimizer step on an all-masked batch. Fail-loud is the better default, so I am not raising that as a finding.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1224,6 +1248,48 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> KVBatchMeta:
         sample_mask = squeeze_trailing_unit_dim(
             tensor_field(data, adv_cfg.sample_mask_field)
         ).float()
+
+        seq_logprob_error_threshold = (
+            self._master_config.grpo.seq_logprob_error_threshold
+        )
+        if seq_logprob_error_threshold is not None:
```
saumishr: **[P3] SC only reports these metrics once a threshold is already set, so there is no way to choose one from an SC run.**

v1 gates the helper call on `skip_prev_logprobs` rather than on the threshold (`grpo.py:3253-3261`) and passes the threshold through as-is. The helper wraps only the masking in `if seq_logprob_error_threshold is not None:` (`grpo.py:2563`), so a `None` threshold yields metrics-with-no-masking, which `grpo.py:3482` logs every step under the comment "Always log sequence-level error metrics (useful for deciding threshold)". Gating the whole block here (and the field append at 1362) on the threshold means an SC operator has to read the error distribution off a v1 run before they can pick a value for an SC one.

Matching v1 is cheaper than it first looks: `prev_logprobs` are already computed on every SC step unless `loss_fn.force_on_policy_ratio` is set, so a metrics-only mode adds one DataPlane column read of the already-materialised `generation_logprobs`, not a forward pass. Either a metrics-only mode (behind its own flag if the column read is unwelcome) or a sentence in the SC docs recording the divergence would close it. Low severity — raising it for the parity decision, not as a defect.

nemo_rl/algorithms/single_controller.py
mehraakash: Addressed in 029475273. The fast SC functional test now sets grpo.seq_logprob_error_threshold=1000 and asserts that sequence-error metrics are emitted while zero sequences are masked. The all-empty chunk/step cases remain covered by focused unit tests, as noted.

nemo_rl/algorithms/single_controller.py
mehraakash: Addressed in 029475273. I removed the overlong-filtering advice; the RuntimeError now points only to grpo.seq_logprob_error_threshold.

nemo_rl/algorithms/single_controller.py
mehraakash: Addressed in 029475273 by matching V1 semantics. When real policy logprobs are available, SC now reads the already-materialized generation_logprobs and reports sequence-error metrics even with a null threshold; masking and sample_mask writeback still occur only when a threshold is set. Added a unit test for metrics-only behavior.

saumishr:
Thanks for addressing the comments. Just a note for other reviewers that we have also validated this with a prod run.

saumishr (approved):

PR #3677, opened 2026-08-17
feat(trtllm): support Qwen3.5 routed-expert block-FP8 refit

shikicloud:
# What does this PR do ?

Enable Qwen3.5 MoE rollout workers to use routed-expert block-FP8 weights with TRT-LLM while policy training remains in BF16. This feature does not change optimizer-state precision.

During each BF16 trainer-to-generation refit, NeMo RL:

- Validates the fused Qwen3.5 expert layout: gate_up_proj=[E,2I,H] and down_proj=[E,H,I].

- Splits the fused expert tensors into the per-expert gate_proj, up_proj, and down_proj weights expected by the TRT-LLM Qwen3.5 mapper.

- Quantizes only routed-expert weights to E4M3 using 128x128 blocks and FP32 weight_scale_inv tensors.

- Keeps attention, routers, shared experts, embeddings, and the output head in BF16.

- Applies the conversion to both collective and IPC/ZMQ refit transports.

- Requires TRT-LLM's incremental-refit lifecycle and fails during setup when the required APIs are unavailable.

- Marks a worker unusable after a failed partial FP8 refit, preventing generation with mixed old and new weights.

- Checks the update result from every TRT-LLM worker instead of only the first worker.

- Delays async rollout collection until the initial FP8 refit is complete, without changing startup behavior for non-FP8 backends.

This implementation currently supports only Transformers models with model_type=qwen3_5_moe. It uses TRT-LLM's MoE backend to preserve FP32 block scales; MXFP8/E8M0 scaling is not used.

# Issues
None.

# Usage
* Enable Qwen3.5 routed-expert FP8 rollout refit with:

```python
policy:
  generation:
    backend: trtllm
    trtllm_cfg:
      precision: fp8
```
The source policy checkpoint is BF16. Direct loading of a pre-quantized ModelOpt FP8 checkpoint is outside the scope of this PR.

The installed TRT-LLM build must provide the Qwen3.5 mapper and incremental-refit lifecycle used by this path.

# How to test

Run the focused Qwen3.5 TRT-LLM FP8 refit unit test:

```python
uv run --group test pytest -q \
  tests/unit/models/generation/trtllm/test_trtllm_fp8.py
```

This test covers the fused expert layout, E4M3 block quantization and FP32 scales, routed-expert filtering, non-aligned shapes, and Qwen3.5-only configuration guards.

**Pre checks**:

**Additional Information**:

*Scope boundaries*:

- Qwen3.5 MoE only.

- Routed experts only; non-expert modules remain BF16.

- E4M3 128x128 block FP8 with FP32 scales, not MXFP8/E8M0.

- Direct ModelOpt FP8 checkpoint loading is not included.

- This PR does not add KV-cache recomputation semantics for in-flight weight updates.

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -4047,6 +4047,10 @@ def async_grpo_train(
     # SGLang async rollouts do not support the async GRPO replay path.
     generation_config = master_config.policy["generation"]
     backend = generation_config.get("backend", "") if generation_config else ""
+    requires_initial_fp8_refit = (
+        backend == "trtllm"
+        and generation_config.get("trtllm_cfg", {}).get("precision") == "fp8"
+    )
     assert backend in ("vllm", "megatron", "trtllm") and _should_use_async_rollouts(
         master_config
     ), (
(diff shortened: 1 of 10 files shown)


nemo_rl/algorithms/grpo.py
```diff
@@ -4238,13 +4242,12 @@ def async_grpo_train(
...
     )
 
-    # Start trajectory collection in background
-    collection_task = trajectory_collector.start_collection.remote(dataloader)
-
-    # Ensure collector knows initial weight version
-    trajectory_collector.set_weight_version.remote(weight_version)
-
-    print("📦 Started continuous background trajectory collection")
+    if not requires_initial_fp8_refit:
```
shuyixiong, line 4245: Could you explain why we need this in fp8 rollout?

shikicloud: FP8 is initialized with load_format="dummy", so it has no usable model weights until the first refit. I can change the name to `requires_initial_weight_load`

PR #3683, opened 2026-08-17
feat(sc): Add SingleController central agent loop for NeMo RL

rohitrango:
## What changed

- Add a configurable central agent loop for NeMo Gym rollouts in the SingleController path.
- Add rollout-tree, tool-call, steering, and termination helpers for the central agent.
- Route central-agent rollouts through a dedicated rollout implementation while retaining existing post-processing and metrics.
- Add configuration documentation and unit coverage for the central-agent and rollout-manager paths.
- Automatically skip reference-policy log-probability calculation when the SingleController KL penalty is disabled.

## Why

This lets NeMo RL own the NeMo Gym agent loop directly instead of starting a separate agent-server process. It keeps the existing NeMo Gym verification and rollout processing behavior while allowing central control over turns, tool execution, steering, and termination.

All components are interface-driven: research workflow is streamlined by adding custom interfaces for termination management, context compaction, steering messages, and serial / parallel tool calls. Most applications do not require modifying the central loop at all.

This is the baseline control flow

image

With the proposed central agent with pluggable components, we can intercept all the `response_api_agents` servers and only use Gym (or other third-party sandboxes) to interact with the sandboxes.

image

## User impact

Users can enable the path with `env.nemo_gym.central_agent.enabled: true`. Existing agent-server rollouts remain the default.

## Validation

Tested on Qwen-4B with (green) and without (red) central agent. 

image


nemo_rl/environments/central_agent_helpers/config.py
```diff
@@ -0,0 +1,121 @@
...
+        if implementation not in SUPPORTED_AGENT_IMPLEMENTATIONS:
+            raise ValueError(
+                f"env.nemo_gym.central_agent is enabled, but agent instance "
+                f"'{top_key}' uses the '{implementation}' implementation, whose loop "
+                f"the central agent does not reproduce. Supported implementations: "
+                f"{list(SUPPORTED_AGENT_IMPLEMENTATIONS)}. Disable central_agent to run "
+                f"this agent through its own NeMo-Gym server."
+            )
+
+        registry[top_key] = {
```
rohitrango, line 105: [`nemo_rl/environments/central_agent_helpers/config.py:105`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L105)

**[BUG]** The registry entry is built from `resources_server` / `model_server` / `central_agent` only, so the agent block's own `max_steps` is dropped and every env falls back to `DEFAULT_CENTRAL_AGENT_CONFIG["termination"]["max_turns"] = 20`.

`max_steps` is a first-class `SimpleAgentConfig` field and is the agent server's loop bound:

```python
# responses_api_agents/simple_agent/app.py:164
if self.config.max_steps and step >= self.config.max_steps:
    break
```
[`simple_agent/app.py:48`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_agents/simple_agent/app.py#L48) (the field) - [`app.py:164`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_agents/simple_agent/app.py#L164) (the break)

32 shipped Gym configs set it, so flipping `central_agent.enabled: true` silently changes trajectory length in both directions: [`circle_click.yaml:13`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/resources_servers/circle_click/configs/circle_click.yaml#L13) `max_steps: 1` becomes a 20-turn rollout, `math_with_code: 3` becomes 20, while `tau2: 200` / `browsecomp: 400` get truncated to 20. Different trajectories and different verifier scores than the agent-server baseline, with no warning.

The stop points line up exactly (`step` increments at the loop top before the model call, `_turns` increments once per turn in `is_termination_criteria_met`, and both compare `>=` after tools run), so `max_turns := max_steps` is a 1:1 mapping. Suggested seeding, with an explicit `central_agent.termination.max_turns` still winning:

```suggestion
        agent_central_agent = merge_central_agent_config(
            central_agent_config, inner.get(CENTRAL_AGENT_KEY)
        )
        # simple_agent's own loop bound (`step >= max_steps`) is the same stop
        # point as `turns >= max_turns`, so inherit it: enabling the central
        # agent should not silently change trajectory length. An explicit
        # central_agent.termination.max_turns still wins.
        agent_max_steps = _to_plain(inner.get("max_steps"))
        if agent_max_steps and not (
            (_to_plain(inner.get(CENTRAL_AGENT_KEY)) or {})
            .get("termination", {})
            .get("max_turns")
        ):
            agent_central_agent["termination"]["max_turns"] = int(agent_max_steps)

        registry[top_key] = {
            "name": top_key,
            "implementation": implementation,
            "resources_server": _to_plain(inner["resources_server"])["name"],
            "model_server": _to_plain(inner["model_server"])["name"],
            "central_agent": agent_central_agent,
        }
```

One caveat on that fix: `merge_central_agent_config` shallow-copies ([`config.py:57`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L57), `merged = dict(_to_plain(base) or {})`), so `resolve_central_agent_config(None)["termination"]` **is** `DEFAULT_CENTRAL_AGENT_CONFIG["termination"]`. Writing into `agent_central_agent["termination"]` would corrupt the module-level default for every later agent, so that line is worth switching to a deep copy first.

nemo_rl/environments/central_agent_helpers/config.py
```diff
@@ -0,0 +1,121 @@
...
+# limitations under the License.
+
+"""Config resolution and the NeMo-Gym agent registry for the central agent loop."""
+
+from typing import Any, Optional
+
+# NeMo-Gym agent implementations whose loop the central agent reproduces. The
+# implementation name is the inner key of a `responses_api_agents` block, i.e.
+# the directory under Gym's `responses_api_agents/`.
+SUPPORTED_AGENT_IMPLEMENTATIONS = ("simple_agent", "non_executing_simple_agent")
```
rohitrango, line 22: [`nemo_rl/environments/central_agent_helpers/config.py:22`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L22)

**[BUG]** `non_executing_simple_agent` is listed as supported, but `implementation` is only ever written into the registry ([`config.py:107`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L107)) and never read by `CentralAgent` - so it gets the identical tool-executing, multi-turn `simple_agent` loop.

Upstream it is strictly single-shot: one POST to the model server, then return. No loop, no `function_call` inspection, no tool POSTs, and no `max_steps` field at all:

```python
# responses_api_agents/non_executing_simple_agent/app.py:88
# Preserve session cookies for /run verification, but do not inspect or execute tool calls.
```
[`non_executing_simple_agent/app.py:61-91`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_agents/non_executing_simple_agent/app.py#L61)

This is not hypothetical - [`structured_outputs_v4.yaml:11`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/resources_servers/structured_outputs/configs/structured_outputs_v4.yaml#L11) uses it, and its dataset (`structured_outputs_v4_tool_call.jsonl`) exists precisely so the emitted tool calls are *scored for schema adherence* rather than run. Under `central_agent` those calls get POSTed to the structured_outputs server and the rollout runs up to 20 turns, changing both the trained token sequence and the reward.

The simplest resolution is to not claim it until the loop shape is branched on:

```suggestion
SUPPORTED_AGENT_IMPLEMENTATIONS = ("simple_agent",)
```

Alternatively, keep it and give it an implementation-specific profile (`termination.max_turns: 1` plus a `mode: "none"` tool manager returning `[]`), and drop the matching claim from the design doc.

nemo_rl/environments/nemo_gym.py
```diff
@@ -474,7 +507,74 @@ async def run_rollouts(
         tokenizer: PreTrainedTokenizerBase,
         timer_prefix: str,
     ) -> AsyncGenerator[tuple[int, dict, dict | None], None]:
-        """Stream postprocessed rollouts as NeMo-Gym tasks complete."""
+        """Stream postprocessed rollouts driven by the NeMo-Gym agent servers."""
+        if self._central_agent_enabled:
```
rohitrango, line 597: [`nemo_rl/environments/nemo_gym.py:511`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/nemo_gym.py#L511)

**[BUG]** This guard is the right idea, but it fires very late for the non-SingleController path. `spinup_nemo_gym_actor` forwards `central_agent` unconditionally ([`nemo_gym.py:986`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/nemo_gym.py#L986), [`:1006`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/nemo_gym.py#L1006)) and is shared by both [`grpo.py:652`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/algorithms/grpo.py#L652) and [`single_controller_utils/setup.py:406`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/algorithms/single_controller_utils/setup.py#L406) - but only the latter passes `central_agent=` on to `RolloutManager`.

So a plain `run_grpo_nemo_gym.py` recipe with `env.nemo_gym.central_agent.enabled: true` gets an actor that has already deleted every agent-server block, then [`rollouts.py:2186`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/experience/rollouts.py#L2186) calls `run_rollouts` and hits this `RuntimeError` - after the full Ray cluster, model load, and Gym venv build. `grep -rn central_agent nemo_rl/algorithms/grpo.py` returns nothing, so nothing catches it earlier.

Consider validating at config-parse time (in `setup_nemo_gym_config`, or in `spinup_nemo_gym_actor` given a `use_single_controller` flag) so the misconfiguration fails before spinup rather than at step 1.

nemo_rl/environments/central_agent_helpers/tool_call_manager/base.py
```diff
@@ -0,0 +1,172 @@
...
+        self._finished: dict[int, dict[str, Any]] = {}
+
+    @property
+    def pending_count(self) -> int:
+        return len(self._pending)
+
+    async def submit_and_collect(
+        self, calls: list[dict[str, Any]]
+    ) -> list[dict[str, Any]]:
+        threshold = min(self._min_returns, len(calls))
```
rohitrango, line 134: [`nemo_rl/environments/central_agent_helpers/tool_call_manager/base.py:134`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/tool_call_manager/base.py#L134)

**[BUG]** With `min_returns < n`, a turn's `function_call_output` items land in a *later* turn node than their `function_call`, which renders as an invalid tool transcript. `agent_loop` appends both to the same node ([`agent_loop.py:119`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L119) and [`:127`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L127)), so a node becomes `[assistant(tool_calls=f,g,h,i), output(c), output(d)]` - outputs belonging to calls issued a turn earlier.

The Gym converter is positional and does not re-pair by `call_id`:

```python
# nemo_gym/responses_converter.py:224
def _format_function_call_output(self, m, state):
    state.flush_assistant()
    converted = NeMoGymChatCompletionToolMessageParam(
        content=m["output"], role="tool", tool_call_id=m["call_id"],
    )
```
[`responses_converter.py:224-237`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/nemo_gym/responses_converter.py#L224), with the assistant buffer at [`:323-337`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/nemo_gym/responses_converter.py#L323)

So an assistant message carrying 3 `tool_calls` is followed by 1 `tool` message. Token contiguity still holds, so `_postprocess_nemo_gym_to_nemo_rl_result` will not catch it - the model just trains on structurally invalid transcripts.

A second issue on this line: `threshold` is compared against `len(self._finished)` at [`:138`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/tool_call_manager/base.py#L138), and `_harvest()` pulls in prior-turn completions - so with 3 stale results already queued and `min_returns: 2`, this returns immediately without waiting on *any* of this turn's calls. `min_returns` gives no guarantee that a fresh result was produced.

Related, on the same theme: [`agent_loop.py:125`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L125) suppresses execution on `incomplete_details` unconditionally while the *stop* is configurable, so `stop_on_incomplete_details: false` leaves the same kind of dangling `function_call` every turn.

Given both, it may be cleanest either to hold a turn's outputs until all of that turn's calls resolve, or to document `mode: parallel` as requiring a chat template tolerant of unpaired calls.

nemo_rl/environments/central_agent_helpers/termination_manager/base.py
```diff
@@ -0,0 +1,85 @@
...
+        if self._stop_on_incomplete_details and turn.response.get("incomplete_details"):
+            self._reason = "incomplete_details"
+        elif (
+            self._stop_on_no_tool_calls
+            and not turn.tool_calls
+            and turn.has_output_message
+            and not turn.steering_messages
+        ):
+            self._reason = "no_tool_calls"
+        elif self._max_malformed is not None and malformed >= self._max_malformed:
```
rohitrango, line 73: [`nemo_rl/environments/central_agent_helpers/termination_manager/base.py:73`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/termination_manager/base.py#L73)

**[BUG]** `malformed >= self._max_malformed` means `max_malformed_tool_calls: 0` terminates every rollout after turn 1 - `0 >= 0` is true whenever the two earlier branches do not fire, regardless of what actually happened. Since the config documents `null` as the disable value, `0` reads as "tolerate no malformed calls", and it fails silently: training just collapses to single-turn rollouts with `central_agent_termination_reason="max_malformed_tool_calls"`.

```suggestion
        elif self._max_malformed is not None and malformed > self._max_malformed:
```

Same theme on the next line: [`:75`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/termination_manager/base.py#L75) `elif self._max_turns and self._turns >= self._max_turns:` uses truthiness, so `max_turns: 0` silently removes the cap entirely. Combined with `stop_on_no_tool_calls: false` and `stop_on_incomplete_details: false`, `is_termination_criteria_met` can never return True, and `CentralAgent.run`'s `while True` ([`agent_loop.py:112`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L112)) has no timeout or watchdog to fall back on. Worth using `is not None` and rejecting non-positive values in `build_termination_manager`, keeping `None` as the documented "no cap" sentinel (consistent with `max_malformed_tool_calls`).

nemo_rl/environments/central_agent_helpers/config.py
```diff
@@ -0,0 +1,121 @@
...
+                f"{list(SUPPORTED_AGENT_IMPLEMENTATIONS)}. Disable central_agent to run "
+                f"this agent through its own NeMo-Gym server."
+            )
+
+        registry[top_key] = {
+            "name": top_key,
+            "implementation": implementation,
+            "resources_server": _to_plain(inner["resources_server"])["name"],
+            "model_server": _to_plain(inner["model_server"])["name"],
+            "central_agent": merge_central_agent_config(
```
rohitrango, line 110: [`nemo_rl/environments/central_agent_helpers/config.py:110`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L110)

**[BUG]** `enabled` is part of `DEFAULT_CENTRAL_AGENT_CONFIG`, so it lands in the merged per-agent block here - but nothing inspects it, and [`:114`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L114), `del global_config_dict[top_key]`, runs unconditionally.

The design doc invites this: "Any `responses_api_agents` block may carry its own `central_agent:` key, which is merged over the config above for rollouts of that agent." A user with several agents will reasonably write `central_agent: {enabled: false}` on one to keep it on the agent-server path. The result is worse than a no-op - that agent's uvicorn process is never started, so it can neither run its own loop nor be reached, and the only symptom is a different trajectory shape.

Consider skipping the block (leaving its server in place) when the merged `enabled` is false, or rejecting a per-agent `enabled` key with a clear error.

nemo_rl/environments/central_agent_helpers/config.py
```diff
@@ -0,0 +1,121 @@
...
+
+# NeMo-Gym agent implementations whose loop the central agent reproduces. The
+# implementation name is the inner key of a `responses_api_agents` block, i.e.
+# the directory under Gym's `responses_api_agents/`.
+SUPPORTED_AGENT_IMPLEMENTATIONS = ("simple_agent", "non_executing_simple_agent")
+
+AGENT_SERVER_TYPE_KEY = "responses_api_agents"
+CENTRAL_AGENT_KEY = "central_agent"
+
+DEFAULT_CENTRAL_AGENT_CONFIG: dict[str, Any] = {
```
rohitrango, line 27: [`nemo_rl/environments/central_agent_helpers/config.py:27`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L27)

**[GUIDELINE]** This is a known-field, user-facing config block, so `.claude/skills/config-conventions/SKILL.md` asks for a `pydantic.BaseModel` ("New user-facing config -> pydantic.BaseModel. Do not add new TypedDict-based config classes") rather than `Dict[str, Any]` ([`nemo_gym.py:127`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/nemo_gym.py#L127)) plus a module-level default dict.

The consequence is the skill's named forbidden pattern - each default now lives here, at the call site, *and* in the docs YAML, so the three can drift:

- [`tool_call_manager/__init__.py:35`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/tool_call_manager/__init__.py#L35) `cfg.get("mode", "serial")`
- [`termination_manager/base.py:55-56`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/termination_manager/base.py#L55) `cfg.get("stop_on_no_tool_calls", True)`
- [`tool_call_manager/base.py:123`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/tool_call_manager/base.py#L123) `cfg.get("min_returns") or 1`
- [`steering_manager/base.py:60`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/steering_manager/base.py#L60) `cfg.get("url_path") or "/steering"`
- [`rollout_tree/__init__.py:30`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/rollout_tree/__init__.py#L30) `cfg.get("type", "linear")`

A `CentralAgentConfig(BaseModel, extra="allow")` with sub-models for `rollout_tree` / `tool_calls` / `steering_message` / `termination` would let all five fallbacks and `DEFAULT_CENTRAL_AGENT_CONFIG` go away, and would fix the shallow-copy aliasing at [`:57`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L57) and the dead `cfg` parameter in `CentralAgent.__init__` in the same change.

nemo_rl/experience/rollout_manager.py
```diff
@@ -522,9 +526,11 @@ async def _run_rollouts(
             results: list[dict | None] = [None for _ in inputs]
             received_row_indices: set[int] = set()
             env_timing_metrics: dict[str, Any] = {}
-            async for result_ref in nemo_gym_env.run_rollouts.options(
-                num_returns="streaming"
-            ).remote(inputs, self._tokenizer, timer_prefix):
+            async for result_ref in (
+                getattr(nemo_gym_env, self._ROLLOUT_METHOD)
```
rohitrango: [`nemo_rl/experience/rollout_manager.py:530`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/experience/rollout_manager.py#L530)

**[GUIDELINE]** `.claude/skills/linting-and-formatting/SKILL.md` says "Do not use reflection when functionality can be achieved without it." String dispatch here also defeats pyrefly on the actor-handle call and pushes the test to assert on a magic string ([`test_rollout_manager.py:348`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/tests/unit/experience/test_rollout_manager.py#L348)).

An overridable one-line method gets the same subclass hook without the `getattr`:

```python
def _rollout_stream(self, nemo_gym_env, inputs, timer_prefix):
    """Streaming actor call that drives the rollouts. Subclasses swap the loop only."""
    return nemo_gym_env.run_rollouts.options(num_returns="streaming").remote(
        inputs, self._tokenizer, timer_prefix
    )
```

with `AsyncNemoGymRolloutImplWithAgent` overriding it to call `run_agent_rollouts`, replacing `_ROLLOUT_METHOD` at [`:423`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/experience/rollout_manager.py#L423) and [`:685`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/experience/rollout_manager.py#L685).

nemo_rl/environments/central_agent_helpers/agent_loop.py
```diff
@@ -0,0 +1,220 @@
...
+
+    ``client`` is a NeMo-Gym ``ServerClient``; only ``post(server_name, url_path,
+    json, cookies)`` is used, and its response only needs ``.status``,
+    ``.cookies``, and ``await .content.read()``. Model output is kept as raw JSON
+    end to end so the ``prompt_token_ids`` / ``generation_token_ids`` /
+    ``generation_log_probs`` fields survive into verify untouched.
+    """
+
+    def __init__(
+        self, client: Any, agent_entry: dict[str, Any], cfg: Optional[dict] = None
```
rohitrango, line 78: [`nemo_rl/environments/central_agent_helpers/agent_loop.py:78`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L78)

**[BUG-low]** The `cfg` parameter is dead - the only production caller ([`nemo_gym.py:559`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/nemo_gym.py#L559)) does not pass it, and neither do any of the tests, so everything relies on the `agent_entry.get("central_agent")` fallback on the next line. Two ways to supply one config, one of them unused.

```suggestion
    def __init__(self, client: Any, agent_entry: dict[str, Any]) -> None:
        self._client = client
        self._resources_server = agent_entry["resources_server"]
        self._model_server = agent_entry["model_server"]
        self._cfg = agent_entry.get("central_agent") or {}
```

`pop_agent_registry` always populates `central_agent`, so the fallback stays safe.

nemo_rl/environments/central_agent_helpers/rollout_tree/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+from nemo_rl.environments.central_agent_helpers.rollout_tree.base import BaseRolloutTree
+from nemo_rl.environments.central_agent_helpers.rollout_tree.linear import (
+    LinearRolloutTree,
+)
+
+ROLLOUT_TREES = {"linear": LinearRolloutTree}
+
+
+def build_rollout_tree(
```
rohitrango, line 25: [`nemo_rl/environments/central_agent_helpers/rollout_tree/__init__.py:25`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/rollout_tree/__init__.py#L25)

**[DOCSTRING]** All four public `build_*` factories are in `__all__`, re-exported from the package `__init__`, and documented as the plug-in surface ("a new implementation is one dict entry"), but none has a docstring - [`build_rollout_tree`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/rollout_tree/__init__.py#L25), [`build_steering_manager`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/steering_manager/__init__.py#L30), [`build_termination_manager`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/termination_manager/__init__.py#L26), [`build_tool_call_manager`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/tool_call_manager/__init__.py#L31). Same for [`function_call_output`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/tool_call_manager/base.py#L24).

The base classes themselves are well documented, so this is only the factories. Each raises `ValueError` on an unknown key, which is worth stating:

```suggestion
def build_rollout_tree(
    cfg: Optional[dict[str, Any]],
    prompt: list[dict[str, Any]],
    model_client: Optional[Callable[..., Any]] = None,
) -> BaseRolloutTree:
    """Build the rollout tree named by ``cfg["type"]``.

    Args:
        cfg: The ``central_agent.rollout_tree`` block; None selects the default.
        prompt: Items seeding the root turn.
        model_client: Optional callable a branching tree uses to expand nodes.

    Returns:
        The tree implementation registered under ``cfg["type"]``.

    Raises:
        ValueError: If ``cfg["type"]`` is not a key of ``ROLLOUT_TREES``.
    """
```

nemo_rl/environments/central_agent_helpers/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+
+Each of the four pluggable pieces has a registry and a factory keyed off its
+config block, so a new implementation is one dict entry:
+
+    ROLLOUT_TREES        <- central_agent.rollout_tree.type
+    TOOL_CALL_MANAGERS   <- central_agent.tool_calls.mode
+    STEERING_MANAGERS    <- central_agent.steering_message.enabled / .type
+    TERMINATION_MANAGERS <- central_agent.termination.type
+
+Design: ``ideas/central-controller/index.html`` in the design-docs repo.
```
rohitrango, line 30: [`nemo_rl/environments/central_agent_helpers/__init__.py:30`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/__init__.py#L30)

**[DOCSTRING]** This points at a path that cannot be resolved from this repo, and Sphinx will render it verbatim. The PR adds the equivalent section to the in-repo design doc, so it can point there instead:

```suggestion
Design: ``docs/design-docs/nemo-gym-integration.md`` (Central Agent Loop section).
```

nemo_rl/environments/central_agent_helpers/agent_loop.py
```diff
@@ -0,0 +1,220 @@
...
+            model_response = await self._call_model(
+                {**create_params, "input": tree.get_active_branch()}
+            )
+            last_response = model_response
+            usage = _merge_usage(usage, model_response.get("usage"))
+            output = model_response.get("output") or []
+            tree.append(output)
+
+            tool_calls = [o for o in output if o.get("type") == "function_call"]
+            # A length-truncated response can carry truncated arguments, so its
```
rohitrango, line 122: [`nemo_rl/environments/central_agent_helpers/agent_loop.py:122`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L122)

**[DOC]** Two small inaccuracies in comments that a future reader would reasonably take at face value.

The three-agent claim does not hold up: `grep -rn incomplete_details responses_api_agents/*/app.py` in Gym hits only cvdp, browsecomp, finance and simple - [`hermes_agent/app.py`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_agents/hermes_agent/app.py) has neither `incomplete_details` handling nor any turn loop, `pi_agent` likewise, and there is no `nanobot` agent in Gym at all. Only `simple_agent` does what is described ([`simple_agent/app.py:120`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/responses_api_agents/simple_agent/app.py#L120)).

```suggestion
            # A length-truncated response can carry truncated arguments, so its
            # calls are never executed (simple_agent breaks on the same
            # condition). Termination still sees them.
```

Separately, [`tool_call_manager/base.py:69-70`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/tool_call_manager/base.py#L69) says a nameless call POSTing to `/` "gets a 404 body back ... the same path an unknown tool takes" - Gym's `SimpleServer` registers a GET liveness route at `/` ([`server_utils.py:424`](https://github.com/NVIDIA-NeMo/Gym/blob/473f446f71ec7c1243eb1517fe2440a2b37fe68b/nemo_gym/server_utils.py#L424), `@app.get("/", include_in_schema=False)`), so a POST there returns **405**, which is a different path from an unknown tool's 404.

docs/design-docs/nemo-gym-integration.md
```diff
@@ -49,6 +49,42 @@ Gym result payloads are needed for a short debugging run.
 
 For complete examples, see `examples/nemo_gym/run_grpo_nemo_gym.py`, `examples/nemo_gym/run_distillation_nemo_gym.py`, and their associated configs under `examples/nemo_gym/`.
 
+### Central Agent Loop (SingleController)
+
+By default each rollout is driven by a NeMo Gym agent server. Setting
+`env.nemo_gym.central_agent.enabled: true` moves that loop into the `NemoGym`
+actor: `RolloutManager` selects `AsyncNemoGymRolloutImplWithAgent`, the actor
+runs the turn loop itself, and no agent-server process is started (their config
+blocks become an internal registry). The verify result is unchanged, so
```
rohitrango, line 58: [`docs/design-docs/nemo-gym-integration.md:58`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/docs/design-docs/nemo-gym-integration.md#L58)

**[DOC]** Three corrections to this block, plus one gap.

1. "The verify result is unchanged" - the schema is unchanged, but [`agent_loop.py:166-173`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L166) merges extra `central_agent_*` / `tool_calls_*` keys in (deliberately, since [`rollout_manager.py:655`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/experience/rollout_manager.py#L655) harvests them into metrics). Wording it as "the verify *response schema* is unchanged" would be exact.
2. [`:83`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/docs/design-docs/nemo-gym-integration.md#L83) "Only the `simple_agent` and `non_executing_simple_agent` implementations are reproduced" - the non-executing one is not actually reproduced (see the `SUPPORTED_AGENT_IMPLEMENTATIONS` comment).
3. [`:75`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/docs/design-docs/nemo-gym-integration.md#L75) documents `max_turns: 20` without noting that it overrides the agent block's `max_steps`, and [`:69-70`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/docs/design-docs/nemo-gym-integration.md#L69) documents `mode: parallel` with no mention that a deferred result leaves a `function_call` with no matching output in the next prompt.

The gap: `git ls-files examples/nemo_gym/` shows no config with `central_agent`, and the SC exemplars under `examples/configs/` are untouched, so this doc block is the only place the config appears. The config-conventions skill asks for the default to be reflected in an exemplar YAML, and a runnable example would give users something to copy from - the working tree already has a well-commented `grpo_qwen4b_sc_central_agent.yaml` that is untracked and not part of the PR.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -288,6 +288,21 @@ def setup_single_controller(
...
 
+    # Mirrors grpo.py: the reference model is only loaded when the KL penalty is
+    # non-zero, so asking for its logprobs without it would fail in
+    # use_reference_model() with a missing reference_state_dict. Mutating
+    # master_config here reaches SingleControllerActor, which reads this flag.
+    if (
+        master_config.loss_fn.reference_policy_kl_penalty <= 0
+        and not grpo_config.skip_reference_policy_logprobs_calculation
+    ):
+        grpo_config.skip_reference_policy_logprobs_calculation = True
```
rohitrango, line 411: [`nemo_rl/algorithms/single_controller_utils/setup.py:299`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/algorithms/single_controller_utils/setup.py#L299)

**[TEST]** This auto-skip has no coverage, and it is the highest-value gap in the PR: it silently mutates `master_config`, and that mutation is what `SingleControllerActor._reference_logprobs_required` reads. It is also the exact complement of the guard in [`validate_single_controller_config`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/algorithms/single_controller_utils/config.py#L129), which *is* tested - so the pair is half-covered.

The change itself traces clean end to end (`init_reference_model = reference_policy_kl_penalty > 0` at [`setup.py:210`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/algorithms/single_controller_utils/setup.py#L210); both advantage estimators gate the KL term on `kl_coef > 0`; SC blocks checkpointing, so there is no resume path) - it just needs a regression test. `tests/unit/single_controller/test_single_controller_setup.py` already has the harness (`_make_master_config` + `patched_factories`), and both branches are reachable from the existing fixture since `reference_policy_kl_penalty` defaults to `0.01` and `skip_reference_policy_logprobs_calculation` to `False`:

```python
def test_zero_kl_penalty_auto_skips_reference_logprobs(self, patched_factories):
    mc = _make_master_config()
    mc.loss_fn.reference_policy_kl_penalty = 0.0
    setup_single_controller(mc, MagicMock(pad_token_id=0))
    # mutated in place: SingleControllerActor reads the flag off master_config
    assert mc.grpo.skip_reference_policy_logprobs_calculation is True


def test_nonzero_kl_penalty_leaves_reference_logprobs_enabled(self, patched_factories):
    mc = _make_master_config()
    setup_single_controller(mc, MagicMock(pad_token_id=0))
    assert mc.grpo.skip_reference_policy_logprobs_calculation is False
```

(Drafted but not executed - the review environment had no container or GPU.)

tests/unit/environments/test_central_agent.py
```diff
@@ -0,0 +1,782 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+import json
+
+import pytest
+from omegaconf import OmegaConf
+
+from nemo_rl.environments.central_agent_helpers import (
```
rohitrango, line 21: [`tests/unit/environments/test_central_agent.py:21`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/tests/unit/environments/test_central_agent.py#L21)

**[TEST]** The suite is well above the repo's bar - real code paths, specific assertions, no tautologies, and correct placement and marks (no `nemo_gym` mark is right, since nothing here imports `nemo_gym`). The gaps cluster in two places.

**Parallel mode is never driven through the loop.** `ParallelToolCallManager` is tested directly ([`:294-376`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/tests/unit/environments/test_central_agent.py#L294), including the `threshold = min(min_returns, n)` edge at `n == 0`), but every `CentralAgent.run` test uses the default serial mode - so the terminate -> `drain()` -> `tree.append` path at [`agent_loop.py:145-146`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L145) is untested. That is exactly the ordering the parallel-mode pairing concern turns on.

**Error paths.** Only `test_loop_raises_on_a_failed_model_call` exists; three of the four `_raise_for_status` call sites are uncovered ([`agent_loop.py:89`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L89) seed_session, [`:168`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L168) verify, plus the steering POST at [`:211`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L211)). A resources server 5xx-ing on seed or verify, or a 404 from a steering endpoint that no shipped server implements, are the most likely real failures, and the error message is the operator's only signal. Also uncovered: the `JSONDecodeError` branch at [`:194`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L194), the `TypeError` half of the malformed-arguments guard at [`tool_call_manager/base.py:80`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/tool_call_manager/base.py#L80) (the comment says "None is a TypeError, handled below", but only `JSONDecodeError` is exercised), and the string-prompt normalization at [`:93`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/agent_loop.py#L93) - `responses_create_params.input` is a bare string in several Gym datasets, and every test passes a list.

Two smaller ones: `pop_agent_registry` is only ever tested with a single agent instance, despite its contract being "remove *every* agent block and index it"; and `resolve_central_agent_config` is only fed plain dicts, never the `DictConfig` it gets in production via `_to_plain`.

Happy to share concrete test bodies for any of these - they were drafted but not executed, since the review environment had no container or GPU.

nemo_rl/environments/nemo_gym.py
```diff
@@ -474,7 +507,74 @@ async def run_rollouts(
...
+        ):
+            yield item
+
+    def _agent_server_tasks(self, nemo_gym_examples: list[dict]) -> Any:
+        """One /run POST per row against its NeMo-Gym agent server."""
+        return self.rch.run_examples(
+            examples=nemo_gym_examples, head_server_config=self.head_server_config
+        )
+
+    def _central_agent_tasks(self, nemo_gym_examples: list[dict]) -> Any:
```
rohitrango, line 639: [`nemo_rl/environments/nemo_gym.py:544`](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/nemo_gym.py#L544)

**[PERF-EVIDENCE]** The two-curve reward comparison in the description is the right first check and settles the accuracy question - nothing more is needed on that axis.

What is still open is throughput, which is what this function actually changes: N per-rollout agent loops move from N uvicorn processes into this single `NemoGym` actor's event loop, where all HTTP fan-out, per-turn `json.loads` of token-id-bearing payloads, and the existing `_postprocess_nemo_gym_to_nemo_rl_result` now share one CPU. The perf claim in [`pop_agent_registry`'s docstring](https://github.com/NVIDIA-NeMo/RL/blob/714a4bbdec01e4d0d71da1c5c50a7776d840f2d2/nemo_rl/environments/central_agent_helpers/config.py#L76) ("one uvicorn process and one uv venv build per agent instance are saved") is also unmeasured.

It would be great to see, at the same `num_prompts_per_step x num_generations_per_prompt` on a multi-turn tool-heavy env (`workplace_assistant` is a good stressor - a single-turn env would not exercise the loop), with and without `central_agent`:

1. `timing/rollout/run_rollouts` and `timing/rollout/postprocess_results_pct` per step (both already emitted by `_stream_rollouts`)
2. end-to-end step time
3. `NemoGym` actor CPU utilisation at peak concurrency, to show the single event loop is not the bottleneck
4. actor spinup wall time, to put a number on the saved venv builds

For what it is worth, the concurrency shape itself matches the existing path - `_agent_server_tasks` calls `run_examples` with no semaphore either, so there is no regression there, just an unmeasured axis.

rohitrango (commented):
Reviewed by a five-agent team (RL/guidelines expert, bug finder, test reviewer, NeMo Gym API expert, devil's advocate). 29 candidate findings were raised and then independently stress-tested; 15 survived verification and are inline below. All analysis is static - the review environment had no container or GPU, so nothing was executed.

**Please rebase on `main` first.** GitHub reports this as `CONFLICTING`. `main` is 40 commits ahead of the merge base (`e496258`), and five files changed on both sides: `nemo_rl/algorithms/single_controller_utils/setup.py`, `nemo_rl/environments/nemo_gym.py`, `nemo_rl/experience/rollout_manager.py`, `pyrefly.toml`, `tests/unit/experience/test_rollout_manager.py`.

This is a valuable change and the design holds up well. Putting the turn loop where NeMo-RL can reach it is the prerequisite for the tree / compaction / steering / centralized-termination surface that the agent-server path structurally cannot provide, the plug-in registry keeps that surface cheap to extend, and the parity work with `simple_agent` is careful - usage merging, cookie carry-forward, and malformed-arguments-as-tool-result all match upstream line for line. The 782-line test file is genuinely good: real code paths, specific assertions, no tautologies. Cleanliness notes: copyright headers are correct on all 12 new files, all 12 are registered in `pyrefly.toml`, and the PR title is valid Conventional Commits.

Three findings look worth resolving before merge - the dropped `max_steps` (which would quietly change trajectory length on 32 shipped Gym envs), the `non_executing_simple_agent` loop shape, and the late failure for non-SingleController configs. The rest are suggestions.

_Generated by Claude Code_

PR #3685, opened 2026-08-17
fix(vllm): resolve unset sampling params to neutral defaults

### LINKED ISSUE #3497: vLLM resolves unset sampling params from the model's generation_config.json, not neutral defaults
## Summary

vLLM resolves any **unset** sampling parameter on an incoming OpenAI request from the *model's* `generation_config.json`, not from neutral defaults. NeMo-RL never overrides this, so an omitted `top_p` / `temperature` / `top_k` / `min_p` / `repetition_penalty` silently picks up a model-specific value that differs from `policy.generation.*`.

For RL this is worse than ordinary drift: top-p/top-k truncate and renormalize the distribution, which changes the logprobs. NeMo-RL mirrors that into the training forward pass — [`dtensor_policy_worker` builds `TrainingSamplingParams` from `generation_cfg["top_p"/"top_k"]`](https://github.com/NVIDIA-NeMo/RL/blob/414b9bb6b69539011e835ce369be8e53aad709bd/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L244) and [`_apply_top_k_top_p_filtering` rescales the logits](https://github.com/NVIDIA-NeMo/RL/blob/414b9bb6b69539011e835ce369be8e53aad709bd/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L571) only when [`_need_top_p_filtering` sees `top_p != 1.0`](https://github.com/NVIDIA-NeMo/RL/blob/414b9bb6b69539011e835ce369be8e53aad709bd/nemo_rl/algorithms/logits_sampling_utils.py#L52). If the engine actually samples at a different top_p than the config records, the train-side rescaling replicates the wrong distribution and the mismatch is invisible.

## Mechanism

Against the pinned `vllm==0.25.1`:

1. [`top_p: float | None = None`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/openai/chat_completion/protocol.py#L219) — an omitted field stays `None`.
2. [`if (top_p := self.top_p) is None: top_p = default_sampling_params.get("top_p", ...)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/openai/chat_completion/protocol.py#L603-L605) — the neutral `1.0` is only the *last* fallback. Same pattern for `temperature`, `top_k`, `min_p`, `repetition_penalty`.
3. [`default_sampling_params = self.model_config.get_diff_sampling_param()`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/openai/chat_completion/serving.py#L168).
4. [`config = {} if src == "vllm" else self.try_get_generation_config()`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/config/model.py#L1499-L1501) — docstring: *"vLLM's neutral defaults if `generation_config="vllm"`, the model's defaults if `generation_config="auto"`"*.
5. [`generation_config: str = "auto"`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/config/model.py#L290) is the default, and NeMo-RL passes neither `generation_config` nor `override_generation_config` into [`AsyncEngineArgs`](https://github.com/NVIDIA-NeMo/RL/blob/414b9bb6b69539011e835ce369be8e53aad709bd/nemo_rl/models/generation/vllm/vllm_worker_async.py#L191) (grep of `nemo_rl/` → 0 hits).

Concretely, `Qwen/Qwen3-30B-A3B-Instruct-2507` (used by `examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml`) ships `top_p: 0.8`, `top_k: 20`, `temperature: 0.7`, while the gym recipes set `top_p: 1.0` / `top_k: null` / `temperature: 1.0`.

## Not always the case — introduced by vLLM v0.8.0

| vLLM | `ModelConfig.generation_config` | effect |
|---|---|---|
| ≤ v0.7.3 | [`Optional[str] = None`](https://github.com/vllm-project/vllm/blob/v0.7.3/vllm/config.py#L244) | [`get_diff_sampling_param` returns `{}`](https://github.com/vllm-project/vllm/blob/v0.7.3/vllm/config.py#L948-L950) — the model's file is ignored |
| ≥ v0.8.0 | `str = "auto"` | the model's `generation_config.json` is read |

## Current mitigation is partial

The async OpenAI server guard is what stands between this and a silent off-policy rollout, and it covers the parameters unevenly:

- **`top_k`** — safe: [`request.top_k = -1`](https://github.com/NVIDIA-NeMo/RL/blob/414b9bb6b69539011e835ce369be8e53aad709bd/nemo_rl/models/generation/vllm/vllm_worker_async.py#L678) explicitly pins it, so vLLM never consults the model default.
- **`temperature`** — safe: [compared strictly](https://github.com/NVIDIA-NeMo/RL/blob/414b9bb6b69539011e835ce369be8e53aad709bd/nemo_rl/models/generation/vllm/vllm_worker_async.py#L682), so `None` trips the assert.
- **`top_p`** — [compared strictly on `main`](https://github.com/NVIDIA-NeMo/RL/blob/414b9bb6b69539011e835ce369be8e53aad709bd/nemo_rl/models/generation/vllm/vllm_worker_async.py#L683), but #3401 coalesces an unset value to a literal `1.0` before comparing, which makes the omitted case pass. Being handled in that PR's review.
- **`min_p` / `repetition_penalty`** — not covered by the guard at all.

Note the guard only protects the **async OpenAI-server** path; it is not a general property of the engine.

## Proposed fix

Pass `generation_config="vllm"` when constructing [`AsyncEngineArgs`](https://github.com/NVIDIA-NeMo/RL/blob/414b9bb6b69539011e835ce369be8e53aad709bd/nemo_rl/models/generation/vllm/vllm_worker_async.py#L191) (and the sync engine equivalent), so unset parameters resolve to vLLM's neutral defaults instead of per-model values. That removes the whole class rather than patching one parameter at the guard, and makes `policy.generation.*` the single source of truth — which is what the train-side logprob rescaling already assumes.

Worth deciding explicitly whether any model-default behavior is wanted for **validation** rollouts before changing it globally.

_(Surfaced during the review of #3401.)_

lonexreb:
# What does this PR do ?

Passes `generation_config="vllm"` in the shared engine kwargs built by `BaseVllmGenerationWorker._load_model` (consumed by both the sync `vllm.LLM` engine and the async `AsyncEngineArgs` path), so sampling parameters NeMo-RL leaves unset on a request (`min_p`, `repetition_penalty`, and formerly `top_p`) resolve to vLLM's **neutral** defaults instead of the model's `generation_config.json`.

Without this, a model shipping non-neutral values (e.g. `Qwen2.5-0.5B-Instruct` with `repetition_penalty: 1.1`) silently samples off-policy while the train-side logprob rescaling replicates the distribution recorded in `policy.generation.*` — the mismatch is invisible to the existing async-server guard, which only covers `top_k`/`temperature`/`top_p`. This is the engine-level fix proposed in the issue: it removes the whole class of drift rather than patching one parameter at the guard, and makes `policy.generation.*` the single source of truth.

Details:
- The new kwarg sits **before** `**vllm_kwargs`, so `generation.vllm_kwargs.generation_config` can deliberately restore model-default behavior (e.g. if that is ever wanted for validation rollouts).
- The async server's explicit `top_p is not None` assert is kept as defense-in-depth; its comment is updated since its old rationale (resolution from `generation_config.json`) no longer applies at the engine level.

# Issues

Fixes #3497

# Usage

No user-facing config change. To intentionally restore per-model sampling defaults:

```yaml
policy:
  generation:
    vllm_kwargs:
      generation_config: "auto"
```

- [x] One-line engine-arg change + comment updates; verified `generation_config` is an `EngineArgs`/`ModelConfig` field on the pinned vllm==0.25.1
- [ ] Functional coverage: existing L1 tests exercise the vLLM engine; `token_mult_prob_error`-style checks guard sampling/train distribution agreement

### DIFF
--- a/nemo_rl/models/generation/vllm/vllm_worker.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker.py
@@ -589,6 +589,13 @@ class BaseVllmGenerationWorker:
             enable_sleep_mode=True,
             # Set disable_log_stats=False so that self.llm.get_metrics() works.
             disable_log_stats=False,
+            # Resolve sampling params NeMo-RL leaves unset (min_p,
+            # repetition_penalty, ...) to vLLM's neutral defaults instead of the
+            # model's generation_config.json, so policy.generation.* stays the
+            # single source of truth for the sampling distribution that the
+            # train-side logprob rescaling assumes. Overridable via
+            # generation.vllm_kwargs.generation_config. See #3497.
+            generation_config="vllm",
             **vllm_kwargs,
         )
 
--- a/nemo_rl/models/generation/vllm/vllm_worker_async.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker_async.py
@@ -747,12 +747,16 @@ class VllmAsyncGenerationWorkerImpl(
             # caller that constructs a non-train GenerationSamplingParams. Multi-turn
             # agents issue their own requests, so this server-side check is the one
             # chokepoint they all pass.
-            # vLLM resolves an unset top_p from the model's generation_config.json
-            # (ModelConfig.generation_config defaults to "auto"), NOT to 1.0, so a
-            # request omitting it would sample off-policy while passing this check.
+            # The engine is constructed with generation_config="vllm" (see
+            # BaseVllmGenerationWorker._load_model), so an unset top_p resolves to
+            # vLLM's neutral 1.0 rather than the model's generation_config.json.
+            # Still require it explicitly: a silently-resolved 1.0 would only match
+            # the on-policy check by coincidence, and an explicit value keeps this
+            # chokepoint independent of the engine-level setting (which
+            # generation.vllm_kwargs.generation_config can override).
             assert request.top_p is not None, (
-                "top_p must be set explicitly on NeMo-RL requests; an unset top_p is "
-                "resolved by vLLM from the model's generation_config.json and would "
+                "top_p must be set explicitly on NeMo-RL requests; an unset top_p "
+                "would be resolved by vLLM instead of the NeMo-RL config and would "
                 "bypass the on-policy sampling check."
             )
             request_top_p = request.top_p


lonexreb:
@terrykong @youngeunkwon0405 — gentle ping when you have bandwidth. This one fixes the off-policy sampling drift from #3497 (model `generation_config.json` silently overriding unset sampling params) at the engine level. All pre-CI checks are green and the branch is mergeable against current main; it still needs a `CI:*` label to run the full tier. Happy to address any feedback or rebase again if it goes stale.

PR #3690, opened 2026-08-18
fix: raise an actionable error when a prepare_for_* step is skipped

### LINKED ISSUE #1141: More helpful error message if prepare_for_*() not called
Internal report:

If someone is making large changes or writing their own training loop. Users may encounter "illegal memory access" if the model is offloaded and not moved to the GPU before that API call. We should either document or give more helpful error messages when the appropriate prepare step is not called.

VishnuR23:
# What does this PR do ?

Turns the opaque CUDA `illegal memory access` you get from skipping a `prepare_for_*()` step into a `RuntimeError` that names both the API you called and the prepare step you missed.

# Issues

Closes #1141

# Usage

Policy workers offload model parameters to CPU between phases (`finish_inference`, `offload_after_refit`), and the matching `prepare_for_*` call onloads them again. Reaching a compute API without that call leaves parameters on CPU while the rest of the step assumes CUDA. As #1141 notes, that's easy to hit when writing a custom training loop, and the resulting failure points nowhere near the missing step.

Before — an `illegal memory access` from somewhere deep in the forward pass. After:

```python
policy.offload_after_refit()
policy.get_logprobs(data)
```
```
RuntimeError: DTensorPolicyWorkerV2.get_logprobs() was called while the model
parameters are still on CPU. Call prepare_for_lp_inference() to onload the model
before get_logprobs(). Parameters are offloaded to CPU between phases (for example
by finish_inference() or offload_after_refit()), and without the onload this
typically fails later as a CUDA 'illegal memory access'.
```

## What changed

`AbstractPolicyWorker._assert_model_onloaded()` checks the device of the model's first parameter and raises naming the API and the required prepare call. It's wired into the compute entry points of all three policy workers:

| Worker | Guarded API | Required call |
|---|---|---|
| DTensor v1 / v2 | `train` | `prepare_for_training` |
| DTensor v1 / v2 | `get_logprobs`, `score`, `get_topk_logits` | `prepare_for_lp_inference` |
| Megatron | `train` | `prepare_for_training` |
| Megatron | `get_logprobs`, `get_topk_logits` | `prepare_for_lp_inference` |

`get_reference_policy_logprobs` delegates to `get_logprobs`, so it's covered without a second check.

The check deliberately does not fire in three cases:
- **FSDP `cpu_offload`** — parameters legitimately stay on CPU during compute, so callers pass `params_may_be_offloaded=self.cpu_offload`. Megatron only has `optimizer_cpu_offload` (params are always onloaded for compute), so it passes nothing.
- **A model with no parameters** — nothing to check.
- **Any non-CPU device** — only a CPU-resident model is treated as offloaded.

Megatron holds a list of virtual pipeline chunks rather than a single module, so `first_parameter_device()` accepts both shapes.

I also added docstrings to the `PolicyInterface` lifecycle methods (`prepare_for_training`, `finish_training`, `prepare_for_lp_inference`), which had none — that covers the "or document" half of the issue for someone reading the interface before writing their own loop.

**Pre checks**:
- [x] Did you write any new necessary tests? — `tests/unit/models/policy/test_prepare_step_guards.py`, 8 tests covering the raise path, the message contents, the module-list shape, and each of the three skip conditions. The real workers are Ray actors needing a GPU, so the guard is exercised through a minimal `AbstractPolicyWorker` subclass.
- [x] Did you add or update any necessary documentation? — interface docstrings; no user-facing config or feature change

# Additional Information

* Local verification (macOS, CPU-only, so GPU paths could not be exercised):
  * `pytest tests/unit/models/policy/test_prepare_step_guards.py` → 8 passed
  * `pytest tests/unit/models/policy` → the failures there (`test_split_api_wrappers.py`, `test_utils.py`, `test_policy_validation.py`) reproduce identically on a clean `main` with `Torch not compiled with CUDA enabled`; none of those files import the modules touched here.
* Scope note: value workers (`nemo_rl/models/value/workers/`) have the same `prepare_for_training` / `prepare_for_inference` lifecycle and could take the same guard. I left them out to keep this focused on the policy path from the issue — happy to extend if you'd prefer it in one change.

### DIFF
--- a/nemo_rl/models/policy/interfaces.py
+++ b/nemo_rl/models/policy/interfaces.py
@@ -148,10 +148,17 @@ class PolicyInterface(ABC):
 
     @abstractmethod
     def prepare_for_training(self, *args: Any, **kwargs: Any) -> None:
+        """Onload the model and optimizer state to GPU so training can run.
+
+        Call this before :meth:`train`. Implementations offload parameters to
+        CPU between phases, so skipping it leaves them on CPU while the training
+        step assumes CUDA.
+        """
         pass
 
     @abstractmethod
     def finish_training(self, *args: Any, **kwargs: Any) -> None:
+        """Release training-only state after a :meth:`train` phase completes."""
         pass
 
     @abstractmethod
@@ -252,7 +259,12 @@ class ColocatablePolicyInterface(PolicyInterface):
 
     @abstractmethod
     def prepare_for_lp_inference(self, keep_train_buffers: bool = False) -> None:
-        """Put the policy in eval mode for logprob inference.
+        """Onload the model and put the policy in eval mode for logprob inference.
+
+        Call this before :meth:`get_logprobs`, :meth:`get_reference_policy_logprobs`,
+        :meth:`get_topk_logits`, or :meth:`score`. As with
+        :meth:`prepare_for_training`, parameters are offloaded to CPU between
+        phases and must be onloaded again first.
 
         Args:
             keep_train_buffers: Leave grad buffers and optimizer state on CUDA
--- a/nemo_rl/models/policy/workers/base_policy_worker.py
+++ b/nemo_rl/models/policy/workers/base_policy_worker.py
@@ -22,8 +22,62 @@ from nemo_rl.models.policy.interfaces import ReferenceLogprobOutputSpec
 from nemo_rl.utils.nsys import wrap_with_nvtx_name
 
 
+def first_parameter_device(
+    model: Any,
+) -> Optional[torch.device]:
+    """Return the device of the model's first parameter, or None if it has none.
+
+    ``model`` is a single ``nn.Module`` on the DTensor workers and a list of
+    virtual pipeline chunks on the Megatron worker, so both are accepted.
+    """
+    modules = model if isinstance(model, (list, tuple)) else [model]
+    for module in modules:
+        for param in module.parameters():
+            return param.device
+    return None
+
+
 class AbstractPolicyWorker:
     """Base class for policy workers with shared functionality."""
+
+    def _assert_model_onloaded(
+        self,
+        api_name: str,
+        required_prepare_call: str,
+        *,
+        params_may_be_offloaded: bool = False,
+    ) -> None:
+        """Fail with an actionable message if the model is still offloaded to CPU.
+
+        Workers offload model parameters to CPU between phases (``finish_inference``,
+        ``offload_after_refit``), and the matching ``prepare_for_*`` call onloads them
+        again. Reaching a compute API without that call leaves parameters on CPU while
+        the rest of the step assumes CUDA, which surfaces as an opaque ``illegal memory
+        access`` rather than anything pointing at the missing step. Check up front so
+        the failure names the call to add.
+
+        Args:
+            api_name: Name of the compute API being entered, used in the message.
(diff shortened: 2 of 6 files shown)


tildahh:
Nice improvement! The parameter-device check gives a much clearer error for offloaded models. One edge case I noticed from reading the lifecycle methods: `prepare_for_lp_inference()` brings parameters onto CUDA, so `_assert_model_onloaded()` passes, but that doesn't necessarily mean the worker is ready for training.

In Megatron, that call releases grad-buffer storage unless `keep_train_buffers=True`, and can also offload optimizer state when `offload_optimizer_for_logprob` is enabled. The DTensor workers can likewise offload optimizer state. A subsequent `train(eval_mode=False)` without `prepare_for_training()` would therefore pass this guard without the training state being restored.

Would it make sense to cover that transition with a training-readiness check and a regression test, while preserving supported CPU-offload configurations and `keep_train_buffers=True` behavior?

VishnuR23:
Thanks — you're right, and I confirmed it against the current code rather than taking it on faith. `prepare_for_lp_inference()` onloads parameters, so `_assert_model_onloaded()` sees a CUDA-resident model and passes, while with `keep_train_buffers=False` it has separately:

- released grad-buffer storage (`megatron_policy_worker.py`, the `not keep_train_buffers and not uses_mxfp8_shared_buffer` branch), and
- moved optimizer state to CPU when `offload_optimizer_for_logprob` is set — Megatron and both DTensor workers.

So `train(eval_mode=False)` from there clears the existing guard with the training state still parked. Pushed a fix.

**What it does.** A `_training_state_parked` flag on `AbstractPolicyWorker`, set at the points that actually park state — the grad-buffer release and the optimizer offload inside each `prepare_for_lp_inference` — and cleared in `prepare_for_training`, which is what restores them. `train()` checks it through `_assert_training_state_restored()` only when `eval_mode=False`, since a forward-only eval pass needs neither the buffers nor the optimizer.

**On preserving the two cases you named.** Setting the flag from the offloading branches rather than from `keep_train_buffers` alone is what keeps both quiet:

- `keep_train_buffers=True` never reaches those branches, so an already-open train step is untouched. The single-controller chunked path (`keep_train_buffers=step_open`) also calls `prepare_for_training` before each chunk, so it clears the flag regardless.
- A configuration that offloads nothing — DTensor with `offload_optimizer_for_logprob=False`, or Megatron on the MXFP8 shared-buffer path — never sets the flag, so CPU-offload setups and no-offload setups both stay silent. Had I keyed it off `not keep_train_buffers`, those would have started raising spuriously.

I also checked the call ordering in `sft.py` and `single_controller.py`: both already call `prepare_for_training()` before training, including after the `train(eval_mode=True)` validation loop, so no existing flow trips this.

**Tests.** The parked-state error and its message, the restored case, the default, plus a structural check that every worker setting the flag also clears it in `prepare_for_training` — that asymmetry is the dangerous one, since a stuck flag would wedge every later train step. I verified that check fails against a worker with the clear removed.

The branch is also rebased onto current `main`.

**One gap I left deliberately**, happy to close it if you'd prefer: `offload_before_refit()` parks optimizer state while leaving parameters on GPU, so in principle it has the same shape. In the refit flow it is followed by `offload_after_refit()`, which moves parameters to CPU and is therefore caught by the device check, so I did not want to widen the blast radius on a path I can't exercise on a GPU here. Say the word and I'll set the flag there too.

PR #3692, opened 2026-08-18
feat: add checkpointing.resume_if_exists to control automatic resumption

### LINKED ISSUE #345: [Feature Request] Add configurable `resume_if_exists` flag to control automatic checkpoint resumption
Hi team,

**Summary**
At the moment, the training scripts automatically resume from the latest checkpoint whenever `last_checkpoint_path` is found. This is convenient, but there are cases where I want to reuse the same `checkpoint_dir` while *not* resuming — i.e., I’d like to start training from scratch even if previous checkpoints are present.

**Why it matters**

* Experimentation often involves rerunning with the same output directory for ease of bookkeeping.
* Deleting or moving checkpoints manually is error‑prone, especially in shared storage or automated pipelines.
* Providing an opt‑in / opt‑out flag keeps the current convenience while adding flexibility.

**Where it shows up**
I noticed this in **GRPO**, but the same logic appears in **DPO** and **SFT**.
Example: in `nemo_rl/algorithms/grpo.py` around line 148 you immediately call

```python
last_checkpoint_path = checkpointer.get_latest_checkpoint_path()
```

without a way to override the behavior.

**Proposed interface**
Add a field under `checkpointing` (or similar) in the config, e.g.

```yaml
checkpointing:
  resume_if_exists: true   # default keeps current behaviour
```

Then gate the logic like:

```python
if master_config["checkpointing"].get("resume_if_exists", True):
    last_checkpoint_path = checkpointer.get_latest_checkpoint_path()
else:
    last_checkpoint_path = None
```

**Benefits**

* Zero disruption for existing users (default = current behaviour).
* Simple implementation; touches only the call site.
* Symmetric with NeMo’s `resume_if_exists` option in other training loops, so it feels familiar to users.

I’m not across the entire code‑base, so the exact wiring may differ, but I hope this illustrates the idea. Happy to refine the proposal or open a PR if you think it makes sense.

Thanks for your consideration!

VishnuR23:
# What does this PR do ?

Adds a `checkpointing.resume_if_exists` flag so a `checkpoint_dir` that already holds checkpoints can be reused for a fresh run without deleting the previous one by hand.

# Issues

Closes #345

# Usage

Today `get_latest_checkpoint_path()` is called unconditionally at setup, so training always resumes from the newest checkpoint in `checkpoint_dir`. This implements the behaviour agreed on the issue:

```yaml
checkpointing:
  enabled: true
  # set to false to start fresh in a checkpoint_dir that already has
  # checkpoints; the existing ones are moved to checkpoint_dir/run_<n>/
  resume_if_exists: true
```

- `resume_if_exists: true` — current behaviour, resume from the latest checkpoint (also fine from a cold start).
- `resume_if_exists: false` — move the existing `step_*` directories under `checkpoint_dir/run_<n>/` and proceed as a cold start.

```console
$ ls results/sft
step_10  step_20

$ uv run examples/run_sft.py --config=examples/configs/sft.yaml checkpointing.resume_if_exists=false
UserWarning: checkpointing.resume_if_exists is false, so 2 existing checkpoint(s) in
results/sft were moved to results/sft/run_0 and this run starts from scratch.
Set resume_if_exists to true to resume from them instead.

$ ls results/sft
run_0

$ ls results/sft/run_0
step_10  step_20
```

Successive cold starts number themselves `run_0`, `run_1`, and so on.

## Design notes

- **Nothing is deleted.** Checkpoints are moved, not removed, so a mistaken flag is recoverable. `run_<n>` does not match the `step_<digits>` pattern that both `get_latest_checkpoint_path()` and `_load_checkpoint_history()` glob for, so an archived run stays on disk while being invisible to resumption and to best-checkpoint selection.
- **Archiving happens once in `CheckpointManager.__init__`**, not at the nine `get_latest_checkpoint_path()` call sites. All nine algorithms (grpo, ppo, sft, dpo, rm, distillation, xtoken, single-controller) pick it up with no changes, and the getter keeps its expected lack of side effects. `CheckpointManager` is only constructed driver-side, once per run, so this runs exactly once — and it is a no-op when there are no `step_*` directories, so a second construction could not double-archive anyway.
- **It runs whatever `enabled` is set to.** Resumption itself does not consult `enabled` — `get_latest_checkpoint_path()` is called before any `enabled` check — so gating the archive on it would silently resume the run the user just asked to start fresh. Happy to flip this if you would rather a disabled-checkpointing run never touch the directory.
- **Absence means resume.** A config written before this option has no key, so only an explicit `false` archives. That keeps every existing config and in-flight run behaving exactly as before.

Per the config conventions the default is written in the exemplar YAMLs rather than in code — 15 base exemplars carry `resume_if_exists: true`; the 7 that inherit their `checkpointing` block via `defaults:` deliberately do not repeat it. The 6 reference configs under `tests/unit/reference_configs/` are updated so `test_reference_configs_up_to_date` stays green.

**Pre checks**:
- [x] Did you write any new necessary tests? — 6 tests in `tests/unit/utils/test_checkpoint.py` covering: key absent (behaves as before), `true`, `false` archives and cold-starts while preserving `training_info.json`, `false` on an empty dir is a no-op, successive runs number `run_0`/`run_1`, and archived runs stay invisible to both `get_latest_checkpoint_path` and `get_best_checkpoint_path`.
- [x] Did you add or update any necessary documentation? — the flag is documented on the `CheckpointingConfig` docstring (which autodoc2 renders) and inline in every exemplar YAML. There is no existing guide section on resume behaviour — `docs/design-docs/checkpointing.md` is specifically about exporting to HF format — so I did not force it in there. Glad to add a proper section if you would like one.

# Additional Information

* Local verification (macOS, CPU-only):
  * `pytest tests/unit/utils/test_checkpoint.py tests/unit/test_config_validation.py tests/unit/test_config_v2.py tests/unit/tools/test_config_cli.py` → 656 passed, 6 skipped. The 10 failures are `test_all_config_no_tp_size_accuracy_issues` on unrelated `grpo-*` / `mopd-*` recipes; that failure set is byte-identical to a clean `main` baseline.
  * `./tools/config_cli.py minimize-check examples/configs/recipes/{llm,vlm}/*.yaml` → clean

### DIFF
--- a/examples/configs/audio_grpo_3B_megatron.yaml
+++ b/examples/configs/audio_grpo_3B_megatron.yaml
@@ -11,6 +11,9 @@ grpo:
 
 checkpointing:
   enabled: true
+  # set to false to start fresh in a checkpoint_dir that already has
+  # checkpoints; the existing ones are moved to checkpoint_dir/run_<n>/
+  resume_if_exists: true
   checkpoint_dir: results/audio_grpo_3B_megatron
   keep_top_k: 10
   save_period: 400
--- a/examples/configs/distillation_math.yaml
+++ b/examples/configs/distillation_math.yaml
@@ -20,6 +20,9 @@ loss_fn:
 
 checkpointing:
     enabled: true
+    # set to false to start fresh in a checkpoint_dir that already has
+    # checkpoints; the existing ones are moved to checkpoint_dir/run_<n>/
+    resume_if_exists: true
     checkpoint_dir: "checkpoints/distillation-${policy.model_name}"
     metric_name: "val:accuracy" # one of "val:" or "train:" followed by the metric name
     higher_is_better: true
--- a/examples/configs/dpo.yaml
+++ b/examples/configs/dpo.yaml
@@ -22,6 +22,9 @@ dpo:
 
 checkpointing:
   enabled: true
+  # set to false to start fresh in a checkpoint_dir that already has
+  # checkpoints; the existing ones are moved to checkpoint_dir/run_<n>/
+  resume_if_exists: true
   checkpoint_dir: "results/dpo"
   metric_name: "val:validation-default_loss"
   higher_is_better: false
--- a/examples/configs/grpo_math_1B.yaml
+++ b/examples/configs/grpo_math_1B.yaml
@@ -112,6 +112,9 @@ loss_fn:
 
 checkpointing:
   enabled: true
+  # set to false to start fresh in a checkpoint_dir that already has
+  # checkpoints; the existing ones are moved to checkpoint_dir/run_<n>/
+  resume_if_exists: true
   checkpoint_dir: "results/grpo"
   metric_name: "val:accuracy" # one of "val:" or "train:" followed by the metric name
   higher_is_better: true
--- a/examples/configs/grpo_math_1B_megatron.yaml
+++ b/examples/configs/grpo_math_1B_megatron.yaml
@@ -30,6 +30,9 @@ loss_fn:
 
 checkpointing:
   enabled: false
+  # set to false to start fresh in a checkpoint_dir that already has
+  # checkpoints; the existing ones are moved to checkpoint_dir/run_<n>/
+  resume_if_exists: true
   checkpoint_dir: "results/grpo_megatron"
   metric_name: "val:accuracy" # one of "val:" or "train:" followed by the metric name
   higher_is_better: true
--- a/examples/configs/grpo_math_1B_megatron_single_controller.yaml
+++ b/examples/configs/grpo_math_1B_megatron_single_controller.yaml
@@ -68,6 +68,9 @@ async_rl:
 
 checkpointing:
   enabled: false
+  # set to false to start fresh in a checkpoint_dir that already has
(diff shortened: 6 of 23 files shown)


VishnuR23:
Rebased onto current `main` and re-validated.

`CheckpointingConfig` was restructured upstream in the meantime — the automodel fields moved out of the TypedDict into `_AUTOMODEL_ONLY_CHECKPOINT_FIELDS`, and `save_data_plane` / `load_replay_buffer` were added — so I resolved the conflict into the new shape rather than reinstating the old block. `resume_if_exists` now sits alongside those two fields, and the archive hook in `__init__` is the only line the constructor gains.

Also picked up the three exemplar configs added since this PR was opened (`grpo_math_1B_dynamo`, `ppo_math_1B_megatron_single_controller`, `vlm_mpo`), so every base exemplar with a full `checkpointing` block documents the flag.

Locally on the rebased branch: `tests/unit/utils/test_checkpoint.py` + `test_config_v2.py` → 94 passed; `ruff` and `minimize-check` clean. The 18 failures in `test_config_validation.py::test_all_config_no_tp_size_accuracy_issues` are on unrelated `grpo-*`/`mopd-*` recipes and are byte-identical to a clean `main` baseline.

Issue #3698, opened 2026-08-18
DFlash/DSpark and EAGLE-3 TTT draft co-training for speculative decoding (Megatron backend)


MarkXCloud:
## Motivation

NeMo-RL already ships a single-pass EAGLE-3 draft-training path (`policy.draft`). When serving RL rollouts with speculative decoding, a *static* drafter's acceptance rate degrades as the policy trains away from it — co-training the drafter alongside the policy (and refitting it into vLLM together with the policy weights) keeps acceptance high and rollout generation fast.

## What I'm working on

Extending the draft co-training path in three directions:

1. **Multi-pass TTT training for EAGLE-3** (`draft.ttt_steps > 1`): sequential draft passes with a two-part trunk+branch flash attention (exact joint-LSE merge, inference-aligned per-pass RoPE offsets), instead of the current single pass.
2. **DFlash/DSpark block drafters** (`draft.method: dflash|dspark`): anchor + mask-token block drafting (bidirectional in-block attention, trunk truncated at the anchor), following the official DFlash contract — the draft owns no lm_head / mask embedding and rides the target's live head, matching vLLM ≥ 0.26 serving.
3. **Training-loop integration**: chunked vocab-parallel soft-CE distillation losses with per-pass/per-slot global normalization, a separate draft grad-norm clipping group, an optional draft-only optimizer param group (FastGRPO-style higher LR), and vLLM refit fixes (drafter lm_head un-sharing, spec-decode max_tokens clamping).

## Scope

This covers the **Megatron** policy backend only — the automodel draft-training path is being worked on separately by others.

## Status

Split into a stacked series of three draft PRs (each PR's diff includes the ones below it until they merge):

1. #3699 — multi-pass TTT training for EAGLE-3 drafts (+ shared infra: chunked distillation CE, draft-only optimizer group, vLLM refit safety). Runs on the currently pinned vLLM.
2. #3712 — DFlash block-draft co-training (vLLM ≥ 0.25 serving).
3. #3713 — DSpark block-draft co-training with the official CE+TV+confidence loss (vLLM ≥ 0.26 serving, `drafter` → `speculator` rename tolerated).

Unit tests included; smoke-tested on a Qwen3-8B GRPO math recipe with eagle3 and dflash drafters.

Related finding, tracked separately: #3716 (split-API train step misses the TP grad all-reduce for replicated params).

Opening this issue for visibility — feedback on the direction is welcome, and happy to align with any related internal work.

PR #3699, opened 2026-08-18
feat(spec-decode): multi-pass TTT training for EAGLE-3 drafts (Megatron)

MarkXCloud:
# What does this PR do ?

Extends the single-pass EAGLE-3 draft co-training path (**Megatron** policy backend; the automodel path is being handled separately) to multi-pass TTT training (`policy.draft.ttt_steps > 1`), co-trained with the policy during GRPO and refit into vLLM alongside the policy weights.

# Issues

Part of #3698 (1/5). Stacked series: this PR → #3712 (DFlash) → #3713 (DSpark) → #4059 (sequence packing + CP) → #4060 (PP).

# Usage

```yaml
policy:
  draft:
    enabled: true
    model_name: <eagle3 draft ckpt or null>
    loss_weight: 0.1
    ttt_steps: 3            # >1 = multi-pass TTT; 1 = stock single pass
    ttt_pass_weights: null  # per-pass alphas, null = uniform
    loss_seq_chunk_size: null # seq-dim tile for the multi-pass loss internals; null = 4096
    lr: 5e-5                # optional draft-only optimizer group
```

**Pre checks**:

# Additional Information

Draft for visibility — actively being worked on. Main pieces:

- **Multi-pass TTT** (`EagleModel.forward_ttt`): sequential draft passes with a two-part trunk+branch flash attention (exact joint-LSE merge), inference-aligned per-pass RoPE offsets, non-detached hidden-state chain between passes.
- **Chunked distillation loss**: `ChunkedDistributedCrossEntropy` keeps fp32 softmax buffers chunk-sized (several TTT passes stack before the shared backward; chunk size via `policy.draft.loss_seq_chunk_size`). `DraftTTTCrossEntropyLossFn` — the multi-pass sibling of the single-pass `DraftCrossEntropyLossFn`, which stays the `ttt_steps == 1` loss (including the packed layout) — does per-pass slicing (pass d predicts x_{i+d+1}), per-pass alphas, and per-pass global valid-count normalization (DP-only all-reduce).
- **Draft-only optimizer group**: optional `draft.{lr,min_lr,weight_decay}` param group with resume-safe param-group identity keys.
- **vLLM refit safety**: un-share the drafter lm_head before loading a trained head (vLLM's share-on-identical-weights alias would let the draft refit silently overwrite the serving target's head); fail loudly when draft weights arrive but no drafter is reachable.

EAGLE-3 runs on the currently pinned vLLM. Smoke-tested on a Qwen3-8B GRPO math recipe.

A suspected TP grad all-reduce gap on the (unexercised) split-API train-step path was found during this work; it is tracked separately with its regression test in #3716 rather than bundled here.

# Evaluation verification
 - Model: Qwen3-8B
 - Benchmark: AIME24
 - Purple line: Eagle3 pretrained by AngelSlim/Qwen3-8B_eagle3
 - Pink line: Eagle3 w/ ttt=3, trained from random initialization
 - Blue line: no speculative decoding
image

Perf:
image

### DIFF
--- a/nemo_rl/algorithms/loss/utils.py
+++ b/nemo_rl/algorithms/loss/utils.py
@@ -45,6 +45,7 @@ def map_teacher_logits_to_draft_vocab(
     d2t: Optional[torch.Tensor],
     vocab_parallel_rank: Optional[int] = None,
     vocab_parallel_group: Optional[torch.distributed.ProcessGroup] = None,
+    seq_chunk_size: Optional[int] = None,
 ) -> torch.Tensor:
     """Restrict full-vocab teacher logits to the draft vocabulary via ``d2t``.
 
@@ -53,6 +54,10 @@ def map_teacher_logits_to_draft_vocab(
     are gathered to the full vocab and re-sliced to this rank's shard of the
     draft vocabulary (the draft output layer is sharded the same way). No-op
     when ``d2t`` is None (full-vocab drafts).
+
+    ``seq_chunk_size`` gathers and subsets in sequence chunks instead of all at
+    once: the full ``[B, S, V_full]`` gather peaks at several GiB at long
+    sequence lengths while only the ``[B, S, draft_vocab/tp]`` subset survives.
     """
     if d2t is None:
         return teacher_logits
@@ -64,15 +69,37 @@ def map_teacher_logits_to_draft_vocab(
             gather_from_tensor_model_parallel_region,
         )
 
-        teacher_logits = gather_from_tensor_model_parallel_region(
-            teacher_logits, vocab_parallel_group
-        )
         tp_size = torch.distributed.get_world_size(vocab_parallel_group)
         local_draft_size = len(d2t) // tp_size
         assert vocab_parallel_rank is not None
         start_index = vocab_parallel_rank * local_draft_size
         end_index = (vocab_parallel_rank + 1) * local_draft_size
         reverse_mapping = reverse_mapping[start_index:end_index]
+
+        if seq_chunk_size is None:
+            teacher_logits = gather_from_tensor_model_parallel_region(
+                teacher_logits, vocab_parallel_group
+            )
+        else:
+            batch_size, seq_size = teacher_logits.shape[:2]
+            subset_teacher = torch.empty(
+                batch_size,
+                seq_size,
+                reverse_mapping.shape[0],
+                dtype=teacher_logits.dtype,
+                device=teacher_logits.device,
+            )
+            for chunk_start in range(0, seq_size, seq_chunk_size):
+                chunk_end = min(seq_size, chunk_start + seq_chunk_size)
+                gathered_chunk = gather_from_tensor_model_parallel_region(
+                    teacher_logits[:, chunk_start:chunk_end].contiguous(),
+                    vocab_parallel_group,
+                )
+                subset_teacher[:, chunk_start:chunk_end] = gathered_chunk[
+                    :, :, reverse_mapping
+                ]
+                del gathered_chunk
+            return subset_teacher
     return teacher_logits[:, :, reverse_mapping]
 
 
@@ -337,10 +364,77 @@ def prepare_loss_input(
             "token_mask": token_mask,
         }
 
+    elif loss_fn.input_type == LossInputType.DRAFT_TTT:
+        # Multi-pass convention: pass-d student logits z^d_i predict x_{i+d+1},
+        # so the matching teacher is the policy's logits at position i+d. The
+        # teacher stays UNSHIFTED here — DraftTTTCrossEntropyLossFn slices it
+        # per pass (pure views) and builds the shifted masks with
+        # draft_pass_token_mask. Slicing replaces the CP-aware roll, so this
+        # path requires CP == 1.
+        if (
+            context_parallel_group is not None
+            and torch.distributed.is_initialized()
+            and torch.distributed.get_world_size(context_parallel_group) > 1
+        ):
+            raise NotImplementedError(
+                "Multi-pass draft distillation requires context_parallel_size == 1."
+            )
+        teacher_logits = map_teacher_logits_to_draft_vocab(
+            logits.detach(),
+            d2t,
+            vocab_parallel_rank=vocab_parallel_rank,
+            vocab_parallel_group=vocab_parallel_group,
+            seq_chunk_size=loss_fn.seq_chunk_size,
+        )
+        loss_input = {
+            "teacher_logits": teacher_logits,
+            "student_logits_by_pass": list(data["student_logits_by_pass"]),
+        }
+
     else:
         raise ValueError(f"Unknown loss function input type: {loss_fn.input_type}")
 
     return loss_input, data
+
+
+def draft_pass_token_mask(token_mask: torch.Tensor, ttt_pass: int) -> torch.Tensor:
+    """Per-pass draft loss mask (before the sample_mask factor).
+
+    Pass ``d`` position ``i`` predicts token ``x_{i+d+1}``, so its validity is
+    ``token_mask[i + d + 1]``: a left shift by ``d + 1`` that also drops the
+    out-of-bounds tail. Returned shape is ``[B, S - d - 1]``, aligned with the
+    ``student[:, :S-d-1]`` / ``teacher[:, d:S-1]`` slices used by
+    ``DraftCrossEntropyLossFn``.
+    """
+    return token_mask[:, ttt_pass + 1 :]
+
+
+def compute_draft_pass_valid_counts(
+    token_mask: torch.Tensor,
+    sample_mask: torch.Tensor,
+    *,
+    ttt_steps: int,
+) -> torch.Tensor:
+    """Local per-pass valid-token counts for the draft TTT loss.
+
+    Returns a fp32 tensor of shape ``[ttt_steps]`` where entry ``d-1`` is this
+    rank's ``#valid pass-d targets``. The caller must all-reduce it over the
+    data-parallel group (and only DP: every CP/TP rank computes the loss over
+    the full sequence/vocab, so reducing over those groups would double-count).
+
+    The loss consumes the vector twice: the gradient denominator is
+    ``sum_d alpha_d * counts[d-1]``, and the per-pass diagnostic metrics divide
+    by ``counts[d-1]`` so that the driver's sum-over-microbatches aggregation
+    (grpo.py logs mb-metric lists with ``np.sum``) reconstructs the global
+    per-token mean instead of summing per-microbatch means.
+    """
+    counts = token_mask.new_zeros((ttt_steps,), dtype=torch.float32)
+    for ttt_pass in range(1, ttt_steps + 1):
+        pass_mask = draft_pass_token_mask(token_mask, ttt_pass) * sample_mask.unsqueeze(
+            -1
+        )
+        counts[ttt_pass - 1] = pass_mask.sum().float()
+    return counts
 
 
 def _pack_input_ids(
--- a/nemo_rl/algorithms/loss/wrapper.py
+++ b/nemo_rl/algorithms/loss/wrapper.py
@@ -19,7 +19,10 @@ import torch
 import torch.distributed
 
 from nemo_rl.algorithms.loss.interfaces import LossFunction
-from nemo_rl.algorithms.loss.loss_functions import DraftCrossEntropyLossFn
+from nemo_rl.algorithms.loss.loss_functions import (
+    DraftCrossEntropyLossFn,
+    DraftTTTCrossEntropyLossFn,
+)
 from nemo_rl.distributed.batched_data_dict import BatchedDataDict
 
 Tensor = TypeVar("Tensor", bound=torch.Tensor)
@@ -243,6 +246,12 @@ class DraftLossWrapper:
       vocabulary, and the token mask is packed with the same per-sequence
       shift; ``student_logits`` must be passed explicitly (it is kept out of
       the data dict so the packing loss wrappers never slice it).
+
+    ``ttt_steps > 1`` selects the multi-pass loss
+    (:class:`DraftTTTCrossEntropyLossFn`), which returns per-pass metrics
+    alongside the loss and needs ``global_draft_pass_counts``.
+    ``draft_loss_kwargs`` are the selected LossFn's remaining ctor kwargs —
+    ``pass_weights`` / ``seq_chunk_size`` for the multi-pass loss.
     """
 
     def __init__(
@@ -251,6 +260,8 @@ class DraftLossWrapper:
         prepare_fn: Optional[Callable[Any, Any]],
         data_dict: BatchedDataDict[Any],
         loss_weight: float = 1.0,
+        draft_loss_kwargs: Optional[dict[str, Any]] = None,
+        global_draft_pass_counts: Optional[torch.Tensor] = None,
         vocab_parallel_rank: Optional[int] = None,
         vocab_parallel_group: Optional[torch.distributed.ProcessGroup] = None,
         context_parallel_group: Optional[torch.distributed.ProcessGroup] = None,
@@ -258,11 +269,13 @@ class DraftLossWrapper:
         cu_seqlens_q_padded: Optional[torch.Tensor] = None,
         d2t: Optional[torch.Tensor] = None,
         student_logits: Optional[torch.Tensor] = None,
+        ttt_steps: int = 1,
     ):
         self.loss_fn = loss_fn
         self.prepare_fn = prepare_fn
         self.data_dict = data_dict
         self.loss_weight = loss_weight
+        self.global_draft_pass_counts = global_draft_pass_counts
         self.vocab_parallel_rank = vocab_parallel_rank
         self.vocab_parallel_group = vocab_parallel_group
         self.context_parallel_group = context_parallel_group
@@ -276,9 +289,24 @@ class DraftLossWrapper:
             raise ValueError("student_logits must be passed explicitly in packed mode.")
         if cu_seqlens_q is None and prepare_fn is None:
             raise ValueError("prepare_fn is required in unpacked mode.")
-        self.draft_loss_fn = DraftCrossEntropyLossFn(
-            vocab_parallel_group=vocab_parallel_group,
-        )
+        self.multi_pass = ttt_steps > 1
+        if cu_seqlens_q is not None and self.multi_pass:
+            raise ValueError(
+                "Multi-pass draft training (policy.draft.ttt_steps > 1) does not "
+                "support sequence packing; the per-pass slicing convention needs "
+                "the unpacked [B, S] layout."
+            )
+        draft_loss_kwargs = draft_loss_kwargs or {}
+        if self.multi_pass:
+            self.draft_loss_fn: Any = DraftTTTCrossEntropyLossFn(
+                vocab_parallel_group=vocab_parallel_group,
+                **draft_loss_kwargs,
+            )
+        else:
+            self.draft_loss_fn = DraftCrossEntropyLossFn(
+                vocab_parallel_group=vocab_parallel_group,
+                **draft_loss_kwargs,
+            )
 
     def _packed_draft_loss(
         self,
@@ -339,6 +367,7 @@ class DraftLossWrapper:
             **kwargs,
         )
 
+        draft_metrics: dict[str, Any] = {}
         if self.cu_seqlens_q is not None:
             draft_loss = self._packed_draft_loss(
                 next_token_logits, data, global_valid_seqs, global_valid_toks
@@ -352,14 +381,24 @@ class DraftLossWrapper:
                 vocab_parallel_group=self.vocab_parallel_group,
                 context_parallel_group=self.context_parallel_group,
             )
-            draft_loss = self.draft_loss_fn(
-                data=data,
-                global_valid_seqs=global_valid_seqs,
-                global_valid_toks=global_valid_toks,
-                **loss_input,
-            )
+            if self.multi_pass:
+                draft_loss, draft_metrics = self.draft_loss_fn(
+                    data=data,
+                    global_valid_seqs=global_valid_seqs,
+                    global_valid_toks=global_valid_toks,
+                    global_draft_pass_counts=self.global_draft_pass_counts,
+                    **loss_input,
+                )
+            else:
+                draft_loss = self.draft_loss_fn(
+                    data=data,
+                    global_valid_seqs=global_valid_seqs,
+                    global_valid_toks=global_valid_toks,
+                    **loss_input,
+                )
         combined_loss = policy_loss + self.loss_weight * draft_loss
         metrics["draft_loss"] = float(draft_loss.detach().item())
+        metrics.update(draft_metrics)
         return combined_loss, metrics
 
 
--- a/nemo_rl/models/generation/vllm/vllm_backend.py
+++ b/nemo_rl/models/generation/vllm/vllm_backend.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
 import gc
 import logging
 import re
@@ -596,8 +597,66 @@ class VllmInternalWorkerExtension:
         because these are dynamic vLLM model classes whose ``load_weights`` /
         ``mtp_start_layer_idx`` members are not visible through ``nn.Module``.
         """
-        draft_owner = getattr(self.model_runner, "drafter", None)
+        # The V1 model runner names the owner `drafter`; the V2 runner
+        # `speculator`. On both, non-last PP ranks carry the attribute as None.
+        draft_owner = getattr(self.model_runner, "drafter", None) or getattr(
+            self.model_runner, "speculator", None
+        )
         return getattr(draft_owner, "model", None) if draft_owner else None
+
+    def _unshare_draft_lm_head(self, draft_model: torch.nn.Module) -> None:
+        """Give the drafter a private lm_head before loading a trained head.
+
+        vLLM's spec-decode proposer aliases the drafter's ``lm_head`` MODULE to
+        the target's when their weights are identical at engine init — and
+        unconditionally for drafters without a ``has_own_lm_head`` flag
+        (``SpecDecodeBaseProposer._maybe_share_lm_head``, vLLM 0.26). For a
+        draft that TRAINS its own head (Eagle), the refit stream would then
+        write the diverged draft head through the alias, silently overwriting
+        the serving target's head (generation no longer matches the trained
+        policy — broken importance ratios / KL). Deep-copy the head into
+        drafter-private storage before such a load; later refits see distinct
+        storage and no-op.
+
+        Only called when the incoming draft weights actually contain an
+        lm_head update: head-less drafts WANT the sharing to persist — their
+        draft logits must track the target's live head. Embedding sharing is
+        likewise left intact.
+        """
+        target_model = getattr(self.model_runner, "model", None)
+        if target_model is not None and hasattr(target_model, "get_language_model"):
+            target_model = target_model.get_language_model()
+        draft_lm_head = getattr(draft_model, "lm_head", None)
+        target_lm_head = getattr(target_model, "lm_head", None)
+        if not isinstance(target_lm_head, torch.nn.Module):
+            return
+        draft_weight = getattr(draft_lm_head, "weight", None)
+        target_weight = getattr(target_lm_head, "weight", None)
+        if not isinstance(draft_weight, torch.Tensor) or not isinstance(
+            target_weight, torch.Tensor
+        ):
+            return
+        if (
+            draft_weight.untyped_storage().data_ptr()
+            != target_weight.untyped_storage().data_ptr()
+        ):
+            return
+        private_lm_head = copy.deepcopy(target_lm_head)
+        # Parameter.__deepcopy__ rebuilds bare Parameters, dropping the attrs
+        # vLLM attaches via set_weight_attrs (weight_loader, output_dim, ...);
+        # without them a later load_weights falls back to
+        # default_weight_loader, which cannot shard under TP. The loaders take
+        # the param explicitly, so rebinding the originals is safe.
+        for private_param, source_param in zip(
+            private_lm_head.parameters(), target_lm_head.parameters()
+        ):
+            private_param.__dict__.update(source_param.__dict__)
+        draft_model.lm_head = private_lm_head
+        print(
+            "[draft] Drafter lm_head was aliased to the target's (vLLM "
+            "share-on-identical-weights); un-shared it into private storage "
+            "before applying draft refit weights."
+        )
 
     def _load_draft_weights(
         self, draft_weights: list[tuple[str, torch.Tensor]]
@@ -607,10 +666,23 @@ class VllmInternalWorkerExtension:
 
         draft_model = self._get_drafter_model()
         if draft_model is None:
-            logger.warning(
-                "[draft] Received draft weights but vLLM drafter is unavailable; skipping draft update."
+            # vLLM places the drafter on the last PP stage only; its absence
+            # on earlier stages is expected (cf. load_mtp_weights_from_disk).
+            if not get_pp_group().is_last_rank:
+                return
+            # The trainer only streams draft.* weights when a draft model is
+            # co-training, so an unreachable drafter here means the weights
+            # would be silently dropped and serving would keep stale draft
+            # weights forever (observed: acceptance rate pinned at exactly 0
+            # while draft_loss fell). Fail loudly instead.
+            raise RuntimeError(
+                "[draft] Received draft weights but no vLLM drafter was found "
+                "on the model runner; the draft refit would be silently "
+                "dropped. Check the speculative_config and the vLLM "
+                "model-runner version compatibility."
             )
-            return
+        if any("lm_head" in name for name, _ in draft_weights):
+            self._unshare_draft_lm_head(draft_model)
         draft_weights = self._trim_vocab_padding(draft_model, draft_weights)
         draft_model.load_weights(weights=draft_weights)
 
--- a/nemo_rl/models/megatron/draft/eagle.py
+++ b/nemo_rl/models/megatron/draft/eagle.py
@@ -12,12 +12,47 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Megatron training implementation of the EAGLE-3 draft model.
+
+This file wraps ModelOpt's ``EagleModule`` and adds two training paths:
+
+- :meth:`EagleModel.forward` runs one causal draft pass when ``ttt_steps == 1``.
+- :meth:`EagleModel.forward_ttt` runs several sequential TTT passes.
+  Each pass feeds its hidden state and shifted token embeddings into the next
+  pass, matching the draft model's self-conditioning during serving.
+
+For a query at anchor ``i`` in TTT pass ``d``, attention can read:
+
+- the pass-1 trunk at positions ``0..i``; and
+- the entry at anchor ``i`` from every pass ``2..d``.
+
+It cannot read a later trunk position or another anchor's branch. This matches
+the serving cache: causal prefill entries followed by one draft-generated entry
+for each speculation depth. :class:`TwoPartTTTAttention` shows the full mask.
+
+The causal trunk uses FlashAttention. The short per-anchor branch uses an
+einsum. Their output and log-sum-exp values are merged to produce the exact
+softmax over both sets of keys; backward uses the same joint values to produce
+the exact gradients.
+
+The implementation uses private FlashAttention functions because the merge
+needs their log-sum-exp values. Their interface is pinned to the compatible
+FlashAttention 2.x versions checked below.
+"""
+
 from __future__ import annotations
 
 from contextlib import contextmanager, nullcontext
-from typing import Optional, Tuple
+from typing import Any, Optional, Tuple
 
+import flash_attn
 import torch
+from flash_attn.flash_attn_interface import (
+    _flash_attn_backward,
+    _flash_attn_forward,
+    flash_attn_func,
+)
+from megatron.core import parallel_state
 from megatron.core.dist_checkpointing.mapping import ShardedStateDict
 from megatron.core.models.common.embeddings import RotaryEmbedding
 from megatron.core.packed_seq_params import PackedSeqParams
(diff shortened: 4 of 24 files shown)


tests/unit/models/megatron/test_draft_optimizer_overrides.py
```diff
@@ -0,0 +1,207 @@
...
+        "lr_mult": 1.0,
+        "is_expert_parallel": False,
+        "is_decoupled_lr": False,
+        "max_lr": max_lr,
+        "min_lr": min_lr,
+        "params": params,
+    }
+
+
+def test_identifier_keys_extended_and_idempotent():
```
zpqiu, line 96: [`tests/unit/models/megatron/test_draft_optimizer_overrides.py:96`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/tests/unit/models/megatron/test_draft_optimizer_overrides.py#L96)

**Good news: upstream already fixed this, so this helper and these tests can go.**

`extend_param_group_identifier_keys_for_resume()` is a no-op against the pinned Megatron-LM. `param_group_identifier_keys` is now built from `ParamGroupOverride.__annotations__` and already contains `max_lr`/`min_lr`/`start_wd`/`end_wd` ([`optimizer.py#L139`](https://github.com/NVIDIA/Megatron-LM/blob/7c9c3a027c503ae9ae1e8ad7b14397abb8269378/megatron/core/optimizer/optimizer.py#L139)). Verified at runtime in all three patched modules:

```
('end_wd','max_lr','min_lr','optimizer','start_wd','wd_mult','lr_mult','is_expert_parallel','is_decoupled_lr')
```

The upstream fix is `27e312ca1` (Megatron-LM #4705), confirmed an ancestor of the pinned SHA. Your patch was correct against the older submodule; `main` bumped it in a0edb2761, so the code is now **redundant rather than wrong** — you can delete `_PARAM_GROUP_IDENTIFIER_EXTRA_KEYS`, the helper, and its call site.

Three tests here currently fail as a result:
- `test_identifier_keys_extended_and_idempotent` — `AssertionError` (hard-codes the stale base tuple)
- `test_resume_group_matching_keeps_policy_and_draft_hyperparams` — `KeyError: 'end_wd'`
- `test_draft_override_indistinct_from_policy_is_rejected` — see the separate note below; that one is unrelated to the submodule.

If you replace them, a behavioural round-trip through `_filter_and_reorder_param_groups` would survive the next Bridge bump better than re-hardcoding the 9-tuple.

Two stale docstring premises to drop with it: `start_wd`/`end_wd` **are** identifier keys, and a missing key does not "raise at load" — [`optimizer.py#L613-L622`](https://github.com/NVIDIA/Megatron-LM/blob/7c9c3a027c503ae9ae1e8ad7b14397abb8269378/megatron/core/optimizer/optimizer.py#L613) maps it to `None` ("Treat missing and explicit None identifier values as equivalent").

nemo_rl/models/megatron/draft/utils.py
```diff
@@ -1252,6 +1258,108 @@ def register_draft_grad_norm_group() -> None:
...
+        elif self.draft_min_lr is not None:
+            draft_override["min_lr"] = float(self.draft_min_lr)
+        if self.draft_weight_decay is not None:
+            draft_override["start_wd"] = float(self.draft_weight_decay)
+            draft_override["end_wd"] = float(self.draft_weight_decay)
+
+        base_config = context.optimizer_config
+        effective_max_lr = draft_override.get("max_lr", base_config.lr)
+        effective_min_lr = draft_override.get("min_lr", base_config.min_lr)
+        if (effective_max_lr, effective_min_lr) == (base_config.lr, base_config.min_lr):
```
zpqiu, line 1330: [`nemo_rl/models/megatron/draft/utils.py:1330`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/draft/utils.py#L1330)

**A weight-decay-only draft override is accepted at the entry point but rejected here.**

`build_draft_optimizer_override_provider` returns a provider when **any** of `lr`/`min_lr`/`weight_decay` is set, but with `weight_decay` alone the override carries no `max_lr`/`min_lr`, so both `effective_*` fall back to the base config and this raises. So `policy.draft.weight_decay: 0.0` with `lr: null` is a supported entry path that hard-fails at setup.

Since `start_wd`/`end_wd` are already identifier keys (see the note on the optimizer test), that config is distinguishable at resume and the collision this guards against can't occur. The guard is worth keeping — the collision is real, upstream documents it at [`optimizer.py#L636-L648`](https://github.com/NVIDIA/Megatron-LM/blob/7c9c3a027c503ae9ae1e8ad7b14397abb8269378/megatron/core/optimizer/optimizer.py#L636) — it just needs to test the whole distinguishing set:

```suggestion
        base_config = context.optimizer_config
        effective_max_lr = draft_override.get("max_lr", base_config.lr)
        effective_min_lr = draft_override.get("min_lr", base_config.min_lr)
        # Must differ from the policy group in at least one of mcore's
        # param_group_identifier_keys, or the two collapse at checkpoint load
        # and swap params. The policy group carries no start_wd/end_wd at all,
        # so setting them on the draft is itself distinguishing.
        distinguishes = (
            "start_wd" in draft_override
            or "end_wd" in draft_override
            or (effective_max_lr, effective_min_lr)
            != (base_config.lr, base_config.min_lr)
        )
        if not distinguishes:
            raise ValueError(
                "[draft] draft (lr, min_lr, weight_decay) must differ from the "
                "policy's in at least one field."
            )
```

Separately, `test_draft_override_indistinct_from_policy_is_rejected` fails on a plain string drift — it expects `'distinct from the policy'`, the message says `"must differ from the policy's"`. Unrelated to the submodule bump.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -607,10 +651,19 @@ def _load_draft_weights(
 
         draft_model = self._get_drafter_model()
         if draft_model is None:
-            logger.warning(
-                "[draft] Received draft weights but vLLM drafter is unavailable; skipping draft update."
+            # The trainer only streams draft.* weights when a draft model is
+            # co-training, so an unreachable drafter here means the weights
+            # would be silently dropped and serving would keep stale draft
+            # weights forever (observed: acceptance rate pinned at exactly 0
```
zpqiu, line 657: [`nemo_rl/models/generation/vllm/vllm_backend.py:657`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/generation/vllm/vllm_backend.py#L657)

**Failing loudly is the right instinct, but there are two configurations where a missing drafter is expected — and one of them has a correct analog in this same file.**

`load_mtp_weights_from_disk` ([`nemo_rl/models/generation/vllm/vllm_backend.py:748`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/generation/vllm/vllm_backend.py#L748), pre-existing) handles this exact state:

```python
draft_model = self._get_drafter_model()
if draft_model is None:
    # vLLM places the speculative drafter only on the last pipeline
    # stage. Its absence is expected on every earlier stage, ...
    if get_pp_group().is_last_rank:
        raise RuntimeError(...)
    return False
```

`_load_weights` calls `_load_draft_weights` on **every** worker, so two disjoint configs reach this raise:

1. **vLLM PP > 1** → drafter exists only on the last rank, by design on both runners ([V1 `gpu_model_runner.py#L570`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/worker/gpu_model_runner.py#L570), [V2 `gpu/model_runner.py#L189`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/worker/gpu/model_runner.py#L189)). Every earlier rank raises. `policy.generation.vllm_cfg.pipeline_parallel_size` is ungated for draft — the guard at [`nemo_rl/models/policy/lm_policy.py:146`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/policy/lm_policy.py#L146) is conjunctive with sequence packing and gates the *training*-side PP.
2. **vLLM PP == 1, dense target** → vLLM selects the V2 runner by default (`_is_default_v2_model_runner_model` returns `... or not is_moe`; eagle3 is on the supported list at [`config/vllm.py#L2068`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/config/vllm.py#L2068)), and V2 names the attribute **`speculator`**, not `drafter` — so `_get_drafter_model` returns `None`.

Note `_get_drafter_model` and the old `logger.warning(...); return` are both pre-existing on `main`; the change here is warn → raise, which makes an existing silent no-op fatal.

Both parts are needed — on V2 `speculator` *exists but is `None`* on non-last ranks, so the fallback alone isn't sufficient:

```suggestion
        draft_model = self._get_drafter_model()
        if draft_model is None:
            # vLLM places the drafter on the last PP stage only; its absence on
            # earlier stages is expected (cf. load_mtp_weights_from_disk).
            if not get_pp_group().is_last_rank:
                return
            raise RuntimeError(
                "[draft] Received draft weights but no vLLM drafter was found "
                "on the model runner; the draft refit would be silently "
                "dropped. Check the speculative_config and the vLLM "
                "model-runner version compatibility."
            )
```
plus in `_get_drafter_model`:
```python
        draft_owner = getattr(self.model_runner, "drafter", None) or getattr(
            self.model_runner, "speculator", None
        )
```

One question so we're not guessing about your runs: does your engine log contain `Using V2 Model Runner` ([`gpu_worker.py#L359`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/worker/gpu_worker.py#L359))? Your AIME24 results imply a reachable drafter, so you may well be on V1 — the fallback is correct either way, but it would tell us whether the acceptance-rate-0 you mention above was this.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -599,6 +599,50 @@ def _get_drafter_model(self) -> Any:
...
+        ):
+            return
+        if (
+            draft_weight.untyped_storage().data_ptr()
+            != target_weight.untyped_storage().data_ptr()
+        ):
+            return
+        import copy
+
+        draft_model.lm_head = copy.deepcopy(target_lm_head)
```
zpqiu, line 639: [`nemo_rl/models/generation/vllm/vllm_backend.py:639`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/generation/vllm/vllm_backend.py#L639)

**`pyrefly` fails on this line, and while you're here there's a subtlety worth guarding.**

The typecheck (a pre-commit hook, `pass_filenames: false`) reports a new error introduced by this line — `main` has 0, this branch has 1:
```
ERROR Argument `Any | None` is not assignable to parameter `value` with type `Module | Tensor`
  in function `torch.nn.modules.module.Module.__setattr__` [bad-argument-type]
  --> nemo_rl/models/generation/vllm/vllm_backend.py:639
```
An `isinstance(target_lm_head, torch.nn.Module)` narrow above the storage check fixes it.

The separate subtlety: `torch.nn.Parameter.__deepcopy__` rebuilds a **bare** Parameter, dropping attributes vLLM attaches via `set_weight_attrs` ([`vocab_parallel_embedding.py#L58`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/layers/vocab_parallel_embedding.py#L58)). Reproduced on torch 2.11.0+cu130:
```
ORIGINAL attrs: ['weight_loader', 'input_dim', 'output_dim']
DEEPCOPY attrs: []
```
`load_weights` resolves the loader off the param ([`models/utils.py#L277`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/models/utils.py#L277)) and would fall back to `default_weight_loader`, which can't shard and asserts under TP>1. The deep-copied head also carries the **target's** `org_vocab_size`, which `_trim_vocab_padding` then reads for the draft.

Worth noting this is currently unreachable, so it's a latent issue rather than something you're hitting: NeMo-RL's own `has_own_lm_head` patch ([`nemo_rl/models/generation/vllm/patches.py:161`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/generation/vllm/patches.py#L161)) sets `draft_vocab_size != vocab_size`, and both shipped drafters are truncated-vocab (`AngelSlim/Qwen3-{1.7B,8B}_eagle3`: `vocab_size=151936`, `draft_vocab_size=32000`), so vLLM never shares the head and this method early-returns. It would bite a full-vocab drafter. Building a fresh module and `copy_()`-ing the data is more robust than deepcopy-plus-reattach.

nemo_rl/models/megatron/draft/eagle.py
```diff
@@ -29,11 +64,370 @@
...
+
+    def forward(
+        self,
+        query: Tensor,
+        key: Tensor,
+        value: Tensor,
+        attention_mask: Optional[Tensor],
+        attn_mask_type: Optional[Any] = None,
+        attention_bias: Optional[Tensor] = None,
+        packed_seq_params: Optional[Any] = None,
```
zpqiu, line 384: [`nemo_rl/models/megatron/draft/eagle.py:384`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/draft/eagle.py#L384)

**`test_module_guards_do_not_require_gpu` fails here — the test is right and the guard is missing.**

The test asserts `pytest.raises(ValueError, match="attention_mask")`, but `forward` never inspects `attention_mask` (or `attention_bias`); the call falls through to `flash_attn_func` and dies with a CPU-backend error instead.

Worth fixing in production rather than relaxing the test: this module is installed permanently on every eagle decoder layer as a drop-in for MCore's `core_attention`, and modelopt builds those layers with `AttnMaskType.arbitrary` — a spec that expects an explicit mask. `forward_ttt` passes `None` today, but silently dropping a caller-supplied padding mask is the wrong default.

```suggestion
        if attention_mask is not None:
            raise ValueError(
                "TTTDraftCoreAttention builds its own trunk/branch mask and "
                "cannot honour an external attention_mask."
            )
        if attention_bias is not None:
            raise ValueError("TTTDraftCoreAttention does not support attention_bias.")
```

One heads-up if you also add the re-entrancy assertion discussed elsewhere: it can't live in this test, because pass 1 reaches `flash_attn_func` immediately so `forward` isn't CPU-drivable. It needs to sit before `self._kv_by_pass.append(...)` and be driven via module state.

nemo_rl/algorithms/loss/utils.py
```diff
@@ -337,12 +364,79 @@ def prepare_loss_input(
...
+        teacher_logits = map_teacher_logits_to_draft_vocab(
+            logits.detach(),
+            d2t,
+            vocab_parallel_rank=vocab_parallel_rank,
+            vocab_parallel_group=vocab_parallel_group,
+            seq_chunk_size=loss_fn.seq_chunk_size,
+        )
+        loss_input = {
+            "teacher_logits": teacher_logits,
+            "student_logits_by_pass": list(data["student_logits_by_pass"]),
```
zpqiu, line 391: [`nemo_rl/algorithms/loss/utils.py:391`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/algorithms/loss/utils.py#L391)

**`test_prepare_loss_input_falls_back_to_single_pass_key` fails here** — it asserts a `student_logits` fallback that this bare subscript doesn't implement (`KeyError: 'student_logits_by_pass'`).

Suggest **deleting the test rather than adding the fallback**: a fallback would let a `ttt_steps=3` run silently train single-pass, which is worse than a loud `KeyError`.

But the test is pointing at something real. `ttt_steps` currently has two sources of truth:
- forward side — the `draft_ttt_steps` parameter, **defaulting to `1`** ([`nemo_rl/models/megatron/train.py:233`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/train.py#L233)), which decides whether `data_dict` gets `student_logits` or `student_logits_by_pass`;
- loss side — `int(draft_cfg.get("ttt_steps", 1) or 1)` ([`nemo_rl/models/megatron/train.py:659`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/train.py#L659)), which decides which loss class is built.

Any caller that omits the parameter while config says `3` produces exactly this `KeyError`, and the default of `1` makes the omission invisible at the call site. Deriving the loss-side branch from what the forward actually produced would make the mismatch unrepresentable — at which point the test is moot rather than merely deleted:

```suggestion
                ttt_steps = int(draft_cfg["ttt_steps"])
                multi_pass = "student_logits_by_pass" in data_dict
                if multi_pass != (ttt_steps > 1):
                    raise RuntimeError(
                        f"draft forward produced "
                        f"{'multi-pass' if multi_pass else 'single-pass'} logits "
                        f"but policy.draft.ttt_steps={ttt_steps}; the "
                        "draft_ttt_steps argument threaded into "
                        "megatron_forward_backward disagrees with the config."
                    )
```

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1507,6 +1539,12 @@ def _train_microbatch_body(
         placeholder_n = torch.tensor(1.0, device="cuda")
 
         draft_enabled = "draft" in self.cfg and self.cfg["draft"]["enabled"]
+        # With TTT enabled the draft's core attention is the multi-pass module,
+        # so the forward MUST run through forward_ttt even here (the plain
+        # single-pass forward would trip its begin_pass guard).
+        draft_ttt_steps = (
+            int(self.cfg["draft"].get("ttt_steps", 1) or 1) if draft_enabled else 1
```
zpqiu, line 1546: [`nemo_rl/models/policy/workers/megatron_policy_worker.py:1546`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1546)

**The split-API path threads `draft_ttt_steps` but not `global_draft_pass_counts`, which scales the draft loss by roughly `ttt_steps`.**

With counts `None`, `DraftTTTCrossEntropyLossFn` falls back to `denominator = global_valid_toks` — but here that's `placeholder_n = 1.0`, so the loss returns `sum_d alpha_d * pass_sum_d`. `_finish_train_step_body` then applies a **single** `scale_gradients(1/global_valid_toks)` to all params, while the draft's correct denominator is `sum_d alpha_d * count_d`. Error factor ≈ `sum_d alpha_d`, i.e. **~3× at the documented `ttt_steps: 3`** (second-order drift with mask density).

Threading the counts in wouldn't fix it — the loss would return an already-normalized value and `finish` would divide again. One scalar can't express two denominators. The placeholder-N contract is exact for the *single-pass* draft precisely because its correct denominator genuinely is `global_valid_toks`; multi-pass is the first draft loss where that stops being true.

You already note this path is unexercised and tracked in #3716, so rather than fixing it here, a guard matching the four this PR already adds would keep it from silently mis-scaling:

```suggestion
        if draft_ttt_steps > 1:
            # The split API returns un-normalized sums and applies one global
            # 1/N at finish_train_step, so it cannot express the draft's
            # per-pass denominator (sum_d alpha_d * count_d) alongside the
            # policy's global_valid_toks.
            raise NotImplementedError(
                "policy.draft.ttt_steps > 1 is not supported on the split "
                "train-step API (single controller); use the batched train() path."
            )
```

nemo_rl/models/megatron/train.py
```diff
@@ -279,8 +281,25 @@ def forward_with_post_processing_fn(
...
+    # target_layer_ids / policy.draft.aux_layer_indices, resolved onto
+    # draft_model.config at build time): the serving-side drafter taps exactly
+    # these policy layers, so capturing the hard-coded defaults instead would
+    # silently train on different features (or break the fc width when the
+    # counts differ).
+    aux_layer_indices = None
+    if draft_model is not None:
+        configured_aux_layers = getattr(
+            getattr(draft_model, "config", None),
+            "eagle_aux_hidden_state_layer_ids",
```
zpqiu, line 295: [`nemo_rl/models/megatron/train.py:295`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/train.py#L295)

**Under PP > 1 this resolves to a different aux-layer list on each stage.**

`draft_model` is attached only to the post-process chunk, so `find_draft_owner_chunk` returns `None` on every non-last stage ([`nemo_rl/models/megatron/draft/utils.py:975`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/draft/utils.py#L975)) and `aux_layer_indices` is `None` there. `HiddenStateCapture.__init__` then falls back to `get_eagle3_aux_hidden_state_layers(num_layers)` ([`nemo_rl/models/megatron/draft/hidden_capture.py:66`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/draft/hidden_capture.py#L66)), while the last stage uses the configured ids. `_gather_distributed` iterates that list posting one send/recv per index ([`nemo_rl/models/megatron/draft/hidden_capture.py:251`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/draft/hidden_capture.py#L251)), so two different lists means mismatched counts — a wrong-tensor match that shapes won't catch, then a desync.

This is **not** reachable on the shipped recipes: neither `AngelSlim/Qwen3-1.7B_eagle3` nor `Qwen3-8B_eagle3` carries `eagle_aux_hidden_state_layer_ids`, so `build_draft_model` falls back to the same `get_eagle3_aux_hidden_state_layers(model_provider.num_layers)` and the lists coincide. The reachable trigger is the documented knob: `policy.draft.aux_layer_indices: [...]` with `pipeline_model_parallel_size > 1` and packing off (which the guard at [`nemo_rl/models/policy/lm_policy.py:146`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/policy/lm_policy.py#L146) permits).

Flagging honestly: this is a static trace — I did not run it on two PP stages. Resolving the indices from a rank-independent source (or all-gathering them over the PP group and asserting agreement) would remove the divergence.

Minor, same block: the nested `getattr(getattr(draft_model, "config", None), ...)` can be a direct read since `build_draft_model` always sets that attribute.

nemo_rl/models/policy/__init__.py
```diff
@@ -502,6 +502,27 @@ class DraftConfig(TypedDict):
     loss_weight: NotRequired[float]
     num_layers: NotRequired[int | None]
     aux_layer_indices: NotRequired[list[int] | None]
+    # Multi-pass TTT draft training. ttt_steps=1 (default) keeps the stock
+    # single-pass path; >1 runs that many sequential draft passes with the
+    # trunk+branch two-part attention (requires CP=1, no sequence parallel,
+    # no sequence packing, no fused linear logprobs). Pass-d RoPE positions
+    # are offset by d-1 (inference-aligned, matching modelopt's reference
+    # TTT) — not configurable.
```
zpqiu, line 510: [`nemo_rl/models/policy/__init__.py:510`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/policy/__init__.py#L510)

**These six new keys need to reach the YAMLs — one of them is already failing CI.**

`test_config_v2.py::test_reference_configs_up_to_date[grpo_math_1B.yaml]` fails today:
```
assert 1 == 0
 +  where 1 = len(['policy.draft.loss_seq_chunk_size'])
```
`examples/configs/grpo_math_1B.yaml` gained the key; `tests/unit/reference_configs/grpo_math_1B.yaml` didn't.

`DraftConfig` is a v1 `TypedDict`, so per the config-conventions skill the exemplar YAML *is* the source of truth for its defaults. Exactly three files need all six keys (the two recipes with a `draft:` block inherit via `defaults:` and must stay minimal, or `configs-minimize-check` will flag them):
- `examples/configs/grpo_math_1B.yaml`
- `examples/configs/ppo_math_1B_megatron_single_controller.yaml`
- `tests/unit/reference_configs/grpo_math_1B.yaml`

Related, same root cause — `ttt_steps` currently carries a hidden default at five call sites (`train.py:659`, `draft/utils.py:1471`, `lm_policy.py:178`, `megatron_policy_worker.py:871` and `:1546`). Once the YAML carries the key these can read it directly. Three details worth folding in:
- `seq_chunk_size: int = 4096` ([`nemo_rl/algorithms/loss/loss_functions.py:155`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/algorithms/loss/loss_functions.py#L155)) puts the real default in Python while the YAML comment says `null = 4096`; better to put `4096` in the exemplar.
- `draft_cfg.get("method", "eagle3")` ([`nemo_rl/models/megatron/train.py:658`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/train.py#L658)) reads a key declared nowhere and set in no YAML, which makes `raise ValueError(f"Unknown draft_method ...")` unreachable. Presumably forward-plumbing for #3712 — worth declaring it or landing it there.
- `or 1` silently turns `ttt_steps: 0` into `1`, while `-1` raises. Dropping `or 1` routes both through the existing `ValueError`.

tests/unit/models/generation/test_vllm_draft_lm_head_unshare.py
```diff
@@ -0,0 +1,90 @@
...
+the serving target's head. ``_unshare_draft_lm_head`` must detect the alias
+and give the drafter private storage — and must not touch anything otherwise.
+"""
+
+import types
+
+import pytest
+import torch
+
+pytest.importorskip("vllm", reason="vllm_backend requires a vLLM install")
```
zpqiu, line 30: [`tests/unit/models/generation/test_vllm_draft_lm_head_unshare.py:30`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/tests/unit/models/generation/test_vllm_draft_lm_head_unshare.py#L30)

**This file never executes in CI** — verified `Running 0 items in this shard`.

`L0_Unit_Tests_Vllm_1.sh` runs two passes: a base run that globs `test_vllm*.py` with `--no-sync` (vLLM is an optional extra, so `importorskip` skips the module), and a `--extra vllm --vllm-only` catch-all that collects only marker-carrying items. With `importorskip` but no marker, the file falls through both. `test_nccl_reshard_backend.py` is the exemplar — it pairs the marker with an `importorskip`, exactly this situation.

Verified fix takes it from 0 collected to 3 passing:
```suggestion
pytestmark = pytest.mark.vllm

pytest.importorskip("vllm", reason="vllm_backend requires a vLLM install")
```

Worth pairing with one assertion, since the current stubs would pass even with the `weight_loader` issue above present (`draft_model` is a `SimpleNamespace`, so `nn.Module.__setattr__` is never exercised, and no `ParallelLMHead` is ever built): after un-sharing, `assert hasattr(draft_model.lm_head.weight, "weight_loader")`. The `_load_draft_weights` raise path is also untested — a `pytest.raises` there would have caught the PP case above.

tests/unit/models/megatron/test_train.py
```diff
@@ -791,6 +791,72 @@ def test_forward_with_draft_model_packed_shifts_ids_and_reembeds(
         )
         assert data_dict["student_logits"] is student_logits
 
+    @patch("megatron.core.transformer.multi_token_prediction.roll_tensor")
+    @patch("nemo_rl.models.megatron.train.get_context_parallel_group")
+    @patch("nemo_rl.models.megatron.train.get_capture_context")
+    @patch("nemo_rl.models.megatron.train.model_forward")
+    def test_forward_with_ttt_draft_model_uses_forward_ttt(
```
zpqiu, line 798: [`tests/unit/models/megatron/test_train.py:798`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/tests/unit/models/megatron/test_train.py#L798)

**This test fails** with `AssertionError` at `train.py:407` (`assert original_seq_length is not None`). Both passing sibling tests pass `original_seq_length=3` to `ProcessedMicrobatch`; this one omits it. Verified passing with:

```suggestion
            packed_seq_params=None,
            cu_seqlens_padded=None,
            original_seq_length=3,
        )
```

Also worth knowing: `pyproject.toml:723` puts `-x` in `addopts`, so a plain local run stops at the first failure and reports 1 of the 7. `--maxfail=100` shows them all — that may be why these went unnoticed.

nemo_rl/algorithms/loss/wrapper.py
```diff
@@ -276,9 +286,37 @@ def __init__(
...
+        if cu_seqlens_q is not None and self.multi_pass:
+            raise ValueError(
+                "Multi-pass draft training (policy.draft.ttt_steps > 1) does not "
+                "support sequence packing; the per-pass slicing convention needs "
+                "the unpacked [B, S] layout."
+            )
+        draft_loss_kwargs = draft_loss_kwargs or {}
+        # Per-method losses are imported in-branch: only the selected one loads.
+        if draft_method == "eagle3":
+            if self.multi_pass:
```
zpqiu, line 300: [`nemo_rl/algorithms/loss/wrapper.py:300`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/algorithms/loss/wrapper.py#L300)

A few low-severity items, grouped so they're easy to skim or skip.

**In-branch imports.** The comment says "only the selected one loads", but both classes live in `nemo_rl/algorithms/loss/loss_functions.py`, so importing either loads the module in full — and this PR removed the pre-existing module-level import to get here. The linting skill allows deferred imports to break a cycle or skip a heavy optional dep, with a one-line comment saying which. Hoisting both back to the top is simpler.

**Stdlib `import copy`** inside `_unshare_draft_lm_head` ([`nemo_rl/models/generation/vllm/vllm_backend.py:637`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/generation/vllm/vllm_backend.py#L637)) — the same skill says deferring stdlib "buys nothing — hoist it".

**Sibling return types.** `DraftCrossEntropyLossFn.__call__ -> Tensor` vs `DraftTTTCrossEntropyLossFn.__call__ -> tuple[Tensor, dict]`, which forces the `if self.multi_pass:` fork below. Returning `(loss, {})` from the single-pass class would collapse it.

**`ttt_pass_weights` length** is validated per-microbatch inside `__call__`, though `ttt_steps` is known at construction — `lm_policy.py` next to the packing guard would fail at setup instead.

**Guard placement.** `ttt_steps > 1` + packing is rejected in `lm_policy.__init__`, but + `sequence_parallel` / `context_parallel_size` only inside `build_draft_model` on the worker. Moving those two up would give all three the same fast failure.

**Pass-1 masking differs between the two paths.** Single-pass masks position `i` by `token_mask[i+1]`; TTT pass 1 uses `token_mask[i+2]` ([`nemo_rl/algorithms/loss/utils.py:400`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/algorithms/loss/utils.py#L400)). Each is internally consistent with its own normalizer and neither looks wrong, but flipping `ttt_steps` 1→2 silently changes what the shared pass optimizes, so an A/B isn't quite like-for-like. A one-line comment saying the difference is deliberate would settle it for the next reader.

**`import flash_attn` at module scope** ([`nemo_rl/models/megatron/draft/eagle.py:48`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/draft/eagle.py#L48)) with a hard version `raise`: `draft/__init__.py` pulls in `eagle.py`, and `megatron_policy_worker.py:2351` imports the draft package unguarded on the refit path, so the gate runs on every Megatron run including `draft.enabled=false`. It can't misfire today (flash-attn is pinned `==2.8.1` in four places), but a future bump — driven by TE/xformers constraints, per the comment at `pyproject.toml:205-207` — would take down all Megatron training with a message about the drafter. Moving the import and check into the `ttt_steps > 1` branch scopes it to the feature.

**Unreachable guard.** `EagleModel.__init__`'s `recompute_granularity` check ([`nemo_rl/models/megatron/draft/eagle.py:460`](https://github.com/NVIDIA-NeMo/RL/blob/e93d5d82955cc414b5c94ada8321e80e78d04a3e/nemo_rl/models/megatron/draft/eagle.py#L460)) reads the draft's own `TransformerConfig`, which `build_draft_model` never sets it on — so it can't fire. Worth keeping as insurance, but the message should say it's the draft's config. The reassuring corollary: policy activation checkpointing does **not** propagate to the draft, so it doesn't block `ttt_steps > 1`.

zpqiu (commented):
Review of the multi-pass TTT draft-training path, focused on correctness and CI state. This was a deep pass — six reviewers across the RL code, Megatron/flash-attn call chains, the vLLM refit path, and the test suite, with findings adversarially challenged before landing here. Roughly 35 raw findings were reduced to the 12 comments below; a dozen were dropped as unreachable, pre-existing, or style, and nine were withdrawn by their own authors under challenge.

**The hardest part of this PR is correct.** The trunk+branch joint-LSE attention was verified three independent ways: analytically against pinned flash-attn 2.8.1 source (the `softmax_lse` natural-log base and the FA-backward-with-joint-`(out, lse)` trick both check out), empirically on an H100 against a dense fp32 full-attention reference (`fwd 3.09e-03`, worst gradient `5.44e-03` in bf16), and by your own tests, which compare against modelopt's `set_multi_step_attention_mask` as an external oracle. `ChunkedDistributedCrossEntropy` was separately confirmed correct on two GPUs against a full-vocab reference. The per-pass RoPE offsets, GQA head grouping, and DP-only count all-reduce all check out too. That's a stronger evidence base than most PRs this size carry, and the AIME24 three-arm plot is a genuinely useful convergence check.

**The main ask is CI.** Seven tests currently fail and `pyrefly` reports a new error. Since `pyproject.toml:723` puts `-x` in `addopts`, a local run stops at the first one — `--maxfail=100` shows them all, which may be why this wasn't visible. Three of the failures are simply the Megatron-LM submodule moving under you (`main` bumped it in a0edb2761 and upstream fixed the underlying issue), so the fix there is deleting code rather than writing it.

Two items are worth a look even though they aren't reachable on the shipped recipes — the vLLM drafter lookup and the PP aux-layer divergence — because both fail silently or fatally rather than loudly, and both have small fixes.

Documentation comments are deliberately omitted, since that box is unchecked. One thing that would round out the evidence when you're ready for review: spec-decode acceptance rate (mean accepted tokens/step) for `ttt_steps=3` vs the pretrained drafter, since that's the headline metric for the rollout speedup the perf plot is showing.

_Generated by Claude Code_

seonjinn (commented):
## Review

Read this branch merged with current `main` (clean merge, 59 files). @zpqiu has already covered the correctness of the joint-LSE trunk+branch attention and the CI state, so this review does not revisit either — everything below is either a config-plumbing defect or a collision with the drafter stack that is landing in parallel.

One framing note before the findings: CI on this PR runs only DCO, the submodule comment, semantic-pull-request and triage. There is no unit-test lane, so the seven failures reported from a local run will not surface here no matter how many times it is re-run, and none of the defects below are catchable by the checks currently on the PR.

### Actionable

**1. Multi-pass normalization is silently wrong on the split train-step path.** `megatron_policy_worker.py:1580` threads `draft_ttt_steps=draft_ttt_steps` into the split step but not `global_draft_pass_counts`, while the monolithic path passes both (`:978` and `:983`). With `ttt_steps > 1` on the split path the loss therefore receives `global_draft_pass_counts=None` and falls back to `denominator = global_valid_toks` (`loss_functions.py:218-219`) — the policy's token count, not the intended `sum_d alpha_d * count_d`. The draft gradient is scaled by the wrong denominator, by roughly the ratio of the two counts, with no error and no metric that would look obviously wrong.

Either compute the counts on the split path as well, or reject the combination outright. I would reject for now: `raise NotImplementedError` when `ttt_steps > 1` and the split path is selected. That also becomes mandatory once #3706 lands — see item 5 of the follow-up comment.

**2. `ttt_steps: 0` is silently coerced to 1.** `int(draft_cfg.get("ttt_steps", 1) or 1)` (`train.py:659`, and the same expression at `megatron_policy_worker.py:872` and `:1547`) turns `0` into `1` because `0` is falsy, while `-1` passes straight through to whatever fails first. A user who writes `ttt_steps: 0` meaning "off" gets a single-pass run and no message. This wants one validated field rather than three copies of a defensive coercion — see item 2 of the follow-up comment.

**3. The weight-decay-only draft optimizer config is rejected.** `DraftOptimizerConfigOverrideProvider.build_config_overrides` (`draft/utils.py:1328-1334`) raises `ValueError: [draft] draft (lr, min_lr) must differ from the policy's.` whenever the effective `(max_lr, min_lr)` equals the policy's. Setting only `draft.weight_decay` leaves both LRs unset, so `effective_max_lr`/`effective_min_lr` fall back to the policy's values and the raise fires — even though the requested override (`start_wd`/`end_wd`) is perfectly distinct and would work. Giving the drafter a different weight decay at the same LR is a reasonable thing to want and is currently unreachable.

The raise exists because the group's resume identity was assumed to be just `(max_lr, min_lr)`. It is not, any more — see item 1 of the follow-up comment, which is also why I would delete this code path rather than patch the condition.

**4. `ttt_pass_weights` length is validated per microbatch, not at startup.** The check lives in `DraftTTTCrossEntropyLossFn.__call__` (`loss_functions.py:191-195`), so a mismatched list is caught on the first forward of the first microbatch rather than at config parse. The length is fully known at config time. Same for the `global_draft_pass_counts` shape check right below it, which is derived from the same `ttt_steps`.

**5. The forward-side pass count and the loss-side pass count come from different places.** The forward selects `forward_ttt` from the `draft_ttt_steps` function parameter, which defaults to `1` (`train.py:233`, used at `:369`), while `DraftLossWrapper` is constructed with `ttt_steps` read from `self.cfg["draft"]` (`train.py:659`, `:683`). Both call sites in this PR pass the kwarg, so the two agree today. But the default makes the divergence silent-by-construction for the next caller: omit the kwarg with `ttt_steps: 3` in the config and the forward produces single-pass output while the loss asks for `student_logits_by_pass`, i.e. a `KeyError` some distance from the cause. Deriving both from one config object removes the trap.

**6. `raise ValueError(f"Unknown draft_method ...")` is unreachable.** `method` is read as `draft_cfg.get("method", "eagle3")` (`train.py:658`) but is declared in no schema and set in no YAML in the repo, so the only reachable value is the default. This is a symptom of the selector duplication rather than a bug on its own — see item 4 of the follow-up comment.

### Nit

`self.seq_chunk_size` defaults to `4096` in `DraftTTTCrossEntropyLossFn.__init__` (`loss_functions.py:155`), but `train.py:668-671` only forwards `loss_seq_chunk_size` when `ttt_steps > 1`. In the single-pass case the key is accepted by the config and then ignored. Harmless today because `DraftCrossEntropyLossFn` does not chunk, but it reads as a knob that works everywhere.

### Verified clean

The guard placement is right: `ttt_steps > 1` with `sequence_packing` is rejected early at `lm_policy.py:176-188`, where the user sees it at setup, and CP / sequence-parallel for `ttt_steps > 1` are rejected in `build_draft_model` (`draft/utils.py:1471-1481`). The per-pass metric denominators are correctly the *global* per-pass counts rather than per-microbatch means (`loss_functions.py:247-259`) — the comment there explaining the ~512x inflation at gbs 512 is worth keeping.

`ChunkedDistributedCrossEntropy`'s memory argument holds: saving the bf16 logits references and recomputing the chunk softmaxes in backward is the right trade when several TTT passes stack before a shared backward, and it mirrors `ChunkedDistributedLogprob`. My only objection to it is that #3703 already ships the same kernel — see the follow-up comment.

### Method note

Nothing except lint was executed: `torch` is not importable in the review environment, so all of the above is read from source on the merged tree. `ruff 0.9.9 check` and `format --check` are both clean on the merged tree (640 files formatted), so the reported CI failures are test failures, not lint. The submodule pin was resolved through `3rdparty/Megatron-Bridge-workspace/Megatron-Bridge` at `b11414c71b15e54d333eb49346ed199f20fa9021`.

seonjinn (commented):
(Duplicate submission — same content as the review directly above. Sorry for the noise.)

seonjinn:
## Integration suggestions for the DFlash/DSpark drafter stack

Thanks again for this work. Would it be okay if I leave a few integration-focused suggestions here?

I am happy for your team to continue owning the EAGLE-3 TTT implementation, and I do not intend to propose changes to the core algorithm — the TTT loss, the joint-LSE trunk+branch attention, and the multi-pass schedule are yours. What I would like to align on is the surrounding plumbing: the typed draft config, the loss-statistics interface, and the optimizer/refit structures, so that this integrates cleanly with the DFlash/DSpark online-training PRs landing alongside it (`#3701 → #3703 → #3706 → #3707 → #3715 → #3723 → #3711 → #3726 → #3744 → #3750 → #3757`).

Concretely, the four conventions I would suggest converging on:

- **a typed draft config with a shared selector**, `speculator_type: eagle3` rather than a separate `method` key
- **EAGLE-specific fields declared on `Eagle3DraftConfig`** (`ttt_steps`, `ttt_pass_weights`, `loss_seq_chunk_size`), so they are validated once at parse time
- **a common draft loss-statistics interface**, `DraftLossStats` with one bin per pass, which the TTT reduction maps onto exactly
- **consistent optimizer and refit structures**, `draft.optimizer.*` and the shared `_load_draft_weights` contract

**On timing, since none of the stack has landed yet.** I am not asking you to depend on unmerged branches, and this PR should not be held waiting on them. The point of raising it now is that both sides can still agree cheaply on a shared shape, rather than discovering it as a conflict at merge time. The ordinary rule applies: whichever lands first stays, the second rebases. The items below are written for the case where the stack lands first, since that is the current plan of record — if it slips and this PR lands first, the same items apply in reverse and **I will carry them on the stack side instead**. Either way, most of the mechanical work below is work I am happy to do rather than hand to you; tell me which parts you would rather own.

In practice, only item 4 is worth doing before the stack merges. The rest is agreement-in-principle.

I have written these as concrete diffs rather than a neutral comparison, because a comparison leaves the decision unmade — not because the choice is settled. If you disagree on any of them, that is exactly the discussion worth having before either side merges.

Everything cited was read from this branch merged with current `main`, and from the stack branches.

---

### 1. Optimizer overrides: converge on #3707's structure

Both PRs add a function with the same name, wired into the same call:

- here: `build_draft_optimizer_override_provider` at `nemo_rl/models/megatron/draft/utils.py:1339`, wired at `nemo_rl/models/megatron/setup.py:2063`
- #3707: same name at `nemo_rl/models/megatron/draft/optimizer.py:82`, wired at `setup.py:2029`

Neither is in `main` — `git show origin/main:nemo_rl/models/megatron/draft/utils.py` has no such symbol, and neither does `main`'s `setup.py`. So this is a genuine duplicate, and the second one to merge gets a conflict in both files.

**Suggestion: use #3707's provider instead of the one here.** That means removing, from `draft/utils.py`:

- `_PARAM_GROUP_IDENTIFIER_EXTRA_KEYS` (`:1263`)
- `extend_param_group_identifier_keys_for_resume` (`:1266`)
- `DraftOptimizerConfigOverrideProvider` (`:1294-1336`)
- `build_draft_optimizer_override_provider` (`:1339-1360`)

and the wiring at `setup.py:2063` plus the import at `setup.py:245`. Then delete `tests/unit/models/megatron/test_draft_optimizer_overrides.py`, which is the file already reported as carrying three failures.

Three reasons the stack's version is the one to keep, beyond merge order:

- **The monkeypatch is already dead.** As noted separately above, `param_group_identifier_keys` is now built from `ParamGroupOverride.__annotations__` upstream and already carries `max_lr`/`min_lr`/`start_wd`/`end_wd`. `extend_param_group_identifier_keys_for_resume` is a no-op against the pinned submodule. #3707 never needed it.
- **The distinctness raise goes with it.** The `ValueError` at `draft/utils.py:1331` requiring `(lr, min_lr) != (policy lr, policy min_lr)` only exists because the identifier tuple used to be too narrow. It is what rejects the weight-decay-only config flagged in the separate comment above. #3707 has no equivalent and does not need one, so deleting the patch deletes the bug.
- **The selector is more robust.** This PR matches parameters by module path, `ParamKey(name="*draft_model.*")` (`:1335`). `main` already tags every draft parameter — `DRAFT_GRAD_NORM_GROUP = "draft"` at `draft/utils.py:1239`, stamped at `:1531`, registered as a separate grad-norm group at `:1254` — and #3707's `_is_draft_parameter` reads that tag. To be clear, the glob is not broken today: the leading `*` absorbs the `module.` prefixes that DDP and `Float16Module` add, and the draft is attached as exactly that attribute at `setup.py:1671`. The problem is that it is a second, independent answer to a question `main` already answers, and the two drift silently — rename the attachment attribute, or reuse the provider for the DFlash/DSpark bodies that #3715/#3723/#3711 add, and the glob selects nothing while the grad-norm grouping keeps working.

**One thing to carry over rather than drop.** The two disagree on what `min_lr` defaults to, and this is a real behavioural difference, not a code-organization one:

- here (`draft/utils.py:1314-1320`): with `lr` set and `min_lr` unset, `min_lr` follows the draft LR, so the drafter holds a constant LR while the policy decays. The comment says that is deliberate.
- #3707 (`draft/optimizer.py:52-54`): `min_lr` falls back to the *policy's* `min_lr`, so the drafter decays on the policy's floor.

If the constant-LR default is what your AIME24 runs used, that is worth saying on #3707 before it merges, because rebasing onto #3707 silently changes it. Both behaviours stay reachable either way by setting `min_lr` explicitly — it is only the default that differs.

### 2. EAGLE fields: declare them on `Eagle3DraftConfig`

This PR adds `ttt_steps`, `ttt_pass_weights` and `loss_seq_chunk_size` to the `DraftConfig` `TypedDict` (`nemo_rl/models/policy/__init__.py:515-523`). #3701 deletes that TypedDict outright and replaces it with `PolicyConfig.draft: NotRequired[Eagle3DraftConfig]` (`policy/__init__.py:568`), a pydantic model in `nemo_rl/models/policy/draft_config.py`.

**Suggestion: declare the three keys as fields on `Eagle3DraftConfig`.** Roughly:

```python
    ttt_steps: Annotated[int, Field(ge=1)] = 1
    ttt_pass_weights: list[float] | None = None
    loss_seq_chunk_size: Annotated[int, Field(gt=0)] | None = None

    @model_validator(mode="after")
    def validate_ttt(self) -> Self:
        if self.ttt_pass_weights is not None and len(self.ttt_pass_weights) != self.ttt_steps:
            raise ValueError(
                f"ttt_pass_weights must have ttt_steps={self.ttt_steps} entries, "
                f"got {len(self.ttt_pass_weights)}"
            )
        return self
```

This is worth doing for its own sake, not just for the merge. It fixes three things reported separately above in one place:

- `Field(ge=1)` replaces `int(draft_cfg.get("ttt_steps", 1) or 1)`, which silently turns `ttt_steps: 0` into `1` while `-1` raises. Both then route through one validator.
- the `ttt_pass_weights` length check currently runs per-microbatch inside `DraftTTTCrossEntropyLossFn.__call__`, though the length is known at config time; as a validator it fails at startup.
- pydantic defaults remove the five hidden `ttt_steps` defaults at `train.py:659`, `draft/utils.py:1471`, `lm_policy.py:178`, `megatron_policy_worker.py:871` and `:1548`, which is what lets a caller omit the forward-side `draft_ttt_steps` and produce the `KeyError: 'student_logits_by_pass'` reported above.

If the flat `lr`/`min_lr`/`weight_decay` keys at `policy/__init__.py:527-529` survive item 1, note that `Eagle3DraftConfig` is declared `extra="allow"` (`draft_config.py:36`) with no unknown-key check, so those three would be accepted and silently ignored rather than rejected — a user asking for a 50x draft LR would get 1x with no error and no failing test. Deleting them per item 1 is what closes that.

### 3. Draft-config reads: subscript to attribute access

Once `draft` is an `Eagle3DraftConfig`, `cfg["draft"]["x"]` raises `TypeError: 'Eagle3DraftConfig' object is not subscriptable` and `.get(...)` raises `AttributeError`. This PR adds new subscript sites on top of the ones `main` already has:

- `nemo_rl/models/megatron/train.py:657-668` (`draft_cfg = self.cfg["draft"]`, then `.get("method")`, `.get("ttt_steps")`, `.get("ttt_pass_weights")`, `.get("loss_seq_chunk_size")`, `draft_cfg["loss_weight"]`)
- `nemo_rl/models/policy/workers/megatron_policy_worker.py:871-873` and `:1547-1548`
- `nemo_rl/models/policy/lm_policy.py:178`
- `nemo_rl/models/megatron/draft/utils.py:1471`

#3701 provides `coerce_draft_config` (`draft_config.py:48`) for exactly the boundary case where a caller hands in a plain mapping, so these become `draft_cfg.ttt_steps` after a single coercion at the entry point rather than defensive `.get()` at each site.

### 4. Selector: one key, `speculator_type`

This PR adds a second drafter selector: `draft_method = draft_cfg.get("method", "eagle3")` (`train.py:658`), threaded into `DraftLossWrapper` as `draft_method=` (`:682`). #3701 already declares the selector as `speculator_type: Literal["eagle3"] = "eagle3"` (`draft_config.py:39`) and dispatches through `_SPECULATOR_FACTORIES` (`nemo_rl/models/megatron/draft/training.py:67-81`), and #3715/#3723/#3711/#3726 extend that same `Literal` for DFlash and DSpark.

`method` is declared in no schema and set in no YAML, which is also why the `raise ValueError(f"Unknown draft_method ...")` below it is unreachable today.

That unreachability stops being harmless once the stack lands. #3757 adds `examples/configs/recipes/llm/grpo-qwen3-8b-1n8g-megatron-dflash.yaml`, where `policy.draft.speculator_type: dflash`. `method` is absent from that file, so `draft_cfg.get("method", "eagle3")` returns `"eagle3"` and this PR builds an EAGLE-3 loss for a DFlash draft — two selectors disagreeing, with the wrong one winning silently. Worth noting separately that `speculator_type` is also the name of a vLLM field (`vllm_config.speculative_config.method`) used in #3744/#3750/#3757; those are different namespaces, and #3757's guide states the vLLM `method` must match `policy.draft.speculator_type`. One NeMo-RL-side selector is the only way that stays legible.

**Suggested change: delete the `method` read and the `draft_method` parameter.** Multi-pass TTT is an EAGLE-3 feature, and `speculator_type` already says the speculator is eagle3, so `ttt_steps > 1` is sufficient to select the multi-pass loss — which is what `DraftLossWrapper.multi_pass` (`wrapper.py:290`) already keys on. That leaves one selector for "which drafter" and one knob for "how many passes", and the multi-pass loss registers through the same factory the DFlash/DSpark work will use.

### 5. #3706 makes the split-path guard mandatory, and points at the eventual fix

Separately noted above that the split train-step API threads `draft_ttt_steps` but not `global_draft_pass_counts` (`megatron_policy_worker.py:1546`), and that a `NotImplementedError` is the right response for now. #3706 turns that from a good idea into a requirement: it routes split-microbatch draft statistics through a new `DraftStepState` that hard-rejects anything other than a single bin with unit weight (`nemo_rl/models/megatron/draft/step_state.py:73-79`). Multi-pass statistics are `[ttt_steps]` bins with per-pass weights `alpha_d`, so after #3706 a `ttt_steps > 1` run on the split path fails with `ValueError: current EAGLE draft step state requires exactly one bin` instead of mis-scaling. Adding the guard here keeps that a clear message at setup rather than a confusing one mid-step.

Worth flagging the constructive half, since it affects how #3706 should be reviewed: the multi-pass shape is exactly what `DraftLossStats`'s `numerators`/`counts`/`weights` triple was built for, with `weights = alpha`. The one-bin restriction in `accumulate` is a deliberate narrowing, not a structural limit. If multi-pass split-step support is wanted later, generalizing that guard to `[ttt_steps]` bins is the natural path and the denominator `sum_d alpha_d * count_d` falls out of `_normalization_scale` as written. Saying so on #3706 now would stop the narrowing from hardening into an assumption.

### 6. Loss statistics: converge on `DraftLossStats`

This is the third duplicate, and the one where the two implementations are closest.

- here: `ChunkedDistributedCrossEntropy` at `nemo_rl/distributed/model_utils.py:270` — "Soft-target cross entropy across TP-sharded vocab, chunked along sequence", saving only the bf16 logits references and recomputing the chunk softmaxes in backward. Called from `DraftTTTCrossEntropyLossFn._per_token_soft_ce` (`nemo_rl/algorithms/loss/loss_functions.py:165-172`) with the `loss_seq_chunk_size` knob.
- #3703: `streaming_vocab_parallel_soft_ce` at `nemo_rl/algorithms/loss/draft.py:324`, dispatching to `_StreamingVocabParallelSoftCE` (`:213`) with the same recompute-in-backward structure, plus a `_CachedVocabParallelSoftCE` fast path (`:149`) when the whole input fits in one tile, and the `token_chunk_size` knob.

Same math, same memory strategy, two copies. #3703's is the superset: it takes the mask, returns per-bin reductions, and tiles over flattened tokens rather than the sequence axis, so it does not degrade when the batch dimension carries the tokens.

**Suggested change: drop `ChunkedDistributedCrossEntropy` and have the per-pass reduction return a `DraftLossStats` with one bin per TTT pass.** The two reductions already line up term for term. In the loop at `loss_functions.py:223-262`, `pass_sum` is the mask-weighted CE sum for pass `d`, `pass_counts[pass_index]` is its global token count, `weight` is `alpha_d`, and the final scalar is

```
total / denominator  =  sum_d alpha_d * pass_sum_d  /  sum_d alpha_d * pass_count_d
```

which is exactly `DraftLossStats.normalized(normalization_counts=global_counts)` — that method computes `(numerators * weights).sum() / ((counts * weights).sum() + 1e-8)` (`draft.py:42-58`). So the per-pass call becomes

```python
stats_d = streaming_vocab_parallel_soft_ce(
    student_logits=student_slice,
    teacher_logits=teacher_slice,
    mask=pass_mask,
    token_chunk_size=self.seq_chunk_size,
    tp_group=self.vocab_parallel_group,
    weights=student_slice.new_tensor([weight]),
)
```

and the per-pass stats stack into a single `[ttt_steps]`-bin `DraftLossStats`. `metrics[f"draft_loss_pass_{d}"]` reads off `numerators[d] / counts[d]`, which is what the current code recomputes by hand.

Three things this buys beyond removing the duplicate:

- It is the concrete form of item 5. The stacked stats are the `[ttt_steps]`-bin object `DraftStepState` would need, so multi-pass split-step support later becomes generalizing #3706's one-bin guard rather than a new data path.
- `nemo_rl/distributed/model_utils.py` stops growing a draft-specific loss kernel. #3703 puts it in `nemo_rl/algorithms/loss/draft.py`, next to the thing that consumes it.
- The `torch.clamp(denominator, min=1.0)` at `loss_functions.py:262` becomes `normalized`'s `+ 1e-8`. Worth a deliberate decision rather than an accident of which file won: clamping at 1.0 silently rescales any step whose global draft token count is below 1, which is only the empty-batch case, but the two conventions disagree there and only one survives the merge.

One thing to check rather than assume: #3703's kernel expects `mask` to match the non-vocab dims of the logits and folds it into the numerator itself, while the current code multiplies the mask in afterwards at `:244`. `pass_mask` is already `draft_pass_token_mask(token_mask, ttt_pass) * sample_mask.unsqueeze(-1)`, i.e. the right shape and semantics, so this should be a straight substitution — but the pass-1 shifted-slot behaviour is worth one parity test against the current implementation before deleting anything.

### 7. The rest of the stack, and where in the order this PR belongs

I also read the seven later PRs (#3715, #3723, #3711, #3726, #3744, #3750, #3757) against this branch. Two have zero file overlap with it (#3723, #3711) and #3726 reaches it only through a hunk #3715 already owns, so the surface is smaller than the PR count suggests. The three that matter are #3744/#3750 in `vllm_backend.py`, and #3757, which touches nine of the same files.

**Merge-order recommendation: land this PR after #3757, i.e. twelfth.** It conflicts materially with only #3701 and #3757, and #3757 is already last — so putting this one after it means rebasing ~700 production lines onto a settled tree, rather than forcing #3757's 121-file diff to absorb this one across nine files.

**The thing most at risk of being silently lost: `_unshare_draft_lm_head`.** Defined at `vllm_backend.py:602-644`, called at `:666`, gated on the weight stream containing an `lm_head`. #3744 rewrites `_load_draft_weights` wholesale and #3750/#3757 carry that same rewrite; none of the three has any counterpart, and all three delete the lines the call sits between. A routine "take theirs" conflict resolution therefore drops it and leaves the function as dead code with no caller. Per the docstring here, vLLM's `SpecDecodeBaseProposer._maybe_share_lm_head` aliases the drafter's `lm_head` to the target's, so refitting a *trained* EAGLE head writes through the alias into the serving target. This is the only place in the twelve PRs that addresses that aliasing, and it is reachable on exactly the configuration this PR targets — the DFlash/DSpark export blocklists strip `lm_head`, so the gate never fires for them, but EAGLE-3 co-training is precisely where it does. Please re-apply the call inside #3744's rewritten body when rebasing. (I have not read vLLM's own source for the aliasing; that half is from the docstring here.)

**Three smaller carry-overs in the same area:**

- **Drop the warn-to-raise at `vllm_backend.py:660` in favour of #3744's.** #3744 reaches the same conclusion at the same site with a typed `SpeculatorRuntimeError` plus PP-ownership awareness. That matters after #3750: an unconditional raise here fires on non-owning pipeline ranks, where #3744's version returns early instead.
- **The packing guard becomes redundant; the CP/SP guards do not.** #3757 replaces the blanket packing check with a capability test against `Eagle3DraftConfig.supports_sequence_packing = False`, which subsumes `lm_policy.py:176-188` — drop this PR's version, it is a textual conflict on the same `if`. The CP and sequence-parallel guards at `draft/utils.py:1477` and `:1481` look equally redundant against the sibling ClassVars, but #3757 enforces those centrally in `_validate_draft_training_setup`, which early-returns unless the speculator is `dflash`/`dspark`. No other site rejecting eagle3 + CP turned up, so **keep those two** rather than deleting them as redundant — worth confirming against the full body of that validator first.
- **Re-apply the `rope_scaling` null handling** at `draft/utils.py:1434-1443` (the "official z-lab configs carry an explicit `rope_scaling: null`" fix). #3757 rewrites neighbouring lines but not those, so it is a small isolated re-apply that is easy to lose in a large rebase.

**One new contract hole, from #3757.** It adds `defer_normalization` to `DraftLossWrapper` and on that path calls `self.draft_loss_fn.loss_stats(...)`, adding `loss_stats` to `DraftCrossEntropyLossFn`. `DraftTTTCrossEntropyLossFn` has only `__call__`, so multi-pass would `AttributeError` there. It looks unreachable on any shipped config today — the deferral comes from the split-step API, which #3757 demands only for `context_parallel_size > 1`, and eagle3 + CP is rejected — but item 6 is what closes it properly: once the per-pass reduction returns a `DraftLossStats`, `loss_stats` is that same object and the method is nearly free to add.

### 8. Mechanical

- `setup.py`: both PRs add the same kwarg to the same `setup_optimizer` call. Textual conflict, trivial — but resolve it after items 1 and 2, not before, or it papers over them.
- Tests: `test_draft_optimizer_overrides.py` (here) and `test_draft_optimizer.py` (#3707) cover the same function under different names. Item 1 deletes the former.
- Module home: `draft/utils.py` is already ~1500 lines; #3707 puts this feature in `draft/optimizer.py`. Item 1 makes that moot.

---

To restate the intent: none of this touches the TTT algorithm, and none of it is a blocker from my side. Items 1, 2, 3, 6 and 8 are mechanical, and I am glad to prepare them as a follow-up on top of whichever branch merges second — say the word and I will open it rather than adding to your queue. Items 4, 5 and 7 are the ones where I would value your judgement, particularly the `min_lr` default in item 1 (if your AIME24 runs relied on the constant draft LR, I would rather change #3707 than change this) and the `_unshare_draft_lm_head` carry-over in item 7, which I think is the single most valuable thing in this PR and the easiest to lose in a rebase.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1507,6 +1539,12 @@ def _train_microbatch_body(
...
+        )
```
seonjinn, line 1547: Finding 1: this path threads `draft_ttt_steps` into `megatron_forward_backward` (`:1578`) but never computes `global_draft_pass_counts`, unlike the monolithic path (`:881`, passed at `:976`). The draft loss then falls back to `denominator = global_valid_toks` — the policy token count rather than `sum_d alpha_d * count_d` — so the draft gradient is mis-scaled with no error and no metric that looks wrong.

Failing fast is the smaller change; computing the counts here would also be correct if you prefer that.

```suggestion
        draft_ttt_steps = (
            int(self.cfg["draft"].get("ttt_steps", 1) or 1) if draft_enabled else 1
        )
        if draft_ttt_steps > 1:
            raise NotImplementedError(
                "policy.draft.ttt_steps > 1 is not supported on the split train-step "
                "path: global_draft_pass_counts is not computed here, so the draft "
                "loss would normalize by the policy token count instead of "
                "sum_d alpha_d * count_d."
            )
```

nemo_rl/models/megatron/train.py
```diff
@@ -598,15 +653,34 @@ def __call__(
...
                     data_dict=data_dict,
-                    loss_weight=float(self.cfg["draft"]["loss_weight"]),
+                    loss_weight=float(draft_cfg["loss_weight"]),
+                    draft_loss_kwargs=draft_loss_kwargs,
+                    global_draft_pass_counts=global_draft_pass_counts,
                     vocab_parallel_rank=get_tensor_model_parallel_rank(),
                     vocab_parallel_group=get_tensor_model_parallel_group(),
                     context_parallel_group=get_context_parallel_group(),
+                    draft_method=draft_method,
+                    ttt_steps=ttt_steps,
```
seonjinn, line 683: Two things folded into one suggestion, both from the earlier comments:

- **`ttt_steps: 0` is silently coerced to 1.** `int(... or 1)` turns `0` into `1` because `0` is falsy, while `-1` passes through to whatever fails first. Explicit validation instead.
- **The `method` selector.** `draft_cfg.get("method", "eagle3")` is declared in no schema and set in no YAML, so `"eagle3"` is the only reachable value and the `Unknown draft_method` raise in `DraftLossWrapper` is dead. `draft_method` already defaults to `"eagle3"` in the wrapper signature (`wrapper.py:268`), so dropping the read here is behaviour-preserving today — and it avoids a second selector disagreeing with `speculator_type` once the DFlash/DSpark configs land.

Happy to drop the second half if you would rather keep `method` for now; the first half stands on its own.

```suggestion
                draft_cfg = self.cfg["draft"]
                raw_ttt_steps = draft_cfg.get("ttt_steps", 1)
                ttt_steps = 1 if raw_ttt_steps is None else int(raw_ttt_steps)
                if ttt_steps < 1:
                    raise ValueError(
                        f"policy.draft.ttt_steps must be >= 1, got {ttt_steps}"
                    )
                # Ctor kwargs for the draft LossFn (uniform shape; DraftLossWrapper
                # splats them into the selected class). Only the multi-pass class
                # takes pass weights.
                draft_loss_kwargs: dict[str, Any] = (
                    {"pass_weights": draft_cfg.get("ttt_pass_weights")}
                    if ttt_steps > 1
                    else {}
                )
                if ttt_steps > 1 and draft_cfg.get("loss_seq_chunk_size") is not None:
                    draft_loss_kwargs["seq_chunk_size"] = int(
                        draft_cfg["loss_seq_chunk_size"]
                    )
                loss_fn_wrapped = DraftLossWrapper(
                    loss_fn=loss_fn_wrapped,
                    prepare_fn=prepare_loss_input_wrapped,
                    data_dict=data_dict,
                    loss_weight=float(draft_cfg["loss_weight"]),
                    draft_loss_kwargs=draft_loss_kwargs,
                    global_draft_pass_counts=global_draft_pass_counts,
                    vocab_parallel_rank=get_tensor_model_parallel_rank(),
                    vocab_parallel_group=get_tensor_model_parallel_group(),
                    context_parallel_group=get_context_parallel_group(),
                    ttt_steps=ttt_steps,
                )
```

seonjinn (commented):
Two of the items from my earlier comments as applicable suggestions, so they are easier to take or reject on the spot. Both are standalone — neither depends on the drafter stack landing.

MarkXCloud:
@zpqiu Thanks for the review — all seven test failures and the pyrefly error are fixed in the updated commit; replied on each comment in-thread.

Test evidence (container run against the pinned submodules, `--maxfail=100` to get past `-x`): default pass 32 passed / 1 skipped (the vllm file's importorskip in the mcore venv); `--mcore-only` 60 passed (incl. the guard test and the modelopt mask oracle); `--vllm-only` 6 passed; pyrefly clean on `vllm_backend.py`.

Acceptance-rate data (`ttt_steps=3` vs the pretrained drafter) is being collected and will be posted here.

Could a maintainer post `/ok to test 5ab691cfa05bfd3d54120f117b1315412591521b` and attach a `CI:L1` label? I don't have label permissions on this repo.

tests/unit/models/megatron/test_draft_optimizer_overrides.py
MarkXCloud: You're right — upstream #4705 covers it. Deleted the helper, the extra-keys tuple, and the call site. Tests reworked into an upstream-premise pin (`{max_lr, min_lr, start_wd, end_wd} ⊆ param_group_identifier_keys` in all three modules) plus a behavioural round-trip through `_filter_and_reorder_param_groups` with the saved order reversed; both stale docstring premises dropped.

nemo_rl/models/megatron/draft/utils.py
MarkXCloud: Took your guard verbatim; the wd-only config is now accepted and covered by a test.

nemo_rl/models/generation/vllm/vllm_backend.py
MarkXCloud: Took both halves — `speculator` fallback plus early return on non-last PP ranks — and all three behaviors are now unit-tested. On V1 vs V2: I'll confirm from the engine logs on the next run; the AIME24 runs had a live drafter (nonzero, moving acceptance), which per your trace is consistent with V1.

nemo_rl/models/generation/vllm/vllm_backend.py
MarkXCloud: Added the `isinstance` narrow (the file is pyrefly-clean now). For the latent attrs issue I kept deepcopy but re-bind the source params' `__dict__` onto the copies, with a test pinning `weight_loader` survival — reattach avoids reproducing `ParallelLMHead`'s ctor args.

nemo_rl/models/megatron/draft/eagle.py
MarkXCloud: Added both raises; `test_module_guards_do_not_require_gpu` passes.

nemo_rl/algorithms/loss/utils.py
MarkXCloud: Deleted the test (agreed a fallback would be worse), and adopted your consistency check — the loss post-processor now raises when `student_logits_by_pass` and `ttt_steps` disagree.

nemo_rl/models/policy/workers/megatron_policy_worker.py
MarkXCloud: Adopted the `NotImplementedError` guard, with the denominator argument in the comment.

nemo_rl/models/megatron/train.py
MarkXCloud: Fixed at the root — new `resolve_draft_aux_layer_ids()` resolves the list rank-independently and is threaded into both forward paths, and `build_draft_model` uses the same resolver so the draft and the capture can't disagree. The nested `getattr` is a direct read now.

nemo_rl/models/policy/__init__.py
MarkXCloud: All six keys are in the two exemplar YAMLs and the reference config (`test_config_v2` passes); `loss_seq_chunk_size: 4096` is the exemplar default; `method` is deleted; `or 1` is gone everywhere with `ttt_steps >= 1` validated once at Policy setup.

tests/unit/models/generation/test_vllm_draft_lm_head_unshare.py
MarkXCloud: Took the fix (marker + importorskip), added the `weight_loader` assertion, and added the `_load_draft_weights` raise-path tests you suggested.

tests/unit/models/megatron/test_train.py
MarkXCloud: Fixed as suggested.

nemo_rl/algorithms/loss/wrapper.py
MarkXCloud: Imports hoisted (wrapper.py and the stdlib `copy`); `ttt_pass_weights` length and the CP/SP guards moved to Policy setup; pass-1 masking now carries a "deliberate" comment; the flash-attn version gate moved from import time into `TTTDraftCoreAttention.__init__`. Deferred: unifying the sibling return types — better done in the `DraftLossStats` convergence discussed with seonjinn.

PR #3701, opened 2026-08-18
refactor(draft): add typed speculator contract

seonjinn:
## Summary

- Introduces typed `policy.draft` configuration and a speculator registry.
- Preserves disabled and existing EAGLE-3 behavior while providing an extension point for additional draft models.
- Serves as the entry point for the DFlash/DSpark online drafter training PR series described below.

## Scope

Configuration and model-construction contracts only. DFlash/DSpark objectives, training integration, distributed weight transport, and runtime refit are handled by the follow-up PRs.

## PR series overview

This is a logical API and validation dependency map. It is not necessarily the branch ancestry or required merge order. Each follow-up PR lists its direct dependencies in its own PR body.

| Series | PR | Purpose | Direct logical dependencies |
|---|---|---|---|
| PR1 | #3701 | Typed speculator configuration and registry contract | None |
| PR2 | #3703 | Streaming vocab-parallel soft cross entropy | None |
| PR3 | #3706 | Split-step normalization parity | #3703 |
| PR4 | #3707 | Optimizer and scheduler overrides for draft updates | #3701 |
| PR5 | #3711 | DSpark auxiliary heads and checkpoint provenance | None |
| PR6 | #3715 | Projected soft cross entropy for DFlash | #3703 |
| PR7 | #3723 | Tensor-parallel DFlash body | #3715 |
| PR8a | #3725 | Bounded multi-pass EAGLE test-time training | #3703, #3706 |
| PR8b | #3726 | DSpark tiled objectives | #3703, #3711 |
| PR9 | #3744 | Fail-closed speculative draft refit lifecycle | None; integration point for trained drafters |
| PR10 | #3750 | Context- and pipeline-parallel draft-weight transport | #3744 |
| PR11 | #3757 | Packed context-parallel DFlash/DSpark co-training | #3744, #3750 |

The main paths are:

- Common training primitives: #3703 → #3706 and #3715.
- DFlash training: #3703 → #3715 → #3723.
- DSpark training: #3703 + #3711 → #3726.
- Distributed online update/refit: #3744 → #3750 → #3757.

## Validation

- Exact head: `8774f1d3ddf62fe60f39853d485486c80b3c0576`.
- Current-head preflight, lint, docs, container-build, and unit-test checks report no failures.

## Prior focused evidence

- OCI-Hsg `6299238`: 15 focused tests passed; Ruff, format, compile, DCO, and targeted Pyrefly passed.

## Remaining

- PR1 has no direct PR dependency.
- Review is required before merge.
- CP transport, packed-sequence co-training, repeated online update/refit, and multi-node validation are tracked by #3750 and #3757 rather than this contract-only PR.


seonjinn:
Self-review completed on exact head `ca736093`.

Resolved:
- Fixed the two Eagle utility typing errors exposed when this PR enrolled the module in Pyrefly.
- Updated the stale registry mock seam found during independent review.

No unresolved correctness, lifecycle, config-compatibility, security, or test findings.

Validation:
- OCI-Hsg job `6299238`: 8 base/config/controller + 7 MCore setup/speculator tests passed.
- Targeted Pyrefly: 0 errors.
- Ruff, formatting, compile, DCO, and GitHub checks passed.

seonjinn:
/review-pr-team

Reviewed exact head `389c077f48e15fe2c501fb63e4dbf939b594f998` against current `main` (`97f391eb`). The synthetic merge is conflict-free.

The review covered the typed `policy.draft.speculator_type` schema, registry and builder dispatch, disabled/no-draft behavior, pre-wrap construction, refit callers, and all seven paths changed by both the PR and current `main`. No remaining issue at or above 80% confidence was found.

OCI-Hsg job `6299238` passed 8 focused config/controller tests, 7 MCore tests, Ruff, format, compile, DCO, and targeted Pyrefly on the functional head. The final commit is import-order-only. The previous Algorithms CI failure was caused by a stale tokenizer test in `main` and is fixed by #3717; it is not a PR-product failure.

No action needed from this self-review. Fresh current-main CI is requested separately.

nemo_rl/models/policy/draft_config.py
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Literal
+
+from pydantic import BaseModel
+
+
+class Eagle3DraftConfig(BaseModel, extra="allow"):
```
seonjinn, line 22: `extra="allow"` plus every field having a default means a typo of a declared key silently no-ops: `Eagle3DraftConfig.model_validate({"enalbed": True})` yields `enabled=False` with no error, where the pre-PR TypedDict path raised loudly. The allow-mode itself is deliberate (the legacy-keys test pins it), and #3757's dflash/dspark config classes go `extra="forbid"`, so the exposure is this class's declared fields only.

Fixed in the follow-up commit with a near-miss validator: extras that closely match a declared field name (difflib, cutoff 0.8) are rejected with a `did you mean 'enabled'?` error, while genuinely novel legacy keys are still preserved (the legacy-keys test stays green; verified empirically for both cases).

seonjinn (commented):

Method: a comprehensive review agent over the diff, then an adversarial pass that re-ran every repro and resolved the judgment calls against downstream branches.

Two candidate findings did not survive adversarial verification and are recorded here so reviewers don't re-derive them: (1) "the registry is bypassed by the provider PRs" -- refuted: #3725/#3726 are based on pre-#3701 main and their scope excludes wiring, while #3757's head demonstrably extends exactly this registry (`_SPECULATOR_FACTORIES` grows to eagle3/dflash/dspark with discriminated configs), so the contract is load-bearing downstream; within this PR alone it is a one-entry dict, which is worth one body sentence. (2) "the CI-refresh claim is false" -- partially: the literal wording ("exact same product tree") is wrong (`c789d498` is the merge-base, ~143 files of main churn apart), but a diff-of-diffs shows this PR's own delta is byte-identical at both SHAs, so the intended claim holds.

**Process, before merge:**

1. Trigger CI on the actual head -- the full L0 suite ran green only on the ancestor `c789d498`; the merge interaction with newer main (including a 334-line upstream change to `megatron_policy_worker.py`) is untested. Needs `/ok to test 70a9673a6b098bc19f181e979ff2d46723c9f6a2` (or on the post-fix head).
2. Reword the body's "exact same product tree" to "this PR's changes are byte-identical to the CI-validated ones".
3. Optional body sentence: the registry ships with a single eagle3 entry by design; #3757 extends it.

Follow-up commit applies the near-miss typo validator (inline comment) plus its tests. Minor nits folded there where cheap; the `draft: null` config variant is exercised by the existing omit-block test.

seonjinn (commented):
## Self-review

Multi-agent review of this branch merged with latest `main` (`8d1ca372f`), including mutation testing over the PR's own suite (35 mutations, 20 caught).

The split is clean and worth stating up front: the pydantic validator is the best-pinned code in the PR — 12 of 13 mutations against `Eagle3DraftConfig` die. Every one of the 15 survivors is at a **boundary**: who calls the validator, and what the caller does with the typed object.

### Actionable

**1. Merge staleness: the packed-sequence branch still subscripts the draft config.** At the PR head there was one draft call site in `nemo_rl/models/megatron/train.py`, correctly converted. `main` has since added a second in the packed-sequence branch, still doing `self.cfg["draft"]["loss_weight"]`, which raises `TypeError: 'Eagle3DraftConfig' object is not subscriptable`. Packing is a supported draft configuration (only packed + PP>1, CP>1 and fused-linear-logprobs are rejected), so packed draft co-training dies at loss-wrapper construction after this merge. Nothing in CI constructs `DraftLossWrapper`, so the suite stays green. Fixed on the branch; a sweep of merge-base vs PR head for every migrated construct found exactly two such sites and no third.

**2. `Policy.__init__` assumes the caller already typed the config.** `PolicyConfig` is a `TypedDict` — a plain dict at runtime — so `config.get("draft")` is not guaranteed to be an `Eagle3DraftConfig`, and every caller building a policy config in-process passes a mapping. `main`'s own guard tests fail against the merged tree with `AttributeError: 'dict' object has no attribute 'enabled'`. Fixed on the branch with a `coerce_draft_config` at the boundary, plus tests pinning the mapping, model, and `None` inputs.

**3. The third entrypoint bypasses the PR's own helper.** `examples/run_grpo.py` and `examples/run_grpo_single_controller.py` were converted to `draft_refit_enabled(...)`, but `examples/nemo_gym/run_grpo_nemo_gym.py` kept `"draft" in config.policy and config.policy["draft"].enabled` — which is also the one form that still raises on a mapping config. Fixed on the branch.

### Suggestions (test gaps, ranked by what a regression would cost)

**`draft_refit_enabled` is never asserted to return `True`.** The only assertion is `draft_refit_enabled(None) is False`, so replacing the entire body with `return False` keeps the suite green. The regression that hides behind this is the expensive one: generation is never told to accept refitted draft weights, the drafter trains, weights are never refit into the generator, and the run produces plausible but stale results with no crash and no failing test. A three-line parametrization over `(None, disabled, enabled)` closes it.

**The draft-disabled path is untested end to end.** Every draft test in the repo, CPU and mcore alike, constructs `enabled=True`. Consequently `build_draft_model`'s disabled guard can be deleted, and `_create_draft_pre_wrap_hook`'s `if draft_speculator is None: return model` neutered, with no failure in either lane. A bug of that shape crashes model setup with `AttributeError: 'NoneType' object has no attribute 'build_model'` for anyone who has a `draft:` block with `enabled: false` — which is the default configuration for every non-draft Megatron user.

**The registry indirection is not exercised.** `_SPECULATOR_FACTORIES` is the headline abstraction here, but hardcoding `Eagle3Speculator(config)` in place of the lookup at `draft/training.py:79-81` changes nothing observable; renaming the key is caught only incidentally via `KeyError`. The tests pin that a dict exists, not that dispatch goes through it. `monkeypatch.setitem` with a fake speculator would pin it, and it becomes load-bearing the moment a second type is added to the `Literal`.

**No test parses a config with the draft actually enabled.** `speculator_type: eagle3` was added to `examples/configs/grpo_math_1B.yaml` and the reference config, both `enabled: false`. The only config in the repo with `enabled: true` is `examples/configs/recipes/llm/grpo-qwen3-1.7b-1n8g-megatron-eagle3.yaml`, and nothing loads it — note it also omits `speculator_type`, relying on the default. Parametrizing the existing `MasterConfig` parse test over every yaml containing a `draft:` block is cheap and kills the largest number of survivors of any single addition.

**`aux_layer_indices` is never asserted to reach the model.** Its default is pinned three ways, but ignoring the field entirely survives — a user sets `[4, 8, 12]`, it validates, it is dropped, and the drafter trains on default layers.

### Nits

The typo detector is one-sided: raising the near-miss cutoff to 0.9 fails a test, lowering it to 0.6 does not. Nothing asserts a legitimately-distinct extra key is *preserved* near the boundary, so the detector can drift aggressive and start rejecting the legacy extensions that `extra="allow"` exists to accept.

`test_eagle3_draft_config_rejects_unknown_speculator_type` uses `pytest.raises(ValidationError, match="eagle3")`, but `"eagle3"` appears in pydantic's canned `Literal` error text regardless of any logic in this PR. Matching on the rejected value, or on `errors()[0]["type"] == "literal_error"`, would assert something.

`test_disabled_resolver_import_does_not_require_megatron_runtime` blocks only `megatron.bridge*`; `megatron.core` is imported unconditionally by the module under test, so the test is weaker than its name.

`tests/unit/models/megatron/test_draft_speculator.py:24` sets `pytestmark = pytest.mark.mcore`, but all three tests are pure Python — two never import megatron and the third patches `build_draft_model`. The mark exiles the PR's only speculator-contract tests to a GPU lane most PRs never run.

### Method note

The suite was executed against a scratch venv with a meta-path stub for `megatron.*`, production code imported from the worktree: 13 passed across the three CPU-runnable files, stable across four orderings. Mutations were applied to a scratch copy; the worktree was never mutated.

seonjinn (commented):

Re-merged this branch with `main` at `4213e0867`. Clean merge, no conflicts. `main` has advanced 109 files (+4118 / -1387) since the base of the previous review (`8d1ca372f`), but none of it reaches a draft code path: `git diff 8d1ca372f 4213e0867` filtered to lines containing `draft` across the six files this PR and `main` both touch (`setup.py`, `policy/__init__.py`, `megatron_policy_worker.py`, `pyrefly.toml`, `test_megatron_generation.py`, `test_megatron_setup.py`) is empty. The `pyrefly.toml` edits land in different sections and merged textually. `ruff 0.9.9 check` and `format --check` are clean on the merged tree (735 files). CI is green, though as before there is no unit-test lane on this PR.

So the merge-staleness class that produced findings 1-3 in the previous review has not recurred. Those findings and their fixes stand as posted.

### One new measurement, on the reach of the typo detector

`_reject_near_miss_extra_keys` (`draft_config.py:32-44`) rejects extras that look like misspellings of a declared field, at `cutoff=0.8`. It is worth stating what that does *not* cover, because the answer bears directly on how the EAGLE-3 work in #3699 will land.

`Eagle3DraftConfig` declares six fields: `speculator_type`, `enabled`, `model_name`, `loss_weight`, `num_layers`, `aux_layer_indices`. Running the detector's own call — `difflib.get_close_matches(key, declared, n=1, cutoff=0.8)` — against the seven flat keys #3699 adds to the draft block returns `[]` for every one:

| key | near match |
|---|---|
| `ttt_steps` | none |
| `ttt_pass_weights` | none |
| `loss_seq_chunk_size` | none |
| `lr` | none |
| `min_lr` | none |
| `weight_decay` | none |
| `method` | none |

None of them is a typo of anything declared, so `extra="allow"` absorbs all seven. A config written against #3699's flat schema therefore validates cleanly under this PR and silently loses every one of those settings — a draft LR override becomes 1x the policy LR with no error and no failing test.

That is the correct behaviour for the detector as specified; it is a typo guard, not an unknown-key guard. The conclusion is just that the guard cannot be the thing that catches schema drift between the two branches, and the EAGLE fields need to be declared as real fields on `Eagle3DraftConfig` rather than left to `extra="allow"`. I have raised that on #3699 as an integration suggestion; noting it here so the reasoning is visible from this side too.

### Also checked

`main` added `_resolve_optimizer_dtype_kwargs` (`setup.py:263-295`), which resolves optimizer dtype strings before `_create_megatron_config` builds `optimizer_kwargs`. It sits close to this PR's wiring but does not interact with it — the draft path contributes `ParamGroupOverride` entries, which carry only `max_lr`/`min_lr`/`start_wd`/`end_wd`/`wd_mult`/`optimizer` and no dtype keys.

### Method note

Nothing executed beyond lint and the `difflib` call above; `torch` is not importable in this environment, so everything else is read from the merged tree.

seonjinn:
/ok to test [d36eb89](https://github.com/NVIDIA-NeMo/RL/pull/3701/commits/d36eb898b7da1bac3d7a21dcadbd68697637b8ef)

nemo_rl/models/megatron/draft/training.py
```diff
@@ -0,0 +1,81 @@
...
+            policy_model_chunk=policy_model_chunk,
+        )
+
+
+_SPECULATOR_FACTORIES: dict[str, type[Eagle3Speculator]] = {
+    "eagle3": Eagle3Speculator,
+}
+
+
+def resolve_draft_speculator(
```
yuekaizhang, line 76: `resolve_draft_speculator` takes `Eagle3DraftConfig | None` and reads `.enabled`/`.speculator_type` directly, unlike `draft_refit_enabled` which calls `coerce_draft_config` first at the API boundary. This makes correctness depend on `Policy.__init__` having already normalized `policy_cfg["draft"]` in place before this runs. Consider having `resolve_draft_speculator` call `coerce_draft_config` too (or accept `Eagle3DraftConfig | Mapping | None` like its sibling) so a raw-dict `policy_cfg` fails with a clear validation error instead of a bare `AttributeError` if this is ever called outside the current call-order contract.

yuekaizhang (commented):
Reviewed the typed speculator config/registry contract. Verified the TypedDict→BaseModel migration is complete (no stale dict-style `draft` config access left anywhere), and that EAGLE-3 behavior is preserved (same hook order, same `build_draft_model` call). One inline point below, plus a doc gap noted as a separate comment (the doc file isn't touched by this diff so GitHub won't let me anchor a line comment to it).

yuekaizhang:
This PR adds `policy.draft.speculator_type` to `Eagle3DraftConfig`, but `docs/guides/eagle3-speculative-decoding.md` (Important Knobs section, around line 158) isn't updated to list it alongside `enabled`/`model_name`/`loss_weight`. Worth a one-line addition noting it's currently pinned to `eagle3` and exists as the extension point for future speculators (DFlash/DSpark).

nemo_rl/models/megatron/draft/training.py
seonjinn: Fixed in 8852e017d. `resolve_draft_speculator` now takes `Eagle3DraftConfig | Mapping[str, Any] | None` and runs `coerce_draft_config` first, so it validates rather than depending on `Policy.__init__` having normalized `policy_cfg["draft"]` in place beforehand.

Your read of the call-order contract is right, and it is the only thing holding the current path up: `PolicyConfig` declares `draft: NotRequired[Eagle3DraftConfig]`, but that is a TypedDict, so at `setup.py:1624` the value is whatever the YAML produced. The single normalization site is `lm_policy.py:125`. Any setup path reached without it hit `AttributeError: 'dict' object has no attribute 'enabled'` on a disabled draft too, not just an enabled one, since the `not config.enabled` short-circuit is itself the attribute access.

Two tests added: `{"enabled": False}` resolves to `None`, and a raw mapping with `enabled=True` resolves to a speculator whose `.config` is a validated `Eagle3DraftConfig` (so `build_draft_model` still receives a model, not the dict).

seonjinn:
Added in 8852e017d, in the Important Knobs list right after `policy.draft.enabled`:

- `policy.draft.speculator_type`: which speculator implementation to build. Currently pinned to `eagle3`, which is also the default, so configs that predate this key keep working unchanged; it exists as the extension point for future speculators (DFlash/DSpark).

Also re-merged latest `main` into this branch (behind=0 now), so the staleness gate that was failing the last CICD run is clear.

zpqiu (dismissed):
Hi @seonjinn, overall, LGTM. The typed config → resolver → builder separation looks reasonable and is consistent with the project’s existing patterns.

One non-blocking suggestion: add tests to verify that when `draft.enabled=false`, the pre-wrap hook returns the original policy models without calling the draft builder.

Have the affected EAGLE-3 co-training nightly tests been run on the current head? I only found focused/unit-test results in the validation notes. If they have, could you share the results? Otherwise, it would be helpful to run them to confirm that the existing training and refit behavior is preserved.

seonjinn:
Thanks both - covering the two open asks and the CI note here.

@zpqiu on the disabled-path test: added in `6687969` as `tests/unit/models/megatron/test_megatron_setup.py::TestDraftSetup::test_draft_pre_wrap_hook_is_identity_when_draft_disabled`. It parametrizes over the three shapes a disabled config can take (`draft` key absent, raw dict `{"enabled": False}`, typed `Eagle3DraftConfig(enabled=False)`) and asserts the hook returns the same chunk list, attaches no `draft_model`, and never calls `build_draft_model`, `get_pg_collection` or `_load_checkpoint_from_path`. It runs with `preload_policy_from_pretrained=True` on purpose, so the preload branch would fire if the short-circuit regressed.

@yuekaizhang on the doc gap: `docs/guides/eagle3-speculative-decoding.md` Important Knobs now lists `policy.draft.speculator_type` right under `enabled` (line 159), noting it is pinned to `eagle3`, that `eagle3` is also the default so pre-existing configs are unchanged, and that it is the extension point for DFlash/DSpark.

On the nightly EAGLE-3 co-training suites: they have **not** been run on this head. The validation notes only cover focused and unit tests, as you read them. I will post the nightly results here once I have them rather than claim coverage I do not have.

Head is now merged with latest `main` (`c49d53e2e`) - clean merge, no conflicts - and `ruff check --select I`, `ruff check` and `ruff format --check` are clean tree-wide. `/ok to test` re-posted against the merged head.

terrykong (approved):

PR #3702, opened 2026-08-18
fix(nemo-gym): honour pad_dynamic_image_shapes on the dedup prompt path

DanialTaheri:
## What

A NeMo-Gym prompt's images are attached in one of two places, and only one of
them read `pad_dynamic_image_shapes`:

| `deduplicate_multimodal_data` | attach site | flag honoured |
|---|---|---|
| off (default) | `nemo_gym.py` — env actor, per turn | yes |
| **on** | `rollouts.py` — driver, once before `repeat_interleave` | **no** |

The dedup call site passed neither value, so the parameter took its default of
`False`. This threads it from `env.nemo_gym` at all three call sites through one
accessor, so both paths process the same prompt under the same rules.

## Why it breaks

The flag decides whether the processor is asked for stacked tensors
(`return_tensors="pt"`) or a ragged CHW list. A prompt carrying **more than one
image at differing resolutions** cannot be stacked — which is exactly what a
dynamic-resolution processor produces, and exactly what the flag exists to
handle.

So the same prompt trained fine with dedup off and raised with it on.

## Evidence

**Cluster A/B** — 16 nodes, Super Omni, `++grpo.deduplicate_multimodal_data=true`,
identical dataset of multi-image prompts at mixed resolutions (multi-image rows
are ~1.3% of the blend, so the dataset was filtered to them; a random sample
would not reach this path).

| arm | job | Slurm | outcome |
|---|---|---|---|
| without fix | 1716862 | **FAILED** (8m17s) | 24 stack errors, never reached step 1 |
| with fix | 1716888 | **COMPLETED** (13m54s) | Step 1/1, rollouts 100%, 0 stack errors |

Failure from the unfixed arm, raised inside `AsyncTrajectoryCollector`:

```
    return torch.stack(value)
RuntimeError: stack expects each tensor to be equal size,
              but got [3, 576, 576] at entry 0 and [3, 512, 576] at entry 1
```

Real dimensions from the blend — two 576-wide images differing in height.

**Local, same config both ways** (`pad_dynamic_image_shapes = True` read from
`env.nemo_gym` in both; only the call site differs):

```
with fix    -> attached OK, imgs_sizes=[[3, 2], [5, 4]]
without fix -> RuntimeError: stack expects each tensor to be equal size
```

## Scope

Opt-in only: `deduplicate_multimodal_data` defaults to `False`, and the helper's
docstring already noted *"Flag-off runs never call this helper."* It needs dedup
on **and** a multi-image prompt **and** differing resolutions, which is why it
has gone unnoticed.

## Tests

`tests/unit/experience/test_initial_gym_image_payload_shapes.py`:

- the default raises on a mixed-resolution prompt — documents the gap
- with the flag, the batch is padded **and** `imgs_sizes` keeps the true
  per-image extents (`[[3,2],[5,4]]`, not the padded `[5,4]` twice), which the
  model needs to crop each image back out

One existing test changed: `fake_attach` in `test_rollouts.py` gains the new
keyword, forced by the signature change.

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -85,7 +85,11 @@ from nemo_rl.distributed.virtual_cluster import (
     prepare_segment_topology,
 )
 from nemo_rl.environments.interfaces import EnvironmentInterface
-from nemo_rl.environments.nemo_gym import should_use_nemo_gym, spinup_nemo_gym_actor
+from nemo_rl.environments.nemo_gym import (
+    get_pad_dynamic_image_shapes,
+    should_use_nemo_gym,
+    spinup_nemo_gym_actor,
+)
 from nemo_rl.experience.interfaces import (
     NEXT_NEMO_GYM_TASK_INDEX_KEY,
 )
@@ -2224,19 +2228,6 @@ def _get_effort_config(master_config: MasterConfig) -> Optional[EffortLevelsConf
     return EffortLevelsConfig.model_validate(effort_dict)
 
 
-def get_pad_dynamic_image_shapes(master_config: MasterConfig) -> bool:
-    """Return env.nemo_gym's pad_dynamic_image_shapes, defaulting to off.
-
-    The NeMo-Gym actor reads this from its own config for the per-turn attach.
-    The initial-payload attach runs in the driver instead, so it has to be read
-    here and passed down, or multi-image prompts would be processed under
-    different rules on the two paths.
-    """
-    if "nemo_gym" not in master_config.env:
-        return False
-    return bool(master_config.env["nemo_gym"].get("pad_dynamic_image_shapes"))
-
-
 def _pad_teacher_logprobs(teacher_logprobs: torch.Tensor, train_S: int) -> torch.Tensor:
     """Right-zero-pad teacher logprobs ``[B, teacher_S]`` to ``train_S``.
 
@@ -2858,7 +2849,7 @@ def grpo_train(
                             batch,
                             processor,
                             pad_dynamic_image_shapes=get_pad_dynamic_image_shapes(
-                                master_config
+                                master_config.env
                             ),
                         )
                     # Repeat batch items
@@ -3888,7 +3879,7 @@ def validate(
                         val_batch,
                         processor,
                         pad_dynamic_image_shapes=get_pad_dynamic_image_shapes(
-                            master_config
+                            master_config.env
                         ),
                     )
                 generation_config = master_config.policy["generation"]
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -2224,6 +2224,19 @@ def _get_effort_config(master_config: MasterConfig) -> Optional[EffortLevelsConf
     return EffortLevelsConfig.model_validate(effort_dict)
 
 
+def get_pad_dynamic_image_shapes(master_config: MasterConfig) -> bool:
+    """Return env.nemo_gym's pad_dynamic_image_shapes, defaulting to off.
+
+    The NeMo-Gym actor reads this from its own config for the per-turn attach.
+    The initial-payload attach runs in the driver instead, so it has to be read
+    here and passed down, or multi-image prompts would be processed under
+    different rules on the two paths.
+    """
+    if "nemo_gym" not in master_config.env:
+        return False
+    return bool(master_config.env["nemo_gym"].get("pad_dynamic_image_shapes"))
+
+
 def _pad_teacher_logprobs(teacher_logprobs: torch.Tensor, train_S: int) -> torch.Tensor:
     """Right-zero-pad teacher logprobs ``[B, teacher_S]`` to ``train_S``.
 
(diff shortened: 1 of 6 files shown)


nemo_rl/algorithms/grpo.py
```diff
@@ -2208,6 +2208,19 @@ def _get_effort_config(master_config: MasterConfig) -> Optional[EffortLevelsConf
     return EffortLevelsConfig.model_validate(effort_dict)
 
 
+def get_pad_dynamic_image_shapes(master_config: MasterConfig) -> bool:
+    """Return env.nemo_gym's pad_dynamic_image_shapes, defaulting to off.
```
aroshanghias-nvd: **[SUGGESTION]** `pad_dynamic_image_shapes` belongs to `env.nemo_gym` and is not GRPO-specific. Could we place this accessor in `nemo_rl/environments/nemo_gym.py` (or the corresponding config module) and have it accept `master_config.env` rather than the full `MasterConfig`? This would keep NeMo Gym configuration interpretation with the environment while allowing the rollout helper to continue receiving the resolved boolean.

aroshanghias-nvd (commented):
One non-blocking code-organization suggestion.

nemo_rl/algorithms/grpo.py
DanialTaheri: Thanks @aroshanghias-nvd . Addressed your comments.

DanialTaheri:
## Self-review (agent-team), + rebased onto `main`

Rebased onto `ea147e38d` — the branch was 11 commits behind and 5 of its 6 files had changed upstream. Every conflict was import-only (`main` split `MasterConfig` into `MasterConfig as GRPOMasterConfig` and added `should_use_nemo_gym` to the same import blocks this PR edits). No logic conflicts.

A review team (RL-correctness, bug-finder, design, tests) went over the diff, and a devil's-advocate pass then tried to refute each finding. **No correctness bug was found in the change.** Six attack vectors came back clean — call-site coverage, `master_config.env` typing, default-vs-actor agreement, truthiness coercion, keyword-only callers, import cycles — and the training-correctness question resolved cleanly: padded pixels never reach the model, because both backends crop off `imgs_sizes` (`modeling_nemotron_omni.py:470`, Automodel `model.py:634`), the HF path that can't take `imgs_sizes` raises rather than proceeding, and the padded tensor never reaches vLLM.

### Fixed in the follow-up commit

**The crash test asserted an exception production never raises.** It expected `RuntimeError` matching an ATen message. The real image processor overrides the caller when shapes differ (`image_processing.py:151`, `tensor_type=(return_tensors if all_same_shape else None)`) and the failure surfaces from `BatchFeature` as **`ValueError`** (`transformers/feature_extraction_utils.py:206`), with the torch error only as `__cause__`. The test passed solely because the fake called `torch.stack` directly. It now asserts the `return_tensors` the processor was handed — the actual behaviour under test, uncoupled from a torch message.

**The fake didn't emit `imgs_sizes`; the real processor always does** (`image_processing.py:113,131,149`), so the guard at `multimodal_utils.py:952` is False in production and the old tests drove the derive-it-here fallback rather than the production branch. Fixed.

**The wiring was untested** — the actual content of the PR. Deleting any of the three `pad_dynamic_image_shapes=` kwargs left the suite green. The helper now takes `env_config` and resolves the flag itself, so there is no per-call-site value left to get wrong; all three sites already held `master_config.env`. Mirrors `get_nemo_gym_thinking_tags(env_config)`.

**New coverage:** config-to-processor wiring at all three flag values, the equal-resolution multi-image case, and an explicit `false`. (The equal-resolution test earned its keep immediately — it caught a fidelity gap in the fake, since `BatchFeature` converts *every* key under `return_tensors="pt"`, not just `pixel_values`.)

### Worth knowing for reviewers

**No committed config exercises this path.** Resolving every YAML under `examples/` through the repo's own loader: `pad_dynamic_image_shapes: true` resolves in 5 configs, all with dedup unset; `deduplicate_multimodal_data: true` resolves in 5, all with the pad flag unset — and only one of those even has `env.nemo_gym`. No `defaults:` chain bridges them and no test script sets either. This is a **forward-looking fix**; low merge risk, but CI will not regression-test it. The cluster A/B reached the path only via `++grpo.deduplicate_multimodal_data=true`.

### Follow-ups — separate, not this PR

- `validate()` repeats before attaching and without `share_immutable_media` (`grpo.py:3869` vs `:3877`), so the processor runs per repeated row. Pre-existing, and currently unreachable — every committed dedup recipe has `val_num_generations_per_prompt: 1`.
- `distillation.py:542-549` never forwards `tokenizer_config` when hand-rolling `NemoGymConfig`. A missing feature in an untouched file (distillation has no multimodal driver plumbing at all), not a defect this PR introduces.

### Dropped after adversarial verification

Two findings were raised and then **refuted empirically** by pulling the real processors from HF and running them: a suspected numpy-survives-`_restore_tensors` regression for equal-resolution multi-image prompts (real processors return torch tensors or lists of tensors — never numpy; flag-on/flag-off payloads are byte-identical), and a suspected silent-corruption hole in `_stack_ragged_pixel_values` (unreachable; the argument conflated the actor's processor with the driver's). Neither touches code in this diff.

### Checks

`ruff check` clean, `ruff format --check` clean, 9 + 5 unit tests passing locally. Full `tests/unit` needs CI — Ray's GCS server won't start on the machine this was reviewed on, and `tests/unit/conftest.py` has a session-scoped autouse `init_ray_cluster`.

yfw (approved):

PR #3703, opened 2026-08-18
feat(draft): stream vocab-parallel soft CE

seonjinn:
## Summary

- Adds vocab-parallel soft cross-entropy for draft distillation.
- Caches single-tile distributions and recomputes larger inputs by tile to bound saved FP32 vocabulary state.
- Returns additive numerator and token-count statistics for global normalization.

## Scope

The existing EAGLE loss keeps its current behavior. This standalone primitive can merge independently.

## Validation

- Exact head: `895e0c2123164119678c76e5b8ea217f025f5de6`
- 27 current-head GitHub checks pass; the docs test is still running.

## Performance evidence

- GB200 TP2 BF16: peak allocation fell 12.5% at 1K/4K tokens and 24.9% at 8K; the 8K recompute route traded this for 51.5% higher isolated-kernel latency. Losses agreed within 5e-6.

## Dependencies

None.


seonjinn:

Reviewed exact head 9cdc84fe with independent bug-finder and devil’s-advocate passes. Cached and multi-tile autograd math, TP reductions, masking/bin scaling, teacher detachment, and saved-tensor contracts are correct. No unresolved correctness, security, or design findings remain.

Resolved during self-review:
- added explicit cached-versus-recompute boundary coverage;
- added dense/BF16 and TP=2 parity for both routes;
- replaced the always-recompute description with the actual hybrid contract;
- disclosed the measured 8,192-token isolated-kernel latency/memory tradeoff.

Exact OCI-Hsg evidence: functional job 6300378 and benchmark job 6300380.

seonjinn:
Self-review note (found while reviewing the stacked #3715, but owned by this PR's diff):

1. **The `global_valid_toks` wiring is only mock-tested.** `tests/unit/algorithms/test_draft_loss_wrapper.py` patches `streaming_vocab_parallel_soft_ce` and asserts only `assert_called_once()` with no argument inspection, so a swapped or dropped `normalization_counts=global_valid_toks.reshape(1)` in `DraftCrossEntropyLossFn.__call__` would pass CI silently. One `assert_called_once_with(...)` (or a call-args check on the normalization argument) closes it.

2. **`DistributedCrossEntropy` becomes dead code.** After this PR rewires `DraftCrossEntropyLossFn` to the streaming path, `nemo_rl/distributed/model_utils.py::DistributedCrossEntropy` has zero callers repo-wide (verified by grep at the stacked head). Either remove it here or state that the cleanup is deferred, so nobody reuses the stale autograd.Function later.

nemo_rl/algorithms/loss/draft.py
```diff
@@ -0,0 +1,404 @@
...
+        )
+    if token_chunk_size < 1:
+        raise ValueError(f"token_chunk_size must be positive, got {token_chunk_size}.")
+    if weights is None:
+        weights = torch.ones(1, dtype=torch.float32, device=mask.device)
+    elif weights.ndim != 1 or weights.shape[0] < 1:
+        raise ValueError(f"weights must be a nonempty vector, got {weights.shape}.")
+    num_bins = weights.shape[0]
+
+    if bin_ids is None:
```
seonjinn, line 373: Caller-supplied `bin_ids` are validated for shape/dtype/device but not range: an id outside `[0, num_bins)` reaches `scatter_add_`/`index_select` and surfaces as a device-side assert on CUDA (loud, but cryptic and batch-fatal) or a RuntimeError on CPU. Unreachable today -- the only wired caller never passes `bin_ids`/`weights` -- but this function is explicitly the multi-bin primitive later PRs build on.

Added in the follow-up commit: a range check on caller-supplied ids with a clear message, placed so the internal single-bin zeros default pays no extra device sync, plus tests for both out-of-range directions.

seonjinn (commented):
Self-review before requesting reviewers (supplements the two findings already posted as a comment: the mock-only `global_valid_toks` wiring test and `DistributedCrossEntropy` becoming dead code).

Method: a focused review agent over the surfaces not already verified by the stacked PRs' reviews, then an adversarial pass with empirical CPU execution. Already verified elsewhere in this series and re-confirmed here: the cached-path max-cancellation algebra (independently re-derived), the exact transitive cached-vs-streaming parity (both routes share a seeded input and the same dense oracle, including the exact `num_tokens == token_chunk_size` boundary), the one-tile FP32 memory bound, and backward-staleness impossibility. The EAGLE single-pass loss is untouched by this diff, as claimed.

One code finding survived (inline, fixed in the follow-up commit). One candidate -- "silent CUDA corruption from out-of-range bins" -- was empirically refuted: CUDA scatter kernels bounds-assert loudly, so the issue is diagnosability, not corruption.

**Process, before merge:** the head has only lightweight checks; the last `/ok to test` targeted a SHA twelve commits back, so the new `test_draft_soft_ce.py` suites have never run in CI. Since this PR is the base of the draft co-training stack, a green run on the actual head gates everything above it. Posting `/ok to test` on the post-fix head.

seonjinn:
CI triage for the L1_Functional_Tests_Megatron_4 failure: `grpo_megatron_generation_topology.sh` trained both steps normally (KL 0.0005-0.0007, topology-aware allocation engaged, both log guards passed) and then exited non-zero during teardown before the metrics-json check could run -- no traceback in the log. Nothing in this PR's diff (draft-loss primitives + `DraftCrossEntropyLossFn` rewiring) is on the plain-GRPO path this test exercises, so this looks like a teardown flake; retriggering on the same head. The `CI quality check` failure is the label gate.

seonjinn (commented):
## Self-review

I ran a multi-agent review over this branch merged with latest `main` (`8d1ca372f`). Six candidate findings came out of the first pass; an adversarial second pass killed or downgraded four of them with measurements. Reporting both halves, because the refutations are the more useful half for anyone reviewing this.

### Actionable

**1. `DistributedCrossEntropy` is now dead code.** This PR deletes the last import of it from `nemo_rl/algorithms/loss/loss_functions.py`, orphaning `nemo_rl/distributed/model_utils.py:213-267` (55 lines). A full-tree grep — no path filters, so docs and YAML are covered — returns exactly one hit: the `class` statement itself. No `__all__`, no re-export through any `__init__.py`, no dynamic lookup, no tests.

I tried to refute this by arguing `model_utils.py` is already full of unreferenced public symbols and one more is repo-normal. That failed: I audited every zero-external-reference candidate in the file (`AllGatherCPTensor`, `ChunkedDistributedHiddenStatesToLogprobs`, `dtensor_from_parallel_logits_to_logprobs`, `from_parallel_hidden_states_to_logprobs`, `_AllReduceSum`, `all_to_all_vp2sq`) and every one has an internal call site. `DistributedCrossEntropy` is the only genuinely unreachable symbol in the file.

Either delete it here or say explicitly that it is kept for out-of-tree consumers.

**2. `token_chunk_size` has no config surface.** It exists only as a function-parameter default (`loss_functions.py:76`, `4096`). `DraftCrossEntropyLossFn` is constructed at `nemo_rl/algorithms/loss/wrapper.py:279` with `vocab_parallel_group` only; there is no `MasterConfig` field, no YAML key, and no `**kwargs` passthrough. The route between cached and streaming is therefore not tunable, only inferable from the batch shape.

Worth noting this is less severe than it first looks: the knob auto-engages, since any microbatch above 4096 tokens streams with no configuration, which is the memory-pressure regime it exists for.

### Nit

`streaming_vocab_parallel_soft_ce` contracts the teacher expectation against raw student logits (`draft.py:251-262`) while the cached route uses the max-shifted form (`draft.py:123`), so the two routes are not algebraically identical. See below for why I am not filing this as a stability bug.

These looked like findings and are not. Measurements are on CPU torch 2.14 against the real modules.

- **"The streaming route is numerically unstable / missing a max-subtraction."** The forward error is governed by `|common logit offset| / CE`, not by absolute logit magnitude. I measured real pre-softmax logits from a Qwen3-family checkpoint across five prompt types (factual, code, math, degenerate repetition, garbage tokens): max `|logit|` 26.60, max `|mean logit|` 2.53, min per-token CE 0.96, giving an observed ratio of 2.64. The ratio needed for even 1e-5 relative error is ~1e3 — three orders of magnitude short. A pure scale sweep (logit std 1 → 1000, max logit up to 4777) held streaming error flat at 1e-7..1e-8, identical to cached; degradation needs an artificial constant offset with O(1) spread, and softmax shift-invariance means training exerts no pressure toward one.

  I also implemented the obvious fix (return `maxima` from `_tile_log_normalizers`, shift the einsum) and measured it against an fp64 oracle. **Gradient error is bit-identical before and after** — 9.79e-9 at C=0, 5.69e-7 at C=1e3, 3.96e-5 at C=1e5. The gradient error originates in the backward's `_tile_distributions` (`draft.py:99-100`), which that change does not touch. The fix repairs the reported loss value and not the thing that drives training.

- **"The vocab-parallel `all_reduce` calls have zero executed coverage."** False. All three mutants survive on a CPU laptop because `tests/unit/distributed/test_draft_soft_ce.py` requires 2 GPUs — but the always-on tier has them. `.github/workflows/cicd-main.yml:754` runs `cicd-unit-tests` on the H100 runner with `cpu-only: "false"`, `.github/actions/test-template/action.yml:177` launches with `--runtime=nvidia --gpus all`, `L0_Unit_Tests_Distributed.sh` runs all of `unit/distributed/`, and `excluded_unit_tests.sh` does not exclude the file. Decisively, that same exclusion file deselects individual `TestTwoGPUCluster::` cases elsewhere with per-case runtimes cited — you do not measure runtimes for tests that skip. "Mutants survive on my machine" is not "there is no coverage."

- **"14 of 15 validation branches are untested."** Grep-accurate and not worth acting on: each is a `raise ValueError` with a self-describing message on a helper with one in-repo caller, so tests would assert message strings against themselves. At most two are plausible caller mistakes (`draft.py:372`, `bin_ids` dtype not `torch.long`; `draft.py:365`, `bin_ids` required when multi-bin).

- **"Zeroing the max-subtraction passes every test."** True but about test strength, not about this code, which is correct. The mutant only diverges past fp32 overflow (max logit ~88 = `ln fp32_max`), which is 3.3x above the real 26.6 — so even a realistic-input test would not catch it.

### Verified clean

Forward and backward of both routes agree with a dense `-softmax(teacher)·log_softmax(student)` oracle. Route selection depends only on batch shape, which is replicated across TP ranks, so it cannot desynchronize a collective. Peak fp32 vocab footprint is `2 × token_chunk_size × V` on both branches at the boundary, and neither route regresses against the `DistributedCrossEntropy` path it replaces, which also held two full fp32 `(T,V)` tensors.

### Still open

No memory or step-time number is attached to the cached-vs-streaming swap. The streaming route issues 3 TP all-reduces per token chunk against 2 total for cached, and its backward recomputes `exp()` over the full vocab per tile. The memory saving plausibly dominates, but it is unmeasured. I will attach peak draft-loss memory and draft-microbatch wall time at real TP and sequence length.

seonjinn (commented):

Re-merged this branch with `main` at `4213e0867`. Clean merge, no conflicts. Of the 109 files (+4118 / -1387) `main` has gained since the previous review base (`8d1ca372f`), exactly one is also touched by this PR — `pyrefly.toml` — and the two edits land in different sections. No overlap at all in `nemo_rl/`. `ruff 0.9.9 check` and `format --check` clean on the merged tree (734 files); CI green.

This is the most isolated of the four PRs: the loss kernel is new code in a new module, so `main` moving does not reach it. The previous review stands unchanged.

### One forward-looking nit

The `bin_ids` range validation (`draft.py:378-390`) calls `int(bin_ids.min())` and `int(bin_ids.max())`, which are two device syncs on CUDA. The comment correctly notes that the internal single-bin default skips this path and pays no sync, so there is no cost on any caller that exists today.

It becomes a per-call cost the moment a multi-bin caller lands — which is the multi-pass TTT reduction, i.e. a hot loop. Worth deciding now whether to gate it behind a debug flag, or to keep it and accept the sync on the grounds that the alternative is a device-side assert deep inside `scatter_add_`. I lean toward keeping it as written and revisiting only if it shows up in a profile; flagging it so the choice is recorded rather than discovered later.

### Checked and clean

I went looking for a device or dtype gap on `weights`, since it is the one argument with no explicit device check alongside `student_logits`/`teacher_logits`/`mask`/`bin_ids`. There is none: `draft.py:417` normalizes it with `weights.to(device=mask.device, dtype=torch.float32)` before it reaches `DraftLossStats`, so a caller passing CPU weights against CUDA logits is handled rather than deferred to a mismatch inside `normalized()`.

### Method note

Lint only; `torch` is not importable here, and the two-GPU test in `tests/unit/distributed/test_draft_soft_ce.py` was not run. Everything else read from the merged tree.

nemo_rl/algorithms/loss/loss_functions.py
```diff
@@ -73,8 +73,10 @@ class DraftCrossEntropyLossFn(LossFunction):
     def __init__(
         self,
         vocab_parallel_group: Optional[torch.distributed.ProcessGroup] = None,
+        token_chunk_size: int = 4096,
```
yuekaizhang: `token_chunk_size=4096` is a hardcoded constructor default that's never surfaced through `DraftCrossEntropyLossConfig` (only `vocab_parallel_group` is declared there), and `DraftLossWrapper` never passes it either. Per the config-conventions guideline, a config value shouldn't rely on a function-parameter default — this tile size directly bounds the FP32 vocab memory footprint (the point of this PR), so it should be threaded through `DraftCrossEntropyLossConfig` with a documented default in the exemplar YAML rather than left as a magic constant.

yuekaizhang (commented):
Traced the forward/backward math for both the cached and streaming vocab-parallel soft-CE paths by hand against the vocab-parallel log-sum-exp identity, and checked the "EAGLE loss keeps current behavior" claim — both check out. Two smaller points below.

nemo_rl/algorithms/loss/draft.py
```diff
@@ -0,0 +1,418 @@
...
+        ctx.save_for_backward(
+            student_logits,
+            teacher_logits,
+            mask,
+            bin_ids,
+            saved_log_normalizers,
+        )
+        # pyrefly: ignore[implicitly-defined-attribute]
+        ctx.token_chunk_size = token_chunk_size
+        ctx.tp_group = tp_group  # pyrefly: ignore[implicitly-defined-attribute]
```
yuekaizhang: `ctx.tp_group` is stored in forward but never read in backward (the saved `log_normalizers` already reflect the cross-TP reduction, so backward doesn't need to communicate). Minor — consider dropping the unused assignment so it doesn't imply backward does TP work.

nemo_rl/algorithms/loss/loss_functions.py
seonjinn: Fixed in 3e4cc829f. `token_chunk_size` is now a real config value rather than a constructor default nothing could reach.

The plumbing, end to end:

- `DEFAULT_DRAFT_TOKEN_CHUNK_SIZE = 4096` in `nemo_rl/algorithms/loss/draft.py`, next to the kernel whose tile it sizes, with the memory-vs-launch-count tradeoff written down.
- `DraftCrossEntropyLossConfig` declares `token_chunk_size: NotRequired[int]`, and `DraftCrossEntropyLossFn.__init__` defaults to the constant instead of a bare `4096`.
- `DraftLossWrapper` takes and forwards it, and both `DraftLossWrapper` construction sites in `nemo_rl/models/megatron/train.py` (packed and unpacked) read `policy.draft.token_chunk_size`.
- `DraftConfig` gains the key, and `examples/configs/grpo_math_1B.yaml` carries `token_chunk_size: 4096` with a comment on when to move it.

You are right that this is the value that matters here: the streaming kernel materializes `token_chunk_size x vocab` in FP32, so it, not the batch, sets the peak the PR is meant to cap. Test added on the forwarding specifically, because a default swallowed at the wrapper hand-off would present as a silent OOM regression at the tile boundary rather than as a config bug.

Adding the key to the exemplar is safe for `tests/unit/test_recipes_and_test_suites.py`: recipe configs merge onto the base, so they just inherit it.

nemo_rl/algorithms/loss/draft.py
seonjinn: Dropped in 3e4cc829f, and replaced with a comment saying why it is absent, so the next reader does not re-add it looking for the missing collective:

```python
# tp_group is deliberately not saved: backward rebuilds both
# distributions from the already TP-reduced log-normalizers, so the
# gradient is purely local and needs no collective.
```

Your reasoning is the reason it is safe to remove: backward recomputes `student_probs - teacher_probs` from `flat_student`/`flat_teacher` and the saved `log_normalizers`, and the normalizers already carry the cross-TP reduction from forward, so every term is local to the shard.

seonjinn:
/ok to test [3e4cc82](https://github.com/NVIDIA-NeMo/RL/pull/3703/commits/3e4cc829fdae6e3a6b951630fdcee38ab87f1093)

zpqiu:
Overall LGTM on the design and approach. 

One **non-blocking follow-up**: `token_chunk_size` currently controls both the FP32 tile size and the cached-vs-recompute decision. That seems reasonable for the current single-pass path. When this is reused for multi-pass TTT, however, several individually small passes could all take the cached route and retain FP32 probabilities until the shared backward. Allowing that caller to force recomputation independently of tile size may be useful, without having to shrink tiles just to select the recompute route. I don’t think this needs to expand the scope of this PR. cc @MarkXCloud

Before merging, please also sync the reference YAML for `policy.draft.token_chunk_size` to address the reference-config test failure.

seonjinn:
@zpqiu thanks - both points.

Reference config: already synced, so the reference-config test should be green. `tests/unit/reference_configs/grpo_math_1B.yaml` carries `policy.draft.token_chunk_size: 4096` matching `examples/configs/grpo_math_1B.yaml`, which is what `test_config_v2.py::test_reference_configs_up_to_date` diffs. If you saw the failure on an older head, it predates that sync.

Decoupling the recompute decision from the tile size: agreed, and agreed it should not expand this PR. Today `token_chunk_size` does carry both roles, and the multi-pass TTT case you describe is real - several small passes each taking the cached route and holding FP32 probabilities until a shared backward is exactly the shape where you would want to force recomputation without shrinking tiles. The clean split is a separate `recompute` selector on the loss config that defaults to the current size-derived behavior, so nothing changes for the single-pass path. Happy to file it as a follow-up issue and reference this thread. cc @MarkXCloud

Head is now merged with latest `main` (`c49d53e2e`), clean merge, lint clean tree-wide, and `/ok to test` re-posted against the merged head.

terrykong (dismissed):

terrykong:
@seonjinn can you rebase and restart CI?

seonjinn:
/ok to test [7de1a99](https://github.com/NVIDIA-NeMo/RL/pull/3703/commits/7de1a99c437cd7134a26c26a417fd39f20732a8d)

seonjinn:
/ok to test [9ca803f](https://github.com/NVIDIA-NeMo/RL/pull/3703/commits/9ca803ff97f6103159428fb23405903cd6b7f37d)

terrykong (approved):

Issue #793, opened 2025-07-29
[multimodal dtensor] Inconsistent logprobs for multimodal models


rohitrango:
**Describe the bug**

When running GRPO for VLMs (Qwen2.5VL, LLaVa, etc.) the logprobs generated by vllm and that by huggingface differ by a margin higher than 1.05. Although the policy converges across different VLMs.

**Steps/Code to reproduce bug**

Run `uv run examples/run_vlm_grpo.py` from PR #712 

**Expected behavior**

**Environment overview (please complete the following information)**

 - Environment location: [local / cluster]
 - Method of install: [pip install or from source].

joyang-nv:
@yuki-97 is vllm upgrading related to this? Should we ask reporter to try again after upgrading?

PR #3704, opened 2026-08-18
feat(refit): integrate ModelExpress weight synchronization

zhengluo-nv:
## Summary

- add `model_express` as a NeMo-RL weight synchronization transport using the public ModelExpress control, trainer, and generator clients
- integrate Megatron publication and vLLM staged installation through existing framework-native worker RPCs
- keep version orchestration and safe points in NeMo-RL while ModelExpress owns tensor selection, manifests, NIXL staging, verification, and vLLM installation
- support Megatron MoE publication with expert parallelism and non-colocated vLLM tensor parallelism

## Architecture

NeMo-RL owns orchestration and safe points. ModelExpress owns rank-local transport resources and engine-specific transfer/install behavior.

```mermaid
flowchart LR
    D["NeMo-RL synchronizer"] -->|"create version"| C["MX control service"]
    D -->|"publish"| T["Megatron trainer clients"]
    T -->|"tensor manifests"| C
    D -->|"update exact version"| G["vLLM generator clients"]
    G -->|"discover + lease"| C
    T -->|"NIXL tensor reads"| G
    G -->|"stage, verify, apply"| V["vLLM model"]
    D -->|"retire + release"| C
```

Lifecycle:

1. Policy actors initialize trainer-side ModelExpress resources before distributed setup; generator clients initialize inside vLLM workers and own their NIXL resources.
2. Each Megatron policy rank binds MX-owned tensor specifications, including TP and expert-parallel placement.
3. The synchronizer creates one version and asks participating trainers to publish their shards.
4. Each vLLM worker stages the exact version, verifies it, and applies it at the framework safe point.
5. The synchronizer retires the version and releases trainer publication before training mutates the source buffers.

## Integration boundary

NeMo-RL owns:

- backend selection and version lifecycle orchestration
- framework-native policy and generation RPC dispatch
- the safe point between generation updates and training mutation

ModelExpress owns:

- trainer and generator NIXL resources, including engine-specific early initialization
- Megatron tensor selection and publication metadata
- exact-version source discovery and leases
- NIXL transfer planning, staging, and digest verification
- graph-safe vLLM installation
- MX-specific listener, NIC, and source-selection configuration

## Dependencies

- public RL trainer/generator clients: [ai-dynamo/modelexpress#648](https://github.com/ai-dynamo/modelexpress/pull/648), merged
- composable RL refit paths: [ai-dynamo/modelexpress#690](https://github.com/ai-dynamo/modelexpress/pull/690), merged
- trainer-only MoE publication: [ai-dynamo/modelexpress#704](https://github.com/ai-dynamo/modelexpress/pull/704), merged
- distributed reshard/QKV reliability: [ai-dynamo/modelexpress#635](https://github.com/ai-dynamo/modelexpress/pull/635), merged
- early generator NIXL bootstrap used by the current E2E: [ai-dynamo/modelexpress#793](https://github.com/ai-dynamo/modelexpress/pull/793), open draft

Current integration heads:

- NeMo-RL: `91846f290415746b6a13f1da44b73a1e5c7e1797`
- ModelExpress: `76a1db3dd511df5a637235dd76ff15e41d08d5a0` (#793)

The NeMo-RL branch is squashed to one GPG-signed commit with DCO sign-off.

### Current-head checks

- DCO, semantic PR title, triage, and submodule checks pass on the live PR head
- NeMo-RL Python compilation and `git diff --check` pass for the final source
- the full NeMo-RL monorepo suite has not been run locally; Linux PR CI remains the unit-test gate

### Qwen3-30B-A3B two-step functional E2E

The pre-rebase NeMo-RL head `46bb9801ce14e81a48239a54addc07f29c807a3e` completed a two-step GRPO run on the nscale cluster with `DRIVER_EXIT_CODE=0`. The rebase preserved the same 25-file PR delta; its only content conflict was the `vllm_backend.py` typing import, resolved by retaining both new-main and ModelExpress imports. The rebased head has not received a repeat GPU run.

- model: `Qwen/Qwen3-30B-A3B`, BF16
- hardware: four nodes, 16×B200
- trainer: TP2/PP1/EP2/ETP1/DP2 on eight GPUs
- generation: four independent vLLM 0.29 TP2 replicas on eight GPUs
- each generator rank staged and installed 40.21 GB from all eight trainer publishers
- both steps completed generation, reward calculation, log-prob calculation, backward, optimizer update, and the subsequent refit

| Measurement | Cold | Warm |
| --- | ---: | ---: |
| NeMo-RL `prepare_for_generation` | 55.92 s | 7.93 s |
| ModelExpress generator refit | 43.94–47.36 s | 5.84–6.00 s |
| Transfer planning | 39.24–40.04 s | ~0 s, plan-cache hit |
| Wire transfer | 4.42–7.71 s | 5.73–5.89 s |
| Full GRPO step | 102.91 s | 19.64 s |

The cold path is dominated by first-time transfer-plan construction. On the warm path, planning is cached and wire transfer accounts for nearly all generator refit time.

Test image:

`nvcr.io/0980761089281446/model-express-dev-containers:nemo3704-mx793-vllm029-41de39d-76a1db3-msync`

Digest: `sha256:e928f88cd7559998a17c65a30a2f64c965bdeb785af6b9b3ffda87c4956205d2`

The exact final NeMo-RL `vllm_backend.py` and `vllm_sparse_refit.py` from `46bb9801` were mounted into this image for the run. This validates the final integration behavior, but the published image is not a self-contained build from the final NeMo-RL SHA.

The workers emitted UCX handler warnings during teardown, but the driver completed successfully and all 16 GPU workers were released after the run.

This is functional multi-step reshard/refit evidence, not a formal performance or numerical-parity result. The 10.25 s warm result reported by NeMo-RL #3632 used a different 32×GB200 topology, payload, software stack, cache state, and timer boundaries, so the numbers should not be compared as an A/B benchmark. Explicit post-transfer parameter equality and deterministic generation parity remain follow-up qualification.

### Earlier dense-model coverage

Earlier Llama-3.1-8B runs covered TP1 lifecycle with redundant DP publishers and TP2 trainer to TP1 generator resharding. Both completed one training step with exact-manifest NIXL staging and installation. Those runs remain useful dense-model coverage but are not current-head qualification.


KavinKrishnan:
Two things I hit reading through this, neither an argument against the direction.

The first is concrete. ai-dynamo/modelexpress#648 deletes `modelexpress/engines/vllm/refit/receiver.py`, and that's the class the `mx_reshard` path in #3632 imports. So whichever of these lands first, the other needs a real fix rather than a rebase, and it'd be good to settle which receiver we're keeping before either merges. On top of that we both add a transport to the same `VllmRefitSelector` literal, the same `VllmRefitConfig`, and the same `create_weight_synchronizer` dispatch, so there's some mechanical overlap in ~11 files either way.

The second is about the E2E. TP1 trainer to TP1 generator doesn't move any shard boundaries, so it isn't really exercising the reshard. The two geometries that caused us trouble were the MoE one (Megatron TP2/EP4 into vLLM TP4) and models where KV heads are fewer than TP -- there the per-rank KV count floors to zero, and most ranks legitimately own no K or V rows at all, so the sparse merge is the whole game. Since this moves Megatron tensor selection into MX, those are the two I'd want green before we swap out a path that's already been through them. Happy to point you at the configs and the reduced 64Q/2KV model if that's useful.

zhengluo-nv:
> deletes modelexpress/engines/vllm/refit/receiver.py, and that's the class the mx_reshard path in https://github.com/NVIDIA-NeMo/RL/pull/3632 imports

Yeah, I want to only expose 3 public clients to external, and the receiver will become an internal class. A modularized modelexpress-rl with a handful of simple public APIs should make future framework integration much easier. We will make these public APIs stable during MX version upgrade.

> The two geometries that caused us trouble were the MoE one (Megatron TP2/EP4 into vLLM TP4) and models where KV heads are fewer than TP

I believe I copied all your code into the new folder. So it should not cause regression on the resharding

zhengluo-nv:
Updated in signed commit `bcd6936d9` based on the Qwen3-30B-A3B E2E.

Changes:
- Stop passing the removed `version_number` argument to the current ModelExpress control client; retry correlation remains in the idempotency key.
- Configure vLLM early NIXL initialization for both `nixl` and `model_express` without presenting ModelExpress as a NeMo checkpoint engine. The vLLM payload is now a transport-neutral `nemo_rl_nixl_preinit` block containing only the NIXL backend settings.

E2E evidence:
- Topology: 6x B200, Megatron TP1/EP4 trainer (4 GPUs) -> vLLM TP2 generator (2 GPUs).
- Model: Qwen3-30B-A3B.
- Exact ModelExpress version `26056408` was published and installed on both TP ranks.
- Each TP rank captured 18,867 reshard copies with 0 unsupported sources and transferred 40.611 GB.
- Both ranks completed prepare/install/release; 4 generations, rewards, logprobs, and one policy-training step completed; the run emitted `NEMO_E2E_COMPLETE`.

Qualification: this validates the functional publish -> TP2 reshard/refit -> generation -> training lifecycle. Both vLLM workers subsequently hit a UCX abort inside ModelExpress NIXL teardown, after the training step completed. That cleanup issue remains MX-owned and is not hidden in this NeMo change. Parameter equality, generation parity, repeated multi-step stability, and performance qualification are still separate gates.

zhengluo-nv:
The early ModelExpress generator bootstrap used by this integration is proposed in ai-dynamo/modelexpress#793 (head 57bb0469db5ed7055e3b76a5f570e1b35ef59d41).

zhengluo-nv:
Bootstrap ownership is now in ModelExpress PR ai-dynamo/modelexpress#793 at 76a1db3dd511df5a637235dd76ff15e41d08d5a0. NeMo-RL only selects the MX vLLM worker and no longer handles local rank or device mapping.

alanliu-nvidia:
Blocking: The first refit fails when CP > 1.

initialize_model_express() skips non-zero CP ranks, so they never call bind_tensors(). However, publish and release still run on all workers, causing CP rank 1 to call publish_version() with unbound tensors.

We can try apply the same [CP-rank gate](https://github.com/NVIDIA-NeMo/RL/blob/dd2bfe7e8d00d040248251e37dc1773ed6d130d5/nemo_rl/models/policy/workers/megatron_policy_worker.py#L2954-L2970) to both methods, preferably through a shared predicate, and add a CP-rank-1 regression test verifying that neither client method is called.

PR #3706, opened 2026-08-19
feat(draft): preserve split-step normalization parity

seonjinn:
## Summary

- Accumulates draft-loss statistics across split microbatches.
- Normalizes draft gradients and metrics using the DP-global draft-token count.
- Preserves pipeline-wide gradient-norm reporting and clears state on abort or retry.

## Scope

Split-step state and lifecycle handling around the soft-CE primitive; no new draft model or runtime integration.

## Validation

- Exact head: `76c2648d75c76b9f107f5de3d28f48cfa879316b`
- 27 current-head GitHub checks pass; remaining unit jobs are queued.

## Prior distributed evidence

- OCI `6327354`, `6326981`, `6327012`, `6327013`: split lifecycle, PP2 grad-norm propagation, and DP2/DP4 uneven/zero-count FP32/BF16 parity passed.

## Dependencies / Remaining

Depends on #3703. The branch-up-to-date check is the only current failure.


seonjinn:

Reviewed the PR3-only range on top of #3703, including split-step loss semantics, DP reduction ordering, draft-only gradient rescaling, metrics, and tests.

Resolved during self-review:
- Preserved the runtime policy-loss metric contract instead of applying an invalid combined-loss normalization.
- Made the reduction expectation device-local for GPU execution.
- Corrected the OCI selector so the TQ draft-grad-norm test is exercised.

No unresolved correctness, security, or design findings remain in the PR3-only range.

Validation: OCI-Hsg job 6299067 completed successfully with 6 base tests, 2 MCore split tests, DP=2 NCCL uneven-count gradient/metric parity, Ruff, and compile checks.

tests/unit/models/megatron/test_draft_step_state.py
```diff
@@ -0,0 +1,145 @@
...
+    "draft_step_state_under_test", _STEP_STATE_PATH
+)
+assert _SPEC is not None and _SPEC.loader is not None
+_MODULE = importlib.util.module_from_spec(_SPEC)
+sys.modules[_SPEC.name] = _MODULE
+_SPEC.loader.exec_module(_MODULE)
+DraftStepState = _MODULE.DraftStepState
+
+
+def _stats(numerator: float, count: float) -> DraftLossStats:
```
seonjinn, line 42: No test asserted split-vs-monolithic equality for the draft loss -- the new suites validate accumulation mechanics in isolation, and the existing GPU parity suite has no draft references. The parity does hold (adversarially verified: `roll_tensor` zeroes the vacated slot, so the draft count equals the policy's shifted-mask count by construction, and a CPU run shows identical losses to machine precision), but nothing guarded it against future prepare-fn drift. Added in the follow-up commit: a CPU unit test that accumulates two split halves through `DraftStepState` and asserts bit-level count equality plus loss parity (1e-12) against the monolithic path, including the shifted-first-slot case.

nemo_rl/algorithms/loss/wrapper.py
```diff
@@ -278,12 +280,25 @@ def __call__(
...
+            metrics[DRAFT_STEP_PAYLOAD_KEY] = DraftStepState.metric_payload(stats)
+        else:
+            draft_loss = self.draft_loss_fn(
+                data=data,
+                global_valid_seqs=global_valid_seqs,
+                global_valid_toks=global_valid_toks,
+                **loss_input,
+            )
         combined_loss = policy_loss + self.loss_weight * draft_loss
         metrics["draft_loss"] = float(draft_loss.detach().item())
```
seonjinn: `"draft_loss"` was a bare string literal coupling this file and the worker's metric aggregation with no shared constant (unlike `DRAFT_STEP_PAYLOAD_KEY` right next to it) -- a rename on either side would silently fall back to wrong-scale aggregation. Promoted to `DRAFT_LOSS_METRIC_KEY` in `step_state.py` in the follow-up commit; the import stays method-local (hoisted above the branch so both paths see it) because the draft package's `__init__` pulls the modelopt chain.

seonjinn (commented):

Method: a focused review agent, then an adversarial pass with empirical CPU execution of the real modules.

The headline claim survived a serious challenge worth recording: a candidate CRITICAL argued the split path normalizes by the draft's own count while the monolithic path uses the policy's `global_valid_toks`, breaking parity. Refuted on four grounds: (1) `roll_tensor(token_mask, shifts=-1)` zeroes the vacated last slot, so the draft mask count equals `sum(token_mask[:, 1:])` -- exactly the policy formula -- per sample, by construction; (2) both counts ride the same DP all-reduce; (3) the gradient scale composition (`scale_gradients(1/n)` then `correct_main_grads x n x 1/draft_count`) nets to `raw/draft_count` in both paths; (4) an empirical CPU run with a mask exercising the shifted first slot produced identical losses (relative diff 0.0). Also verified clean: zero-count guards, single-DP-collective accumulation, abort/retry state freshness, PP grad-norm collective symmetry for non-owner ranks.

Follow-up commit: the split-vs-monolithic parity regression test (inline) and the `draft_loss` shared constant (inline).

**Notes:** CI on this head is green except two non-code failures (an HF-CDN 429 download flake and the label-gated quality check). One pre-existing limitation worth a body line: draft + CP>1 fails the loss's mask/logits shape check identically in both paths -- not a parity regression, but worth stating. #3725 stacks on this PR and relies on the `scale_gradients -> correct_main_grads -> optimizer.step()` finalization order; flag any reordering there.

seonjinn:
CI triage for 76c2648d7: both failures were in the review follow-up commit itself — (1) the new split-parity test asserted 1e-12 on an fp32-accumulated value (chunked addition order differs from monolithic by ~2e-7; relaxed to 1e-6 with a comment, since the parity claim is identical normalization semantics, not bitwise equality), and (2) the wrapper test's megatron-free step_state stub predated the new DRAFT_LOSS_METRIC_KEY constant (added to the stub). Fixed on the new head.

seonjinn (commented):
## Self-review

Multi-agent review of this branch merged with latest `main` (`8d1ca372f`), then an adversarial pass. The adversarial pass retracted more than it confirmed, including one finding of my own that I had already committed to the branch and have now reverted. Reporting both halves.

### Actionable

**1. The PR description overstates the scope of the fix.** It says the change preserves split-step normalization parity, without naming which configuration was actually broken. For the default `TOKEN_LEVEL` loss this PR is a numerical no-op: `placeholder_n = torch.tensor(1.0)` (`megatron_policy_worker.py:1514`) is passed as both `global_valid_seqs` and `global_valid_toks` (`:1541-1542`), so the pre-PR draft loss was already an un-normalized sum, and `n_true = global_valid_toks` (`:1630`) matches the draft denominator, making the deferral correction 1.0. Gradients match the monolithic path to ~6e-9 both before and after.

The real fix is for `SEQUENCE_LEVEL` and GSPO (`policy.token_level_loss: false`), where `n_true = global_valid_seqs` and the draft gradient was inflated by roughly the mean sequence length. That is worth stating plainly in the description — a reader currently cannot tell whether their configuration was affected.

**2. Packed sequences silently ignored `defer_normalization`.** Deferral is nested inside the unpacked branch of `DraftLossWrapper.__call__`, so a packed split step would fall through to the non-deferred path without any signal. Packed draft training is reachable — packing is allowed with draft (only PP>1 is blocked), the split path supports packing, and `defer_draft_normalization=True` is hardcoded at `megatron_policy_worker.py:1510`. Added a constructor `ValueError` plus a test rather than leaving it to fall through.

Noting the counter-argument for the record: for `TOKEN_LEVEL` the draft and policy denominators are equal, so the deferral correction is exactly 1.0 and packed + draft + split arguably works today, which makes the guard stricter than strictly necessary. I have kept the guard, on the grounds that a configuration that happens to work because two denominators coincide is not a contract.

### Suggestions (tests, not defects)

Mutation testing over the six affected test files (30 mutations, 24 caught) left three worth closing:

- The degenerate empty-batch case is untested and mildly inconsistent. `scale_gradients` applies `1/n_safe` while `correct_main_grads` receives `n_true` (`megatron_policy_worker.py:1638-1646`); the two differ only when `n_true == 0`, where the current code zeroes draft grads instead of applying the arithmetic inverse. Draft grads are near-zero in an empty batch so there is no practical impact, but replacing `n_true` with `n_safe` passes every test, and a reader has nothing to consult. One `TestFinish` case with `local_valid_toks = 0` asserting the intended behavior, whichever convention is preferred.
- `test_draft_loss_wrapper_defers_raw_stats_for_split_step` patches `DraftStepState.metric_payload` and asserts it was called. That stubs out the one seam it is meant to cover; the real payload never runs. Asserting on `metrics[DRAFT_STEP_PAYLOAD_KEY].stats` values instead would pin the seam.
- No test asserts `"draft_grad_norm" not in out` when the draft state is inactive, although the sibling `grad_norm` key has exactly that negative assertion.

**"The zero-match assert in `correct_main_grads` catches a tagging failure."** This was mine, it was wrong, and it is reverted. I had added a `RuntimeError` when the state is active but no parameter carries `grad_norm_group='draft'`. `active` is job-wide, not rank-local: the payload reaches every pipeline rank through `broadcast_loss_metrics_from_last_stage` (`megatron_policy_worker.py:1571`) and the accumulate loop has no stage guard, while the draft module is attached only to the post-process chunk. Every non-last pipeline rank would have raised, turning draft + PP>1 into a hard crash. The decisive evidence is 60 lines below, where `grad_norms_by_group.get("draft")` under the same `active` flag is deliberately `None`-tolerant for exactly those ranks. A second mechanism makes it worse: under virtual pipeline the draft lives in `model[-1]` while the worker is handed `model[0]` (`setup.py:2092`), so the assert would fire on the owner rank too.

**"The split-path denominator drops `weights` and nothing notices."** Mechanically true — deleting the `* self._weights` at `step_state.py:118-124` leaves the suite green — but inert. `accumulate` hard-rejects any non-unit weight vector (`:78-79`), so `_weights` is always ones and the multiply is by 1. There is no reachable behavior to pin. If multi-bin draft loss ever lands, the accumulate guard fires first.

**"The payload's `detach`/`clone` are unpinned, risking a retained autograd graph in the step-scoped accumulator."** The accumulator is defended twice: `accumulate` independently calls `.detach().clone()` on the first payload (`:81-83`) and `.detach()` on every subsequent one (`:94-95`). Dropping the detach in `metric_payload` cannot leak a graph into the accumulator or alias a caller buffer. `metric_payload`'s detach is still worth keeping — the payload is pickled through `broadcast_object_list` — but the stated OOM failure mode does not exist.

**"`counts_for_reduction` returns a length-0 tensor when inactive, so the DP all_reduce can desynchronize."** No trigger found. `active` is set from a PP-group broadcast and DP-group members share a PP stage, so activity is uniform; empty shards are rejected upstream at `preshard.py:77`. Two reviewers plus the adversarial pass failed to construct a reachable case.

**"The split-parity tests use equal token counts per microbatch, so they cannot distinguish the fixed formula from the buggy one."** False. Both use unequal counts — 7 vs 8, and 5 vs 15 — and the formulas diverge on those inputs (2.3008 fixed vs 2.3107 mean-of-means) well outside the asserted `rtol=1e-6`.

### Method note

The suite was executed on CPU with stubs for the unimportable heavy dependencies: 142 passed, zero skips, stable across three random orderings. `tests/unit/distributed/test_draft_soft_ce.py` was not run — it requires 2 GPUs. The adversarial analysis of pipeline and optimizer semantics is read from source, not executed, since megatron-core is not importable in the review environment.

seonjinn (commented):

Re-merged this branch with `main` at `4213e0867`. Clean merge, no conflicts. `main` has advanced 109 files (+4118 / -1387) since the previous review base (`8d1ca372f`); the only files it shares with this PR are `megatron_policy_worker.py` and `pyrefly.toml`. `main`'s two hunks in the worker are at `:430` and `:524`, far from the split-step normalization path this PR rewrites (`:1500-1650`), and contain no `draft` lines at all — `git diff 8d1ca372f 4213e0867 -- nemo_rl/models/policy/workers/megatron_policy_worker.py | grep '^[+-].*draft'` is empty. `ruff 0.9.9 check` and `format --check` clean (736 files); CI green.

Nothing in the previous review changes. The reverted grad-tagging assert stays reverted, and the packed × deferred-normalization guard stays in.

### One thing worth recording, now that #3699 is in review alongside this

`DraftStepState.accumulate` (`step_state.py:72-79`) rejects any payload that is not a single bin with unit weight. That is the right narrowing for what ships today, but it is also what will reject #3699's multi-pass TTT statistics, which are `[ttt_steps]` bins with per-pass weights `alpha_d`. After this PR merges, a `ttt_steps > 1` run on the split train-step path fails with `ValueError: current EAGLE draft step state requires exactly one bin` rather than mis-normalizing silently — which is an improvement on the current behaviour there, and I have said so on #3699.

The point I want on the record here is that the restriction is a call-site convention, not a structural limit. `DraftLossStats` and `streaming_vocab_parallel_soft_ce` are already multi-bin; `_normalization_scale` (`step_state.py:115-124`) computes `sum_d w_d * count_d` and needs no change; and the weight-drift check at `:92-93` already generalizes. So if multi-pass split-step support is wanted later, the work is relaxing the two guards in `accumulate` and widening `set_global_counts`'s `counts.shape != (1,)` check — not a new data path.

Recording it so the narrowing does not harden into an assumption that the step state is inherently one-bin.

### Method note

Lint only on this pass. The CPU suite result from the previous review (142 passed, three orderings) was not re-run; `torch` is not importable in this environment and nothing in `main`'s delta touches the code those tests cover.

yuekaizhang:
Reviewed the split-step normalization/lifecycle logic. Verified the grad-normalization math (`scale_gradients` + `correct_main_grads` composing to the correct DP-global `1/global_draft_count` scale), zero/uneven draft-count handling, PP grad-norm reporting, and state clearing on abort/retry against both the code and the unit tests — all correct. LGTM.

seonjinn:
Merged latest `main` into this branch (68565423e), which clears the `CONFLICTING` state GitHub was showing and puts the branch at behind=0.

Four files conflicted, all with the same shape: draft-side code from this PR against MTP-side code that landed on `main`, inside the same function but operating on disjoint parameter sets. Resolution was "keep both" in every case, not a choice between them:

- `nemo_rl/models/policy/tq_policy.py`: `_aggregate_train_results` surfaces both `mtp_metrics` and `draft_grad_norm`.
- `nemo_rl/models/policy/workers/megatron_policy_worker.py`: in `_finish_train_step_body`, the draft main-grad correction and the MTP denominator rescale both run after `scale_gradients`, and both `draft_grad_norm` and `mtp_grad_norm` get reduced across the model-parallel group.
- `tests/unit/models/policy/test_split_api_wrappers.py`: split back into `test_finish_surfaces_draft_grad_norm` and `test_finish_propagates_mtp_metrics`.
- `tests/unit/models/policy/test_megatron_split_state.py`: both suites preserved.

One thing worth calling out for review: `test_without_draft_payload_still_reduces_draft_grad_norm` asserts `len(reduce_calls) == 3`, and the merge adds a fourth `reduce_max_stat_across_model_parallel_group` call. The count still holds because the new call is gated on `state["mtp_enabled"]`, which `_setup_open_step` leaves false, so the assertion is still testing what it was written to test rather than passing by accident.

No reviewer comments outstanding on this one; `ruff@0.9.9 check` and `format --check` are clean.

seonjinn:
/ok to test [fb138a7](https://github.com/NVIDIA-NeMo/RL/pull/3706/commits/fb138a77f22548c06ea6601392a76fefce689702)

terrykong (dismissed):

seonjinn:
/ok to test [6d5c106](https://github.com/NVIDIA-NeMo/RL/pull/3706/commits/6d5c106be00f90dc37db2f6f572ee6986411a8a8)

seonjinn:
/ok to test [53a9e12](https://github.com/NVIDIA-NeMo/RL/pull/3706/commits/53a9e122e507abfebe7e4cd054a0ebef7cd3df78)

seonjinn:
/ok to test [0cbe19c](https://github.com/NVIDIA-NeMo/RL/pull/3706/commits/0cbe19cb373579f98eaec2a57bc5b31fca9d9550)

seonjinn:
/ok to test [0cbe19c](https://github.com/NVIDIA-NeMo/RL/pull/3706/commits/0cbe19cb373579f98eaec2a57bc5b31fca9d9550)

terrykong (approved):

PR #3707, opened 2026-08-19
feat(draft): support optimizer schedule overrides

seonjinn:
## Summary

- Adds opt-in learning-rate, minimum-LR, and weight-decay overrides for draft parameters.
- Applies overrides only to tagged draft parameters.
- Preserves existing optimizer behavior when overrides are unset or disabled.

## Scope

Draft normalization and bias parameters use zero weight decay. Target optimizer groups and the global scheduler contract remain unchanged.

## Validation

- Exact head: `f658f69b972732fe28596959e8dd6dc0829b66fc`
- 27 current-head GitHub checks pass; remaining unit jobs are queued.

## Focused evidence

- Real optimizer/scheduler trajectories and deterministic draft-update tests passed; latest-main composition passed config, optimizer, Ruff, format, and Pyrefly checks.

## Dependencies / Remaining

Depends on #3701. The branch-up-to-date check is the only current failure.


seonjinn:

Reviewed the PR4-only range on top of #3701, including config validation, parameter selection, Megatron Bridge override semantics, disabled/default behavior, checkpoint surface, and tests.

No unresolved correctness, security, design, or test findings remain. The final branch contains two signed-off commits and only six PR4 files.

Validation: OCI-Hsg job 6300114 completed successfully with 8 config/setup tests, 6 MCore provider/setup tests, Ruff 0.9.9, format, and configured Pyrefly with zero errors.

nemo_rl/models/megatron/draft/optimizer.py
```diff
@@ -0,0 +1,87 @@
...
+            minimum_lr = context.optimizer_config.min_lr
+        if minimum_lr is not None and minimum_lr > self.draft_optimizer.lr:
+            raise ValueError(
+                "draft optimizer lr must be at least the inherited optimizer min_lr"
+            )
+
+        draft_override = ParamGroupOverride(max_lr=self.draft_optimizer.lr)
+        if minimum_lr is not None:
+            draft_override["min_lr"] = minimum_lr
+        if self.draft_optimizer.weight_decay is not None:
```
seonjinn, line 71: The draft `weight_decay` override composes with Megatron's standard overrides, which still set `wd_mult=0.0` for norm/bias/1-D params -- and the scheduler applies `weight_decay = get_wd(group) * wd_mult`, so a user-configured draft decay silently stays zero for draft norms and biases. That matches the main model's convention (arguably working as designed), but the config surface promised "draft-only ... weight_decay" with no caveat and no test covered a draft param matching the standard predicate.

Follow-up commit adds the explaining comment here and a test that pins the composition: a 1-D draft bias gets both `start_wd=0.02` and `wd_mult=0.0`, netting zero effective decay.

seonjinn (commented):

Method: a focused review agent (with anchors verified against actual file reads) cross-checking every scheduler/override mechanic against the vendored Megatron-LM source; the real-scheduler LR trajectory in the test suite was independently recomputed and matches.

Verified clean: parameter tagging is attribute-based over the whole draft tree (provider-agnostic -- naming conventions from the eagle-ttt/dspark provider PRs cannot miss it); the unset/disabled path is traced to byte-identical optimizer behavior (no extra param group, so pre-PR checkpoints resume unchanged); warmup/decay stay on the single shared step counter; and the min-LR inversion guard is present and tested.

One finding (inline, fixed in the follow-up commit): the weight-decay override silently composes to zero for draft norm/bias params via Megatron's standard `wd_mult=0.0` -- now documented and pinned by a test.

**Process:** rebase onto #3701's current head (`8774f1d3d` added a near-miss config-typo validator to `Eagle3DraftConfig`, which this PR extends -- no conflict expected). No unit/L0 CI has run on this exact head; posting `/ok to test` on the post-fix head.

seonjinn:
CI triage: the 13 L0 shard failures were one collection error — this test file imports `megatron.bridge` at module level, which breaks pytest COLLECTION on every non-mcore shard (the `mcore` mark only deselects at run time). First real CI run on this PR surfaced it. Fixed on the new head with `pytest.importorskip` before the megatron imports.

seonjinn (commented):
## Self-review

Multi-agent review of this branch merged with latest `main` (`8d1ca372f`), followed by an adversarial pass. The adversarial pass killed the only HIGH the first pass produced, so the retractions below are the more useful half of this report.

### Actionable

**1. Merge staleness: the packed-sequence branch still subscripts the draft config.** This PR converts the draft config to a pydantic model, and at the PR head there was one draft call site in `nemo_rl/models/megatron/train.py`, correctly converted. `main` has since added a second, in the packed-sequence branch, which still does `self.cfg["draft"]["loss_weight"]`. Subscripting the model raises `TypeError: 'Eagle3DraftConfig' object is not subscriptable`. This is not a dead branch — packing is a supported draft configuration (only packed + PP>1, CP>1 and fused-linear-logprobs are rejected), so packed draft co-training dies at loss-wrapper construction after this merge. Nothing in CI constructs `DraftLossWrapper`, so it is invisible to the suite.

A sweep of merge-base vs PR head for every construct this PR migrated found exactly two such sites, both now fixed on the branch; there is no third.

**2. `Policy.__init__` assumes the caller already typed the config.** `PolicyConfig` is a `TypedDict`, i.e. a plain dict at runtime, so `config.get("draft")` is not guaranteed to be an `Eagle3DraftConfig` — every caller that builds a policy config in-process passes a mapping. `main`'s own guard tests fail against the merged tree with `AttributeError: 'dict' object has no attribute 'enabled'`. Fixed on the branch with a `coerce_draft_config` at the boundary plus tests that pin the mapping, model, and `None` inputs.

**3. The only production wiring line for this feature has no test.** `nemo_rl/models/megatron/setup.py:2029` passes `optimizer_config_override_provider=build_draft_optimizer_override_provider(policy_cfg.get("draft"))`, and that kwarg appears exactly once in the repo. The provider itself is well covered — `tests/unit/models/megatron/test_draft_optimizer.py` constructs it directly, exercises real `ParamKey.matches`, real `combine_param_group_overrides`, and asserts a real `OptimizerParamScheduler` LR trajectory. But no test goes through `setup_optimizer`, so deleting line 2029 silently disables the entire feature with a green suite. Worth one test asserting the provider arrives non-`None` when `draft.optimizer` is configured.

### Nit

`_is_draft_parameter` (`nemo_rl/models/megatron/draft/optimizer.py:36`) compares `grad_norm_group` against the string literal `"draft"` while `DRAFT_GRAD_NORM_GROUP = "draft"` (`draft/utils.py:1238`) is the constant stamped onto the parameter at `:1407`. The values agree today, so this is not a bug. Both modules already import from `draft_config`, so there is no cycle blocking the constant.

The guide at `docs/guides/eagle3-speculative-decoding.md:48-51` documents the `draft:` block without the new `optimizer:` key — the string `optimizer` does not appear in that file at all.

**"Draft and policy param groups collide in `param_group_identifier_keys`, so a checkpoint round trip silently trains the drafter at the policy LR."** This was filed as HIGH and it is wrong. `param_group_identifier_keys` is a 9-tuple that includes `max_lr` and `min_lr`:

```python
param_group_identifier_keys = (
    *_param_group_override_keys(),
    'lr_mult',
    'is_expert_parallel',
    'is_decoupled_lr',
)
```

where `_param_group_override_keys()` returns `tuple(sorted(ParamGroupOverride.__annotations__.keys()))` = `max_lr, min_lr, start_wd, end_wd, wd_mult, optimizer`. Read at Megatron-LM `7c9c3a027c503ae9ae1e8ad7b14397abb8269378`, `megatron/core/optimizer/optimizer.py:139-149` — resolved through this branch's own submodule chain (Megatron-Bridge `b11414c71b15e54d333eb49346ed199f20fa9021` → Megatron-LM `7c9c3a02`), not from a nearby rev. The 4-tuple that would produce this bug exists at no revision reachable from this PR.

A narrower collision is mechanically real — with `draft.optimizer.lr` equal to the policy `lr` and both `min_lr` and `weight_decay` omitted, the draft group and the policy default group do land on identical tuples. It still cannot corrupt anything: the scheduler reads only `start_wd`, `end_wd`, `max_lr`, `min_lr`, `wd_mult`, and all five are identifier keys, so any two colliding groups are schedule-identical by construction and swapping their state changes no LR or WD. The configuration is also self-defeating — it reproduces the default schedule exactly, and setting `weight_decay` to differentiate it emits `start_wd`/`end_wd` and breaks the tie.

Related: the comment at `optimizer.py:628-633` describing silent clobbering on duplicate identifiers is upstream documenting a bug class they **closed** by widening the tuple, with a regression test at `tests/unit_tests/optimizer/test_param_group_identifier_keys.py:122`. It is not evidence of a live bug.

**"MuP overrides conflict with the draft `max_lr` override."** Mechanism is real and the failure would be a loud `ValueError` at optimizer construction, but MuP is gated on `use_mup`, and `use_mup` / `mup_width_mult` appear nowhere in `nemo_rl/` or `examples/`. Unreachable via any supported config.

### Verified clean

The provider grants overrides only to parameters carrying the draft grad-norm group, so policy parameters cannot inherit draft scheduler state. `_get_param_groups` having zero repo-wide references is expected — it is megatron-internal, reached via `get_megatron_optimizer`.

### Method note

Nothing in the adversarial pass was executed: `torch` is not importable in the review environment, so the checkpoint analysis above is read from pinned source objects (`git show <sha>:<path>`) rather than observed at runtime. The param-group save/load round trip is owned and tested upstream.

seonjinn (commented):

Re-merged this branch with `main` at `4213e0867`. Clean merge, no conflicts. `main` has advanced 109 files (+4118 / -1387) since the previous review base (`8d1ca372f`), and shares six files with this PR, but none of its hunks carry a `draft` line — `git diff 8d1ca372f 4213e0867` filtered on `^[+-].*draft` across `setup.py`, `policy/__init__.py` and `megatron_policy_worker.py` is empty. `ruff 0.9.9 check` and `format --check` clean on the merged tree (737 files); CI green.

The previous review stands. One citation to refresh: the single production wiring line has moved to `setup.py:2065` after the merge (it was `:2029`), and it is still the only occurrence of `optimizer_config_override_provider` in the repo, so the untested-wiring point from that review is unchanged.

### `main`'s new optimizer dtype resolution does not reach this path — checked

`main` added `_resolve_optimizer_dtype_kwargs` (`setup.py:263-295`) and routed `config["megatron_cfg"]["optimizer"]` through it in `_create_megatron_config`, turning YAML dtype strings (`fp8`, `bf16`, ...) into `torch` dtypes, with `exp_avg_dtype` / `exp_avg_sq_dtype` / `store_param_remainders` newly declared on `MegatronOptimizerConfig`.

That is adjacent to this PR but does not interact with it. The draft path contributes `ParamGroupOverride` entries, whose keys are `max_lr`, `min_lr`, `start_wd`, `end_wd`, `wd_mult` and `optimizer` — no dtype keys — and `DraftOptimizerConfig` declares only `lr`, `min_lr` and `weight_decay`. There is no route by which a draft override reaches the dtype resolver, and no draft-side equivalent is needed: precision-aware moments are a whole-optimizer setting, not a per-param-group one.

Worth noting the asymmetry it highlights, though: `DraftOptimizerConfig` is `extra="forbid"` (`draft_config.py:21`) while `Eagle3DraftConfig` is `extra="allow"`. That is the right split — the optimizer block is new surface with no legacy to preserve — but it means a misspelled key under `draft.optimizer` is rejected while the same misspelling one level up is not.

### One new nit: the `min_lr` error message hides the fallback

`build_config_overrides` (`optimizer.py:51-57`) falls back to the policy's `min_lr` when `draft.optimizer.min_lr` is unset, then raises if that inherited floor exceeds the draft LR:

```
draft optimizer lr must be at least the inherited optimizer min_lr
```

The mechanism is right — a group whose `max_lr` sits below its `min_lr` is nonsense, and failing at construction is better than a silently clamped schedule. The message is what I would change. A user who wants a *lower* draft LR than the policy (plausible for a small drafter on a decayed policy schedule) sets `draft.optimizer.lr: 1e-6` under a policy `min_lr: 1e-5`, gets this error, and has nothing pointing at the fix, which is to set `draft.optimizer.min_lr` explicitly. The message does not mention that the floor was inherited rather than configured, or name either value.

Something like `draft optimizer lr ({lr}) is below the inherited min_lr ({minimum_lr}) taken from the policy optimizer; set policy.draft.optimizer.min_lr to decouple the draft schedule` would make the failure self-resolving.

### Method note

Lint only on this pass. `torch` and `megatron.core` are not importable in this environment, so the merge analysis and the dtype-path check are read from the merged tree and from `main`'s diff, not executed.

nemo_rl/models/megatron/draft/optimizer.py
```diff
@@ -0,0 +1,92 @@
...
+
+class DraftOptimizerConfigOwner(Protocol):
+    """Draft config fields required to opt in to optimizer overrides."""
+
+    enabled: bool
+    optimizer: DraftOptimizerConfig | None
+
+
+def _is_draft_parameter(parameter: torch.nn.Parameter) -> bool:
+    return getattr(parameter, "grad_norm_group", None) == "draft"
```
yuekaizhang: This re-hardcodes the literal `"draft"` instead of importing `DRAFT_GRAD_NORM_GROUP` from `nemo_rl.models.megatron.draft.utils` (the single source of truth used when tagging params via `param.grad_norm_group = DRAFT_GRAD_NORM_GROUP`). No import cycle blocks this. If the tag value ever changes, this comparison would silently stop matching and draft params would silently fall back to the target-model optimizer schedule instead of raising — worth importing the constant instead.

yuekaizhang (commented):
Cross-checked the override mechanism against the actual Megatron-Bridge/Megatron-LM source for parameter-group tagging leakage, the "no-op when unset" claim, the zero-weight-decay carve-out, and checkpoint-resume interaction — all four check out correctly. One maintainability point below.

nemo_rl/models/megatron/draft/optimizer.py
seonjinn: Fixed in 067b71491. `_is_draft_parameter` now imports `DRAFT_GRAD_NORM_GROUP` from `nemo_rl.models.megatron.draft.utils` and compares against it; the four literals in `tests/unit/models/megatron/test_draft_optimizer.py` were switched over too, so the test would no longer paper over a divergence between tag and predicate.

Confirmed there is no cycle: `draft/utils.py` has no module-level import of `draft/optimizer.py`, and `draft/__init__.py` does not pull `optimizer` in, so `optimizer -> utils` is a plain one-way edge.

The failure mode you describe is the right one to design against. The predicate is the only thing standing between the draft params and the policy schedule, and a miss is silent by construction: `build_config_overrides` would return no draft entry, the params would fall through to the target-model group, and training would carry on at the policy LR with nothing logged.

Two smaller things in the same commit, since they sit in the code path this comment points at:

- The inherited-`min_lr` error now names the knob that decouples the schedule (`policy.draft.optimizer.min_lr`) instead of just reporting a bound the user never set.
- Added coverage for the non-inherited branch of that guard. `DraftOptimizerConfig` already rejects `min_lr > lr` at validation, so the branch is only reachable via `model_construct`; the test says so in its docstring rather than implying it is a normal path.

Latest `main` is merged in as well (behind=0).

seonjinn:
/ok to test [067b714](https://github.com/NVIDIA-NeMo/RL/pull/3707/commits/067b714915122dcbd98553cf7cc200030a39f811)

seonjinn:
/ok to test [841a816](https://github.com/NVIDIA-NeMo/RL/pull/3707/commits/841a8161093bcae4505dd4a4e8b6fddf0e24c555)

terrykong (dismissed):

nemo_rl/models/megatron/draft/optimizer.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def build_config_overrides(
+        self, context: OptimizerConfigOverrideProviderContext
+    ) -> dict[ParamKey, ParamGroupOverride] | None:
+        """Combine standard Megatron groups with the opt-in draft schedule."""
+        overrides = super().build_config_overrides(context) or {}
+        minimum_lr = self.draft_optimizer.min_lr
+        if minimum_lr is None:
+            minimum_lr = context.optimizer_config.min_lr
+        if minimum_lr is not None and minimum_lr > self.draft_optimizer.lr:
```
terrykong: **1 action item.**

**TL;DR — this check catches `min_lr > lr` but nothing checks `lr` against the scheduler's `lr_warmup_init`, and the error message steers you into that second case: do what it says and the draft silently starts warmup *above* its own peak and ramps down.**

The check at [optimizer.py#L52-L66](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/optimizer.py#L52-L66) covers one bound and not the other. Warmup ramps every group from the scheduler-wide `init_lr` up to that group's own `max_lr` ([optimizer_param_scheduler.py#L225-L233](https://github.com/NVIDIA/Megatron-LM/blob/1e7598cbfae888cdd3d741a351aae588d56f66c0/megatron/core/optimizer_param_scheduler.py#L225-L233)), and Bridge wires `init_lr=scheduler_config.lr_warmup_init` ([optim.py#L441](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/5ed97996cc2b422904d18179375b6d7366915097/src/megatron/bridge/training/optim.py#L441)). Megatron's own `assert self.init_lr <= self.max_lr` ([optimizer_param_scheduler.py#L152](https://github.com/NVIDIA/Megatron-LM/blob/1e7598cbfae888cdd3d741a351aae588d56f66c0/megatron/core/optimizer_param_scheduler.py#L152)) only ever sees the scheduler's own `max_lr` — there is no per-group version of it. The draft override can only carry the peak LR, the floor LR and the weight decay ([`ParamGroupOverride`](https://github.com/NVIDIA/Megatron-LM/blob/1e7598cbfae888cdd3d741a351aae588d56f66c0/megatron/core/optimizer_param_scheduler.py#L18-L37)), so nothing downstream catches it either.

What makes this worth fixing is that the existing error text recommends the step that opens the hole. [L60-L64](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/optimizer.py#L60-L64) says ", which was inherited from the policy optimizer; set `policy.draft.optimizer.min_lr` to decouple the draft schedule". So it takes two guided steps:

1. Set a gentler draft `lr` and leave `min_lr` unset → it inherits the policy's `min_lr` and this raises.
2. Follow the message and set a small `min_lr` → it passes, and the draft now warms down from above its own peak.

Against [`examples/configs/grpo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/examples/configs/grpo_math_1B.yaml#L313-L314) (`lr_warmup_iters: 13`, `lr_warmup_init: 5.0e-7`), step 2 lands here:

```yaml
  draft:
    enabled: true
    optimizer:
      lr: 1.0e-7
      min_lr: 1.0e-8
```

1. The existing check passes — `1e-8 > 1e-7` is false.
2. Warmup step n gives `5e-7 + (1e-7 - 5e-7)*n/13`, so the draft starts at 5.0e-7 — 5x its configured peak — and ramps *down* to 1e-7 by step 13, while the policy group ramps *up* from 5e-7 to 5e-6.
3. The warmup branch returns before any `min_lr` clamp, so the draft LR is unclamped through those steps.
4. `lr_decay_style: "constant"` then pins the draft at 1e-7. No error, no warning.

The likelier mistake is already caught, and no shipped config reaches this by inheritance — of the 26 that set `lr_warmup_init`, none has `min_lr < lr_warmup_init`, so an unset `min_lr` always trips the existing raise first.

### AI-1

Add the warmup case directly after the existing `min_lr` check, in this same function. Warn rather than raise: the post-warmup schedule is exactly what the user asked for, so raising would reject a config whose steady state is correct. [`warnings.warn` is already the convention next door in `setup.py`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/setup.py#L939).

Not a committable `suggestion` block: it also needs `import warnings` at the top of the file, so applying the block alone would not compile.

```python
        # Warmup ramps every group from the scheduler-wide lr_warmup_init up to
        # that group's max_lr, so a draft peak below lr_warmup_init warms DOWN
        # from above its own target. Megatron asserts init_lr <= max_lr for the
        # scheduler's own max_lr only; there is no per-group equivalent.
        warmup_steps = context.scheduler_config.lr_warmup_steps or 0
        warmup_init = context.scheduler_config.lr_warmup_init
        if warmup_steps > 0 and warmup_init > self.draft_optimizer.lr:
            warnings.warn(
                f"draft optimizer lr ({self.draft_optimizer.lr}) is below the policy "
                f"schedule's lr_warmup_init ({warmup_init}), so the draft group starts "
                f"warmup above its own peak learning rate and ramps down for the first "
                f"{warmup_steps} steps. Lower "
                f"policy.megatron_cfg.scheduler.lr_warmup_init, or raise "
                f"policy.draft.optimizer.lr.",
                stacklevel=2,
            )
```

`or 0` is needed because `lr_warmup_steps` defaults to `None` and `None > 0` raises `TypeError`. `lr_warmup_init` is a plain `float = 0.0` and never `None`, so it needs no guard, and strict `>` is right because equal values are a flat warmup rather than a warm-down. The `warmup_steps > 0` gate mirrors [`get_lr`'s own test](https://github.com/NVIDIA/Megatron-LM/blob/1e7598cbfae888cdd3d741a351aae588d56f66c0/megatron/core/optimizer_param_scheduler.py#L230).

This blocks nothing legitimate: `lr_warmup_init` defaults to 0.0 and `DraftOptimizerConfig.lr` must be `> 0`, so the warning is inert unless someone set a nonzero warmup init above the draft LR.

<details><summary>Why the warmup length is only known here</summary>

`lr_warmup_steps` cannot be read from config — Megatron declares it [`lr_warmup_steps: int | None = field(init=False, default=None)`](https://github.com/NVIDIA/Megatron-LM/blob/1e7598cbfae888cdd3d741a351aae588d56f66c0/megatron/training/config/training_config.py#L204), and only Bridge's `_calculate_scheduler_steps` fills it in ([config.py#L1783-L1844](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/5ed97996cc2b422904d18179375b6d7366915097/src/megatron/bridge/training/config.py#L1783-L1844)), resolving three mutually exclusive spellings and scaling by global batch size. That is why the check reads it off `context` here instead of earlier at config-validation time.
</details>

examples/configs/grpo_math_1B.yaml
```diff
@@ -336,6 +336,9 @@ policy:
     loss_weight: 0.1
     num_layers: null
     aux_layer_indices: null
+    # Optional draft-only optimizer schedule. Set a mapping with `lr` and
+    # optional `min_lr`/`weight_decay` to enable it.
+    optimizer: null
```
terrykong: **1 action item.**

**TL;DR — the Eagle3 guide never mentions `policy.draft.optimizer`, so the two rules that decide what a draft schedule actually does — an unset `min_lr` inherits the policy's as an absolute floor, and warmup/decay shape cannot be set per group — live only inside an error string.**

Anchored on this new `optimizer: null` line because the guide is not in this PR's diff.

The guide's "Important Knobs" list at [eagle3-speculative-decoding.md#L156-L164](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/docs/guides/eagle3-speculative-decoding.md#L156-L164) documents `policy.draft.enabled`, `.speculator_type`, `.model_name` and `.loss_weight`, but not `.optimizer` — `grep -c optimizer` on that file returns 0, and the example `draft:` block at [L48](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/docs/guides/eagle3-speculative-decoding.md#L48) omits it too. The list does carry `speculator_type`, added by the immediately-preceding PR #3701, so the guide was current one PR ago. (`num_layers` and `aux_layer_indices` are missing from it as well — the list was already incomplete, so this is "please add the new knob", not "you forgot".)

The missing bullet matters less than the two rules that have no home outside the raise:

1. Leaving `draft.optimizer.min_lr` unset inherits `policy.megatron_cfg.optimizer.min_lr` as an **absolute floor**, not as a preserved `lr`→`min_lr` ratio ([optimizer.py#L52-L54](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/optimizer.py#L52-L54)). With this config's policy `lr: 5.0e-6` / `min_lr: 5.0e-7` (a 10x decay), a draft `lr: 1.0e-6` decays only 2x.
2. The only parts of the schedule this config can set per group are the peak LR, the floor LR and the weight decay — those are what the draft override writes into [`ParamGroupOverride`](https://github.com/NVIDIA/Megatron-LM/blob/1e7598cbfae888cdd3d741a351aae588d56f66c0/megatron/core/optimizer_param_scheduler.py#L18-L37). Warmup length, warmup starting LR, decay length and decay style are shared with the policy schedule and cannot be set per group.

Worth one line in the same bullet: `weight_decay` reaches draft weight matrices only — draft norms and biases keep zero decay, because Megatron's standard `wd_mult=0.0` group still applies. The code comment at [optimizer.py#L72-L78](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/optimizer.py#L72-L78) already says this correctly; it just isn't in the guide.

### AI-1

Add `policy.draft.optimizer` to the guide's Important Knobs list, with those rules. Not a committable `suggestion` block — the target file is not in this PR's diff.

```markdown
- `policy.draft.optimizer` — optional per-group LR / weight-decay schedule for the draft
  parameters. `null` (the default) trains the draft on the policy's own schedule.
  - `lr` / `min_lr`: the draft's peak and floor LR. If `min_lr` is unset it inherits
    `policy.megatron_cfg.optimizer.min_lr` as an absolute floor, not as a ratio — so a
    policy `lr: 5.0e-6` / `min_lr: 5.0e-7` (10x decay) with a draft `lr: 1.0e-6` gives
    the draft only a 2x decay.
  - `weight_decay`: applies to draft weight matrices only. Draft norms and biases keep
    zero decay, same as the main model.
  - Only the peak LR, floor LR and weight decay can be set per group. Warmup length,
    warmup starting LR, decay length and decay style are shared with the policy schedule.
```

nemo_rl/models/megatron/draft/optimizer.py
```diff
@@ -0,0 +1,101 @@
...
+                predicate=ParamPredicate(
+                    name="draft_parameter",
+                    fn=_is_draft_parameter,
+                )
+            )
+        ] = draft_override
+        return overrides
+
+
+def build_draft_optimizer_override_provider(
```
terrykong: **1 action item.**

**TL;DR — `build_draft_optimizer_override_provider` reads `.enabled` / `.optimizer` straight off its argument, while its two siblings both call `coerce_draft_config` first, so this one function relies on `Policy.__init__` having normalized the config instead of enforcing it itself.**

The sibling's docstring states the contract outright ([draft/training.py#L76-L91](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/training.py#L76-L91)):

> `PolicyConfig` is a TypedDict, so `policy_cfg["draft"]` is only an `Eagle3DraftConfig` because `Policy.__init__` normalizes it in place. Coerce here as well so this stays correct for callers that build a `PolicyConfig` by hand.

Two calls, same file, same argument:

- [setup.py#L1904](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/setup.py#L1904): `resolve_draft_speculator(policy_cfg.get("draft"))` → calls `coerce_draft_config`
- [setup.py#L2362](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/setup.py#L2362): `build_draft_optimizer_override_provider(policy_cfg.get("draft"))` → raw attribute access

The third sibling, [`draft_refit_enabled`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/policy/draft_config.py#L76-L81), coerces too.

The [`DraftOptimizerConfigOwner` Protocol](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/optimizer.py#L30-L34) exists only to type that raw access. It has one implementer ever, is not `@runtime_checkable`, and fronts no dispatch — unlike [`DraftSpeculator`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/training.py#L34-L46), which fronts the [`_SPECULATOR_FACTORIES`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/training.py#L69-L71) registry.

### AI-1

Take `Eagle3DraftConfig | Mapping[str, Any] | None`, coerce first, and delete the Protocol:

```python
def build_draft_optimizer_override_provider(
    draft_config: Eagle3DraftConfig | Mapping[str, Any] | None,
) -> DraftOptimizerConfigOverrideProvider | None:
    """Build an override provider only when a draft schedule is configured."""
    coerced = coerce_draft_config(draft_config)
    if coerced is None or not coerced.enabled or coerced.optimizer is None:
        return None
    return DraftOptimizerConfigOverrideProvider(coerced.optimizer)
```

Not a committable `suggestion` block: it touches three separate regions — the [import](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/optimizer.py#L27), the [Protocol deletion](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/draft/optimizer.py#L30-L34), and this function. Applying the block alone would leave the now-unused `Protocol` import behind.

**Context — no action.** This is a "match the sibling" ask, not a crash report. [setup.py#L2161](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/setup.py#L2161) already does `policy_cfg["draft"].enabled` about 200 lines earlier in the same function with no early return in between, and that line predates this PR — a raw dict fails there first. Also: the fix is to delete the Protocol, not to make it `@runtime_checkable` or give it a second implementer.

tests/unit/models/megatron/test_draft_optimizer.py
```diff
@@ -0,0 +1,282 @@
...
+    ]
+
+
+def test_draft_lr_override_produces_finite_stable_toy_updates() -> None:
+    def run(*, draft_lr: float | None) -> tuple[list[float], float, float]:
+        policy_parameter = torch.nn.Parameter(torch.tensor([1.0]))
+        draft_parameter = torch.nn.Parameter(torch.tensor([1.0]))
+        draft_group: dict[str, object] = {"params": [draft_parameter]}
+        if draft_lr is not None:
+            draft_group.update({"max_lr": draft_lr, "min_lr": draft_lr / 10})
```
terrykong: **1 action item.**

**TL;DR — [`test_draft_lr_override_produces_finite_stable_toy_updates`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/tests/unit/models/megatron/test_draft_optimizer.py#L253-L282) hand-writes the override into the param group instead of building it, so it never constructs `DraftOptimizerConfigOverrideProvider` and no bug in this PR can make it fail.**

The group is written directly on this line:

```python
draft_group.update({"max_lr": draft_lr, "min_lr": draft_lr / 10})
```

That leaves only `torch.optim.SGD` and Megatron's `OptimizerParamScheduler` under test — both upstream. Compare [`test_real_scheduler_applies_distinct_draft_lr_trajectory`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/tests/unit/models/megatron/test_draft_optimizer.py#L226-L250), which builds the group from the provider through [`_combined_overrides_for_parameter`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/tests/unit/models/megatron/test_draft_optimizer.py#L46-L60) — that is the right shape.

### AI-1

Build the draft group from the provider, or delete the test — as written it asserts upstream behaviour. Not a `suggestion` block: the suite could not be run on the reviewing machine (no Linux/GPU environment), so this is a sketch rather than a committable edit. One way to do it:

```python
draft_group: dict[str, object] = {"params": [draft_parameter]}
if draft_lr is not None:
    draft_parameter.grad_norm_group = DRAFT_GRAD_NORM_GROUP
    provider = DraftOptimizerConfigOverrideProvider(
        DraftOptimizerConfig(lr=draft_lr, min_lr=draft_lr / 10)
    )
    draft_group.update(
        _combined_overrides_for_parameter(provider, draft_parameter, "draft.weight")
    )
```

nemo_rl/models/megatron/setup.py
```diff
@@ -2356,6 +2359,9 @@ def composed_peft_hook(model: list[MegatronModule]) -> list[MegatronModule]:
             scheduler_config=megatron_cfg.scheduler,
             model=model,
             use_gloo_process_groups=megatron_cfg.dist.use_gloo_process_groups,
+            optimizer_config_override_provider=build_draft_optimizer_override_provider(
```
terrykong: **1 action item — low, a note rather than a must-add.**

**TL;DR — this line is the only thing connecting the provider to training, and no test touches it, so deleting it turns the whole feature off with a green suite.**

`optimizer_config_override_provider` appears exactly once under `nemo_rl/` — [here](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/setup.py#L2362) — and zero times under `tests/`. The provider itself is well covered by [test_draft_optimizer.py](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/tests/unit/models/megatron/test_draft_optimizer.py#L100-L282); the untested part is only the hand-off.

The honest cost: a real test needs `get_model`, `setup_optimizer` and `ProcessGroupCollection` all patched, and [test_megatron_setup.py](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/tests/unit/models/megatron/test_megatron_setup.py#L1-L40) is pure-config today — it never calls the model-building path. So this is worth doing only if it turns out cheap.

### AI-1

If the scaffolding is cheap enough, patch [`setup_optimizer`](https://github.com/NVIDIA-NeMo/RL/blob/841a8161093bcae4505dd4a4e8b6fddf0e24c555/nemo_rl/models/megatron/setup.py#L2357) and assert it receives a `DraftOptimizerConfigOverrideProvider` when `policy.draft.optimizer` is set, and `None` when it is not. Not a `suggestion` block — it is a new test plus scaffolding, not an edit to this line.

seonjinn:
/ok to test [67307b6](https://github.com/NVIDIA-NeMo/RL/pull/3707/commits/67307b651e35586a3d00f7370818e6f415ad7654)

seonjinn:
/ok to test [bbb7ebc](https://github.com/NVIDIA-NeMo/RL/pull/3707/commits/bbb7ebc5501c97bb11ecbcecae3f1b9b7597a1ad)

terrykong (approved):
approved, but @kajalj22 to confirm the cicd-main.yml change

PR #3709, opened 2026-08-19
fix: non-contiguous top-p logits and null keep_top_k

mferrato:
# What does this PR do ?

Fixes two correctness issues in policy scoring and checkpoint configuration: non-contiguous top-p logits are flattened safely, and `checkpointing.keep_top_k` accepts `null` as documented.

# Issues

No linked issue.

# Usage

No API changes are required:

- Top-p policy scoring accepts non-contiguous `[..., vocab_size]` logits.
- `checkpointing.keep_top_k: null` validates successfully and retains all periodic checkpoints.

**Pre checks**:
- [x] Did you write any new necessary tests? Added regression coverage for non-contiguous top-p logits; existing checkpoint coverage verifies that `keep_top_k=None` retains all checkpoints.
- [x] Did you run the unit tests and functional tests locally? Focused unit tests were run locally, and repository `CI:L1` was requested for the expanded scope. Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [x] Did you add or update any necessary documentation? No documentation update is required because these changes correct existing behavior without changing the public API. Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information

The logits regression uses a genuinely non-contiguous `(2, 5, 17)` tensor and checks `top_k=None` and `top_k=5` against a contiguous reference with exact equality. The focused CPU cases passed locally, and the same failure and fix were validated separately on an NVIDIA H100.

The checkpoint runtime already implements and tests the documented `None`-keeps-all behavior; this change aligns the configuration type with that behavior so YAML `null` is accepted.

Ruff formatting, lint checks, and Python syntax compilation passed for the original focused change. Repository `CI:L1` status is tracked in the PR checks.


yuekaizhang:
The fix itself looks right: `reshape` at the flatten site is exactly what's needed (the top-p scoring path receives strided `[..., :-1, :]` logits slices in multi-sequence batches), and the regression test covers the non-contiguous case. The remaining `.view(original_shape)` calls at `logits_sampling_utils.py:207-208` are fine as-is — `filtered_logits`/`keep_mask` are freshly allocated with `torch.empty(...)` so they are always contiguous.

This crash is one of three NeMo-RL items from a customer handoff on Nemotron 3.5 Nano RLVR training (the run that surfaced this bug). The other two are still present on `main` — since you're in this area, could you pick them up here or in follow-ups?

1. **Refit silently rounds FP32 router-bias buffers to BF16** — `nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py:1088`: `prepare_refit_info` reports `(shape, self.dtype)` for every tensor, and the streaming paths cast everything to `self.dtype` (see the comment at `:1083` and `dtensor_params_generator(self.model, self.dtype)` at `:1128`/`:1167`). For NemotronH MoE checkpoints this rounds the 23 FP32 `e_score_correction_bias` routing buffers to BF16 on every refit; the customer's diagnostics identified these as exactly the tensors diverging across the policy→vLLM boundary (expert selection flips; they observed gen KL ~109–135 with all-zero rewards until they patched the dtype through the refit metadata and payload). Suggested fix: report each tensor's native dtype in `prepare_refit_info` and skip the cast for buffers whose dtype differs from the model compute dtype — the vLLM-side manifest already carries per-tensor dtype.
2. **`checkpointing.keep_top_k` rejects `null` while the docstring promises `None` semantics** — `nemo_rl/utils/checkpoint.py:126` declares `keep_top_k: NotRequired[int]`, but the docstring at `:108` says "If None, all checkpoints are kept". Suggested fix: `NotRequired[int | None]` plus a keep-all branch, mirroring `ft_keep_latest_k: NotRequired[int | None]` a few lines below.

Happy to review either change; item 1 has an Automodel-side counterpart (keeping the buffers FP32 in the training model) that we can coordinate separately.

yuekaizhang:
@mferrato Following up on the earlier comment — I pushed the two companion fixes onto this branch (maintainer edits are enabled; happy to split them into a separate PR if you'd rather keep this one single-purpose):

- `15a2438d` **fix(checkpoint): allow `keep_top_k: null`** — the runtime and docstring already support None-keeps-all; only the TypedDict annotation (`NotRequired[int]`) made MasterConfig validation reject a YAML `null`. One-line type fix.
- `a697d802` **fix(refit): stream `_keep_in_fp32`-declared tensors in native dtype** — `dtensor_params_generator` blanket-cast every tensor to the policy compute dtype and `prepare_refit_info` declared `self.dtype` for every entry, rounding NemotronH's 23 fp32 `e_score_correction_bias` MoE routing buffers to bf16 on every refit. The fix honors the HF/Automodel `_keep_in_fp32_modules(_strict)` declarations: matching FQNs are streamed in their native dtype and the manifest declares that dtype, so sender bytes and receiver reconstruction stay consistent on both the IPC-ZMQ and collective transports (both unpack per-name from the manifest — `vllm_backend.py` `state_dict_info[key]`). Models without declarations see no behavior change. Unit test added next to the existing `TestDTensorParamsGenerator` cases.

All three fixes in this PR now correspond 1:1 to the NeMo-RL items from the customer RLVR handoff. Since the scope grew beyond the title, maintainers may want to retitle, e.g. `fix: non-contiguous top-p logits, keep_top_k null, and fp32-buffer refit fidelity`.

/ok to test a697d802ba5d3bbdb7e7287749e4d239ec75d2c8

yuekaizhang:
@mferrato Could you help fix the CI CD pr title error? Thanks.

yuekaizhang:
Update on the CI failures:

- **PR title length**: fixed directly — retitled to `fix: non-contiguous logits, null keep_top_k, and fp32-buffer refit dtype` (72 chars, under the 80-char limit enforced by `semantic-pull-request`). `semantic-pull-request` now passes.
- **codecov/patch** (75% < 80% target): `prepare_refit_info`'s new native-dtype branch had no direct test, only its `dtensor_params_generator` counterpart did. Added `test_prepare_refit_info_keeps_declared_buffer_native_dtype` in `9edaf9db`, mirroring the existing `test_keep_in_fp32_declared_buffer_keeps_native_dtype` test.
- **gb200_L1_Functional_Tests_Megatron_1**: failed on a HuggingFace CDN `429 Too Many Requests` during model download — unrelated to this PR's changes; should pass on retry.

/ok to test 9edaf9db3d2610884ac5a0757b0060ab019669c0

yuekaizhang:
/ok to test [4e463cd](https://github.com/NVIDIA-NeMo/RL/pull/3709/commits/4e463cd65c6e867ff9ffb99ee39b47fbf56844b1)

yuekaizhang:
Dropped the fp32-buffer refit dtype fix from this PR — #3540 already preserves the FP32 `e_score_correction_bias` MoE router buffer during DTensor refit, so this PR's more general `_keep_in_fp32_modules(_strict)` mechanism was a duplicate/overlapping fix. Reverted the corresponding commit and its test (`d12d7e2e`), retitled the PR to reflect the reduced scope: `fix: non-contiguous top-p logits and null keep_top_k`.

This PR now only contains the two independent fixes:
- Non-contiguous top-p logits in policy scoring.
- `checkpointing.keep_top_k: null` accepted by config validation.

/ok to test d12d7e2e10a97cb1b9c58c0e0e9c07622c73c50e

yuki-97 (approved):
Both code changes look right to me. The `reshape` is the correct fix at the flatten site — `next_token_logits[:, :-1]` in [`get_next_token_logprobs_from_logits`](https://github.com/NVIDIA-NeMo/RL/blob/d12d7e2e8b648d0f12c07f96eecef818888f41c0/nemo_rl/distributed/model_utils.py#L1679) is genuinely non-viewable once batch > 1, the trailing [`.view(original_shape)`](https://github.com/NVIDIA-NeMo/RL/blob/d12d7e2e8b648d0f12c07f96eecef818888f41c0/nemo_rl/algorithms/logits_sampling_utils.py#L209-L210) calls stay safe, and the same fix also covers the chunked slice at [`dtensor_policy_worker.py:1255`](https://github.com/NVIDIA-NeMo/RL/blob/d12d7e2e8b648d0f12c07f96eecef818888f41c0/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L1255). `keep_top_k: NotRequired[int | None]` matches the runtime's existing `None` branches and its own [docstring](https://github.com/NVIDIA-NeMo/RL/blob/d12d7e2e8b648d0f12c07f96eecef818888f41c0/nemo_rl/utils/checkpoint.py#L108).

One thing left over from the scope reduction: the PR description still describes the reverted third fix. "What does this PR do", "Usage", the "new necessary tests" answer, and "Additional Information" all claim `_keep_in_fp32_modules`-declared tensors keep their native dtype during refit and that a refit regression test was added — that commit was reverted in d12d7e2e. The title was updated right after the revert but the body wasn't, and the body is what lands in the squash-merge commit message.

Not a blocker for merge — could you update the description to match the two remaining fixes?

PR #3715, opened 2026-08-19
feat(draft): add projected soft CE for DFlash

seonjinn:
## Summary

- Adds vocab-parallel projected soft cross-entropy over a caller-owned target head for DFlash.
- Adds slot/position mapping, body-only export validation, and bounded tiled recomputation.
- Validates TP metadata consistently while avoiding a redundant metadata collective.

## Scope

Projected-loss and export primitives only. Trainer and model-construction integration remain outside this PR.

## Validation

- Exact head: `7700a077497942c2b43bee4341d1815f23b8b510`
- Current-head L0/L1 suites, lint, docs, container builds, and coverage uploads pass.

## Prior focused evidence

- OCI `6325973`: 18 algorithm, 10 TP2, and 6 MCore export tests passed. The 8K multi-tile case reduced peak allocation by 24%; no blanket speed claim is made.

## Dependencies / Remaining

Depends on #3703. The branch-up-to-date and patch-coverage checks still fail; the aggregate quality check reflects them. The current head is unsigned, so the next substantive update should be signed.


nemo_rl/algorithms/loss/draft.py
```diff
@@ -0,0 +1,934 @@
...
+    teacher_logits: torch.Tensor,
+    sample_rows: torch.Tensor,
+    label_positions: torch.Tensor,
+    loss_mask: torch.Tensor,
+    position_decay: float,
+    token_chunk_size: int,
+    tp_group: torch.distributed.ProcessGroup | None,
+) -> DraftLossStats:
+    """Map DFlash block slots to the live target head and indexed teacher rows."""
+    _tp_assert_projected_metadata_agreement(
```
seonjinn, line 840: `dflash_projected_vocab_parallel_soft_ce` runs `_tp_assert_projected_metadata_agreement` here, then unconditionally calls `projected_streaming_vocab_parallel_soft_ce`, which runs it again (line 679). The second pass validates only tensors that are identical to, or deterministically derived from, inputs this first pass already exact-matched -- `mask`/`bin_ids`/`weights` come from the exact-checked `loss_mask`/`sample_rows`/`label_positions` plus the header-checked `position_decay` scalar -- so it adds no safety while doubling the header `all_gather` plus broadcast/all_reduce traffic on every training step under TP.

The first check is load-bearing: it is what makes the pre-`index_select` raises symmetric across ranks. Only the inner repeat is redundant. Consider a private skip-validation entry point into the projected function for this caller.

nemo_rl/algorithms/loss/draft.py
```diff
@@ -0,0 +1,934 @@
...
+    output_weight: torch.Tensor,
+    teacher_logits: torch.Tensor,
+    sample_rows: torch.Tensor,
+    label_positions: torch.Tensor,
+    loss_mask: torch.Tensor,
+    position_decay: float,
+    token_chunk_size: int,
+    tp_group: torch.distributed.ProcessGroup | None,
+) -> DraftLossStats:
+    """Map DFlash block slots to the live target head and indexed teacher rows."""
```
seonjinn: Two contracts the trainer PR will need are tested but stated nowhere at this API surface:

1. **Slot-0 override.** This function unconditionally forces `effective_loss_mask[:, 0] = False` (line 892) -- slot 0 is the verified anchor and is always excluded regardless of what the caller passed in `loss_mask[:, 0]`. A caller setting `loss_mask[:, 0] = True` gets a silent no-op.
2. **Normalization shape.** The returned `DraftLossStats.counts` is a rank-local `(gamma,)` vector; the caller must all_reduce it across DP/CP before passing it as `normalization_counts` to `.normalized()`. That differs from the scalar `global_valid_toks` pattern the base draft-CE path uses, and nothing here says so.

One docstring paragraph covering both would save the next author a debugging session.

seonjinn (commented):

Method: six independent review agents over the diff, then an adversarial pass that re-verified every citation against the head and re-checked CI claims live. The diff-vs-main is inflated by the unmerged base PR #3703; review scope here is the 7 files / +1586 lines unique to this PR (the projected/DFlash path). Findings anchored in #3703-owned files are posted on #3703 instead.

The core math survived adversarial verification cleanly: the soft-CE algebra was hand-derived and matches the pre-existing `DistributedCrossEntropy` behavior exactly; token-axis tiling is mathematically identical to the untiled path (not an approximation); teacher/target-head gradient isolation is enforced by `.detach()` at every use site; and all TP collectives are unconditional and symmetric -- the high-rank rejection genuinely fires on all ranks before `broadcast`, verified by the TP=2 test that patches `torch.distributed.broadcast` to fail.

**Needs action before merge:**

1. **CI has not run on this head.** `304d031e` carries only DCO/copyright/secrets/semantic/submodule checks; the last `/ok to test` targeted `6a2cc27`, two commits back -- and on that SHA, Lint check and CI quality check both FAILED. The PR body's "Ruff and format checks passed on the exact head" has no corresponding check-run. Needs a fresh `/ok to test 304d031ef31477604f924d0e81c5848ac98faf24`.
2. **Label external measurements.** The OCI `6325973` test counts and the 24% peak-allocation reduction were measured on the pre-head feature stack; the PR body should say so or the numbers should be re-run on the exact head.

**Deliberate scope decisions worth one sentence each in the PR description:** no `LossFunction` wrapper for the DFlash path yet (there is a test asserting `DFlashProjectedLossFn` does not exist -- intentional, but the contract the wrapper PR must satisfy is undocumented; see the inline comment); `token_chunk_size` config plumbing deferred to the trainer PR.

**Posted on #3703 instead (owned by the base PR):** the `global_valid_toks` wiring is only mock-tested (`assert_called_once()` with no argument check), and `DistributedCrossEntropy` in `nemo_rl/distributed/model_utils.py` becomes dead code after the rewiring.

nemo_rl/models/megatron/draft/utils.py
```diff
@@ -46,6 +46,9 @@
     "pytorch_model.bin.index.json",
 ]
 _HF_SNAPSHOT_IGNORE_PATTERNS = ["*.pt", "*.pth", "*.ckpt"]
+_DFLASH_FORBIDDEN_EXPORT_COMPONENTS = frozenset(
```
yuekaizhang, line 54: `_DFLASH_FORBIDDEN_EXPORT_COMPONENTS` doesn't include `eagle_output_layer`, which is the actual component name this file uses for the copied/target-owned LM head (see `eagle_module.eagle_output_layer.weight` and `_get_draft_output_layer`). If this validator is meant to run on a native (non-HF-remapped) state dict, it will miss the exact head it's supposed to reject. Either add `eagle_output_layer` to the forbidden set or clarify (in the docstring and a test) that this only validates already-HF-mapped state dicts.

nemo_rl/algorithms/loss/loss_functions.py
```diff
@@ -72,8 +72,10 @@ class DraftCrossEntropyLossFn(LossFunction):
     def __init__(
         self,
         vocab_parallel_group: Optional[torch.distributed.ProcessGroup] = None,
+        token_chunk_size: int = 4096,
```
yuekaizhang: `token_chunk_size` defaults at the constructor and is never threaded through from the caller's config — this is the forbidden "function-parameter default" pattern from the config-conventions skill (same issue as in #3703's `DraftCrossEntropyLossFn`). If this is meant to be tunable, source it from the loss config with a documented default; if it's intentionally an internal constant, consider a module-level named constant instead of a public constructor default that looks configurable but isn't wired up.

nemo_rl/algorithms/loss/draft.py
```diff
@@ -0,0 +1,951 @@
...
+        if any(
+            tensor is not None and tensor.ndim > max_header_dimensions
+            for _, tensor in tensors
+        ):
+            raise ValueError(
+                "projected soft-CE tensors support at most four dimensions."
+            )
+        return
+
+    dtype_codes = {
```
yuekaizhang: Unlisted dtypes (e.g. `torch.complex64`, `torch.float8_e4m3fn`) all map to the same sentinel `-1` via `.get(tensor.dtype, -1)`. Two ranks passing different unlisted dtypes for the same tensor would both encode as `-1` and pass this header-agreement check undetected, only to fail later (or silently misbehave) in the actual arithmetic. Low real-world likelihood given current dtypes in play (bf16/fp32/fp16), but worth an explicit "unsupported dtype" raise instead of silently coding unknowns to a shared sentinel.

yuekaizhang (commented):
Traced the projected vocab-parallel soft-CE math end-to-end (log-sum-exp decomposition, teacher-expectation term, cached/streaming/projected backward paths) against a from-scratch derivation — exact, not an approximation. TP-metadata-agreement-before-collectives ordering correctly avoids the classic deadlock pattern. Three points below.

isomap (commented):
Reviewed the changes and existing review threads. I have no additional code comments. Full CI on the final rebased head still needs to pass before merging.

nemo_rl/models/megatron/draft/utils.py
seonjinn: Fixed in 236f359. `_DFLASH_FORBIDDEN_EXPORT_COMPONENTS` now also carries `eagle_output_layer`, so the validator rejects the copied target head on a raw Megatron/EAGLE state dict, not just an HF-remapped one. The set now reads `{lm_head, output_layer, eagle_output_layer, mask_embedding, mask_token}` and a comment states that both naming conventions must be covered because the validator runs on raw state dicts too. `tests/unit/models/megatron/test_dflash_export_contract.py` covers the native name.

nemo_rl/algorithms/loss/loss_functions.py
seonjinn: Fixed in 236f359 (comment added in e511070). `token_chunk_size` is now sourced from config rather than being a constructor default that looks configurable but isn't:

- `DEFAULT_DRAFT_TOKEN_CHUNK_SIZE` is a documented module-level constant in `nemo_rl/algorithms/loss/draft.py`, with the memory/launch-count tradeoff spelled out.
- `DraftCrossEntropyLossFn.__init__` and `DraftLossWrapper.__init__` default to that constant instead of a bare literal.
- `policy.draft.token_chunk_size` is declared on `DraftConfig` and threaded through both `DraftLossWrapper(...)` construction sites in `nemo_rl/models/megatron/train.py`.

The same wiring is applied in #3703, #3706, #3726 and #3757 so the key means the same thing in every branch that carries this code.

nemo_rl/algorithms/loss/draft.py
seonjinn: Fixed in 236f359. The hand-written dtype table is gone. Header codes are now built from every `torch.dtype` the installed torch exposes, sorted by name so all ranks agree on the numbering, and `_dtype_header_code` raises `ValueError` for anything not in that map instead of coding unknowns to a shared `-1`. Two ranks passing different unlisted dtypes can no longer agree by accident.

seonjinn:
Thanks - the three inline points from @yuekaizhang are all addressed and replied to in-thread (forbidden-export set now covers `eagle_output_layer`; `token_chunk_size` sourced from `policy.draft.token_chunk_size` instead of a constructor-only default; dtype header codes built from the installed `torch.dtype` set with an explicit raise instead of a shared `-1` sentinel).

On the CI half: the head was two commits behind `main` (#3438, #3978). Merged latest `main` (`c49d53e2e`) - clean merge, no conflicts - and `ruff check --select I`, `ruff check` and `ruff format --check` are clean tree-wide. `/ok to test` re-posted against `369ffab9c8fea36f96ddbeb22af832dce1a2867c`.

nemo_rl/algorithms/loss/draft.py
```diff
@@ -426,3 +688,294 @@ def streaming_vocab_parallel_soft_ce(
...
+    elif bin_ids.shape != mask.shape:
+        raise ValueError(
+            f"bin_ids must match mask, got {bin_ids.shape} and {mask.shape}."
+        )
+    elif bin_ids.dtype != torch.long:
+        raise TypeError(f"bin_ids must use torch.long, got {bin_ids.dtype}.")
+    elif bin_ids.device != mask.device:
+        raise ValueError(
+            f"bin_ids and mask must share a device, got {bin_ids.device} and {mask.device}."
+        )
```
yuki-97: nit: this chain copies the sibling's shape / dtype / device checks but stops one short — [draft.py:651-662](https://github.com/NVIDIA-NeMo/RL/blob/b15c930ab4e8cf53894d5b3bee5ec76ff0da3aaf/nemo_rl/algorithms/loss/draft.py#L651-L662) also validates the range, with a comment giving the reason: out-of-range bins "would otherwise surface as a device-side assert (CUDA) or RuntimeError (CPU) deep inside `scatter_add_`".

The effect is that the same malformed input reports differently depending on a performance knob. With `num_tokens <= token_chunk_size` this delegates to `streaming_vocab_parallel_soft_ce` and the caller gets `ValueError: bin_ids must lie in [0, N)`; above the threshold it reaches [draft.py:515](https://github.com/NVIDIA-NeMo/RL/blob/b15c930ab4e8cf53894d5b3bee5ec76ff0da3aaf/nemo_rl/algorithms/loss/draft.py#L515) directly. Nothing reaches it today — the DFlash wrapper builds `bin_ids` with max `gamma - 1` against `num_bins = gamma` — so this is about the contract the next caller will meet.

**Suggest fix:** mirror the sibling's check here, or say in the docstring that range validation is the caller's job on this path and why (the sibling's own comment notes the check costs a device sync).

tests/unit/algorithms/test_projected_draft_soft_ce.py
```diff
@@ -0,0 +1,452 @@
...
+    num_tokens, hidden_size, vocab_size = 5, 3, 17
+    full_teacher_logits = torch.randn(
+        context_length,
+        vocab_size,
+        generator=generator,
+        dtype=torch.bfloat16,
+    )
+    selected_teacher_logits = full_teacher_logits.index_select(
+        0, torch.tensor([1, 7, 42, context_length - 1, 7])
+    )
```
yuki-97: nit: this assertion cannot fail. `index_select` allocates a fresh `(5, 17)` tensor, so `full_teacher_logits` is never reachable from the function under test and `max(saved_storage_bytes)` is bounded by the 5-row input regardless of what the implementation saves. Both parametrize rows pass the same 5-row tensor and only enlarge the right-hand side.

The regression the test is named for — saving a full-context intermediate for backward — would ship green. Passing a view instead makes the storage observable, since `untyped_storage().nbytes()` on a view reports the whole buffer:

```suggestion
    selected_teacher_logits = full_teacher_logits[1:6]
```

PR #3717, opened 2026-08-19
fix: fix missing tokenizer in unit test

yuki-97:
fix the unit test fail comes from the silent merge conflict of https://github.com/NVIDIA-NeMo/RL/pull/3668 + https://github.com/NVIDIA-NeMo/RL/pull/3391.

validated `test_setup_initializes_noncolocated_dynamo_with_nemo_gym` and `test_setup_starts_nemo_gym_for_trtllm` added in #3391 could pass.

### DIFF
--- a/tests/unit/algorithms/test_grpo.py
+++ b/tests/unit/algorithms/test_grpo.py
@@ -2407,6 +2407,7 @@ def test_setup_initializes_noncolocated_dynamo_with_nemo_gym(monkeypatch) -> Non
         env_configs=master_config.env,
         base_urls=["http://dynamo-wrapper.example/v1"],
         model_name=master_config.policy["model_name"],
+        tokenizer=tokenizer,
         enable_router_replay=False,
         routed_experts_dtype="int16",
         use_fastokens=False,
@@ -2770,13 +2771,15 @@ def test_setup_starts_nemo_gym_for_trtllm(monkeypatch, mock_grpo_components):
 
     dataset = MagicMock()
     dataset.__len__ = MagicMock(return_value=1)
-    result = grpo_mod.setup(master_config, MagicMock(), dataset, None)
+    tokenizer = MagicMock()
+    result = grpo_mod.setup(master_config, tokenizer, dataset, None)
 
     assert result[2] is nemo_gym_actor
     spinup_nemo_gym_actor.assert_called_once_with(
         env_configs=master_config.env,
         base_urls=["http://trtllm.example/v1"],
         model_name="test-model",
+        tokenizer=tokenizer,
         enable_router_replay=False,
         routed_experts_dtype="int16",
         use_fastokens=False,


yuki-97:
## Self-review — `/review-pr-team`

**Verdict: no blocking findings.** 4 agents (RL expert, test reviewer, bug finder, devil's advocate). 3 candidate findings raised; all three fell below the confidence×impact bar and none land on a line in this diff.

### CI

All checks green on 9cfeea9d5a5d6b9c866dbdf781e79993dd89b4db — 78 passed, 13 skipped, 0 failed.

The relevant one is `L0_Unit_Tests_Algorithms`, which runs `unit/algorithms/`. Neither fixed test carries a pytest mark (and `test_grpo.py` has no module-level `pytestmark`), so neither is routed to the mcore or nemo_gym shards, and neither appears in the `FAST=1` deselect list in `excluded_unit_tests.sh`. Both therefore ran in that shard and passed — confirming the two assertions this PR fixes.

`L0_Unit_Tests_Environments` and `L0_Unit_Tests_Nemo_Gym` are also green, which is consistent with the `test_nemo_gym_health.py` item below being latent rather than broken.

Review analysis itself was static — no GPU and no pytest/uv on the host where the skill ran, so conclusions below come from reading source, AST checks, and the #3668/#3391 patches. CI supplies the execution evidence.

### Root cause confirmed

Both tests are red on main. [`spinup_nemo_gym_actor`](https://github.com/NVIDIA-NeMo/RL/blob/9cfeea9d5a5d6b9c866dbdf781e79993dd89b4db/nemo_rl/environments/nemo_gym.py#L1056) takes `tokenizer` as keyword-only with no default (added in #3668) and [`grpo.py:751`](https://github.com/NVIDIA-NeMo/RL/blob/9cfeea9d5a5d6b9c866dbdf781e79993dd89b4db/nemo_rl/algorithms/grpo.py#L751) passes it, but the two `assert_called_once_with` blocks added in #3391 still list the pre-#3668 kwarg set. #3668 never touched `tests/unit/algorithms/test_grpo.py`, so nothing conflicted textually — clean merge, stale assertions.

### Checked

- **No shadowing in either hunk.** The asserted `tokenizer` is the same object handed to `setup`, so both assertions test real forwarding rather than passing vacuously. In `test_setup_starts_nemo_gym_for_trtllm` the `MagicMock()` had to be bound to a name for that to hold, which is what the second hunk does.
- **Completeness.** Repo-wide sweep of `spinup_nemo_gym_actor`: 2 production call sites, 3 test assertion sites. [`single_controller_utils/setup.py:344`](https://github.com/NVIDIA-NeMo/RL/blob/9cfeea9d5a5d6b9c866dbdf781e79993dd89b4db/nemo_rl/algorithms/single_controller_utils/setup.py#L344) already passes `tokenizer=` (#3668 updated it with its own test), so no third stale assertion remains.
- **No production bug masked by this fix.** Specifically checked whether the single-controller path was missing the kwarg — that would be a runtime `TypeError` a green suite would hide. It is not.
- `setup`'s `tokenizer` is non-Optional and built unconditionally by every entrypoint, so there is no `None` path into `_spinup_nemo_gym`.

### Follow-up (not this PR)

#3668 also *removed* `run_rollouts`' positional `tokenizer`, and one caller still has the old shape: [`test_nemo_gym_health.py:82`](https://github.com/NVIDIA-NeMo/RL/blob/9cfeea9d5a5d6b9c866dbdf781e79993dd89b4db/tests/unit/environments/test_nemo_gym_health.py#L82) calls `run_rollouts([...], None, "timing/x")`, which now binds `timer_prefix=None` and `deduplicate_multimodal_data="timing/x"`. It stays green because [`_require_spinup()`](https://github.com/NVIDIA-NeMo/RL/blob/9cfeea9d5a5d6b9c866dbdf781e79993dd89b4db/nemo_rl/environments/nemo_gym.py#L591) raises before either argument is read — which is exactly what that test asserts — so nothing is broken today. Out of scope here; happy to fold it in if a reviewer prefers.

### Considered and dropped

- `AsyncNemoGymRolloutImpl._tokenizer` ([rollout_manager.py:722](https://github.com/NVIDIA-NeMo/RL/blob/9cfeea9d5a5d6b9c866dbdf781e79993dd89b4db/nemo_rl/experience/rollout_manager.py#L722)) became write-only when #3668 deleted its only read. Pre-existing, no runtime effect, and `RolloutManager._tokenizer:1132` is the same.
- PR title `fix:` vs `test:` — the contributing skill mandates Conventional Commits *format*, not type-by-content.

_Generated by Claude Code_

kajalj22 (approved):

PR #3721, opened 2026-08-20
perf(data): parallelize NeMo Gym image encoding

youngeunkwon0405:
# What does this PR do?

Parallelizes local-image conversion in `encode_images_in_examples` without changing its public API or supported payload shapes.

The function now:
- groups image parts by their exact local source string;
- encodes each unique source once with a bounded `ThreadPoolExecutor(max_workers=8)`;
- closes each decoded PIL image as soon as encoding finishes to release its pixel buffer;
- mutates the NeMo Gym payload only on the caller thread after all unique sources have encoded successfully; and
- reuses one data-URL string for every part sharing a source, reducing the payload serialized across the Ray boundary.

The worker count is capped at eight to bound peak decoded-image memory. In a CPU microbenchmark using 1024 x 1024 PNGs, the thread pool provided approximately 3.7-3.9x speedup. Deduplication is a separate gain: GRPO repeats each prompt per generation, so a group size of `N` previously encoded the same image approximately `N` times and now encodes it once.

HTTP(S) and existing `data:` URLs remain pass-through values, and local paths plus `file://` URLs keep the existing PNG data-URL behavior.

# Issues

None.

# Validation

- `uvx --from ruff==0.9.9 ruff check nemo_rl/data/multimodal_utils.py tests/unit/data/test_multimodal_image_encoding.py`
- `uvx --from ruff==0.9.9 ruff format --check nemo_rl/data/multimodal_utils.py tests/unit/data/test_multimodal_image_encoding.py`
- `uv run --extra nemo_gym --group test pytest -q --noconftest tests/unit/data/test_multimodal_image_encoding.py` - 7 passed on the initial patch
- Refreshed GitHub `Lint check` on `634f5450fecf30ba09c6495261586e01ec370ee2` - passed, including Pyrefly

- [x] Added focused tests
- [x] Ran focused Linux unit tests
- [x] No documentation update is needed because this is an internal performance change with no new configuration or public API


tests/unit/data/test_multimodal_image_encoding.py
```diff
@@ -86,6 +90,71 @@ def test_encode_images_passes_through_http_and_data_urls():
...
+        observed_max_workers.append(kwargs.get("max_workers"))
+        return original_executor(*args, **kwargs)
+
+    monkeypatch.setattr(multimodal_utils, "resolve_to_image", tracking_resolve)
+    monkeypatch.setattr(multimodal_utils, "ThreadPoolExecutor", tracking_executor)
+
+    encode_images_in_examples(examples)
+
+    assert Counter(resolve_calls) == Counter({first: 1, second: 1})
+    assert observed_max_workers == [8]
```
youngeunkwon0405: [`tests/unit/data/test_multimodal_image_encoding.py:130`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/tests/unit/data/test_multimodal_image_encoding.py#L130)

**[TEST]** This assertion (and the `..._uses_eight_workers` name) hardcodes the value of [`NEMO_GYM_IMAGE_ENCODE_MAX_WORKERS`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/nemo_rl/data/multimodal_utils.py#L40), so retuning the constant would break a green test for a non-behavioral reason. Asserting against the constant keeps the valuable property (“the pool is bounded”) without duplicating the literal — consider also renaming to `..._uses_a_bounded_thread_pool`. Verified passing locally with this change.

```suggestion
    assert observed_max_workers == [
        multimodal_utils.NEMO_GYM_IMAGE_ENCODE_MAX_WORKERS
    ]
```

tests/unit/data/test_multimodal_image_encoding.py
```diff
@@ -86,6 +90,71 @@ def test_encode_images_passes_through_http_and_data_urls():
     assert parts[2]["image_url"] == data_url
 
 
+def test_encode_images_deduplicates_sources_and_uses_eight_workers(
```
youngeunkwon0405: [`tests/unit/data/test_multimodal_image_encoding.py:93`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/tests/unit/data/test_multimodal_image_encoding.py#L93)

**[TEST]** The most user-visible semantic change in this PR is currently unpinned: previously a failing source left earlier parts already rewritten in place, while the new code [mutates parts only after every source encodes](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/nemo_rl/data/multimodal_utils.py#L1092) — all-or-nothing (we verified this empirically; a missing path still raises `FileNotFoundError` unchanged, with zero parts mutated). A future refactor back to per-item encoding would silently reintroduce partial mutation. Consider one small test: one good source + one nonexistent path, `pytest.raises(FileNotFoundError)`, then assert both parts still hold their original path strings.

tests/unit/data/test_multimodal_image_encoding.py
```diff
@@ -86,6 +90,71 @@ def test_encode_images_passes_through_http_and_data_urls():
...
+
+    assert Counter(resolve_calls) == Counter({first: 1, second: 1})
+    assert observed_max_workers == [8]
+    for example in examples:
+        parts = example["responses_create_params"]["input"][0]["content"]
+        assert parts[0]["image_url"] == expected[first]
+        assert parts[1]["image_url"] == expected[second]
+
+
+def test_encode_single_image_source_closes_image_on_error(monkeypatch):
```
youngeunkwon0405: [`tests/unit/data/test_multimodal_image_encoding.py:137`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/tests/unit/data/test_multimodal_image_encoding.py#L137)

**[TEST]** This pins close-on-error only; nothing pins close-on-success — deleting the [`finally: image.close()`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/nemo_rl/data/multimodal_utils.py#L884) leaves the suite green (checked via mutation). Since the helper exists to bound decoded-image memory across 8 worker threads, a success-path twin of this test would lock that in (verified: passes on baseline, fails with the `finally` removed):

```python
def test_encode_single_image_source_closes_image_on_success(monkeypatch):
    class CloseTrackingImage:
        def __init__(self):
            self.closed = False

        def close(self):
            self.closed = True

    image = CloseTrackingImage()
    monkeypatch.setattr(multimodal_utils, "resolve_to_image", lambda _: image)
    monkeypatch.setattr(
        multimodal_utils, "image_to_data_url", lambda _: "data:image/png;base64,AA"
    )
    assert (
        multimodal_utils._encode_single_image_source("x.png")
        == "data:image/png;base64,AA"
    )
    assert image.closed
```

nemo_rl/data/multimodal_utils.py
```diff
@@ -877,6 +879,15 @@ def image_to_data_url(image: Image.Image, fmt: str = "PNG") -> str:
     return f"data:image/{fmt.lower()};base64,{encoded}"
 
 
+def _encode_single_image_source(source: str) -> str:
+    """Resolve and encode one local image source."""
```
youngeunkwon0405: [`nemo_rl/data/multimodal_utils.py:883`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/nemo_rl/data/multimodal_utils.py#L883)

**[DOCSTRING]** Tiny wording nit: the helper accepts anything [`resolve_to_image`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/nemo_rl/data/multimodal_utils.py#L840) accepts (including http/data URLs) — only the caller restricts it to local sources.

```suggestion
    """Resolve and encode one image source."""
```

nemo_rl/data/multimodal_utils.py
```diff
@@ -36,6 +37,7 @@
 MULTIMODAL_CONTENT_TYPES = frozenset(
     {"input_image", "image", "image_url", "video", "audio"}
 )
+NEMO_GYM_IMAGE_ENCODE_MAX_WORKERS = 8
```
youngeunkwon0405, line 43: [`nemo_rl/data/multimodal_utils.py:40`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/nemo_rl/data/multimodal_utils.py#L40)

**[PERF-EVIDENCE]** This is a nice win — we reproduced ~3.7–3.9× from the thread pool alone on 1024² PNGs, and the dedup win is structural: `BatchedDataDict.repeat_interleave` deep-copies parts per generation, so the same file was previously re-encoded once per generation in a GRPO group. It would be helpful to capture that in the PR description: (a) the dedup factor (≈`num_generations_per_prompt`) stated separately from the threading factor, (b) one line on why `max_workers=8` (e.g. bounding peak decoded-image memory), and (c) softening “closes the worker-owned PIL image” — `convert("RGB")` already detached the file handle, so `close()` frees the pixel buffer early rather than an fd. A bonus worth a line: all parts sharing a source now reference one `str` object, which also shrinks the pickled payload across the Ray boundary.

nemo_rl/data/multimodal_utils.py
```diff
@@ -1062,7 +1075,27 @@ def encode_images_in_examples(nemo_gym_examples: list[dict]) -> list[dict]:
                     continue
                 if url.startswith(("http://", "https://", "data:")):
                     continue
-                part["image_url"] = image_to_data_url(resolve_to_image(url))
+                parts_by_source.setdefault(url, []).append(part)
```
youngeunkwon0405: [`nemo_rl/data/multimodal_utils.py:1069`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/nemo_rl/data/multimodal_utils.py#L1069)

**[Question, optional]** Pre-existing, but since this PR reworked the loop: the filter here accepts only `type == "input_image"` / key `image_url`, while the sibling [`extract_input_image_sources_from_responses_messages`](https://github.com/NVIDIA-NeMo/RL/blob/6957387c3a9912df92296df121c9a7f25480864f/nemo_rl/data/multimodal_utils.py#L908) also accepts `image`/`image_url` types and `image`/`image_url`/`url` keys — so a `{"type": "image_url"}` part with a local path would ship unencoded. Is the narrower filter deliberate here? No change requested — just checking whether it deserves a clarifying comment.

tests/unit/data/test_multimodal_image_encoding.py
youngeunkwon0405: Updated the test name to describe the bounded-pool contract and now assert against `NEMO_GYM_IMAGE_ENCODE_MAX_WORKERS` instead of duplicating the literal. Fixed in 634f5450f.

tests/unit/data/test_multimodal_image_encoding.py
youngeunkwon0405: Added a regression test with one valid image and one missing source. It verifies that the exception propagates and neither payload part is mutated. Fixed in 634f5450f.

tests/unit/data/test_multimodal_image_encoding.py
youngeunkwon0405: Added the success-path twin and shared close-tracking test helper. Both successful encoding and encoding failure now verify that the decoded image is closed. Fixed in 634f5450f.

nemo_rl/data/multimodal_utils.py
youngeunkwon0405: Updated the helper docstring to say `image source`; the caller, rather than the helper, is what restricts this path to local inputs. Fixed in 634f5450f.

nemo_rl/data/multimodal_utils.py
youngeunkwon0405: Updated the PR description with the separate deduplication factor, the 3.7-3.9x threading result, the eight-worker decoded-memory bound, the pixel-buffer close wording, and data-URL string reuse across the Ray payload.

nemo_rl/data/multimodal_utils.py
youngeunkwon0405: Yes, the narrower filter is deliberate. This function preserves the existing outbound NeMo Gym Responses request wire format (`input_image` plus `image_url`); the sibling extractor handles broader returned-message variants. Added an inline comment to make that boundary explicit in 634f5450f.

nemo_rl/data/multimodal_utils.py
```diff
@@ -1062,7 +1078,27 @@ def encode_images_in_examples(nemo_gym_examples: list[dict]) -> list[dict]:
...
+
+    sources = list(parts_by_source)
+    if sources:
+        with ThreadPoolExecutor(
```
aroshanghias-nvd, line 1103: **[RESOURCE ACCOUNTING]** `NemoGym` does not declare a runtime `num_cpus` reservation, while this pool may actively use up to eight cores—consistent with the reported 3.7–3.9× CPU speedup. Could multiple `NemoGym` actors be colocated on the same node? If so, should the worker count be derived from an explicitly reserved/configured CPU budget, or could we include representative multi-actor measurements showing that eight workers do not cause contention?

aroshanghias-nvd (commented):
One non-blocking operational concern:

nemo_rl/data/multimodal_utils.py
youngeunkwon0405: Thanks for raising this. In my understanding, the current GRPO path creates a single NemoGym actor per controller, so multiple NemoGym actors are not normally colocated within one run. The executor is also active only during the image-preparation phase.
Configuring the NemoGym actor’s num_cpus would affect Ray scheduling, placement, and other resource-management behavior, so that requires a significantly broader discussion with the NeMo Gym team. I’d prefer to keep this PR focused on parallelizing image encoding with a bounded eight-worker pool.

DanialTaheri (approved):
looks good to me

DanialTaheri:
@yfw @terrykong  can we merge this? It is a harmless optimization for image encoding.

youngeunkwon0405:
Hi @terrykong, can I get your help with merging this PR? It is a ~20-line logic change that parallelizes image encoding for the VLM models. I already did it through `review-pr-team`, the CI run passed, and I got approval from @DanialTaheri. Thanks for your help always!

terrykong (approved):

PR #3722, opened 2026-08-20
fix(dtensor-v2): keep the KL reference anchored to base weights across resumes

yuekaizhang:
## What does this PR do?

Cherry-pick of f186c3be from #3643 (which was merged into the `muse-glimmer-support` branch, not `main`) so the fix lands on `main`.

In `dtensor_policy_worker_v2`, `setup_model_and_optimizer` restores `weights_path` internally, so capturing the reference model from its output re-anchors the KL reference to the resumed step on every resume instead of the original base policy. Each resume then grants the policy a fresh KL drift budget, defeating the anchor over multi-window runs.

Fix: when a reference model is needed on a resumed run, defer the NeMo RL checkpoint load — build the model from the pristine base weights first, capture the reference from that, then load the checkpoint. This matches the ordering already used by the v1 `dtensor_policy_worker` (reference captured before `load_checkpoint`).

## Issues
Regression introduced in the v2 worker refactor; v1 worker is unaffected.


nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -323,7 +323,16 @@ def __init__(
...
+        # Set up model and optimizer.
+        # When a reference policy is needed on a resumed run, defer the NeMo RL
+        # checkpoint load: setup_model_and_optimizer restores `weights_path`
+        # internally, and capturing the reference from the restored model
+        # re-anchors the KL reference to the resumed step on every resume (a
+        # rolling anchor). Each resume then grants the policy a fresh drift
+        # budget, which defeats the anchor over multi-window runs. Instead,
+        # build the model from the pristine base weights, capture the
+        # reference, then load the checkpoint below.
+        defer_checkpoint_load = init_reference_model and weights_path is not None
```
yuekaizhang: **1 action item.**

[`nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py:335`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py#L335) — PR-introduced. With deferred loading, [`setup_model_and_optimizer`'s else branch](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/automodel/setup.py#L862-L864) prints `No weights path provided. Loaded base HF weights via from_pretrained (default policy init)` on every KL-anchored resume, immediately before the worker loads the real checkpoint — a contradictory log pair exactly where resume triage looks first.

### AI-1
Print an explanatory line on the deferred path (and, as a nit, use truthiness to match [setup.py's own guard](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/automodel/setup.py#L854) — no real producer passes `""` ([`get_resume_paths`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/utils/checkpoint.py#L190) returns `Path | None`), so this is defensive consistency only):

```suggestion
        defer_checkpoint_load = init_reference_model and bool(weights_path)
        if defer_checkpoint_load:
            print(
                "Deferring NeMo RL checkpoint load until after the KL reference "
                "is captured from base weights; the 'No weights path provided' "
                "message from setup_model_and_optimizer is expected on this path."
            )
```

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -350,11 +359,22 @@ def __init__(
...
 
-        # Initialize reference model if requested
+        # Initialize reference model if requested. With deferred loading the
+        # model still holds the base (model_name) weights here, so the KL
+        # reference stays anchored to the same policy across resumes.
         self.reference_model_state_dict = None
         if init_reference_model:
             self.reference_model_state_dict = setup_reference_model_state(self.model)
 
+        if defer_checkpoint_load:
```
yuekaizhang, line 374: **1 action item.**

**TL;DR — the exact branch this PR fixes (`init_reference_model=True` + `weights_path` set) has zero test coverage anywhere in `tests/`, so the rolling-anchor regression could silently come back.**

[`nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py:369`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py#L369) — PR-introduced behavior, pre-existing coverage gap. The only v2 resume test, [`test_dtensor_v2_checkpoint_save_and_load`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/tests/unit/models/policy/test_dtensor_worker_v2.py#L450), constructs both policies with `init_reference_model=False`; the only `init_reference_model=True` call sites in the tree are megatron tests.

### AI-1
Add the mock-based tests below to `tests/unit/models/policy/test_dtensor_worker_v2.py` (plus `import nemo_rl.models.policy.workers.dtensor_policy_worker_v2 as worker_mod` under the existing try-import). Verified: 3 passed on this PR's head in ~19s (automodel venv), and the key test fails against pre-fix `main` (setup receives the checkpoint path, no deferred load) — so it pins the contract, not the implementation.

```python
def _init_v2_worker_mocked(monkeypatch, *, init_reference_model, weights_path, optimizer_path):
    """Run DTensorPolicyWorkerV2Impl.__init__ with all heavy deps mocked.

    Returns (worker, call_log, setup_mock, load_checkpoint_mock).
    """
    from unittest.mock import MagicMock

    call_log = []

    monkeypatch.setattr(worker_mod, "apply_transformer_engine_patch", lambda: None)
    monkeypatch.setattr(worker_mod.ray, "get_gpu_ids", lambda: [0])
    monkeypatch.setattr("nemo_rl.distributed.numa_utils.bind_to_gpu_numa", lambda gpu_id: None)
    monkeypatch.setattr(
        "nemo_rl.models.automodel.setup.get_tokenizer",
        lambda cfg, get_processor=False: MagicMock(name="tokenizer"),
    )

    # 13-tuple unpacked as runtime config at the end of __init__.
    runtime_config = (
        "model_class", "model_config", {}, False, "attn_impl", None, False,
        1.0, False, False, False, None, False,
    )
    monkeypatch.setattr(worker_mod, "validate_and_prepare_config", lambda **kw: runtime_config)
    monkeypatch.setattr(worker_mod, "setup_distributed", lambda **kw: MagicMock())
    monkeypatch.setattr(worker_mod.torch.distributed, "get_rank", lambda: 0)
    monkeypatch.setattr(worker_mod, "maybe_preinit_nixl_checkpoint_engine", lambda cfg: None)

    load_checkpoint_mock = MagicMock(side_effect=lambda **kw: call_log.append("load_checkpoint"))

    def fake_init_checkpoint_manager(self, config_updates=None, checkpoint_root=None):
        self.checkpoint_manager = MagicMock()
        self.checkpoint_manager.load_checkpoint = load_checkpoint_mock

    monkeypatch.setattr(
        DTensorPolicyWorkerV2Impl, "_init_checkpoint_manager", fake_init_checkpoint_manager
    )

    # 10-tuple unpacked as model_and_optimizer_state.
    model_and_optimizer_state = (
        MagicMock(name="model"), MagicMock(name="optimizer"), MagicMock(name="scheduler"),
        False, False, False, "model_class", "model_config", None, False,
    )
    setup_mock = MagicMock(
        side_effect=lambda **kw: (
            call_log.append("setup_model_and_optimizer"), model_and_optimizer_state
        )[1]
    )
    monkeypatch.setattr(worker_mod, "setup_model_and_optimizer", setup_mock)

    ref_state = {"ref": "state"}
    monkeypatch.setattr(
        worker_mod,
        "setup_reference_model_state",
        lambda model: (call_log.append("setup_reference_model_state"), ref_state)[1],
    )

    config = {"model_name": "base-model", "tokenizer": {}, "dtensor_cfg": {}, "generation": {}}
    worker = object.__new__(DTensorPolicyWorkerV2Impl)
    DTensorPolicyWorkerV2Impl.__init__(
        worker, config,
        weights_path=weights_path, optimizer_path=optimizer_path,
        init_optimizer=True, init_reference_model=init_reference_model,
    )
    return worker, call_log, setup_mock, load_checkpoint_mock


@pytest.mark.automodel
@pytest.mark.skipif(not NEMO_AUTOMODEL_AVAILABLE, reason="nemo_automodel not available")
def test_dtensor_v2_resume_with_reference_model_defers_checkpoint_load(monkeypatch):
    """On resume with a KL reference, the reference must be captured from base
    weights (checkpoint load deferred until after the capture)."""
    worker, call_log, setup_mock, load_mock = _init_v2_worker_mocked(
        monkeypatch, init_reference_model=True,
        weights_path="/ckpt/weights", optimizer_path="/ckpt/optim",
    )
    # (a) base weights used for setup: checkpoint paths not passed through.
    assert setup_mock.call_args.kwargs["weights_path"] is None
    assert setup_mock.call_args.kwargs["optimizer_path"] is None
    # (b) reference captured BEFORE the checkpoint load.
    assert call_log == [
        "setup_model_and_optimizer", "setup_reference_model_state", "load_checkpoint",
    ]
    assert worker.reference_model_state_dict == {"ref": "state"}
    # (c) checkpoint still loaded, with the original paths.
    assert load_mock.call_args.kwargs["weights_path"] == "/ckpt/weights"
    assert load_mock.call_args.kwargs["optimizer_path"] == "/ckpt/optim"
    assert load_mock.call_args.kwargs["model"] is worker.model
    assert load_mock.call_args.kwargs["optimizer"] is worker.optimizer


@pytest.mark.automodel
@pytest.mark.skipif(not NEMO_AUTOMODEL_AVAILABLE, reason="nemo_automodel not available")
def test_dtensor_v2_resume_without_reference_model_passes_paths_through(monkeypatch):
    worker, call_log, setup_mock, load_mock = _init_v2_worker_mocked(
        monkeypatch, init_reference_model=False,
        weights_path="/ckpt/weights", optimizer_path="/ckpt/optim",
    )
    assert setup_mock.call_args.kwargs["weights_path"] == "/ckpt/weights"
    assert setup_mock.call_args.kwargs["optimizer_path"] == "/ckpt/optim"
    load_mock.assert_not_called()
    assert worker.reference_model_state_dict is None


@pytest.mark.automodel
@pytest.mark.skipif(not NEMO_AUTOMODEL_AVAILABLE, reason="nemo_automodel not available")
def test_dtensor_v2_fresh_run_with_reference_model_does_not_defer(monkeypatch):
    worker, call_log, setup_mock, load_mock = _init_v2_worker_mocked(
        monkeypatch, init_reference_model=True, weights_path=None, optimizer_path=None,
    )
    assert setup_mock.call_args.kwargs["weights_path"] is None
    load_mock.assert_not_called()
    assert worker.reference_model_state_dict == {"ref": "state"}
```

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -350,11 +359,22 @@ def __init__(
...
+            self.checkpoint_manager.load_checkpoint(
```
yuekaizhang: **No action needed in this PR — 1 follow-up.**

[`nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py:370`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py#L370) — adjacent maintainability note, not something this PR broke. This deferred call is kwarg-for-kwarg identical to [the internal load inside `setup_model_and_optimizer`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/automodel/setup.py#L855-L861). The megatron path already grew a [`pre_load_checkpoint_hook`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/megatron/setup.py#L1867-L1868) threaded around its load (relied on by [`megatron_quant_policy_worker.py`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/policy/workers/megatron_quant_policy_worker.py#L174-L185)); if automodel setup gains an analogous pre-load step, this duplicated call silently bypasses it on exactly the resume-with-reference path.

### Follow-up
File an issue to give `setup_model_and_optimizer` a `defer_checkpoint_load: bool = False` (skip the internal load + suppress the misleading else-print) so the load kwargs live in one place.

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -332,8 +341,8 @@ def __init__(
             checkpoint_manager=self.checkpoint_manager,
             is_vlm=self.is_vlm,
             init_optimizer=init_optimizer,
-            weights_path=weights_path,
-            optimizer_path=optimizer_path,
+            weights_path=None if defer_checkpoint_load else weights_path,
```
yuekaizhang, line 349: **2 action items** (PR metadata, not code).

[`nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py:335`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py#L335) — this changes the KL penalty magnitude on every resumed run with `reference_policy_kl_penalty > 0` (resume `weights_path` flows only from [`get_resume_paths`](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/nemo_rl/algorithms/grpo.py#L1120), so all KL-anchored resumes hit this branch).

### AI-1
Link the supporting evidence from the original PR #3643 / commit f186c3be in the PR description (e.g. the resumed-run KL curve showing the anchor no longer rolls). No new convergence run needed — the direction of effect is provably the documented intent.

### AI-2
Apply the `CI:L1` label ([label tiers](https://github.com/NVIDIA-NeMo/RL/blob/79c519e583a80037343573e1e6f5cfddf092e361/docs/ci-cd.md)) so functional tests run on this PR.

yuekaizhang (commented):
Self-review via the team-review skill (5 agents: rl-expert, bug-finder, test-agent, design-reviewer, devil's advocate), per the new self-review-before-human-review process.

Summary: the fix itself was verified correct and necessary — deferred load is kwarg-identical to the internal one, `setup_reference_model_state` copies to pinned CPU memory (no aliasing), ordering matches the v1 worker's existing semantics, and the megatron/value workers are unaffected. Devil's advocate: 2 findings confirmed, 3 downgraded, 0 disputed. Linters (pre-commit on the changed file): pass. All actionable findings are inline below.

_Generated by Claude Code_

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuekaizhang: fixed in 358a19d77cc547ef5a91216a10fea3b45708b0d5

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuekaizhang: fixed in 358a19d77cc547ef5a91216a10fea3b45708b0d5

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuekaizhang: fixed in 358a19d77cc547ef5a91216a10fea3b45708b0d5 — AI-1: PR description now has an Evidence section linking #3643 / f186c3be37600c0140726951bc7e05ecc6726e68; AI-2: CI:L1 label applied.

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -350,11 +365,22 @@ def __init__(
...
             self.reference_model_state_dict = setup_reference_model_state(self.model)
 
+        if defer_checkpoint_load:
+            self.checkpoint_manager.load_checkpoint(
+                model=self.model,
+                weights_path=weights_path,
+                optimizer=self.optimizer,
+                optimizer_path=optimizer_path,
+                scheduler=self.scheduler,
+            )
```
yuki-97: This block is byte-identical to the worker's own [`load_checkpoint`](https://github.com/NVIDIA-NeMo/RL/blob/358a19d77cc547ef5a91216a10fea3b45708b0d5/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py#L1387-L1398) — same five kwargs, same order.

v1 already goes through the method at [`dtensor_policy_worker.py:527`](https://github.com/NVIDIA-NeMo/RL/blob/358a19d77cc547ef5a91216a10fea3b45708b0d5/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L527), which is the ordering this PR cites as precedent. Reusing it keeps one place to update if the manager signature moves, and the new tests are unaffected — they assert on `checkpoint_manager.load_checkpoint` kwargs, which are unchanged.

```suggestion
        if defer_checkpoint_load:
            self.load_checkpoint(weights_path, optimizer_path)
```

tests/unit/models/policy/test_dtensor_worker_v2.py
```diff
@@ -1003,3 +1004,164 @@ def test_multiple_cp_buffers_sequence_dim_replication(
...
+        "attn_impl",
+        None,
+        False,
+        1.0,
+        False,
+        False,
+        False,
+        None,
+        False,
+    )
```
yuki-97: nit: this mock and the one at line 1068 return bare positional tuples where production returns the [`RuntimeConfig`](https://github.com/NVIDIA-NeMo/RL/blob/358a19d77cc547ef5a91216a10fea3b45708b0d5/nemo_rl/models/automodel/config.py#L41-L73) and [`ModelAndOptimizerState`](https://github.com/NVIDIA-NeMo/RL/blob/358a19d77cc547ef5a91216a10fea3b45708b0d5/nemo_rl/models/automodel/config.py#L76-L92) NamedTuples — `False, False, False, "model_class"` can't be read without counting positions, and a swap of two same-typed bools would be silent. A field added to either also surfaces as `not enough values to unpack` at [`dtensor_policy_worker_v2.py:385`](https://github.com/NVIDIA-NeMo/RL/blob/358a19d77cc547ef5a91216a10fea3b45708b0d5/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py#L385), not here.

**Suggest fix:** construct them by keyword — NamedTuple is a tuple subclass, so it's a drop-in for the positional unpack, and [`test_automodel_setup.py:607`](https://github.com/NVIDIA-NeMo/RL/blob/358a19d77cc547ef5a91216a10fea3b45708b0d5/tests/unit/models/automodel/test_automodel_setup.py#L607) already does this. Add both classes to the existing try-import block: [`config.py:20`](https://github.com/NVIDIA-NeMo/RL/blob/358a19d77cc547ef5a91216a10fea3b45708b0d5/nemo_rl/models/automodel/config.py#L20) imports `PeftConfig` from `nemo_automodel` at module level.

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -323,7 +323,22 @@ def __init__(
             },
         )
 
-        # Set up model and optimizer
+        # Set up model and optimizer.
+        # When a reference policy is needed on a resumed run, defer the NeMo RL
```
yuki-97, line 327: can we make this feature as a configurable config and default false (to keep current behavior)? since some exps are intend to load ckpt as the reference model.

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
```diff
@@ -332,8 +347,8 @@ def __init__(
             checkpoint_manager=self.checkpoint_manager,
             is_vlm=self.is_vlm,
             init_optimizer=init_optimizer,
-            weights_path=weights_path,
-            optimizer_path=optimizer_path,
+            weights_path=None if defer_checkpoint_load else weights_path,
```
yuki-97, line 349: could you add a functional test and some wandb curves to show this feature works well? e.g. after resuming the KL is obviously smaller than w/o the feature.

yuki-97 (commented):
thanks for adding this, feels this is a feature rather than a fix, left some comments.

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuekaizhang: Sure.


Attached training curve from meta/muse-glimmer-30B

Red: no kl penalty
Green: kl penalty 0.01 but without the PR （See left figure, around step 120, resuming training）
Blue: kl penalty 0.01 with the PR

image

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuekaizhang: fixed in 198064fe6a8593b1f6021ea2addcea4e2b2a39e5

tests/unit/models/policy/test_dtensor_worker_v2.py
yuekaizhang: fixed in 198064fe6a8593b1f6021ea2addcea4e2b2a39e5

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuekaizhang: fixed in e18d99aef4e757b59b7cfdb8ef2f6cccc0449494 — added policy config bool ref_model_from_base_weights, default false (resume keeps loading the checkpoint before the reference is captured, byte-for-byte the current behavior; a new unit test pins it). true opts into the deferred load. Named it ref_model_from_base_weights rather than ref_model_load_from_ckpt so the name describes the opt-in state instead of negating the default (ref_model_load_from_ckpt: false would read as if the default did not load the reference from the checkpoint, which is the opposite of what it does).

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuekaizhang: fixed in 0eaa949b9f2f5544cbf7b89381aa136608a2a6c2 — added tests/functional/grpo_automodel_ref_anchor_resume.sh (registered in L1_Functional_Tests_AutoModel.sh): an uninterrupted 3-step baseline checkpointing step_2, then a default resume and a flag-on resume from copies of that checkpoint. It asserts step-3 train/kl_penalty collapses at the resume boundary by default (reference re-anchored to the checkpoint) and stays continuous with the baseline when ref_model_from_base_weights=true. Verified locally on 2xH100: baseline 9.4e-2, default resume 3.7e-3 (noise floor), flag-on resume 1.1e-1. wandb curves are in the earlier reply on this PR.

yuekaizhang:
/ok to test [25bfb18](https://github.com/NVIDIA-NeMo/RL/pull/3722/commits/25bfb18e12adf5084f6ac6be65caa490533ea42b)

nemo_rl/models/policy/__init__.py
```diff
@@ -579,6 +579,14 @@ class PolicyConfig(TypedDict):
...
     draft: NotRequired[DraftConfig | DraftConfigDisabled]
     pretrained_checkpoint: NotRequired[PretrainedCheckpointConfig]
+    # If true, the KL reference model is captured from the pristine base
+    # (model_name) weights even when resuming from a NeMo RL checkpoint, so the
+    # reference stays anchored to the same policy across resumes. If false
+    # (default), the reference is captured from the restored checkpoint weights,
+    # i.e. it re-anchors to the resumed step on every resume.
+    # Currently consumed by the DTensor v2 (automodel) path only; the Megatron
+    # path always anchors the reference to `pretrained_checkpoint`.
+    ref_model_from_base_weights: NotRequired[bool]
```
yuki-97: The shipped default `false` documents behaviour that two of the three backends cannot produce. [grpo_math_1B_megatron.yaml:2](https://github.com/NVIDIA-NeMo/RL/blob/25bfb18e12adf5084f6ac6be65caa490533ea42b/examples/configs/grpo_math_1B_megatron.yaml#L2) inherits this key from grpo_math_1B.yaml and resolves it to `false`, but Megatron rebuilds the reference from `pretrained_checkpoint` unconditionally ([megatron/setup.py:479](https://github.com/NVIDIA-NeMo/RL/blob/25bfb18e12adf5084f6ac6be65caa490533ea42b/nemo_rl/models/megatron/setup.py#L479) takes no `weights_path`; [:2048](https://github.com/NVIDIA-NeMo/RL/blob/25bfb18e12adf5084f6ac6be65caa490533ea42b/nemo_rl/models/megatron/setup.py#L2048)), and DTensor v1 captures at [dtensor_policy_worker.py:481](https://github.com/NVIDIA-NeMo/RL/blob/25bfb18e12adf5084f6ac6be65caa490533ea42b/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L481) before its only load at [:527](https://github.com/NVIDIA-NeMo/RL/blob/25bfb18e12adf5084f6ac6be65caa490533ea42b/nemo_rl/models/policy/workers/dtensor_policy_worker.py#L527). Both are always base-anchored, i.e. `true` — and neither this comment nor the YAML one mentions v1 at all.

Note the bind before guarding on it: `true` is what those two already do, so rejecting truthy would fail configs that behave exactly as asked. The value they cannot honour is `false`, which is the default. Could you file an issue for checkpoint-anchored references on Megatron and link it here? Happy to help on that side if useful.

nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py
yuekaizhang: Updated in 58b78860494531c7015ddcea87fa0c1c8b4df0b6 — per our discussion this is now handled as an unconditional bug fix: the ref_model_from_base_weights flag is removed and the v2 worker always captures the KL reference from base weights on resume (matching the v1 worker and Megatron ordering). The functional test now asserts KL continuity across the resume boundary directly.

yuki-97 (approved):

PR #3723, opened 2026-08-20
feat(draft): add tensor-parallel DFlash body

seonjinn:
## Summary

- Adds a checkpoint-compatible tensor-parallel DFlash body.
- Uses block-only attention with deterministic block planning.
- Fails synchronously before projection when ranks disagree on input shape or topology.

## Supported contract

- Supported: TP1/TP2/TP4 with CP1.
- Checkpoint ownership and TP forward/gradient paths are covered.
- CP>1, draft-body sequence parallelism, and full PP integration are out of scope.

## Validation

- Exact head: `d43a5b6cd82bfd348289a112fa02708765cd130e`
- Exact-head Ruff lint, import-order, and format checks passed.

## Prior GPU evidence

- Prior stacked-head suite: 56 focused tests passed. External 4×GB200 coverage exercised TP2, TP4, TP2×PP2, malformed-input, and CP-rejection paths; TP2×PP2 is external evidence, not a unit-test topology.

## Dependencies / Remaining

#3715 adds training integration; #3744 adds generation/refit integration. CP-local attention requires a separate design.


tests/unit/distributed/test_dflash_body_tp.py
```diff
@@ -0,0 +1,577 @@
...
+        DFlashBody(
+            _config(),
+            tp_group=torch.distributed.group.WORLD,
+            parallel_config=parallel_config,
+        )
+
+    assert parallel_config.sequence_parallel is True
+
+
+def test_tp2_pp2_last_stage_forward_gradient_and_replica_ids(
```
seonjinn: `test_tp2_pp2_last_stage_forward_gradient_and_replica_ids` doesn't exercise pipeline parallelism: `_pp2_world` is a plain 4-rank `init_process_group` world, and this test splits it into two independent TP=2 groups with the entire body gated by `if rank >= 2:` -- that's DP2xTP2 with test code simulating "only the last stage runs", not TP2xPP2 (no `pipeline_model_parallel_size`, no stage split, no send/recv, and `dflash.py` itself contains no PP logic).

The subgroup-owned `replica_id` assertions are genuinely valuable, so the test should stay -- but rename it (e.g. `test_tp2_subgroup_forward_gradient_and_replica_ids`) and drop the "TP=2 with PP=2 covers last-stage forward" line from the PR body, or add a real PP test that initializes Megatron pipeline groups and asserts non-last-stage behavior.

nemo_rl/models/megatron/draft/dflash.py
```diff
@@ -0,0 +1,738 @@
...
+    dflash_block_only_attention,
+)
+from nemo_rl.models.megatron.draft.block_plan import DFlashBatchPlan
+
+if TYPE_CHECKING:
+    from megatron.core.dist_checkpointing.mapping import ShardedStateDict
+
+
+_ShardedOffsets = tuple[tuple[int, int, int], ...]
+_INPUT_ERROR_MESSAGES = (
```
seonjinn, line 50: Of the 8 `_INPUT_ERROR_MESSAGES` branches, only codes 0, 1, and 7 are exercised by tests (`test_dflash_model.py:408,414`; `test_dflash_body_tp.py:420`). Codes 2-3 (device mismatches) and 4-6 -- all three `TypeError` branches in `_INPUT_TYPE_ERROR_CODES` -- have no test, so the TypeError-vs-ValueError split and the device checks are unguarded against regression in this hand-rolled TP-synchronized validation. Five one-liner `pytest.raises` cases in `test_dflash_model.py` would close this.

nemo_rl/models/megatron/draft/dflash.py
```diff
@@ -0,0 +1,738 @@
...
+    "DFlash inputs must use a floating dtype",
+    "DFlash query positions must use torch.int64",
+    "DFlash plan and input shapes must match across tensor-parallel ranks",
+)
+_INPUT_TYPE_ERROR_CODES = frozenset({4, 5, 6})
+
+
+@dataclass(frozen=True, slots=True)
+class DFlashBodyConfig:
+    """Checkpoint-compatible DFlash body dimensions."""
```
seonjinn: The docstring says "Checkpoint-compatible" without naming the checkpoint. The actual pin (`z-lab/Qwen3-8B-DFlash-b16` @ `9b41424b7109f9c5413454f481b09a82b85333f4`) lives only in `test_dflash_checkpoint_schema.py`, so someone editing these defaults has no in-module signal they're tied to an external HF artifact. Add the repo+revision to this docstring, mirroring the fix applied to `dspark.py` in #3711.

nemo_rl/models/megatron/draft/dflash.py
```diff
@@ -0,0 +1,738 @@
...
+                "DFlash tensor parallel group size"
+            )
+        self.parallel_config = parallel_config
+        partitioned_dimensions = (
+            self.config.hidden_size,
+            self.config.intermediate_size,
+            self.config.num_attention_heads,
+            self.config.num_key_value_heads,
+        )
+        if any(
```
seonjinn, line 450: Because `partitioned_dimensions` includes `num_key_value_heads` (8 by default), this check silently caps supported TP at 8 with no KV-head-duplication fallback (unlike mainstream Megatron GQA-TP). The loud failure is correct; suggest naming the constraint in the message or a `DFlashBodyConfig` docstring line ("tensor_parallel_size must divide num_key_value_heads; max TP=8 for the pinned artifact") so a user hitting TP=16 understands it's a design ceiling, not a bug.

seonjinn (commented):

Method: six independent review agents over the diff, then an adversarial pass that re-verified every citation at the head, re-ran CI checks live, and empirically tested the two highest-severity candidate bugs with a scratch torch 2.13 CPU venv. Both were refuted by running the claimed repros: the short-sequence block-plan IndexError claim misses the `torch.where(block_valid, ...)` masking at `block_plan.py:145-150` (repro runs clean), and the flex_attention shared-KV batch=1 call is an officially supported broadcast when `block_mask` is non-None (bit-identical to expanded KV, and the CUDA tests push num_blocks=4/6 through the public dispatch). The core math, collectives, and checkpoint plumbing survived cleanly: the dense scalar-loop attention oracle is genuinely independent, TP failure tests use per-rank-divergent inputs, real `ColumnParallelLinear`/`RowParallelLinear` are used (so the #3711 TP-attribute bug class doesn't reproduce), and `sharded_state_dict` threads `tp_group` down to leaf norms (so the eagle.py-documented replica_id collapse doesn't either).

What remains is labeling and evidence accuracy:

**Needs action before merge:**

1. **CI has never run on this head.** `0f097a2c2` carries meta checks only; the two `/ok to test` comments target SHAs five commits back. Worse, the last time Lint ran on this branch (`29cc303e3`) it FAILED, CI quality failed at both stale SHAs, and `L1_Functional_Tests_Megatron_4` failed at `125a168dc`. The body's "Ruff and format checks passed on the exact head" has no backing check-run -- if that was a local run, say so. Needs a fresh `/ok to test 0f097a2c25e39fe813f354a829c0929b0750219e` and a green run.
2. **Reword the PP claim.** "TP=2 with PP=2 covers last-stage forward" is not substantiated by the unit suite (see the inline comment) -- the 4xGB200 runs are external evidence and should be labeled as such.

**Description nits:** "56 focused tests" -- the files define 54 test functions; if 56 counts parametrized collection, state the basis. `test_dflash_body_efficiency_contract.py` is a structural/AST contract (the real memory-bound evidence is the CUDA-gated tests in `test_dflash_block_attention.py`) -- consider renaming. One sentence confirming bidirectional (non-causal) intra-block attention is the intended DFlash semantics, and a mention of the new `tools/benchmark_dflash_block_attention.py`, would round out the description.

isomap (commented):
Reviewed the changes and existing review threads. I have no additional code comments. Full CI on the final rebased head still needs to pass before merging.

seonjinn:
Thanks for the pass. Nothing to change on the code side, so this is just the CI half of your note.

The head was two commits behind `main` (#3438 HybridEP recipes, #3978 MOPD full-vocabulary distillation). Merged latest `main` (`c49d53e2e`) in `446bd6b`; the merge was clean, no conflicts. Local gate on the merged head: `ruff check --select I`, `ruff check` and `ruff format --check` all clean across the tree.

`/ok to test` re-posted against `446bd6b96665492fcf3a904ac45a57a2e449dfec` so full CI runs on this head rather than the pre-merge one. All four earlier review threads are resolved.

One heads-up on reading the results: `main` itself is currently red on `L0_Unit_Tests_Megatron_Policy_1` (`test_megatron_split_state.py::TestTrainMicrobatch::test_forwards_model_owned_packing_flags[model-owned-cp-slicing]`), an upstream semantic conflict this branch inherits by merging. If that job is the only failure, it is not from this PR.

Issue #3694, opened 2026-08-18
MXFP8 rollout support for Qwen3.5 (grouped-MoE expert refit)


MarkXCloud:
# **Summary**

#2744 adds FP8 rollout + training for Qwen3.5 (dense and MoE), but only for the **blockwise** recipe (e4m3, `[128, 128]` weight blocks) . MXFP8 currently does not work for Qwen3.5: the grouped-MoE expert refit path only implements blockwise quantization, so an MXFP8 refit of a Qwen3.5 MoE model fails at the first refit with:

```
NotImplementedError: MXFP8 refit does not support grouped MoE expert weights.
```

(guard + unit test added in #2744). Dense Qwen3.5 (9B) does not hit the guard but is unvalidated under MXFP8.

# **Proposal**

End-to-end MXFP8 rollout support for Qwen3.5, which needs:

1. An MXFP8 counterpart of `_expand_grouped_moe_expert_to_fp8` in `nemo_rl/models/generation/vllm/quantization/fp8.py`: per-expert e4m3 data + E8M0 scales (32-element blocks along the reduction dim), emitted under the `*_scale_from_checkpoint` naming the dense MXFP8 path already uses (including `pad_flashinfer_scale_k` handling).
2. Verification that per-expert MXFP8 scale entries route correctly through vLLM's `RoutedExperts` expert mapping with the ModelOpt MXFP8 MoE method (the blockwise path needed per-expert re-emission because the fused loader has no scale path — the same question applies here).
3. The same exclusion behavior as blockwise: `quantization_ignored_layer_kws` (GDN projections + vision tower) and bf16 pass-through for ignored experts.
4. Tests and a recipe: extend the grouped-expert unit tests in `tests/unit/models/generation/test_vllm_fp8_quantization.py` (currently a `pytest.raises` guard for MXFP8) with value-level checks, and add/validate an MXFP8 recipe variant once the path works.

PR #2744, opened 2026-06-09
fix(fp8): support Qwen3.5 blockwise FP8 rollout and training

MarkXCloud:
Summary:
- Add Qwen3.5 DAPO Megatron recipes for BF16 and FP8 rollout/training.
- Fix grouped MoE expert FP8 refit for weights that do not end with .weight, so block quantization is applied instead of skipped.
- Re-emit grouped MoE experts through vLLM's per-expert refit path so FP8 scale_inv tensors are loaded correctly.
- Support both vLLM Ray executor module paths for older and newer vLLM versions.
- Use Megatron Bridge provider freeze_config to keep unused VL vision parameters out of Megatron FP8 distributed training buckets for text-only runs.

PR #3724, opened 2026-08-20
feat(refit): support grouped MoE MXFP8 weights

seonjinn:
## Summary

Add MXFP8 rollout refit support for Qwen3.5 grouped MoE weights.

Split fused gate/up and down expert slabs into per-expert projections, quantize them to E4M3 values with E8M0 scales, and load them through vLLM's existing expert mapping. Scale names follow the legacy or native reload API as appropriate.

## Changes

- Support grouped MoE MXFP8 refit while preserving BF16 passthrough for excluded experts.
- Handle vLLM kernel factories with or without the `layer` argument.
- Add focused conversion tests and a Qwen3.5 recipe; its end-to-end test remains disabled pending validation.

This PR contains only the original grouped-MoE feature. Trainer-side prequantization, cached loader routes, and IPC optimizations from #3294 are excluded. The prerequisite grouped-FP8 support from #2744 is already on main.

## Validation

- After removing #3294, the code tree matches the original standalone head `c22a14d29`.
- CPU recipe checks: 3 passed, 1 skipped (vLLM unavailable).
- Ruff lint/format, shell syntax, and diff checks passed.
- The broader FP8 test suite could not be validated locally: 103 skipped and one existing test failed to import vLLM. GPU refit and end-to-end validation of this standalone revision remain pending.

Addresses #3694.


seonjinn:
Self-review summary (agent team + adversarial verification over the delta unique to this PR vs the #2744 base — fp8.py grouped-MXFP8 path, the test, and the doc line).

Verified clean: the fused gate/up slab split and expert iteration are byte-identical to the already-validated blockwise-FP8 sibling (`_expand_grouped_moe_expert_to_fp8`); the E8M0 math is fully delegated to the existing `quantize_mxfp8_weight` helper (including the zero-scale byte-0 -> byte-1 workaround), so no new quantization math is introduced; the `*_scale_from_checkpoint` naming matches both the non-grouped MXFP8 path and the NCCL-reshard receiver; per-expert `.contiguous()` keeps refit peak memory at ~1/num_experts of the slab, same philosophy as the FP8 sibling. A suspected ordering hazard (the per-expert scale params being lazily created only by `process_weights_after_loading_mxfp8_moe`) was traced and refuted: the fp8 patches install via the patched `collective_rpc` before the first worker RPC, so the params exist before any refit.

Fixed in the follow-up commit: the new test compared bf16 mock inputs against fp32 source slices with `torch.equal` (always False on dtype mismatch), and used a `fp32 -> e4m3` expectation while production rounds `fp32 -> bf16 -> e4m3` — the test as previously committed would have failed on its first real run (unit tests haven't executed on this head yet). Also one pre-existing ruff-0.9.9 format violation in the same file.

Before undrafting:
- Strengthen the mock: the value check currently validates routing/plumbing against a bare-cast fake; a small CPU reference E8M0 quantizer (mirroring the FP8 sibling's `_assert_dequant_close` approach) would make it a real numerics test.
- Rebase onto the current #2744 tip (3 commits behind; none touch this path, so trivial).
- The already-planned GPU unit run + Qwen3.5 e2e.

docs/guides/models/qwen/qwen3-5.md
```diff
@@ -64,6 +64,7 @@ authoritative settings.
 | Qwen3.5-9B-Base | LLM | GRPO | Megatron | 1n8g | [`grpo-qwen3.5-9b-1n8g-megatron-fp8.yaml`](../../../../examples/configs/recipes/llm/grpo-qwen3.5-9b-1n8g-megatron-fp8.yaml) |
 | Qwen3.5-35B-A3B-Base | LLM | GRPO | Megatron | 2n8g | [`grpo-qwen3.5-35ba3b-2n8g-megatron-ep16tp2cp2.yaml`](../../../../examples/configs/recipes/llm/grpo-qwen3.5-35ba3b-2n8g-megatron-ep16tp2cp2.yaml) |
 | Qwen3.5-35B-A3B-Base | LLM | GRPO | Megatron | 2n8g | [`grpo-qwen3.5-35ba3b-2n8g-megatron-ep16tp2-fp8.yaml`](../../../../examples/configs/recipes/llm/grpo-qwen3.5-35ba3b-2n8g-megatron-ep16tp2-fp8.yaml) |
+| Qwen3.5-35B-A3B-Base | LLM | GRPO | Megatron | 4n4g | [`grpo-qwen3.5-35ba3b-4n4g-megatron-ep16tp2-mxfp8-trtllm.yaml`](../../../../examples/configs/recipes/llm/grpo-qwen3.5-35ba3b-4n4g-megatron-ep16tp2-mxfp8-trtllm.yaml) |
```
yuki-97: The row added at line 67 joins the table this page introduces as the **Long-Run Convergence Validated** set ([qwen3-5.md:26-30](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/docs/guides/models/qwen/qwen3-5.md#L26-L30)), but this PR's own [`disabled.txt:81-84`](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/tests/test_suites/disabled.txt#L81-L84) keeps the recipe manual "until an end-to-end run establishes its gen_kl_error/reward bounds", and its driver caps at [`MAX_STEPS=20`](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/tests/test_suites/llm/grpo-qwen3.5-35ba3b-4n4g-megatron-ep16tp2-mxfp8-trtllm.sh#L8) rather than the full-length run the stage definition at [:24](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/docs/guides/models/qwen/qwen3-5.md#L24) requires.

The one other table row currently in `disabled.txt` is there for a vLLM hang, not for missing bounds — so the table's convention today is "validated, one of them currently red", not "anything we ship".

**Suggest fix:**

1. Mark line 67 as not yet convergence-validated — a footnote under the table or a status marker on the row — so it agrees with `disabled.txt`.
2. Or hold the row out until the validation run lands.

docs/guides/models/qwen/qwen3-5.md
```diff
@@ -247,8 +248,12 @@ FP8 noise.
...
 ### Limitations
 
-- MXFP8 is not yet supported; tracked in
-  [#3694](https://github.com/NVIDIA-NeMo/RL/issues/3694).
+- Dense Qwen3.5 MXFP8 rollout has not been validated end to end.
+- Grouped-MoE MXFP8 refit is supported through both weight-transfer paths:
+  NCCL Reshard uses the receiver-side conversion added in
+  [#3477](https://github.com/NVIDIA-NeMo/RL/pull/3477), while colocated CUDA IPC
+  and legacy reloads expand fused expert slabs before calling vLLM's native
+  per-expert loader.
```
yuki-97: nit: bullet 2 states a capability rather than a limitation, and reading it next to bullet 1 implies the grouped-MoE path *is* validated — while [`disabled.txt:81-84`](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/tests/test_suites/disabled.txt#L81-L84) keeps that run manual until its bounds exist. The mechanism claim itself checks out; it is the placement and the missing validation caveat.

It also drops the last in-tree pointer to [#3694](https://github.com/NVIDIA-NeMo/RL/issues/3694), which is still open and whose body covers the dense gap bullet 1 names.

**Suggest fix:**

1. Keep a limitation bullet saying the grouped-MoE MXFP8 path is implemented but not yet validated end to end, and move the two-path mechanism sentence into the FP8 section body.
2. Re-link #3694 on the dense bullet.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -854,6 +859,31 @@ def _expand_grouped_moe_expert_to_fp8(key, weight):
     return entries
 
 
+def _expand_grouped_moe_expert_to_mxfp8(
+    key: str, weight: torch.Tensor, *, refit_with_reload_api: bool
+) -> list[tuple[str, torch.Tensor]]:
+    """Expand a grouped Qwen3.5 MoE slab into per-expert MXFP8 entries."""
+    base, proj = key.rsplit(".", 1)
```
yuki-97: nit: lines 866-874 are byte-identical to the split in [`_expand_grouped_moe_expert_to_fp8` at :839-847](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/nemo_rl/models/generation/vllm/quantization/fp8.py#L839-L847), so the grouped-slab layout contract — gate first, split at `shape[1] // 2` — is now asserted in two places with nothing tying them together.

**Suggest fix:** extract `_split_grouped_moe_shards(key, weight)` returning `(base, shards)` and call it from both; the per-expert loop, the quantizer choice, the `.contiguous()` decision and the scale suffix all stay at the call sites.

tests/test_suites/disabled.txt
```diff
@@ -78,5 +78,10 @@ tests/test_suites/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation-colocated-resh
 tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_generation-noncolocated-async-gym-quick.sh
 tests/test_suites/llm/grpo-nanov3-30BA3B-4n4g-megatron_generation-colocated-reshard-async-gym-quick.sh
 
+# Qwen3.5 grouped-MoE MXFP8 validation. Keep this 4x4-GPU, 20-step run
+# manual until an end-to-end run establishes its gen_kl_error/reward bounds
+# and the recurring GB200 suite has budget.
+tests/test_suites/llm/grpo-qwen3.5-35ba3b-4n4g-megatron-ep16tp2-mxfp8-trtllm.sh
```
yuki-97: Nothing in the PR shows this path working end to end — this entry keeps the only e2e driver manual until its bounds exist, and the PR body lists GPU refit and end-to-end validation as pending.

**Suggest fix:** run the 20-step validation once and paste the result into the PR description (reward, loss and `gen_kl_error`), so the thresholds at [`...mxfp8-trtllm.sh:36-37`](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/tests/test_suites/llm/grpo-qwen3.5-35ba3b-4n4g-megatron-ep16tp2-mxfp8-trtllm.sh#L36-L37) are measured for this recipe rather than inherited from the 2n8g block-FP8 one, and the row added to the [Example Recipes table](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/docs/guides/models/qwen/qwen3-5.md#L67) is backed by a run.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -1149,6 +1179,18 @@ def create_weights_mxfp8_moe(
     )
 
 
+def _make_fp8_moe_kernel_compat(make_fp8_moe_kernel, layer, **kwargs):
```
yuki-97: #4080 touches both of this helper's call sites and resolves the same problem by dropping `layer=` unconditionally, with the pin at 0.29. Once this branch rebases onto that, the shim reproduces #4080's plain call through `inspect.signature`, and its `if accepts_layer:` arm ([:1188](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/nemo_rl/models/generation/vllm/quantization/fp8.py#L1188)) can no longer be taken on any pinned version.

**Suggest fix:** at rebase time, take #4080's call at [`:1255`](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/nemo_rl/models/generation/vllm/quantization/fp8.py#L1255) and [`:1504`](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/nemo_rl/models/generation/vllm/quantization/fp8.py#L1504), and drop this helper, the `import inspect` at [:15](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/nemo_rl/models/generation/vllm/quantization/fp8.py#L15) (its only other use), and [`test_make_fp8_moe_kernel_compat_matches_vllm_signature`](https://github.com/NVIDIA-NeMo/RL/blob/3a88b8863395973036cdd96fe08d3e18d49501b3/tests/unit/models/generation/test_vllm_fp8_quantization.py#L188).

yuki-97 (commented):
Thanks for supporting this. The PR lifts the grouped-MoE block on MXFP8 refit by expanding the fused expert slab per-expert into the `gate_proj` / `up_proj` / `down_proj` params vLLM actually registers, which lets the pre-refit rejection guard go away, and ships a Qwen3.5-35BA3B EP16/TP2 recipe on top of it.

**Scope:** the grouped-MoE MXFP8 expansion in `quantization/fp8.py`, the removed pre-refit guard, and the new recipe / driver / doc surface.

**Checked:**

- The expansion is anchored to what vLLM does rather than to what the layout looks like — the gate/up split at `shape[1] // 2` matches vLLM's own `chunk(2, dim=1)` de-fusion with index 0 as `w1`/`gate_proj`, and both scale suffixes land on registered params through `RoutedExperts`' prefix rewrite.
- Keeping the new branch inside `_is_fp8_grouped_moe_expert` is the other half of that: experts excluded by `num_{first,last}_layers_in_bf16` still pass through as the untouched bf16 slab, so the rewrite is scoped to the experts vLLM actually built as FP8.
- Every weight-transfer path the deleted `assert_refit_unsupported_grouped_moe_params` used to gate. Colocated IPC, collective legacy, collective native-reload and non-sharded checkpoint-engine all reach the new expander; sharded NIXL still rejects FP8 models outright; and NCCL reshard was never covered by that guard in the first place, since it implies non-colocated and therefore skips `prepare_refit_info` entirely.
- The vLLM calls against the pinned source at 0.25.1 and 0.26.0, and the kernel factory at every release through 0.30.0 — signature, kwarg names, return shape.

**Remaining comments** are on end-to-end evidence for the new recipe and the overlap with #4080.

**nit, not blocking merging** — the doc puts the grouped-MoE capability under `### Limitations`, and the slab layout contract is now spelled out in two places.

PR #3725, opened 2026-08-20
[Do not merge into Main] feat(draft): add bounded multi-pass Eagle TTT 

seonjinn:
## Summary

- Adds a bounded multi-pass EAGLE-3 TTT provider.
- Uses structured MCore attention without retaining dense pass logits or sequence-square masks.
- Validates rotary-position support when the session is created.

## Scope

Covers the provider, pass state, attention, loss, and split-step training primitives. Higher-level configuration and recipes remain out of scope.

## Validation

- Exact head: `fdf2a971be0b396ed7282a9a99c79a6af24a5e7b`
- Ruff and Pyrefly pass.
- Focused CPU suites: 50 passed / 11 CUDA-skipped; TP suite: 2 passed / 1 CUDA-skipped.

## Prior GPU evidence

- Prior 4×GB200 focused suite: 139 passed with 1 optional-dependency skip.

## Dependencies / Remaining

Depends on #3703 and #3706. CUDA memory cases remain GPU-only validation.


nemo_rl/models/megatron/draft/eagle_ttt.py
```diff
@@ -0,0 +1,1547 @@
...
+        packed_seq_params: object | None,
+    ) -> Tensor:
+        del attn_mask_type
+        if self.layout is None or self.plan is None or self._storage_plan is None:
+            raise RuntimeError(
+                f"EAGLE TTT layer {self.layer_number} called outside an active pass"
+            )
+        if self._called:
+            raise RuntimeError(
+                f"EAGLE TTT layer {self.layer_number} executed twice in one pass"
```
seonjinn, line 691: The one-shot pass lifecycle (`begin_pass` -> exactly-one `forward` -> `finish_pass`) is incompatible with activation recompute. Empirically reproduced on CPU (torch 2.13): wrapping an armed `EagleTTTCoreAttention.forward` in `torch.utils.checkpoint.checkpoint` and calling `backward()` raises `RuntimeError: EAGLE TTT layer 0 called outside an active pass` (or `executed twice in one pass` when recompute fires before `finish_pass`), for both `use_reentrant` modes; the same forward/backward without checkpoint passes.

Since this module slots in as `core_attention` inside a real MCore `SelfAttention`, both `recompute_granularity="full"` (the convention we already rely on at 235B scale to avoid activation OOM) and `"selective"` (which checkpoints core attention specifically) re-enter this forward during backward and crash deterministically on step 1. Idempotent re-entry is not a trivial fix -- a second forward would re-run `from_trunk`/`append_branch` and corrupt `self.state` -- so failing fast seems right: at `MCoreEagleTTTSession.begin` (or arm time), inspect the layer config and raise a clear error when any recompute granularity is enabled, and document the constraint. A ~30-line CPU unit test pins the behavior.

nemo_rl/models/megatron/draft/eagle_ttt.py
```diff
@@ -0,0 +1,1547 @@
...
+    if query.shape[2] != plan.sequence_length:
+        raise ValueError("query sequence length must match the attention plan")
+    if query.device != state.trunk_key.device or query.dtype != state.trunk_key.dtype:
+        raise ValueError("query and retained state must share device and dtype")
+    if layout is not None:
+        if layout.valid_tokens.shape != (query.shape[0], query.shape[2]):
+            raise ValueError("sequence layout must match query batch and sequence axes")
+        if layout.valid_tokens.device != query.device:
+            raise ValueError("sequence layout and query must share a device")
+    _expand_gqa(state.trunk_key, query_heads=query.shape[1])
```
seonjinn: `_expand_gqa(state.trunk_key, query_heads=query.shape[1])` discards its result -- it exists only to raise early on a head-count mismatch, but pays a full `repeat_interleave` copy of the trunk key (`[batch, query_heads, seq, head_dim]`) on every pass of every layer, on both backends (it runs before the `query.is_cuda` branch). At the sequence lengths this module targets (tests go to 262k) that's a multi-hundred-MB transient in a file whose whole premise is bounding memory. Both downstream paths already handle GQA themselves (dense re-expands internally, flex uses `enable_gqa`, `_merge_branch` reshapes into groups), so a modulo check suffices:

```python
if query.shape[1] % state.trunk_key.shape[1] != 0:
    raise ValueError("query heads must be divisible by key/value heads")
```

nemo_rl/models/megatron/draft/eagle_ttt.py
```diff
@@ -0,0 +1,1547 @@
...
+    def from_trunk(
+        cls,
+        *,
+        trunk_key: Tensor,
+        trunk_value: Tensor,
+        pass_count: int,
+        max_passes: int,
+        activation_budget_bytes: int,
+    ) -> EagleTTTState:
+        """Validate the complete bound before retaining any supplied tensor."""
```
seonjinn: "Validate the complete bound before retaining any supplied tensor" overstates what this check does: the `EagleTTTStoragePlan` built here uses defaults `layer_count=1, hidden_size=0, rope_dim=0`, so it validates only this layer's multi-pass K/V plus mask metadata against the full budget. The complete bound (all layers, hidden, rope, loss) is enforced by the provider's `_storage_plan` and, at runtime, by the resource ledger. Suggest rewording to "Validate the per-layer K/V bound..." so a future edit doesn't mistake this for the enforcement point.

seonjinn (commented):

Method: six review agents over the PR8a-only diff (8 files vs the #3703+#3706 base), then an adversarial pass that re-verified citations, ran two empirical checks on CPU torch, and re-checked CI/PR state live.

The core math held up well under adversarial verification: the multi-pass staircase visibility and the RoPE-relative shift alignment in `_merge_branch` were re-derived against an independently written rotate-half oracle and match to machine precision (4.4e-16, fp64); label alignment (`teacher_offset = pass_index + 1`) was re-worked by hand; the cross-pass detach asymmetry is pinned by an explicit test; and the TP/SP integration follows the clean pattern -- `EagleTTTCoreAttention` owns zero parameters, so the sharded-checkpoint bug classes from earlier PRs in this series are structurally inapplicable. The SP test (real NCCL `SelfAttention` with `sequence_parallel=True`, output + input grads + parameter grads vs a global dense oracle) is the strongest distributed test in the series so far.

The one substantive finding is the activation-recompute incompatibility (inline, empirically reproduced): with `recompute_granularity=full` or `selective`, backward re-enters the armed forward and crashes on step 1.

**Process, before merge:**

1. **No `/ok to test` has ever been issued on this PR** -- zero substantive CI on any commit. Needs a fresh `/ok to test 292e29d31a6276c0e4823c37d94f7c6e917f2fc1` (after the rebase below, on the new head).
2. **Rebase required:** both stacked dependencies have *diverged* (not just advanced) -- #3706's head (`40badeea`) and #3703's head (`5133c28d`) no longer contain this PR's base `660ee3c6`, so the rebase may not be clean.
3. Body accuracy: the Ruff/pyrefly/4xGB200 numbers are self-reported (no check-runs exist); "139 passed" spans the full 13-file stacked suite, not this PR's 5 test files (~91 collected) -- worth stating the basis.
4. Below inline threshold: `MCoreEagleTTTSession` is never driven past `pass_count=2` off-GPU, and the "pinned_modelopt" test name pins an internal formula, not an external ModelOpt oracle -- consider a rename or a small modelopt-driven geometry diff test later.

nemo_rl/models/megatron/draft/eagle_ttt.py
```diff
@@ -0,0 +1,1547 @@
...
+        try:
+            for layer in self.layers:
+                layer.arm(
+                    layout=session_layout,
+                    storage_plan=session_storage_plan,
+                    resource_ledger=resource_ledger,
+                    packed_seq_params=packed_seq_params,
+                )
+                armed.append(layer)
+            rotary_module = getattr(self.model, "rotary_pos_emb", None)
```
seonjinn: `getattr(self.model, "rotary_pos_emb", None)` silently proceeds with no rotary table when the attribute is absent or renamed. The branch-merge shift itself stays self-consistent without RoPE (it's a pure index gather), but a RoPE-configured model that loses this attribute in a refactor would silently train un-rotated -- a quality-only, no-crash divergence from the target. Since every intended wiring constructs rotary unconditionally, consider requiring it (`raise ValueError("EAGLE TTT session requires model.rotary_pos_emb")`) or an explicit `allow_no_rope` opt-out.

tests/unit/distributed/test_eagle_ttt_sp.py
```diff
@@ -0,0 +1,640 @@
...
+        assert [flag.item() for flag in gathered_flags] == [1] * world_size
+        assert error == "TP ranks must agree on EAGLE TTT packed sequence parameters"
+        session.reset()
+    finally:
+        if parallel_state.model_parallel_is_initialized():
+            parallel_state.destroy_model_parallel()
+        dist.destroy_process_group()
+
+
+@pytest.mark.skipif(
```
seonjinn, line 591: These real-MCore tests (three here, plus the two TP-session tests in `test_eagle_ttt_tp.py`) gate only on NCCL and `device_count >= 2`, but import `megatron.core` unconditionally once the skipif passes -- on a 2-GPU shard without the mcore extra they error instead of skipping. The repo registers an `mcore` marker and the sibling contract test in this PR carries it; adding `@pytest.mark.mcore` here keeps the gating consistent.

tests/unit/models/megatron/test_eagle_ttt_attention.py
```diff
@@ -0,0 +1,785 @@
...
+            expected_gradient,
+            atol=8e-2,
+            rtol=8e-2,
+        )
+        assert actual_gradient.isfinite().all()
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is required")
+@pytest.mark.parametrize("sequence", [8_192, 32_768])
+def test_cuda_long_context_has_no_saved_sequence_square_tensor(
```
seonjinn: This is the only byte-measured memory test, and it runs `pass_count=1` -- zero branches -- so a per-branch O(seq^2) regression under multi-pass would be invisible to it, and the multi-pass CUDA tests check saved-tensor *shapes*, which a differently-shaped quadratic leak evades. Since "bounded multi-pass" is this PR's headline, consider parametrizing the byte-measured assertion over `pass_count` in {1, 4}.

seonjinn:
Merged latest `main` (`b03da0f469`) and re-fired CI on `be80ab656b`.

`main` now carries the typed speculator contract from #3701, so this branch had to be adapted rather than merged verbatim. Three changes, one of which git merged clean but broke at runtime:

**1. `nemo_rl/models/policy/__init__.py` — took `main`'s deletion.**
#3701 removed the `DraftConfig` / `DraftConfigDisabled` `TypedDict`s and replaced them with `Eagle3DraftConfig(BaseModel, extra="allow")` in `nemo_rl/models/policy/draft_config.py`. This branch's only addition to that file was the `token_chunk_size: NotRequired[int]` field, so the whole conflicted block is dropped in favour of the new contract. `policy.draft.token_chunk_size` still validates and still reaches the loss wrapper, as a pydantic extra.

**2. `pyrefly.toml` — sorted union.**
Both sides inserted into the same sorted list under `nemo_rl/models/megatron/draft/`: `eagle_provider.py` / `eagle_ttt.py` / `step_state.py` from here, `training.py` / `utils.py` from `main`.

**3. `nemo_rl/models/megatron/train.py` — `.get()` no longer exists on the draft config.**
This one is worth calling out because it was **not** a merge conflict. `policy_cfg["draft"]` used to be a `TypedDict`; after #3701 it is a pydantic model. Two sites here still read

```python
self.cfg["draft"].get("token_chunk_size", DEFAULT_DRAFT_TOKEN_CHUNK_SIZE)
```

which git merged clean, passes lint, and raises `AttributeError` only on a draft-enabled run. Rewritten as `getattr(self.cfg["draft"], "token_chunk_size", DEFAULT_DRAFT_TOKEN_CHUNK_SIZE)`, which preserves the exact prior semantics.

**Known gap, deliberately not fixed here.** `token_chunk_size` is left undeclared on `Eagle3DraftConfig`, so #3701's `_reject_near_miss_extra_keys` validator cannot catch a typo of it - `token_chunk_sze` would silently fall back to 4096 instead of raising. Declaring the field properly would mean importing `DEFAULT_DRAFT_TOKEN_CHUNK_SIZE` from `nemo_rl/algorithms/loss/draft.py`, which pulls torch into `draft_config.py` - a module #3701 deliberately kept to pydantic and stdlib only. Happy to declare it with a literal default here instead if reviewers prefer the typo protection over the duplicated constant.

PR #3726, opened 2026-08-20
feat(draft): add DSpark tiled objectives

seonjinn:
## Summary

- Adds tiled hard-CE, total-variation, and confidence objectives for DSpark.
- Trains hard CE on sampled draft labels and applies position weights to numerators and denominators.
- Keeps target logits and the live output head stop-gradient.
- Supports split vocabularies, bounded-memory chunking, and synchronized TP contract failures.

## Scope

Covers DSpark objectives, heads, providers, and distributed objective tests. Training-loop configuration is handled downstream.

## Validation

- Exact head: `42cefa8113a2d79dd1efe0dc81fe4b90fa2f6273`

## Prior focused evidence

- OCI `6392604`: 42 focused tests passed. The local objective suite passed 32 tests; earlier TP/DP and real-MCore checkpoint coverage also passed on 4×GB200.

## Dependencies

Depends on #3703 and #3711.


tests/unit/algorithms/test_dspark_objective.py
```diff
@@ -0,0 +1,931 @@
...
+@pytest.mark.parametrize(
+    "loss_weights",
+    [
+        pytest.param((1.0, 0.0, 0.0), id="ce_only"),
+        pytest.param((0.0, 1.0, 0.0), id="tv_only"),
+        pytest.param((0.0, 0.0, 1.0), id="confidence_only"),
+        pytest.param((1.75, 0.375, 2.25), id="mixed_non_unit"),
+    ],
+)
+def test_dspark_objective_matches_dense_component_oracles(
```
seonjinn, line 368: The only unequal-tile test checks storage retention, not values, and this oracle test runs a fixed `token_chunk_size=2` over 6 rows. The adversarial pass empirically confirmed the code IS chunk-invariant (chunks {1,2,4,7,15,100}: bit-equal numerators, counts, and all grads), so this is purely a regression guard -- parametrizing this test over `token_chunk_size in (1, 2, 4, 100)` catches a future off-by-one in the `end = min(start + chunk, N)` boundary or a boundary-scatter bug. Applied in the follow-up commit.

nemo_rl/models/megatron/draft/dspark_provider.py
```diff
@@ -0,0 +1,320 @@
...
+            valid_mask=valid_mask,
+            slot_bins=slot_bins,
+            slot_weights=slot_weights,
+            configured_tp_group=self.markov_head.tensor_parallel_group,
+            requested_tp_group=tp_group,
+        )
+        if self.markov_head.tensor_parallel_group is not tp_group:
+            raise ValueError("objective TP group must match the DSpark head TP group")
+        if draft_hidden.ndim != 3:
+            raise ValueError("draft_hidden must have shape [blocks, slots, hidden]")
```
seonjinn, line 183: Five raise sites here have no test coverage (grep of all five message strings over tests/ returns zero hits): the ndim/matrix/vocab-shape/dtype/device guards in `objective_stats` and the `confidence_with_markov requires confidence_enabled` factory guard. All are single-rank CPU-testable; five `pytest.raises` cases added in the follow-up commit.

nemo_rl/algorithms/loss/dspark.py
```diff
@@ -0,0 +1,945 @@
...
+            )
+            tile_embedding_gradient = torch.zeros(
+                (end - start, markov_rank),
+                dtype=torch.float32,
+                device=draft_hidden.device,
+            )
+            previous_embeddings_fp32 = previous_embeddings.detach().float()
+            for vocab_start in range(
+                0,
+                local_vocab_size,
```
seonjinn, line 561: The inner vocab-gradient-chunk loop reused the name `vocab_start` -- the same name as the shard-offset the rest of `backward` reads via `ctx.vocab_start`. All live reads were correct, but a future `ctx.`-drop typo would silently pick up a stale chunk offset instead of failing. Renamed to `chunk_start`/`chunk_end` in the follow-up commit.

nemo_rl/models/megatron/draft/dspark_provider.py
```diff
@@ -0,0 +1,320 @@
...
+        confidence_logits = None
+        if self.confidence_head is not None:
+            markov_embeddings = None
+            if self.confidence_head.with_markov:
+                safe_previous_token_ids = torch.where(
+                    valid_mask,
+                    previous_token_ids,
+                    torch.zeros_like(previous_token_ids),
+                ).clamp(0, self.markov_head.target_vocab_size - 1)
+                markov_embeddings = self.markov_head.markov_w1(safe_previous_token_ids)
```
seonjinn, line 216: This path reads `markov_w1` directly without `_CopyToTensorParallelRegion`, and that is correct by construction -- every input here is TP-replicated full-sequence, so each rank computes the identical, complete gradient; wrapping would over-count `markov_w1.grad`'s confidence component by the TP world size (the TP2 test asserts the grad against a dense oracle, and a factor-2 error is far outside its rtol). The invariant deserved a comment, added in the follow-up commit: if a future change (e.g. PR11's sequence-sharded packing) delivers sharded `draft_hidden`, the metadata-agreement and slot-shape checks in `dspark_tiled_objective` fail loudly before anything silent can happen.

seonjinn (commented):

Method: five review agents over the PR8b-only diff (6 files vs base `a38c9e281`), then an adversarial pass that re-verified citations, git-verified all stacking claims, and empirically executed the loss on CPU torch (chunk-size invariance and the normalization algebra).

The math survived every adversarial check: the position-weight denominator symmetry (`normalized()` applies weights to numerators and counts alike -- re-derived and executed on a 2-bin example), the TV shard-sum exactness under vocab-parallel TP, the hard-CE and TV gradients (hand-derived, match the code), stop-gradient placement on `target_logits` and the live head at every use site, and the additive-DP-stats contract (which, unlike earlier PRs in this series, is actually exercised by a DP2 test that all-reduces counts, including a zero-valid-slot rank). The confidence path's direct `markov_w1` read -- the exact pattern flagged as unsafe on #3711 -- was adjudicated by three agents and the adversarial pass: it is correct here because all inputs are TP-replicated (see the inline note), and wrapping it would over-count by the TP world size.

A follow-up commit applies the four small findings: chunk-size parametrization of the dense-oracle test, five raise-site tests, a shadowed-name rename in `backward`, and the TP-replication invariant comment.

**Process, before merge:**

1. **Rebase onto #3711's current head (`54a3d9b4e`).** This stack's `models/megatron/draft/dspark.py` is the pre-fix copy: `markov_w2` lacks `set_tensor_model_parallel_attributes` (grad-norm undercount under TP clipping) and `sharded_state_dict` lacks the tp_group/dp_cp_group fallback. The GPU results quoted in the body were produced against the stale copy; the unit tests assert grads directly (not clip norms), so they should survive the rebase, but re-run the TP suite after.
2. `/ok to test` has never been issued on this PR -- no CI has run the suite on any commit. Issue it on the post-rebase head.
3. Reconcile the "42 focused tests on OCI" figure (static collection of the two GPU files is 20; the "32 local" figure reconciles exactly).
4. The #3703 dependency is fine, contrary to first appearances: its content is embedded byte-identically as rebased commits in the base. Optionally say "dependencies embedded by rebase" in the body.
5. 30-commit fixup history -- squash before merge.

PR #3727, opened 2026-08-20
feat(sc): support non-colocated MInf

tdene:
# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -312,6 +310,65 @@ def _build_trainer(
     return trainer, time.perf_counter() - t0
 
 
+def _build_trainer_then_megatron_generation(
```
yuki-97: **[BUG]** The new functional test is red on both SC shards: [_spinup_gym](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/algorithms/single_controller_utils/setup.py#L397) raises `ConfigError: 1 model endpoint(s) never answered within 600s`.

Both lanes look the same in the logs. The reserved URL is published, the trainer worker finishes within ~10s, and then nothing is logged until the timeout — `MegatronPolicyWorker` appears exactly once in each run, so no generation-side worker is ever created and nothing binds the reserved port.

| | URL reserved | trainer init_start → init_complete | ConfigError |
|---|---|---|---|
| h100 | 17:38:22 | 17:39:01 → 17:39:12 | 17:48:48 |
| gb200 | 17:00:18 | 17:00:47 → 17:00:53 | 17:10:38 |

(Timestamps from the `INFO:nemo_rl.utils.nvml` copies of the `[GPU_DIAG]` markers; the bare-print copies of the same markers are flushed at actor teardown and read ~9 minutes later.)

tests/functional/grpo_megatron_generation_gym_single_controller.sh
```diff
@@ -0,0 +1,116 @@
...
+    data.train.data_path=$TRAIN_PATH \
+    data.validation.data_path=$VALIDATION_PATH \
+    ++data_plane.enabled=true \
+    ++data_plane.impl=transfer_queue \
+    ++data_plane.backend=simple \
+    ++data_plane.storage_capacity=1000000 \
+    ++data_plane.num_storage_units=2 \
+    ++data_plane.claim_meta_poll_interval_s=0.5 \
+    ++data_plane.global_segment_size=549755813888 \
+    ++data_plane.local_buffer_size=68719476736 \
```
yuki-97: Same `data_plane` nesting problem as the exemplar (https://github.com/NVIDIA-NeMo/RL/pull/3727#discussion_r3859137778) — after ffbf33f30 (#2935) these flat keys are never read, so `simple.num_storage_units` is absent and TQ bootstrap fails once the readiness timeout is fixed. Rationale in that thread; the fix here:

```suggestion
    ++data_plane.simple.storage_capacity=1000000 \
    ++data_plane.simple.num_storage_units=2 \
    ++data_plane.claim_meta_poll_interval_s=0.5 \
```

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
```diff
@@ -0,0 +1,70 @@
...
+# (values as documented in examples/configs/grpo_math_1B.yaml).
+data_plane:
+  enabled: true
+  impl: transfer_queue
+  backend: "simple"                  # TQ storage backend ('simple' or 'mooncake_cpu')
+  storage_capacity: 1000000          # max samples retained per partition
+  num_storage_units: 2               # storage shards
+  claim_meta_poll_interval_s: 0.5    # blocking-claim poll cadence
+  global_segment_size: 549755813888  # 512 GiB — used when backend == "mooncake_cpu"
+  local_buffer_size:   68719476736   # 64 GiB  — used when backend == "mooncake_cpu"
```
yuki-97: **[BUG]** This block was correct when you wrote it — it is a verbatim copy of `grpo_math_1B.yaml`'s pre-nesting form, comments and alignment included. ffbf33f30 (#2935, Aug 23) moved these keys under `simple:` / `mooncake_cpu:` and left the flat spelling permanently unread; your branch picked that commit up in the Aug 25 rebase, but the config did not follow.

With `backend: simple` and no `simple:` block, [backend_config](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/data_plane/interfaces.py#L148) validates `SimpleStorageConfig` against `{}`, and [num_storage_units](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/data_plane/interfaces.py#L58) has no default — TQ bootstrap fails.

#2935 also re-based the two mooncake values to *per client process* and shrank them (512 GiB / 64 GiB → 64 GiB / 4 GiB) and added `reuse_registered_buffers` / `staging_buffer_size`, so the copied numbers are stale even read as mooncake values. The suggestion is a verbatim mirror of [grpo_math_1B.yaml:545-559](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/examples/configs/grpo_math_1B.yaml#L545-L559), which is what the header comment on line 45 already promises — including `${mul:2, ${cluster.num_nodes}}`, whose resolver is registered globally and already used by the gym base itself.

Same edit on the `++data_plane.*` overrides at [the functional test's :98-102](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/tests/functional/grpo_megatron_generation_gym_single_controller.sh#L98-L102) (https://github.com/NVIDIA-NeMo/RL/pull/3727#discussion_r3859137769); #2935 swept the sibling scripts already — [grpo_async_gym_single_controller.sh:97-98](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/tests/functional/grpo_async_gym_single_controller.sh#L97-L98).

```suggestion
  claim_meta_poll_interval_s: 0.5    # blocking-claim poll cadence
  # Backend-specific blocks: only the one named by `backend` is read, and an
  # absent block means that backend's defaults (see SimpleStorageConfig /
  # MooncakeCpuConfig in nemo_rl/data_plane/interfaces.py).
  simple:
    storage_capacity: 1000000        # max samples retained per partition
    num_storage_units: ${mul:2, ${cluster.num_nodes}}  # TQ wants >= 2 per node
  mooncake_cpu:
    # Per client process — see MooncakeCpuConfig before raising these.
    global_segment_size: 68719476736 # 64 GiB/process
    local_buffer_size:    4294967296 # 4 GiB/process
    reuse_registered_buffers: true   # reuse RDMA-registered buffers
    staging_buffer_size:   268435456 # 256 MiB/pool slot; bigger transfers bypass the pool
  # observability:                    # NotRequired
  #   enabled: false
```

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
```diff
@@ -0,0 +1,70 @@
...
+  sampler:
+    name: in_order
+    # 0 = fully synchronous. No off-policy data is admitted, so the inherited
+    # use_importance_sampling_correction=false stays valid.
+    max_lookahead_versions: 0
+  min_groups_for_streaming_train: ${grpo.num_prompts_per_step}
+  max_inflight_prompts: ${grpo.num_prompts_per_step}
+
+checkpointing:
+  # SC does not support checkpointing yet (setup raises when it is enabled).
```
yuki-97: This comment contradicts the code, and the guide's own limitation bullet: [main's own bullet](https://github.com/NVIDIA-NeMo/RL/blob/d938da394063bf32381d84d819c60f8826eaa66b/docs/guides/single-controller.md#L180) now reads "Validation is not yet supported (setup raises on `val_period > 0`, `val_at_start`, or `val_at_end`)" — #3773 dropped the checkpointing claim independently.

[setup_single_controller](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/algorithms/single_controller_utils/setup.py#L645-L649) only raises for validation, and builds a `CheckpointManager` unconditionally — the doc is right and this comment is stale.

```suggestion
```

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -389,6 +446,54 @@ def _maybe_inject_megatron_train_iters(master_config: MasterConfig) -> None:
     policy_config["megatron_cfg"]["train_iters"] = grpo_config.max_num_steps
 
 
+def _maybe_apply_megatron_generation_overrides(
```
yuki-97: `async_rl.generation_fleet_health.enabled=true` on the megatron backend is knowable from config alone, but it is only rejected at [_maybe_attach_fleet_health](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/algorithms/single_controller_utils/setup.py#L930) via [the base NotImplementedError](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/models/generation/interfaces.py#L480-L483) — by which point both worker groups and the HTTP server are already built.

This function is the pure-config gate, so rejecting it next to the colocated check above makes it free.

tests/functional/grpo_megatron_generation_gym_single_controller.sh
```diff
@@ -0,0 +1,116 @@
...
+    policy.dtensor_cfg.enabled=false \
+    policy.megatron_cfg.enabled=true \
+    policy.megatron_cfg.tensor_model_parallel_size=1 \
+    policy.megatron_cfg.pipeline_model_parallel_size=1 \
+    policy.megatron_cfg.expert_model_parallel_size=1 \
+    policy.megatron_cfg.context_parallel_size=1 \
+    policy.megatron_cfg.sequence_parallel=false \
+    policy.generation.backend=megatron \
+    policy.generation.mcore_generation_config.expose_http_server=true \
+    policy.generation.max_new_tokens=128 \
```
yuki-97, line 63: **[BUG]** These two lines are the first config anywhere to make `max_new_tokens != max_total_sequence_length`, and that combination exposes [_generation_max_seq_len](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/algorithms/single_controller_utils/setup.py#L421-L422): the megatron branch returns `max_new_tokens` where vLLM/SGLang return the context window, so `RolloutManager` gets 128 instead of 512.

Its docstring says megatron "has no dedicated field", but [max_model_len](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/models/generation/megatron/config.py#L34) is required and is what the engine receives — [megatron_worker.py:273](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/models/generation/megatron/megatron_worker.py#L273) is `max_sequence_length=mcore_generation_config["max_model_len"]`, while `max_new_tokens` is only `num_tokens_to_generate` at [:616](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/models/generation/megatron/megatron_worker.py#L616). On the Gym path this only skews `truncated` ([rollout_manager.py:1027](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/experience/rollout_manager.py#L1027)); on the native path [:533](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/experience/rollout_manager.py#L533) computes `max_env_tokens = self._max_seq_len - input_lengths - gen_token_count`, which goes negative and blanks every env observation.

The branch was unreachable before this PR (the deleted `_build_clusters` assert), so this is its first caller.

**Suggest fix:** return `generation_config["mcore_generation_config"]["max_model_len"]` and drop the docstring's third clause.

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
```diff
@@ -0,0 +1,70 @@
+# GRPO on the NeMo-Gym workplace-assistant environment via the SingleController path,
+# using non-colocated Megatron Inference.
+# Gym rollouts go through the persistent Megatron engine's OpenAI-compatible server,
+# Smoke-test scale; Qwen3-0.6B on one node with 2 GPUs (1 training + 1 generation);
+# the CI-run variant of this setup is
+# tests/functional/grpo_megatron_generation_gym_single_controller.sh.
+defaults: "grpo_qwen3_30ba3b_instruct.yaml"
```
yuki-97, line 9: A dense Qwen3-0.6B smoke recipe inheriting the 30B-A3B MoE base silently trains with **GSPO**, not GRPO.

[grpo_qwen3_30ba3b_instruct.yaml:15-27](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml#L15-L27) is explicitly MoE-scoped ("We use GSPO rather than GRPO for MoE models") and sets `sequence_level_importance_ratios: true`, `ratio_clip_min/max: 3e-4` (vs 0.2), `token_level_loss: false`, `reference_policy_kl_penalty: 0`. This file overrides none of them. The functional test then overrides two of them back on the CLI (`loss_fn.reference_policy_kl_penalty=0.01`, `loss_fn.use_importance_sampling_correction=true`), so the exemplar and the test it claims to mirror do not train the same way.

The dense, megatron-enabled, plain-GRPO grandparent [grpo_workplace_assistant_nemotron_nano_v2_9b.yaml](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml) looks like the base you want — or override the `loss_fn` block here explicitly.

The `logger` block inherits the same way: with no override here the run reports `wandb.project: grpo-workplace-assistant` / `name: nemotron-nano-v2-9b-workplace-assistant` ([base:336-346](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L336-L346)). Both same-depth siblings override it — [swe1.yaml:233-240](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/examples/nemo_gym/grpo_qwen3_30ba3b_thinking_swe1.yaml#L233-L240).

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
```diff
@@ -0,0 +1,70 @@
+# GRPO on the NeMo-Gym workplace-assistant environment via the SingleController path,
+# using non-colocated Megatron Inference.
+# Gym rollouts go through the persistent Megatron engine's OpenAI-compatible server,
+# Smoke-test scale; Qwen3-0.6B on one node with 2 GPUs (1 training + 1 generation);
+# the CI-run variant of this setup is
+# tests/functional/grpo_megatron_generation_gym_single_controller.sh.
```
yuki-97: Nothing in the repo loads this file, which is why the `data_plane` nesting above could ship broken.

The script named here loads a different config — [tests/functional/grpo_megatron_generation_gym_single_controller.sh:60](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/tests/functional/grpo_megatron_generation_gym_single_controller.sh#L60) is `--config .../grpo_qwen3_30ba3b_instruct.yaml` — and re-specifies every value on the CLI. No unit test globs `examples/nemo_gym/*.yaml` either: [test_config_v2.py:50](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/tests/unit/test_config_v2.py#L50) and [test_config_validation.py:100](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/tests/unit/test_config_validation.py#L100) both resolve `examples/configs`. Delete this file and every check stays green — no `defaults:` resolution, no `${grpo.num_prompts_per_step}` interpolation, no `MasterConfig` validation.

Pointing the script at this YAML and dropping the duplicated overrides would make the header true and give the exemplar CI coverage in one move.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -389,6 +446,54 @@ def _maybe_inject_megatron_train_iters(master_config: MasterConfig) -> None:
...
+
+    if generation_config["colocated"]["enabled"]:
+        raise NotImplementedError(
+            "SC does not support colocated Megatron generation currently."
+        )
+
+    mcore_cfg = cast(MCoreGenerationConfig, generation_config)[
+        "mcore_generation_config"
+    ]
+    if use_nemo_gym and not mcore_cfg["expose_http_server"]:
```
yuki-97: [should_use_nemo_gym](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/algorithms/single_controller_utils/setup.py#L688) runs seven lines earlier and already asserts this exact key for the megatron backend ([nemo_gym.py:99-116](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/nemo_rl/environments/nemo_gym.py#L99-L116)), so `use_nemo_gym` being true already implies `expose_http_server` is truthy — this guard can never fire and can just go.

The new `megatron_gym_without_http_server` case should go with it: it only passes because it patches `should_use_nemo_gym` out, so it is pinning a `ValueError` no run can produce.

docs/guides/single-controller.md
```diff
@@ -40,6 +40,21 @@ uv run examples/run_grpo_single_controller.py --config <your-sc.yaml>
...
+      megatron_cfg:
+        enabled: true
+      generation:
+        backend: "megatron"
+        colocated:
+          enabled: false
+          resources:
+            num_nodes: 1
+            gpus_per_node: 4
+    ```
```
yuki-97, line 58: This block is introduced as the exemplar for "a NeMo-Gym run with the OpenAI server exposed", but it omits `expose_http_server` — and the gym base defaults it to false ([grpo_qwen3_30ba3b_instruct.yaml:82](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml#L82)), so copying it verbatim fails the Gym assert at setup. `gpus_per_node` also disagrees with the linked file ([:33-42](https://github.com/NVIDIA-NeMo/RL/blob/c9c667a7b13e69d7a496013c1e9647c3bfe403d0/examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml#L33-L42) uses 1).

```suggestion
    ```yaml
    policy:
      megatron_cfg:
        enabled: true
      generation:
        backend: "megatron"
        mcore_generation_config:
          expose_http_server: true  # required for NeMo-Gym rollouts
        colocated:
          enabled: false
          resources:
            num_nodes: 1
            gpus_per_node: 1  # inference GPUs; remainder go to training
    ```
```

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -678,30 +798,69 @@ def _build_generation_then_trainer(
...
-                [ray.get(generation_router.base_url.remote())]
-                if generation_router is not None
-                else generation.dp_openai_server_base_urls
-            ),
+            base_urls=gym_spinup_base_urls,
             tokenizer=tokenizer,
         )
 
-    if colocated:
+    if megatron_backend:
```
yuki-97: After a rebase this branch skips the critic. #3773 folded value construction into `_build_trainer_and_value` ([main setup.py:716-741](https://github.com/NVIDIA-NeMo/RL/blob/d938da394063bf32381d84d819c60f8826eaa66b/nemo_rl/algorithms/single_controller_utils/setup.py#L716-L741)), and both of main's build-task shapes return it — the results region unpacks four values on the colocated leg and three on the other ([main :847-852](https://github.com/NVIDIA-NeMo/RL/blob/d938da394063bf32381d84d819c60f8826eaa66b/nemo_rl/algorithms/single_controller_utils/setup.py#L847-L852)).

This branch is chosen ahead of `elif colocated:` and calls bare `_build_trainer`, returning a 3-tuple. `value` has no `None` initializer on main and is consumed unconditionally at [:956](https://github.com/NVIDIA-NeMo/RL/blob/d938da394063bf32381d84d819c60f8826eaa66b/nemo_rl/algorithms/single_controller_utils/setup.py#L956) as `value_handle=value`, so every megatron-generation SC run — GRPO included — ends in an unpack error or `UnboundLocalError`.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -389,6 +446,54 @@ def _maybe_inject_megatron_train_iters(master_config: MasterConfig) -> None:
...
+    if not (
+        "megatron_cfg" in policy_config and policy_config["megatron_cfg"]["enabled"]
+    ):
+        raise ValueError(
+            "policy.generation.backend='megatron' requires the Megatron trainer "
+            "(policy.megatron_cfg.enabled=true): refit transfers weights via Megatron's reshard "
+            "collective from the Megatron trainer."
+        )
+
+    if generation_config["colocated"]["enabled"]:
```
yuki-97: This guard becomes unreachable after a rebase. #3773 added an unconditional colocated rejection to `validate_single_controller_config` ([main config.py:751-757](https://github.com/NVIDIA-NeMo/RL/blob/d938da394063bf32381d84d819c60f8826eaa66b/nemo_rl/algorithms/single_controller_utils/config.py#L751-L757)) — it did not exist at the merge-base — and that validator is the first statement of `setup_single_controller`, so it always fires before this function runs.

Two of the new parametrize cases go with it: `megatron_colocated` and `gym_on_sglang` both build `colocated=True` configs, so after the rebase they hit main's generic `ValueError` instead of the exception they assert.

**Suggest fix:** drop this block and the `megatron_colocated` case, and flip `gym_on_sglang` to `colocated=False`.

tests/functional/grpo_megatron_generation_gym_single_controller.sh
yuki-97: want to double check is it intended?

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -312,6 +310,65 @@ def _build_trainer(
...
+        train_cluster,
+        master_config,
+        tokenizer,
+        processor,
+        weights_path=weights_path,
+        optimizer_path=optimizer_path,
+    )
+
+    t0 = time.perf_counter()
+    generation = MegatronGeneration(
```
yuki-97: maybe a noob q, why mcore-inf can't start in parallel even in non-colocated case?
if it's for the refit, we always do refit before first step.
https://github.com/NVIDIA-NeMo/RL/blob/d938da394063bf32381d84d819c60f8826eaa66b/nemo_rl/algorithms/single_controller.py#L303

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -389,6 +446,54 @@ def _maybe_inject_megatron_train_iters(master_config: MasterConfig) -> None:
...
+        # invalidate_kv_cache (a base-class no-op for MegatronGeneration).
+        prior_mode = mcore_cfg.get("kv_cache_management_mode", "persist")
+        if prior_mode != "recompute":
+            print(
+                f"kv_cache_management_mode overridden '{prior_mode}' -> 'recompute' by "
+                f"async_rl.recompute_kv_cache_after_weight_updates=True."
+            )
+        # pyrefly: ignore[typed-dict-key-error]
+        mcore_cfg["kv_cache_management_mode"] = "recompute"
+        async_config.recompute_kv_cache_after_weight_updates = False
```
yuki-97: I'm a bit confused here.

1. why we always need to set `recompute_kv_cache_after_weight_updates=False`?
2. I'm not 100% sure but just guess you mean when `recompute_kv_cache_after_weight_updates=True`, `kv_cache_management_mode` should be `recompute`? if so, I think we should error loudly instead of silently auto set, otherwise people change the config but nothing happen.
3. if `recompute_kv_cache_after_weight_updates=False`, is there such limit need to set `kv_cache_management_mode=recompute`?

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: can we move this function and rename to something like `validate_settings` (after update there's no override, all things are just raise) to somewhere under `nemo_rl/models/generation/megatron/megatron_generation.py`?

something like below:
1. add it as a `classmethod` like `verify_served_address`
2. also add it to `GenerationInterface` and default pass since currently only mcore-inf use it.
3. add an util like `resolve_generation_class(generation_config)` at `nemo_rl/models/generation/__init__.py`.
4. get the cls and `generation_cls.validate_settings()` at `setup_single_controller`.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -628,6 +741,13 @@ def setup_single_controller(
     # live generation to front. None is also the correct value whenever the router
     # is disabled or NeMo-Gym is not in play -- it is Gym that needs one stable URL.
     generation_router = None
+    megatron_backend = generation_config["backend"] == "megatron"
+    megatron_reserved_url = None
+    megatron_port_holder = None
+    reserved_http_server_port = None
+    if megatron_backend:
+        # Normally set inside _build_generation, which megatron skips.
+        generation_config["model_name"] = master_config.policy["model_name"]
```
yuki-97, line 942: curious why mcore-inf can't use `_build_generation`? if there's some special logic, you can also add at there.

yuki-97 (commented):
Non-colocated MInf on SC — the setup wiring and the new unit coverage read well. Fourteen comments.

**Rebase first.** This no longer merges: #3773 (SC PPO) landed on main and rewrote the same `setup.py` / `single_controller.py` / `L1_Functional_Tests_SingleController.sh` / `single-controller.md` surface, and renamed `test_single_controller_setup.py` to `test_setup.py` — the file this PR adds 242 lines to. Two comments below exist only because of that rebase: the megatron build-task branch bypasses the new critic construction, and the megatron-specific colocated guard is subsumed by a generic one main now applies. Both take two of the new test cases with them.

On the red CI: both lanes hit Gym's 600s endpoint timeout, but differently — on h100 the trainer finishes in ~50s and then nothing is logged for 9.5 minutes with no generation worker ever created; on gb200 the trainer's own init takes 9m45s. Details in the first comment.

Separately: the PR description is still the untouched template. Worth a short summary of the approach, and — since this lands a new generation backend on the SC path — the usual convergence evidence (reward, loss, token_mult_prob_error), rather than only the functional test's `median(gen_kl_error) < 1.3` / `max(reward) > 0`.

tests/functional/L1_Functional_Tests_SingleController.sh
```diff
@@ -36,6 +36,7 @@ run_test() {
 
 run_test fast uv run --no-sync bash ./tests/functional/grpo_dp_single_controller.sh
 run_test fast uv run --no-sync bash ./tests/functional/grpo_async_gym_single_controller.sh
+run_test fast uv run --no-sync bash ./tests/functional/grpo_megatron_generation_gym_single_controller.sh
```
yuki-97, line 40: besides functional test, could you also add a nightly test (which is longer than the functional one) and share the results of that in the PR description to show it works well?

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: So this is an annoying conflict between pre-existing arguments.

Megatron Inference has always had three modes: `persist`, `recompute`, `offload`.     
Prior to Megatron Inference support, NeMo RL basically supported the modes `persist` and `recompute`. But it was done with different arguments, at different levels in the nested config.

When I made the big Megatron Inference PR, I intentionally kept them as separate flags to minimize conflicts, overlaps, etcetera.

I'm going to go ahead and just merge all these flags so that we can completely eliminate this code-block you're commenting on. Let's look at it together after that commit.

tdene: See 28218a6515b0b85f5545c06517487289eb788656

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: Fixed

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: Fixed

docs/guides/single-controller.md
tdene: Fixed!

tests/functional/grpo_megatron_generation_gym_single_controller.sh
tdene: Not intended at all. Fixed!

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
tdene: Done!

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
tdene: Done!

tests/functional/grpo_megatron_generation_gym_single_controller.sh
tdene: Done!

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
tdene: I don't agree with this.

The vllm SC test also uses the same base. This test being added here is meant to be an exact mirror of the vllm SC test, so it can be compared apples-to-apples.

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: Resolved

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: Fixed

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: Done

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
tdene: I've changed the CI test so that it loads this file, and dropped the duplicated overrides.

tests/functional/L1_Functional_Tests_SingleController.sh
tdene: Done, pipeline running; when that finishes I can edit the results into the PR description.

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: I believe I found the error and it will pass this time.

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: This is related to #3777 which is being resolved by #3864

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: This is related to #3777 which is being resolved by #3864

The fact that we have to build in a different order made the diff very ugly when I tried just editing `_build_generation`. But once #3864 goes in, all of that can be cleaned up.

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
yuki-97: fixed in 6e493568d58e49120ac48a06ef1c4b597400070f

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
yuki-97: fixed in e3223248a4c5ea830a69ffa16ea511cc58a98d97

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: fixed in d3602a3da33f4e304e85512dce180f6db7b2572b

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: fixed in d3602a3da33f4e304e85512dce180f6db7b2572b

examples/nemo_gym/grpo_qwen3_0_6b_megatron_generation_single_controller.yaml
yuki-97: fixed in 6e493568d58e49120ac48a06ef1c4b597400070f

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: fixed in d3602a3da33f4e304e85512dce180f6db7b2572b

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: fixed in 28218a6515b0b85f5545c06517487289eb788656

docs/guides/single-controller.md
yuki-97: fixed in 6e493568d58e49120ac48a06ef1c4b597400070f

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: fixed in f2b9a3014d811b52c12c8ad3ee94e774b54d0da2

tests/functional/grpo_megatron_generation_gym_single_controller.sh
yuki-97: fixed in e3223248a4c5ea830a69ffa16ea511cc58a98d97

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: fixed in f2b9a3014d811b52c12c8ad3ee94e774b54d0da2

tests/functional/grpo_megatron_generation_gym_single_controller.sh
yuki-97: fixed in 6e493568d58e49120ac48a06ef1c4b597400070f

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -226,6 +280,11 @@ def __init__(
         if not (skip_weight_load and gen_fp8_cfg and gen_fp8_cfg["enabled"]):
```
yuki-97: **[BUG]** SC + NeMo-Gym + megatron non-colocated + `mcore_generation_config.fp8_cfg.enabled=true` fails setup with a misleading error, and the fp8 gate itself reads the wrong config key.

SC always passes `skip_weight_load=True` for the dedicated inference model (setup.py:996) and never calls `sync_weights()` before the gym cross-check at setup.py:1118 (`MegatronGeneration.verify_served_address(generation.dp_openai_server_base_urls, megatron_reserved_url)`). When fp8 is enabled, this deferral (line 279-281) skips `prepare_for_generation()`, so `dp_openai_server_base_urls` is still `[]` — the check raises `RuntimeError("... serving []")` instead of a clear config error. (grpo.py's non-SC path is safe only because it syncs weights before this same check.)

Separately: this gate reads `self.cfg["mcore_generation_config"].get("fp8_cfg")` — the generation-side key — but the dedicated inference model is actually built from the *merged* `policy_config["megatron_cfg"]` dict (megatron_generation.py:257-260), and the worker reads fp8 from that merged dict (megatron_policy_worker.py:545). So fp8 enabled only via `policy.megatron_cfg.fp8_cfg` (not overridden under `mcore_generation_config`) is invisible to this gate — eager CUDA-graph capture happens pre-refit, the exact regression #3731 was deferring engine init to avoid.

Neither of this PR's shipped recipes sets fp8_cfg on this path, and the one functional test that would exercise it (grpo_megatron_mxfp8_refit_gb200.sh) is disabled by this PR's own hotfix commit — so this is real but currently untested.

**Suggest fix:** gate on the merged config (`self._policy_config["megatron_cfg"].get("fp8_cfg")`, already computed at line 257-260) instead of the generation-side key, and skip/guard the `verify_served_address` check when the fp8 deferral is active.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -234,10 +236,6 @@ def _build_clusters(
         return cluster, cluster, teacher_topology
 
     # Non-colocated: split node into train + inference clusters.
```
yuki-97, line 241: **[BUG]** Non-colocated megatron generation with `cluster.segment_size` set (multi-node student split) still crashes here with "only supports vllm or sglang generation", even though this PR's purpose is to add non-colocated megatron support.

This PR deletes the old `assert backend != "megatron"` guard for the non-colocated branch, but this backend whitelist a few lines down (reachable when `segment_size is not None` and `policy_nodes > 1`) was never extended to include megatron. All of this PR's own shipped recipes are single-node (`policy_nodes == 1`, `segment_size` unset), so CI never exercises this path.

This looks like an intentional scope-limit (docs only claim single-node non-colocated support) rather than a shipped regression — flagging so it's either extended here or tracked as a known follow-up rather than silently discovered later.

tests/test_suites/disabled.txt
```diff
@@ -40,3 +40,10 @@ tests/test_suites/vlm/mopd-nemotron-super-omni-120ba12b-4n8g-smoke.v1.sh
 # nightly budget is at 3922 of its 3928 GPU-hour cap and this run's ~27
 # GPU-hours would exceed it. Move to nightly.txt when the budget has room.
 tests/test_suites/llm/grpo-llama3.1-8b-instruct-2n8g-ready-first-single-controller.sh
+
+# The SC sync math pair (vLLM twin + Megatron Inference) runs temporarily on
+# GB200 as 1n4g variants (see nightly_gb200.txt) while cw_dfw H100 capacity is
+# unavailable. Restore these 1n8g entries to nightly.txt when H100 nightlies
```
yuki-97: **[TEST]** This PR disables a pre-existing, unrelated H100 nightly (the vLLM SC sync recipe's 1n8g entry, removed from nightly.txt:178) while also parking its own new 1n8g megatron-generation twin here — net effect: only the GB200 1n4g variant stays live for either recipe.

The comment above ("The SC sync math pair ... runs temporarily on GB200 as 1n4g variants ... while cw_dfw H100 capacity is unavailable") implies the vLLM entry moved to GB200, but `grpo-qwen2.5-math-1.5b-instruct-1n4g-megatron-single-controller-sync.sh` was already in nightly_gb200.txt on main before this PR — it didn't move as part of this change. Net effect is a coverage loss for the vLLM recipe's H100/1n8g perf-assertion (`mean(timing/train/total_step_time) < 25`) that this PR doesn't call out.

Worth confirming this H100 nightly removal is intentional (capacity-driven, same as the new megatron entry) rather than an incidental side effect, and correcting the comment if so.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -139,6 +140,59 @@ def reserve_http_server_address(
         node_ip, port = ray.get(holder.address.remote())
         return f"http://{node_ip}:{port}/v1", port, holder
 
+    @classmethod
+    def validate_settings(
```
yuki-97: thanks for update, this shape looks good!

I think we can just pass `master_config` as the input args instead of detail things, since different rollout backend may need to check different things.

for `master_config`, maybe need to use `TYPE_CHECKING` or just don't `master_config: MasterConfig` if there's circle input.

also `use_nemo_gym` seems not be used here.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -944,6 +965,41 @@ def _build_generation_then_trainer(
...
+        real weights over the refit collective.
+
+        Returns:
+            A tuple of (MegatronGeneration, TQPolicy trainer, TQValue critic or
+            None, per-phase wall times keyed as "gen_time", "trainer_time" and
+            "value_time").
+        """
+        time_metrics = {}
+
+        trainer, value, train_side_metrics = _build_trainer_and_value()
```
yuki-97: FYI we'll call `trainer.prepare_for_training()` at the end of `_build_trainer_and_value()`, will this affect mcore-inf?

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: https://github.com/NVIDIA-NeMo/RL/pull/3768 added the topology support into v2, want to check whether mcore-inf needs to do something with it?

tests/test_suites/disabled.txt
```diff
@@ -40,3 +40,10 @@ tests/test_suites/vlm/mopd-nemotron-super-omni-120ba12b-4n8g-smoke.v1.sh
...
+# run again.
+tests/test_suites/llm/grpo-qwen2.5-math-1.5b-instruct-1n8g-megatron-single-controller-sync.sh
```
yuki-97: from offline thread I feel that's more like a CI pipeline issue instead of the script itself?
I think we can update the comments above just saying we haven't validated it and will move out from disabled once we validated.

tdene: Yes, sorry for not clarifying @yuki-97!

I made that commit just temporarily, so that I could launch some pipelines and get a comparison. As soon as I have results, I am reverting that commit.

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: I didn't realize topology support was in v2. Let me rework this.

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: I don't see it affecting MInf in this branch, non-colocated.

For the colocated PR, we'll see. I'll handle it there if it does affect.

nemo_rl/models/generation/megatron/megatron_generation.py
tdene: Done in 61f1e2a66

nemo_rl/algorithms/single_controller_utils/setup.py
tdene: Done in 44c13e6, thanks for pointing this out!

tests/functional/L1_Functional_Tests_SingleController.sh
```diff
@@ -37,6 +37,7 @@ run_test() {
 run_test fast uv run --no-sync bash ./tests/functional/grpo_dp_single_controller.sh
 run_test fast uv run --no-sync bash ./tests/functional/ppo_async_single_controller.sh
 run_test fast uv run --no-sync bash ./tests/functional/grpo_async_gym_single_controller.sh
+run_test fast uv run --no-sync bash ./tests/functional/grpo_megatron_generation_gym_single_controller.sh
```
yuki-97: This runs Megatron Inference ungated on the GB200 shard.

[`L1_Functional_Tests_Megatron_4.sh:39`](https://github.com/NVIDIA-NeMo/RL/blob/0cc133cceaa931afda855843f5becc07f61cf9d7/tests/functional/L1_Functional_Tests_Megatron_4.sh#L39) skips every MInf test at `compute_cap >= 100` for a documented IMA, and `compute_cap` appears nowhere else under `tests/functional/`. `gb200_L1_Functional_Tests_SingleController` is a live job, so this shard would hit the IMA where Megatron_4 cleanly skips.

**Suggest fix:** hoist `megatron_generation_supported` into a shared helper and wrap this line.

yuki-97 (approved):

PR #3731, opened 2026-08-20
feat(megatron): support MXFP8 inference and BF16-to-MXFP8 refit

wdykas:
# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


tdene (dismissed):

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -197,7 +197,7 @@ def _initialize_inference_engine(self, mcore_generation_config: dict) -> None:
             ),
             logging_step_interval=logging_step_interval,
             num_speculative_tokens=num_speculative_tokens,
-            logprobs_mode="processed_logprobs",
+            logprobs_mode=mcore_generation_config["logprobs_mode"],
```
terrykong, line 294: **3 action items.**

**TL;DR — `logprobs_mode` is now a required key, and the three configs that reach this line never got it, so a megatron async-gym test and a nightly recipe die with `KeyError: 'logprobs_mode'`.**

PR-introduced. The bare subscript here is *correct* — per `config-conventions` the default belongs in the YAML, not a `.get()` fallback. Only the configs need updating.

**Worth being precise about the two `nemo_gym` files:** both are `backend: "vllm"` themselves ([`grpo_nanov3.yaml:210`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/examples/nemo_gym/grpo_nanov3.yaml#L210); `grpo_qwen3_30ba3b_instruct.yaml` inherits it from [its parent](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L227)), so neither reaches this line on its own. Each is the **base config for a megatron consumer that flips the backend**, and each defines the `mcore_generation_config` block that consumer inherits — a block that already carries every other megatron-only key (`num_speculative_tokens`, `refit_backend`, `parsers`, `expose_http_server`). That block is where the new key belongs; putting it on the two consumers instead would leave the shared block incomplete for the next one. Neither file's `defaults:` chain reaches the exemplars this PR updated.

Not suggestion blocks: none of these files is in this PR's diff.

**AI-1** — [`examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml#L79): add `logprobs_mode: processed_logprobs` after `num_speculative_tokens: 0` (L79). Consumer: [`grpo_megatron_generation_async_gym.sh`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/tests/functional/grpo_megatron_generation_async_gym.sh#L66), which overrides `policy.generation.backend=megatron` and runs unconditionally at [`L1_Functional_Tests_Megatron_4.sh:56`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/tests/functional/L1_Functional_Tests_Megatron_4.sh#L56).

**AI-2** — [`examples/nemo_gym/grpo_nanov3.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/examples/nemo_gym/grpo_nanov3.yaml#L247): same line at L247. Consumer: [`grpo-nanov3-30BA3B-2n8g-megatron_generation-async-gym.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/examples/configs/recipes/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation-async-gym.yaml#L19) (`backend: megatron`), registered in [`nightly.txt:119`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/tests/test_suites/nightly.txt#L119).

**AI-3** — [`test_megatron_worker.py:744`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/tests/unit/models/policy/test_megatron_worker.py#L744), `create_megatron_test_config` — a megatron test config, so it hits this line directly.

**Context — no action.** No config test catches this: `GenerationConfig` never declares `mcore_generation_config`, so pydantic never descends into it.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -163,8 +163,10 @@ def __init__(
             skip_weight_load=skip_weight_load,
         )
 
-        # Start the persistent inference engine + HTTP server during construction.
-        self.prepare_for_generation()
+        # Skip-load models do not have their final refit weight buffers yet. Defer
+        # engine initialization so CUDA graphs capture the persistent buffers.
+        if not skip_weight_load:
```
terrykong, line 224: **No action needed in this PR** — flagging so it is not lost, because #3569 has to land after this and will need to account for it.

**This PR leaves `dp_openai_server_base_urls` empty for non-colocated runs.** [`grpo.py:1213`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/algorithms/grpo.py#L1213) passes `skip_weight_load=not colocated_inference`, so this guard is False for every non-colocated run and [`L169`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/generation/megatron/megatron_generation.py#L169) never runs — and [`L252-261`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/generation/megatron/megatron_generation.py#L252-L261) is the only place that list is ever assigned. [`grpo.py:1287`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/algorithms/grpo.py#L1287) then reads it during setup and hands `[]` to NeMo-Gym.

#3569 ("overlap NeMo-Gym spinup with MInf load") fixes this properly: it reserves the server address up front and spins Gym up against that reserved URL, so the list is no longer on the critical path.

**One thing for whoever rebases #3569 on top of this**: it keeps a post-init consistency check —

```python
served_urls = policy_generation.dp_openai_server_base_urls
if served_urls != [reserved_url]:
    raise RuntimeError("Megatron server came up at a different address than the one pre-published to NeMo Gym: ...")
```

That check still depends on `prepare_for_generation` having run, because `base_url` is only set inside `_setup_openai_api_server`. #3569 changes which port that function uses, not when it runs. With the deferral added here, the list is `[]` for non-colocated and the check raises. Worth reconciling when the two meet.

Also [`grpo.py:1285`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/algorithms/grpo.py#L1285)'s comment ("The Megatron inference engine must be up before its server URLs exist") no longer describes the code below it.

nemo_rl/models/megatron/data.py
```diff
@@ -196,6 +256,9 @@ def get_microbatch_iterator(
     if seq_length_key is None and cfg["sequence_packing"]["enabled"]:
         seq_length_key = "input_lengths"
 
+    if not cfg["sequence_packing"]["enabled"]:
+        pad_factor = _get_non_packed_sequence_pad_factor(cfg)
```
terrykong, line 260: **1 action item.**

PR-introduced. This runs for the dense path *before* the dynamic-batching branch, and `_get_non_packed_sequence_pad_factor` bare-subscripts [`cfg["megatron_cfg"]`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/data.py#L176). [`test_get_microbatch_iterator_dynamic_batching`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/tests/unit/models/megatron/test_megatron_data.py#L1209-L1212) passes a cfg with only `dynamic_batching` and `sequence_packing`, so it raises `KeyError: 'megatron_cfg'` in the mcore unit lane. The two sibling tests in that file were updated; this one was missed.

Production is unaffected — both callers pass a full `PolicyConfig`.

**Action:** add the keys to that test's cfg. Not a suggestion block — L1209 is not in this PR's diff.
```python
        cfg = {
            "dynamic_batching": {"enabled": True},
            "sequence_packing": {"enabled": False},
            "make_sequence_length_divisible_by": 1,
            "megatron_cfg": {
                "tensor_model_parallel_size": 1,
                "sequence_parallel": False,
                "context_parallel_size": 1,
                "fp8_cfg": {"enabled": False},
            },
        }
```

tests/functional/grpo_megatron_mxfp8_refit_gb200.sh
```diff
@@ -0,0 +1,111 @@
...
+    policy.train_global_batch_size=8 \
+    policy.train_micro_batch_size=1 \
+    policy.logprob_batch_size=1 \
+    policy.max_total_sequence_length=512 \
+    ++policy.megatron_cfg.train_iters=2 \
+    policy.megatron_cfg.tensor_model_parallel_size=1 \
+    policy.megatron_cfg.expert_model_parallel_size=2 \
+    policy.megatron_cfg.sequence_parallel=false \
+    policy.megatron_cfg.activation_checkpointing=true \
+    policy.megatron_cfg.fp8_cfg.enabled=false \
```
terrykong, line 60: **1 action item, 2 follow-ups.**

**TL;DR — with training-side FP8 off the dense pad factor resolves to 1, so none of this PR's new padding code runs in this test.**

PR-introduced. The recipe inherits `make_sequence_length_divisible_by` = 1 ([`grpo_math_1B.yaml:341`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/examples/configs/grpo_math_1B.yaml#L341)); the script sets TP=1 and `sequence_parallel=false`; CP is 1; and this line disables training FP8. So the factor is `lcm(1, 1) = 1` and [`_pad_sequence_aligned_tensors`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/data.py#L196) returns at its `multiple <= 1` guard. All the MXFP8 here is on the generation engine.

This override is also a no-op — the recipe already inherits `fp8_cfg.enabled: false` ([`L309`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/examples/configs/grpo_math_1B.yaml#L309)).

**AI-1:** enable training-side FP8 (or set `policy.make_sequence_length_divisible_by` above 1) in one invocation, so the padding and the slice-back execute once end to end.

**Follow-up 1:** nothing asserts the un-padding — `grep -rn original_seq_length tests/` returns only the padding-side assertion at [`test_megatron_data.py:213`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/tests/unit/models/megatron/test_megatron_data.py#L213).

**Follow-up 2:** only the `lcm(1, 32)` leg of the new helper is covered; TP+SP is untested.

nemo_rl/models/megatron/train.py
```diff
@@ -612,6 +615,7 @@ def __call__(
         data_dict: BatchedDataDict[Any],
         input_ids: torch.Tensor,
         cu_seqlens_padded: torch.Tensor,
+        original_seq_length: Optional[int] = None,
```
terrykong: **1 action item, 1 follow-up.**

PR-introduced, low severity. The `None` default is not merely unused — it resolves to the *wrong* value. `data_dict` is the object [`_pad_sequence_aligned_tensors`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/data.py#L344) mutated in place (the iterator yields that same object, [`data.py:128-145`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/data.py#L128-L145)), so the fallback reads the **padded** width and means "don't strip the padding" — exactly what this parameter exists to prevent. Unreachable today because `forward_with_post_processing_fn` always passes it, which is the argument for not leaving a wrong answer in the default.

```suggestion
        original_seq_length: int,
```
Then delete the `if original_seq_length is None:` block at [L634-635](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/train.py#L634-L635), and mirror both edits in `TopkLogitsPostProcessor` ([param L703](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/train.py#L703), [fallback L722-723](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/train.py#L722-L723) — keep `unpacked_seqlen`, the packed branch still uses it).

**Follow-up:** neither `Args:` block documents the new parameter ([L625-628](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/train.py#L625-L628), [L710-712](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/train.py#L710-L712)).

nemo_rl/models/generation/megatron/megatron_generation.py
terrykong, line 224: **No action needed in this PR** — not caused by it, and not yours to fix. Tracked in #3777, noted here for context since it sits in the same init path.

Megatron is the only generation backend that initializes **policy then generation** in sequence ([`grpo.py:1275-1280`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/algorithms/grpo.py#L1275-L1280)). vLLM, SGLang, TRT-LLM and Dynamo all go through [`initialize_generation_with_policy`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/algorithms/grpo.py#L1217), which runs the two concurrently when inference is not colocated.

#3569 removes the NeMo-Gym part of the serialization but keeps this inner one, so after it lands:

Whether megatron *can* parallelize is the open question in #3777: the comment at [`grpo.py:1276`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/algorithms/grpo.py#L1276) says inference waits so "checkpoint conversion completes", which holds colocated — but non-colocated passes `policy=None` and `skip_weight_load=True`, and `weights_path` comes from the checkpointer ([L1121](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/algorithms/grpo.py#L1121)) rather than from `init_policy()`.

terrykong (commented):
Reviewed by a team of six agents (RL/config, megatron-core upstream, bug scan, design, tests, plus an adversarial pass). Six findings below.

**The core change is right, and worth stating explicitly:** deferring engine init under `skip_weight_load` fixes a real ordering bug, not a cosmetic one. Upstream states that `prepare_swap_model_weights` "quantizes the target decoder weights to persistent FlashInfer MXFP8Tensor buffers (whose addresses are later baked into CUDA graphs)" ([refit.py:307-311](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/resharding/refit.py#L307-L311)), and pre-PR the engine was built in `__init__`, before refit ever created those buffers. The docstring updates on `preinit_nvshmem_collective` are correspondingly correct.

The dense-padding change adds something genuinely missing too: the non-packed path had **no FP8 token alignment** before this PR, so dense + MXFP8 silently fell back to bf16 all-gathers (TE warns `"Cannot quantize input tensor. Performing all-gather in high precision."`) and slow-path quantize kernels. To be precise about scope, the `make_sequence_length_divisible_by` leg of the new helper is *not* new — that rounding already happened upstream in [`batched_message_log_to_flat_message`](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/data/llm_message_utils.py#L355-L358) — so it acts as a defensive backstop. The FP8 alignment is the new capability.

Two of the three critical items land on the same active L1 test, `tests/functional/grpo_megatron_generation_async_gym.sh`.

Several candidate findings were investigated and **dropped** rather than posted, and one is worth recording so it doesn't resurface: the claim that `_get_non_packed_sequence_pad_factor` should multiply its constraints instead of using `lcm`. That is wrong. Under sequence parallelism Megatron all-gathers to the full sequence before the GEMM ([layers.py:597-606](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/tensor_parallel/layers.py#L597-L606)), and upstream's own FP8 padder pads *after* gathering ([fp8_utils.py:946-953](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/fp8_utils.py#L946-L953)) — so the FP8 and TP-scatter constraints apply to the same quantity and compose by `lcm`. CP is separately impossible on the dense path ([setup.py:861-863](https://github.com/NVIDIA-NeMo/RL/blob/dfdf6894d5380f13f8b9c285315aa0f2ee99dafb/nemo_rl/models/megatron/setup.py#L861-L863) asserts sequence packing). The packed sibling multiplies because CP ranks genuinely never re-gather; that asymmetry is correct, not a smell. **`lcm` is the exact minimum here and the helper is right as written.**

Note on verification: this review is static analysis only — no GPU was available and `uv sync` is Linux-only on the reviewing host, so no tests or linters were executed. Nothing below is claimed as reproduced at runtime.

_Generated by Claude Code_

nemo_rl/models/megatron/data.py
wdykas: addressed

nemo_rl/models/megatron/train.py
wdykas: addressed

terrykong (dismissed):

wdykas:
/ok to test [f316afa](https://github.com/NVIDIA-NeMo/RL/pull/3731/commits/f316afa39e8f8b80cfac0a6a728b441ff475f842)

wdykas:
/ok to test [f316afa](https://github.com/NVIDIA-NeMo/RL/pull/3731/commits/f316afa39e8f8b80cfac0a6a728b441ff475f842)

wdykas:
/ok to test [917ca58](https://github.com/NVIDIA-NeMo/RL/pull/3731/commits/917ca585f506b807343a0a9811de50eaa2c07e80)

wdykas:
/ok to test [da7bee9](https://github.com/NVIDIA-NeMo/RL/pull/3731/commits/da7bee9dee8b2c78943f6cd506a012cd0eee4b52)

nemo_rl/models/generation/megatron/megatron_generation.py
terrykong, line 224: **1 action item — and the edit is in `grpo.py`, not in this PR's code.**

**TL;DR — move the six-line address check from setup to just after the first `sync_weights()`. Where it sits now it fires on a healthy state and stops every non-colocated NeMo-Gym run at setup.**

*This replaces my earlier note on this line.* I said #3569 would land after this PR. It merged first, so that note is out of date — this is the current version.

**What this PR changes.** On `main`, the engine and its HTTP server start while the workers are being built ([main megatron_generation.py:223](https://github.com/NVIDIA-NeMo/RL/blob/eaf02ef6baab63eee51886c7cbd85a83fe23d7d2/nemo_rl/models/generation/megatron/megatron_generation.py#L223)). This PR waits instead, until weights arrive by refit. That is the right call — CUDA graphs then capture the real MXFP8 buffers.

**What it runs into.** #3569 checks, during setup, that the engine came up at the address it reserved ([main grpo.py:1394-1398](https://github.com/NVIDIA-NeMo/RL/blob/eaf02ef6baab63eee51886c7cbd85a83fe23d7d2/nemo_rl/algorithms/grpo.py#L1394-L1398)):

```python
served_urls = policy_generation.dp_openai_server_base_urls
if served_urls != [reserved_url]:
    raise RuntimeError(...)
```

That list is only filled when the engine starts ([L252-261](https://github.com/NVIDIA-NeMo/RL/blob/da7bee9dee8b2c78943f6cd506a012cd0eee4b52/nemo_rl/models/generation/megatron/megatron_generation.py#L252-L261)). With the wait added here, the engine starts about 270 lines later, at [`sync_weights()` — main grpo.py:1669](https://github.com/NVIDIA-NeMo/RL/blob/eaf02ef6baab63eee51886c7cbd85a83fe23d7d2/nemo_rl/algorithms/grpo.py#L1669). So the check runs first and sees `[]`.

**This is a false alarm, not a catch.** At that moment `[]` means *not yet*, not *wrong*. The engine does come up at the reserved address — rank 0 adopts the very socket that was reserved and serves on it at the first refit. NeMo-Gym does not need the address live before then either: `NemoGym._spinup` is deferred on purpose so it can start "once the ... server URLs are available, overlapping with ... model loading", and `health_check()` polls Gym's own processes rather than the policy URL. So nothing is actually broken — the check just turns a healthy state into a hard failure.

**Action:** move those six lines from [main grpo.py:1394](https://github.com/NVIDIA-NeMo/RL/blob/eaf02ef6baab63eee51886c7cbd85a83fe23d7d2/nemo_rl/algorithms/grpo.py#L1394-L1398) to just after [`sync_weights()`](https://github.com/NVIDIA-NeMo/RL/blob/eaf02ef6baab63eee51886c7cbd85a83fe23d7d2/nemo_rl/algorithms/grpo.py#L1669). Nothing else needs to change — NeMo-Gym already receives the reserved address, and the port stays held the whole time. After the move the check still does its real job: catching an engine that came up somewhere other than the reserved address. Not a suggestion block, because those lines are in `grpo.py` and this PR does not touch that file.

Walkthrough with a diagram: https://terrykong.github.io/gh-pages-poc/terryk/pr-3731-gym-url.html

tests/functional/grpo_megatron_mxfp8_refit_gb200.sh
```diff
@@ -0,0 +1,115 @@
...
+    grpo.num_prompts_per_step=2 \
+    grpo.num_generations_per_prompt=4 \
+    grpo.max_num_steps=2 \
+    env.math.num_workers=2 \
+    loss_fn.reference_policy_kl_penalty=0.0 \
+    policy.train_global_batch_size=8 \
+    policy.train_micro_batch_size=1 \
+    policy.logprob_batch_size=1 \
+    policy.max_total_sequence_length=512 \
+    policy.make_sequence_length_divisible_by=32 \
```
terrykong: **1 action item** — reopening the coverage point from the previous review. The thread was resolved, but the path is still inert; I don't think this landed the way it was intended.

`make_sequence_length_divisible_by` is **also** applied upstream when the training batch is built ([`grpo.py:3155-3161`](https://github.com/NVIDIA-NeMo/RL/blob/da7bee9dee8b2c78943f6cd506a012cd0eee4b52/nemo_rl/algorithms/grpo.py#L3155-L3161) → [`llm_message_utils.py:355-358`](https://github.com/NVIDIA-NeMo/RL/blob/da7bee9dee8b2c78943f6cd506a012cd0eee4b52/nemo_rl/data/llm_message_utils.py#L355-L358)). [`grpo.py:2240`](https://github.com/NVIDIA-NeMo/RL/blob/da7bee9dee8b2c78943f6cd506a012cd0eee4b52/nemo_rl/algorithms/grpo.py#L2240) states the invariant: *"train_data is padded to roundup(max(S_i), make_sequence_length_divisible_by)"*.

So setting it to 32 here means the batch already arrives 32-aligned:

1. pad factor = `lcm(32, 1) = 32` (training FP8 is still off on [L60](https://github.com/NVIDIA-NeMo/RL/blob/da7bee9dee8b2c78943f6cd506a012cd0eee4b52/tests/functional/grpo_megatron_mxfp8_refit_gb200.sh#L60), TP=1, SP off)
2. microbatch width == batch width — `sequence_packing: false` and `dynamic_batching: False`, so slicing is along the batch dim only
3. `_round_up_to_multiple(seq_len, 32) == seq_len` → `padding == 0` → [`_pad_sequence_aligned_tensors`](https://github.com/NVIDIA-NeMo/RL/blob/da7bee9dee8b2c78943f6cd506a012cd0eee4b52/nemo_rl/models/megatron/data.py#L196) returns at its **second** guard

The first guard (`multiple <= 1`) is now cleared, but the padding and the slice-back still never execute. My earlier comment offered this option without noting that the same knob is applied upstream — that one is on me.

**Action:** drop `policy.make_sequence_length_divisible_by=32` (letting it fall back to the inherited `1`) **and** enable training-side FP8 — `policy.megatron_cfg.fp8_cfg.enabled=true` with `++policy.megatron_cfg.fp8_cfg.fp8_recipe=mxfp8`. Then the factor is `lcm(1, 32) = 32` against an unrounded batch and real padding occurs.

Note both edits are needed: keeping `=32` while enabling FP8 gives `lcm(32, 32) = 32` against a 32-aligned batch, which is still `padding == 0`. Not a suggestion block because the two lines are not adjacent (L54 and L60).

wdykas:
@terrykong CI has passed whenever you have time

terrykong (commented):
Follow-up on the new head (`da7bee9`). The previous round is in good shape — `logprobs_mode` across all three configs, the `megatron_cfg` KeyError, and the `original_seq_length` changes are all correctly done, and the new un-padding and TP+SP tests assert real values rather than just shapes. L0 is green.

One thing has changed since then, and it is not caused by anything in this PR: **#3569 merged first.** This branch is 2 commits behind `main` and does not contain it yet. After the rebase, the address check that PR added runs before this PR's deferred engine start, sees an empty list, and stops every non-colocated NeMo-Gym run at setup. Nothing is actually wrong when it fires — the engine does come up at the reserved address, just later. The fix is to move six lines in [`grpo.py`](https://github.com/NVIDIA-NeMo/RL/blob/eaf02ef6baab63eee51886c7cbd85a83fe23d7d2/nemo_rl/algorithms/grpo.py#L1394-L1398); details in the inline comment.

Do not read the current green CI as covering this. `fast_L1_Functional_Tests_Megatron_4` runs the async-gym test and passes today, but only because #3569's check is not in this branch yet. A rebase plus a re-run of that one lane settles it either way.

Diagram and walkthrough: https://terrykong.github.io/gh-pages-poc/terryk/pr-3731-gym-url.html

Also noted, no action needed here: two further test configs (`test_megatron_generation.py`, `test_vllm_generation.py`) needed `make_sequence_length_divisible_by` and were fixed in follow-up commits. My earlier comment named only one of the three sites — I swept production configs but not test configs, so that gap was found by L0 rather than by review.

_Generated by Claude Code_

wdykas:
/ok to test [632dfb0](https://github.com/NVIDIA-NeMo/RL/pull/3731/commits/632dfb00e0a352d6800700e21ce07ce0f1ff4171)

terrykong (approved):

PR #3732, opened 2026-08-20
feat(ppo): support NeMo Gym rollouts

bg51717:
# What does this PR do ?

Adds NeMo Gym rollout support to synchronous and asynchronous PPO. PPO can now use the same Gym environment, rollout, validation, reward-processing, and task-cursor infrastructure as GRPO.

## Design

The main changes are:

- Add a dedicated PPO NeMo Gym launcher and example recipe.
- Route PPO training and validation through NeMo Gym for both synchronous and asynchronous PPO.
- Save and restore the Gym task cursor with async replay-buffer state.
- Support Gym reward penalties, effort shaping, sample masking, response logging, and message-level advantage overrides.
- Share Gym helpers with GRPO, clean up Gym/generation resources on exit, and add documentation and tests.

# Issues

N/A

# Usage

Use the included recipe with the dedicated launcher:

```bash
uv run examples/nemo_gym/run_ppo_nemo_gym.py \
    --config examples/nemo_gym/ppo_math_rlvr_nemo_gym.yaml \
    data.train.data_path=/path/to/train.jsonl \
    data.validation.data_path=/path/to/validation.jsonl
```

Set `ppo.async_ppo.enabled=true` together with the existing non-colocated Async PPO requirements to use asynchronous Gym rollouts. See `examples/nemo_gym/ppo_math_rlvr_nemo_gym.yaml` for the complete configuration.

**Pre checks**:

# Test

- Full relevant unit suite: **1117 passed, 6 skipped**, covering PPO, shared async
utilities, GRPO compatibility, configuration, checkpoint resolution, and value-worker
checkpointing.
- Synchronous PPO + NeMo Gym functional: completed 10 training steps and passed logprob-
mismatch and validation-accuracy gates.
- Async PPO + NeMo Gym functional: completed 10 training steps and passed logprob-mismatch,
trajectory-age, validation-accuracy, and Gym cursor checkpoint gates.
- Async GRPO + NeMo Gym functional: passed as a regression check for the shared collector
and Gym helper paths.


nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -934,6 +935,7 @@ async def _iter_rollout_groups(
                 max_rollout_turns=None,
                 greedy=False,
                 reward_penalty_config=self.master_config.reward_penalties,
+                effort_config=get_nemo_gym_effort_config(self.master_config.env),
```
bg51717, line 940: **1 action item.**

**TL;DR — this line activates effort shaping for *async GRPO*, not just PPO, changing reward numerics on 8 shipped recipes that no CI job covers.**

[`trajectory_collector.py:938`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/async_utils/trajectory_collector.py#L938) — PR-introduced.

This is the shared collector, so the new kwarg reaches async GRPO too. Pre-PR it was absent, defaulting to `None` ([`rollouts.py:2324`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/experience/rollouts.py#L2324)), so [`_apply_effort_shaping` early-returned](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/experience/rollouts.py#L472). Now the guard passes and it [rewrites the reward in place](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/experience/rollouts.py#L506).

To be clear: **this is a fix, not a regression.** Sync GRPO has always passed `effort_config`; async silently dropped it, so these recipes have been ignoring their own config. Please don't revert it — and it can't be scoped to PPO without institutionalizing the bug.

Affected (all `async_grpo.enabled: true`, `low_weight: 0.1`, `low_ub: 15000`): the 7 `examples/nemo_gym/nemotron-3-ultra/*.yaml` plus `nemotron-3.5-lightning/rlvr.yaml`. `nemotron-3-super/stage1_rlvr.yaml` has `low_weight: 0.0`, so it stays a no-op.

Magnitude — both branches are live:

```
len=     0  ->  reward x1.100      (bonus, bounded at +10%)
len= 14999  ->  reward x1.000
len= 49152  ->  reward -0.228      (penalty, additive)
len=196608  ->  reward -1.211      (mopd, swe_teacher)
```

At the 196608 cap the additive penalty can invert the sign of a [0,1] task reward.

The plumbing *is* tested — [`test_async_utils.py:2167`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/tests/unit/algorithms/test_async_utils.py#L2167) asserts the forward. What's untested is the numeric consequence: every async test mocks the rollout, and `grpo_async_gym.sh` sets no `effort_levels`, so nothing in CI would surface the shift.

**Action:** call this out in the PR description and in the design doc's "Async Collector Invariants" section, and consider adding `effort_levels` to `tests/functional/grpo_async_gym.sh` to close the end-to-end gap. The ask holds either way — if your RLVR prompts carry the `low_string` marker, 8 recipes change numerically; if they don't, those 8 carry dead config worth deleting.

nemo_rl/algorithms/ppo.py
```diff
@@ -405,6 +474,26 @@ def setup(
...
+        or ppo_config.malformed_thinking_advantage is not None
+    ):
+        raise ValueError(
+            "ppo.invalid_tool_call_advantage and "
+            "ppo.malformed_thinking_advantage require the NeMo-Gym path "
+            "(env.should_use_nemo_gym=true)"
+        )
+    if enable_nemo_gym and backend != "vllm":
+        raise NotImplementedError(
+            "PPO NeMo Gym currently requires the vLLM generation backend"
```
bg51717, line 494: **1 action item.**

[`ppo.py:494`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L494) — PR-introduced.

The new Gym validation block guards `reward_penalties`, the two advantage overrides, and the backend — but not `ppo.reward_scaling` / `ppo.reward_shaping`. Async PPO [rejects both outright](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L347), so sync is the odd one out.

This bites by *inheriting a default*, not by opting in. `examples/configs/ppo_math_1B.yaml` ships [`reward_shaping.enabled: true`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/examples/configs/ppo_math_1B.yaml#L37) and [`reward_scaling.enabled: true`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/examples/configs/ppo_math_1B.yaml#L56), where GRPO's exemplar ships both `false`. A recipe deriving from it with `env.should_use_nemo_gym: true` — rather than copying `ppo_math_rlvr_nemo_gym.yaml`, which does disable both — silently gets:

1. [`scale_rewards`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L1622) remaps the Gym reward `[0,1]` to `[-1,1]`, so a reward zeroed by `reward_penalties` becomes a full **-1.0**, not "no reward".
2. [`apply_reward_shaping`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/reward_functions.py#L145) finds no `response_token_lengths` on a Gym batch, so it scans `message_log` and **breaks on the first assistant message**.

Measured on a 3-turn rollout (100/100/13000 assistant tokens): rewards `[1.0, 0.0]` become `[1.0, -1.0]`, and the overlong penalty evaluates length 100 instead of 13,200 — so it never fires. No error at any point.

**Action:** extend the guard. This blocks nothing legitimate — async already rejects both, and every in-tree Gym recipe sets them false.

```suggestion
    if enable_nemo_gym and backend != "vllm":
        raise NotImplementedError(
            "PPO NeMo Gym currently requires the vLLM generation backend"
        )
    if enable_nemo_gym and ppo_config.reward_scaling.enabled:
        raise NotImplementedError(
            "ppo.reward_scaling is not supported with NeMo Gym rollouts; "
            "NeMo Gym owns reward computation"
        )
    if enable_nemo_gym and ppo_config.reward_shaping.enabled:
        raise NotImplementedError(
            "ppo.reward_shaping is not supported with NeMo Gym rollouts; "
            "its response-length term only sees the first assistant turn"
        )
```

nemo_rl/algorithms/ppo.py
```diff
@@ -3054,29 +3302,71 @@ def validate(
...
+            if should_use_nemo_gym(master_config):
+                generation_config = {
+                    **master_config.policy["generation"],
+                    "stop_token_ids": None,
+                    "stop_strings": None,
+                }
+                val_sampling_params = GenerationSamplingParams(
+                    temperature=generation_config["val_temperature"],
+                    top_p=generation_config["val_top_p"],
+                    top_k=generation_config["val_top_k"],
```
bg51717, line 3349: **1 action item.**

[`ppo.py:3329`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L3329) — PR-introduced.

`val_top_k` flows into `GenerationSamplingParams`, and the Gym path [hard-asserts `not sampling_params.top_k`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/experience/rollouts.py#L2425). `val_top_k` appears nowhere else in `ppo.py`, so there's no setup-time guard — and with `val_at_start: true` this dies after the full cluster and Gym stack are already up. Shipped defaults are safe (`top_k: null`), so it only bites a user who sets it.

Two traps worth knowing before fixing:

- **Don't copy GRPO's guard.** [`grpo.py:471-489`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/grpo.py#L471) nests `assert not val_top_k` inside a `val_sampling_overridden` *inequality* test. With the shipped `val_top_k: ${.top_k}` interpolation the two are equal, so the guard never fires — GRPO carries this hole today.
- **Checking `val_top_k` alone isn't enough.** When `sampling_params is None` the rollout [builds them from the generation config](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/experience/rollouts.py#L2421), and `from_generation_config` reads the **train** `top_k` (its docstring: "Build the train-time sampling profile"). Both sync `ppo_train` and the async collector call without `sampling_params`, so `top_k: 20` with an explicit `val_top_k: null` still crashes — at the first *training* rollout.

**Action:** add an unconditional two-key assert next to the other Gym guards at `ppo.py:492-495`. Scoping this honestly: it converts a late crash into a setup-time one with an actionable message; it does not prevent a silent wrong result.

```python
    if enable_nemo_gym:
        assert not generation_config["top_k"] and not generation_config["val_top_k"], (
            "policy.generation.top_k/val_top_k are not supported on the NeMo-Gym "
            "rollout path, which only honors temperature/top_p."
        )
```

examples/run_ppo.py
```diff
@@ -159,7 +122,7 @@ def main() -> None:
     async_config = config.ppo.async_ppo
     async_ppo_enabled = async_config.enabled
     if async_ppo_enabled:
-        _validate_async_ppo_config(config, policy_generation)
+        validate_async_ppo_config(config, policy_generation)
```
bg51717: **1 action item.** Best read together with the note on `ppo.py:494` — the two interact.

[`run_ppo.py:125`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/examples/run_ppo.py#L125) — PR-introduced.

`validate_async_ppo_config` dropped its `"NeMo Gym rollout"` entry from `unsupported_features`. That's correct for the new launcher, but `examples/run_ppo.py` calls the *same* validator and can't run Gym — so a configuration that previously failed fast now proceeds.

`nemo_gym` **is** registered in [`ENV_REGISTRY`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/environments/utils.py#L53), so `setup_response_data` builds `task_to_env` from a bare `NemoGym` actor that never went through [`_spinup` + `set_tokenizer`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/environments/nemo_gym.py#L1206) — while the correctly-initialized actor from `setup()` is discarded into `_nemo_gym`. Two actors; the broken one wins. Pre-PR this raised `NotImplementedError: NeMo Gym rollout is not supported with async PPO` at startup.

Explicit on scope: the `_nemo_gym` discard itself is **not** this PR's problem — `run_grpo.py:153` and `run_distillation.py:93` do the same. Only the lost guard is PR-introduced.

**Action:** fail fast in `main()`, right after `MasterConfig(**config)`.

```python
    if config.env.get("should_use_nemo_gym"):
        raise ValueError(
            "env.should_use_nemo_gym=true is not supported by examples/run_ppo.py; "
            "use examples/nemo_gym/run_ppo_nemo_gym.py"
        )
```

**Interaction worth deciding deliberately:** the new `enable_nemo_gym and backend != "vllm"` guard at `ppo.py:494` is reachable *only* through this entrypoint — via `run_ppo_nemo_gym.py`, `setup_nemo_gym_config` sets `vllm_cfg.async_engine` first and would `KeyError` on a non-vLLM config. If you add the guard above, that one becomes dead code. Worth resolving as one change rather than two that cancel out.

nemo_rl/algorithms/ppo.py
```diff
@@ -458,6 +547,9 @@ def setup(
         assert val_dataset is not None, (
             "Validation dataset is required if validation is enabled"
         )
+        assert ppo_config.val_batch_size is not None, (
```
bg51717, line 560: **1 action item.**

[`ppo.py:550`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L550) — PR-introduced.

This PR widens `max_val_samples` to `int | None` alongside `val_batch_size`, but `setup()` only asserts `val_batch_size`. [`validate()` asserts both](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L3306).

So `ppo.max_val_samples: null` with a normal `val_batch_size` passes setup, allocates the full cluster, then dies at the first validation on a bare `AssertionError`. Reachable because `examples/nemo_gym/ppo_math_rlvr_nemo_gym.yaml` teaches `max_val_samples: null` as the pattern, and only the Gym launcher writes the value back.

**Action:** add the sibling assert.

```suggestion
        assert ppo_config.val_batch_size is not None, (
            "ppo.val_batch_size must be set when validation is enabled"
        )
        assert ppo_config.max_val_samples is not None, (
            "ppo.max_val_samples must be set when validation is enabled"
        )
```

tests/unit/algorithms/test_ppo.py
```diff
@@ -2470,10 +2679,13 @@ def test_async_ppo_completed_resume_exits_before_actor_start(monkeypatch):
     value_model.shutdown.assert_called_once()
 
 
-def test_async_ppo_initial_refit_failure_cleans_up_actors(monkeypatch):
+def test_async_ppo_restores_gym_cursor_and_cleans_up_on_initial_refit_failure(
```
bg51717, line 2794: **1 action item.**

[`test_ppo.py:2682`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/tests/unit/algorithms/test_ppo.py#L2682) — PR-introduced.

This covers only the case where `rollouts.pt` is *ahead of* the replay-buffer metadata (7 vs 5), where [`max()`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L2475) returns the same value from either source. The replay-buffer branch is never exercised.

Proven by mutation: replacing `ppo.py:2475-2478` with `int(rollouts_state.get(NEXT_NEMO_GYM_TASK_INDEX_KEY, 0))` — deleting the replay-buffer source entirely — leaves the suite at **97 passed, 7 skipped**.

What slips through: resuming from a checkpoint written before `rollouts.pt` existed silently restarts `_ng_task_index` at 0 and reuses Gym task identities — which the design doc this PR edits explicitly promises against ("checkpointed, restored, and cross-checked against buffered trajectories so task identities are not reused after restart").

**Action:** parametrize over both sources. Verified passing 3/3, and the `(None, 5, 5)` case fails under the mutation.

```python
@pytest.mark.parametrize(
    ("rollouts_cursor", "replay_cursor", "expected_cursor"),
    [(7, 5, 7), (None, 5, 5), (3, 9, 9)],
)
def test_async_ppo_restores_gym_cursor_and_cleans_up_on_initial_refit_failure(
    monkeypatch, tmp_path, rollouts_cursor, replay_cursor, expected_cursor
):
    ...
    replay_actor.load_from_path.remote.return_value = {
        NEXT_NEMO_GYM_TASK_INDEX_KEY: replay_cursor
    }
    torch.save({}, checkpoint_path / "replay_buffer.pt")
    if rollouts_cursor is not None:
        torch.save(
            {NEXT_NEMO_GYM_TASK_INDEX_KEY: rollouts_cursor},
            checkpoint_path / "rollouts.pt",
        )
    ...
    assert (
        collector_type.options.return_value.remote.call_args.kwargs[
            "next_nemo_gym_task_index"
        ]
        == expected_cursor
    )
```

tests/unit/algorithms/test_ppo.py
```diff
@@ -716,6 +717,8 @@ def _make_ppo_loop_batch(
                         "role": "assistant",
                         "content": "answer-0",
                         "token_ids": torch.tensor([2, 3]),
+                        "generation_logprobs": torch.zeros(2),
```
bg51717, line 721: **1 action item.**

[`test_ppo.py:720`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/tests/unit/algorithms/test_ppo.py#L720) — PR-introduced.

`_make_ppo_loop_batch` never sets `mask_sample`, so the new [`apply_mask_sample_filter`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L1647) calls in sync and async PPO are only ever exercised on the no-op path — the helper returns 0 in every PPO test. `test_utils.py:38` covers the helper in isolation, but nothing checks the PPO wiring.

That wiring is **placement-sensitive**, which is what makes it worth a test: the call mutates `repeated_batch["loss_multiplier"]`, which is read exactly once, 34 lines later, at [`train_data["sample_mask"]`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L1681).

Proven by mutation: moving the sync call to *after* `train_data` is built leaves **97 passed, 7 skipped**. The resulting bug is silent — Gym-flagged samples keep contributing to the loss while `metrics["num_mask_sample_filtered"]` still reports them masked. Both new functional scripts run `++env.should_mask_flagged_samples=true`, so this is the shipped configuration.

**Action:** thread a `mask_sample` knob through the harness and assert it reaches `sample_mask`. Verified passing sync and async; fails under the mutation.

```python
@pytest.mark.parametrize("async_mode", [False, True])
def test_ppo_train_masks_gym_flagged_samples_out_of_the_loss(monkeypatch, async_mode):
    """mask_sample must land in sample_mask, i.e. before train_data is built."""
    harness = _run_mock_ppo_train(
        monkeypatch, async_mode=async_mode, max_num_steps=1, ppo_epochs=1,
        seq_logprob_error_threshold=None, use_nemo_gym=True,
        mask_sample=(False, True),
    )
    train_data = harness.policy.train.call_args.args[0]
    torch.testing.assert_close(train_data["sample_mask"], torch.tensor([1.0, 0.0]))
```

Needs a 3-line `mask_sample` parameter on `_make_ppo_loop_batch` / `_run_mock_ppo_train`, threaded into the `DummyLoader` batches and `replay_actor.sample.remote`.

tests/unit/algorithms/test_ppo.py
```diff
@@ -1950,13 +2089,79 @@ def test_noncolocated_vllm_builds_separate_clusters_and_collective(monkeypatch):
...
-    value_model = result[2]
+    value_model = result[3]
     value_model.finish_training.assert_called_once_with()
     policy.prepare_for_training.assert_called_once_with()
     policy.prepare_refit_info.assert_called_once_with()
     generation.prepare_refit_info.assert_called_once_with({"state": "dict"})
 
 
+def test_vllm_setup_spins_up_nemo_gym_with_deferred_model_load(monkeypatch):
+    """PPO exposes vLLM URLs before loading so Gym startup can overlap."""
```
bg51717: **1 action item.**

[`test_ppo.py:2100`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/tests/unit/algorithms/test_ppo.py#L2100) — PR-introduced.

This exercises only the non-colocated Gym branch. The colocated path — [`init_vllm_then_policy_value` / `results["vllm_policy_value"]`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L1002) — has no unit coverage, and it's exactly what `tests/functional/ppo_nemo_gym.sh` runs (`policy.generation.colocated.enabled=true`). The only colocated setup test doesn't set `should_use_nemo_gym`, so it takes the `else` branch.

What slips through: any mismatch between the `init_tasks` key and the result unpacking (renaming `"vllm_policy_value"` on one side only) raises `KeyError` at setup for every colocated Gym run — today that surfaces only in a 2-GPU L1 job. Confirmed: renaming just the assignment leaves the non-colocated param green and fails the colocated one.

**Action:** parametrize over colocation — verified passing both params.

```python
@pytest.mark.parametrize("colocated", [False, True])
def test_vllm_setup_spins_up_nemo_gym_with_deferred_model_load(monkeypatch, colocated):
    config = _make_noncolocated_setup_config()
    config.env.update(should_use_nemo_gym=True, nemo_gym={})
    if colocated:
        config.policy["generation"]["colocated"] = {
            "enabled": True,
            "resources": {"num_nodes": None, "gpus_per_node": None},
        }
    ...  # rest unchanged
```

**Context — no action.** This covers branch *execution*, not the serialize-before-parallel ordering that prevents GPU OOM. Collapsing the colocated branch into the parallel one also leaves every test green, and that invariant resists a deterministic assertion under `ThreadPoolExecutor` — a source comment stating it is probably the better tool.

tests/unit/algorithms/test_async_utils.py
```diff
@@ -1456,6 +1461,7 @@ def test_collector_selects_ppo_config(self):
         assert collector.algorithm_config is master_config.ppo
         assert collector.async_config is async_config
         assert collector.async_config.max_trajectory_age_steps == 3
+        assert collector.master_config.reward_penalties is reward_penalties
```
bg51717: **1 action item, 1 follow-up.**

[`test_async_utils.py:1464`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/tests/unit/algorithms/test_async_utils.py#L1464) — PR-introduced.

This assertion is tautological. `collector.master_config` *is* the `master_config` object constructed two lines earlier, so it asserts only that `PPOMasterConfig.model_construct` stored what it was handed. It cannot fail regardless of collector behavior.

The behavior it looks like it targets — the collector forwarding `reward_penalty_config` into the rollout call — is already genuinely asserted at [`test_async_utils.py:2165`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/tests/unit/algorithms/test_async_utils.py#L2165) inside `fake_rollouts`.

**Action:** delete the line.

### Follow-up

[`test_async_utils.py:1404`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/tests/unit/algorithms/test_async_utils.py#L1404) uses `reward_penalties = SimpleNamespace()` where [line 1441](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/tests/unit/algorithms/test_async_utils.py#L1441) in the same file uses `RewardPenaltyConfig(...)`. A fieldless stand-in would raise `AttributeError` rather than exercise the real defaults if collector code ever reads `penalize_*`. `RewardPenaltyConfig()` is already imported at line 54.

examples/configs/grpo_math_1B.yaml
```diff
@@ -70,10 +70,7 @@ reward_penalties:
   penalize_empty_final_answer: false
   penalize_unwanted_tokens: false
   penalize_malformed_think_tag: false
-  # Optional model/tokenizer-specific token IDs. Add token IDs that should
-  # be penalized when emitted by the model. Unwanted token IDs must be
-  # specified explicitly; think-tag IDs are inferred when each tag encodes
-  # to one token. Example: {unwanted: [2], think_open: 12, think_close: 13}
+  # Optional token IDs used by unwanted-token and think-tag checks.
```
bg51717: **1 action item.**

[`grpo_math_1B.yaml:73`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/examples/configs/grpo_math_1B.yaml#L73) — PR-introduced, and out of scope for this PR.

This replaces four documented lines on a **GRPO** exemplar with a one-liner, deleting both the worked example and the rule that unwanted token IDs must be specified explicitly. Nothing about PPO + NeMo Gym requires editing this file's comments, and the config-conventions guidance asks for purpose plus valid values.

**Action:** restore the original text here, and use the same fuller wording for the new copy at `examples/configs/ppo_math_1B.yaml:74` so the two exemplars stay in sync.

```suggestion
  # Optional model/tokenizer-specific token IDs. Add token IDs that should
  # be penalized when emitted by the model. Unwanted token IDs must be
  # specified explicitly; think-tag IDs are inferred when each tag encodes
  # to one token. Example: {unwanted: [2], think_open: 12, think_close: 13}
```

examples/configs/ppo_math_1B.yaml
```diff
@@ -61,6 +61,18 @@ ppo:
   # Log train/generation probability mismatch. Set a threshold to mask
   # high-error sequences; null keeps metrics-only behavior.
   seq_logprob_error_threshold: null
+  # Optional NeMo-Gym assistant-message advantage overrides.
+  invalid_tool_call_advantage: null
```
bg51717, line 66: **1 action item.**

[`ppo_math_1B.yaml:65`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/examples/configs/ppo_math_1B.yaml#L65) — PR-introduced.

The new keys share one generic header, where the identical GRPO keys at [`grpo_math_1B.yaml:51-54`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/examples/configs/grpo_math_1B.yaml#L51) each carry their own comment with a concrete value and the null-disables semantics. Since YAML is the source of truth for defaults, a reader here has no indication of what magnitude to pick or that the keys require the Gym path.

Worth noting in the comment that the value lands on **normalized** advantages (`adv_estimator.normalize_advantages: true` a few lines above) — that's the context a user needs to choose a sensible number, and PPO has no `advantage_clip_*` knob to catch an outlier.

**Action:** match the GRPO wording.

```suggestion
  # Advantage value assigned to invalid tool call tokens (e.g. -5.0 to penalize). null disables.
  # Requires the NeMo-Gym rollout path (env.should_use_nemo_gym=true).
  invalid_tool_call_advantage: null
  # Advantage value assigned to tokens with malformed <think>/</think> tags (e.g. -5.0). null disables.
  malformed_thinking_advantage: null
```

nemo_rl/algorithms/utils.py
```diff
@@ -209,6 +211,122 @@ def wrapper(*args, **kwargs):  # type: ignore
...
+
+
+def apply_message_level_advantage_penalties(
+    train_data: BatchedDataDict[Any],
+    message_logs: Sequence[Sequence[Mapping[str, Any]]],
+    invalid_tool_call_advantage: float | None,
+    malformed_thinking_advantage: float | None,
+    log_config: bool = False,
+) -> dict[str, float | int]:
+    """Overwrite actor advantages for Gym-flagged assistant-message token spans.
```
bg51717, line 239: **1 action item, 1 follow-up.**

[`utils.py:239`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/utils.py#L239) — PR-introduced.

Moving this from a private `grpo.py` helper to a public cross-algorithm util dropped its Google-style docstring. `log_config` and the five-key return shape are now undocumented on a function used by both PPO and GRPO, in a pyrefly-checked module.

**Action:** restore the fuller docstring.

```suggestion
    """Overwrite actor advantages for Gym-flagged assistant-message token spans.

    For each assistant message flagged by the NeMo-Gym detector as an invalid
    tool call or malformed thinking, overwrite that message's advantage span in
    ``train_data["advantages"]`` with the configured value. No-op when both
    override values are ``None``. The value-model ``returns`` are intentionally
    left unchanged.

    Args:
        train_data: Training batch; ``advantages`` is modified in place.
        message_logs: Batch of message logs with per-message Gym flags.
        invalid_tool_call_advantage: Advantage assigned to invalid tool calls.
        malformed_thinking_advantage: Advantage assigned to malformed thinking.
        log_config: If True, print the configured penalty values once.

    Returns:
        Penalty rate/count metrics, or an empty dict when no override is configured.
    """
```

### Follow-up

[`nemo_rl/utils/logger.py:103`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/utils/logger.py#L103) has the same regression and arguably matters more, because the name reads backwards from the effect — every call site is `if not should_log_nemo_gym_responses(...)`. The dropped text was the only place explaining that `true` means *skip* the local JSONL, and that W&B tables are controlled independently by `log_nemo_gym_full_result_tables`. One sentence there would restore it.

nemo_rl/algorithms/ppo.py
```diff
@@ -801,6 +894,31 @@ def init_sglang():
         pg.finish_generation()
         return pg, time.perf_counter() - t0
 
+    def init_policy_and_value():
+        """Initialize policy and value workers serially on the training GPUs."""
+        policy, policy_time = init_policy()
+        policy.offload_to_cpu()
```
bg51717, line 916: **1 action item.** Behavior is identical here; only the rationale was lost.

[`ppo.py:900`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L900) — PR-introduced.

The `init_policy_and_value` extraction is a clean refactor, but it dropped both load-bearing comments. The second recorded a non-obvious hazard that can't be recovered from the code: without `finish_training()`, the value worker's `__init__` runs asynchronously inside the Ray actor and can overlap vLLM generation, causing GPU OOM. A future reader now sees two unexplained calls and could remove either.

**Action:** restore both comments inside the extracted function.

```suggestion
    def init_policy_and_value():
        """Initialize policy and value workers serially on the training GPUs."""
        policy, policy_time = init_policy()
        # Block until the policy worker's __init__ completes and offload to
        # CPU, freeing GPU for value model initialization. Policy will be
        # reloaded before the vLLM refit step below.
        policy.offload_to_cpu()

        print("  ⚙️  Initializing value model for GAE...", flush=True)
        value_model, value_time = init_value()
        # Block until the value worker's __init__ completes and offload
        # model + optimizer to CPU. Without this, __init__ runs asynchronously
        # in the Ray actor and may overlap with vLLM generation, causing
        # GPU OOM.
        value_model.finish_training()
        print(f"  ✓ Value model initialized in {value_time:.2f}s", flush=True)
        return policy, policy_time, value_model, value_time
```

nemo_rl/algorithms/ppo.py
```diff
@@ -1718,6 +1908,27 @@ def ppo_train(
                         val_metrics, total_steps + 1, prefix="validation"
                     )
 
+                if should_use_nemo_gym(
```
bg51717, line 1931: **No action needed** — recording why this is correct, so nobody "fixes" it later.

[`ppo.py:1911`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/ppo.py#L1911) — PR-introduced.

This gates the new `train_data_step*.jsonl` on `should_use_nemo_gym(...) and not should_log_nemo_gym_responses(...)`, while its three siblings — async PPO at `ppo.py:3149`, [sync GRPO](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/grpo_sync.py#L1239), and [async GRPO](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/nemo_rl/algorithms/grpo.py#L5006) — use the second condition alone. That asymmetry looks like an oversight and isn't.

Pre-PR, sync `ppo_train` wrote no train-data JSONL at all. Dropping the `should_use_nemo_gym` conjunct would resolve to `not bool(None)` on a non-Gym run and start writing a per-step file for **every existing non-Gym sync PPO job** — a new artifact and disk cost on runs this PR shouldn't touch. Scoping it to the Gym path is the conservative, correct call.

**Action (optional):** a one-line comment here stating the gate is Gym-only on purpose would stop a future reader from unifying it.

docs/guides/ppo.md
```diff
@@ -63,7 +63,7 @@ Set `ppo.async_ppo.enabled: true` to overlap rollout generation with training. A
 
 Async PPO reuses the trajectory collector, replay buffer, and weight-versioning infrastructure described in the [Async GRPO guide](async-grpo.md); this section focuses on PPO-specific behavior and constraints.
 
-Async PPO requires non-colocated vLLM generation with `vllm_cfg.async_engine: true`, `loss_fn.use_importance_sampling_correction: true`, and `loss_fn.force_on_policy_ratio: false`. Dynamic sampling, reward scaling, reward shaping, multiple dataloaders, NeMo Gym, colocated generation, and FP8 KV-scale synchronization are not supported yet.
+Async PPO requires non-colocated vLLM generation with `vllm_cfg.async_engine: true`, `loss_fn.use_importance_sampling_correction: true`, and `loss_fn.force_on_policy_ratio: false`. Dynamic sampling, `ppo.reward_scaling`, `ppo.reward_shaping`, multiple dataloaders, colocated generation, and FP8 KV-scale synchronization are not supported yet. NeMo Gym rollouts are supported through [run_ppo_nemo_gym.py](../../examples/nemo_gym/run_ppo_nemo_gym.py), including Gym reward penalties, effort-level shaping, sample masking, and message-level advantage overrides.
```
bg51717: **No action needed** — one option if you want it.

[`ppo.md:66`](https://github.com/NVIDIA-NeMo/RL/blob/1a7ac065d28c6ee818f5887740ab6f22e1979e07/docs/guides/ppo.md#L66) — PR-introduced.

The sentence is accurate, and the preceding sentence already lists all three async requirements, so the doc isn't misleading. The gap is in the recipe rather than the prose: `examples/nemo_gym/ppo_math_rlvr_nemo_gym.yaml` inherits `max_num_epochs: 100000`, `use_importance_sampling_correction: false`, and `colocated.enabled: true` and overrides none of them, so `run_ppo_nemo_gym.py ppo.async_ppo.enabled=true` hits three sequential fail-fasts. `tests/functional/ppo_async_gym.sh` confirms this by passing all three as CLI overrides.

**Action (optional):** either note that the example recipe is configured for synchronous PPO and point at `ppo_async_gym.sh` for the async delta, or ship a small async variant of the recipe. Purely a convenience call.

bg51717 (commented):
## Team review — PPO + NeMo Gym

Reviewed by a coordinated agent team (RL, NeMo Gym, tests, bug scan, design) followed by an adversarial verification pass. `pre-commit run --all-files` passes cleanly.

**The one finding with reach outside this PR** is `trajectory_collector.py:938`: because the async collector is shared, enabling effort shaping for async PPO also enables it for async GRPO, where it had been silently dropped. That's the right fix — it can't be scoped to PPO without institutionalizing the bug — but it moves reward numerics on 8 shipped recipes and no CI job covers them, so it's worth an explicit callout rather than leaving it to be discovered from stored reference metrics.

Everything else is PPO-local: three paths that fail late and should fail at setup (`reward_scaling`/`reward_shaping` on a Gym batch, `top_k`/`val_top_k`, `max_val_samples`), two test-coverage gaps demonstrated by mutation (both mutants leave all 97 tests green), and a small cluster of comments/docstrings lost during the refactor.

**Checked and found correct**, so these don't need re-litigating: the `grpo.py` to `utils.py`/`rollouts.py` extraction is behavior-preserving, and the helpers take the narrowest config slice with zero algorithm conditionals; `rollouts.py` is purely additive (+69/-1, the deletion being an import line) so GRPO's paths are untouched and the relocated `RewardPenaltyConfig` validator fires at exactly the same lifecycle point; the Gym task-cursor save/restore is a faithful port of GRPO's including the `max()` reconciliation and the missing-`rollouts.pt` degradation; no test coverage was lost in the extraction (`TestMaskSampleFilter` relocated verbatim, and net coverage increases since `_get_effort_config` and `_should_log_nemo_gym_responses` had none before); and `apply_message_level_advantage_penalties` correctly leaves `returns` untouched — the most PPO-specific correctness question in the diff, since PPO has a critic and GRPO doesn't.

Worth saying explicitly: moving `validate_async_ppo_config` out of the launcher into `ppo.py` is better than the GRPO analog, which still inlines its guards in `run_grpo_nemo_gym.py` and can drift between its two launchers.

A few things were investigated and deliberately **not** raised: `policy.router_replay.enabled` with PPO already fails loudly at the first train step, so the hardcoded `enable_router_replay=False` needs no guard; `examples/run_ppo.py` discarding the Gym actor matches `run_grpo.py` and `run_distillation.py` exactly, so it is a repo-wide convention rather than this PR's debt; and an empty validation set is rejected by `StatefulDataLoader` before any division, so there is no `ZeroDivisionError`.

Held as follow-ups: `distillation.py` silently ignoring `env.nemo_gym.effort_levels` (same defect class, not introduced here), and `sync_rollout_actor.py:257` still hand-inlining the helper this PR extracted.

_Generated by Claude Code_

bg51717:
/ok to test [4999ad4](https://github.com/NVIDIA-NeMo/RL/pull/3732/commits/4999ad4a3e4b07ecfd6e2d6d8e99d08c57a030e0)

nemo_rl/algorithms/async_utils/trajectory_collector.py
bg51717: @yuki-97 .This PR needs to forward effort_levels through the shared async collector for Async PPO. It also fixes Async GRPO, which previously ignored its configured effort shaping, so the affected recipes will now have different reward numerics. Would you prefer to keep and document/test that Async GRPO fix in this PR, or track the recipe rebaseline and end-to-end coverage separately?

nemo_rl/algorithms/ppo.py
bg51717: Agreed. I added setup-time guards for PPO reward scaling and reward shaping on the NeMo Gym path, with targeted coverage for both invalid configurations before worker setup.

nemo_rl/algorithms/ppo.py
bg51717: Agreed. I added unconditional setup-time guards for both training `top_k` and validation `val_top_k` on the NeMo Gym path, with targeted coverage for each key.

nemo_rl/algorithms/ppo.py
bg51717: Agreed. I added the sibling setup assertion for `ppo.max_val_samples` and targeted coverage for an unresolved value before worker initialization.

tests/unit/algorithms/test_ppo.py
bg51717: Agreed. I parameterized the restore test to cover `rollouts.pt` being ahead, a missing `rollouts.pt`, and replay metadata being ahead, so both cursor sources and the `max()` reconciliation are protected.

tests/unit/algorithms/test_ppo.py
bg51717: Agreed. I threaded `mask_sample` through the PPO loop harness and added sync/async coverage verifying that Gym-flagged samples are zeroed in the training `sample_mask` before the loss is built.

examples/configs/grpo_math_1B.yaml
bg51717: Agreed. I restored the original GRPO token-ID guidance and used the same complete wording and example in the PPO exemplar.

examples/configs/ppo_math_1B.yaml
bg51717: Agreed. I replaced the generic header with field-specific guidance matching GRPO, including example penalty values, `null` semantics, and the NeMo Gym requirement.

nemo_rl/algorithms/utils.py
bg51717: Agreed. I restored the complete shared advantage-helper docstring and clarified that `should_log_nemo_gym_responses=true` delegates full response logging to Gym, skips the local training JSONL, and is independent of W&B full-result tables.

nemo_rl/algorithms/ppo.py
bg51717: Agreed. I restored both initialization-order comments inside `init_policy_and_value`, including the asynchronous value-worker initialization and GPU-residency rationale for `finish_training()`.

PR #3738, opened 2026-08-20
fix(vlm): propagate NeMo Gym dynamic image padding

youngeunkwon0405:
# What does this PR do?

Propagates the existing NeMo-Gym `pad_dynamic_image_shapes` option through the initial deduplicated multimodal attachment path.

When `grpo.deduplicate_multimodal_data=true`, GRPO attaches the initial image-model inputs before repeating prompts. That path previously called `attach_image_model_inputs_to_message` without forwarding the environment's dynamic-padding option. A heterogeneous multi-image example could therefore attempt to stack different image resolutions before rollout even when the NeMo-Gym environment requested padding.

Before this fix, propagation stopped at the algorithm-to-rollout-helper boundary:

```text
master_config.env.nemo_gym.pad_dynamic_image_shapes
  -> grpo_train / validate / AsyncTrajectoryCollector._process_batch
  -> attach_initial_nemo_gym_image_payloads(batch, processor)
  -> attach_image_model_inputs_to_message(...)
       pad_dynamic_image_shapes=False  # implicit default
```

The synchronous training, validation, and asynchronous collection call sites invoked `attach_initial_nemo_gym_image_payloads` without the flag. That helper then invoked `attach_image_model_inputs_to_message` without the flag as well, so the target function silently used its existing `False` default. This early deduplicated attachment shortcut is separate from the normal NeMo-Gym environment preprocessing path, which already carried the option correctly.

After this fix, the complete chain is:

```text
master_config.env.nemo_gym.pad_dynamic_image_shapes
  -> grpo_train / validate / AsyncTrajectoryCollector._process_batch
  -> attach_initial_nemo_gym_image_payloads(..., pad_dynamic_image_shapes=...)
  -> attach_image_model_inputs_to_message(..., pad_dynamic_image_shapes=...)
```

This change:

- adds a keyword-only `pad_dynamic_image_shapes` argument to `attach_initial_nemo_gym_image_payloads`;
- forwards `env.nemo_gym.pad_dynamic_image_shapes` from synchronous training, validation, and asynchronous trajectory collection; and
- preserves the existing `False` default for every caller that does not opt in.

This PR is independent of the 32-GPU recipe PR and independent of #3721.

# Issues

None.

# Validation

- `uvx --from ruff==0.9.9 ruff check nemo_rl/algorithms/async_utils/trajectory_collector.py nemo_rl/algorithms/grpo.py nemo_rl/experience/rollouts.py tests/unit/experience/test_rollouts.py`
- `uvx --from ruff==0.9.9 ruff format --check nemo_rl/algorithms/async_utils/trajectory_collector.py nemo_rl/algorithms/grpo.py nemo_rl/experience/rollouts.py tests/unit/experience/test_rollouts.py`
- `PYTHONPATH=. uv run --extra nemo_gym --group test pytest -q --noconftest tests/unit/experience/test_rollouts.py -k attach_initial_nemo_gym_image_payloads` — 3 passed
- GPU Slurm validation job `6370288` — `COMPLETED 0:0`
- Ten-step private NeMo-Gym sync and async runs completed without mixed-image padding failures: [sync W&B](https://wandb.ai/nvidia/youngeunk-nano-omni-perf-study/runs/3kd1t7yh), [async W&B](https://wandb.ai/nvidia/youngeunk-nano-omni-perf-study/runs/d9ie6duc)

- [x] Added focused unit coverage for flag propagation and the default behavior
- [x] Ran focused Linux GPU-container tests
- [x] No documentation update is needed because this fixes existing internal option propagation without adding a new public option


nemo_rl/algorithms/grpo.py
```diff
@@ -2841,7 +2841,15 @@ def grpo_train(
                         master_config.grpo.deduplicate_multimodal_data
                         and should_use_nemo_gym(master_config)
                     ):
-                        attach_initial_nemo_gym_image_payloads(batch, processor)
+                        attach_initial_nemo_gym_image_payloads(
+                            batch,
+                            processor,
+                            pad_dynamic_image_shapes=bool(
+                                master_config.env.get("nemo_gym", {}).get(
```
youngeunkwon0405: [`nemo_rl/algorithms/grpo.py:2848`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/algorithms/grpo.py#L2848)

**[GUIDELINE]** The chained `.get("nemo_gym", {}).get("pad_dynamic_image_shapes", False)` pattern is one the config-conventions guideline asks us to avoid (chained `.get()` with hidden defaults at each level). Since `pad_dynamic_image_shapes` is `NotRequired[bool]` ([`nemo_gym.py:215`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/environments/nemo_gym.py#L215)), the established read is truthiness with no non-`None` default, as the env actor itself does: [`nemo_gym.py:471`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/environments/nemo_gym.py#L471) `bool(cfg.get("pad_dynamic_image_shapes"))`.

The same expression is also repeated verbatim at [`grpo.py:3880`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/algorithms/grpo.py#L3880) and [`trajectory_collector.py:525`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/algorithms/async_utils/trajectory_collector.py#L525). Consider a small helper next to the existing analog [`get_nemo_gym_thinking_tags`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/experience/rollouts.py#L1777), which resolves both points and gives the flag-read a cheap unit-test target:

```python
def get_nemo_gym_pad_dynamic_image_shapes(env_config: dict[str, Any]) -> bool:
    """Read ``env.nemo_gym.pad_dynamic_image_shapes``; absent means False."""
    nemo_gym_config = env_config.get("nemo_gym")
    if isinstance(nemo_gym_config, dict):
        return bool(nemo_gym_config.get("pad_dynamic_image_shapes"))
    return False
```

Behavior is identical either way — low-severity polish, not a bug.

tests/unit/experience/test_rollouts.py
```diff
@@ -115,21 +115,24 @@ class _Processor:
...
+        batch, processor, pad_dynamic_image_shapes=True
+    )
     rollouts_mod.attach_initial_nemo_gym_image_payloads(batch, processor)
 
     assert len(calls) == 1
     assert calls[0][0] is batch["message_log"][0][0]
     assert calls[0][1][0].size == (2, 3)
     assert calls[0][2] is processor
+    assert calls[0][3] is True
     assert batch["message_log"][0][0]["pixel_values"] is attached
```
youngeunkwon0405, line 134: [`tests/unit/experience/test_rollouts.py:136`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/tests/unit/experience/test_rollouts.py#L136)

**[TEST]** The default-`False` forwarding isn't actually exercised: the second (no-flag) call at [line 129](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/tests/unit/experience/test_rollouts.py#L129) is a no-op because the idempotency guard at [`rollouts.py:113`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/experience/rollouts.py#L113) skips messages whose `pixel_values` is already a `PackedTensor`, so `fake_attach` never sees the default. A regression that hard-codes `pad_dynamic_image_shapes=True` in the helper would still pass. A fresh batch closes the gap (verified passing locally):

```suggestion
    assert batch["message_log"][0][0]["pixel_values"] is attached

    # A fresh batch with the flag omitted must forward the False default.
    default_batch = _initial_gym_image_batch()
    rollouts_mod.attach_initial_nemo_gym_image_payloads(default_batch, processor)
    assert len(calls) == 2
    assert calls[1][3] is False
```

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -520,7 +520,13 @@ def _process_batch(self, batch: BatchedDataDict[DatumSpec]) -> None:
                 self._stamp_nemo_gym_task_indices(rollout_batch)
                 if self._deduplicate_multimodal_data:
                     attach_initial_nemo_gym_image_payloads(
-                        rollout_batch, self.processor
+                        rollout_batch,
+                        self.processor,
+                        pad_dynamic_image_shapes=bool(
```
youngeunkwon0405: [`nemo_rl/algorithms/async_utils/trajectory_collector.py:525`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/algorithms/async_utils/trajectory_collector.py#L525)

**[TEST — optional]** The unit test covers the helper's forwarding, but none of the three call sites' config reads are tested. `grpo_train`/`validate` sit too deep in the training loop to unit-test cheaply (not worth it), but `TestAsyncTrajectoryCollector` in `tests/unit/algorithms/test_async_utils.py` already unit-tests `_process_batch` with the same monkeypatched-thread approach (e.g. the gym-stamping test at [`test_async_utils.py:1831`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/tests/unit/algorithms/test_async_utils.py#L1831)), so one parametrized test covers the shared read expression for real against the real `MasterConfig.env` (verified passing locally, 2 passed; the `False` leg runs with no `"nemo_gym"` key, so it exercises both `.get` defaults):

```python
    @pytest.mark.parametrize("pad_flag", [False, True])
    def test_process_batch_forwards_gym_pad_dynamic_image_shapes(
        self, monkeypatch, pad_flag
    ):
        """Dedup-enabled Gym forwards env.nemo_gym.pad_dynamic_image_shapes."""

        class RemoteMethod:
            def remote(self, *args, **kwargs):
                return 2

        class FakeReplayBuffer:
            get_trajectories_needed = RemoteMethod()

        class RecordingThread:
            def __init__(self, *, target, daemon):
                assert daemon
                self.target = target

            def start(self):
                self.target()

            def is_alive(self):
                return False

        target_weight = 11
        collector = self.create_local_collector(replay_buffer=FakeReplayBuffer())
        collector.running = True
        collector._deduplicate_multimodal_data = True
        if pad_flag:
            collector.master_config.env = {
                "nemo_gym": {"pad_dynamic_image_shapes": True}
            }

        def reserve_target(generation_weight_version):
            collector._generating_targets.add(target_weight)
            return target_weight

        async def capture_batch(**kwargs):
            collector._release_target(target_weight)

        attach_flags = []

        def fake_attach(batch, processor, *, pad_dynamic_image_shapes=False):
            attach_flags.append(pad_dynamic_image_shapes)

        collector._get_next_target_for_generation = reserve_target
        collector._run_rollout_batch_worker = capture_batch
        monkeypatch.setattr(trajectory_collector_mod.ray, "get", lambda value: value)
        monkeypatch.setattr(
            trajectory_collector_mod, "should_use_nemo_gym", lambda config: True
        )
        monkeypatch.setattr(
            trajectory_collector_mod,
            "attach_initial_nemo_gym_image_payloads",
            fake_attach,
        )
        monkeypatch.setattr(
            trajectory_collector_mod._threading, "Thread", RecordingThread
        )

        collector._process_batch(self.create_mock_batch(size=2))

        assert attach_flags == [pad_flag]
```

nemo_rl/experience/rollouts.py
```diff
@@ -77,6 +77,8 @@
 def attach_initial_nemo_gym_image_payloads(
     batch: BatchedDataDict[DatumSpec],
     processor: Any,
+    *,
+    pad_dynamic_image_shapes: bool = False,
 ) -> None:
     """Attach initial Gym image tensors once, before prompt repeat.
```
youngeunkwon0405, line 86: [`nemo_rl/experience/rollouts.py:83`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/experience/rollouts.py#L83)

**[DOCSTRING — nit]** The new `pad_dynamic_image_shapes` parameter isn't documented. The docstring has no `Args:` section at all (pre-existing), so this is optional polish; if you touch it, something like:

```
    Args:
        pad_dynamic_image_shapes: Forwarded ``env.nemo_gym.pad_dynamic_image_shapes``;
            normalizes heterogeneous image tensor shapes while retaining exact sizes.
```

nemo_rl/experience/rollouts.py
```diff
@@ -77,6 +77,8 @@
 def attach_initial_nemo_gym_image_payloads(
     batch: BatchedDataDict[DatumSpec],
     processor: Any,
+    *,
+    pad_dynamic_image_shapes: bool = False,
```
youngeunkwon0405: [`nemo_rl/experience/rollouts.py:81`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/experience/rollouts.py#L81)

**[OBSERVATION — pre-existing, out of PR scope]** While auditing callers of this helper we noticed the single-controller path never calls it: [`SyncRolloutActor`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/experience/sync_rollout_actor.py#L264) passes `deduplicate_multimodal_data` into the gym rollout (which then [omits initial media](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/environments/nemo_gym.py#L690)), and [`_reattach_original_multimodal_payloads`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/experience/rollouts.py#L2423) restores from a message log that never had payloads attached — so SC + gym + dedup would silently drop initial image tensors. Dedup with the data plane is already rejected in `grpo.setup` ([`grpo.py:386`](https://github.com/NVIDIA-NeMo/RL/blob/bd2996e6def18e4f6ba70c112100b71c83c21922/nemo_rl/algorithms/grpo.py#L386)), but `validate_single_controller_config` has no dedup check, so the misconfiguration fails silently instead of loudly. Suggest triaging separately: adding the dedup guard to SC config validation seems like the cheap fix. Not a blocker for this PR.

youngeunkwon0405 (commented):
This PR was reviewed by a coordinated team of 4 specialized agents (RL-codebase expert, test reviewer, independent bug scanner, and a devil's-advocate verification pass). Summary:

- **The fix is correct and needed.** Pre-PR, the dedup attachment path always invoked the processor with stacking semantics, breaking heterogeneous multi-image gym prompts even when the environment requested padding; the gym-actor side already honored the flag. All three production call sites are updated and no others exist.
- **No bugs found.** Config key path, `master_config.env` typing at all three call sites, and the downstream `attach_image_model_inputs_to_message` keyword signature were all verified against the code. Ruff check/format pass. The updated unit test was run locally and passes.
- The 5 inline comments are all low-severity polish (config-read convention + shared helper, two optional verified test hardenings, a docstring nit) plus one pre-existing out-of-scope observation. **Nothing blocks merge.**

Validation evidence (W&B sync/async runs + Slurm job) is appropriate for a correctness fix — thanks for including it.

_Generated by Claude Code_

tests/unit/experience/test_rollouts.py
youngeunkwon0405: Addressed in 36ed0446. The test now creates a fresh batch after the idempotency assertion and verifies that the second real attach call forwards pad_dynamic_image_shapes=False. Validation: OCI-HSG job 6383465 completed 0:0 with 7 passed and 43 deselected.

nemo_rl/algorithms/grpo.py
youngeunkwon0405: Addressed in 36ed0446. Added get_nemo_gym_pad_dynamic_image_shapes next to get_nemo_gym_thinking_tags, replaced all three chained config reads in GRPO training, validation, and the async trajectory collector, and added focused coverage for missing, empty, false, and true configurations. Validation: OCI-HSG job 6383465 completed 0:0 with 7 passed and 43 deselected; Ruff 0.9.9 lint and format checks also passed.

nemo_rl/experience/rollouts.py
youngeunkwon0405: Out of PR scope so will ignore atm

PR #3741, opened 2026-08-20
feat(sc): apply reward penalties to Gym rollouts

macandro96:
# What does this PR do ?

Support reward penalties for gym rollouts.

# Issues

# Usage
```
# Reward-zeroing penalties applied to NeMo-Gym rollout results.
reward_penalties:
  penalize_duplicated_reasoning: false
  penalize_empty_final_answer: false
  penalize_unwanted_tokens: false
  penalize_malformed_think_tag: false
  # Optional model/tokenizer-specific IDs. Example:
  # token_ids: {unwanted: [2], think_open: 12, think_close: 13}
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/single_controller_utils/config.py
+++ b/nemo_rl/algorithms/single_controller_utils/config.py
@@ -35,7 +35,12 @@ from nemo_rl.algorithms.async_utils.staleness_sampler import (
     SamplerConfig,
     required_buffer_capacity_for_config,
 )
-from nemo_rl.algorithms.grpo import GRPOConfig, GRPOLoggerConfig
+from nemo_rl.algorithms.grpo import (
+    _REWARD_PENALTY_FLAGS,
+    GRPOConfig,
+    GRPOLoggerConfig,
+    RewardPenaltyConfig,
+)
 from nemo_rl.algorithms.loss import ClippedPGLossConfig
 from nemo_rl.algorithms.loss.loss_functions import MseValueLossConfig
 from nemo_rl.algorithms.opd import OnPolicyDistillationConfig
@@ -586,6 +591,7 @@ class MasterConfig(BaseModel, extra="allow"):
     logger: GRPOLoggerConfig
     cluster: ClusterConfig
     checkpointing: CheckpointingConfig
+    reward_penalties: RewardPenaltyConfig = Field(default_factory=RewardPenaltyConfig)
     data_plane: DataPlaneConfig
     async_rl: AsyncRLConfig
     on_policy_distillation: Optional[OnPolicyDistillationConfig] = None
@@ -932,6 +938,15 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
 
     async_config = master_config.async_rl
     algo_cfg = algo_config(master_config)
+
+    reward_penalties_enabled = any(
+        getattr(master_config.reward_penalties, flag) for flag in _REWARD_PENALTY_FLAGS
+    )
+    if reward_penalties_enabled and not master_config.env.get("should_use_nemo_gym"):
+        raise ValueError(
+            "reward_penalties require the NeMo-Gym rollout path "
+            "(env.should_use_nemo_gym=true) on SingleController"
+        )
 
     if algo_cfg.num_prompts_per_step < async_config.min_groups_for_streaming_train:
         raise ValueError(
--- a/nemo_rl/experience/rollout_manager.py
+++ b/nemo_rl/experience/rollout_manager.py
@@ -57,7 +57,9 @@ from nemo_rl.experience.rollouts import (
     _find_routed_experts_template,
     _tensorize_by_key,
     attach_static_multimodal_payload,
+    apply_reward_penalties,
     calculate_rewards,
+    compute_reward_penalty_metrics,
 )
 from nemo_rl.models.generation.interfaces import (
     GenerationConfig,
@@ -765,6 +767,7 @@ class AsyncNemoGymRolloutImpl:
         max_rollout_turns: int,
         generation_config: GenerationConfig,
         mask_env_flagged_samples: bool = True,
+        reward_penalty_config: Optional[dict[str, Any]] = None,
         # Optional so direct construction does not have to carry the resiliency wiring;
         # RolloutManager always passes both explicitly.
         timeouts: Optional[RolloutTimeouts] = None,
@@ -783,6 +786,7 @@ class AsyncNemoGymRolloutImpl:
         self._max_rollout_turns = max_rollout_turns
         self._generation_config = generation_config
         self._mask_env_flagged_samples = mask_env_flagged_samples
+        self._reward_penalty_config = reward_penalty_config
         self._timeouts = timeouts if timeouts is not None else RolloutTimeouts()
         self._max_gym_row_attempts = (
             retry_policy
@@ -1018,10 +1022,11 @@ class AsyncNemoGymRolloutImpl:
             # All N rollouts share the same input prompt; tensorize one copy.
             prompt_message_log = completed_results[0]["input_message_log"]
             _tensorize_by_key(prompt_message_log, "token_ids")
-            # Convert results to completions.
-            completions = [
-                self._result_to_completion(result) for result in completed_results
-            ]
+            # Apply penalties before Completion captures each result's reward, while
+            # preserving the batch-level counts used by legacy Gym metrics.
+            completions, penalty_counts = self._results_to_completions(
+                completed_results
+            )
 
         # Compute rollout metrics.
         with timer.time(f"{timer_prefix}/compute_metrics"):
@@ -1030,36 +1035,57 @@ class AsyncNemoGymRolloutImpl:
             )
             # Same helper the batched path uses, so the two cannot drift apart.
             rollout_metrics.update(_effort_shaping_metrics(shaping))
+            rollout_metrics.update(
+                self._compute_reward_penalty_metrics(
+                    penalty_counts, len(completed_results)
+                )
+            )
 
         rollout_metrics.update(env_timing_metrics)
 
         return completions, prompt_message_log, rollout_metrics
 
-    def _result_to_completion(self, result: dict) -> Completion:
-        """Convert one run_rollouts result dict into a Completion."""
-        # Tensorize token fields.
-        _tensorize_by_key(result["message_log"], "token_ids")
-        _tensorize_by_key(
-            [m for m in result["message_log"] if m["role"] == "assistant"],
-            "generation_logprobs",
-        )
-        # Calculate truncation.
-        truncated = (
-            sum(len(m["token_ids"]) for m in result["message_log"]) == self._max_seq_len
-        )
-
-        # Same gate as the batched path: when masking is off, drop the env
-        # mask flag so later batch building never sees it.
-        if not self._mask_env_flagged_samples:
-            (result["full_result"].get("instance_config") or {}).pop(
-                "mask_sample", None
+    def _results_to_completions(
+        self, results: list[dict]
+    ) -> tuple[list[Completion], dict[str, int]]:
+        """Apply configured penalties and convert a Gym result batch."""
+        for result in results:
+            _tensorize_by_key(result["message_log"], "token_ids")
+            _tensorize_by_key(
+                [m for m in result["message_log"] if m["role"] == "assistant"],
+                "generation_logprobs",
             )
 
-        return Completion(
-            message_log=result["message_log"],
-            env_extras=result["full_result"],
-            truncated=truncated,
-            reward=float(result["full_result"]["reward"]),
+            # Same gate as the batched path: when masking is off, drop the env
+            # mask flag so later batch building never sees it.
+            if not self._mask_env_flagged_samples:
+                (result["full_result"].get("instance_config") or {}).pop(
+                    "mask_sample", None
+                )
+
+        penalty_counts = apply_reward_penalties(results, self._reward_penalty_config)
+        completions = []
+        for result in results:
+            truncated = (
+                sum(len(m["token_ids"]) for m in result["message_log"])
+                == self._max_seq_len
+            )
+            completions.append(
+                Completion(
+                    message_log=result["message_log"],
+                    env_extras=result["full_result"],
+                    truncated=truncated,
+                    reward=float(result["full_result"]["reward"]),
+                )
+            )
+        return completions, penalty_counts
+
+    def _compute_reward_penalty_metrics(
+        self, penalty_counts: dict[str, int], num_results: int
+    ) -> dict[str, float]:
+        """Return enabled penalty rates using the legacy Gym metric names."""
+        return compute_reward_penalty_metrics(
+            penalty_counts, num_results, self._reward_penalty_config
         )
 
     def _compute_rollout_metrics(
@@ -1154,6 +1180,7 @@ class RolloutManager:
         generation_config: Optional[GenerationConfig] = None,
         use_nemo_gym: bool = False,
         mask_env_flagged_samples: bool = True,
+        reward_penalty_config: Optional[dict[str, Any]] = None,
         tq_buffer: Optional[TQReplayBuffer] = None,
         timeouts: Optional[RolloutTimeouts] = None,
         retry_policy: Optional[RolloutRetryPolicy] = None,
@@ -1193,6 +1220,7 @@ class RolloutManager:
             generation_config=generation_config,
             # Only used by AsyncNemoGymRolloutImpl; AsyncRolloutImpl ignores it.
             mask_env_flagged_samples=mask_env_flagged_samples,
+            reward_penalty_config=reward_penalty_config,
             # None means "no deadlines", which is what async_rl's own defaults resolve
             # to; callers that have a config pass the resolved values in.
             timeouts=timeouts if timeouts is not None else RolloutTimeouts(),
--- a/nemo_rl/experience/rollout_manager.py
+++ b/nemo_rl/experience/rollout_manager.py
@@ -56,8 +56,8 @@ from nemo_rl.experience.rollouts import (
     _effort_shaping_metrics,
     _find_routed_experts_template,
     _tensorize_by_key,
-    attach_static_multimodal_payload,
     apply_reward_penalties,
+    attach_static_multimodal_payload,
     calculate_rewards,
     compute_reward_penalty_metrics,
 )
@@ -1085,7 +1085,9 @@ class AsyncNemoGymRolloutImpl:
     ) -> dict[str, float]:
         """Return enabled penalty rates using the legacy Gym metric names."""
         return compute_reward_penalty_metrics(
-            penalty_counts, num_results, self._reward_penalty_config
+            penalty_counts,
+            num_results,
+            self._reward_penalty_config,
         )
 
     def _compute_rollout_metrics(
--- a/tests/unit/experience/test_rollout_manager.py
+++ b/tests/unit/experience/test_rollout_manager.py
@@ -612,12 +612,16 @@ def _mask_gate_result():
 
 
 def test_result_to_completion_keeps_mask_flag_when_gate_on():
-    completion = _nemo_gym_impl(True)._results_to_completions([_mask_gate_result()])[0][0]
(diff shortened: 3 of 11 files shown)


nemo_rl/experience/rollout_manager.py
```diff
@@ -980,48 +998,79 @@ async def _run_rollouts(
             # All N rollouts share the same input prompt; tensorize one copy.
             prompt_message_log = completed_results[0]["input_message_log"]
             _tensorize_by_key(prompt_message_log, "token_ids")
-            # Convert results to completions.
-            completions = [
-                self._result_to_completion(result) for result in completed_results
-            ]
+            # Apply penalties before Completion captures each result's reward, while
+            # preserving the batch-level counts used by legacy Gym metrics.
+            completions, penalty_counts = self._results_to_completions(
```
macandro96, line 1027: **1 action item.**

**TL;DR — this line rewires `_run_rollouts` to call `_results_to_completions` directly, which breaks a pre-existing stream-ordering test that monkey-patches `_result_to_completion` (singular). CI will fail.**

Not in this diff, but the test that breaks is [`tests/unit/experience/test_rollouts.py::test_rollout_manager_consumes_stream_and_restores_input_order`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/tests/unit/experience/test_rollouts.py#L2123). It monkey-patches the singleton at [line 2191](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/tests/unit/experience/test_rollouts.py#L2191):
```python
manager._result_to_completion = lambda result: result["value"]
```
Under the new wiring, the patched attribute is never read — `_run_rollouts` at [`rollout_manager.py:1003`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/experience/rollout_manager.py#L1003) calls `self._results_to_completions(completed_results)` directly. The real `_results_to_completions` at [`rollout_manager.py:1032`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/experience/rollout_manager.py#L1032) then reads `result["message_log"]` on the stub result dicts, which only carry `{"value", "input_message_log"}`.

Reproduced locally on this head:
```
FAILED tests/unit/experience/test_rollouts.py::test_rollout_manager_consumes_stream_and_restores_input_order
nemo_rl/experience/rollout_manager.py:1003: in _run_rollouts
nemo_rl/experience/rollout_manager.py:1032: in _results_to_completions
E  KeyError: 'message_log'
```
The sibling `test_rollout_manager_rejects_duplicate_stream_rows` still passes because it fails earlier on the duplicate-detection check.

### AI-1
In `tests/unit/experience/test_rollouts.py`, replace the monkey-patch at line 2191 (there's no way to attach a `suggestion` block to a file outside this diff — please apply manually):
```python
manager._results_to_completions = lambda results: (
    [r["value"] for r in results],
    {},
)
```
Delete the surrounding `_result_to_completion` line if the test no longer references it. Then re-run `cd tests && uv run pytest unit/experience/test_rollouts.py::test_rollout_manager_consumes_stream_and_restores_input_order -x` to confirm green.

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -679,6 +680,21 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
...
     num_prompts_per_step = master_config.grpo.num_prompts_per_step
+
+    reward_penalties_enabled = any(
+        (
+            master_config.reward_penalties.penalize_duplicated_reasoning,
+            master_config.reward_penalties.penalize_empty_final_answer,
+            master_config.reward_penalties.penalize_unwanted_tokens,
+            master_config.reward_penalties.penalize_malformed_think_tag,
+        )
+    )
```
macandro96, line 944: **1 action item.**

**TL;DR — this SC validator open-codes the four penalty-flag names, duplicating the non-SC helper [`_raise_if_reward_penalties_enabled_without_nemo_gym`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/algorithms/grpo.py#L1966) which already iterates the canonical [`_REWARD_PENALTY_FLAGS`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/algorithms/grpo.py#L233) tuple. A future 5th penalty added to `_REWARD_PENALTY_FLAGS` will be auto-picked up by the non-SC validator and `apply_reward_penalties`, but silently skipped here — SC users could enable it with `env.should_use_nemo_gym=false` and get no error.**

[`single_controller_utils/config.py:37`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/algorithms/single_controller_utils/config.py#L37) already imports from `nemo_rl.algorithms.grpo`, so extending that import to include `_REWARD_PENALTY_FLAGS` creates no new cycle.

### AI-1
One way — of several:
```suggestion
    reward_penalties_enabled = any(
        getattr(master_config.reward_penalties, flag)
        for flag in _REWARD_PENALTY_FLAGS
    )
    if reward_penalties_enabled and not master_config.env.get("should_use_nemo_gym"):
        raise ValueError(
            "reward_penalties require the NeMo-Gym rollout path "
            "(env.should_use_nemo_gym=true) on SingleController"
        )
```
(Add `_REWARD_PENALTY_FLAGS` to the existing `from nemo_rl.algorithms.grpo import ...` at line 37.) Alternatively, promote `_raise_if_reward_penalties_enabled_without_nemo_gym` itself into a shared `nemo_rl/algorithms/reward_penalty_utils.py` and call it from both validators.

nemo_rl/experience/rollout_manager.py
```diff
@@ -55,6 +56,21 @@
...
+    ),
+    "empty_final_answer": (
+        "penalize_empty_final_answer",
+        "empty_final_answer_rate",
+    ),
+    "unwanted_token": ("penalize_unwanted_tokens", "unwanted_token_rate"),
+    "malformed_think_tag": (
+        "penalize_malformed_think_tag",
+        "malformed_think_tag_rate",
+    ),
```
macandro96: **1 action item.**

**TL;DR — `_REWARD_PENALTY_METRICS` here is byte-for-byte identical to the function-local `_PENALTY_METRICS` at [`nemo_rl/experience/rollouts.py:2765-2779`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/experience/rollouts.py#L2765) inside `run_nemo_gym_rollout_sync`. Any future rename (or new penalty) needs updating both files, and drift would be silent — the two paths would emit different metric names for the same penalty count.**

Not PR-introduced; the duplication is created by this PR. Devil-advocate confirmed the two dicts contain the same 4 (flag, metric_name) pairs.

### AI-1
Promote the mapping to a module-level constant in `nemo_rl/experience/rollouts.py`, import it here, and drop the local copy. Even better — since `_compute_reward_penalty_metrics` (rollout_manager.py:1062-1073) mirrors the metric-emission loop inside `run_nemo_gym_rollout_sync`, expose a shared `compute_reward_penalty_metrics(counts, num_results, config)` helper next to `apply_reward_penalties` (rollouts.py:1936) and reduce this file's method to a one-line delegate. That collapses both duplications in one go.

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -542,6 +542,7 @@ class MasterConfig(BaseModel, extra="allow"):
     logger: GRPOLoggerConfig
     cluster: ClusterConfig
     checkpointing: CheckpointingConfig
+    reward_penalties: RewardPenaltyConfig = Field(default_factory=RewardPenaltyConfig)
```
macandro96, line 594: **1 action item.**

**TL;DR — this new field is now a first-class knob on `MasterConfig`, but the SC exemplar YAML doesn't advertise it — users discovering the SC recipe won't know reward penalties exist.**

The non-SC exemplar [`examples/configs/grpo_math_1B.yaml:67-77`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/examples/configs/grpo_math_1B.yaml#L67) already has a mirrored `reward_penalties:` block. Per the `config-conventions` skill, YAML is the source of truth for defaults and exemplar YAML must be updated when a new field lands.

### AI-1
Add a `reward_penalties:` block to [`examples/configs/grpo_math_1B_megatron_single_controller.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/examples/configs/grpo_math_1B_megatron_single_controller.yaml) mirroring the non-SC exemplar shape (all four flags `false`, `token_ids:` commented out with an inline example).

tests/unit/experience/test_rollout_manager.py
```diff
@@ -422,6 +426,98 @@ def test_result_to_completion_drops_mask_flag_when_gate_off():
...
+            },
+            [{"type": "message", "content": [{"text": "answer"}]}],
+            {"has_malformed_thinking": True},
+            None,
+            "malformed_think_tag",
+            "malformed_think_tag_rate",
+        ),
+    ],
+)
+def test_nemo_gym_reward_penalties_match_legacy_rewards_counts_and_metrics(
```
macandro96, line 702: **No action needed — suggestion only.**

All 4 parametrized cases run with `num_results=1`, so the metric math reduces to `1/1 = 1.0`. A future refactor that swapped numerator/denominator or dropped the `/ num_results` division in `_compute_reward_penalty_metrics` at [`rollout_manager.py:1072`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/experience/rollout_manager.py#L1072) wouldn't be caught.

Optional: add one case with `num_results=3` where only 1 sample triggers, asserting `metric_name: 1/3`. Trivial — synthesize a counts dict and call `impl._compute_reward_penalty_metrics(counts, 3)` directly.

macandro96 (commented):
Nice work extending the reward-penalty feature to the SingleController Gym path. Overall the wiring is clean: resolving `reward_penalty_config` once at setup enables the fail-fast test, and the batched `_results_to_completions` refactor keeps the singleton hook. Tests for the new SC surface pass locally (16 + 35).

**Findings from the parallel review team (bug-finder, rl-expert, test-agent, design-reviewer, devil-advocate):**
- 1 pre-existing test regression that will break CI (inline).
- 3 DRY / config-conventions suggestions (inline).
- 1 test-coverage suggestion (inline).
- Body: the PR description is the unfilled auto-template — please add a one-line summary and link the non-SC parity motivation (existing feature at [`grpo.py:2960`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/algorithms/grpo.py#L2960) / [`:3893`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/algorithms/grpo.py#L3893) / [`trajectory_collector.py:936`](https://github.com/NVIDIA-NeMo/RL/blob/5d8707afe7cb6a851cabf01e267340379340f59e/nemo_rl/algorithms/async_utils/trajectory_collector.py#L936)).

Pre-commit passed clean. Local pytest runs:
- `tests/unit/experience/test_rollout_manager.py`: 16 passed
- `tests/unit/single_controller/test_single_controller_setup.py`: 35 passed
- `tests/unit/experience/test_rollouts.py::test_rollout_manager_consumes_stream_and_restores_input_order`: **FAILED** (see inline)

_Generated by Claude Code_

tests/unit/experience/test_rollout_manager.py
macandro96: addressed in [bd60fb7](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/bd60fb71fc80174907211ec9a6f4e26972759748)

nemo_rl/algorithms/single_controller_utils/config.py
macandro96: addressed in [bd60fb7](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/bd60fb71fc80174907211ec9a6f4e26972759748)

nemo_rl/experience/rollout_manager.py
macandro96: addressed in [bd60fb7](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/bd60fb71fc80174907211ec9a6f4e26972759748)

nemo_rl/algorithms/single_controller_utils/config.py
macandro96: addressed in [bd60fb7](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/bd60fb71fc80174907211ec9a6f4e26972759748)

nemo_rl/experience/rollout_manager.py
macandro96: addressed in [bd60fb7](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/bd60fb71fc80174907211ec9a6f4e26972759748)

nemo_rl/experience/rollout_manager.py
```diff
@@ -980,47 +984,75 @@ async def _run_rollouts(
...
+                self._compute_reward_penalty_metrics(
+                    penalty_counts, len(completed_results)
+                )
+            )
 
         rollout_metrics.update(env_timing_metrics)
 
         return completions, prompt_message_log, rollout_metrics
 
     def _result_to_completion(self, result: dict) -> Completion:
```
yuki-97: `_result_to_completion` lost its only production caller in this diff — `_run_rollouts` now calls `_results_to_completions` directly. It's referenced only by the two tests at test_rollout_manager.py:419,424.

**Suggest fix:** delete it and have those two tests call `_results_to_completions([result])[0][0]` directly, mirroring how test_rollouts.py was updated in this same PR to mock `_results_to_completions` instead of `_result_to_completion`.

yuki-97: `_result_to_completion` lost its only production caller in this diff — `_run_rollouts` now calls `_results_to_completions` directly. It's referenced only by the two tests at test_rollout_manager.py:419,424.

**Suggest fix:** delete it and have those two tests call `_results_to_completions([result])[0][0]` directly, mirroring how test_rollouts.py was updated in this same PR to mock `_results_to_completions` instead of `_result_to_completion`.

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -542,6 +547,7 @@ class MasterConfig(BaseModel, extra="allow"):
...
```
yuki-97, line 594: main has since added PPO support to SingleController (#3773, landed d938da394) — [MasterConfig here](https://github.com/NVIDIA-NeMo/RL/blob/4f653c4dce903598f659f17969c654a050b3919f/nemo_rl/algorithms/single_controller_utils/config.py#L541-L556) is now `grpo: Optional[GRPOConfig] = None` / `ppo: Optional[PPOConfig] = None`, not the required `grpo: GRPOConfig` this PR is based on.

After rebasing, `reward_penalties` needs to land on that new shape, and [examples/configs/ppo_math_1B_megatron_single_controller.yaml](https://github.com/NVIDIA-NeMo/RL/blob/4f653c4dce903598f659f17969c654a050b3919f/examples/configs/ppo_math_1B_megatron_single_controller.yaml) (also added in #3773) should probably get the same `reward_penalties:` block for parity — NeMo-Gym rollouts apply to the PPO SC path too, not just GRPO.

yuki-97 (commented):
Traced the reward-penalty wiring for SC end-to-end (RewardPenaltyConfig -> validate_single_controller_config -> resolve_reward_penalty_config -> RolloutManager -> apply_reward_penalties -> compute_reward_penalty_metrics) — logic is correct and the prior self-review threads all genuinely landed. Two minor items below: a dead private wrapper left over from the streaming refactor, and a heads-up that main has since added PPO to SingleController, so this needs a small parity follow-up after rebase.

nemo_rl/experience/rollout_manager.py
macandro96: addressed in [2fe8a47](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/2fe8a47d68a3b3d98fdb72afc939f4be360810fc)

nemo_rl/algorithms/single_controller_utils/config.py
macandro96: addressed in [2fe8a47](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/2fe8a47d68a3b3d98fdb72afc939f4be360810fc)

macandro96:
/ok to test [2fe8a47](https://github.com/NVIDIA-NeMo/RL/commit/2fe8a47d68a3b3d98fdb72afc939f4be360810fc)

nemo_rl/experience/rollout_manager.py
```diff
@@ -1030,36 +1035,59 @@ async def _run_rollouts(
...
-        truncated = (
-            sum(len(m["token_ids"]) for m in result["message_log"]) == self._max_seq_len
-        )
-
-        # Same gate as the batched path: when masking is off, drop the env
-        # mask flag so later batch building never sees it.
-        if not self._mask_env_flagged_samples:
-            (result["full_result"].get("instance_config") or {}).pop(
-                "mask_sample", None
+    def _results_to_completions(
```
yuki-97, line 1048: **[nit]** `_result_to_completion` is gone, but one dangling reference to it survives: the wiring comment at [test_effort_shaping.py:208](https://github.com/NVIDIA-NeMo/RL/blob/2fe8a47d68a3b3d98fdb72afc939f4be360810fc/tests/unit/test_effort_shaping.py#L208) still says shaping "must land before `_result_to_completion`".

The ordering it documents is still correct — [`_apply_effort_shaping`](https://github.com/NVIDIA-NeMo/RL/blob/2fe8a47d68a3b3d98fdb72afc939f4be360810fc/nemo_rl/experience/rollout_manager.py#L1019) runs before the call at [:1027](https://github.com/NVIDIA-NeMo/RL/blob/2fe8a47d68a3b3d98fdb72afc939f4be360810fc/nemo_rl/experience/rollout_manager.py#L1027) — only the name is stale.

**Suggest fix:** update that comment to name `_results_to_completions`.

yuki-97 (commented):
Re-reviewed at 2fe8a47d68a3b3d98fdb72afc939f4be360810fc. Both open threads check out against the current head:

- [Dead `_result_to_completion` wrapper](https://github.com/NVIDIA-NeMo/RL/pull/3741#discussion_r3879672322) — deleted, and both mask-gate tests now call `_results_to_completions([result])[0][0]`. A repo-wide grep at head finds no definition and no caller left.
- [PPO parity after the rebase](https://github.com/NVIDIA-NeMo/RL/pull/3741#discussion_r3879672332) — `reward_penalties` sits on the new `grpo`/`ppo`-Optional MasterConfig, the PPO SC exemplar carries the block, and neither `validate_single_controller_config` nor `resolve_reward_penalty_config` reads `master_config.grpo`, so the PPO SC path genuinely gets the penalties rather than silently dropping them.

One nit below, nothing blocking.

`L0_Unit_Tests_Other` is red at this head, needs to be fixed before merge.

macandro96:
/ok to test [5e48ef6](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/5e48ef683550fe49423a22417e85024cf661dc21)

nemo_rl/experience/rollout_manager.py
macandro96: addressed in [5e48ef6](https://github.com/NVIDIA-NeMo/RL/pull/3741/commits/5e48ef683550fe49423a22417e85024cf661dc21)

yuki-97 (approved):

PR #2443, opened 2026-05-08
fix: fix skip_reference_policy_logprobs_calculation and skip_prev_logprobs

jinglinglingling:
## Related issues

- Fixes #1968
- Closes #1957
- Part of #1906

## Summary

Consolidates three open PRs onto current `main` and addresses review
feedback so they can ship together cleanly:

- **#2177** — skip `prev_logprobs` computation when `force_on_policy_ratio=True`
- **#2174** — guard the GRPO training loop against calling `policy.get_reference_policy_logprobs()` when the reference model was never loaded (issue #1968 Bug 1)
- **#2178** — derive `init_reference_model = (kl_penalty > 0)` to skip loading the reference model when KL penalty is zero

## Changes on top of those three PRs

- Move the `skip_reference_policy_logprobs_calculation` assert into `setup()` so misconfiguration fails before any GPU work (per @yfw on #2177).

- Auto-enable `grpo.skip_reference_policy_logprobs_calculation=True` when `loss_fn.reference_policy_kl_penalty == 0`, so existing recipes that have `kl_penalty=0` without explicitly setting the skip flag (e.g. `examples/configs/recipes/llm/grpo-deepscaler-1.5b-8K.yaml`) stop crashing inside `use_reference_model()` (per @yuki-97 / @terrykong on #2178).

- Add two parametrized unit tests in `tests/unit/algorithms/test_grpo.py` covering both `grpo_train` and `async_grpo_train`:
  - `test_grpo_train_skips_reference_policy_logprobs_when_configured` guards #1968 / #2174 / #2178.
  - `test_grpo_train_skips_prev_logprobs_when_force_on_policy_ratio` guards #2177.

- Drop the worker-level guards from #2174 — the grpo.py-level skip already prevents the bad call paths, so the worker-layer fallbacks are dead code.

PR #3742, opened 2026-08-21
fix(sc): gate reference logprobs on KL penalty

jinglinglingling:
## Summary

This PR forward-ports the Zero-KL reference-logprob gating introduced in [#2443](https://github.com/NVIDIA-NeMo/RL/pull/2443) to NeMo RL v2 (single-controller async).

- gate v2 SingleController reference-logprob requests on both a positive `reference_policy_kl_penalty` and the explicit skip flag
- fetch only the policy/reference logprob data-plane columns produced for each train step
- reject non-finite/negative KL coefficients and reward-KL configurations that would omit required policy logprobs
- cover initialization, controller calls, train-field propagation, and invalid configuration combinations

This forward-ports the Zero-KL reference-logprob gating behavior from #2443 to the current v2 single-controller async path.

## Test plan
- [x] affected SingleController and setup suites: 56 passed
- [x] broader affected CPU suites (controller, setup, checkpointing, split API): 102 passed
- [x] targeted invalid KL coefficient and data-plane field tests passed
- [x] `pre-commit run --all-files`

GPU/runtime qualification was not run locally because this environment does not expose a GPU.

### DIFF
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -206,6 +206,8 @@ class SingleControllerActor:
             master_config.loss_fn.force_on_policy_ratio
             and self._algo_cfg.seq_logprob_error_threshold is None
         )
+        # _build_trainer initializes the reference model only for a positive KL
+        # penalty, so the controller must use the same gate before requesting it.
         self._reference_logprobs_required = bool(
             master_config.loss_fn.reference_policy_kl_penalty > 0
             and not self._algo_cfg.skip_reference_policy_logprobs_calculation
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -101,7 +101,7 @@ from nemo_rl.algorithms.single_controller_utils.utils import (
 from nemo_rl.data.interfaces import DatumSpec
 from nemo_rl.data_plane import DATA_PLANE_CHECKPOINT_SCHEMA_VERSION, KVBatchMeta
 from nemo_rl.data_plane.async_utils import call_data_plane
-from nemo_rl.data_plane.schema import DP_CALIB_INPUT_FIELDS
+from nemo_rl.data_plane.schema import DP_CALIB_INPUT_FIELDS, DP_TRAIN_FIELDS
 from nemo_rl.distributed.batched_data_dict import BatchedDataDict
 from nemo_rl.distributed.refit_watchdog import RefitAborted, is_refit_context_lost
 from nemo_rl.environments.nemo_gym import should_use_nemo_gym
@@ -147,6 +147,20 @@ def _pooled_opd_metrics(
         "on_policy_distillation/adv_mean": mean,
         "on_policy_distillation/adv_std": math.sqrt(max(variance, 0.0)),
     }
+
+
+def _train_fields_for_step(
+    *,
+    policy_logprobs_required: bool,
+    reference_logprobs_required: bool,
+) -> tuple[str, ...]:
+    """Return only the data-plane columns produced for this train step."""
+    return tuple(
+        field
+        for field in DP_TRAIN_FIELDS
+        if (policy_logprobs_required or field != "prev_logprobs")
+        and (reference_logprobs_required or field != "reference_policy_logprobs")
+    )
 
 
 @ray.remote(num_cpus=1, num_gpus=0)  # pragma: no cover
@@ -213,6 +227,10 @@ class SingleControllerActor:
             and not self._algo_cfg.skip_reference_policy_logprobs_calculation
         )
         self._teacher_logprobs_required = opd_module.is_opd_enabled(master_config)
+        self._train_fields = _train_fields_for_step(
+            policy_logprobs_required=self._policy_logprobs_required,
+            reference_logprobs_required=self._reference_logprobs_required,
+        )
         self._dp_client = actor_args.dp_client
         self._gen: Generation = actor_args.gen_handle
         self._trainer: TQPolicy = actor_args.trainer_handle
@@ -1876,6 +1894,7 @@ class SingleControllerActor:
                                     await asyncio.to_thread(
                                         self._trainer.train_microbatches_from_meta,
                                         train_meta,
+                                        train_fields=self._train_fields,
                                     )
                                     # A PPO step is one chunk: nothing to
                                     # accumulate, so close every epoch here.
--- a/nemo_rl/algorithms/single_controller_utils/config.py
+++ b/nemo_rl/algorithms/single_controller_utils/config.py
@@ -1020,6 +1020,21 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
             "loss_fn.reference_policy_kl_penalty=0."
         )
 
+    if (
+        master_config.loss_fn.use_kl_in_reward
+        and reference_policy_kl_penalty > 0
+        and master_config.loss_fn.force_on_policy_ratio
+        and algo_cfg.seq_logprob_error_threshold is None
+    ):
+        raise ValueError(
+            "loss_fn.use_kl_in_reward=true with a nonzero "
+            "loss_fn.reference_policy_kl_penalty requires policy logprobs, but "
+            "loss_fn.force_on_policy_ratio=true without "
+            "seq_logprob_error_threshold skips them. Set "
+            "loss_fn.force_on_policy_ratio=false or configure "
+            "seq_logprob_error_threshold."
+        )
+
     # ``env`` is required in production configs, but model_construct-based unit
     # configs can omit it. Only apply rollout-path validation when it is present.
     env_config = getattr(master_config, "env", None)
--- a/nemo_rl/algorithms/loss/loss_functions.py
+++ b/nemo_rl/algorithms/loss/loss_functions.py
@@ -15,7 +15,7 @@
 from typing import Any, NotRequired, Optional, TypedDict, TypeVar
 
 import torch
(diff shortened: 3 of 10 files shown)


jinglinglingling (commented):
## Self-review

**No action needed — all high-confidence findings from the self-review were resolved and re-reviewed.**

### Scope

Reviewed the complete v2 Zero-KL reference-logprob gating change with parallel RL-correctness, bug, test, design, and comment reviewers, followed by an independent devil's-advocate pass.

### Findings resolved

- Fixed the zero-KL producer/consumer mismatch: the trainer now fetches only logprob columns produced for the current step.
- Added direct train-pump coverage for all four policy/reference logprob gate combinations.
- Constrained `reference_policy_kl_penalty` to finite non-negative values.
- Added setup-time rejection for reward-KL configurations that would skip required policy logprobs.

All fixes landed in e06a1b7342855ebeb7d36139853121477bd27533 and the final adversarial pass returned LGTM with no remaining findings at or above the 80% confidence threshold.

### Validation

- 56 final SingleController/setup tests passed.
- 102 broader affected CPU tests passed across controller, setup, checkpointing, and split-policy APIs.
- `pre-commit run --all-files` passed, including Ruff, formatting, and Pyrefly.
- PR is conflict-free.

### Limitation

GPU/runtime qualification was not run locally because this environment does not expose a GPU. Full NVIDIA-runner validation and reviewer approval remain pending.

_Generated by Cursor Agent using the upstream `/review-pr-team` workflow._

nemo_rl/algorithms/single_controller.py
```diff
@@ -127,8 +141,15 @@ def __init__(
             master_config.loss_fn.force_on_policy_ratio
             and master_config.grpo.seq_logprob_error_threshold is None
         )
-        self._reference_logprobs_required = not bool(
-            master_config.grpo.skip_reference_policy_logprobs_calculation
+        # _build_trainer initializes the reference model only for a positive KL
+        # penalty, so the controller must use the same gate before requesting it.
+        self._reference_logprobs_required = bool(
```
yfw, line 225: [`single_controller.py:146`](https://github.com/NVIDIA-NeMo/RL/blob/e06a1b734f9db1503d93b73eb00965d472cb79ae/nemo_rl/algorithms/single_controller.py#L146)

**1 action item.**

**TL;DR — this predicate is already on `main` verbatim; the thing to watch is the conflict resolution, because keeping your side reintroduces `master_config.grpo`, which is `None` on an SC PPO run.**

[aa739509c](https://github.com/NVIDIA-NeMo/RL/commit/aa739509ce3772a18a4d02da651b86c9e898e8cb) (#3771, merged 2026-08-24) landed exactly this change — the code is identical, this hunk just adds the explanatory comment above it. Since then #3773 moved the algorithm block behind `algo_config()`, so the line on `main` reads `... and not self._algo_cfg.skip_reference_policy_logprobs_calculation`.

Simulating the rebase with `git merge-file` puts the two versions in a single conflict block, byte-identical apart from `master_config.grpo` vs `self._algo_cfg` — and **your side looks like the superset**, because it carries the `self._train_fields = ...` lines that `main` doesn't have. Resolving by keeping it restores the `master_config.grpo` read. `MasterConfig.grpo` is `Optional[GRPOConfig] = None` and `validate_algorithm_block` enforces that exactly one of `grpo`/`ppo` is set, so on a PPO run that becomes:

```
AttributeError: 'NoneType' object has no attribute 'skip_reference_policy_logprobs_calculation'
```

raised in `SingleControllerActor.__init__` before training starts.

**Action:** keep `main`'s line unchanged and append only your `self._train_fields = _train_fields_for_step(...)` assignment, feeding it from `self._algo_cfg` rather than `master_config.grpo`.

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -758,6 +758,21 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
             "loss_fn.reference_policy_kl_penalty=0."
         )
 
+    if (
+        master_config.loss_fn.use_kl_in_reward
+        and reference_policy_kl_penalty > 0
+        and master_config.loss_fn.force_on_policy_ratio
+        and master_config.grpo.seq_logprob_error_threshold is None
```
yfw: [`config.py:765`](https://github.com/NVIDIA-NeMo/RL/blob/e06a1b734f9db1503d93b73eb00965d472cb79ae/nemo_rl/algorithms/single_controller_utils/config.py#L765)

**1 action item — after the rebase.**

Same root cause as the `__init__` comment: `master_config.grpo` is `None` on an SC PPO run, so this dereference raises `AttributeError: 'NoneType' object has no attribute 'seq_logprob_error_threshold'` instead of the `ValueError` you intend. Short-circuiting keeps it narrow — it needs `use_kl_in_reward and kl > 0 and force_on_policy_ratio` to all hold first — but that is precisely the misconfiguration this guard exists to report, so the user gets a traceback instead of your (good) remediation message.

**Action:** post-rebase, read the active algorithm block instead:

```python
and algo_cfg.seq_logprob_error_threshold is None
```

`algo_cfg = algo_config(master_config)` is already bound near the top of `validate_single_controller_config` on `main`, and the sibling guard just above yours already uses `algo_cfg.skip_reference_policy_logprobs_calculation`. `PPOConfig` carries `seq_logprob_error_threshold` too, so it's valid on both branches. Worth dropping the `grpo.` prefix from the message text as well — `main` already did that for the neighbouring guard.

I've deliberately not attached this as a committable `suggestion`: `algo_cfg` doesn't exist in this branch yet, so applying it here would `NameError`. It's only correct once you're on top of `main`.

The guard itself is a good addition — see the review summary for why SC genuinely lacks this check.

tests/unit/single_controller/test_single_controller.py
```diff
@@ -401,6 +510,10 @@ def _train_pump_controller(*, sampler) -> object:
     ctrl._advantage_cfg = AdvantageConfig()
     ctrl._policy_logprobs_required = False
     ctrl._reference_logprobs_required = False
+    ctrl._train_fields = single_controller._train_fields_for_step(
```
yfw: [`test_single_controller.py:513`](https://github.com/NVIDIA-NeMo/RL/blob/e06a1b734f9db1503d93b73eb00965d472cb79ae/tests/unit/single_controller/test_single_controller.py#L513)

**1 action item.**

**TL;DR — all three `single_controller` test files this PR edits were renamed on `main`, and this four-line hunk is the one that's easiest to lose in the move and the most expensive to lose.**

Renames since your base: `test_single_controller.py` → `test_single_controller_actor.py`, `test_sc_checkpointing.py` → `test_checkpointing.py`, `test_single_controller_setup.py` → `test_setup.py`.

**Why this hunk specifically matters.** On `main`, `_train_pump_controller` builds the actor with `object.__new__(controller_cls)` — `__init__` never runs — and hand-sets the flags (`_policy_logprobs_required`, `_reference_logprobs_required`, `_teacher_logprobs_required`) but not `_train_fields`. Once `_train_pump` reads `train_fields=self._train_fields`, every test built through that fixture raises `AttributeError: 'SingleControllerActor' object has no attribute '_train_fields'`. That's **17 call sites**, including the whole `_ppo_train_pump_controller` family. Your diff already adds the fix here; it just has to survive the rename.

Checklist for the `train_fields` kwarg on the renamed `test_single_controller_actor.py`:

| site | status |
|---|---|
| `_train_pump_controller` — add `_train_fields` (this hunk) | yours, must survive the move |
| `_NoOpTrainer.train_microbatches_from_meta` | covered by this PR |
| `_EpochRecordingTrainer.train_microbatches_from_meta` | **not covered** — a third fake, added on `main` after your base. Update rather than delete; its `calls` list is asserted elsewhere |
| `trainer.train_microbatches_from_meta.assert_called_once_with(valid_meta)` | **not covered** — the trainer is `MagicMock(spec=_NoOpTrainer)`, so this needs `assert_called_once_with(valid_meta, train_fields=ctrl._train_fields)` |

`test_checkpointing.py`'s fake is covered by your `test_sc_checkpointing.py` hunk, and the two new `test_setup.py` tests port over verbatim — `_make_master_config(loss_cfg=...)` and `_build_trainer`'s signature are both unchanged on `main`.

One dedup while you're in here: `aa739509c` added `test_reference_logprobs_required_only_when_kl_enabled` to this same file, with an identical fixture body and the same first three parametrize rows. Only your 4th row and the `_train_fields` assertions are new — worth folding those into the existing test rather than shipping both.

nemo_rl/algorithms/single_controller.py
jinglinglingling: Resolved in ff0df3f0c: kept main’s `self._algo_cfg.skip_reference_policy_logprobs_calculation` predicate and added only the `_train_fields` initialization.

nemo_rl/algorithms/single_controller_utils/config.py
jinglinglingling: Resolved in ff0df3f0c: the guard now uses `algo_cfg.seq_logprob_error_threshold`, and the remediation text is algorithm-neutral.

tests/unit/single_controller/test_single_controller.py
jinglinglingling: Resolved in ff0df3f0c: migrated the changes to the renamed test files, initialized `_train_fields` in the fixture, updated `_EpochRecordingTrainer` and the mock assertion, and folded the new cases into the existing test. The affected suite passes (339 passed, 44 GPU-only skipped).

yfw (approved):

PR #3744, opened 2026-08-21
fix(vllm): harden speculative draft refit lifecycle

seonjinn:
# What does this PR do ?

Hardens the vLLM speculative-draft refit lifecycle.

- Resolves Eagle3, DFlash, and DSpark capabilities through one typed interface.
- Verifies exact target/draft weight coverage before completing refit.
- Fails closed on partial transfer, bulk receive, or FP8 KV-cache post-processing failures.
- Releases IPC resources deterministically and preserves ordered draft manifests.

## Supported contract

- No configuration change is required.
- DFlash and DSpark generation remain limited to CP1 and PP1.
- Trainer-side CP transport and packed co-training are handled by #3750 and #3757.

# Issues

No issue is closed by this PR.

# Usage

Existing Eagle3, DFlash, and DSpark generation configurations use the common refit lifecycle automatically.

**Pre checks**:

## Prior exact-head GPU evidence

- OCI-HSG `6406866`: PASS (`COMPLETED 0:0`) on the predecessor exact head, covering 11 speculator-runtime, 37 vLLM refit-lifecycle, and 18 policy IPC/lifecycle tests.

# Validation / Remaining

- Exact head: `8a03385f2320ed3dd1677a30a52856f54556da3f`
- All completed GitHub checks pass; the H100 container build is still running.
- This is a correctness change; no throughput improvement is claimed.


seonjinn (commented):
## Contributor self-review (`/review-pr-team` checklist)

Reviewed exact head `ef32181adfa88b6d11426ea3d0fe1b8622e930b7` against current `main` `8863ebeff172fc3e1d14ab6cdbd0af626f477c01`. The branch is mergeable, has one `CI:L1` label, and has no unresolved review threads.

### Findings

- Resolved before posting: FP8 KV-cache post-processing failures occurred outside the fail-closed refit lifecycle guard. A worker that had already applied target weights could therefore report the post-processing error without marking itself unusable, allowing a later refit attempt. Head commit `ef32181a` moves the post-processing step under the lifecycle guard and adds a regression test proving that the next refit is rejected.
- No actionable findings remain in the reviewed diff.

### Verification

- OCI-HSG gate job `6406866`: `COMPLETED (0:0)` on the exact reviewed head, using vLLM 0.25.1 and OpenAI 2.25.0.
- Pyrefly: 0 errors.
- Speculator runtime suite: 11 passed.
- vLLM backend suite: 37 passed.
- Policy IPC/lifecycle suite: 18 passed.
- Local focused regression checks: 11 passed; final lifecycle pair: 2 passed.
- GitHub DCO, semantic-title, secrets, copyright, and submodule checks are green.

### Scope and residual risk

The focused GPU gate validates the vLLM 0.25.1 refit/lifecycle paths changed by this PR. End-to-end training performance/convergence and distributed trainer CP > 1 remain outside this review's validation scope.

Independent reviewer-agent capacity was unavailable because PR9-PR11 work was already running in parallel, so this is a degraded direct full-diff audit rather than an independently adversarial review wave.

_Generated by Codex contributor self-review_

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1185,7 +1321,7 @@ def _recv_one_param(param_info, group, stream):
             # canonical [gate; up] bulk write above defers to here, and the MTP
             # drafter's mirror of the same. The FP8 KV-cache per-layer k/v scales
             # are finalized by the lifecycle on exit.
-            finalize()
+            finalize(False)
```
seonjinn: Supplemental to the earlier self-review, from a lifecycle-checklist pass: the bulk xferdtensor receive here writes directly into live vLLM params on per-stage CUDA streams, but the `_weight_update_lifecycle` guard is entered only afterwards (line ~1306). A mid-stream failure therefore leaves the model partially overwritten without setting `_refit_unusable_reason`, so the next refit attempt proceeds against an inconsistent model -- the exact stale-state window the ipc/collective transports close by doing their weight-mutating work inside the guard. Fixed in the follow-up commit by entering the lifecycle before the bulk receive (the CM has no transport-specific entry behavior, so this only widens the fail-closed window).

nccl_reshard_refit had no outer try/except, so its exceptions propagated raw through `collective_rpc` while ipc/collective uniformly convert non-fatal transport errors to `return False` via `_weight_update_errors_are_fatal()`. No test exercised an nccl_reshard failure path. The follow-up commit mirrors the ipc/collective error contract and adds a fail-closed test: a bulk-receive failure returns False, poisons the worker, and every subsequent attempt keeps returning False.

seonjinn (commented):
Supplemental self-review (the earlier one on this PR stands; this pass ran the PR9 lifecycle checklist -- repeated refit, fail-closed windows, stale-state reuse, error-contract consistency -- specifically against surfaces it did not cover).

Verified clean beyond the two inline findings: `speculator_runtime.py` holds no cross-refit mutable state that survives a failure; `_refit_unusable_reason` has no reset path (poisoning is permanent until worker restart, which is the intended semantics); a second successful refit starts from clean state; and worker restart reinitializes without partial-state carryover.

Note for the stack: #3750 builds on this head and touches the same transport -- it should rebase once the follow-up commit lands.

seonjinn:
CI triage: the L0_Unit_Tests_Vllm_1-3 failures on 8a03385f2 were a pre-existing signature drift this PR's first real CI run surfaced — `VllmQuantInternalWorkerExtension._weight_update_lifecycle`'s inner `finalize()` took zero arguments while the base lifecycle callers pass `finalize(finalize_draft)`, so every real-quant refit test hit `TypeError`. Fixed in the new head: the override now matches the signature and fails loudly if draft finalization is ever requested on the real-quant path (unsupported there).

seonjinn:
Second signature-drift in the same subclass, same root cause as the finalize fix: the base `_load_weights` gained a `coverage` keyword this PR, but `VllmQuantInternalWorkerExtension._load_weights` (and three test stubs) kept the old signature -- every real-quant IPC/sync refit threw `TypeError` and reported 507 missing keys. Fixed on the new head: the override accepts `coverage` and records the full delivered name set (real-quant deliberately filters some scale entries out of the actual load, but they are handled); stubs updated. Audited the subclass for further drift against the base's changed surfaces -- none remaining.

nemo_rl/models/generation/vllm/vllm_backend.py
```diff
@@ -1135,59 +1271,76 @@ def _recv_one_param(param_info, group, stream):
...
+                    )
+                torch.cuda.empty_cache()
 
-            torch.cuda.empty_cache()
+                # Finalize post-load weight processing: dense Linear + attention/MLA,
+                # the per-MoE-backend w13 layout (FlashInfer CUTLASS/TRTLLM) that the
+                # canonical [gate; up] bulk write above defers to here, and the MTP
+                # drafter's mirror of the same. The FP8 KV-cache per-layer k/v scales
+                # are finalized by the lifecycle on exit.
+                finalize(False)
```
yuekaizhang: `nccl_reshard_refit` now goes through `_weight_update_lifecycle` but always calls `finalize(False)`, and `_receive_and_load_misc_params` doesn't pass `coverage`. If a `speculative_config` (eagle3/dflash/dspark) is active, nothing in `nccl_reshard_utils.validate_master_config` rejects this combination, so draft weights could load via the legacy `_draft_runtime_adapter is None` branch without `process_weights_after_loading` ever running on them — a silent, partially-finalized draft model on every refit cycle. Please add an explicit config-validation violation for `speculative_config` being set on this transport (matching the existing `real_quant`/`enable_eplb` pattern), or wire `coverage`/draft finalization through this path the same way `ipc`/`collective` now do.

yuekaizhang (commented):
Reviewed the fail-closed refit lifecycle hardening — the core machinery (`_weight_update_lifecycle`, `ModelUpdateCoverage` exact-input validation, permanent `_refit_unusable_reason` poisoning, `_require_refit_usable`/`_mark_refit_unusable` on `except BaseException`) is solid and well-tested. One real gap found in the "does every failure path actually fail closed" scrutiny, given that's this PR's explicit goal.

nemo_rl/modelopt/models/generation/vllm_quant_backend.py
```diff
@@ -482,7 +482,14 @@ def _weight_update_lifecycle(
...
 
-        def finalize() -> None:
+        def finalize(finalize_draft: bool) -> None:
+            # Match the base lifecycle's finalize(finalize_draft) signature.
+            # Real-quant refits do not support co-trained draft finalization;
+            # fail loudly rather than silently skipping it.
+            if finalize_draft:
+                raise RuntimeError(
+                    "ModelOpt real-quant refit does not support draft finalization"
+                )
```
isomap: With the check moved to `prepare_refit_info()`, this late rejection is redundant and can be removed.

```suggestion
```

nemo_rl/modelopt/models/generation/vllm_quant_backend.py
isomap, line 547: Could we reject real-quant manifests containing draft weights during preparation, before any weights are loaded? Please apply this together with the suggestion removing the late check in finalize().

```suggestion
        if self._is_real_quant_model() and any(
            name.startswith("draft.") for name in state_dict_info
        ):
            raise RuntimeError(
                "ModelOpt real-quant refit does not support draft finalization"
            )
        super().prepare_refit_info(state_dict_info)
```

Could we also add a regression test for rejecting this manifest during preparation?

isomap (commented):
Suggest moving the real-quant/draft compatibility check to preparation so it fails before loading weights. The two suggestions are intended to be applied together.

nemo_rl/models/generation/vllm/vllm_backend.py
seonjinn: Fixed in 3ad5841 by adding the explicit config-validation violation, matching the `real_quant` / `enable_eplb` pattern. `check_nccl_reshard_refit_support` now rejects `policy.generation.vllm_kwargs.speculative_config.method` in `{eagle3, dflash, dspark}` on this transport, with a comment explaining why: `nccl_reshard_refit` finalizes the target only and `_receive_and_load_misc_params` loads without a coverage object, so the drafter would never see `process_weights_after_loading`. MTP is deliberately not covered by the check, since it finalizes through `_mtp_drafter_refit_enabled` rather than the draft manifest. Regression added in `tests/unit/weight_sync/test_nccl_reshard_utils.py`.

nemo_rl/modelopt/models/generation/vllm_quant_backend.py
seonjinn: Removed in 3ad5841. `finalize()` now just `del finalize_draft`s with a comment noting that the flag can only be `True` when the manifest carries draft weights, which `prepare_refit_info` now rejects up front.

nemo_rl/modelopt/models/generation/vllm_quant_backend.py
seonjinn: Applied in 3ad5841, together with the removal of the late check. `prepare_refit_info` now rejects a real-quant manifest containing `draft.`-prefixed entries before `super().prepare_refit_info(...)` runs, so nothing moves before the failure. Regression added in `tests/unit/models/generation/test_vllm_modelopt_real_quant_config.py`.

seonjinn:
The real-quant/draft compatibility check now runs in `prepare_refit_info()` before any weights move, and the redundant late rejection is gone - both replied to in-thread with the regressions added.

Head is merged with latest `main` (`c49d53e2e`), clean merge, lint clean tree-wide, and `/ok to test` re-posted against `82cf47b6d54863876fdc6ec4acccc916b75e76e4`.

PR #3748, opened 2026-08-21
fix(logging): aggregate rollout histograms before logging

yfw:
# What does this PR do?

Rollout metric helpers created W&B `Histogram` objects before async prompt-group aggregation. The aggregator treated them as non-numeric values and produced lists of per-group histograms, resulting in oversized W&B history rows and unsupported list values for other logging backends.

Keep raw observations backend-agnostic, flatten them into one step-level distribution, and route histogram keys through `Logger.log_histogram`. Remove duplicate GRPO logging loops while preserving the existing ISL, OSL, total-length, and per-agent histogram names.

This means:

- W&B continues to receive `generation_metrics/histogram/*`, now as one flattened distribution per step.
- The raw `train/histogram/*` list-valued series are no longer sent through the scalar logging path.
- MLflow no longer receives non-serializable W&B histogram objects. Native MLflow histogram summaries are intentionally deferred to a follow-up PR.
- TensorBoard now records these distributions with its native histogram API.
- Validation distributions, when enabled by a future caller, use `generation_metrics/validation/histogram/*` rather than aliasing training series.

# Issues

None.

# Testing

- [x] `pytest -q tests/unit/algorithms/test_grpo.py tests/unit/utils/test_logger.py tests/unit/experience/test_rollouts.py` (272 passed)
- [x] `ruff check` on changed Python files
- [x] `pyrefly check`

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -89,6 +89,7 @@ from nemo_rl.environments.nemo_gym import should_use_nemo_gym, spinup_nemo_gym_a
 from nemo_rl.experience.interfaces import (
     NEXT_NEMO_GYM_TASK_INDEX_KEY,
 )
+from nemo_rl.experience.metric_utils import is_histogram_metric
 from nemo_rl.experience.rollouts import (
     EffortLevelsConfig,
     attach_initial_nemo_gym_image_payloads,
@@ -4033,7 +4034,7 @@ def aggregate_rollout_metrics(
     """
     aggregated = {}
     for k, v in per_group_metrics.items():
-        if "histogram" in k.split("/"):
+        if is_histogram_metric(k):
             aggregated[k] = [observation for group in v for observation in group]
         elif not isinstance(v[0], (int, float)):
             aggregated[k] = v
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -3686,20 +3686,6 @@ def grpo_train(
                     logger,
                 )
 
-            # Plot ISL/OSL/ISL+OSL histograms to wandb
-            if (
-                master_config.policy["generation"]
-                .get("vllm_cfg", {})
-                .get("async_engine", False)
-            ):
-                for metric_name in metrics.keys():
-                    if metric_name.startswith("histogram/"):
-                        logger.log_histogram(
-                            metrics[metric_name],
-                            total_steps + 1,
-                            f"generation_metrics/{metric_name}",
-                        )
-
             print("\n📊 Training Results:")
 
             print(f"  • Loss: {metrics['loss']:.4f}")
@@ -4032,6 +4018,7 @@ def aggregate_rollout_metrics(
     """Aggregate rollout metrics from multiple trajectory groups.
 
     Different metric types are aggregated according to their semantics:
+    - Histogram observations: flattened into one step-level distribution
     - Metrics ending with "/min" or starting with "min_" (excluding "_rate" suffix): take the minimum
     - Metrics ending with "/max" or starting with "max_" (excluding "_rate" suffix): take the maximum
     - "total_turns": summed
@@ -4046,7 +4033,9 @@ def aggregate_rollout_metrics(
     """
     aggregated = {}
     for k, v in per_group_metrics.items():
-        if not isinstance(v[0], (int, float)):
+        if "histogram" in k.split("/"):
+            aggregated[k] = [observation for group in v for observation in group]
+        elif not isinstance(v[0], (int, float)):
             aggregated[k] = v
         elif k.endswith("/min") or (k.startswith("min_") and not k.endswith("_rate")):
             aggregated[k] = min(v)
@@ -5249,20 +5238,6 @@ def async_grpo_train(
                     ],
                     logger,
                 )
-
-            # Plot ISL/OSL/ISL+OSL histograms to wandb
-            if (
-                master_config.policy["generation"]
-                .get("vllm_cfg", {})
-                .get("async_engine", False)
-            ):
-                for metric_name in metrics.keys():
-                    if metric_name.startswith("histogram/"):
-                        logger.log_histogram(
-                            metrics[metric_name],
-                            step + 1,
-                            f"generation_metrics/{metric_name}",
-                        )
 
             print("\n📊 Training Results:")
             print(f"  • Loss: {metrics['loss']:.4f}")
--- a/nemo_rl/utils/logger.py
+++ b/nemo_rl/utils/logger.py
@@ -1049,8 +1049,28 @@ class Logger(LoggerInterface):
             step_metric: Optional name of a field in metrics to use as step instead
                          of the provided step value (currently only needed for wandb)
         """
+        histogram_metrics = {
+            name: value
+            for name, value in metrics.items()
+            if "histogram" in name.split("/")
+        }
+        metrics_to_log = {
+            name: value
+            for name, value in metrics.items()
+            if name not in histogram_metrics
+        }
+
+        for name, values in histogram_metrics.items():
+            # Preserve the established namespace for per-turn generation
+            # histograms; other distributions inherit the caller's prefix.
+            if name.startswith("histogram/"):
+                histogram_name = f"generation_metrics/{name}"
+            else:
+                histogram_name = f"{prefix}/{name}" if prefix else name
+            self.log_histogram(values, step, histogram_name)
+
         for logger in self.loggers:
-            logger.log_metrics(metrics, step, prefix, step_metric, step_finished)
+            logger.log_metrics(metrics_to_log, step, prefix, step_metric, step_finished)
 
     def log_hyperparams(self, params: Mapping[str, Any]) -> None:
         """Log hyperparameters to all enabled backends.
--- a/nemo_rl/utils/logger.py
+++ b/nemo_rl/utils/logger.py
@@ -42,6 +42,7 @@ from torch.utils.tensorboard import SummaryWriter
 
 from nemo_rl.data.interfaces import LLMMessageLogType
 from nemo_rl.distributed.batched_data_dict import BatchedDataDict
+from nemo_rl.experience.metric_utils import is_histogram_metric
 
 # Flag to track if rich logging has been configured
 _rich_logging_configured = False
@@ -191,7 +192,8 @@ class TensorboardLogger(LoggerInterface):
 
     def log_histogram(self, histogram: list[Any], step: int, name: str) -> None:
         """Log histogram metrics to Tensorboard."""
-        return
+        if len(histogram) > 0:
+            self.writer.add_histogram(name, np.asarray(histogram), step)
 
     def log_hyperparams(self, params: Mapping[str, Any]) -> None:
         """Log hyperparameters to Tensorboard.
@@ -1048,11 +1050,16 @@ class Logger(LoggerInterface):
             prefix: Optional prefix for metric names
             step_metric: Optional name of a field in metrics to use as step instead
                          of the provided step value (currently only needed for wandb)
+
+        Note:
+            Metrics identified by :func:`is_histogram_metric` are routed through
+            :meth:`log_histogram` instead of each backend's ``log_metrics``.
+            ``histogram/*`` keys retain their established
+            ``generation_metrics/`` namespace; other distributions inherit
+            ``prefix``.
         """
         histogram_metrics = {
-            name: value
-            for name, value in metrics.items()
-            if "histogram" in name.split("/")
+            name: value for name, value in metrics.items() if is_histogram_metric(name)
         }
         metrics_to_log = {
             name: value
@@ -1064,7 +1071,11 @@ class Logger(LoggerInterface):
             # Preserve the established namespace for per-turn generation
             # histograms; other distributions inherit the caller's prefix.
             if name.startswith("histogram/"):
-                histogram_name = f"generation_metrics/{name}"
+                histogram_name = (
+                    f"generation_metrics/{name}"
+                    if prefix in ("", None, "train")
+                    else f"generation_metrics/{prefix}/{name}"
+                )
             else:
                 histogram_name = f"{prefix}/{name}" if prefix else name
             self.log_histogram(values, step, histogram_name)
--- a/tests/unit/algorithms/test_grpo.py
+++ b/tests/unit/algorithms/test_grpo.py
@@ -5089,6 +5089,17 @@ class TestAggregateRolloutMetrics:
         result = aggregate_rollout_metrics(metrics)
         assert result["some_list_metric"] == [["a", "b"], ["c", "d"]]
 
+    def test_histogram_observations_are_flattened(self):
+        """Per-group observations become one bounded step-level distribution."""
+        metrics = {
+            "agent/reward/histogram": [[0.1], [0.2, 0.3]],
+            "histogram/gen_tokens_length": [[10, 20], [30]],
+        }
+        result = aggregate_rollout_metrics(metrics)
+
+        assert result["agent/reward/histogram"] == [0.1, 0.2, 0.3]
+        assert result["histogram/gen_tokens_length"] == [10, 20, 30]
+
     def test_mixed_metrics(self):
         """Full integration test with a realistic mix of metric types."""
         metrics = {
--- a/tests/unit/algorithms/test_grpo.py
+++ b/tests/unit/algorithms/test_grpo.py
@@ -5100,6 +5100,34 @@ class TestAggregateRolloutMetrics:
         assert result["agent/reward/histogram"] == [0.1, 0.2, 0.3]
         assert result["histogram/gen_tokens_length"] == [10, 20, 30]
 
+    def test_histogram_substring_keys_still_average(self):
+        """Histogram-like substrings do not identify distributions."""
+        metrics = {
+            "histogram_bucket_count": [8, 10],
+            "reward/histogram_p95": [1.0, 3.0],
+        }
+        result = aggregate_rollout_metrics(metrics)
+
+        assert result["histogram_bucket_count"] == 9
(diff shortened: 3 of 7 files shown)


nemo_rl/utils/logger.py
```diff
@@ -1049,8 +1049,28 @@ def log_metrics(
...
+        }
+
+        for name, values in histogram_metrics.items():
+            # Preserve the established namespace for per-turn generation
+            # histograms; other distributions inherit the caller's prefix.
+            if name.startswith("histogram/"):
+                histogram_name = f"generation_metrics/{name}"
+            else:
+                histogram_name = f"{prefix}/{name}" if prefix else name
+            self.log_histogram(values, step, histogram_name)
```
yfw, line 1081: **1 action item.**

**TL;DR — after this change W&B is the only backend that renders a rollout histogram; TensorBoard, enabled in 85 shipped configs, silently drops every one despite supporting them natively.**

PR-introduced consequence of an adjacent, pre-existing gap. [`log_histogram`](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/utils/logger.py#L124) is declared abstract, but three of four backends implement it as a bare `return` — [TensorBoard:192](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/utils/logger.py#L192), [SwanLab:491](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/utils/logger.py#L491), [MLflow:951](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/utils/logger.py#L951). That was fine while `log_histogram` was an *additional* W&B-flavoured channel; this line makes it the sole route for every rollout distribution.

Measured on the installed mlflow 3.15.1 / wandb 0.28.1:

- **MLflow gains** — pre-PR `*/histogram` held a `wandb.Histogram` object that went straight to `mlflow.log_metrics`, giving `TypeError: Object of type Histogram is not JSON serializable`. That crash is gone. Nice, and worth stating in the PR body.
- **MLflow loses** — `train/histogram/{gen,input,total}_tokens_length/{mean,p50,p90,max}` (12 series) on the *sync* rollout paths, where the value was already a flat list and [`_summarize_list`](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/utils/logger.py#L816) reduced it. The async trainers lose nothing; they were logging nothing there.
- **TensorBoard** — no data change (it only ever printed `Warning: Skipping metric ... (unsupported type: list)`), but that per-step spam is now gone.

### AI-1
Because `aggregate_rollout_metrics` now guarantees a flat numeric list in every trainer, both stubs become ~3 lines and would give these backends *more* than they had before this PR:

```python
# MLflowLogger.log_histogram
stats = _summarize_list(histogram)
if stats:
    mlflow.log_metrics(
        {f"{name}/{k}": v for k, v in stats.items()}, step=step, run_id=self.run_id
    )

# TensorboardLogger.log_histogram
self.writer.add_histogram(name, np.asarray(histogram), step)
```

Not a `suggestion` block because those lines fall outside this PR's diff hunks. Entirely reasonable as a follow-up issue if you'd rather keep this PR tight — flagging it because this PR is what makes the gap load-bearing.

nemo_rl/utils/logger.py
```diff
@@ -1049,8 +1049,28 @@ def log_metrics(
             step_metric: Optional name of a field in metrics to use as step instead
                          of the provided step value (currently only needed for wandb)
         """
```
yfw, line 1060: **1 action item.**

`log_metrics` no longer forwards every key it is handed — histogram-segment keys are diverted to `log_histogram` and never reach the backends' `log_metrics` — but the docstring still reads as a pure fan-out. PR-introduced contract change on a cross-module public interface.

I applied this exact block and re-ran the gates: 269 tests pass, `ruff-format` and `pyrefly` clean.

```suggestion

        Note:
            Metrics whose key has a ``histogram`` path segment are routed to
            :meth:`log_histogram` instead of being forwarded to each backend's
            ``log_metrics``. ``histogram/*`` keys are namespaced under
            ``generation_metrics/``; other distributions inherit ``prefix``.
            Backends with no histogram implementation (TensorBoard, MLflow,
            SwanLab) discard them.
        """
```

nemo_rl/utils/logger.py
```diff
@@ -1049,8 +1049,28 @@ def log_metrics(
...
+        histogram_metrics = {
+            name: value
+            for name, value in metrics.items()
+            if "histogram" in name.split("/")
```
yfw: **1 action item** — paired with the identical edit in `nemo_rl/algorithms/grpo.py:4036` (separate comment). Applying only one lets the two predicates drift apart, which is exactly the class of bug this PR fixes.

**TL;DR — `"histogram" in name.split("/")` matches a `histogram` segment anywhere, so a Gym env supplying a field named `histogram` turns `<agent>/histogram/mean` (a float) into a flatten target and raises `TypeError: 'float' object is not iterable`.**

PR-introduced narrowing: pre-PR that key fell to the final `else` in `aggregate_rollout_metrics` → `sum(v)/len(v)` and aggregated fine.

Per-agent metric names are built from **env-controlled** keys — [`for key in agent_extras[0].keys()`](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/experience/rollout_manager.py#L1089) feeding `calculate_single_metric(values, n, f"{agent_name}/{key}")` — so an env emitting a numeric `env_extras["histogram"]` produces `<agent>/histogram/{mean,max,min,median,stddev}`, all floats.

**I could not construct this with any in-tree env** (`grep -rn 'env_extras\[' nemo_rl/` has no `histogram` hit), so this is a robustness gap rather than a live bug — but the trigger is third-party env data, not in-repo code.

The tighter predicate matches every real producer (`{key}/histogram`, `histogram/{gen,input,total}_tokens_length`) and still correctly flattens the pathological `<agent>/histogram/histogram`:

```suggestion
            if name.startswith("histogram/") or name.endswith("/histogram")
```

Verified: 269 tests pass, `pyrefly` clean.

nemo_rl/algorithms/grpo.py
```diff
@@ -4046,7 +4033,9 @@ def aggregate_rollout_metrics(
     """
     aggregated = {}
     for k, v in per_group_metrics.items():
-        if not isinstance(v[0], (int, float)):
+        if "histogram" in k.split("/"):
```
yfw: **1 action item** — the paired half of the `nemo_rl/utils/logger.py:1055` comment. Both predicates must describe the same key set; if they drift, the aggregator flattens a key the logger then forwards as a raw list, which is the original bug.

Same reasoning as the other half: a float-valued `<agent>/histogram/mean` reaches this branch and raises `TypeError: 'float' object is not iterable`, where pre-PR it averaged correctly.

```suggestion
        if k.startswith("histogram/") or k.endswith("/histogram"):
```

Verified: 269 tests pass, `pyrefly` clean.

Optional, and deliberately *not* offered as a suggestion because it needs an import outside this diff: the two predicates could share one `is_histogram_key(name)` helper so they cannot drift at all.

nemo_rl/utils/logger.py
```diff
@@ -1049,8 +1049,28 @@ def log_metrics(
...
+            name: value
+            for name, value in metrics.items()
+            if name not in histogram_metrics
+        }
+
+        for name, values in histogram_metrics.items():
+            # Preserve the established namespace for per-turn generation
+            # histograms; other distributions inherit the caller's prefix.
+            if name.startswith("histogram/"):
+                histogram_name = f"generation_metrics/{name}"
```
yfw: **1 action item** — low priority; not reachable today.

`prefix` is silently discarded for `histogram/*` keys, so every caller writes to the same `generation_metrics/histogram/*` series regardless of context. Keeping the established train name is right — I am not asking for a rename.

The reason it is worth one line anyway: **three** `validate()` functions independently compute a `gen_metrics` that contains `histogram/*` and then throw it away — [grpo](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/algorithms/grpo.py#L3892), [ppo](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/algorithms/ppo.py#L3070), [distillation](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/algorithms/distillation.py#L1238). Any one of three one-line "let's also report validation ISL/OSL" edits would silently alias validation onto the train series at the same step, with no error. I confirmed no current call site sends a `histogram/*` key with a non-train prefix, so this is prevention, not a bug report.

One way to do it — of several; this keeps the existing train key byte-identical and rejects nothing legitimate:

```suggestion
            if name.startswith("histogram/"):
                histogram_name = (
                    f"generation_metrics/{name}"
                    if prefix in ("", None, "train")
                    else f"generation_metrics/{prefix}/{name}"
                )
```

Verified: 269 tests pass, `ruff-format` and `pyrefly` clean.

tests/unit/algorithms/test_grpo.py
```diff
@@ -5089,6 +5089,17 @@ def test_non_numeric_passed_through(self):
...
+    def test_histogram_observations_are_flattened(self):
+        """Per-group observations become one bounded step-level distribution."""
+        metrics = {
+            "agent/reward/histogram": [[0.1], [0.2, 0.3]],
+            "histogram/gen_tokens_length": [[10, 20], [30]],
+        }
+        result = aggregate_rollout_metrics(metrics)
+
+        assert result["agent/reward/histogram"] == [0.1, 0.2, 0.3]
+        assert result["histogram/gen_tokens_length"] == [10, 20, 30]
```
yfw, line 5101: **1 action item** — optional test hardening.

The new test pins the happy path, but nothing asserts the *precision* this PR introduced: `k.split("/")` rather than a plain `"histogram" in k` substring test. These three cases lock that in, and the middle one is the regression guard for the env-controlled-key crash raised on `nemo_rl/algorithms/grpo.py:4036`.

I ran these against the tightened predicate from that comment: **269 passed**, `ruff-format` clean. Note the middle test only passes *with* that tightening — it is the paired test for it.

```suggestion
        assert result["histogram/gen_tokens_length"] == [10, 20, 30]

    def test_histogram_substring_keys_still_average(self):
        """Only a whole "histogram" path segment flattens; substrings average."""
        metrics = {
            "histogram_bucket_count": [8, 10],
            "reward/histogram_p95": [1.0, 3.0],
        }
        result = aggregate_rollout_metrics(metrics)

        assert result["histogram_bucket_count"] == 9
        assert result["reward/histogram_p95"] == 2.0

    def test_per_agent_histogram_stats_are_not_flattened(self):
        """An env field literally named "histogram" yields scalar stat keys."""
        metrics = {
            "myagent/histogram/mean": [1.0, 2.0],
            "myagent/histogram/histogram": [[1.0], [2.0]],
        }
        result = aggregate_rollout_metrics(metrics)

        assert result["myagent/histogram/mean"] == 1.5
        assert result["myagent/histogram/histogram"] == [1.0, 2.0]

    def test_empty_histogram_groups_flatten_to_empty(self):
        """A step where every group produced no observations stays loggable."""
        result = aggregate_rollout_metrics({"histogram/gen_tokens_length": [[], []]})

        assert result["histogram/gen_tokens_length"] == []
```

yfw (commented):
Reviewed by a team of five agents plus an adversarial pass. **The fix is correct and necessary — no blockers.** Everything below is a suggestion.

Pre-PR, [`aggregate_rollout_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/algorithms/grpo.py#L4015) hit `not isinstance(v[0], (int, float))` on a `list[wandb.Histogram]` and passed it straight through, so every async trainer wrote a list of Histogram objects into a single W&B history row — the "oversized rows" in the description are real. It also crashed MLflow (`TypeError: Object of type Histogram is not JSON serializable`, measured on mlflow 3.15.1) and printed a per-step TensorBoard warning for every histogram key. This PR fixes all three, and moving `wandb` out of [`metric_utils.py`](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/experience/metric_utils.py#L21) also drops a hard `wandb` import from a module that loads inside every rollout actor.

**One win the description undersells.** The three deleted loops were gated on `policy.generation.vllm_cfg.async_engine`, but these keys are produced by the *async rollout* path, and [`should_use_async_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/1d7194ab6bb75c8bcdbc4e1347f6883518c7fbb6/nemo_rl/models/generation/interfaces.py#L229) also selects it for `megatron`, `sglang`, `dynamo` and `trtllm` — none of which set `vllm_cfg.async_engine`. So the old gate was wrong for four of five async backends, and shipped recipes like `grpo-llama3.2-1b-instruct-1n8g-megatron_generation.yaml` (`backend: megatron`) get these histograms for the first time.

**Worth a line in the description before this leaves draft**, since both are user-visible W&B changes: `generation_metrics/histogram/*` becomes unconditional (new series for sync/megatron/sglang runs), and in the async-engine case the raw-list key `train/histogram/gen_tokens_length` disappears from history entirely, leaving only `generation_metrics/...`.

Gates run locally on a clean tree at this SHA: `pre-commit run --all-files` passes all 8 hooks (ruff, ruff-format, taplo, pyrefly); `tests/unit/{algorithms/test_grpo,utils/test_logger,experience/test_rollouts}.py` → 269 passed. Every `suggestion` block below was applied verbatim and re-verified against those same gates, then reverted.

_Generated by Claude Code_

nemo_rl/utils/logger.py
yfw: Implemented native TensorBoard histogram logging in b11617b2b. Per the PR scope, MLflow histogram summaries will follow in a separate PR; the PR description now documents that deferral and the MLflow serialization fix.

tests/unit/algorithms/test_grpo.py
yfw: Added the requested regression coverage in b11617b2b, including substring precision, an environment field named histogram, empty distributions, validation namespacing, and TensorBoard logging.

nemo_rl/utils/logger.py
yfw: Addressed in b11617b2b with a shared is_histogram_metric helper, restricted to histogram/* and */histogram keys, so environment fields named histogram retain scalar aggregation.

nemo_rl/utils/logger.py
yfw: Addressed in b11617b2b. The Logger.log_metrics docstring now documents histogram routing and namespace behavior.

nemo_rl/utils/logger.py
yfw: Addressed in b11617b2b. Training keeps the existing generation_metrics/histogram/* names, while non-training prefixes are preserved.

nemo_rl/algorithms/grpo.py
yfw: Addressed in b11617b2b. Both aggregation and logging now use the same shared is_histogram_metric predicate.

nemo_rl/utils/logger.py
yfw: fixed in b11617b2be84ad93068555ea539ecc19e480858b

nemo_rl/utils/logger.py
yfw: fixed in b11617b2be84ad93068555ea539ecc19e480858b

nemo_rl/algorithms/grpo.py
yfw: fixed in b11617b2be84ad93068555ea539ecc19e480858b

nemo_rl/utils/logger.py
yfw: fixed in b11617b2be84ad93068555ea539ecc19e480858b

tests/unit/algorithms/test_grpo.py
yfw: fixed in b11617b2be84ad93068555ea539ecc19e480858b

nemo_rl/utils/logger.py
yfw: partially fixed in b11617b2be84ad93068555ea539ecc19e480858b — TensorBoard is done, MLflow and SwanLab are not.

**Addressed:** `TensorboardLogger.log_histogram` now calls `writer.add_histogram` with an empty-input guard, plus two tests. That was the high-value half — TensorBoard is enabled in 85 shipped configs. I probed the new implementation against the degenerate inputs real rollouts produce (all-identical rewards, a single observation, int token lengths, `nan`, `inf`): all succeed, so the missing `try/except` that `WandbLogger.log_histogram` carries isn't needed here.

**Still remaining, low priority — no action needed:** `MLflowLogger.log_histogram` and `SwanlabLogger.log_histogram` are still bare `return`, so MLflow keeps losing `train/histogram/{gen,input,total}_tokens_length/{mean,p50,p90,max}` (12 series, sync rollout paths only). Deferring is a fair call and the PR description documents it clearly. Noting it here only so the thread records what's outstanding.

Also worth noting for the record: taking `is_histogram_metric` into `metric_utils.py` went further than I asked. I'd flagged the shared helper as optional prose because it needed an import outside the diff — having both consumers call one predicate means they now cannot drift, which was the actual risk.

terrykong (approved):

PR #3750, opened 2026-08-21
feat(refit): support context-parallel draft weight transport

seonjinn:
# What does this PR do ?

Adds context- and pipeline-parallel-aware online draft weight transport and refit on top of #3744.

- Maps train-side CP replicas onto the corresponding MCore TP/PP/CP transport lane.
- Transports metadata and dtype-bucketed draft tensors with lane-wide failure consensus.
- Preflights refits, cached payloads, and ModelOpt real-quant routes before shared collectives.
- Rejects unsupported generation topology before collective initialization.

## Supported contract

- Train-side draft transport supports mapped TP/PP/CP lanes.
- vLLM generation remains limited to `context_parallel_size: 1`.
- Invalid manifests, payloads, route mismatches, and unsupported generation CP fail closed.

# Issues

No issue is closed by this stacked PR.

**Pre checks**:

# Validation

- Exact head: `ea57fed901ee8b5bf8e52b64a583251361995a89`.
- Dependency #3744: `8a03385f2320ed3dd1677a30a52856f54556da3f` (ancestor).
- OCI-HSG `6442377`: PASS (`COMPLETED 0:0`) for the exact head, including a four-rank NCCL CP2 non-owner payload-allocation failure and lane-wide `PAYLOAD_INVALID` consensus.
- The exact head is also included in PR11's terminal full gate `6442673` and packed DFlash/DSpark E2E `6443086`, both PASS.
- This is a correctness feature; no throughput improvement is claimed.

- Reconciled stack head (2026-08-22): `8f397564ecef0df597c62033f8b31cc2eef3d6fa` merges current #3744 `d43a5b6cd82bfd348289a112fa02708765cd130e` into the prior #3750 head.
- OCI-HSG reconciliation gate `6456112` is running the focused then exact affected full matrix from this stack; the prior `6456037` stopped before tests because the base image lacked TransformerEngine on its default path.
- `[PERF-EVIDENCE]` pending; this reconciliation claims no throughput improvement.


seonjinn:

Reviewed exact head `3e36e9708ed38f8172189e49c3dd4bc390a44af4` against stacked dependency #3744 at `ef32181adfa88b6d11426ea3d0fe1b8622e930b7`.

Review scope included CP/PP rank ownership, typed tensor-only draft transport, metadata and payload ordering, standard collective/IPC/NCCL-reshard refit, ModelOpt real-quant routing, repeated refits, generation CP validation, and failure behavior across train and generation workers.

### Findings resolved during self-review

- **Critical:** a draft exporter failure in only one PP lane could let healthy train lanes or generation workers enter later WORLD/model-update collectives and hang. Fixed with tensor-only train-WORLD failure consensus, per-refit draft preflight/cache, and a scalar readiness broadcast as the first shared train/generation model-update collective.
- **Warning:** ModelOpt real-quant refit bypassed the PP-lane `draft_metadata_only` path. Fixed by forwarding both metadata mode and cached draft weights through the inherited lane-aware transport.
- **Warning:** metadata preparation originally propagated WORLD error text from inside `_meta_tensor_alloc_context`, which only preserved PP control broadcasts. Fixed by preflighting metadata-only draft weights outside that context and passing the cache into the meta enumeration.

No unresolved finding at the review confidence threshold remains on the exact head.

### Verification

- Exact head, public fork head, and PR API head all matched before posting.
- All 14 stacked commits have valid signatures and DCO sign-offs; worktree, recursive submodules, and `git diff --check` are clean.
- OCI-HSG job `6409883`: **PASS**, 47:08, exit 0.
  - Static: focused Ruff, Pyrefly baseline-debt guard (64 candidate diagnostics versus 71 baseline), and focused unit suite.
  - Distributed: 2 nodes / 8 GPUs, TP2×PP2×CP2; typed buckets, repeated refits, owner failure release, exact MCore lane selection, and one-lane-only exporter failure consensus all passed.
  - Immutable container SHA-256: `6940409542de6669f77e91c7ce7aac0ef7e91bd56839772e1ae7efc371718d44`.

### Remaining boundaries

- Training CP > 1 is supported; vLLM generation CP > 1 remains explicitly rejected.
- Failures before payload collectives now fail together. An unrecoverable NCCL failure during an already-started tensor payload remains communicator-fatal.
- Cached draft payloads live through NCCL bulk transfer for the duration of one refit; review found no concrete failure, but this is the main residual memory-lifetime risk.

The PR remains Draft because it is stacked on #3744.

tests/unit/models/megatron/test_draft_refit.py
```diff
@@ -0,0 +1,692 @@
...
+        dist.all_reduce(fingerprint_min, op=dist.ReduceOp.MIN)
+        dist.all_reduce(fingerprint_max, op=dist.ReduceOp.MAX)
+        torch.testing.assert_close(fingerprint_min, fingerprint_max, rtol=0, atol=0)
+    finally:
+        draft_module.export_eagle_weights_to_hf = original_export
+        parallel_state.destroy_model_parallel()
+
+
+@pytest.mark.mcore
+def test_tp2_pp2_cp2_single_lane_draft_failure_reaches_every_rank(
```
seonjinn, line 613: Grep-verified: no test exercises the WIRE_INVALID/PROTOCOL_ERROR/PAYLOAD_INVALID paths -- exactly the between-control-and-payload window the failure-consensus design exists for. The follow-up commit adds pure unit tests for the wire-decode failure surface (trailing metadata, truncated names, truncated tensor metadata -> WIRE_INVALID; status-code mapping incl. unknown -> PROTOCOL_ERROR). Still open as a cheap distributed ask: one mp.spawn case forcing payload materialization to fail on a non-owner rank, pinning the PAYLOAD_INVALID consensus at draft/utils.py:1026-1032 (the single-lane exporter-failure test covers the pre-control window only).

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -253,13 +296,17 @@ def _collect_mtp_hf_layer_names(conversion_tasks: Optional[list]) -> set[str]:
 
 
 @contextmanager
-def _meta_tensor_alloc_context():
+def _meta_tensor_alloc_context(
+    *,
+    control_broadcast_group: Optional[torch.distributed.ProcessGroup] = None,
```
seonjinn, line 305: The `control_broadcast_group` escape hatch became unreachable within this PR's own history: it was introduced for preflight-inside-the-meta-context, then a later commit moved preflight outside, and the sole caller always passes precomputed `draft_metadata`, so no int64/uint8 protocol broadcast fires inside the context anymore. Kept (it defends Bridge-driven export paths) but now documented as such in the follow-up commit so nobody assumes it's load-bearing.

seonjinn (commented):
Supplemental self-review (the existing self-review comment on this head stands; every candidate finding was deduped against it and only genuinely additive items are posted).

Method: three review agents (protocol table M1-M7/T/C/R/B/P mapped with line refs, reshard placement math verified byte-for-byte against Megatron-Core's RankGenerator, test/CI audit), then an adversarial pass.

Adversarially refuted, recorded so reviewers need not re-derive: (a) "readiness token doesn't cover the bulk xferdtensor" -- true but pre-existing scope (#3744 territory), the token strictly narrows the hang window, and the gen-side bulk receive is now fail-closed by #3744's follow-up commit; a train-side post-bulk token is a reasonable follow-up, not a blocker. (b) "no cross-CP value-consistency hash" -- refuted by parity with the DP convention: no refit path value-hashes replicated dimensions anywhere; optimizer-synced replica trust is the codebase invariant.

**Process, before undrafting:**

1. **Rebase onto #3744's new head `8a03385f2`** (its follow-up moved the nccl_reshard bulk receive inside the fail-closed lifecycle guard and unified the error contract). Textual conflict risk is low (non-overlapping hunks), but one semantic decision surfaces: after the rebase, this PR's preflight consumer sits OUTSIDE the new try/except+lifecycle guard, so a train-signaled preflight failure would raise raw instead of following the raise-or-return-False contract -- decide deliberately whether preflight belongs inside the guard.
2. No GPU CI has run on GitHub for this head (fork-vetting gate); the OCI-HSG pass is external evidence. `/ok to test` on the post-rebase head.
3. Two added quant-worker tests monkeypatch the method under test and assert the same kwarg back -- consider strengthening or dropping.
4. Worker-level repeated-refit (two full refit calls through the cached path) is untested; the low-level utils loop covers group reuse only.

seonjinn:
Delta re-review of f85be2ef2..ea57fed90: the non-owner PAYLOAD_INVALID consensus test is genuine (fingerprint-matched errors on every PP-lane rank, zero wasted broadcasts) and the rebase onto #3744 is otherwise clean. One gap the rebase resolved the wrong way: `packed_broadcast_preflight_consumer` sat OUTSIDE both the new `try/except` and the `_weight_update_lifecycle` guard, so a train-signaled preflight failure escaped raw — skipping the raise-or-return-False contract and never poisoning the worker — and the new nccl_reshard test monkeypatched the preflight to never fail, hiding exactly this. Fixed in b4f7744da: the preflight now runs inside the guard, plus a fail-closed preflight-failure test (returns False, poisons, stays False).

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -335,11 +335,12 @@ def build_mesh_info(
     rank_offset: int,
     tp_size: int = 1,
     ep_size: int = 1,
+    cp_size: int = 1,
```
yuekaizhang, line 337: This branch is stale relative to current `main` (merge-base predates #3463). On current `main`, `lm_policy.py`'s `LmPolicy.__init__` has a guard: `if draft_enabled and config["megatron_cfg"]["context_parallel_size"] > 1: raise ValueError(...)`, and `DFlashBody.__init__` also hard-rejects `context_parallel_size > 1`. Neither this PR nor #3744 touches either guard. So once this stack rebases onto current `main`, the CP-aware transport added here (`build_mesh_info(cp_size=...)`, `cp_rank`-tagged lanes) is currently unreachable through the standard config path — `LmPolicy.__init__` fails loudly before the draft model or refit path is ever constructed with CP>1. Please confirm whether the `lm_policy.py` guard is meant to be relaxed as part of landing CP support here (and that the hidden-state-capture/per-segment-shift assumption its comment references has actually been addressed by the block-attention/DFlash path), or clarify what config surface exercises this new `cp_size` plumbing today.

yuekaizhang (commented):
Reviewed the CP/PP-aware draft weight transport. The collective/consensus machinery (`broadcast_draft_weights_from_pp_owner`, `_reduce_lane_consensus`, `_raise_draft_refit_failure_on_all_train_ranks`) is unusually careful — every failure path is reached via a symmetric all-reduce/broadcast by all participating ranks before anyone raises, so I could not find a hang/deadlock scenario. CP is correctly modeled as `Replicate()` (not `Shard()`) in `build_mesh_info`/`get_placements`. One integration gap found.

nemo_rl/weight_sync/nccl_reshard_utils.py
```diff
@@ -805,13 +836,15 @@ def _build_train_src_meshes(num_gpus: int, rank_offset: int, stage_pp: int):
...
             ep_size=1,
+            cp_size=cp_size,
             pp_size=stage_pp,
         )
         expert_mesh, expert_dim_map = build_mesh_info(
             num_gpus,
             rank_offset=rank_offset,
             tp_size=1,
             ep_size=ep_size,
+            cp_size=cp_size,
```
isomap: Could you set cp_size=1 for the expert mesh in _build_train_src_meshes()? Pinned MCore (14346b65) uses a separate expert rank generator with cp=1; training CP should remain unchanged.
A four-rank TP1/PP1/EP2/CP2/ETP1 regression through build_nccl_reshard_refit_info() shows source-mesh EP coordinates [0,0,1,1], whereas actual MCore EP ranks are [0,1,0,1]. The synthetic transfer reconstructs experts [0,1,0,1] instead of [0,1,2,3]. The recorded run passes after changing only that caller line.
Please also update test_build_refit_info_replicates_train_context_parallelism: its expert-mesh expectations currently encode the incorrect layout. After the fix, that test’s expert mesh should be (4,2) with [Replicate(), Shard(0)].

seonjinn: Applied in 18736c0. `_build_train_src_meshes()` now passes `cp_size=1` for the expert mesh only; training CP is untouched. A comment records why: MCore lays experts out with a separate rank generator whose CP size is 1 and folds the CP ranks into the expert data-parallel dim, so emitting a CP axis here makes CP - not EP - the innermost dim and `global_rank % ep_size` recovers the wrong EP coord.

`test_build_refit_info_replicates_train_context_parallelism` is updated accordingly: the non-expert mesh keeps `(2, 2, 2)` with `Replicate()` on the CP axis, and the expert mesh is now `(4, 2)` with `[Replicate(), Shard(0)]` and a flattened rank tensor of `[0..7]`, so rank `r` owns expert coord `r % 2`.

nemo_rl/weight_sync/nccl_reshard_utils.py
seonjinn: No code change here; this PR is transport-only groundwork, and the guard you point at is relaxed one PR later.

The branch now carries latest `main`, so the guards you describe are present. The `lm_policy.py` blanket CP>1 draft guard is intentionally left alone in this PR and is replaced in #3757 by a capability gate rather than being deleted: `draft_config.py` grows a `supports_context_parallel` ClassVar that is `False` for `Eagle3DraftConfig` and `True` for `DFlashDraftConfig` / `DSparkDraftConfig`, and `lm_policy.py` gates on that instead of on `draft_enabled`. The eagle3 hidden-state-capture / per-segment-shift assumption its comment references is therefore still enforced - eagle3 keeps rejecting CP>1.

On top of that, #3757's `megatron_policy_worker.py` validates `context_parallel_size in (1, 2, 4)` and requires `sequence_packing.enabled=true` whenever CP>1, so the config surface that exercises this `cp_size` plumbing is DFlash/DSpark with packed sequences.

`DFlashBody.__init__`'s own `context_parallel_size > 1` raise stays as is, and that is deliberate: `nemo_rl/models/megatron/draft/training.py` builds the body's `ModelParallelConfig` without setting `context_parallel_size`, so it defaults to 1 and the body runs on the un-sharded sequence while the target is CP-sharded.

So today, on this PR alone, the new `cp_size` plumbing is reachable only from the unit tests; #3757 is what opens the config path. Happy to reorder the two if you would rather see the gate land with the transport.

seonjinn:
The `cp_size=1` expert-mesh fix is in and replied to in-thread, together with the updated `test_build_refit_info_replicates_train_context_parallelism` expectations.

Head is merged with latest `main` (`c49d53e2e`), clean merge, lint clean tree-wide, and `/ok to test` re-posted against `ae7f72c5a4cdc165cc1a90161e858ac7ea9ecf61`.

One thread is deliberately left open: the ordering question on the `lm_policy.py` CP>1 draft guard. This PR is transport-only groundwork and #3757 is what turns that blanket guard into a `supports_context_parallel` capability gate, so whether to reorder the two is your call, not something I want to resolve unilaterally.

PR #3751, opened 2026-08-21
feat: performance optimization of clipped pg loss

CarlosGomes98:
# What does this PR do ?

Optimize the clipped pg loss computation by:
- reducing cpu-gpu sync points
- Adding "minimal" option that avoids computing diagnostic metrics in the loss function
- adds torch.compile to a compute-intensive section of the loss

**Behaviour Change**
This PR also corrects loss reporting in the V1 DTensor policy worker. This is an intentional behavior change and should be considered when comparing loss curves across versions.
In training mode, the V1 worker currently reports global_loss after multiplying the loss by dp_size * cp_size. That scaling is used only to compensate for FSDP gradient averaging during backward and should not affect the reported objective. Consequently, the reported loss is inflated when dp_size * cp_size > 1.
The updated implementation reports the unscaled mathematical loss instead. This is also consistent with the V2 DTensor worker, which scales a separate tensor for backward while retaining the original unscaled loss for reporting.

## Measured improvement

Measured a reduction in loss computation form 2.3ms -> 0.1ms. Can become significant with high GA.

The diff is large but most of it is indentation changes / moving stuff around.

## Additional:
- Added tests with "golden values" for the loss to make sure numerics are not impacted by these changes. These were mostly for the PR itself, I'm happy to remove before merging, or keep if its useful in the future.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


CarlosGomes98:
/ok to test [3cfc600](https://github.com/NVIDIA-NeMo/RL/pull/3751/commits/3cfc6005a0145781453fede0188a1946d8ae8a20)

nemo_rl/models/policy/workers/dtensor_policy_worker.py
```diff
@@ -959,7 +959,10 @@ def train(
                             loss.backward()
 
                     if num_valid_samples > 0:
-                        mb_losses.append(loss.item())
+                        # Metrics were materialized together by the loss;
+                        # undo this worker's per-global-batch scaling without
+                        # synchronizing the loss tensor again.
+                        mb_losses.append(loss_metrics["loss"] * num_global_batches)
```
CarlosGomes98, line 970: In training mode the old `mb_losses.append(loss.item())` was taken *after* `loss *= self.dp_size * self.cp_size`, so on multi-GPU the reported `global_loss` will now shift by that `dp_size*cp_size` factor. If the new (un-inflated) value is the intended fix, it'd be worth calling out in the PR description, since existing loss curves will step-change; if metric parity was intended, multiply by `self.dp_size * self.cp_size` here too. Either way worth a multi-GPU sanity check, since no test covers `dp_size*cp_size > 1` (the same applies to the mirrored line in `dtensor_policy_worker_v2.py`).

CarlosGomes98 (commented):
From the /review-pr skill, as requested:
One question on the multi-GPU loss metric; core loss-fn optimization LGTM.

CarlosGomes98:
/ok to test [bb2c403](https://github.com/NVIDIA-NeMo/RL/pull/3751/commits/bb2c403783228719e94b274f11a12b17de5b46e7)

CarlosGomes98:
/ok to test [5d6e31b](https://github.com/NVIDIA-NeMo/RL/pull/3751/commits/5d6e31b0949402329fc6b429df5cad4a446ba030)

nemo_rl/models/policy/workers/dtensor_policy_worker.py
CarlosGomes98: In the V1 worker, the old global_loss was taken after the in-place dp_size * cp_size multiplication. That factor seems to exist to compensate for FSDP’s averaged gradient reduction during backward; it should not affect the reported mathematical objective. I believe the previous global_loss was incorrectly scaled and the new uninflated value is the correct behavior.

However, this is still an observable change for configurations where dp_size * cp_size > 1. I added a two-GPU parity assertion checking that global_loss matches the aggregated "loss" metric independently of backward scaling.

Importantly, V2 did not have exactly the same reporting bug: it scales a separate tensor for backward while returning the original unscaled result. That reinforces that I think this is now the correct behaviour

CarlosGomes98:
/ok to test [4ab30dd](https://github.com/NVIDIA-NeMo/RL/pull/3751/commits/4ab30dd19f796d06becbac00ad1b77a2b2ff2615)

CarlosGomes98:
/ok to test [ba8fae9](https://github.com/NVIDIA-NeMo/RL/pull/3751/commits/ba8fae93921c72bd3839e2fd94e6362543729d60)

yuki-97 (commented):
@bg51717 could you help review?

nemo_rl/models/policy/workers/dtensor_policy_worker.py
```diff
@@ -964,7 +964,10 @@ def train(
...
                         all_mb_metrics.append(loss_metrics)
```
bg51717: **1 action item.**

**Under sequence packing, each dummy microbatch now adds `num_global_batches ×` a real microbatch to the reported loss — it used to add 0.**

`loss_metrics` is computed before the `mb_idx < iterator_len` guard, so on a dummy it never gets divided by `num_global_batches`; `num_valid_samples` is set *inside* the guard, so it's stale-positive and this line still fires. The old `loss.item()` read the tensor after `loss *= 0`, hence 0. Gradients are unaffected — only `train/loss` is wrong.

docs/guides/grpo.md
```diff
@@ -628,6 +628,12 @@ Under tensor parallelism (TP), enabling top-p or top-k adds communication overhe
 ## Metrics
 This feature is controlled by the parameters `wandb_name` and `tb_name`. We track a few metrics during training for scientific experimentation and to validate correctness as the run progresses.
 
+`loss_fn.metrics_level` defaults to `full`, preserving all diagnostics described
```
bg51717, line 631: **1 action item.** [`grpo.md:682`](https://github.com/NVIDIA-NeMo/RL/blob/ba8fae93921c72bd3839e2fd94e6362543729d60/docs/guides/grpo.md#L682) says `sampling_importance_ratio` "is always calculated", which `minimal` makes false — it's only emitted when `use_importance_sampling_correction` is on ([loss_functions.py#L394](https://github.com/NVIDIA-NeMo/RL/blob/ba8fae93921c72bd3839e2fd94e6362543729d60/nemo_rl/algorithms/loss/loss_functions.py#L394)).

bg51717 (commented):
LGTM overall.

Speeds up the clipped PG loss three ways: the per-metric `.item()` calls are batched into a single host transfer, the actor objective gets an opt-in `torch.compile` path, and `metrics_level: minimal` lets users skip the diagnostic-only metrics. The objective math and the diagnostic metrics move out of `loss_functions.py` into a new `nemo_rl/algorithms/loss/clipped_pg.py`.

nemo_rl/models/policy/workers/dtensor_policy_worker.py
CarlosGomes98: Makes sense, added an explicit check for dummy_batches and a test for this specific sequence packing case

docs/guides/grpo.md
CarlosGomes98: Fixed the doc

CarlosGomes98:
/ok to test [c0bdc27](https://github.com/NVIDIA-NeMo/RL/pull/3751/commits/c0bdc27d508a559dfdcc0a091b410fca157d45ff)

CarlosGomes98:
/ok to test [e760706](https://github.com/NVIDIA-NeMo/RL/pull/3751/commits/e7607060b3bdd411aa92b92dcf3190929ace9fa6)

nemo_rl/algorithms/loss/wrapper.py
```diff
@@ -381,7 +381,9 @@ def __call__(
                 **loss_input,
             )
         combined_loss = policy_loss + self.loss_weight * draft_loss
-        metrics["draft_loss"] = float(draft_loss.detach().item())
+        draft_loss_value = float(draft_loss.detach().item())
+        metrics["loss"] += self.loss_weight * draft_loss_value
```
RayenTian, line 385: Could we preserve metrics["loss"] as the policy loss and report the combined objective under a separate total_loss metric?
This silently changes the meaning of train/loss for draft training, making historical curves and existing loss thresholds no longer directly comparable.

nemo_rl/algorithms/loss/loss_functions.py
```diff
@@ -378,6 +407,20 @@ def __init__(
             "probs_ratio_clamped_min": MetricNormalizer.NONE,
             "probs_ratio_clamped_max": MetricNormalizer.NONE,
         }
+        if self.metrics_level == "minimal":
+            minimal_metrics = {
+                "loss",
+                "kl_penalty",
+                "num_valid_samples",
+                "positive_nll_loss",
```
RayenTian, line 415: Could we keep `token_mult_prob_error` in minimal mode? It is an important signal for training–generation logprob mismatch, and omitting it also disables the existing > 1.05 diagnostic plot. Please compute and export it in both modes while keeping the other diagnostics optional.

nemo_rl/algorithms/grpo.py
```diff
@@ -3933,7 +3933,7 @@ def _grpo_train_impl(
                 reduction_op="sum"
             )  # type: ignore
             # track example with high token mult prob error above 1.05
-            if metrics["token_mult_prob_error"] > 1.05:
+            if metrics.get("token_mult_prob_error", 0.0) > 1.05:
```
RayenTian, line 3936: If we keep `token_mult_prob_error` in minimal metric set, then the 0.0 fallback isn't needed any more.

nemo_rl/algorithms/grpo.py
```diff
@@ -3966,7 +3966,15 @@ def _grpo_train_impl(
...
             if "draft_loss" in metrics:
                 print(f"  • Draft Loss: {metrics['draft_loss']:.4f}")
-            print(f"  • Generation KL Error: {metrics['gen_kl_error']:.4f}")
+            generation_kl_error = metrics.get("gen_kl_error")
+            print(
+                "  • Generation KL Error: "
+                + (
+                    f"{generation_kl_error:.4f}"
+                    if generation_kl_error is not None
+                    else "N/A"
```
RayenTian, line 3975: Could we use "not reported" instead of "N/A" here? In minimal mode, this metric is intentionally omitted, and "N/A" could be mistaken for missing data or a computation failure.

nemo_rl/algorithms/grpo.py
```diff
@@ -5879,7 +5887,15 @@ def _flush_collector_telemetry() -> None:
...
```
RayenTian, line 5896: Same as this [comment](https://github.com/NVIDIA-NeMo/RL/pull/3751/changes#r4060714964).

nemo_rl/algorithms/grpo_sync.py
```diff
@@ -1337,7 +1337,7 @@ def grpo_train_sync(
                 del log_data
 
             timing_metrics: dict = timer.get_timing_metrics(reduction_op="sum")  # type: ignore
-            if metrics["token_mult_prob_error"] > 1.05:
+            if metrics.get("token_mult_prob_error", 0.0) > 1.05:
```
RayenTian, line 1340: Same as this [comment](https://github.com/NVIDIA-NeMo/RL/pull/3751/changes#r4060690514).

nemo_rl/algorithms/grpo_sync.py
```diff
@@ -1368,7 +1368,15 @@ def grpo_train_sync(
...
             if "draft_loss" in metrics:
                 print(f"  • Draft Loss: {metrics['draft_loss']:.4f}")
-            print(f"  • Generation KL Error: {metrics['gen_kl_error']:.4f}")
+            generation_kl_error = metrics.get("gen_kl_error")
+            print(
+                "  • Generation KL Error: "
+                + (
+                    f"{generation_kl_error:.4f}"
+                    if generation_kl_error is not None
+                    else "N/A"
```
RayenTian, line 1377: Same as this [comment](https://github.com/NVIDIA-NeMo/RL/pull/3751/changes#r4060714964).

tests/unit/algorithms/test_loss_functions.py
```diff
@@ -2984,6 +2985,135 @@ def test_nll_loss_advertises_counts_as_none(self):
         assert norms["num_unmasked_tokens"] is MetricNormalizer.NONE
         assert norms["num_valid_samples"] is MetricNormalizer.NONE
 
+    def test_minimal_metrics_preserve_loss_and_drop_diagnostics(self):
```
RayenTian, line 2988: Could we extend the minimal-mode tests to check that the returned metric keys match metric_normalizations, with the correct normalization for each metric? These are maintained separately, so adding a metric to the output without updating its normalization could pass the current tests.

RayenTian:
Thanks for adding the optimization @CarlosGomes98 ! Left some comments.

PR #3753, opened 2026-08-21
fix(grpo): correct sequence importance ratio metric

CarlosGomes98:
# What does this PR do ?

Fixes an issue in the metric computation of importance ratio

The issue seems to be that the importance weights are reshaped to [B, 1] in order to be applied to all the tokens in each sequence, but are left that way when they are then reused together with the sample mask. That broadcasts to [B, B]. The whole matrix then gets summed and divided by the number of valid samples, which results in an error. e.g. for 2 samples, both valid the value will be 2x as large as it should be.

The training loss is unnafected, but the metric is wrong

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


yuki-97 (approved):
good catch and thanks for the fix!

PR #3754, opened 2026-08-21
fix(vllm): apply FP8 patches with RayExecutorV2

### LINKED ISSUE #3755: [Bug] Multi-GPU FP8 refit patches are skipped with vLLM 0.25.1 RayExecutorV2
## Summary

NeMo-RL's custom vLLM FP8 patches are not applied to internal vLLM workers when:

- `precision: fp8`
- `tensor_parallel_size * pipeline_parallel_size > 1`
- vLLM's Ray executor backend is used
- `VLLM_USE_RAY_V2_EXECUTOR_BACKEND` is unset or set to `1`

vLLM 0.25.1 defaults `VLLM_USE_RAY_V2_EXECUTOR_BACKEND` to `1`, so Ray-backed engines select `RayExecutorV2`. NeMo-RL currently installs its multi-GPU FP8 worker patches by monkey-patching only `RayDistributedExecutor.collective_rpc`, which is the V1 executor and is not used under the 0.25.1 default.

As a result, the FP8 configuration and refit-specific patches never reach the V2 Ray worker processes.

Related vLLM 0.25 upgrade tracking: #3412.

## Affected code

`monkey_patch_vllm_ray_executor()` currently imports and patches only the V1 executor:

https://github.com/NVIDIA-NeMo/RL/blob/737c4d2cd629d19e5c576b67ef2a6105b6b1bbb3/nemo_rl/models/generation/vllm/quantization/fp8.py#L95-L121

```python
from vllm.v1.executor.ray_executor import RayDistributedExecutor

original_run_workers = RayDistributedExecutor.collective_rpc
...
RayDistributedExecutor.collective_rpc = patched_run_workers
```

In vLLM 0.25.1, the environment variable defaults to `1`:

https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/envs.py#L866-L870

and `Executor.get_class()` selects `RayExecutorV2`:

https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/abstract.py#L60-L68

## Root cause

The V1 and V2 worker initialization paths differ.

### V1

`RayDistributedExecutor` exposes `self.workers`, and its early `collective_rpc` calls happen before model initialization. NeMo-RL intercepts the first call and invokes `apply_fp8_patches()` in every Ray worker.

### V2

`RayExecutorV2` creates lightweight `RayWorkerProc` actors and later calls:

```python
actor.initialize_worker.remote(...)
```

`RayWorkerProc.initialize_worker()` then calls `super().__init__()`, which enters `WorkerProc.__init__()` and immediately performs:

```python
self.worker.init_device()
self.worker.load_model()
```

Relevant vLLM sources:

- https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/ray_executor_v2.py#L139-L170
- https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/ray_executor_v2.py#L425-L448
- https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/multiproc_executor.py#L594-L634

Therefore, patching `RayExecutorV2.collective_rpc` would still be too late: the model has already been loaded and processed by the time the MQ-based RPC path becomes available.

## Expected behavior

Before each internal vLLM worker loads its FP8 model, NeMo-RL should:

1. Propagate the `FP8Config` into that worker process.
2. Install the FP8 Linear/MoE/MXFP8/KV-cache patches.
3. Preserve parameter identity and `weight_loader` metadata required by refit.
4. Set the worker-local `global_fp8_config`.

This should work with both V1 and V2 Ray executors.

## Actual behavior

With the V2 executor:

- `apply_fp8_patches()` is never invoked in the internal Ray workers.
- Worker-local `global_fp8_config` remains `None`.
- Model loading uses unpatched upstream `process_weights_after_loading()` methods.
- Refit-required parameter identity and `weight_loader` behavior are not installed.
- The first FP8 weight update may fail when `fp8.load_weights()` accesses `global_fp8_config.is_mx`, or fail later because required loaders/scales were replaced or removed.

Initial model loading or the first rollout may appear to work because upstream vLLM supports serving FP8 models. The failure is exposed when NeMo-RL performs a weight refit.

## Scope

The issue affects both synchronous and asynchronous generation when all of the following are true:

```text
precision == fp8
TP * PP > 1
distributed_executor_backend == ray
VLLM_USE_RAY_V2_EXECUTOR_BACKEND != 0
```

This includes:

- Block FP8 and MXFP8.
- Dense and MoE models.
- Colocated and non-colocated generation.
- FP8 KV cache when used with FP8 weights.

It does not affect:

- `TP=1, PP=1`.
- Non-FP8 generation.
- Runs explicitly setting `VLLM_USE_RAY_V2_EXECUTOR_BACKEND=0`.

## Why existing tests did not catch this

All FP8 recipes currently listed in the standard nightly suite resolve to:

```text
vLLM TP=1
vLLM PP=1
model_parallel_size=1
```

Although those jobs may use many GPUs or nodes, they run multiple single-GPU vLLM replicas. They take the direct `apply_fp8_patches()` branch and never exercise either Ray executor.

The FP8 unit tests also use `model_parallel_size=1` and test the patch contents rather than executor selection and cross-process installation.

The existing checked-in FP8 configurations with `TP * PP > 1` are in example, release, or performance suites rather than the standard FP8 nightly coverage.

## Proposed fix

Keep the existing V1 path and add a V2-specific pre-model-load hook.

For V2, patch `RayWorkerProc.initialize_worker()` before `RayExecutorV2` creates its actors:

```python
original_initialize_worker = RayWorkerProc.initialize_worker

def patched_initialize_worker(self, *args, **kwargs):
    global fp8_patches_applied

    if not fp8_patches_applied:
        apply_fp8_patches(None, fp8_config)

    return original_initialize_worker(self, *args, **kwargs)
```

The patch must run before `original_initialize_worker()` because that call ultimately loads the model.

The implementation should also be idempotent and avoid wrapping either executor method more than once.

## Acceptance criteria

- [ ] FP8 patches are installed in every V2 Ray worker before model loading.
- [ ] Multi-GPU FP8 refit succeeds with the vLLM 0.25.1 default executor.
- [ ] Both sync and async vLLM engines are covered.
- [ ] The existing V1 path continues to work with `VLLM_USE_RAY_V2_EXECUTOR_BACKEND=0`.
- [ ] The single-GPU path continues to work.
- [ ] Unit tests verify executor selection and patch-before-model-load ordering.
- [ ] Add or update a test using at least `vLLM TP=2` that completes one real FP8 weight refit.

## Workaround

Force the V1 executor for affected FP8 recipes:

```yaml
policy:
  generation:
    vllm_cfg:
      env_vars:
        VLLM_USE_RAY_V2_EXECUTOR_BACKEND: "0"
```

This restores the old patch path but does not provide native RayExecutorV2 support.

zpqiu:
## Summary

- apply NeMo-RL FP8 patches in `RayWorkerProc.initialize_worker` before vLLM initializes and loads the model
- preserve the existing Ray V1 and single-GPU paths
- add regression coverage for the vLLM 0.25.1 default `RayExecutorV2` selection and patch ordering

## Problem

vLLM 0.25.1 enables `VLLM_USE_RAY_V2_EXECUTOR_BACKEND` by default. Multi-GPU FP8 initialization currently patches only `RayDistributedExecutor.collective_rpc`, so the hook is inactive when vLLM selects `RayExecutorV2`. The internal workers can therefore initialize and load their models without the NeMo-RL FP8 refit patches.

## Fix

When `RayExecutorV2` is selected, wrap `RayWorkerProc.initialize_worker` and call the existing `apply_fp8_patches` implementation before the original worker initialization. No new configuration or fallback path is introduced.

## Testing

- focused regression against vLLM 0.25.1
  - before fix: failed because worker initialization ran while `fp8_patches_applied` was false
  - after fix: both Ray V1 and Ray V2 cases pass (`2 passed`)
- recipe config validation: `295 passed`
- Ruff: passed
- FP8 nightly regressions:

| Recipe | Generation path | Result |
| --- | --- | --- |
| Qwen3 8B FP8 KV-cache, TP2 | Ray V2 | 60/60 steps, metric pass; median `train/token_mult_prob_error` 1.105 |
| Qwen3 8B FP8 KV-cache, TP1 control | Single GPU | 60/60 steps, metric pass; median `train/token_mult_prob_error` 1.075 |
| Qwen3.5 35B-A3B FP8, TP2 | Ray V2 | 20/20 steps, nightly metric checker pass |
| Qwen3.5 9B FP8, TP1 | Single GPU | 20/20 steps, nightly metric checker pass |

W&B project: https://wandb.ai/nvidia/nemorl-pr3754

Fixes #3755


examples/configs/recipes/llm/grpo-qwen3-8b-base-1n8g-fp8-kvcache-megatron.yaml
```diff
@@ -33,6 +33,7 @@ policy:
     vllm_cfg:
       precision: fp8
       kv_cache_dtype: fp8
+      tensor_parallel_size: 2
```
zpqiu: [`grpo-qwen3-8b-base-1n8g-fp8-kvcache-megatron.yaml:36`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/examples/configs/recipes/llm/grpo-qwen3-8b-base-1n8g-fp8-kvcache-megatron.yaml#L36)

Thanks for running the nightly regression on this — the wandb run linked in the description covers it, so the convergence question is answered.

One process item that is easy to miss because the PR page looks green: **the repo's own CI has not run on this commit yet.** `check-runs` for `b63518519` contains only DCO / copyright / secrets / submodule / semantic-PR / triage — no `lint-check`, no `CI_QA_Gate`. The main pipeline triggers on a push to the `pull-request/[0-9]+` mirror branch ([`cicd-main.yml#L16-L20`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/.github/workflows/cicd-main.yml#L16-L20)), which copy-pr-bot has not created yet, so it never fired.

Two things needed:
1. A vetter has to trigger CI (per the copy-pr-bot comment at the top of this PR).
2. This PR has no `CI:*` label, so even once CI runs it resolves to `TEST_LEVEL="none"` ([`cicd-main.yml#L188-L190`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/.github/workflows/cicd-main.yml#L188-L190)) and the quality gate stays red. Since this line changes a convergence-gated nightly's generation TP, the cicd skill's table points to **`CI:L2`**.

Since the wandb project isn't resolvable outside NVIDIA, it might be worth pasting the headline numbers into the PR body — the median `train/token_mult_prob_error` at TP=2 and the wall-clock against `NUM_MINUTES=240` — so reviewers without access can see the result without asking.

Minor, non-blocking: this is the only recipe in `examples/configs/recipes/` with `kv_cache_dtype: fp8`, so switching it to TP=2 converts that coverage rather than adding to it — nothing exercises FP8 KV-cache at TP=1 afterwards. Fine if intended, just worth being deliberate about.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -94,6 +95,20 @@ def my_run_engine_core(*args, **kwargs):
...
+        if envs.VLLM_USE_RAY_V2_EXECUTOR_BACKEND:
+            from vllm.v1.executor.ray_executor_v2 import RayWorkerProc
+
+            original_initialize_worker = RayWorkerProc.initialize_worker
+
+            def patched_initialize_worker(self, *args, **kwargs):
+                if not fp8_patches_applied:
+                    apply_fp8_patches(None, fp8_config)
+
+                return original_initialize_worker(self, *args, **kwargs)
```
zpqiu, line 115: [`fp8.py:104`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/nemo_rl/models/generation/vllm/quantization/fp8.py#L104)

This guard can't do its job inside the worker. `patched_initialize_worker` is a nested function, so cloudpickle ships it into each `RayWorkerProc` actor **by value together with a snapshot of the globals it references**. In the actor, `fp8_patches_applied` is therefore frozen at its export-time value, while `apply_fp8_patches` — pickled by reference — mutates the *real* module dict at [`fp8.py:201`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/nemo_rl/models/generation/vllm/quantization/fp8.py#L201). We measured this: the wrapper's `__globals__` is not the module `__dict__`, and a second `initialize_worker` call on one actor re-enters and trips [`assert not fp8_patches_applied`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/nemo_rl/models/generation/vllm/quantization/fp8.py#L141).

**Not reachable on vLLM 0.25.1** — `initialize_worker` has exactly one call site, [`ray_executor_v2.py#L434`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/ray_executor_v2.py#L434), once per actor, and `.options(...)` sets no `max_restarts`. So this is hardening, not a blocker. Worth doing anyway because it diverges from the v1 branch 15 lines below, which correctly declares `global` and writes the flag back ([`fp8.py:119-126`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/nemo_rl/models/generation/vllm/quantization/fp8.py#L119-L126)), and because #3755 asked for the implementation to "be idempotent and avoid wrapping either executor method more than once".

Resolving the module through `sys.modules` reads the actor's live flag instead of the snapshot:

```suggestion
            def patched_initialize_worker(self, *args, **kwargs):
                # Ray cloudpickles this closure into each RayWorkerProc actor
                # along with a *snapshot* of the globals it references, so a bare
                # `fp8_patches_applied` read here would be frozen at its
                # export-time value. Resolve through the module object instead so
                # the guard sees the flag apply_fp8_patches actually sets.
                from nemo_rl.models.generation.vllm.quantization import fp8 as _fp8

                if not _fp8.fp8_patches_applied:
                    _fp8.apply_fp8_patches(None, fp8_config)

                return original_initialize_worker(self, *args, **kwargs)
```

(The existing test's `monkeypatch.setattr(fp8, "apply_fp8_patches", ...)` still works with this form.) A matching `_nrl_fp8_wrapped` guard before the rebind would cover the "wrap once" half of that acceptance criterion.

nemo_rl/models/generation/vllm/quantization/fp8.py
```diff
@@ -94,6 +95,20 @@ def my_run_engine_core(*args, **kwargs):
 
 def monkey_patch_vllm_ray_executor(fp8_config):
     if fp8_config.model_parallel_size > 1:
+        if envs.VLLM_USE_RAY_V2_EXECUTOR_BACKEND:
```
zpqiu, line 102: [`fp8.py:98`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/nemo_rl/models/generation/vllm/quantization/fp8.py#L98)

Could you add a couple of lines of comment here recording *why* a local class assignment is sufficient? It is genuinely non-obvious — the v1 branch right below deliberately uses a **remote** call ([`fp8.py:122`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/nemo_rl/models/generation/vllm/quantization/fp8.py#L122)) to get its patches into the workers, so a reader is likely to assume this branch is broken. We initially did.

It works because [`ray.remote(RayWorkerProc)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/ray_executor_v2.py#L379) wraps the class in a dynamic subclass that Ray pickles **by value**, and Ray unconditionally copies inherited methods onto that subclass — [`actor.py#L2684`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L2684) calls `_inject_tracing_into_class`, which does [`inspect.getmembers(_cls, is_function_or_method)`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/util/tracing/tracing_helper.py#L540) and [`setattr(_cls, name, wrapped_method)`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/util/tracing/tracing_helper.py#L600).

That is an undocumented Ray implementation detail, and it is load-bearing here. It would silently stop applying if `initialize_worker` ever became a `staticmethod`/`classmethod` or a generator — both are skipped by that copy loop ([`#L545`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/util/tracing/tracing_helper.py#L545), [`#L548`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/util/tracing/tracing_helper.py#L548)).

Mitigating, and worth saying in the comment too: the failure would be **loud, not silent numerics** — with `global_fp8_config` left `None` in the worker, the first refit raises `AttributeError` on `global_fp8_config.is_mx`. So this is maintainability rather than a correctness risk.

tests/unit/models/generation/test_vllm_fp8_quantization.py
```diff
@@ -519,6 +519,56 @@ def fake_patch(path, _replacement):
...
+    monkeypatch.setattr(RayWorkerProc, "initialize_worker", fake_initialize_worker)
+    # Record the V1 method with monkeypatch so the direct class assignment made by
+    # monkey_patch_vllm_ray_executor() is undone even when this regression test fails.
+    monkeypatch.setattr(
+        RayDistributedExecutor,
+        "collective_rpc",
+        RayDistributedExecutor.collective_rpc,
+    )
+
+    fp8.monkey_patch_vllm_ray_executor(fp8_config)
```
zpqiu, line 973: [`test_vllm_fp8_quantization.py:563`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/tests/unit/models/generation/test_vllm_fp8_quantization.py#L563)

This is the only place in the entire test suite that calls the real `monkey_patch_vllm_ray_executor` — the other two references, [line 55](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/tests/unit/models/generation/test_vllm_fp8_quantization.py#L55) and [line 446](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/tests/unit/models/generation/test_vllm_fp8_quantization.py#L446), are `monkeypatch.setattr` stubs. So the v1 `RayDistributedExecutor.collective_rpc` path now has no regression coverage at all, and this PR adds an early [`return` at `fp8.py:110`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/nemo_rl/models/generation/vllm/quantization/fp8.py#L110) directly in front of it.

That is the one acceptance criterion from #3755 still open: *"The existing v1 path continues to work with `VLLM_USE_RAY_V2_EXECUTOR_BACKEND=0`."*

Cheap fix: parametrize this test over `VLLM_USE_RAY_V2_EXECUTOR_BACKEND` in `{"0", "1"}` and assert the v1 case rebinds `RayDistributedExecutor.collective_rpc` while the v2 case rebinds `RayWorkerProc.initialize_worker`. That also removes the dependency on upstream's default noted below.

tests/unit/models/generation/test_vllm_fp8_quantization.py
```diff
@@ -519,6 +519,56 @@ def fake_patch(path, _replacement):
     assert all(patcher.started for patcher in fp8.fp8_state.vllm_patches)
 
 
+def test_multi_gpu_fp8_patches_ray_v2_worker_before_model_load(
+    fp8_module, monkeypatch
+):
```
zpqiu: [`test_vllm_fp8_quantization.py:522`](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/tests/unit/models/generation/test_vllm_fp8_quantization.py#L522)

`ruff-format` will reformat this — the signature is exactly 88 characters and fits on one line. Verified with the pinned hook version: `uvx ruff@0.9.9 format --diff` reports "1 file would be reformatted". `ruff-format` is a pre-commit hook at rev `v0.9.9` and pre-commit runs in the `lint-check` job, so this will fail once CI is unblocked. (`ruff check` is clean.)

```suggestion
def test_multi_gpu_fp8_patches_ray_v2_worker_before_model_load(fp8_module, monkeypatch):
```

Unrelated nit on the same test, take it or leave it: [lines 538-539](https://github.com/NVIDIA-NeMo/RL/blob/b63518519ef19e8a5c3b26e64d4145ec72dc6149/tests/unit/models/generation/test_vllm_fp8_quantization.py#L538-L539) do `monkeypatch.delenv(...)` then `assert envs.VLLM_USE_RAY_V2_EXECUTOR_BACKEND is True`, which makes the branch under test depend on upstream's default rather than on the test. vLLM has already flipped that default once (`"0"` in v0.20.0 to `"1"` by v0.24.0). `monkeypatch.setenv(..., "1")` would pin the branch; keep the default check as a separate assertion if you want it as a canary.

zpqiu (commented):
Reviewed by a team of five agents (RL codebase, vLLM/Ray upstream, bug finder, test, devil's advocate).

**The fix is correct — nice catch on the root cause.** Four of us independently verified the mechanism end-to-end against real vLLM 0.25.1 + Ray 2.56.1: a driver-side `RayWorkerProc.initialize_worker` assignment *does* reach the Ray actor processes. Inside the actor (different pid), `global_fp8_config` is populated and `Fp8LinearMethod.process_weights_after_loading` is NeMo-RL's replacement. `initialize_worker` also looks like the only clean hook `RayExecutorV2` offers — `RayWorkerProc` extends `WorkerProc` rather than v1's `RayWorkerWrapper`, so there is no `execute_method` to RPC into, and the inherited `collective_rpc` rides `rpc_broadcast_mq`, which is only live *after* `initialize_worker` — too late to patch before ([`load_model()`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/multiproc_executor.py#L634)).

We also confirmed the new unit test is a genuine regression test: it fails against the pre-fix module and passes with the fix, and the full file is green (`23 passed`).

The comments below are hardening, evidence and CI — none of them challenge the approach.

The nightly regression linked in the description covers the convergence side. Separately, and only because it is invisible from the PR page: the repo CI pipeline has not fired on this commit yet — see the inline comment on the recipe change.

_Generated by Claude Code_

tests/unit/models/generation/test_vllm_fp8_quantization.py
```diff
@@ -519,6 +519,65 @@ def fake_patch(path, _replacement):
...
+        closure=patched_initialize_worker.__closure__,
+    )
+    worker_initialize_worker(object(), 0, {})
+    worker_initialize_worker(object(), 0, {})
+
+    assert events == [
+        ("apply_fp8_patches", fp8_config),
+        ("initialize_worker", None),
+        ("initialize_worker", None),
+    ]
```
zpqiu: [`test_vllm_fp8_quantization.py:522-578`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/tests/unit/models/generation/test_vllm_fp8_quantization.py#L522-L578)

This is still the only real call to `monkey_patch_vllm_ray_executor` in the suite ([L55](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/tests/unit/models/generation/test_vllm_fp8_quantization.py#L55) and [L446](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/tests/unit/models/generation/test_vllm_fp8_quantization.py#L446) are `monkeypatch.setattr` stubs), so the V1 `collective_rpc` branch now sitting behind the early [`return` at fp8.py:114](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/nemo_rl/models/generation/vllm/quantization/fp8.py#L114) has no unit assertion. To be precise: it is not uncovered everywhere — `grpo-qwen3.5-35ba3b-2n8g-megatron-ep16tp2-fp8.yaml` pins the V1 executor and exercises it nightly — but nothing gates this PR on it, so #3755's last acceptance criterion ("the existing v1 path continues to work with `VLLM_USE_RAY_V2_EXECUTOR_BACKEND=0`") is unpinned.

We wrote the parametrized version and mutation-tested it rather than just proposing it. Against 11 mutations of `fp8.py`, the current test catches **5**; the version below catches **11**. The ones that slip through today are all real breakage: never installing the V1 hook, removing the early `return`, ignoring the env var, and — the sharpest one — changing the forward to `original_initialize_worker(self)`, which breaks production hard since [`local_rank` and `env_vars` are required positionals](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/ray_executor_v2.py#L139-L145) yet passes silently because the fake swallows `*args`.

It needs no Ray cluster and no GPU: [`envs.__getattr__`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/envs.py#L1972) re-reads the environment on every access, so `monkeypatch.setenv` selects the branch directly. The `envs.disable_envs_cache()` line guards a latent order dependency — [`enable_envs_cache()`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/envs.py#L1990) memoizes those lookups once any engine is built in-process, which would make `setenv` a silent no-op; it does not bite today (this file sorts before the one that builds a real `LLM()`), and there is in-repo precedent at `test_vllm_message_queue_port.py:172`.

Verified on this commit: **24 passed**, `ruff@0.9.9 format` already-formatted, `ruff check` clean. Since it rewrites the signature, it also resolves the `ruff-format` failure flagged in the other thread.

```suggestion
@pytest.mark.parametrize(
    "use_ray_v2", ["1", "0"], ids=["ray_executor_v2", "ray_executor_v1"]
)
def test_multi_gpu_fp8_patches_before_model_load(fp8_module, monkeypatch, use_ray_v2):
    """Both Ray executors must receive the FP8 patches before worker/model init."""
    from vllm import envs
    from vllm.v1.executor.abstract import Executor
    from vllm.v1.executor.ray_executor import RayDistributedExecutor
    from vllm.v1.executor.ray_executor_v2 import RayExecutorV2, RayWorkerProc

    fp8 = fp8_module
    events = []
    fp8_config = fp8.FP8Config(model_parallel_size=2)
    vllm_config = types.SimpleNamespace(
        parallel_config=types.SimpleNamespace(distributed_executor_backend="ray")
    )

    # vLLM memoizes env lookups once an engine has been built in-process, which
    # would make setenv below a silent no-op and quietly test one branch twice.
    envs.disable_envs_cache()
    monkeypatch.setenv("VLLM_USE_RAY_V2_EXECUTOR_BACKEND", use_ray_v2)
    uses_v2 = use_ray_v2 == "1"
    assert envs.VLLM_USE_RAY_V2_EXECUTOR_BACKEND is uses_v2
    assert Executor.get_class(vllm_config) is (
        RayExecutorV2 if uses_v2 else RayDistributedExecutor
    )

    def fake_apply_fp8_patches(_worker, config):
        events.append(("apply_fp8_patches", config))
        fp8.fp8_patches_applied = True

    def fake_initialize_worker(_worker, *args, **kwargs):
        events.append(("initialize_worker", args))
        assert fp8.fp8_patches_applied, (
            "RayExecutorV2 started worker/model initialization before NeMo-RL "
            "installed its FP8 patches"
        )

    def fake_collective_rpc(_executor, *_args, **_kwargs):
        events.append(("collective_rpc", None))
        assert fp8.fp8_patches_applied, (
            "RayDistributedExecutor started worker/model initialization before "
            "NeMo-RL installed its FP8 patches"
        )

    monkeypatch.setattr(fp8, "apply_fp8_patches", fake_apply_fp8_patches)
    # monkey_patch_vllm_ray_executor() rebinds these by raw class assignment with
    # no cleanup of its own, so register both with monkeypatch to undo the rebind
    # even when this regression test fails.
    monkeypatch.setattr(RayWorkerProc, "initialize_worker", fake_initialize_worker)
    monkeypatch.setattr(RayDistributedExecutor, "collective_rpc", fake_collective_rpc)

    fp8.monkey_patch_vllm_ray_executor(fp8_config)

    if uses_v2:
        assert RayDistributedExecutor.collective_rpc is fake_collective_rpc, (
            "the V1 executor must be left unpatched when the V2 backend is active"
        )
        patched_initialize_worker = RayWorkerProc.initialize_worker
        # cloudpickle reconstructs nested functions with a distinct globals dict.
        worker_initialize_worker = types.FunctionType(
            patched_initialize_worker.__code__,
            patched_initialize_worker.__globals__.copy(),
            closure=patched_initialize_worker.__closure__,
        )
        worker_initialize_worker(object(), 0, {})
        worker_initialize_worker(object(), 0, {})

        assert events == [
            ("apply_fp8_patches", fp8_config),
            ("initialize_worker", (0, {})),
            ("initialize_worker", (0, {})),
        ]
    else:
        assert RayWorkerProc.initialize_worker is fake_initialize_worker, (
            "the V2 worker hook must be left unpatched when the V1 backend is active"
        )

        # execute_method(fn, cfg) ends up calling fn(worker, cfg) upstream, so pass
        # the worker through rather than None to mirror apply_fp8_patches(self, cfg).
        def make_worker():
            worker = types.SimpleNamespace()

            def fake_execute_method_remote(fn, config):
                fn(worker, config)
                return object()

            worker.execute_method = types.SimpleNamespace(
                remote=fake_execute_method_remote
            )
            return worker

        monkeypatch.setattr(fp8, "ray", types.SimpleNamespace(get=lambda _future: None))
        executor = types.SimpleNamespace(workers=[make_worker()])
        RayDistributedExecutor.collective_rpc(executor, "init_device")
        RayDistributedExecutor.collective_rpc(executor, "init_device")

        assert events == [
            ("apply_fp8_patches", fp8_config),
            ("collective_rpc", None),
            ("collective_rpc", None),
        ]
```

If you would rather not take the rewrite, the formatting fix stands alone and is worth doing either way — `uv run ruff format tests/unit/models/generation/test_vllm_fp8_quantization.py` collapses the 3-line signature onto one line. `lint-check` runs `pre-commit run --all-files`, so it will go red as soon as CI is unblocked.

examples/configs/recipes/llm/grpo-qwen3-8b-base-1n8g-fp8-kvcache-megatron.yaml
zpqiu: [`grpo-qwen3-8b-base-1n8g-fp8-kvcache-megatron.yaml:36`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/examples/configs/recipes/llm/grpo-qwen3-8b-base-1n8g-fp8-kvcache-megatron.yaml#L36)

Adding generation TP here is what puts this recipe on the new V2 arm, which makes it the only nightly covering it. That surfaces a question about two sibling recipes rather than anything wrong with this line.

Both still pin the legacy executor, added together in 43119293 (#2744) with no explanatory comment:
- [`grpo-qwen3.5-35ba3b-2n8g-megatron-ep16tp2-fp8.yaml:40`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/examples/configs/recipes/llm/grpo-qwen3.5-35ba3b-2n8g-megatron-ep16tp2-fp8.yaml#L40) — `precision: fp8` at `tensor_parallel_size: 2`, i.e. exactly the FP8 + TP>1 + Ray combination #3755 describes. This pin is load-bearing: it is currently the only thing exercising the V1 branch.
- [`grpo-qwen3.5-9b-1n8g-megatron-fp8.yaml:43`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/examples/configs/recipes/llm/grpo-qwen3.5-9b-1n8g-megatron-fp8.yaml#L43) — this one resolves to `tensor_parallel_size: 1` (the leaf overrides the parent's 4), so `model_parallel_size == 1` and it takes the single-GPU branch, never reaching either executor. The pin is inert here.

Were these workarounds for the bug this PR fixes? If so, is unpinning them in scope, or better as a follow-up? Either answer is fine — a one-line comment recording the intent would stop them reading as cargo cult later. We could not tell from the history, hence asking rather than assuming.

zpqiu (commented):
Re-review of `ff255743` by a team of six agents. **The fix is correct and the right shape — no bugs found.**

We verified the mechanism six independent ways rather than reasoning about it: real `cloudpickle` round-trips, `_extract_code_globals` on the patched function returning `[]` (so there is nothing left for cloudpickle to freeze — it is complete *by construction*, not just correct in practice), a live Ray actor run, an `importlib` A/B against the pre-fix revision, a harness reproducing `RayExecutorV2._init_executor`, and an 11-mutant matrix. Zero contradictions. `ff255743` genuinely closes the cloudpickle-globals issue from the last round.

**Worth stating because it is invisible from the diff: this repairs more than the recipe it edits.** Five recipes already on `main` run FP8 with `tensor_parallel_size > 1` and no `VLLM_USE_RAY_V2_EXECUTOR_BACKEND` override, so they were silently on the unpatched V2 arm:

| recipe | TP | suite |
|---|---|---|
| `grpo-minimax-m27-dapo-8n8g-automodel` | 8 | release |
| `performance/grpo-deepseek-v3-64n8g-fp8-async-1off` | 16 | performance |
| `performance/grpo-qwen3-235b-16n4g-mxfp8-rollout` | 4 | performance_gb200 |
| `performance/grpo-qwen3-235b-32n4g-async-1off-mxfp8-rollout` | 4 | performance_gb200 |
| `performance/grpo-qwen3-32b-4n4g-mxfp8-rollout` | 2 | performance_gb200 |

Three of those are async. We traced the async path to the same branch by construction (`init_fp8` installs `EngineCoreProc.run_engine_core`, and `monkey_patch_vllm_ray_executor` runs in each EngineCore child with the same `model_parallel_size` and a live env read) — but to be explicit, that half is a code trace plus a measured env-timing result, **not** an end-to-end run on 16-64 node recipes.

It also answers why a config change rides along in a bugfix: on `main` this recipe had no `tensor_parallel_size` line, so it inherited 1 and sat on the single-GPU branch touching neither executor. `tensor_parallel_size: 2` is precisely what moves it onto the V2 arm — the YAML edit *is* what gives the fix nightly coverage.

The only substantive gap is unit coverage of the arm this PR did not touch. Two inline comments below, plus replies on the existing threads. Findings we considered and dropped: an executor-selection hazard (unreachable — `tp*pp > 1` forces the `"ray"` backend through the same expression that gates the FP8 branch), a suspected test state-leak (disproven empirically), and several nits already raised last round.

_Generated by Claude Code_

examples/configs/recipes/llm/grpo-qwen3-8b-base-1n8g-fp8-kvcache-megatron.yaml
zpqiu: Re-checked on `ff255743`: check-runs are still only DCO / copyright / secrets / submodule / semantic-PR / triage, and `pull-request/3754` still 404s, so the main pipeline hasn't fired. The PR also still carries no label, so even once it runs `TEST_LEVEL` resolves to `none` and the quality gate stays red. Needs a vetter's `/ok to test ff2557435e22` plus a `CI:L2` label.

On convergence — since the wandb project isn't resolvable outside NVIDIA, it's worth pasting the numbers straight into the description. They already exist: **TP=2 median `train/token_mult_prob_error` 1.105 vs TP=1 control 1.075**, both comfortably under the `< 2.0` gate, 60/60 steps. Two lines and the gated-nightly question is answerable without wandb access. (Withdrawing the wall-clock ask from the earlier comment — now that it's clear the TP bump exists to put this recipe on the V2 arm rather than to tune throughput, timing numbers aren't the right ask.)

nemo_rl/models/generation/vllm/quantization/fp8.py
zpqiu: Confirmed fixed on `ff255743` — thanks for picking this up.

We didn't take the reasoning on trust. After the fix the patched function references **zero** module globals (`_extract_code_globals` returns `[]`), so there is nothing left for cloudpickle to freeze — the function-local import at [`fp8.py#L106-L109`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/nemo_rl/models/generation/vllm/quantization/fp8.py#L106-L109) resolves through the actor's own `sys.modules` at call time. That makes it complete by construction, not just correct in practice. Independently reproduced three more ways: a real `cloudpickle` `dumps`/`loads` round-trip, a live Ray actor executing `initialize_worker` twice, and a mutation reverting only this body — which fails the test with a spurious second `apply_fp8_patches`.

The second invocation you added at [`L571-L572`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/tests/unit/models/generation/test_vllm_fp8_quantization.py#L571-L572) is what turns it into a real guard; with a single call both revisions pass.

nemo_rl/models/generation/vllm/quantization/fp8.py
zpqiu: The new comment at [`L104-L105`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/nemo_rl/models/generation/vllm/quantization/fp8.py#L104-L105) is a good addition and does document the cloudpickle-globals subtlety — though this thread was about the line nine below it: why assigning `RayWorkerProc.initialize_worker` on the *driver* ([`L113`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/nemo_rl/models/generation/vllm/quantization/fp8.py#L113)) reaches the actor processes at all.

We confirmed the mechanism against the installed Ray 2.56.1, and it turns out to be **two** things rather than one: `ray.remote()` builds a dynamic subclass whose `__qualname__` doesn't resolve, which is what forces cloudpickle to export it *by value* ([`actor.py#L2648`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L2648)); and `_inject_tracing_into_class` then copies inherited methods — including the patched one — onto that subclass via `inspect.getmembers` + `setattr` ([`tracing_helper.py#L540`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/util/tracing/tracing_helper.py#L540), [`#L600`](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/util/tracing/tracing_helper.py#L600)). Both are unconditional; the tracing flag only gates the wrapper body, so this can't silently no-op. It would stop applying if `initialize_worker` became a static/classmethod ([skipped here](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/util/tracing/tracing_helper.py#L545)).

Worth a couple of lines mainly because the file next door already sets that precedent: [`patches.py#L85-L108`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/nemo_rl/models/generation/vllm/patches.py#L85-L108) carries a `.. note::` explaining the 0.20→0.25 executor default flip, and [`#L596-L630`](https://github.com/NVIDIA-NeMo/RL/blob/ff2557435e2226c234904dedb69ac58bb261fc75/nemo_rl/models/generation/vllm/patches.py#L596-L630) states its branch rationale outright. Same package, same class of load-bearing subtlety.

zpqiu (commented):
Re-reviewed at `e3af1ddc`. **All four items from the previous round are addressed, and each verifies independently — no new findings.** Nice turnaround.

**Test rewrite.** The landed test is byte-identical to the suggested version. Full file: **24 passed**. We re-ran the mutation matrix against it rather than assuming, and it now catches **6/6**, each failing in the correct arm for the correct reason:

| mutant | v1 leg | v2 leg |
|---|---|---|
| v1 hook never installed | FAIL | pass |
| early `return` removed | pass | FAIL |
| v2 reverted to pre-`ff255743` bare globals | pass | FAIL |
| v2 hook never installed | pass | FAIL |
| v1 patches applied after the original call | FAIL | pass |
| v2 forward drops `*args/**kwargs` | pass | **FAIL** |

That last one is the one worth calling out: `original_initialize_worker(self)` is a hard production break (`local_rank` and `env_vars` are [required positionals](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/executor/ray_executor_v2.py#L139-L145)) and the *previous* version of this test did not catch it. The `("initialize_worker", args)` assertion is doing real work.

**Ray transport comment.** Verified clause-by-clause against the installed Ray 2.56.1 — accurate. For the record, the two mechanisms are independent and both load-bearing: `_modify_class` leaves `__qualname__` unresolvable so cloudpickle exports the subclass [by value](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/actor.py#L2654), and `_inject_tracing_into_class` [copies inherited methods onto it](https://github.com/ray-project/ray/blob/ray-2.56.1/python/ray/util/tracing/tracing_helper.py#L600) — unconditionally; the tracing flag gates only the wrapper body. A two-process A/B confirmed that with `_modify_class` alone the patch does *not* survive.

**Unpinning the two recipes.** This was our suggestion, so we checked it rather than assuming it was safe — and it holds up. The 35b ran 2-node on RayExecutorV2 (16 `RayWorkerProc` actors, zero `RayDistributedExecutor`), 20/20 steps, no OOM/NCCL errors/tracebacks, and cleared both of its gates: `mean(train/gen_kl_error)` **0.00224** (< 0.004) and `max(train/reward)` **0.738** (> 0.5). The 9b cleared **0.00133** (< 0.003) and **0.754**. Worth noting they did not merely squeak under the bar. Each driver records the FP8 verification baseline its gate was calibrated from — first-20-step mean `0.0024` for the 35b and `0.0014` for the 9b — and both runs came in slightly *below* those, so moving the 35b onto V2 produced no measurable convergence change on the metric it is actually gated on. The honest limit of that evidence: there is no pinned-V1 run of the same recipe at the same commit in the artifacts, so this is a match against the recorded baseline rather than a controlled V1-vs-V2 A/B. The 9b unpin is a functional no-op either way — it resolves to `tensor_parallel_size: 1`, so `model_parallel_size == 1` takes the single-GPU branch and never constructs a Ray executor.

On the coverage question we raised last time: the V1 `collective_rpc` arm now has no runtime coverage anywhere. We swept all 250 recipe YAMLs with the repo's own resolver to confirm. We think that is the right trade rather than a regression — vLLM 0.25.1 defaults the backend to V2, so the pin was keeping the flagship FP8 nightly on the *non-default* executor while the default path went unexercised. Both arms now have unit coverage.

**Lint.** `ruff@0.9.9 format --diff` → `2 files already formatted`; `check` → `All checks passed!`. The formatting failure from last round is resolved by the rewrite.

**Still outstanding (process, not code):** CI has not fired on `e3af1ddc` — check-runs are still only DCO / copyright / secrets / submodule / semantic-PR / triage, and the PR has no `CI:*` label, so `TEST_LEVEL` would resolve to `none`. Needs a vetter's `/ok to test e3af1ddc6a53` plus `CI:L2`.

_Generated by Claude Code_

examples/configs/recipes/llm/grpo-qwen3-8b-base-1n8g-fp8-kvcache-megatron.yaml
guyueh1: @zpqiu what's the motivation of adding TP=2 in this PR? i think in this benchmark the inference can comfortably fit with TP=1

zpqiu: @guyueh1  This issue was initially triggered when vLLM used TP>1. However, this regression was not detected when we bumped vLLM because we did not have any FP8 nightly tests with TP > 1. Therefore, I changed this test to cover both the FP8 linear and KV-cache patches.

We recently added a new Qwen3.5 nightly test with TP > 1, so we can restore this file to TP1 as well.

zpqiu: I have restored it to TP1.

sharonyu-115 (approved):

PR #3757, opened 2026-08-21
feat(draft): support packed context-parallel co-training

seonjinn:
# What does this PR do?

Adds packed context-parallel online co-training for DFlash and DSpark on top of #3750, including the reviewed lower-stack fixes and Qwen3-30B-A3B asymmetric attention support.

- Defines one packed layout across logical, global, CP-local, and target-SP coordinates.
- Keeps draft attention and loss within sample boundaries, including zero-owner ranks.
- Reconstructs required target-SP values with autograd-safe TP collectives.
- Normalizes draft gradients and metrics across DP x CP without gathering full CP tensors.
- Supports checkpoint, export, packed weight transfer, ModelOpt real-quant refit, and online DFlash/DSpark lifecycle paths.
- Supports asymmetric Qwen attention projections where hidden size differs from query projection size, including TP2 export/refit.

## Supported contract

- Target training: TP2, CP1/CP2/CP4, PP1.
- CP greater than 1 requires sequence packing; packed target sequence parallelism is supported.
- Generation: CP1, PP1.
- DFlash and DSpark draft bodies keep internal sequence parallelism disabled.
- Unsupported TP/PP/CP, unpacked CP greater than 1, unsupported provider routes, and DSpark vocabulary mismatches fail before training.
- Packed transport guarantees ordered framing and bounded draining for synchronous Python iterator/export/packing/unpack/load-callback failures. CUDA stream, allocation, kernel, OOM, and NCCL failures are communicator-fatal and require caller teardown.

## Review map

1. `nemo_rl/models/megatron/draft/sequence_layout.py`, `context_parallel.py`, and `sequence_parallel.py`: packed CP coordinates and collectives.
2. Draft body/provider and loss integration: DFlash/DSpark packing, Qwen asymmetric projections, TP-safe metadata, optimizer overrides, and objective validation.
3. `nemo_rl/utils/packed_tensor.py`: framed packed transfer and synchronous failure propagation.
4. `nemo_rl/modelopt/models/generation/vllm_quant_backend.py`: target-first real-quant finalization followed by co-trained draft post-load finalization.
5. Focused unit and distributed tests under `tests/unit/`.

## Stack and validation

- Exact head: `03cc0e28ef1325b1c1f7ea050db1a35e48a8018c`.
- Dependencies: #3750 `3c9218b8e18c43f34ad4ee9eb12e58ed4ef5b01c`; #3744 `3c3fc104e5525f7eccf2eabbcf50c89fc9ff2059` (both ancestors).
- All 17 finalization commits are signed and DCO-compliant; independent exact-range review is clean with 0 critical, 0 warnings, and 0 nits.
- Local final-head checks: Ruff, format, py_compile, diff-check; packed failure tests 2 passed and 4 CUDA-skipped; ModelOpt lifecycle test 1 passed; projected soft-CE focused suite 17 passed in a dependency-light Torch environment.
- Historical OCI-HSG gates: focused CP normalization/metrics `6442622` PASS; Gloo/NCCL CP1/2/4 matrix `6442673` PASS; packed online TP2xCP2 DFlash+DSpark E2E `6443086` PASS.
- Qwen3-30B-A3B matched 20-step runs: DFlash `6459668` and DSpark `6459669`, both COMPLETED 0:0 with state-dict 0/0, CUDA graph, and all 20 training steps.
- Final exact-head 2-node TP2xPP2xCP2 and expanded focused gate is pending and will be recorded here when terminal.

## Checklist

- [x] Contributor guidelines followed.
- [x] Necessary tests added or updated.
- [x] Documentation and supported/fail-closed contracts updated.
- [ ] Final exact-head OCI-HSG distributed gate terminal receipt.
- [ ] Required maintainer approval and NVIDIA CI.


nemo_rl/models/megatron/draft/dflash.py
```diff
@@ -0,0 +1,724 @@
...
+                error_code = 5
+            elif plan.query_positions.dtype != torch.int64:
+                error_code = 6
+            else:
+                error_code = len(_INPUT_ERROR_MESSAGES)
+
+        if self.tensor_parallel_size > 1:
+            if self.tp_group is None:
+                raise RuntimeError("DFlash tensor parallel group is unavailable")
+            synchronized_error = torch.tensor(
```
seonjinn: Commit 56494c30f replaced the cross-TP-rank shape agreement from the #3723 version of this validation with a MIN-reduce of the scalar error code alone -- and deleted its regression test (`test_tp2_rank_local_valid_shapes_fail_before_first_projection`, -57 lines) in the same commit, with no replacement anywhere. Ranks holding different-but-locally-valid shapes now proceed silently into the sharded projections instead of failing together. The follow-up commit restores the full-metadata all-gather (shapes + plan dimensions), the `must match across tensor-parallel ranks` error code, and the deleted regression test (adapted to the current harness).

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -121,6 +141,183 @@
...
+    mcore_generation_config = (
+        generation_config.get("mcore_generation_config")
+        if generation_config is not None
+        else None
+    )
+    generation_pipeline_size = (
+        mcore_generation_config.get("pipeline_model_parallel_size")
+        if mcore_generation_config is not None
+        else None
+    )
```
seonjinn, line 300: This check read only the raw `mcore_generation_config` override, but the runtime merges the training megatron config into the inference config -- so CP>1 training with the Megatron generation backend and an OMITTED override would silently run generation at CP>1 instead of tripping this validation. Fixed in the follow-up commit: when the generation backend is `megatron` and the override key is absent, the training PP/CP value is used as the default for this check (vLLM generation does not inherit megatron parallelism, so the default is backend-gated and the supported CP2-training + vLLM-generation recipes are unaffected).

seonjinn (commented):
Self-review before requesting reviewers -- the widest pass in this series: eight review agents mapped to the design doc's PR11 review allocation (packed-coordinate mappings, loss/normalization algebra, collective ordering/SP, support envelope, cross-PR composition, tests, line-level, CI/hygiene), then two independent adversarial verifiers (math lens and protocol lens) that re-opened every citation, re-derived the algebra, and executed empirical checks.

**What survived is above (5 inline, all fixed in the follow-up commit).** Equally important is what did NOT survive, recorded so human reviewers need not re-derive it:

- The scariest candidate -- "eager `train()` reduces draft counts over DP-only while the deferred path uses DP x CP" -- is real asymmetry but UNREACHABLE: `_validate_draft_training_entrypoint` rejects draft+CP>1 on the sync path (its docstring anticipates exactly this hazard), and at CP=1 the groups coincide. The deferred path's normalization algebra was re-derived independently and is CP-invariant.
- "Eagle exempt from draft validation" -- eagle3 cannot take the split API (`supports_context_parallel=False`) and the train() fallback raises on all ranks; the residue is only that eagle3+target-SP is unenforced (pre-existing).
- Zero-owner CP ranks: verified EMPIRICALLY that backward traverses the zero-element graph via the deliberate `gathered_dependency` edge -- collective counts stay symmetric.
- The packed coordinate systems (logical/global/CP-local, owner map, labels, RoPE frames) were hand-traced on boundary examples and re-executed: sample-boundary isolation is structural, the zigzag split matches Megatron's `get_batch_on_this_cp_rank` convention (a pinning unit test is a cheap follow-up worth adding), and CP1 degrades to exact identity.
- `draft_sample_ids` dual-derivation concern: refuted -- both entrypoints stamp via the same helper.

**Process, before undrafting:**

1. **Restack.** Every dependency branch carries review fixes this tree predates: 3701 `8774f1d3d`, 3703 `895e0c212`, 3706 `76c2648d7`, 3707 `f658f69b9`, 3711 `54a3d9b4e`+, 3715 `7700a0774`, 3723 `abc74d779`+, 3726 `42cefa811`, 3744 `8a03385f2`, 3750 `f85be2ef2` (current drift vs dep tips: 1/1/14/18 commits for 3744/3750/3711/3723). The embedded copies in this tree lack all of them -- notably the dspark TP-attribute and checkpoint-group fixes and the eagle-ttt recompute guard class.
2. No substantive CI has ever run on this PR (fork gate; no `/ok to test` in history). Issue it on the post-restack head.
3. Follow-up tests worth scheduling: real 2-process CP test of `_finish_train_step_body`; a zigzag-vs-`get_batch_on_this_cp_rank` cross-check; real-distributed zero-owner coverage for the DSpark loss path (currently DFlash-attention-only).

nemo_rl/models/megatron/draft/diagnostics.py
```diff
@@ -0,0 +1,132 @@
...
+    delta = abs(after[0] - probe.before[0]) + abs(after[1] - probe.before[1])
+    return DraftUpdateResult(
+        before=probe.before,
+        after=after,
+        grad_l2=probe.grad_l2,
+        checksum_delta=delta,
+    )
+
+
+def require_draft_update(result: DraftUpdateResult) -> None:
```
seonjinn, line 120: Under packed context parallelism a microbatch can legitimately contain zero eligible draft windows (block-plan eligibility at CP/sample boundaries), which makes `grad_l2 == 0` a consistent no-op rather than a failure -- but this raised unconditionally, so any run with `update_probe_enabled=true` would crash on the first such microbatch. The follow-up commit distinguishes consistent no-ops (zero grad AND unchanged params: allowed) from genuinely inconsistent evidence (a parameter change without a gradient, or a gradient without a change: still raised), with tests for all four quadrants (verified empirically on CPU).

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -857,12 +1084,29 @@ def train(
...
+                if draft_provider is not None:
+                    draft_normalization_counts = draft_provider.normalization_counts(
+                        batch,
+                        optimizer_step=int(self.scheduler.num_steps),
+                    )
+                    if draft_normalization_counts is not None:
+                        draft_normalization_counts = (
+                            _all_reduce_draft_normalization_counts(
+                                draft_normalization_counts,
+                                group=parallel_state.get_data_parallel_group(),
```
seonjinn: Defensive only: this eager-path reduction used `get_data_parallel_group()` (DP-only) while the deferred path's identical operation uses `with_context_parallel=True` (DP x CP). Adversarial verification showed the asymmetry is currently unreachable -- `_validate_draft_training_entrypoint` rejects draft+CP>1 on this path, and at CP=1 the two groups are identical -- so this was never a live wrong gradient. The follow-up commit aligns the group anyway (plus a comment), so a future change to the entrypoint gate cannot silently shrink this denominator.

tests/unit/models/policy/test_megatron_split_state.py
```diff
@@ -528,6 +560,283 @@ def test_picks_global_valid_seqs_for_sequence_level_loss(self, mock_module_symbo
...
+    @pytest.mark.parametrize(
+        ("cp_numerators", "cp_counts"),
+        [
+            ([40.0], [10.0]),
+            ([0.0, 40.0], [0.0, 10.0]),
+            ([0.0, 10.0, 0.0, 30.0], [0.0, 2.0, 0.0, 8.0]),
+        ],
+        ids=["cp1", "cp2_zero_owner", "cp4_uneven_zero_owners"],
+    )
+    def test_draft_metric_is_cp_invariant_before_replica_leader_filtering(
```
seonjinn: This test's "CP invariant" name overclaimed: the `all_reduce` is mocked to inject the expected global counts, so it pins the normalization WIRING (global counts reach the metric before replica-leader filtering), not true CP invariance. Renamed and documented in the follow-up commit; a real 2-process gloo test of `_finish_train_step_body` under CP is the follow-up that would close it (also covering the split x CP2/4 path the docs mandate).

seonjinn:
Re-review at head 443e7243a (previous review was at 815b7834d; delta = the restack merges plus +236 test lines):

**Delta verdict — both remaining test asks are genuinely closed.** `test_megatron_split_cp_distributed.py` is a real `mp.spawn` multi-process test of the production `_finish_train_step_body` (only `parallel_state` accessors are pointed at manually built real groups; the collectives themselves run), parametrized CP1/2/4 with zero-owner ranks, and its hardcoded expected value was independently re-derived by hand from the `step_state` formula — it matches exactly and cancels `world_size`, so it is a true oracle, not an impl mirror. This closes the earlier mock-stipulated-invariance finding. The new `test_draft_refit` case injects a payload-materialization failure on a non-owner rank only and verifies fingerprint-matched errors on every PP-lane rank with zero wasted broadcasts — the exact between-control-and-payload window flagged before.

**Composition check:** all five fixes from the previous review survived the restack merges (verified by marker), and no conflict artifacts remain. One item still open: the restack brought in the #3744/#3750 fix commits, but the embedded copies of the PR1–8 dependency content still predate their review fixes (spot-checked: `dspark.py` TP attributes, `draft.py` bin_ids range check, `draft_config.py` typo validator — all absent). Also note two fixes landed on dependency branches after this restack: #3711 `1deb4beba` (consumer-gated embedding TP reduction) and #3750 `b4f7744da` (preflight inside the fail-closed guard) — the next restack should pick up both plus the PR1–8 fix commits.

nemo_rl/models/megatron/data.py
```diff
@@ -36,6 +40,65 @@
...
+    seq_length_key: Optional[str],
+    processed_inputs: "ProcessedInputs",
+    delegate_pack_to_model: bool,
+) -> Optional["DraftSequenceLayout"]:
+    if not cfg["sequence_packing"]["enabled"] or "draft_sample_ids" not in data_dict:
+        return None
+    if delegate_pack_to_model:
+        raise NotImplementedError(
+            "draft sequence layout requires canonical NeMo-RL sequence packing"
+        )
```
isomap, line 94: TQWorkerMixin.train_presharded() adds draft_sample_ids even when draft training is disabled. With sequence packing enabled, an ordinary TQ batch for a self-packing model such as Qwen3-VL therefore reaches this exception through train_presharded → train → get_microbatch_iterator.

Please gate draft-layout construction on an enabled provider that requires it, rather than treating ID presence as proof that draft training is active. Add a regression for draft-disabled TQ training with model-owned packing.

nemo_rl/algorithms/grpo.py
```diff
@@ -3171,6 +3220,7 @@ def grpo_train(
                             "sample_mask": repeated_batch["loss_multiplier"],
                         }
                     )
+                    _preserve_draft_sample_ids(train_data, repeated_batch)
```
isomap, line 3491: The synchronous NeMo Gym path reconstructs final_batch without draft_sample_ids and replaces repeated_batch with that result. This preservation call therefore has no IDs to copy. With DFlash/DSpark and data_plane.enabled=false, prepare_batch() subsequently fails with “training requires stable plan inputs: draft_sample_ids”.

Please carry the IDs through Gym postprocessing in the corresponding restored row order, or reject this combination before rollout if unsupported. Add a regression through actual sync rollout, training-data construction, and both providers' prepare_batch().

nemo_rl/models/megatron/data.py
seonjinn: Fixed in 87434d3. `_get_packed_draft_sequence_layout` no longer treats ID presence as proof that draft training is on. It now gates on a new `_draft_training_enabled(cfg)` helper that reads `cfg["draft"].enabled` (accepting both the pydantic config object and a plain dict, since the draft config is a `TypedDict` on some of the sibling branches) and returns `False` when the key is absent. The `NotImplementedError` for `delegate_pack_to_model` is therefore only reachable when a provider is actually enabled, and an ordinary TQ batch for a self-packing model passes straight through.

The helper's docstring records why the old test was wrong: `TQWorkerMixin.train_presharded` attaches `draft_sample_ids` to every presharded batch.

Regression added in `tests/unit/models/megatron/test_megatron_data.py::TestMakeProcessedMicrobatchIterator::test_make_processed_microbatch_iterator_skips_draft_layout_when_draft_disabled`. It runs with `delegate_pack_to_model=True`, `sequence_packing.enabled=True` and `draft_sample_ids` present, parametrized over the three draft-disabled shapes (`draft` key absent, provider object with `enabled=False`, dict with `enabled=False`), and asserts the layout is `None` and `_build_draft_sequence_layout` is never called.

nemo_rl/algorithms/grpo.py
seonjinn: Fixed in 87434d3, with one caveat on the test scope that I want to be explicit about.

The sync Gym path now preserves the IDs instead of dropping them: `grpo_train` binds `nemo_gym_rollout_result.final_batch` to a local, calls `_preserve_draft_sample_ids(gym_final_batch, repeated_batch)` and only then rebinds `repeated_batch`. A comment records the assumption that makes this positional copy legal - Gym rebuilds the batch from its own result rows but `run_nemo_gym_rollout_sync` restores input row order, so row `i` of the rebuilt batch is row `i` of the input.

Because that assumption is load-bearing, `_preserve_draft_sample_ids` no longer copies blindly. It now raises `ValueError` when `target.size != len(sample_ids)`, so a postprocessing change that drops or reorders rows fails loudly instead of silently labelling the wrong rows.

Caveat: I did not add the end-to-end regression you asked for - the requested path through an actual sync rollout, training-data construction and both providers' `prepare_batch()` is not covered. What I added in `tests/unit/algorithms/test_grpo.py` exercises the preservation helper directly:

- `test_preserve_draft_sample_ids_survives_nemo_gym_batch_rebuild` builds a Gym-style rebuilt batch that keeps only the columns Gym models, and asserts the IDs come through in the original row order.
- `test_preserve_draft_sample_ids_rejects_row_count_mismatch` asserts the new guard fires.

So the reported failure mode is fixed and the invariant is pinned, but the integration-level coverage is still missing. If you would rather block on the full end-to-end test I can add it; it needs a sync-rollout fixture that does not exist in the unit suite today.

seonjinn: Added the requested sync NeMo Gym regression. It runs one `grpo_train` iteration for both DFlash and DSpark, verifies `draft_sample_ids` survive the Gym batch rebuild and training-data construction, and invokes each real provider's `prepare_batch`. I also fixed NCCL-reshard draft finalization so draft payloads initialize the runtime manifest and run post-load processing.

Issue #3695, opened 2026-08-18
[Async PPO] cleanup tracking issue


yuki-97:
Tracking issue for loose ends in Async PPO, introduced in #3410.
Items will be checked off or split into separate issues as they land.

### Known missing items

- [ ] **The PPO branch of `AsyncTrajectoryCollector.__init__` hard-codes three settings that GRPO reads from config.** [`trajectory_collector.py:111-113`](https://github.com/NVIDIA-NeMo/RL/blob/bac4a8c46ee14d4c3dfb774be9577bacbb92ccf5/nemo_rl/algorithms/async_utils/trajectory_collector.py#L111-L113) sets `_deduplicate_multimodal_data = False`, `_debug_payload_metrics = False`, `_max_generation_failures = 0`, while the GRPO branch ten lines above reads all three from `GRPOConfig.deduplicate_multimodal_data`, `GRPOConfig.debug_payload_metrics`, and `AsyncGRPOConfig.max_generation_failures`. `PPOConfig` / `AsyncPPOConfig` carry no equivalent fields, so none of the three is reachable from a recipe. The `max_generation_failures = 0` one has runtime consequences: async PPO aborts on the first generation-worker exception with no way to opt into tolerating transient failures. Add the fields to `PPOConfig` / `AsyncPPOConfig` (mirroring the GRPO names and YAML docs) and read them here. (Flagged during #3410 review.)

- [ ] **Wire up (or keep fail-loud) the features `_validate_async_ppo_config` currently rejects.** [`run_ppo.py:48-88`](https://github.com/NVIDIA-NeMo/RL/blob/bac4a8c46ee14d4c3dfb774be9577bacbb92ccf5/examples/run_ppo.py#L48-L88) raises `NotImplementedError` for each of:
  - `ppo.max_num_epochs != -1` — async PPO has no epoch limit; `max_num_steps` is the only length control. Same gap async GRPO has in #3175.
  - NeMo-Gym rollout — the shared collector supports gym on the GRPO path, so this is a wiring gap rather than a missing capability.
  - FP8 KV-scale synchronization — same gap tracked for the Single Controller in #2625; the driver-side compute + forward is what's missing.
  - Dynamic sampling, reward scaling, reward shaping, multiple dataloaders.

Add anything else here as it comes up.

tianyi-zhang-02:
Took the first item — #3760.

[...]

PR #3760, opened 2026-08-21
fix(ppo): let async PPO reach the three settings the collector hard-coded

tianyi-zhang-02:
## What does this PR do?

Lets async PPO configure three collector settings that were hard-coded on its branch:

- `deduplicate_multimodal_data`
- `debug_payload_metrics`
- `async_ppo.max_generation_failures`

Defaults stay unchanged (`false`, `false`, and `0`). A zero failure budget still fails on the first generation-worker exception; a positive value now permits the requested number of failures.

This is the first item in #3695. Async PPO already requires a non-colocated asynchronous vLLM backend, so the refresh keeps the configuration path small and does not add a second backend guard.

## Validation

Current head `afa5a3f2f39fce2e3b3c3e16edb6c8f92130e42c`, rebased onto upstream `main` at `90a2a212d503455d8590be5c8de3cb989d3425b0`.

- full `tests/unit/algorithms/test_async_utils.py`: **117 passed**
- covers PPO config selection, payload/debug/failure forwarding, Gym and non-Gym batching, failure thresholds, replay/refit behavior, and collector telemetry
- pytest 9.1.1 with the repository-locked `pytest-asyncio` 1.4.0

The suite runs with a local CPU Ray cluster; no GPU workload is required for these configuration and collector regressions.

### DIFF
--- a/examples/configs/ppo_math_1B.yaml
+++ b/examples/configs/ppo_math_1B.yaml
@@ -34,6 +34,10 @@ ppo:
     max_trajectory_age_steps: 1
     # Requires policy_training_start_step > 0. null uses max_trajectory_age_steps.
     warmup_generation_lead_steps: null
+    # Number of generation-worker failures tolerated before aborting.
+    # 0 (default) = fail on the first worker exception. Increase only
+    # when transient generation errors are expected and acceptable to drop.
+    max_generation_failures: 0
     in_flight_weight_updates: false
     recompute_kv_cache_after_weight_updates: false
     # true discards partial restored rows and fills from subsequent prompts.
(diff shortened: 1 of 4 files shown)


tianyi-zhang-02:
One asymmetry worth pointing at before it comes up in review: the GRPO branch asserts `async_grpo is not None` and the PPO branch has no equivalent guard. That is deliberate — `GRPOConfig.async_grpo` is typed `AsyncGRPOConfig | None`, while `PPOConfig.async_ppo` is `AsyncPPOConfig` with a `default_factory`, so it can never be None. Reading `async_config.max_generation_failures` on the PPO branch needs no guard.

Also merged current main in :)

PR #3761, opened 2026-08-21
test(grpo): cover zero-age startup barrier

jthomson04:
## Summary

- rebase onto current main after upstream PR #3435 superseded the original production change
- cover max_trajectory_age_steps == 0 in the existing startup-pipeline helper test
- verify startup does not query an impossible lookahead target when no lookahead window can exist

This test-only diff removes the uncovered production line that caused the previous Codecov patch failure.

## Testing

- pytest -q tests/unit/algorithms/test_grpo.py -k startup_pipeline_ready (7 passed)
- ruff check tests/unit/algorithms/test_grpo.py
- ruff format --check tests/unit/algorithms/test_grpo.py
- git diff --check

### DIFF
--- a/tests/unit/algorithms/test_grpo.py
+++ b/tests/unit/algorithms/test_grpo.py
@@ -2028,17 +2028,21 @@ def test_should_use_nemo_gym_requires_dynamo_token_wrapper() -> None:
         "generating_targets",
         "current_step_ready",
         "step",
+        "max_trajectory_age_steps",
         "max_num_steps",
         "expected",
         "expected_queries",
     ),
     [
-        pytest.param(True, [], True, 3, 10, True, 1, id="complete"),
-        pytest.param(False, [4], True, 3, 10, True, 1, id="reserved-only"),
-        pytest.param(False, [], True, 3, 10, False, 1, id="neither"),
-        pytest.param(False, [5, 6], True, 3, 10, False, 1, id="other-target-reserved"),
-        pytest.param(False, [], True, 9, 10, True, 0, id="last-step"),
-        pytest.param(False, [], False, 3, 10, False, 0, id="current-not-ready"),
+        pytest.param(True, [], True, 3, 1, 10, True, 1, id="complete"),
+        pytest.param(False, [4], True, 3, 1, 10, True, 1, id="reserved-only"),
+        pytest.param(False, [], True, 3, 1, 10, False, 1, id="neither"),
+        pytest.param(
+            False, [5, 6], True, 3, 1, 10, False, 1, id="other-target-reserved"
+        ),
+        pytest.param(False, [], True, 3, 0, 10, True, 0, id="no-lookahead-window"),
+        pytest.param(False, [], True, 9, 1, 10, True, 0, id="last-step"),
+        pytest.param(False, [], False, 3, 1, 10, False, 0, id="current-not-ready"),
     ],
 )
 def test_startup_pipeline_ready(
@@ -2046,11 +2050,12 @@ def test_startup_pipeline_ready(
     generating_targets,
     current_step_ready,
     step,
+    max_trajectory_age_steps,
     max_num_steps,
     expected,
     expected_queries,
 ):
-    """Startup needs a complete or claimed lookahead except at the last step."""
+    """Startup needs a complete or claimed lookahead only when one can exist."""
     replay_buffer = MagicMock()
     replay_buffer.has_complete_batch.remote.return_value = lookahead_complete
     collector_status = {"generating_targets": generating_targets}
@@ -2062,7 +2067,7 @@ def test_startup_pipeline_ready(
             current_step_ready=current_step_ready,
             step=step,
             num_prompts_per_step=8,
-            max_trajectory_age_steps=1,
+            max_trajectory_age_steps=max_trajectory_age_steps,
             max_num_steps=max_num_steps,
         )
 


tests/unit/algorithms/test_grpo.py
```diff
@@ -41,6 +41,7 @@
     _get_grpo_save_state,
     _initial_grpo_save_state,
     _initial_policy_generation_stale,
+    _needs_initial_lookahead_barrier,
     _needs_hf_refit_handshake,
```
jthomson04: [`tests/unit/algorithms/test_grpo.py:44`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/tests/unit/algorithms/test_grpo.py#L44)

This breaks the isort pre-commit hook, so the lint job will go red. `pyproject.toml` only selects `["D", "F"]`, which is why the PR's own `ruff check` comes back clean — but [`.pre-commit-config.yaml:15-16`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/.pre-commit-config.yaml#L15-L16) runs a **second** ruff pass:

```yaml
    - id: ruff
      args: ["check", "--select", "I", "--fix"]
```

and CI runs [`pre-commit run --all-files --show-diff-on-failure`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/.github/workflows/cicd-main.yml#L387). Confirmed with the pinned version — `uvx ruff@0.9.9 check --select I --diff` on this file requests exactly one change, this swap (`_needs_h…` sorts before `_needs_i…`):

```suggestion
    _needs_hf_refit_handshake,
    _needs_initial_lookahead_barrier,
```

tests/unit/algorithms/test_grpo.py
```diff
@@ -115,6 +116,23 @@ def test_save_async_replay_buffer_checkpoint(tmp_path):
...
+    }
+
+    assert not _needs_initial_lookahead_barrier(replay_buffer_restored=False, **args)
+    assert _needs_initial_lookahead_barrier(replay_buffer_restored=True, **args)
+    assert not _needs_initial_lookahead_barrier(
+        replay_buffer_restored=True,
+        max_trajectory_age_steps=1,
+        step=3,
+        max_num_steps=4,
+    )
```
jthomson04: [`tests/unit/algorithms/test_grpo.py:119`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/tests/unit/algorithms/test_grpo.py#L119)

The test doesn't pin the whole predicate. I mutation-tested it, and **two independent mutations survive**:

- deleting `and max_trajectory_age_steps > 0` from the helper entirely
- changing `step + 1 < max_num_steps` to `step + 2 < max_num_steps`

The `max_age == 0` case is reachable and semantically load-bearing, not a can't-happen input: [`AsyncGRPOConfig.max_trajectory_age_steps`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/grpo.py#L189) is a plain `int = 1` with no `ge=` constraint (contrast [`AsyncPPOConfig`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/ppo.py#L122), which uses `Field(default=1, ge=1)`). With age 0, `_is_valid_for_target` requires `version == target`, so `step + 1` genuinely cannot exist before the first refit — blocking on it would *create* a deadlock rather than prevent one. That's exactly why the conjunct is there, and nothing currently holds it in place.

One parametrize covers both gaps and kills both mutations. Verified passing locally (5 passed), and clean under `ruff@0.9.9 format` + `--select I`. Renaming also matches the sibling convention (`test_needs_hf_refit_handshake`) and fixes the current name being narrower than what it asserts:

```suggestion
@pytest.mark.parametrize(
    "replay_buffer_restored, max_trajectory_age_steps, step, max_num_steps, expected",
    [
        # A fresh collector owns both `step` and `step + 1`, so startup can
        # enter training without waiting for the lookahead to fill.
        (False, 1, 0, 4, False),
        (True, 1, 0, 4, True),
        # max_age 0 requires a trajectory's weight version to equal its target,
        # so `step + 1` can only be generated after the first refit -- blocking
        # on it would create the deadlock instead of preventing one.
        (True, 0, 0, 4, False),
        # The lookahead only exists while `step + 1` is still a training step.
        (True, 1, 2, 4, True),
        (True, 1, 3, 4, False),
    ],
)
def test_needs_initial_lookahead_barrier(
    replay_buffer_restored: bool,
    max_trajectory_age_steps: int,
    step: int,
    max_num_steps: int,
    expected: bool,
) -> None:
    assert (
        _needs_initial_lookahead_barrier(
            replay_buffer_restored=replay_buffer_restored,
            max_trajectory_age_steps=max_trajectory_age_steps,
            step=step,
            max_num_steps=max_num_steps,
        )
        is expected
    )
```

nemo_rl/algorithms/grpo.py
```diff
@@ -4446,15 +4461,20 @@ def async_grpo_train(
...
         if current_step_ready:
-            # Keep the async pipeline one step ahead before entering training.
-
             # A restored buffer may already contain `step`, allowing startup to
             # consume it before the collector generates `step + 1`. After refit,
             # the collector advances to targets starting at `step + 2`, leaving
             # `step + 1` permanently missing. Wait for the initial collector,
             # whose range includes both steps, to complete the lookahead first.
+            # A fresh collector already owns both targets, so it can continue
+            # the lookahead while the first training step runs.
```
jthomson04: [`nemo_rl/algorithms/grpo.py:4469`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/grpo.py#L4469)

The added sentence gives a reason that applies equally to the case that *does* wait, so it doesn't explain the split. [`initial_weight_version = start_step`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/async_utils/trajectory_collector.py#L156-L159) for restored runs too, and [`_calculate_target_weights`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/async_utils/trajectory_collector.py#L208-L217) returns the inclusive window `[step .. step+lead]` whenever `current_weight_version == initial_weight_version` — so a *restored* collector also "already owns both targets."

The real discriminator is arrival timing, not window membership: a restored buffer can satisfy `has_complete_batch(step)` at wait-iteration 0, before the collector has reserved anything. Since this comment is the only place the correctness argument for a deadlock fix is written down, it's worth making it the distinguishing one:

```suggestion
            # A restored buffer can already contain `step` at the first wait
            # iteration, letting startup break before the collector has
            # reserved `step + 1`. After refit the collector advances to
            # targets starting at `step + 2`, leaving `step + 1` permanently
            # missing, so wait for the initial collector to fill it first.
            # A fresh run's buffer starts empty, so the loop cannot break
            # until the collector has produced `step` -- by which point it has
            # already reserved `step + 1`, the same slack the pipeline relies
            # on at every subsequent step.
```

nemo_rl/algorithms/grpo.py
```diff
@@ -4446,15 +4461,20 @@ def async_grpo_train(
...
             max_num_steps = master_config.grpo.max_num_steps
-            need_lookahead = max_trajectory_age_steps > 0 and step + 1 < max_num_steps
+            need_lookahead = _needs_initial_lookahead_barrier(
+                replay_buffer_restored=replay_buffer_restore_metadata is not None,
```
jthomson04: [`nemo_rl/algorithms/grpo.py:4473`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/grpo.py#L4473)

Two small ripple effects from making the barrier conditional here:

1. **#3066's regression-test docstring is now stale.** [`test_async_utils.py:1223-1224`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/tests/unit/algorithms/test_async_utils.py#L1223-L1224) still states the invariant unconditionally:
   > The fix is a startup pipeline barrier: before breaking, also require `has_complete_batch(N+1)` to be True (or `N+1 >= max_steps`).

   That now holds only for restored buffers. The test itself still passes (it asserts buffer readiness values and does simulate a restore), but it's the canonical prose explanation of the barrier — worth adding "for a restored buffer," ideally pointing at `_needs_initial_lookahead_barrier`.

2. **Optional:** this expression is ~200 lines from where the metadata is produced ([grpo.py:4259](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/grpo.py#L4259)), and nothing tests the wiring — the unit test covers the pure predicate only. Computing `replay_buffer_restored = replay_buffer_restore_metadata is not None` as a named local at the restore site would make it locally verifiable by reading. (An integration test through `mock_async_grpo_infrastructure` would also catch it, but that needs ~10 lines of surgery on a fixture shared by 12 tests, which isn't proportionate here.)

For what it's worth, `is not None` is the right check and worth keeping deliberately: [`load_from_path`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/async_utils/replay_buffer.py#L365-L391) always returns a populated dict and raises on a malformed checkpoint, so a truthiness check would be a latent bug.

nemo_rl/algorithms/grpo.py
```diff
@@ -350,6 +350,21 @@ def _get_grpo_save_state(
     return GRPOSaveState(**state_values)
 
 
+def _needs_initial_lookahead_barrier(
```
jthomson04, line 410: [`nemo_rl/algorithms/grpo.py:353`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/grpo.py#L353)

Non-blocking question: async PPO has the structurally identical barrier and now disagrees with GRPO about when it applies. [`ppo.py:2344-2346`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/ppo.py#L2344-L2346) still asserts the invariant this PR relaxes —

```python
# The initial collector is the only window that can generate
# both `step` and `step + 1`. Fill both before the first refit.
need_lookahead = step + 1 < max_training_steps
```

— over the same `AsyncTrajectoryCollector` and `ReplayBuffer`, and async PPO [also restores the replay buffer](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/ppo.py#L2232-L2249). It currently has no way to tell, since it discards `load_from_path`'s return value.

Since the new predicate is already a free function over scalars, PPO could share it. Narrowing PPO here would be scope expansion though — a follow-up issue, or one line noting that PPO stays conservative deliberately, would be enough to keep the next reader from reading the omission as an oversight.

(To be clear, PPO's *lack* of a `max_trajectory_age_steps > 0` guard is correct and not a gap: [`AsyncPPOConfig`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/ppo.py#L122) constrains it with `Field(default=1, ge=1)`, so PPO structurally cannot reach 0, whereas [GRPO's](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/grpo.py#L189) has no such constraint.)

jthomson04 (commented):
Reviewed by a team of five agents (RL-codebase expert, bug-finder, test-agent, devil's advocate, plus collation). Findings were adversarially challenged before posting — 4 candidate findings were dropped as noise, and one was dropped as factually wrong.

**The core change looks correct**, and I traced it rather than taking the premise on faith. [`_calculate_target_weights`](https://github.com/NVIDIA-NeMo/RL/blob/510f51575802dd0d16ec7a88c65ec279a6a38242/nemo_rl/algorithms/async_utils/trajectory_collector.py#L208-L217) returns `[v+1 .. v+lead]` at every non-initial weight version, and `weight_version` only advances at refit — so target `v+1` is *always* generated concurrently with training step `v`. A fresh run's step 0 therefore has exactly the same one-training-step of slack that every subsequent step already relies on; if that were unsafe, training would stall at every step, not just the first. The restored case is genuinely different: the buffer arrives pre-filled, so the wait loop can break at iteration 0 before the collector has reserved anything. The asymmetry the PR encodes is real.

One item turns CI red (import ordering); the rest are suggestions.

Verified locally: the new test passes, and the two `has_complete_batch` / config facts cited below were checked against source. Notably, **two independent mutations to the new predicate survive the added test** — details inline.

_Generated by Claude Code_

tests/unit/algorithms/test_grpo.py
jthomson04: Fixed in 8646691b6. The imports now place _needs_hf_refit_handshake before _needs_initial_lookahead_barrier. The explicit Ruff import-order check passed on all four changed files.

nemo_rl/algorithms/grpo.py
jthomson04: Fixed in 8646691b6. The regression-test docstring now scopes the barrier to restored buffers and names _needs_initial_lookahead_barrier. I also compute replay_buffer_restored once near restore handling with the deliberate is not None check. The regression test passed.

nemo_rl/algorithms/grpo.py
jthomson04: Kept PPO behavior unchanged in 8646691b6. I added one line at the PPO barrier to state that PPO deliberately keeps the conservative wait for fresh and restored buffers.

nemo_rl/algorithms/grpo.py
jthomson04: Fixed in 8646691b6. The comment now explains the timing difference: a restored buffer can satisfy the current step before the collector reserves the next target, while a fresh buffer cannot exit the wait until collection has started.

jthomson04:
/ok to test [8646691](https://github.com/NVIDIA-NeMo/RL/pull/3761/commits/8646691b6afb88c45d19b1446bca9d494b6657d8)

jthomson04:
/ok to test [16c92b2](https://github.com/NVIDIA-NeMo/RL/commit/16c92b2d664567580527a1be9630d0c934d6c860)

PR #3762, opened 2026-08-21
feat(megatron): expose mcore inference async_sched_mode and cover it

shanmugamr1992:
## Summary

`async_sched_mode` is already plumbed through `policy.generation.mcore_generation_config` into the dynamic inference engine's `InferenceConfig` as of #3920, so this PR no longer adds that wiring. What is left:

* Document the key on `MCoreGenerationSpecificArgs`, including the fact that mcore drops to a *non-overlapped async* ordering (not legacy) for steps it cannot overlap.
* Default the shared `grpo_math_1B.yaml` exemplar to `async`, so the existing colocated and non-colocated Megatron generation smoke tests exercise async scheduling. This is a behaviour change for recipes inheriting the exemplar.
* Report the engine's cumulative async-scheduler step count from `finish_generation`, and require it to be positive in both smoke tests. This confirms async mode reached and ran in the engine; the counter includes steps that fall back to the non-overlapped async ordering and does not quantify overlap.
* Reuse the existing smoke tests instead of adding dedicated ten-step jobs, keeping L1 runtime unchanged. The non-colocated test also covers scheduling after weight refit/resume.

## Test plan

* `ruff format --check` and `ruff check` on the touched Python files
* `bash -n` on the modified functional scripts
* Parse the exemplar and reference YAML configs with PyYAML
* Earlier dedicated colocated and non-colocated 10-step runs passed on GB200 with `max(train/token_mult_prob_error) < 1.05` and positive async-scheduler counters
* Existing colocated and non-colocated generation smoke tests in CI


shanmugamr1992:
/ok to test [8dda4a2](https://github.com/NVIDIA-NeMo/RL/pull/3762/commits/8dda4a286d917e206c25a16b2334486a96ae6f91)

nemo_rl/models/generation/megatron/config.py
```diff
@@ -62,6 +62,14 @@ class MCoreGenerationSpecificArgs(TypedDict):
     # TODO: Unify `kv_cache_management_mode` and `recompute_kv_cache_after_weight_updates`.
     kv_cache_management_mode: Literal["persist", "offload"]
 
+    # Dynamic-batching scheduler mode. Options:
+    # - 'legacy': resolve the previous forward pass before preparing the next one.
+    # - 'async': overlap the scheduling phases by preparing the next forward pass
+    #   before resolving the previous one, hiding scheduler CPU time behind GPU
+    #   compute. mcore falls back to legacy ordering for any step that cannot
+    #   overlap (prefill, paused requests, KV-cache pressure).
```
terrykong: [`config.py:69-70`](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/nemo_rl/models/generation/megatron/config.py#L69-L70)

**1 action item.**

PR-introduced. mcore does not fall back to *legacy* ordering here. In `async` mode the per-step fallback is a second async ordering, `_run_async_sched_step_no_overlap` — [`sample -> update -> admit -> forward`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2807), which still launches the forward at the end of step N and resolves it at the start of step N+1. Legacy is [`_run_legacy_step`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L3080), which resolves everything inside one step, and the dispatcher only reaches it when the mode is `LEGACY`.

### AI-1

```suggestion
    #   compute. For any step it cannot overlap (prefill, paused requests,
    #   KV-cache pressure) mcore drops to a non-overlapped async ordering --
    #   not to legacy.
```

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -483,6 +489,21 @@ def finish_generation(self, *, release_gpu: bool = True) -> None:
         if self.is_generation_colocated and not release_gpu:
             return
         print(f"[Rank {self.rank}] finishing generation", flush=True)
+
+        # mcore silently falls back to legacy ordering for any step it cannot
+        # overlap, so report its cumulative overlap-step counter. Without this
+        # the only observable difference between the two modes is throughput.
```
terrykong: [`megatron_worker.py:493-495`](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/nemo_rl/models/generation/megatron/megatron_worker.py#L493-L495)

**1 action item, 1 optional follow-up.**

**TL;DR --- `async_sched_step_count` counts every step run in async mode, including the ones that did not overlap, so calling it an "overlap-step counter" overstates what the assertion proves.**

PR-introduced (wording only --- the check itself is worth keeping).

The counter has exactly one increment site, [`text_generation_controller.py:2787`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2787) inside `_build_async_sched_step_result`. All three async branches reach it --- [no-overlap](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2893), [overlap](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2974), [overlap-MTP](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L3065) --- while `_run_legacy_step` never touches it.

So a nonzero value does prove `async_sched_mode` reached the engine, which is the regression the tests are there to catch. It does not prove any step overlapped.

### AI-1

```suggestion
        # mcore counts every step it runs in async mode here, including steps it
        # could not overlap, so this is "steps run in async mode", not "steps
        # that overlapped". A nonzero value proves async_sched_mode reached the
        # engine: the legacy path never increments this counter.
```

Both test scripts carry the same sentence and echo `async scheduling overlap steps` --- worth the same swap at [colocated L53-54](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_colocated_async_sched.sh#L53-L54) / [L65](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_colocated_async_sched.sh#L65) and [non-colocated L58-59](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_non_colocated_async_sched.sh#L58-L59) / [L70](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_non_colocated_async_sched.sh#L70).

### Follow-up (optional)

`(cumul)` in the printed label is only true under `kv_cache_management_mode: "persist"`. Under `recompute`, [`reinitialize_inference_state_buffers`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/contexts/dynamic_context.py#L1515) calls `reset_metadata()` without `preserve_counters`, which [zeroes the counter](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/contexts/dynamic_context.py#L2778) on every refit --- so it becomes "steps since the last refit". No shipped config hits this today, so this is only a note. If you do rename the label, the `grep` pattern in **both** scripts has to change with it, so I left it out of the suggestion above to keep that commit self-contained.

tests/functional/grpo_megatron_generation_colocated_async_sched.sh
```diff
@@ -0,0 +1,65 @@
...
+    policy.train_global_batch_size=4 \
+    policy.logprob_batch_size=4 \
+    policy.train_micro_batch_size=1 \
+    policy.generation.backend=megatron \
+    policy.generation.mcore_generation_config.async_sched_mode=async \
+    cluster.gpus_per_node=2 \
+    grpo.max_num_steps=10 \
+    grpo.val_period=10 \
+    grpo.max_val_samples=8 \
+    grpo.val_batch_size=8 \
```
terrykong: [`tests/functional/grpo_megatron_generation_colocated_async_sched.sh:35-38`](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_colocated_async_sched.sh#L35-L38)

**1 action item.**

**TL;DR --- these two tests run 10 training steps where every sibling runs 2, and they turn on validation without asserting any validation metric, on the heaviest Megatron shard.**

PR-introduced. Sibling step counts in this same shard: [topology L47](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_topology.sh#L47) = 2, [non_colocated L36](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_non_colocated.sh#L36) = 2, [colocated_reshard L40](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_colocated_reshard.sh#L40) = 2, [colocated_async_grpo L40](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_colocated_async_grpo.sh#L40) = 3 --- and that last one [writes down why it needs 3](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_colocated_async_grpo.sh#L64).

Nothing under test needs 10. The counter is cumulative and reaches roughly 460 per training step by your own numbers (4604 over 10), so 2 steps still lands in the hundreds. The non-colocated suspend/refit path is already exercised by step 2.

Two things make the validation overrides worth removing as well: the base megatron config already has `val_period: 10`, so a 2-step run never validates and the overrides only exist because of the 10; and neither script asserts a `validation/*` metric, while the one sibling that does enable validation [asserts one](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_colocated_async_grpo.sh#L60). As written the validation pass costs time and checks nothing.

This lands on `L1_Functional_Tests_Megatron_4.sh`, which after this PR runs 12 tests against 4 / 4 / 5 in Megatron shards 1-3.

### AI-1

```suggestion
    grpo.max_num_steps=2 \
```

Then change the metric line at [L51](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_colocated_async_sched.sh#L51) to `'"2" in data["train/loss"]'` --- `check_metrics.py` evaluates that as a plain key-membership test over step keys, so it just means "step 2 was logged". Same four edits in the non-colocated script ([L40-43](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_non_colocated_async_sched.sh#L40-L43), [L56](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/tests/functional/grpo_megatron_generation_non_colocated_async_sched.sh#L56)).

If 10 steps *is* deliberate, a one-line header comment saying why --- the standard `colocated_async_grpo.sh` already meets --- would settle it.

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -295,6 +296,11 @@ def _initialize_inference_engine(self, mcore_generation_config: dict) -> None:
             max_requests=max_requests,
         )
 
+        if "async_sched_mode" in mcore_generation_config:
+            inference_config.async_sched_mode = AsyncScheduleMode(
+                mcore_generation_config["async_sched_mode"]
+            )
```
terrykong: [`megatron_worker.py:299-302`](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/nemo_rl/models/generation/megatron/megatron_worker.py#L299-L302)

**1 action item.**

PR-introduced, consistency only --- the behaviour is correct as written.

`async_sched_mode` is the only field set by mutating `inference_config` after it is built. Every other optional key in this same function goes through `.get()` and then the constructor: [`max_requests` L296](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/nemo_rl/models/generation/megatron/megatron_worker.py#L296) and [`logging_step_interval` L293](https://github.com/NVIDIA-NeMo/RL/blob/8dda4a286d917e206c25a16b2334486a96ae6f91/nemo_rl/models/generation/megatron/megatron_worker.py#L293). It is a plain [`InferenceConfig` dataclass field](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/config.py#L391) with default `AsyncScheduleMode.LEGACY`, so it can be passed the same way.

(The `inference_cuda_graph_scope` block just below is not a precedent --- it writes `gen_model.config`, a different object, and is not an `InferenceConfig` field at all.)

### AI-1

Not a suggestion block because the edit spans two regions --- add the kwarg inside the `InferenceConfig(...)` call and delete lines 299-302:

```python
            max_requests=max_requests,
            async_sched_mode=AsyncScheduleMode(
                mcore_generation_config.get("async_sched_mode", "legacy")
            ),
        )
```

To be clear on what this does *not* change: I checked `InferenceConfig.__post_init__` and it touches this field only to coerce the enum ([config.py L434](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/inference/config.py#L434)), so the current code skips no validation, and the mutation already happens before the context and engine are built. This is purely about matching the pattern five lines above.

examples/configs/grpo_math_1B.yaml
```diff
@@ -411,6 +411,7 @@ policy:
       enable_prefix_caching: false  # Reuse KV blocks across requests sharing a prompt prefix.
       max_tokens: 16384 # Maximum number of tokens to use in a single step. Analogous to vllm's max_num_batched_tokens
       kv_cache_management_mode: "persist"  # KV cache lifecycle across suspend/resume. Options: "persist", "offload". To select "recompute", set grpo.async_grpo.recompute_kv_cache_after_weight_updates=true.
+      async_sched_mode: "legacy"  # Dynamic-batching scheduler mode. Options: "legacy", "async". "async" hides scheduler CPU time behind GPU compute.
```
lmcafee-nvidia: can we make this value `"async"`?

tests/functional/grpo_megatron_generation_colocated_async_sched.sh
```diff
@@ -0,0 +1,65 @@
...
+uv run tests/json_dump_tb_logs.py $LOG_DIR --output_path $JSON_METRICS
+
+uv run tests/check_metrics.py $JSON_METRICS \
+    'max(data["train/token_mult_prob_error"]) < 1.05' \
+    '"10" in data["train/loss"]'
+
+# mcore quietly falls back to legacy ordering for any step it cannot overlap, so
+# a nonzero overlap-step count is the only signal that async_sched_mode=async
+# actually took effect rather than being accepted and ignored.
+ASYNC_SCHED_STEPS=$(grep -o 'mcore async scheduling steps (cumul): [0-9]*' $RUN_LOG | grep -o '[0-9]*$' | sort -n | tail -1 || true)
```
lmcafee-nvidia, line 75: `ASYNC_SCHED_STEPS` has zero correlation with whether we went through the overlap path or not. `ASYNC_SCHED_STEPS` is backed by `context.async_sched_step_count`, which is incremented for both the no-overlap and the overlap paths, so some other signal would be required to quantify just the overlap path.

tests/functional/L1_Functional_Tests_Megatron_4.sh
```diff
@@ -53,6 +53,8 @@ if megatron_generation_supported; then
     run_test      uv run --no-sync bash ./tests/functional/grpo_megatron_generation_topology.sh
     run_test      uv run --no-sync bash ./tests/functional/grpo_megatron_generation_non_colocated.sh
     run_test      uv run --no-sync bash ./tests/functional/grpo_megatron_generation_colocated_reshard.sh
+    run_test      uv run --no-sync bash ./tests/functional/grpo_megatron_generation_colocated_async_sched.sh
```
lmcafee-nvidia: once the shared recipe defaults to async, the existing colocated and non-colocated two-step smoke tests exercise these same paths. please move the nonzero `ASYNC_SCHED_STEPS` check into those scripts and remove the two new ten-step jobs and their runner entries; the first generation cycle already reports about 414 scheduling steps, and the second non-colocated cycle covers the post-training refit/resume path.

shanmugamr1992: Done in 6724e0a77. I moved the positive scheduler-counter check into `grpo_megatron_generation.sh` and `grpo_megatron_generation_non_colocated.sh`, removed both dedicated ten-step scripts, and removed their runner entries.

examples/configs/grpo_math_1B.yaml
shanmugamr1992: Done in 6724e0a77. The shared exemplar now defaults `async_sched_mode` to `"async"`, and the v1 reference config matches it.

tests/functional/grpo_megatron_generation_colocated_async_sched.sh
shanmugamr1992: Agreed—the counter does not quantify overlap. In 6724e0a77 I corrected the worker/test wording: it counts steps processed by the async scheduler, including steps that fall back to legacy ordering. The smoke tests now use a positive value only to confirm that async mode reached and ran in the engine.

lmcafee-nvidia (approved):

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -493,6 +499,20 @@ def finish_generation(self, *, release_gpu: bool = True) -> None:
         if self.is_generation_colocated and not release_gpu:
             return
         print(f"[Rank {self.rank}] finishing generation", flush=True)
+
+        # Report how many steps the async scheduler processed. This includes
+        # steps that used legacy ordering because they could not overlap.
```
terrykong: [`megatron_worker.py:503-504`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/nemo_rl/models/generation/megatron/megatron_worker.py#L503-L504)

**1 action item.**

Thanks for the rewrite --- "This includes steps that used ... because they could not overlap" fixes the part that mattered most: the counter is not an overlap counter. One word is still off. Those steps do not use *legacy* ordering.

All three async branches --- [no-overlap](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2990), [overlap](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L3071), [overlap-MTP](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L3162) --- reach the single increment site at [`text_generation_controller.py:2878`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2878), and `_run_legacy_step` never touches it. That is exactly why a nonzero value proves `async_sched_mode` reached the engine. Calling the fallback "legacy" undercuts the very thing the check is proving.

### AI-1

```suggestion
        # Report how many steps the async scheduler processed. This includes
        # steps that ran a non-overlapped async ordering because they could
        # not overlap -- the legacy path never increments this counter.
```

Same swap in both smoke tests ([here](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/tests/functional/grpo_megatron_generation.sh#L47) and [here](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/tests/functional/grpo_megatron_generation_non_colocated.sh#L50)).

One note, no action: `(cumul)` in the printed label is only true under `kv_cache_management_mode: "persist"`. Under `recompute`, [`reinitialize_inference_state_buffers`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/contexts/dynamic_context.py#L1512) calls `reset_metadata()` without `preserve_counters`, which zeroes the counter on each refit. Nothing shipped hits that today. If you do rename the label, the `grep` in both smoke tests has to change with it, so I left it out of the suggestion.

examples/configs/grpo_math_1B.yaml
```diff
@@ -415,6 +415,7 @@ policy:
       enable_prefix_caching: false  # Reuse KV blocks across requests sharing a prompt prefix.
       max_tokens: 16384 # Maximum number of tokens to use in a single step. Analogous to vllm's max_num_batched_tokens
       kv_cache_management_mode: "persist"  # KV cache lifecycle across suspend/resume. Options: "persist", "offload", "recompute".
+      async_sched_mode: "async"  # Dynamic-batching scheduler mode. Options: "legacy", "async". "async" hides scheduler CPU time behind GPU compute.
```
terrykong: [`grpo_math_1B.yaml:418`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/examples/configs/grpo_math_1B.yaml#L418)

**1 action item, 1 question.**

The default changed from `legacy` to `async`, so every recipe inheriting this file now gets the async scheduler. mcore's own default is [`LEGACY`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/config.py#L391).

### AI-1

The Summary still says the exemplar keeps the mcore default `legacy`, "so existing behaviour is unchanged". Please update it.

### Question

Do you have numbers where `async` beat `legacy`? Worth one line in the description now that it is the default.

Also worth checking that the spec-decoding recipes inheriting this file still start --- mcore's async guard [requires one MTP depth per speculative token](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/engines/dynamic_engine.py#L1173).

nemo_rl/models/generation/megatron/config.py
```diff
@@ -61,6 +61,14 @@ class MCoreGenerationSpecificArgs(TypedDict):
     # - "recompute": cache is dropped and rebuilt on resume
     kv_cache_management_mode: Literal["persist", "offload", "recompute"]
 
+    # Dynamic-batching scheduler mode. Options:
+    # - 'legacy': resolve the previous forward pass before preparing the next one.
+    # - 'async': overlap the scheduling phases by preparing the next forward pass
+    #   before resolving the previous one, hiding scheduler CPU time behind GPU
+    #   compute. mcore falls back to legacy ordering for any step that cannot
+    #   overlap (prefill, paused requests, KV-cache pressure).
```
terrykong: [`config.py:68-69`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/nemo_rl/models/generation/megatron/config.py#L68-L69)

**1 action item.**

The fallback is not legacy ordering. It is a second async ordering, [`_run_async_sched_step_no_overlap`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2901). [`_run_legacy_step`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L3177) runs only when the mode is `legacy`.

```suggestion
    #   compute. For any step it cannot overlap (prefill, paused requests,
    #   KV-cache pressure) mcore drops to a non-overlapped async ordering --
    #   not to legacy.
```

Same wording in [`megatron_worker.py:504`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/nemo_rl/models/generation/megatron/megatron_worker.py#L504), [`grpo_megatron_generation.sh:47`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/tests/functional/grpo_megatron_generation.sh#L47), [`grpo_megatron_generation_non_colocated.sh:50`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/tests/functional/grpo_megatron_generation_non_colocated.sh#L50).

nemo_rl/models/generation/megatron/megatron_worker.py
terrykong: [`megatron_worker.py:503-504`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/nemo_rl/models/generation/megatron/megatron_worker.py#L503-L504)

**1 action item.**

Same wording as the `config.py` comment: these steps do not use legacy ordering. The legacy path never increments this counter --- [only the async branches do](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2878).

```suggestion
        # Report how many steps the async scheduler processed. This includes
        # steps that ran a non-overlapped async ordering because they could
        # not overlap -- the legacy path never increments this counter.
```

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -305,6 +306,11 @@ def _initialize_inference_engine(self, mcore_generation_config: dict) -> None:
             max_requests=max_requests,
         )
 
+        if "async_sched_mode" in mcore_generation_config:
+            inference_config.async_sched_mode = AsyncScheduleMode(
+                mcore_generation_config["async_sched_mode"]
+            )
```
terrykong: [`megatron_worker.py:309-312`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/nemo_rl/models/generation/megatron/megatron_worker.py#L309-L312)

**1 action item.**

This is the only field set by mutating `inference_config` after it is built. The other optional keys go through the constructor --- [`max_requests`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/nemo_rl/models/generation/megatron/megatron_worker.py#L306), [`logging_step_interval`](https://github.com/NVIDIA-NeMo/RL/blob/76aa4f067764294cc69a441135c08e10899ed810/nemo_rl/models/generation/megatron/megatron_worker.py#L303). This one can too; it is a plain [dataclass field](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/inference/config.py#L391).

Not a suggestion block --- the edit spans two regions (add the kwarg, delete 309-312):

```python
            max_requests=max_requests,
            async_sched_mode=AsyncScheduleMode(
                mcore_generation_config.get("async_sched_mode", "legacy")
            ),
        )
```

terrykong (commented):
Re-reviewed at `76aa4f0`.

Folding the check into the existing smoke tests is a better shape than two dedicated jobs --- my earlier comment about the 10-step runs no longer applies and is dropped.

One new thing: this also flips the exemplar default from `legacy` to `async`, so the PR is a behaviour change rather than a new opt-in knob. See the comment on `grpo_math_1B.yaml`.

Verified against the Megatron-Core pin (`731b7914`).

_Generated by Claude Code_

nemo_rl/models/generation/megatron/config.py
terrykong: fixed in fef810f769791843a3539a22a50c744727bb9934

nemo_rl/models/generation/megatron/megatron_worker.py
terrykong: fixed in fef810f769791843a3539a22a50c744727bb9934

nemo_rl/models/generation/megatron/megatron_worker.py
terrykong: fixed in fef810f769791843a3539a22a50c744727bb9934

terrykong (dismissed):

nemo_rl/models/generation/megatron/config.py
```diff
@@ -82,6 +82,15 @@ class MCoreGenerationSpecificArgs(TypedDict):
...
     kv_cache_management_mode: Literal["persist", "offload", "recompute"]
 
+    # Dynamic-batching scheduler mode. Options:
+    # - 'legacy': resolve the previous forward pass before preparing the next one.
+    # - 'async': overlap the scheduling phases by preparing the next forward pass
+    #   before resolving the previous one, hiding scheduler CPU time behind GPU
+    #   compute. For any step it cannot overlap (prefill, paused requests,
+    #   KV-cache pressure) mcore drops to a non-overlapped async ordering --
+    #   not to legacy.
+    async_sched_mode: NotRequired[Literal["legacy", "async"]]
```
cspades: I think this is a duplicate of https://github.com/NVIDIA-NeMo/RL/pull/3762/changes#diff-48def2e93bfd7f7d5963538feace671003e937a390cc3d43dd6ee7c35f52a8a4R51

Delete the above one!

nemo_rl/models/generation/megatron/megatron_worker.py
cspades, line 392: I think my multimodal PR already added this support: https://github.com/NVIDIA-NeMo/RL/commit/13a10647ebbf0f940d2b06ea41800b3f2fb46099

cspades:
@shanmugamr1992 could you merge your changes against main branch, async scheduling is already exposed but you added more documentation and tests, which should be merged, and the logic should be unified since right now it's being set twice.

cspades (changes requested):
Currently this PR branch sets async scheduling twice in different ways, we only need 1 way.

nemo_rl/models/generation/megatron/config.py
shanmugamr1992: Done in 0cab3dcda. Rebased onto main, dropped the duplicate declaration, and moved the doc comment onto the single field that came in with #3920 (`config.py:51`).

nemo_rl/models/generation/megatron/megatron_worker.py
shanmugamr1992: Done in 0cab3dcda. Removed the constructor kwarg; the mode is now set only through the `inference_overrides` path from 13a1064. That also drops the `.get("async_sched_mode", "legacy")` fallback, so MCore's `InferenceConfig` default stays the single source of truth when the key is absent.

Worth noting the duplication was not just redundant: `InferenceConfig(**inference_overrides, ..., async_sched_mode=...)` raises `TypeError: got multiple values for keyword argument` whenever the key is set, which the exemplar now always does.

shanmugamr1992:
@cspades addressed in 0cab3dcda (rebased onto `main`, single squashed commit).

* **Duplicate config field** — removed. `async_sched_mode` is declared once, on the field that arrived with #3920; the doc comment moved onto it.
* **Duplicate plumbing** — removed. The mode is set only via the `inference_overrides` path from 13a1064, so `InferenceConfig` receives it once and MCore's own default applies when the key is absent. The stale `.get("async_sched_mode", "legacy")` call-site fallback is gone too.
* The rebase also had to move the colocated assertion: #3570 deleted `grpo_megatron_generation.sh`, so the check now lives in `grpo_megatron_generation_colocated.sh` alongside the one in `grpo_megatron_generation_non_colocated.sh`.

What remains in the PR is the documentation, the exemplar default of `async`, and the scheduler-counter assertion in the two smoke tests.

cspades (dismissed):

terrykong (dismissed):
approved, but @shanmugamr1992 can you fix the linting issues and retry the CI?

shanmugamr1992:
/ok to test [12aa4c8](https://github.com/NVIDIA-NeMo/RL/pull/3762/commits/12aa4c87feafc9da21cd42ed909168980e456c0f)

terrykong (approved):

PR #3765, opened 2026-08-21
feat: log toolcall and thinktag violation rate into single controller

arnavk-nvidia:
# What does this PR do ?
Logs metrics of toolcall and think tag violation rate in single controller similar to the one existing in grpo.py.
In single controller the message log gets dropped hence the metrics are calculated in payload itself.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/single_controller_utils/utils.py
+++ b/nemo_rl/algorithms/single_controller_utils/utils.py
@@ -114,7 +114,7 @@ def reduce_advantage_pump_metrics(
     Returns:
         Step-level reward, advantage, token-count, optional sequence
         log-probability error metrics, and per-sample violation counts.
-    
+
     """
     out: dict[str, float] = {}
     if rewards:
--- a/nemo_rl/algorithms/single_controller_utils/utils.py
+++ b/nemo_rl/algorithms/single_controller_utils/utils.py
@@ -94,6 +94,7 @@ def reduce_advantage_pump_metrics(
     rewards: list[torch.Tensor],
     masked_advantages: list[torch.Tensor],
     sequence_lengths: list[int],
+    *,
     seq_logprob_error_metrics: list[dict[str, float]] | None = None,
     num_invalid_tool_calls: list[int] | None = None,
     num_malformed_thinking: list[int] | None = None,
--- a/nemo_rl/algorithms/single_controller_utils/utils.py
+++ b/nemo_rl/algorithms/single_controller_utils/utils.py
@@ -95,6 +95,9 @@ def reduce_advantage_pump_metrics(
     masked_advantages: list[torch.Tensor],
     sequence_lengths: list[int],
     seq_logprob_error_metrics: list[dict[str, float]] | None = None,
+    num_invalid_tool_calls: list[int] | None = None,
+    num_malformed_thinking: list[int] | None = None,
+    num_assistant_messages: list[int] | None = None,
 ) -> dict[str, float]:
     """Reduce per-step accumulators from _advantage_stage into step scalars.
 
@@ -104,10 +107,14 @@ def reduce_advantage_pump_metrics(
         sequence_lengths: All input_lengths trained on this step.
         seq_logprob_error_metrics: Sequence-error metrics and their aggregation
             counts, one record per streaming chunk.
+        num_invalid_tool_calls: Per-sample invalid tool-call counts.
+        num_malformed_thinking: Per-sample malformed-thinking counts.
+        num_assistant_messages: Per-sample assistant message counts (rate denominator).
 
     Returns:
-        Step-level reward, advantage, token-count, and optional sequence
-        log-probability error metrics.
+        Step-level reward, advantage, token-count, optional sequence
+        log-probability error metrics, and per-sample violation counts.
+    
     """
     out: dict[str, float] = {}
     if rewards:
@@ -126,6 +133,15 @@ def reduce_advantage_pump_metrics(
         out["total_num_tokens"] = float(sum(sequence_lengths))
     if seq_logprob_error_metrics:
         out.update(_reduce_seq_logprob_error_metrics(seq_logprob_error_metrics))
+    n_asst = sum(num_assistant_messages or [])
+    if n_asst:
+        n_invalid = sum(num_invalid_tool_calls or [])
+        n_malformed = sum(num_malformed_thinking or [])
+        out["invalid_tool_call_rate"] = n_invalid / n_asst
+        out["malformed_thinking_rate"] = n_malformed / n_asst
+        out["num_invalid_tool_calls"] = float(n_invalid)
+        out["num_malformed_thinking"] = float(n_malformed)
+        out["num_assistant_messages"] = float(n_asst)
     return out
 
 
--- a/tests/unit/experience/test_payload.py
+++ b/tests/unit/experience/test_payload.py
@@ -132,7 +132,15 @@ def test_record_to_train_batch_preserves_routed_experts_in_tq_payload() -> None:
     packed_rows = list(packed_routes.unbind())
     assert torch.equal(packed_rows[0], expected_routes[0])
     assert torch.equal(packed_rows[1], expected_routes[1])
-    assert tags == [{"weight_version": 3}, {"weight_version": 3}]
+    no_violations = {
+        "num_invalid_tool_calls": 0,
+        "num_malformed_thinking": 0,
+        "num_assistant_messages": 1,
+    }
+    assert tags == [
+        {"weight_version": 3, **no_violations},
+        {"weight_version": 3, **no_violations},
+    ]
 
 
 def test_record_to_train_batch_omits_routed_experts_when_absent() -> None:
@@ -190,3 +198,32 @@ def test_record_to_train_batch_backfills_routes_for_failed_completion() -> None:
     _, fields, _ = pack_payload(train_batch, weight_version=3, group_id="group")
     assert "routed_experts" in fields
     assert list(fields["routed_experts"].unbind())[1].shape == (2, 2, 2)
+
+
+def test_pack_payload_stamps_violation_counts_on_tags() -> None:
+    """Flagged assistant turns are counted per row; ungenerated turns are ignored."""
+    completions = [_completion(route_start=10, reward=1.0), _failed_completion()]
+    completions[0].message_log[1]["is_invalid_tool_call"] = True
+    completions[0].message_log[1]["has_malformed_thinking"] = True
+
+    train_batch = record_to_train_batch(
+        _record(completions),
+        pad_value_dict={"token_ids": 0, "input_ids": 0},
+    )
+    _, fields, tags = pack_payload(train_batch, weight_version=7, group_id="g")
+
+    assert "violation_counts" not in fields
+    assert tags == [
+        {
+            "weight_version": 7,
+            "num_invalid_tool_calls": 1,
+            "num_malformed_thinking": 1,
+            "num_assistant_messages": 1,
+        },
+        {
+            "weight_version": 7,
+            "num_invalid_tool_calls": 0,
+            "num_malformed_thinking": 0,
+            "num_assistant_messages": 0,
+        },
+    ]
--- a/tests/unit/experience/test_payload.py
+++ b/tests/unit/experience/test_payload.py
@@ -201,10 +201,14 @@ def test_record_to_train_batch_backfills_routes_for_failed_completion() -> None:
 
 
 def test_pack_payload_stamps_violation_counts_on_tags() -> None:
-    """Flagged assistant turns are counted per row; ungenerated turns are ignored."""
-    completions = [_completion(route_start=10, reward=1.0), _failed_completion()]
+    """Each flag lands in its own counter; a row that never generated counts zero."""
+    completions = [
+        _completion(route_start=10, reward=1.0),
+        _completion(route_start=30, reward=1.0),
+        _failed_completion(),
+    ]
     completions[0].message_log[1]["is_invalid_tool_call"] = True
-    completions[0].message_log[1]["has_malformed_thinking"] = True
+    completions[1].message_log[1]["has_malformed_thinking"] = True
 
     train_batch = record_to_train_batch(
         _record(completions),
@@ -217,6 +221,12 @@ def test_pack_payload_stamps_violation_counts_on_tags() -> None:
         {
             "weight_version": 7,
             "num_invalid_tool_calls": 1,
+            "num_malformed_thinking": 0,
+            "num_assistant_messages": 1,
+        },
+        {
+            "weight_version": 7,
+            "num_invalid_tool_calls": 0,
             "num_malformed_thinking": 1,
             "num_assistant_messages": 1,
         },
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -82,6 +82,7 @@ from nemo_rl.data_plane.schema import DP_CALIB_INPUT_FIELDS
 from nemo_rl.distributed.batched_data_dict import BatchedDataDict
 from nemo_rl.environments.nemo_gym import should_use_nemo_gym
 from nemo_rl.experience.failures import RolloutStall
+from nemo_rl.experience.payload import VIOLATION_TAG_KEYS
 from nemo_rl.experience.rollout_manager import RolloutOutcome
 from nemo_rl.models.generation.sglang.sglang_generation import SGLangGeneration
 from nemo_rl.models.generation.vllm import VllmGeneration
@@ -312,6 +313,7 @@ class SingleControllerActor:
             "masked_advantages": [],
             "sequence_lengths": [],
             "seq_logprob_error_metrics": [],
+            **{key: [] for key in VIOLATION_TAG_KEYS},
         }
         self._opd_stat_sum = 0.0
         self._opd_stat_sumsq = 0.0
@@ -1960,6 +1962,10 @@ class SingleControllerActor:
             The updated batch metadata and whether the batch contains at least
             one valid training token.
         """
+        for tag in meta.tags or []:
+            for key in VIOLATION_TAG_KEYS:
+                self._step_log_dict[key].append(int(tag.get(key, 0)))
+
         if self._advantage_estimator is None:
             return meta, True
         adv_cfg = self._advantage_cfg
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -1964,7 +1964,7 @@ class SingleControllerActor:
         """
         for tag in meta.tags or []:
             for key in VIOLATION_TAG_KEYS:
-                self._step_log_dict[key].append(int(tag.get(key, 0)))
+                self._step_log_dict.setdefault(key, []).append(int(tag.get(key, 0)))
 
         if self._advantage_estimator is None:
             return meta, True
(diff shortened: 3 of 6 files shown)


tests/unit/single_controller/test_sc_utils_helpers.py
```diff
@@ -204,6 +204,24 @@ def test_seq_logprob_error_metrics_are_reduced_across_streaming_chunks(
...
+            num_assistant_messages=[2, 1, 1],
+        )
+        assert out["invalid_tool_call_rate"] == pytest.approx(0.5)
+        assert out["malformed_thinking_rate"] == pytest.approx(0.25)
+        assert out["num_invalid_tool_calls"] == pytest.approx(2.0)
+        assert out["num_malformed_thinking"] == pytest.approx(1.0)
+        assert out["num_assistant_messages"] == pytest.approx(4.0)
+
+    def test_no_assistant_messages_omits_violation_metrics(self) -> None:
+        assert reduce_advantage_pump_metrics([], [], [], [0], [0], [0]) == {}
```
arnavk-nvidia: [tests/unit/single_controller/test_sc_utils_helpers.py:223](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/tests/unit/single_controller/test_sc_utils_helpers.py#L223)

**1 action item.**

### AI-1
`reduce_advantage_pump_metrics`'s 4th positional parameter is [`seq_logprob_error_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/nemo_rl/algorithms/single_controller_utils/utils.py#L93-L100), not `num_invalid_tool_calls`. Positionally, `[0]` lands as `seq_logprob_error_metrics=[0]`, which is truthy, so `_reduce_seq_logprob_error_metrics([0])` runs and does `record[count_key]` on the int `0` → `TypeError: 'int' object is not subscriptable`. Verified by extracting the reduction logic and running it standalone — it raises, it does not return `{}`. This will fail in CI.

**Action:** use keyword args so positional drift can't happen again:

```suggestion
    def test_no_assistant_messages_omits_violation_metrics(self) -> None:
        assert (
            reduce_advantage_pump_metrics(
                [],
                [],
                [],
                num_invalid_tool_calls=[0],
                num_malformed_thinking=[0],
                num_assistant_messages=[0],
            )
            == {}
        )
```

arnavk-nvidia (commented):
Automated review found one bug that will fail CI.

_Generated by Claude Code_

tests/unit/single_controller/test_sc_utils_helpers.py
arnavk-nvidia: fixed in c51e9ffde86f841ec427e682ccfcb0251f011588

tests/unit/experience/test_payload.py
```diff
@@ -190,3 +198,32 @@ def test_record_to_train_batch_backfills_routes_for_failed_completion() -> None:
...
+            "num_malformed_thinking": 1,
+            "num_assistant_messages": 1,
+        },
+        {
+            "weight_version": 7,
+            "num_invalid_tool_calls": 0,
+            "num_malformed_thinking": 0,
+            "num_assistant_messages": 0,
+        },
+    ]
```
yfw: **1 action item.**

**TL;DR — both flags are set on the same message, so swapping the two counters in `_violation_counts` leaves this test green; and the docstring's "ungenerated turns are ignored" is untested and also false.**

PR-introduced test. The underlying guard behaviour is pre-existing and shared with the `grpo.py` analog.

**Blind to a counter swap.** [Lines 206-207](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/tests/unit/experience/test_payload.py#L206) set `is_invalid_tool_call` and `has_malformed_thinking` on the *same* assistant turn, so the expected tag is symmetric (`1`/`1`). A mutant that increments `num_malformed_thinking` on the invalid-tool-call branch and vice versa produces byte-identical output — verified: mutant survives, 4 passed.

**The docstring claim is false.** [`record_to_train_batch` calls `add_grpo_token_loss_masks_and_generation_logprobs`](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/nemo_rl/experience/payload.py#L101) before [`_violation_counts` runs](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/nemo_rl/experience/payload.py#L120), and that function [backfills `generation_logprobs` onto every message that lacks it](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/nemo_rl/algorithms/grpo.py#L1939-L1942). So the `"generation_logprobs" not in message` guard at [payload.py:47](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/nemo_rl/experience/payload.py#L47) is never True, and prompt-history assistant turns land in the denominator. Measured: a log with one ungenerated plus one generated assistant turn tags `num_assistant_messages: 2`. The test misses it because [`_failed_completion()`](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/tests/unit/experience/test_payload.py#L163) has no assistant message at all.

**Please don't "fix" that by moving `_violation_counts` above line 101.** `grpo.py` has the identical ordering ([backfill 3150](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/nemo_rl/algorithms/grpo.py#L3150) -> [counting 3315](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/nemo_rl/algorithms/grpo.py#L3315), [same predicate 2040](https://github.com/NVIDIA-NeMo/RL/blob/76b70ce14103e0779a9615db63aea905af761b48/nemo_rl/algorithms/grpo.py#L2040)), so the two denominators match — which is the parity this PR exists for. Only the docstring is wrong.

### AI-1
**Action:** put each flag on its own row so the two counters are distinguishable, and reword the docstring. Verified passing locally.

```suggestion
def test_pack_payload_stamps_violation_counts_on_tags() -> None:
    """Each flag lands in its own counter; a row that never generated counts zero."""
    completions = [
        _completion(route_start=10, reward=1.0),
        _completion(route_start=30, reward=1.0),
        _failed_completion(),
    ]
    completions[0].message_log[1]["is_invalid_tool_call"] = True
    completions[1].message_log[1]["has_malformed_thinking"] = True

    train_batch = record_to_train_batch(
        _record(completions),
        pad_value_dict={"token_ids": 0, "input_ids": 0},
    )
    _, fields, tags = pack_payload(train_batch, weight_version=7, group_id="g")

    assert "violation_counts" not in fields
    assert tags == [
        {
            "weight_version": 7,
            "num_invalid_tool_calls": 1,
            "num_malformed_thinking": 0,
            "num_assistant_messages": 1,
        },
        {
            "weight_version": 7,
            "num_invalid_tool_calls": 0,
            "num_malformed_thinking": 1,
            "num_assistant_messages": 1,
        },
        {
            "weight_version": 7,
            "num_invalid_tool_calls": 0,
            "num_malformed_thinking": 0,
            "num_assistant_messages": 0,
        },
    ]
```

nemo_rl/algorithms/single_controller_utils/utils.py
```diff
@@ -104,10 +107,14 @@ def reduce_advantage_pump_metrics(
...
+        num_malformed_thinking: Per-sample malformed-thinking counts.
+        num_assistant_messages: Per-sample assistant message counts (rate denominator).
 
     Returns:
-        Step-level reward, advantage, token-count, and optional sequence
-        log-probability error metrics.
+        Step-level reward, advantage, token-count, optional sequence
+        log-probability error metrics, and per-sample violation counts.
+    
     """
```
yfw, line 119: **1 action item** — the lint job is red on this line.

[`utils.py:117`](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/algorithms/single_controller_utils/utils.py#L117) — PR-introduced whitespace-only line before the closing `"""`.

`uv run --group dev pre-commit run --all-files` -> `trim trailing whitespace ... Failed ... Fixing nemo_rl/algorithms/single_controller_utils/utils.py`, and `ruff format --check` -> `Would reformat: nemo_rl/algorithms/single_controller_utils/utils.py`. (`ruff check` and `pyrefly check` are both clean on this PR.)

### AI-1
**Action:** drop the stray line.

```suggestion
        log-probability error metrics, and per-sample violation counts.
    """
```

nemo_rl/algorithms/single_controller_utils/utils.py
```diff
@@ -95,6 +95,9 @@ def reduce_advantage_pump_metrics(
     masked_advantages: list[torch.Tensor],
     sequence_lengths: list[int],
     seq_logprob_error_metrics: list[dict[str, float]] | None = None,
```
yfw, line 98: **1 action item** — small, and it prevents a repeat of the positional-arg bug you just fixed in c51e9ffde86f841ec427e682ccfcb0251f011588.

[`utils.py:97`](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/algorithms/single_controller_utils/utils.py#L97) — PR-introduced: this PR grew the tail of the signature from one optional to four adjacent, same-typed `list[...] | None` parameters.

The repo's own rule, [`linting-and-formatting/SKILL.md:68`](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/.agents/contributor-skills/linting-and-formatting/SKILL.md#L68): *"When a function has multiple parameters of the same type that could easily be swapped by mistake, use a bare `*` to force keyword-only arguments starting from where the ambiguity begins. This prevents callers from accidentally transposing arguments."*

Your fix corrected the one call site; this closes the class. Safe to add — I replayed every call form against the new signature: the production caller splats kwargs ([single_controller.py:1081](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/algorithms/single_controller.py#L1081)), and the remaining positional calls pass at most 3 args, which stay positional under this change. Verified: full `test_utils.py` + `test_payload.py` green, `ruff format`/`ruff check` clean.

### AI-1
**Action:**

```suggestion
    *,
    seq_logprob_error_metrics: list[dict[str, float]] | None = None,
```

tests/unit/experience/test_payload.py
yfw, line 239: **1 action item.**

**TL;DR — both flags are set on the same message, so swapping the two counters in `_violation_counts` leaves this test green; and the docstring's "ungenerated turns are ignored" is untested and also false.**

PR-introduced test. The underlying guard behaviour is pre-existing and shared with the `grpo.py` analog.

**Blind to a counter swap.** [Lines 206-207](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/tests/unit/experience/test_payload.py#L206) set `is_invalid_tool_call` and `has_malformed_thinking` on the *same* assistant turn, so the expected tag is symmetric (`1`/`1`). A mutant that increments `num_malformed_thinking` on the invalid-tool-call branch and vice versa produces byte-identical output — verified: mutant survives, 4 passed.

**The docstring claim is false.** [`record_to_train_batch` calls `add_grpo_token_loss_masks_and_generation_logprobs`](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/experience/payload.py#L101) before [`_violation_counts` runs](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/experience/payload.py#L120), and that function [backfills `generation_logprobs` onto every message that lacks it](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/algorithms/grpo.py#L1939-L1942). So the `"generation_logprobs" not in message` guard at [payload.py:47](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/experience/payload.py#L47) is never True, and prompt-history assistant turns land in the denominator. Measured: a log with one ungenerated plus one generated assistant turn tags `num_assistant_messages: 2`. The test misses it because [`_failed_completion()`](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/tests/unit/experience/test_payload.py#L163) has no assistant message at all.

**Please don't "fix" that by moving `_violation_counts` above line 101.** `grpo.py` has the identical ordering (backfill then count, [same predicate](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/algorithms/grpo.py#L2040)), so the two denominators match — which is the parity this PR exists for. Only the docstring is wrong.

### AI-1
**Action:** put each flag on its own row so the two counters are distinguishable, and reword the docstring. Verified passing locally.

```suggestion
def test_pack_payload_stamps_violation_counts_on_tags() -> None:
    """Each flag lands in its own counter; a row that never generated counts zero."""
    completions = [
        _completion(route_start=10, reward=1.0),
        _completion(route_start=30, reward=1.0),
        _failed_completion(),
    ]
    completions[0].message_log[1]["is_invalid_tool_call"] = True
    completions[1].message_log[1]["has_malformed_thinking"] = True

    train_batch = record_to_train_batch(
        _record(completions),
        pad_value_dict={"token_ids": 0, "input_ids": 0},
    )
    _, fields, tags = pack_payload(train_batch, weight_version=7, group_id="g")

    assert "violation_counts" not in fields
    assert tags == [
        {
            "weight_version": 7,
            "num_invalid_tool_calls": 1,
            "num_malformed_thinking": 0,
            "num_assistant_messages": 1,
        },
        {
            "weight_version": 7,
            "num_invalid_tool_calls": 0,
            "num_malformed_thinking": 1,
            "num_assistant_messages": 1,
        },
        {
            "weight_version": 7,
            "num_invalid_tool_calls": 0,
            "num_malformed_thinking": 0,
            "num_assistant_messages": 0,
        },
    ]
```

yfw (commented):
Thanks for closing this gap — the SC path had no violation telemetry while the sync GRPO path did, and riding the documented [`tags` sidecar](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/data_plane/interfaces.py#L117-L122) with metric names matching `grpo.py` is the right shape for it. Nice quick turnaround on the keyword-arg fix in c51e9ffde86f841ec427e682ccfcb0251f011588 too.

**2 action items, 1 follow-up.** Findings re-verified against c51e9ff after the force-push.

**The branch is still red on tests it doesn't touch.** `addopts` carries `-x`, so a plain run stops at the first failure and hides the rest:

```
$ uv run --group test pytest tests/unit/single_controller/ tests/unit/experience/ -q --maxfail=0
7 failed, 702 passed
```

### AI-1 — `_train_pump_controller` needs the new `_step_log_dict` keys

All 7 remaining failures are `test_train_pump_*` / `test_advantage_stage_*` in `tests/unit/single_controller/test_single_controller_actor.py` — a file this PR doesn't touch — and they share one cause:

```
    for tag in meta.tags or []:
        for key in VIOLATION_TAG_KEYS:
>           self._step_log_dict[key].append(int(tag.get(key, 0)))
E           KeyError: 'num_invalid_tool_calls'
nemo_rl/algorithms/single_controller.py:1967: KeyError
```

The new drain loop hard-indexes `_step_log_dict`. Production seeds every key at [single_controller.py:313-319](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/algorithms/single_controller.py#L313-L319), but [`_train_pump_controller`](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/tests/unit/single_controller/test_single_controller_actor.py#L1061) hand-builds it with four at [line 1112](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/tests/unit/single_controller/test_single_controller_actor.py#L1112), and its metas carry `tags`, so the loop actually runs. That file is unchanged from `main`, so this is a regression from the production change rather than a stale test.

**Action:** add the keys to the fixture, plus the import. Verified locally — `45 passed` on that file, from 7 failed.

```python
# tests/unit/single_controller/test_single_controller_actor.py
from nemo_rl.experience.payload import VIOLATION_TAG_KEYS
...
    ctrl._step_log_dict = {
        "rewards": [],
        "masked_advantages": [],
        "sequence_lengths": [],
        "seq_logprob_error_metrics": [],
        **{key: [] for key in VIOLATION_TAG_KEYS},
    }
```

Please fix in this PR — CI stays red otherwise. Worth fixing in the fixture rather than by moving the drain loop below the `_advantage_estimator is None` early-out: the other `_advantage_stage` tests pass only because their metas have no `tags`, which is a fragile reason to be green.

### AI-2 — one comment on `VIOLATION_TAG_KEYS`

`VIOLATION_TAG_KEYS` is now load-bearing in four independent places: the tuple itself, the `__init__` seed, the `reduce_advantage_pump_metrics` signature, and that test fixture. Nothing links them statically — the `**self._step_log_dict` splat at [single_controller.py:1081](https://github.com/NVIDIA-NeMo/RL/blob/c51e9ffde86f841ec427e682ccfcb0251f011588/nemo_rl/algorithms/single_controller.py#L1081) is opaque to both ruff and pyrefly. The `KeyError` above is one failure mode; adding a 4th key to the tuple without editing the `utils.py` signature is the other, and it surfaces as `TypeError: ... unexpected keyword argument` at the first train step.

**Action:** a one-line comment on the tuple ("keep in sync with `reduce_advantage_pump_metrics` kwargs"). Explicitly **not** asking for a refactor — collapsing the three kwargs into a single dict parameter moves the complexity rather than removing it, and it would break the deliberate metric-name parity with `grpo.py` that motivates this PR.

### Follow-up — an assertion on the drained counts

Nothing currently asserts what the drain loop does: the other `_advantage_stage` tests build metas without `tags`, so it runs zero iterations, and the 7 above only enter it incidentally. Since you're editing that fixture anyway, an assertion on the drained counts would also pin the `tag.get(key, 0)` default for rows enqueued before this feature — the rolling-upgrade case. Fine as a follow-up if you'd rather keep this PR tight.

### Context — no action

`ruff check` and `pyrefly check` are clean on all changed files. We looked at these and are explicitly **not** asking for them: popping the non-tensor `violation_counts` key out of the `BatchedDataDict` (it's inert — `pack_payload` filters on `isinstance(v, torch.Tensor)` and the batch is dropped right after); using `stamp_tags` instead of `pack_payload` (wrong lifecycle — `stamp_tags` is for scalars derived later in the train loop); and unifying `_violation_counts` with grpo's counting loop (that loop also writes penalty advantages and threads `token_offset`, so merging them would be over-abstraction).

_Generated by Claude Code_

nemo_rl/algorithms/single_controller_utils/utils.py
yfw: fixed in 8c2ba3dbe1bd02372f55ca7f9ea6e5cba01f9690

yfw (approved):

PR #3766, opened 2026-08-22
fix(grpo): support environment sample masking in single controller

macandro96:
# What does this PR do ?

Support https://github.com/NVIDIA-NeMo/RL/pull/3163 for single controller path

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -434,6 +434,7 @@ class SingleControllerActor:
         self._step_log_dict: dict[str, list] = {
             "rewards": [],
             "masked_advantages": [],
+            "num_mask_sample_filtered": [],
             "sequence_lengths": [],
             "seq_logprob_error_metrics": [],
             **{key: [] for key in VIOLATION_TAG_KEYS},
@@ -1856,8 +1857,9 @@ class SingleControllerActor:
                     if self._is_ppo and not has_valid_training_tokens:
                         raise RuntimeError(
                             "SingleController has no valid response tokens after "
-                            "filtering. Check ppo.seq_logprob_error_threshold to "
-                            "avoid an optimizer step with an empty batch."
+                            "filtering. Check seq_logprob_error_threshold, "
+                            "overlong_filtering, and environment mask_sample flags "
+                            "to avoid an optimizer step with an empty batch."
                         )
 
                     # ---- 3. Train the model -- train_microbatches_from_meta ----
@@ -1988,8 +1990,9 @@ class SingleControllerActor:
                     if not step_open:
                         raise RuntimeError(
                             "SingleController has no valid response tokens after "
-                            "filtering. Check grpo.seq_logprob_error_threshold to "
-                            "avoid an optimizer step with an empty batch."
+                            "filtering. Check seq_logprob_error_threshold, "
+                            "overlong_filtering, and environment mask_sample flags "
+                            "to avoid an optimizer step with an empty batch."
                         )
 
                     with self._timer.time("policy_training"):
@@ -3203,6 +3206,18 @@ class SingleControllerActor:
         sample_mask = squeeze_trailing_unit_dim(
             tensor_field(data, adv_cfg.sample_mask_field)
         ).float()
+        mask_sample = squeeze_trailing_unit_dim(
+            tensor_field(data, adv_cfg.mask_sample_field)
+        ).bool()
+        truncated = squeeze_trailing_unit_dim(
+            tensor_field(data, adv_cfg.truncated_field)
+        ).bool()
+
+        num_mask_sample_filtered = int(mask_sample.sum().item())
+        self._step_log_dict["num_mask_sample_filtered"].append(num_mask_sample_filtered)
+        final_sample_mask = sample_mask * (~mask_sample).to(sample_mask.dtype)
+        if self._algo_cfg.overlong_filtering:
+            final_sample_mask = final_sample_mask * (~truncated).to(sample_mask.dtype)
 
         seq_logprob_error_threshold = self._algo_cfg.seq_logprob_error_threshold
         # Match the legacy path: whenever real policy logprobs are available,
@@ -3212,7 +3227,7 @@ class SingleControllerActor:
             masking_data = BatchedDataDict(
                 {
                     "token_mask": token_mask,
-                    "sample_mask": sample_mask,
+                    "sample_mask": final_sample_mask,
                     "prev_logprobs": tensor_field(
                         data,
                         adv_cfg.policy_logprobs_field,
@@ -3224,7 +3239,7 @@ class SingleControllerActor:
                 }
             )
             num_valid_seqs_before = float(
-                ((token_mask[:, 1:] * sample_mask.unsqueeze(-1)).sum(dim=-1) > 0)
+                ((token_mask[:, 1:] * final_sample_mask.unsqueeze(-1)).sum(dim=-1) > 0)
                 .sum()
                 .item()
             )
@@ -3233,9 +3248,9 @@ class SingleControllerActor:
                 rewards=rewards,
                 seq_logprob_error_threshold=seq_logprob_error_threshold,
             )
-            sample_mask = masking_data["sample_mask"]
+            final_sample_mask = masking_data["sample_mask"]
             num_valid_seqs_after = float(
-                ((token_mask[:, 1:] * sample_mask.unsqueeze(-1)).sum(dim=-1) > 0)
+                ((token_mask[:, 1:] * final_sample_mask.unsqueeze(-1)).sum(dim=-1) > 0)
                 .sum()
                 .item()
             )
@@ -3246,7 +3261,7 @@ class SingleControllerActor:
             seq_error_metrics["_num_valid_seqs_after"] = num_valid_seqs_after
             self._step_log_dict["seq_logprob_error_metrics"].append(seq_error_metrics)
 
-        mask = token_mask * sample_mask.unsqueeze(-1)
+        mask = token_mask * final_sample_mask.unsqueeze(-1)
 
         repeated_batch: dict[str, torch.Tensor] = {
             "total_reward": rewards,
@@ -3332,8 +3347,8 @@ class SingleControllerActor:
             self._opd_stat_count += int(valid.numel())
 
         fields_to_put = {adv_cfg.output_field: advantages}
-        if seq_logprob_error_threshold is not None:
-            fields_to_put[adv_cfg.sample_mask_field] = sample_mask
+        if not torch.equal(final_sample_mask, sample_mask):
+            fields_to_put[adv_cfg.sample_mask_field] = final_sample_mask
         new_fields = [adv_cfg.output_field]
         if returns is not None:
             fields_to_put[adv_cfg.returns_field] = returns
@@ -3360,6 +3375,8 @@ class SingleControllerActor:
             adv_cfg.token_mask_field,
             adv_cfg.sample_mask_field,
             *adv_cfg.repeated_batch_fields,
+            adv_cfg.mask_sample_field,
+            adv_cfg.truncated_field,
         ]
         if self._message_level_advantage_penalties_enabled:
             fields.extend(
--- a/nemo_rl/experience/payload.py
+++ b/nemo_rl/experience/payload.py
@@ -27,7 +27,9 @@ from nemo_rl.data_plane.column_io import TOKEN_ALIGNED_FIELDS
 from nemo_rl.data_plane.schema import (
     INVALID_TOOL_CALL_MASK,
     MALFORMED_THINKING_MASK,
+    MASK_SAMPLE,
     ROUTED_EXPERTS_FIELD,
+    TRUNCATED,
 )
 from nemo_rl.distributed.batched_data_dict import BatchedDataDict
 from nemo_rl.experience.interfaces import PromptGroupRecord
@@ -102,8 +104,9 @@ def record_to_train_batch(
             flags for configured advantage penalties.
 
     Returns:
-        BatchedDataDict with input_ids, input_lengths, generation_logprobs, token_mask,
-        sample_mask, prompt_ids_for_adv, total_reward, violation counts, and optional
+        BatchedDataDict with input_ids, input_lengths, generation_logprobs,
+        token_mask, an all-ones sample_mask, the raw mask_sample and truncated
+        flags, prompt_ids_for_adv, total_reward, violation counts, and optional
         routed experts and message-violation masks.
     """
     # Lazy imports: grpo and llm_message_utils transitively pull
@@ -113,7 +116,10 @@ def record_to_train_batch(
         extract_initial_prompt_messages,
     )
     from nemo_rl.data.llm_message_utils import batched_message_log_to_flat_message
-    from nemo_rl.experience.rollouts import backfill_missing_routed_experts
+    from nemo_rl.experience.rollouts import (
+        _mask_sample_flags,
+        backfill_missing_routed_experts,
+    )
 
     completions = record.completions
     n = len(completions)
@@ -146,6 +152,8 @@ def record_to_train_batch(
     total_reward = torch.tensor(
         [float(c.reward) for c in completions], dtype=torch.float32
     )
+    mask_sample = _mask_sample_flags(c.env_extras for c in completions)
+    truncated = torch.tensor([c.truncated for c in completions], dtype=torch.bool)
     sample_mask = torch.ones(n, dtype=torch.float32)
 
     train_data: dict[str, Any] = {
@@ -155,6 +163,8 @@ def record_to_train_batch(
         "token_mask": flat["token_loss_mask"],
         "sample_mask": sample_mask,
         "prompt_ids_for_adv": prompt_flat["token_ids"],
+        MASK_SAMPLE: mask_sample,
+        TRUNCATED: truncated,
         "total_reward": total_reward,
         _VIOLATION_COUNTS_KEY: violation_counts,
     }
--- a/docs/guides/single-controller.md
+++ b/docs/guides/single-controller.md
@@ -241,6 +241,6 @@ The SC path is still under active development. Feature gaps are tracked in [issu
 - Generation backend: vLLM and Megatron generation are supported; SGLang and TRT-LLM have not been tested on SC.
 - Validation is not yet supported (setup raises on `val_period > 0`, `val_at_start`, or `val_at_end`); checkpointing is.
 - (PPO) Rollout drop budgets — `async_rl.rollout_failure.max_skipped_prompts` and `max_consecutive_dropped_prompts` must both be `0`. A drop shortens the step, and the critic shards it against the configured `value.train_global_batch_size` rather than its actual size, so setup rejects a non-zero budget. The resiliency layer stays available on GRPO.
-- Reward shaping and sample filtering — `overlong_filtering`, `reward_shaping`, `reward_scaling`, and `use_dynamic_sampling` are implemented on neither algorithm block, so setup rejects them rather than silently skipping the shaping.
+- Reward shaping — `reward_shaping`, `reward_scaling`, and `use_dynamic_sampling` are implemented on neither algorithm block, so setup rejects them rather than silently skipping the shaping. Environment-flagged sample masking and `overlong_filtering` are supported.
 - The `windowed` sampler has no `over_sampling_ratio` cap — over-produced groups aged past the window are evicted, wasting rollout compute.
 - The drain gate in refit is not yet supported.
--- a/nemo_rl/algorithms/single_controller_utils/config.py
+++ b/nemo_rl/algorithms/single_controller_utils/config.py
@@ -791,7 +791,6 @@ def _validate_algo_settings(master_config: MasterConfig) -> None:
     unsupported = [
         name
         for name, enabled in (
-            ("overlong_filtering", algo_cfg.overlong_filtering),
             ("use_dynamic_sampling", algo_cfg.use_dynamic_sampling),
             ("reward_scaling", algo_cfg.reward_scaling.enabled),
             ("reward_shaping", algo_cfg.reward_shaping.enabled),
@@ -1156,6 +1155,8 @@ class AdvantageConfig:
     sample_mask_field: str = "sample_mask"
     invalid_tool_call_mask_field: str = INVALID_TOOL_CALL_MASK
     malformed_thinking_mask_field: str = MALFORMED_THINKING_MASK
+    mask_sample_field: str = "mask_sample"
+    truncated_field: str = "truncated"
     repeated_batch_fields: list[str] = field(default_factory=list)
     policy_logprobs_field: str = "prev_logprobs"
     generation_logprobs_field: str = "generation_logprobs"
--- a/nemo_rl/algorithms/single_controller_utils/utils.py
+++ b/nemo_rl/algorithms/single_controller_utils/utils.py
@@ -96,6 +96,7 @@ def reduce_advantage_pump_metrics(
     sequence_lengths: list[int],
     *,
     seq_logprob_error_metrics: list[dict[str, float]] | None = None,
+    num_mask_sample_filtered: list[int] | None = None,
     num_invalid_tool_calls: list[int] | None = None,
     num_malformed_thinking: list[int] | None = None,
     num_assistant_messages: list[int] | None = None,
@@ -108,13 +109,16 @@ def reduce_advantage_pump_metrics(
         sequence_lengths: All input_lengths trained on this step.
         seq_logprob_error_metrics: Sequence-error metrics and their aggregation
             counts, one record per streaming chunk.
+        num_mask_sample_filtered: Environment-flagged sample counts, one per
+            streaming chunk.
         num_invalid_tool_calls: Per-sample invalid tool-call counts.
         num_malformed_thinking: Per-sample malformed-thinking counts.
         num_assistant_messages: Per-sample assistant message counts (rate denominator).
 
     Returns:
         Step-level reward, advantage, token-count, optional sequence
-        log-probability error metrics, and per-sample violation counts.
+        log-probability error metrics, the num_mask_sample_filtered count, and
+        per-sample violation counts.
 
     """
     out: dict[str, float] = {}
@@ -132,6 +136,8 @@ def reduce_advantage_pump_metrics(
             out["advantages/min"] = 0.0
     if sequence_lengths:
         out["total_num_tokens"] = float(sum(sequence_lengths))
+    if num_mask_sample_filtered is not None:
+        out["num_mask_sample_filtered"] = float(sum(num_mask_sample_filtered))
     if seq_logprob_error_metrics:
         out.update(_reduce_seq_logprob_error_metrics(seq_logprob_error_metrics))
     n_asst = sum(num_assistant_messages or [])
--- a/nemo_rl/data_plane/schema.py
+++ b/nemo_rl/data_plane/schema.py
@@ -28,6 +28,8 @@ GLOBAL_FORWARD_PAD_SEQLEN = "global_forward_pad_seqlen"
 INPUT_IDS = "input_ids"
 INPUT_LENGTHS = "input_lengths"
 SAMPLE_MASK = "sample_mask"
+MASK_SAMPLE = "mask_sample"
+TRUNCATED = "truncated"
 META_IDX = "meta_idx"
 
 # Token-aligned message-violation fields consumed by SingleController advantages.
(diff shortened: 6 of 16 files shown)


tianyi-zhang-02:
Heads-up: I had #3786 open doing an overlapping thing, and yours came first and covers more — I've said so there and will cut mine down to the non-overlapping remainder rather than compete with it.

Two things from reading yours that might be useful:

**The `advantage_mask` split.** I checked whether keeping it over the full prompt group is load-bearing for GRPO, since my version narrowed `sample_mask` before the advantage stage. It isn't a value difference: `GRPOAdvantageEstimator.compute_advantage` passes `torch.ones_like(rewards)` as the valid-mask into `calculate_baseline_and_std_per_prompt`, and uses `mask` only for `advantages.expand(mask.shape)`. So both arrangements give identical advantages. Yours is still the safer one — it stops depending on that internal detail — just noting it in case a reviewer asks whether the split is required.

**One gap that survives this PR.** The *prompt*-level `DatumSpec.loss_multiplier` is still dropped on the SC path. It's set to 0.0 when the prompt itself exceeded the length budget before generation ran, and `record_to_train_batch` never reads it — `git grep loss_multiplier nemo_rl/experience/payload.py` is empty on `main`, and this PR doesn't add it. So an over-long *prompt* still trains at full weight after this lands. Completion-independent, so it doesn't fit naturally in your `mask_sample`/`truncated` shape — happy to send it separately on top of yours. :)

tianyi-zhang-02:
Correction to what I wrote above — I said #3766's coverage is a superset of this PR's. It isn't, and the difference matters in your favour, @macandro96.

I checked #3766's `config.py` hunk properly afterwards. It adds two `AdvantageConfig` field names and nothing else:

```
+    mask_sample_field: str = "mask_sample"
+    truncated_field: str = "truncated"
```

`("overlong_filtering", algo_cfg.overlong_filtering)` is still in the `unsupported` list at `single_controller_utils/config.py:739` on `main`, and #3766 does not remove it. So on that branch a config with `grpo.overlong_filtering: true` raises `NotImplementedError` at setup and its truncation-filter path is never reached. Its env-flagged `mask_sample` path works; the truncated half is unreachable until something takes that entry off the list, which is what this PR does.

So the accurate split is:

| | env `mask_sample` | truncated / `overlong_filtering` | `num_mask_sample_filtered` | prompt-level `loss_multiplier` |
|---|---|---|---|---|
| **#3766** | yes | implemented but gated off | yes | no |
| **#3786** (this) | yes | yes, and unblocks the knob | no | yes |

Which doesn't change my conclusion — #3766 came first, its column-based shape is the better one, and I'd rather not have two mechanisms for one rule. But the right outcome isn't "close this"; it's that #3766 needs the `unsupported`-list removal to make its own truncation filter live, and the prompt-level `loss_multiplier` is still dropped by both.

Happy to send those two as a small PR on top of #3766 once it lands, and close this. Or, if it's easier, take the two lines from here directly — no attribution needed. :)

nemo_rl/algorithms/single_controller.py
```diff
@@ -1704,6 +1706,17 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> tuple[KVBatchMeta, bool]:
...
+        mask_sample = squeeze_trailing_unit_dim(
+            tensor_field(data, adv_cfg.mask_sample_field)
+        ).bool()
+        truncated = squeeze_trailing_unit_dim(
+            tensor_field(data, adv_cfg.truncated_field)
+        ).bool()
+
+        # Advantage statistics use the complete prompt group. Environment,
+        # overlong, and sequence-error masks are composed only for training.
+        advantage_mask = token_mask * sample_mask.unsqueeze(-1)
```
yfw: [`single_controller.py:1718`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller.py#L1718)

**1 action item.**

**TL;DR — the split's stated purpose is already guaranteed by the estimators themselves, so it buys nothing there; what it does instead is widen the normalization pool for `reinforce_plus_plus` (and, post-rebase, for SC PPO's shipped `gae` + `normalize_advantages: true`), changing the advantages of the rows that *do* train.**

PR-introduced, and two separable effects:

- **vs pre-PR SC — a regression.** At the merge-base, `mask` was built *after* the seq-logprob-error block ([b591652 L1752](https://github.com/NVIDIA-NeMo/RL/blob/b591652624ce24a3e2eaff6e40aa02cea9c567f9/nemo_rl/algorithms/single_controller.py#L1752), [L1781](https://github.com/NVIDIA-NeMo/RL/blob/b591652624ce24a3e2eaff6e40aa02cea9c567f9/nemo_rl/algorithms/single_controller.py#L1781)). `advantage_mask` is now built *before* it, so seq-error-masked rows re-enter the advantage mask.
- **vs the legacy paths — a divergence.** [`grpo.py:3292`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/grpo.py#L3292) and [`grpo_sync.py:876`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/grpo_sync.py#L876) both build `mask` from the *fully filtered* `sample_mask` before calling the estimator.

The comment at L1716-1717 justifies the split as *"Advantage statistics use the complete prompt group."* That property already holds unconditionally — every estimator hardcodes its own valid-mask for the per-prompt baseline: [GRPO `torch.ones_like(rewards)`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/advantage_estimator.py#L85), [GDPO `torch.ones_like(...)`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/advantage_estimator.py#L155), [R++ `torch.ones_like(rewards)`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/advantage_estimator.py#L238). All three then use `mask` **only** for [`advantages.expand(mask.shape)`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/advantage_estimator.py#L98). So for `grpo`/`gdpo` the split is a literal no-op.

Where it is not a no-op is the estimators that use `mask` as a *statistics denominator*:

- `reinforce_plus_plus` — [`adv_mean = (adv * mask).sum() / mask.sum()`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/advantage_estimator.py#L262). Reachable on SC today: the estimator is built through the unrestricted [`_create_advantage_estimator`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller_utils/setup.py#L784) and there is no `adv_estimator` gate anywhere on the SC path.
- **post-rebase, SC PPO** — main routes PPO runs to [`ppo._create_advantage_estimator`](https://github.com/NVIDIA-NeMo/RL/blob/32b20cc775d9361233463c3cbbb3cbf7df4050b4/nemo_rl/algorithms/single_controller_utils/setup.py#L700), and the shipped [`ppo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/examples/configs/ppo_math_1B.yaml#L44) uses `gae` with `normalize_advantages: true` (line 47), which whitens through [`masked_mean(rewards, mask)`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/advantage_estimator.py#L351) at [`:494`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/advantage_estimator.py#L494).

One group of 4, rewards `[0, 0, 1, 10]`, row 3 dropped by `mask_sample`, R++ normalization run verbatim:

```
narrow mask (legacy):  [-0.707, -0.707, +1.414]
wide mask (this PR):   [-0.654, -0.654, -0.416]
```

The **retained** reward-1.0 row flips sign, because the dropped outlier stays in the denominator.

**Action:** collapse the split — compose all three filters into `final_sample_mask` first, then pass a single `token_mask * final_sample_mask.unsqueeze(-1)` to both `compute_advantage` and the loss (the merge-base shape, and what `grpo_sync.py:876` does). This regresses nothing, because no group baseline ever reads `mask`.

/cc @tianyi-zhang-02 — this answers the question you raised about whether the split is load-bearing. You were right that it is not for `grpo`/`gdpo`, and for exactly the reason you gave; it *is* load-bearing for the globally-normalizing estimators.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1771,28 +1782,45 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> tuple[KVBatchMeta, bool]:
...
+        # is the exact mask used when global_valid_toks and the loss are built.
+        final_mask = token_mask * final_sample_mask.unsqueeze(-1)
+        has_valid_training_tokens = bool(final_mask[:, 1:].bool().any().item())
+
         fields_to_put = {adv_cfg.output_field: advantages}
-        if seq_logprob_error_threshold is not None:
-            fields_to_put[adv_cfg.sample_mask_field] = sample_mask
+        if (
+            seq_logprob_error_threshold is not None
+            or self._master_config.grpo.overlong_filtering
```
yfw: [`single_controller.py:1820`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller.py#L1820)

**1 action item** — post-rebase only; nothing is wrong at this HEAD.

This read and the one at [L1807](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller.py#L1807) will raise `AttributeError: 'NoneType' object has no attribute 'overlong_filtering'` on a SingleController **PPO** run once you rebase. Main converted every `.grpo` read in this file to `self._algo_cfg` ([`main:155`](https://github.com/NVIDIA-NeMo/RL/blob/32b20cc775d9361233463c3cbbb3cbf7df4050b4/nemo_rl/algorithms/single_controller.py#L155) — and [`main:1983`](https://github.com/NVIDIA-NeMo/RL/blob/32b20cc775d9361233463c3cbbb3cbf7df4050b4/nemo_rl/algorithms/single_controller.py#L1983) is this very line, the neighbouring `seq_logprob_error_threshold` read), while [`MasterConfig.grpo`](https://github.com/NVIDIA-NeMo/RL/blob/32b20cc775d9361233463c3cbbb3cbf7df4050b4/nemo_rl/algorithms/single_controller_utils/config.py#L543) is `Optional[GRPOConfig] = None` and `validate_algorithm_block` guarantees exactly one algorithm block is set.

I simulated the merge (`git merge-file` against merge-base `b591652` and main `32b20cc`): it yields 4 conflicts, and this line lands inside one — **but L1807 merges silently**, leaving it as one of only two `_master_config.grpo` reads left in the file.

**Action:** use `self._algo_cfg.overlong_filtering` at **both** sites when you rebase. [`PPOConfig.overlong_filtering`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/ppo.py#L183) exists, so it is a drop-in.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1771,28 +1782,45 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> tuple[KVBatchMeta, bool]:
...
         self._step_log_dict["masked_advantages"].append(
             response_advantages.detach().cpu()
         )
 
+        num_mask_sample_filtered = int(mask_sample.sum().item())
+        self._step_log_dict["num_mask_sample_filtered"].append(num_mask_sample_filtered)
+        final_sample_mask = final_sample_mask * (~mask_sample).to(
+            final_sample_mask.dtype
+        )
+        if self._master_config.grpo.overlong_filtering:
```
yfw: [`single_controller.py:1807`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller.py#L1807)

**1 action item** — post-rebase.

Corroborating @tianyi-zhang-02's point in the thread above, with two data points that make this easy to miss. On `main`, `overlong_filtering` is still in the `unsupported` list at [`config.py:739`](https://github.com/NVIDIA-NeMo/RL/blob/32b20cc775d9361233463c3cbbb3cbf7df4050b4/nemo_rl/algorithms/single_controller_utils/config.py#L739) and `_validate_algo_settings` raises `NotImplementedError`. It is not conditional — `validate_single_controller_config` calls it, which is the first statement of `setup_single_controller`, which `examples/run_grpo_single_controller.py` calls unconditionally. [`docs/guides/single-controller.md:185`](https://github.com/NVIDIA-NeMo/RL/blob/32b20cc775d9361233463c3cbbb3cbf7df4050b4/docs/guides/single-controller.md#L185) documents it as rejected too.

1. **`config.py` merges with zero conflicts** (`git merge-file` against `b591652`/`32b20cc`, exit 0) — so the rebase puts no marker anywhere near line 739 and the entry survives untouched.
2. **The PR's tests stay green regardless** — they inject `_master_config = SimpleNamespace(grpo=SimpleNamespace(...))` directly and never construct a `MasterConfig`, so setup validation never runs. Neither the merge nor CI will surface it.

**Action:** when you rebase, drop the `("overlong_filtering", algo_cfg.overlong_filtering)` entry and update the docs line — otherwise this entire `truncated` branch is unreachable and `grpo.overlong_filtering: true` still raises at setup.

**Context — no action.** One thing worth knowing when you do: SC's `truncated` only matches legacy's `hit_max_tokens` on the **vLLM** backend. [`_generation_max_seq_len`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller_utils/setup.py#L351) dispatches on `generation_config["backend"]`, whereas legacy dispatches on which sub-config key is present ([`rollouts.py:2625-2641`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/experience/rollouts.py#L2625-L2641)); for `megatron` and `sglang` the two resolve to different quantities. Pre-existing, and only observable once this filter is live.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1771,28 +1782,45 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> tuple[KVBatchMeta, bool]:
...
+
         fields_to_put = {adv_cfg.output_field: advantages}
-        if seq_logprob_error_threshold is not None:
-            fields_to_put[adv_cfg.sample_mask_field] = sample_mask
+        if (
+            seq_logprob_error_threshold is not None
+            or self._master_config.grpo.overlong_filtering
+            or num_mask_sample_filtered > 0
+        ):
+            fields_to_put[adv_cfg.sample_mask_field] = final_sample_mask
```
yfw: [`single_controller.py:1818-1823`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller.py#L1818-L1823)

**1 action item, 1 optional.**

**TL;DR — both disjuncts this PR adds to the write-back gate are unpinned: delete either one, or drop the `overlong_filtering` guard at L1807 entirely, and all four of the PR's advantage-stage tests still pass.**

If a disjunct is ever lost, `final_sample_mask` is computed, counted into `num_mask_sample_filtered`, folded into `has_valid_training_tokens` — and then never written back, so training proceeds on the unmasked batch with no error. That is the silent-wrong-result shape, not a crash.

Mutation-tested against a pinned snapshot of this HEAD, driving the PR's own fixtures. Three mutants survive: dropping `or num_mask_sample_filtered > 0`, dropping `or self._master_config.grpo.overlong_filtering`, and dropping the `if ...overlong_filtering:` guard at [L1807](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller.py#L1807). Six other mutants in the same method are correctly killed, so this is a targeted hole, not thin coverage generally.

The cause is [`test_single_controller.py:350`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/tests/unit/single_controller/test_single_controller.py#L350): the only test exercising the new masks sets `seq_logprob_error_threshold=2.0` **and** `overlong_filtering=True` **and** a non-empty `mask_sample`, so all three disjuncts are true at once and none is isolated.

**### AI-1 — please add a test that isolates each disjunct.** Verified passing on unmodified HEAD, and red under all three mutations above. Add to `tests/unit/single_controller/test_single_controller.py` (it uses only names already imported there); this can't be a `suggestion` block because the insertion point is outside every diff hunk in that file.

```python
@pytest.mark.parametrize(
    "overlong_filtering, mask_sample, truncated, expected_sample_mask",
    [
        (False, [True, False], [True, True], [0.0, 1.0]),
        (True, [False, False], [False, True], [1.0, 0.0]),
    ],
    ids=["env_mask_only", "overlong_only"],
)
def test_advantage_stage_writes_sample_mask_without_seq_logprob_threshold(
    overlong_filtering: bool,
    mask_sample: list[bool],
    truncated: list[bool],
    expected_sample_mask: list[float],
) -> None:
    batch_size, sequence_length = 2, 5
    data = TensorDict(
        {
            "prompt_ids_for_adv": torch.zeros(
                batch_size, sequence_length, dtype=torch.long
            ),
            "total_reward": torch.tensor([1.0, 0.0]),
            "token_mask": torch.ones(batch_size, sequence_length),
            "sample_mask": torch.ones(batch_size),
            "mask_sample": torch.tensor(mask_sample),
            "truncated": torch.tensor(truncated),
        },
        batch_size=[batch_size],
    )
    data_plane = _AdvantageDataPlane(data)
    estimator = _MaskRecordingAdvantageEstimator()

    controller_cls = SingleControllerActor.__ray_metadata__.modified_class
    ctrl = object.__new__(controller_cls)
    ctrl._dp_client = data_plane
    ctrl._advantage_cfg = AdvantageConfig()
    ctrl._advantage_estimator = estimator
    ctrl._policy_logprobs_required = False
    ctrl._reference_logprobs_required = False
    ctrl._master_config = SimpleNamespace(
        grpo=SimpleNamespace(
            seq_logprob_error_threshold=None,
            overlong_filtering=overlong_filtering,
        )
    )
    ctrl._step_log_dict = {
        "rewards": [],
        "masked_advantages": [],
        "num_mask_sample_filtered": [],
        "sequence_lengths": [],
        "seq_logprob_error_metrics": [],
    }
    meta = KVBatchMeta(
        partition_id="rollout_data",
        task_name="train",
        sample_ids=[f"sample-{i}" for i in range(batch_size)],
        fields=list(data.keys()),
    )

    _, has_valid_training_tokens = asyncio.run(ctrl._advantage_stage(meta))

    assert has_valid_training_tokens
    assert data_plane.written_fields is not None
    assert "sample_mask" in data_plane.written_fields
    assert torch.equal(
        data_plane.written_fields["sample_mask"],
        torch.tensor(expected_sample_mask),
    )
    assert estimator.mask is not None
    assert estimator.mask.all()
```

**### Optional — one way of several** to make that mutation class impossible, by deriving the gate from the data instead of restating the three conditions:

```suggestion
        if not torch.equal(final_sample_mask, sample_mask):
            fields_to_put[adv_cfg.sample_mask_field] = final_sample_mask
```

Caveat worth knowing before you take it: this is **not** byte-for-byte behaviour-preserving. With `seq_logprob_error_threshold` set but nothing actually masked, HEAD writes `sample_mask` and this does not. The written value is identical and `sample_mask` is not added to `new_fields` in either version, so nothing downstream can observe the difference — but no test covers that case either way. Suite is 361 passed with or without it.

nemo_rl/experience/payload.py
```diff
@@ -80,6 +81,18 @@ def record_to_train_batch(
     total_reward = torch.tensor(
         [float(c.reward) for c in completions], dtype=torch.float32
     )
+    mask_sample = torch.tensor(
+        [
+            bool(
+                ((c.env_extras or {}).get("instance_config") or {}).get(
```
yfw: [`payload.py:87`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/experience/payload.py#L87)

**1 action item** — low severity.

This re-implements [`_extract_mask_sample_flags`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/experience/rollouts.py#L266-L278), and it is the *same dict* in both cases — `Completion.env_extras` is assigned `result["full_result"]` at [`rollout_manager.py:1027`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/experience/rollout_manager.py#L1027), so the two differ only in the accessor.

The two copies have **already drifted textually**: `rollouts.py:271` uses the string literal `"mask_sample"`, while this one uses the imported `MASK_SAMPLE` schema constant. That is the concrete hazard — the `instance_config` key path now lives in two places that are already not written the same way.

**Action:** extract a shared helper (e.g. `def _mask_sample_flags(extras: Iterable[dict | None]) -> torch.Tensor`) and have both call it — one generator expression at each call site (`(r["full_result"] for r in results)` / `(c.env_extras for c in completions)`). No import-cycle risk: this function already lazy-imports `backfill_missing_routed_experts` from `experience.rollouts` at [L54](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/experience/payload.py#L54), so the helper can ride that same import.

nemo_rl/experience/payload.py
```diff
@@ -40,8 +40,9 @@ def record_to_train_batch(
         pad_value_dict: Field-name → pad value used by batched_message_log_to_flat_message.
 
     Returns:
-        BatchedDataDict with input_ids, input_lengths, generation_logprobs, token_mask,
-        sample_mask, prompt_ids_for_adv, total_reward, and optional routed_experts.
+        BatchedDataDict with input_ids, input_lengths, generation_logprobs,
+        token_mask, the final-loss sample_mask seed, raw environment/truncation
+        masks, prompt_ids_for_adv, total_reward, and optional routed_experts.
```
yfw: [`payload.py:43-45`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/experience/payload.py#L43-L45)

**1 action item** — low severity, docs only.

The rewritten `Returns:` describes the two new columns as "the final-loss sample_mask seed, raw environment/truncation masks", which names neither key and is not greppable — and it *removed* the literal `sample_mask` that was there before. It's also slightly misleading: `sample_mask` here is literally `torch.ones(n)`.

```suggestion
        BatchedDataDict with input_ids, input_lengths, generation_logprobs,
        token_mask, an all-ones sample_mask, the raw mask_sample and truncated
        flags, prompt_ids_for_adv, total_reward, and optional routed_experts.
```

Same class, no suggestion possible (outside the hunk): [`reduce_advantage_pump_metrics`'s `Returns:`](https://github.com/NVIDIA-NeMo/RL/blob/bf663944a29e3ebf83b248422006e07c276b2860/nemo_rl/algorithms/single_controller_utils/utils.py#L111-L113) still omits the `num_mask_sample_filtered` key it now always emits.

yfw (commented):
Team review of the mask-composition change: 6 agents (RL expert, bug finder, test agent, design reviewer, comment reviewer, devil's advocate), all findings adversarially challenged before posting. Thanks for this — the env `mask_sample` gap on the SingleController path is a real parity hole and the column-based shape is the right way to close it.

**Please rebase first.** This branch is currently `CONFLICTING` and 35 commits behind `main`. Three of the six comments below are *only* about what happens after that rebase, and two of them describe changes that no merge conflict and no CI run will point at — I simulated the merge to check.

The description is the unfilled template; a sentence on which recipe you ran, and on whether the advantage-mask split is intentional, would resolve most of what's below.

Deliberately **not** asking for convergence curves: the one open question (the advantage-mask split) is semantic rather than empirical — a curve wouldn't tell us whether normalizing over dropped rows is intended. Worth answering yes/no first.

Verified clean and worth noting so nobody re-raises them: `pre-commit run --all-files` passes every hook; the `mask_sample` key path matches Gym and the legacy reader exactly; `env.should_mask_flagged_samples=false` is already honoured upstream at `rollout_manager.py:1021` so no second config knob was needed; adding both fields to `PROMOTE_1D_FIELDS` was mandatory, not optional (the Mooncake adapter rejects undeclared dense-1D tensors) and the round-trip test pins it; `num_mask_sample_filtered` matches the legacy metric name and emission semantics; and `AdvantageConfig` is correctly a `@dataclass` of internal column names, so no exemplar-YAML update is owed.

_Generated by Claude Code_

macandro96:
/ok to test [48121bb](https://github.com/NVIDIA-NeMo/RL/pull/3766/commits/48121bbc9df1299a6fcc3610998355b30eda1e4c)

nemo_rl/experience/payload.py
macandro96: addressed in [48121bb](https://github.com/NVIDIA-NeMo/RL/pull/3766/commits/48121bbc9df1299a6fcc3610998355b30eda1e4c)

nemo_rl/experience/payload.py
macandro96: addressed in [48121bb](https://github.com/NVIDIA-NeMo/RL/pull/3766/commits/48121bbc9df1299a6fcc3610998355b30eda1e4c)

nemo_rl/algorithms/single_controller.py
macandro96: addressed in [48121bb](https://github.com/NVIDIA-NeMo/RL/pull/3766/commits/48121bbc9df1299a6fcc3610998355b30eda1e4c)

nemo_rl/algorithms/single_controller.py
macandro96: addressed in [48121bb](https://github.com/NVIDIA-NeMo/RL/pull/3766/commits/48121bbc9df1299a6fcc3610998355b30eda1e4c)

nemo_rl/algorithms/single_controller.py
macandro96: addressed in [48121bb](https://github.com/NVIDIA-NeMo/RL/pull/3766/commits/48121bbc9df1299a6fcc3610998355b30eda1e4c)

nemo_rl/algorithms/single_controller.py
macandro96: addressed in [48121bb](https://github.com/NVIDIA-NeMo/RL/pull/3766/commits/48121bbc9df1299a6fcc3610998355b30eda1e4c)

yfw: fixed in 48121bbc9df1299a6fcc3610998355b30eda1e4c

nemo_rl/algorithms/single_controller.py
yfw: fixed in 48121bbc9df1299a6fcc3610998355b30eda1e4c

nemo_rl/algorithms/single_controller.py
yfw: fixed in 48121bbc9df1299a6fcc3610998355b30eda1e4c

nemo_rl/algorithms/single_controller.py
yfw: fixed in 48121bbc9df1299a6fcc3610998355b30eda1e4c

nemo_rl/experience/payload.py
yfw: fixed in 48121bbc9df1299a6fcc3610998355b30eda1e4c

nemo_rl/experience/payload.py
yfw: fixed in 48121bbc9df1299a6fcc3610998355b30eda1e4c

yfw (dismissed):

macandro96:
/ok to test [5c540c3](https://github.com/NVIDIA-NeMo/RL/commit/5c540c34bced5c2d6f53cd114ee18bf5ebec3cd3)

macandro96:
/ok to test [7482514](https://github.com/NVIDIA-NeMo/RL/commit/748251432c4432e9b13e355b4e6bae334c8fb4c0)

yfw (approved):

PR #3768, opened 2026-08-22
feat: add MOPD to single-controller text path

yfw:
# What does this PR do ?

Adds text-only Multi-Teacher On-Policy Distillation (MOPD) support to the new SingleController GRPO path.

Previously, MOPD was supported only through the legacy controller. This ports the feature to the TransferQueue-based SingleController flow while keeping teacher inference non-colocated.

## Implementation

- Validates the SingleController MOPD configuration before allocating resources:
  - `on_policy_distillation.enabled=true`
  - `grpo.adv_estimator.name=opd`
  - NeMo Gym rollouts
  - non-colocated teachers
- Reserves topology-constrained policy and teacher resources before other workers can consume them.
- Creates and attaches teacher worker groups to the SingleController data plane.
- Routes each completed prompt group to its teacher using NeMo Gym `agent_ref` metadata.
- Runs teacher inference as a required post-write TQ enrichment stage:
  1. rollout tensors are written to TQ;
  2. teacher workers fetch their presharded rows;
  3. teachers write `teacher_reference_logprobs` back to the same rows;
  4. the rollout becomes available for training.
- Builds sequence-packing or dynamic-batching plans globally on the driver using `logprob_mb_tokens`.
- Serializes requests targeting the same physical teacher while allowing distinct teachers to run concurrently.
- Handles teacher DP padding, cleanup, cancellation, rollback, and grouped enrichment failures.
- Feeds student and teacher logprobs into the OPD advantage estimator.
- Pools OPD metrics exactly across unequal streaming chunks using sum, sum-of-squares, and token count.
- Warms the long-lived TQ partition schema before concurrent writers start.

## Limitations

- This PR supports text-only MOPD. Multimodal/VLM teacher inputs are not yet forwarded through the SingleController TQ path.
- MOPD currently requires NeMo Gym and non-colocated teacher workers.
- Reference-logprob field narrowing is handled separately by #3742.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...


nemo_rl/models/policy/teacher_worker_group.py
```diff
@@ -240,6 +246,82 @@ def __init__(
...
+
+    def get_logprobs_from_meta(self, meta: KVBatchMeta) -> None:
+        """Dispatch a TQ metadata batch; workers fetch and write teacher logprobs."""
+        if not meta.sequence_lengths:
+            raise ValueError(
+                f"Teacher {self.alias!r} requires sequence_lengths in TQ metadata"
+            )
+        sequence_pad_multiple = (
+            1 if self.use_sequence_packing else (self.sequence_length_pad_multiple)
+        )
```
yfw, line 282: **1 action item — please fix in this PR.**

**TL;DR — with dynamic batching enabled, the teacher's stamped fetch width can be smaller than the driver's planned microbatch length, so the first teacher dispatch crashes in `torch.narrow` and aborts the run.**

[`teacher_worker_group.py:275`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/teacher_worker_group.py#L275-L287) — PR-introduced (the dyn-batching plumbing in `TeacherWorkerGroup` is new in this diff).

Background: the driver plans dynamic-batch microbatches from metadata alone, rounding each microbatch length up to `sequence_length_round` with a deliberately non-binding width cap ([`preshard.py:99-109`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/data_plane/preshard.py#L99-L109)); workers then fetch padded to `GLOBAL_FORWARD_PAD_SEQLEN` and narrow tensors to the planned lengths. The student upholds `mb_len <= fetched width` by stamping `round_up(max_len, max(pad_mult, sequence_length_round))` in [`TQPolicy._stamp_pad_seqlen`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/tq_policy.py#L202-L217); the teacher stamp here uses only the tp/cp multiple.

Walked failure (executed, red on this branch):
1. `dynamic_batching.sequence_length_round=64` (the base exemplar default), teacher tp=1 -> multiple 1; batch seqlens `[100, 60]`.
2. Planner emits `micro_batch_lengths=[[128]]` (`round_up(100, 64)`); teacher stamps width `round_up(100, 1) = 100`.
3. Worker `_fetch` materializes width 100; `truncate_tensors` -> `RuntimeError: start (0) + length (128) exceeds dimension size (100)`.
4. Enrichment fails -> `PostWriteEnrichmentError` -> run aborts (post-write failures are correctly not retried).

The shipped `-pack` recipe never reaches this branch, and `test_get_logprobs_from_meta_builds_global_dynamic_batch_plan` uses `sequence_length_round=1` — the one value that hides the mismatch.

### AI-1
```suggestion
        sequence_pad_multiple = (
            1 if self.use_sequence_packing else (self.sequence_length_pad_multiple)
        )
        if self.use_dynamic_batches:
            # Mirror TQPolicy._stamp_pad_seqlen: the driver's dynamic-batching
            # plan rounds microbatch lengths to sequence_length_round, so the
            # fetched width must cover that round or truncate_tensors narrows
            # past the fetched tensor.
            sequence_pad_multiple = max(
                sequence_pad_multiple,
                int(self.cfg["dynamic_batching"]["sequence_length_round"]),
            )
```
Verified: a probe driving the real `get_logprobs_from_meta` fails on the current code (`planned mb length 128 exceeds fetched width 100`) and passes with this edit; all 46 teacher/OPD unit tests stay green; ruff/ruff-format clean; pyrefly unchanged. Consider also strengthening the dyn-batch unit test to `sequence_length_round=64` with an assertion that the stamped pad covers every planned microbatch length.

**Context — no action.** This pre-dispatch preamble is a structural near-copy of [`TQPolicy._logprob_dispatch`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/tq_policy.py#L270-L326); sharing the stamp/packing-resolution helpers (idempotence included — the teacher version also overrides an incoming stamp where the student's is idempotent) would prevent this class of drift.

nemo_rl/data_plane/worker_mixin.py
```diff
@@ -525,6 +526,41 @@ def get_reference_policy_logprobs_presharded(
...
+            raise RuntimeError(
+                "SingleController teacher batching requires driver-provided global "
+                "micro_batch_indices and micro_batch_lengths; local worker planning "
+                "can desynchronize data-parallel collectives."
+            )
+        data = self._attach_or_repack_pack_metadata(data, meta)
+        result: BatchedDataDict[Any] = self.get_logprobs(  # type: ignore[attr-defined]
+            data=data,
+            micro_batch_size=micro_batch_size,
+        )
```
yfw, line 555: **1 action item.**

**TL;DR — with an MoE student running `policy.router_replay.enabled=true`, every SC MOPD teacher call raises, because the teacher inherits the student's router-replay config but its TQ fetch never includes `routed_experts`.**

[`worker_mixin.py:552`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/data_plane/worker_mixin.py#L552-L556) — PR-introduced.

Router replay exists to keep logprobs consistent between the student's inference (rollout) and training passes; a frozen teacher has no training pass to stay consistent with, so replay is irrelevant for teachers regardless of the student's setting.

Chain to the crash: the teacher deep-copies the student policy config and sanitizes dtensor/peft/draft/quant but not `router_replay` ([`teacher_worker_group.py:159-179`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/teacher_worker_group.py#L159-L179)); the teacher meta narrows fields to `["input_ids", "input_lengths"]` ([`teacher_worker_group.py:281`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/teacher_worker_group.py#L281)); this call inherits the default [`require_router_replay=True`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1800) and [the gate](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/workers/megatron_policy_worker.py#L124-L141) raises whenever enabled AND required AND routes absent. Committed r3 recipes do enable router replay; none combine it with SC MOPD today, so this is a plausible-config, fail-loud crash rather than something the shipped recipe hits.

### AI-1
Disable router replay in the teacher config build, next to the other student-feature scrubs in [`teacher_worker_group.py:159-179`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/teacher_worker_group.py#L159-L179):

```python
        # Router replay keeps the student's rollout and training logprobs
        # consistent; a frozen teacher has no training pass, and its TQ fetch
        # never carries routed_experts, so replay must stay off.
        if "router_replay" in cfg:
            cfg["router_replay"]["enabled"] = False
```

Verified: 46 teacher/OPD tests green, ruff/ruff-format clean with this applied. It is backend-agnostic and the teacher worker never even initializes replay state.

**Context — no action.** The alternative — passing `require_router_replay=False` at this call site, mirroring the frozen-reference path ([`megatron_policy_worker.py:1145`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1145)) — also works (verified), but has two hazards: it is a paired edit (the fake in `test_teacher_worker_presharded_entrypoint_writes_teacher_tq_field` TypeErrors on the new kwarg), and DTensor workers' `get_logprobs` does not accept the kwarg, so it plants a latent `TypeError` in this shared mixin — safe today only because teachers are Megatron-only by construction. Doing both (sanitize + kwarg with the paired test edit) is fine too, for symmetry with the reference path.

tests/test_suites/llm/mopd-qwen3-1.7b-3n8g-megatron-pack-single-controller.sh
```diff
@@ -0,0 +1,60 @@
...
+    logger.tensorboard_enabled=True \
+    checkpointing.enabled=False \
+    checkpointing.checkpoint_dir=$CKPT_DIR \
+    "$@" \
+    2>&1 | tee $RUN_LOG
+
+uv run tests/json_dump_tb_logs.py $LOG_DIR --output_path $JSON_METRICS
+
+if [[ $(jq 'to_entries | .[] | select(.key == "train/token_mult_prob_error") | .value | keys | map(tonumber) | max' $JSON_METRICS) -ge $MAX_STEPS ]]; then
+    uv run tests/check_metrics.py $JSON_METRICS \
```
yfw, line 52: **1 action item (evidence, not code).**

[`mopd-...-single-controller.sh:52`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/tests/test_suites/llm/mopd-qwen3-1.7b-3n8g-megatron-pack-single-controller.sh#L39-L58) — this gate is a self-distillation smoke test (student == teacher, loss stays near zero), which proves the plumbing but by design cannot demonstrate distillation gains or catch a systematic bias in the teacher column.

### AI-1
Could you add to the PR description a short side-by-side run (5+ steps) of this SC recipe vs the legacy `mopd-qwen3-1.7b-3n8g-megatron-pack.sh` on the same data — the `train/loss`, `train/token_mult_prob_error`, and `train/on_policy_distillation/{adv_mean,adv_std,teacher_student_logprob_gap_mean}` traces should overlay — plus one step-time or `teacher_logprob_time_s` number so the serialized-teacher cost of the required post-write enrichment stage is on record?

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -165,19 +225,84 @@ def _build_clusters(
...
+                node_id: topology[node_id] for node_id in remaining_ids
+            }
+            generation_config_dict = cast(dict[str, Any], generation_config)
+            if backend == "vllm":
+                vllm_cfg = generation_config_dict["vllm_cfg"]
+                gpus_per_instance = vllm_cfg["tensor_parallel_size"] * vllm_cfg.get(
+                    "pipeline_parallel_size", 1
+                )
+            else:
+                gpus_per_instance = generation_config_dict["sglang_cfg"].get(
```
yfw, line 297: **1 action item (low — diagnostics only).**

[`setup.py:276`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller_utils/setup.py#L270-L278) — PR-introduced. This else-branch assumes `sglang_cfg` for **any** non-vllm backend, and `_build_clusters` runs before `_build_generation`'s clean rejection ([`"only supports vllm or sglang"`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller_utils/setup.py#L378-L381)). A trtllm/unknown-backend config with `cluster.segment_size` set, non-colocated, `policy_nodes > 1` dies with a bare `KeyError: 'sglang_cfg'` instead of the intended `ValueError` (Megatron can't reach it — [assert at `setup.py:216`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller_utils/setup.py#L216)). The config is invalid either way — wrong error, not wrong behavior. The legacy twin in `grpo.py` dispatches all five backends here, so there is no sibling precedent for the gap.

### AI-1
Make the dispatch explicit — `elif backend == "sglang": ...` / `else: raise ValueError(...)` with the same message `_build_generation` uses.

nemo_rl/algorithms/opd.py
```diff
@@ -200,6 +208,244 @@ def get_teacher_routing_metrics(
...
+        meta: KVBatchMeta,
+        record: PromptGroupRecord,
+    ) -> KVBatchMeta:
+        """Write teacher logprobs before the replay-buffer slot becomes ready."""
+        alias, group_alias = self._resolve_teacher(record)
+        started_at = time.perf_counter()
+        # asyncio cannot cancel a running thread. Shield and explicitly drain
+        # it so replay-buffer rollback never clears rows while teacher workers
+        # are still fetching from or writing to those rows.
+        enrichment_task = asyncio.create_task(
```
yfw: **1 action item (low).**

[`opd.py:391`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/opd.py#L388-L403) — PR-introduced. Same-teacher serialization holds a `threading.Lock` *inside* the `asyncio.to_thread` thread, so every waiter occupies a default-executor slot (max `min(32, cpus + 4)`). With `async_rl.max_inflight_prompts` above the pool size and one slow teacher, later `to_thread` work — including the train pump's `finish_train_step` ([`single_controller.py:1107`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller.py#L1107)) — queues FIFO behind blocked enrichment threads. This is a transient stall of roughly (queued enrichments ahead) x (teacher inference time), not a deadlock (the lock holder progresses via `ray.get` with no executor dependency). The shipped recipe (8 inflight) is unaffected; 32+ prompts/step is an ordinary setting.

### AI-1
Acquire a per-group-alias `asyncio.Lock` in `enrich` *before* `asyncio.to_thread`, so waiters cost zero threads and only actively running teacher inferences occupy the pool.

nemo_rl/algorithms/opd.py
```diff
@@ -200,6 +208,244 @@ def get_teacher_routing_metrics(
...
+
+        teacher_model_by_agent_name = self._opd_cfg["teacher_model_by_agent_name"]
+        self._teacher_batches += 1
+        self._teacher_samples += meta.size
+        self._teacher_logprob_time_s += total_time_s
+        self._teacher_inference_time_s += inference_time_s
+        self._teacher_lock_wait_time_s += lock_wait_s
+        self._aliases_seen.add(alias)
+        self._models_seen.add(teacher_model_by_agent_name[alias])
+        record.rollout_metrics["teacher_logprob_time"] = total_time_s
```
yfw: **1 action item (low).**

[`opd.py:414`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/opd.py#L414) — PR-introduced. This write is dead on the SC path: the enricher runs inside `commit()` after [`record_to_train_batch`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/async_utils/replay_buffer.py#L812) has already tensorized the record, per-group rollout-metrics logging is an explicit [TODO](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller.py#L1254-L1256), and all `.rollout_metrics` readers are legacy-path only. The timing itself isn't lost — `drain_metrics()` reports `teacher_logprob_time_s`.

### AI-1
Delete the line, or add a one-line comment that it is forward-provisioning for the SC rollout-metrics TODO — as written it looks load-bearing but isn't.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1789,6 +1852,15 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> tuple[KVBatchMeta, bool]:
         self._step_log_dict["masked_advantages"].append(
             response_advantages.detach().cpu()
         )
+        if self._teacher_logprobs_required:
+            valid = response_advantages.detach().double()
+            self._opd_stat_sum += float(valid.sum())
+            self._opd_stat_sumsq += float((valid * valid).sum())
+            self._opd_stat_count += int(valid.numel())
+        else:
+            estimator_metrics = getattr(self._advantage_estimator, "last_metrics", {})
```
yfw: **1 action item (low).**

[`single_controller.py:1861`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller.py#L1859-L1863) — PR-introduced. This else-branch is dead in every configuration: `last_metrics` exists only on `OPDAdvantageEstimator` ([`advantage_estimator.py:574`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/advantage_estimator.py#L574), pre-existing), and config validation forces OPD estimator <-> `on_policy_distillation.enabled` ([`config.py:768-798`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller_utils/config.py#L768-L798)), so an estimator with `last_metrics` always takes the pooled-stats branch instead. And if a future estimator ever made it live, the drain at [`:1114-1121`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller.py#L1114-L1121) averages chunks **unweighted** — exactly the bias your pooled sum/sumsq/count path (which is correct, and nicely tested) exists to avoid.

### AI-1
Delete the branch and `_advantage_metric_values` (7 test fixtures currently have to initialize it), or keep it with a one-line comment stating it is reserved and weight the drain by token count.

nemo_rl/data_plane/schema.py
```diff
@@ -44,6 +44,18 @@
     "sample_mask",
 )
 
+# Full known tensor schema for SingleController's long-lived rollout partition.
+# The initial rollout put writes the first seven payload fields; later stages add
+# student/reference logprobs, advantages, and the MOPD teacher column. Registering
+# their names once before concurrent producers start avoids TransferQueue's lazy
+# field-name registration race.
+SC_ROLLOUT_SCHEMA_FIELDS = (
```
yfw, line 52: **1 action item (nit).**

The teacher input subset `["input_ids", "input_lengths"]` is inlined at [`opd.py:316`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/opd.py#L316), [`opd.py:334`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/opd.py#L334), and [`teacher_worker_group.py:281`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/teacher_worker_group.py#L281); the exact analog [`LP_SEED_FIELDS`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/data_plane/schema.py#L60) names the student fetch subset once.

### AI-1
Add `TEACHER_LP_FIELDS = ("input_ids", "input_lengths")` here beside `LP_SEED_FIELDS` and reference it from the three sites — then multimodal MOPD (currently gated at [`setup.py:683-687`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller_utils/setup.py#L683-L687)) becomes a one-tuple change plus the gate.

tests/unit/models/policy/test_teacher_worker_group.py
```diff
@@ -77,3 +90,195 @@ def test_create_teacher_configs_deduplicates():
...
+    kwargs = call.kwargs
+    assert call.args == ("get_teacher_logprobs_presharded",)
+    assert [shard.sample_ids for shard in kwargs["meta"]] == [["a"], ["b"]]
+    assert all(
+        shard.extra_info[GLOBAL_FORWARD_PAD_SEQLEN] == 6 for shard in kwargs["meta"]
+    )
+    worker_group.get_all_worker_results.assert_called_once_with("futures")
+
+
+def test_get_logprobs_from_meta_builds_global_dynamic_batch_plan(monkeypatch):
```
yfw, line 193: **1 action item (test).**

The teacher's sequence-packing dispatch branch ([`teacher_worker_group.py:291-295`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/models/policy/teacher_worker_group.py#L291-L295)) has no test — and it is the branch the shipped `-pack` nightly actually runs. Its load-bearing behaviors (max tokens from `sequence_packing.logprob_mb_tokens`, not `train_mb_tokens`; `GLOBAL_FORWARD_PAD_SEQLEN` skipping pad-multiple rounding because the packer owns padding) currently have no guard.

### AI-1
Packing-branch sibling of the dyn-batch test below — verified passing against this branch (all imports already at the top of this file); the parameters are discriminating (pad multiple 16 vs raw max 8; train 999 vs logprob 16):

```python
def test_get_logprobs_from_meta_builds_global_sequence_packing_plan(monkeypatch):
    """Teacher packing plans use logprob tokens and skip global pad rounding."""
    import nemo_rl.models.policy.teacher_worker_group as teacher_module
    from nemo_rl.models.policy.teacher_worker_group import TeacherWorkerGroup

    class Sharding:
        def get_axis_size(self, axis):
            assert axis == "data_parallel"
            return 2

    captured = {}
    real_shard_meta_for_dp = teacher_module.shard_meta_for_dp

    def capture_plan(meta, **kwargs):
        captured.update(kwargs)
        captured["meta"] = meta
        return real_shard_meta_for_dp(meta, **kwargs)

    monkeypatch.setattr(teacher_module, "shard_meta_for_dp", capture_plan)
    worker_group = MagicMock()
    teacher = object.__new__(TeacherWorkerGroup)
    teacher.alias = "teacher"
    teacher.use_sequence_packing = True
    teacher.use_dynamic_batches = False
    teacher.sequence_length_pad_multiple = 16
    teacher.sequence_packing_args = {
        "algorithm": "modified_first_fit_decreasing",
        "input_key": "input_ids",
        "input_lengths_key": "input_lengths",
        "sequence_length_pad_multiple": 16,
    }
    teacher.cfg = {
        "sequence_packing": {
            "enabled": True,
            "train_mb_tokens": 999,
            "logprob_mb_tokens": 16,
        }
    }
    teacher.sharding_annotations = Sharding()
    teacher.worker_group = worker_group
    teacher._micro_batch_size = 1
    meta = KVBatchMeta(
        partition_id="rollout_data",
        task_name="train",
        sample_ids=["a", "b", "c", "d"],
        fields=["input_ids", "input_lengths"],
        sequence_lengths=[8, 1, 7, 2],
    )

    teacher.get_logprobs_from_meta(meta)

    assert captured["dynamic_batching_args"] is None
    assert captured["sequence_packing_args"]["max_tokens_per_microbatch"] == 16
    # The packer owns padding, so the global forward pad is the raw max seqlen.
    assert captured["meta"].extra_info[GLOBAL_FORWARD_PAD_SEQLEN] == 8
    shards = worker_group.run_all_workers_sharded_data.call_args.kwargs["meta"]
    assert sorted(sample_id for shard in shards for sample_id in shard.sample_ids) == [
        "a",
        "b",
        "c",
        "d",
    ]
    assert all(MICRO_BATCH_INDICES in shard.extra_info for shard in shards)
    assert all(MICRO_BATCH_LENGTHS in shard.extra_info for shard in shards)
```

tests/unit/single_controller/test_single_controller.py
```diff
@@ -555,6 +566,111 @@ def test_advantage_stage_skips_preexisting_empty_mask_without_seq_threshold() ->
...
+
+        def get_samples(self, sample_ids, partition_id, select_fields):
+            del sample_ids, partition_id
+            assert "teacher_reference_logprobs" in select_fields
+            assert "generation_logprobs" in select_fields
+            return TensorDict(
+                {
+                    "prompt_ids_for_adv": torch.zeros(2, 3, dtype=torch.long),
+                    "total_reward": torch.zeros(2),
+                    "token_mask": torch.ones(2, 3),
```
yfw: **1 action item (test).**

This all-ones `token_mask` makes the pooled-stat assertions mask-insensitive: `count == 6 == numel` holds whether or not the accumulation respects the mask, so a regression from [`masked_select`](https://github.com/NVIDIA-NeMo/RL/blob/7e7d72130a112fe336978c03ae38e28773feb546/nemo_rl/algorithms/single_controller.py#L1850-L1859) to unmasked accumulation would go undetected. The source is currently correct — this is a regression guard, not a bug.

### AI-1
Use a partial mask and mask-sensitive expectations (verified passing end-to-end against the real `_advantage_stage`):
- `"token_mask": torch.tensor([[1.0, 1.0, 1.0], [1.0, 0.0, 0.0]])`
- `assert ctrl._opd_stat_count == 4`, `ctrl._opd_stat_sum == pytest.approx(1.0)`, `ctrl._opd_stat_sumsq == pytest.approx(0.25)`

The `torch.allclose(teacher - prev, full((2, 3), 0.25))` assertion and `has_valid_training_tokens` are unaffected.

yfw (commented):
Team review by 5 coordinated agents (rl-expert, bug-finder, test-agent, design-reviewer, devil's advocate) at head 7e7d72130.

**Overall: high-quality port — recommend addressing the two bugs before merge; nothing here is release-blocking.** All 173 affected unit tests pass locally on this head, lint is green (thanks for the quick import-sort fix), recipe/nightly registration follows the conventions, and docstrings are complete.

Worth calling out as good design, explicitly: the typed `PostWriteEnrichmentError` enrichment seam (the buffer stays OPD-agnostic, failures classify by exception type, and the rollout retry budget is correctly not spent on post-write failures); `TQTeacherLogprobCoordinator` being fully unit-testable without Ray/GPUs; reuse of the legacy teacher reserve/create path with the init-ordering discipline pinned by a test; and the fail-fast MOPD validation before any resource allocation.

Findings summary: 2 bugs on config paths adjacent to (not hit by) the shipped recipe — a teacher dynamic-batching crash (executed repro, verified fix attached) and a router-replay gate that breaks MoE-student MOPD — plus 9 low-severity suggestions (1 perf-evidence ask, 2 dead-code notes, 1 stall note, 1 error-type nit, 2 design/tracking notes, 2 verified test additions). Devil's-advocate pass: 10 findings confirmed, 1 dropped as unreachable, 1 folded.

_Generated by Claude Code_

tests/test_suites/llm/mopd-qwen3-1.7b-3n8g-megatron-pack-single-controller.sh
yfw: Ran the requested 5-step legacy vs. SingleController comparison. Both used the same base recipe, Qwen3-1.7B student/teacher, source dataset, seed 42, batch sizes, and checkpointing disabled.

| Metric (steps 1-5) | Legacy | SingleController |
|---|---|---|
| `train/loss` | `[0, 0, 0, -1.22e-14, 0]` | `[0, 0, 0, 0, 0]` |
| `train/token_mult_prob_error` | `[1.02048, 1.01890, 1.01935, 1.02117, 1.02045]` | `[1.02080, 1.01905, 1.01825, 1.02054, 1.02103]` |
| OPD `adv_mean` | zero except `4.51e-12` numerical noise at step 4 | all zero |
| OPD `adv_std` | zero except `9.74e-10` numerical noise at step 4 | all zero |
| OPD teacher/student gap mean | zero except `4.51e-12` numerical noise at step 4 | all zero |
| total step time (s) | `[11.52, 4.82, 4.44, 4.96, 5.95]` | `[23.60, 2.02, 5.83, 5.43, 5.44]` |
| teacher logprob time (s) | `[2.00, 1.57, 0.42, 0.11, 0.18]` | `[22.57, 0, 1.23, 0.91, 0.90]` |

The loss and OPD traces overlay at self-distillation tolerance. Median token probability error is `1.020450` legacy vs. `1.020543` SC (delta `9.4e-5`).

The SC teacher timing is attributed asynchronously: step 1 accounts for two 32-sample enrichments and step 2 accounts for none (`teacher_samples=[64,0,32,32,32]`, 160 total). The first `22.57s` bucket includes `16.44s` of same-teacher lock wait; steady-state nonzero enrichment is `0.90-1.23s`.

Both jobs completed successfully:
- Legacy: Slurm `16450510`, exit `0:0`, 10m18s
- SingleController: Slurm `16449390`, exit `0:0`, 9m34s

These were independent rollouts, so generated token counts are not bitwise identical despite the common source dataset/configuration and seed; this establishes behavioral parity rather than identical-example parity.

Environment caveat: the SC test used PR head plus the isolated upstream removal of the obsolete per-actor TQ `runtime_env` patch from #2935, because the nightly container already includes TQ. No MOPD implementation code differed; the legacy run used the same checkout.

nemo_rl/models/policy/teacher_worker_group.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

nemo_rl/data_plane/worker_mixin.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

nemo_rl/algorithms/single_controller_utils/setup.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

nemo_rl/algorithms/opd.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

nemo_rl/algorithms/opd.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

nemo_rl/algorithms/single_controller.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

nemo_rl/data_plane/schema.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

tests/unit/models/policy/test_teacher_worker_group.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

tests/unit/single_controller/test_single_controller.py
yfw: fixed in b0b91a30293cd8afced9ccc373ad5c4af54b0d98

tests/test_suites/llm/mopd-qwen3-1.7b-3n8g-megatron-pack-single-controller.sh
yfw: Thanks — this fully addresses the ask. The identically-zero SC gap/adv traces are exactly the alignment signal this comment was after (a misrouted or shifted teacher column would show up as a nonzero gap even in self-distillation), the token_mult_prob_error medians agree to 9.4e-5, and the steady-state 0.90–1.23s enrichment with the cold-start/lock-wait breakdown puts the serialized-teacher cost on record. Resolving.

tests/test_suites/llm/mopd-qwen3-1.7b-3n8g-megatron-pack-single-controller.sh
```diff
@@ -0,0 +1,60 @@
...
+        'abs(median(data["train/loss"])) < 0.05' \
+        'median(data["train/token_mult_prob_error"]) < 1.1' \
+        'max(data["train/on_policy_distillation/teacher_batches"]) > 0' \
+        'max(data["train/on_policy_distillation/teacher_samples"]) > 0' \
+        'max(data["train/on_policy_distillation/teacher_model_unique"]) == 1'
```
terrykong: **1 action item.**

**TL;DR — the gate proves the teacher stage *ran*, not that it wrote *aligned* numbers: a teacher column shifted by one token still has a near-zero mean, so every current assert passes.**

[`mopd-...-single-controller.sh:52-57`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/tests/test_suites/llm/mopd-qwen3-1.7b-3n8g-megatron-pack-single-controller.sh#L52-L57) — PR-introduced.

The three new asserts are counters. The only value check is `abs(median(train/loss)) < 0.05`, and the OPD advantage is `teacher - student`, whose *mean* is what the loss tracks ([`advantage_estimator.py:621-626`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/advantage_estimator.py#L621-L626)). Shift the teacher column by a token and the mean stays near zero while the spread blows up — so loss, `teacher_batches`, `teacher_samples` and `teacher_model_unique` all still pass.

`adv_std` is the one shipped metric that moves under that failure, and it is a direct product of this PR's new pooling ([`single_controller.py:104`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/single_controller.py#L104)), logged under the `train/` prefix ([`single_controller.py:1250`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/single_controller.py#L1250)).

The threshold is free on your own measurements: the 5-step comparison you posted has SC `adv_std` at 0 for all five steps (legacy peaked at 9.74e-10), so `< 0.01` leaves about seven orders of margin. Use `< 0.05` if you want more room — a token shift produces O(1) `adv_std` either way.

**Action:** add one assert to the gate.

```suggestion
    uv run tests/check_metrics.py $JSON_METRICS \
        'abs(median(data["train/loss"])) < 0.05' \
        'median(data["train/token_mult_prob_error"]) < 1.1' \
        'max(data["train/on_policy_distillation/teacher_batches"]) > 0' \
        'max(data["train/on_policy_distillation/teacher_samples"]) > 0' \
        'max(data["train/on_policy_distillation/teacher_model_unique"]) == 1' \
        'max(data["train/on_policy_distillation/adv_std"]) < 0.01'
```

Not run here — this review had no linux/GPU environment.

nemo_rl/algorithms/opd.py
```diff
@@ -200,6 +208,243 @@ def get_teacher_routing_metrics(
...
+            "on_policy_distillation/teacher_samples": float(self._teacher_samples),
+            "on_policy_distillation/teacher_logprob_time_s": self._teacher_logprob_time_s,
+            "on_policy_distillation/teacher_inference_time_s": self._teacher_inference_time_s,
+            "on_policy_distillation/teacher_lock_wait_time_s": self._teacher_lock_wait_time_s,
+            "on_policy_distillation/teacher_alias_unique": float(alias_unique),
+            "on_policy_distillation/teacher_model_unique": float(model_unique),
+            "on_policy_distillation/teacher_alias_to_model_compression": float(
+                model_unique / max(alias_unique, 1)
+            ),
+        }
```
terrykong: **1 action item (low — metrics hygiene, not correctness).**

**TL;DR — `drain_metrics` keeps a second copy of the three routing-metric formulas that `get_teacher_routing_metrics` already owns, and the copy production actually runs is the untested one.**

[`opd.py:422-445`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/opd.py#L422-L445) — PR-introduced. Two problems, one edit.

**1. Duplicate formula.** [`get_teacher_routing_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/opd.py#L186-L208) emits the same three keys and the same `model_unique / max(alias_unique, 1)` ratio, 230 lines up. It has **no production caller** — its only reference in the tree is [`test_opd.py:876`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/tests/unit/algorithms/test_opd.py#L876). So the tested copy is dead and the live copy is untested: redefine "compression" in one and CI stays green while the dashboard drifts.

**2. Idle steps log zeros.** `_train_pump` drains every step ([`single_controller.py:1124`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/single_controller.py#L1124)), but enrichment is attributed asynchronously, so a step can legitimately see zero enrichments — your own posted run shows `teacher_samples=[64,0,32,32,32]`. On step 2 this logs `teacher_alias_unique=0`, `teacher_model_unique=0`, `compression=0.0`, which reads as "the teachers vanished" rather than "nothing ran this step". A ratio of 0 is also not a ratio. The sibling in this same feature already does the right thing: `_pooled_opd_metrics` returns `{}` when there is nothing to report ([`single_controller.py:98-99`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/single_controller.py#L98-L99)).

The counters and timings should keep reporting 0 — only the three cardinality/ratio keys should drop out.

This does not affect the nightly gate: `max(...teacher_model_unique) == 1` reduces over the steps present, and `check_metrics.max` folds `value.values()`, so omitting step 2 leaves `{1:1, 3:1, 4:1, 5:1}` -> `1`.

**Action:** delegate to the existing helper and gate the cardinality keys.

```suggestion
        metrics = {
            "on_policy_distillation/teacher_batches": float(self._teacher_batches),
            "on_policy_distillation/teacher_samples": float(self._teacher_samples),
            "on_policy_distillation/teacher_logprob_time_s": self._teacher_logprob_time_s,
            "on_policy_distillation/teacher_inference_time_s": self._teacher_inference_time_s,
            "on_policy_distillation/teacher_lock_wait_time_s": self._teacher_lock_wait_time_s,
        }
        if self._teacher_batches:
            # Cardinality describes what actually ran. On an idle step, 0 reads
            # as "zero teacher models" rather than "no activity" -- omit it,
            # the way _pooled_opd_metrics does.
            metrics.update(
                get_teacher_routing_metrics(
                    sorted(self._aliases_seen),
                    self._opd_cfg["teacher_model_by_agent_name"],
                )
            )
```

Follow-through, not in the block above because it spans three other lines: `_models_seen` becomes dead once the helper derives models from aliases. Drop its init ([`opd.py:253`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/opd.py#L253)), its `.add` ([`opd.py:413`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/opd.py#L413)), and its `.clear()` ([`opd.py:444`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/opd.py#L444)).

Not run here — no linux/GPU environment.

tests/unit/algorithms/test_opd.py
```diff
@@ -319,6 +321,420 @@ def test_compute_teacher_logprobs_default_alias_fallback_routes():
...
+
+    async def run_both():
+        await asyncio.gather(
+            coordinator.enrich(_teacher_meta("math", 1, 4), _teacher_record("math")),
+            coordinator.enrich(_teacher_meta("code", 1, 4), _teacher_record("code")),
+        )
+
+    asyncio.run(run_both())
+
+    assert max_active == 1
```
terrykong: **1 action item (test).**

**TL;DR — nothing proves distinct teachers run *concurrently*, so collapsing the per-teacher locks into one shared lock would pass every test in this PR while silently serializing independent teachers.**

[`test_opd.py:675-735`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/tests/unit/algorithms/test_opd.py#L675-L735) — PR-introduced.

This test puts two aliases behind **one** physical group and asserts `max_active == 1`, which pins the serialization half well. The other half — the PR description's "while allowing distinct teachers to run concurrently" — has no test at all. `_teacher_locks` is deliberately keyed on the physical group ([`opd.py:243-246`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/opd.py#L243-L246)); a regression to a single shared `asyncio.Lock` is invisible to the whole suite.

**Action:** add the complementary test. Use a `threading.Barrier`, not `sleep` — asserting a *positive* overlap needs both threads to genuinely coexist, and a sleep-based version can flake on a loaded CI box. With a barrier, a shared-lock regression fails crisply with `BrokenBarrierError` instead of intermittently.

`dp_size=1` with `batch_size=1` gives `remainder == 0`, so `_enrich_sync` skips the padding block entirely and never touches `dp_client` — the same reason the neighbouring test gets away with `object()`.

```suggestion
    assert max_active == 1


def test_tq_teacher_enrichment_runs_distinct_teachers_concurrently():
    """Distinct physical teachers hold distinct locks, so they overlap."""
    import asyncio
    import threading

    from nemo_rl.algorithms import opd

    barrier = threading.Barrier(2)

    class BarrierTeacher(_MockTeacherWorkerGroup):
        def get_logprobs_from_meta(self, meta):
            del meta
            # Blocks until BOTH teachers are inside inference at once. One
            # shared lock would serialize them and trip BrokenBarrierError.
            barrier.wait(timeout=5)

    coordinator = opd.TQTeacherLogprobCoordinator(
        dp_client=object(),
        teacher_worker_groups={
            "primary": BarrierTeacher(dp_size=1),
            "secondary": BarrierTeacher(dp_size=1),
        },
        alias_to_group_alias={"math": "primary", "code": "secondary"},
        on_policy_distillation_cfg={
            "teacher_model_by_agent_name": {
                "math": "/ckpt/math",
                "code": "/ckpt/code",
            }
        },
    )

    async def run_both():
        await asyncio.gather(
            coordinator.enrich(_teacher_meta("math", 1, 4), _teacher_record("math")),
            coordinator.enrich(_teacher_meta("code", 1, 4), _teacher_record("code")),
        )

    asyncio.run(run_both())

    metrics = coordinator.drain_metrics()
    assert metrics["on_policy_distillation/teacher_model_unique"] == 2.0
```

Not run here — no linux/GPU environment, so please confirm it passes before merging.

tests/unit/models/policy/test_teacher_worker_group.py
```diff
@@ -77,3 +91,324 @@ def test_create_teacher_configs_deduplicates():
...
+    assert captured["dynamic_batching_args"]["max_tokens_per_microbatch"] == 256
+    assert captured["meta"].extra_info[GLOBAL_FORWARD_PAD_SEQLEN] == 128
+    shards = worker_group.run_all_workers_sharded_data.call_args.kwargs["meta"]
+    assert sorted(sample_id for shard in shards for sample_id in shard.sample_ids) == [
+        "a",
+        "b",
+    ]
+    assert all(MICRO_BATCH_INDICES in shard.extra_info for shard in shards)
+    assert all(MICRO_BATCH_LENGTHS in shard.extra_info for shard in shards)
+    assert len({len(shard.extra_info[MICRO_BATCH_LENGTHS]) for shard in shards}) == 1
```
terrykong: **1 action item (test).**

**TL;DR — this assert can never fail: `len(MICRO_BATCH_LENGTHS)` is the *chunk* count, which is always 1 on the logprob path, so it pins nothing.**

[`test_teacher_worker_group.py:255`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/tests/unit/models/policy/test_teacher_worker_group.py#L255) — PR-introduced (added in b0b91a302).

`MICRO_BATCH_LENGTHS` is a list **of chunks**, each chunk itself a list of bin lengths: [`batched_data_dict.py:694`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/distributed/batched_data_dict.py#L694) appends `chunk_sharded_micro_lengths[shard_idx]` once per chunk. And `shard_meta_for_dp` passes `batch_size=None`, so [`batched_data_dict.py:534`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/distributed/batched_data_dict.py#L534) sets `batch_size = total_batch_size` and `num_chunks == 1`.

So the outer length is `1` for every shard, the set is always `{1}`, and the assert holds no matter what the packer did.

The property worth pinning is the **inner** length — that both DP ranks got the same number of microbatches. That is what `bin_count_multiple=shards` ([`batched_data_dict.py:602-603`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/distributed/batched_data_dict.py#L602-L603)) plus the `[shard_idx::shards]` round-robin ([`batched_data_dict.py:655-659`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/distributed/batched_data_dict.py#L655-L659)) exist to guarantee, and unequal bin counts across DP ranks desync Megatron's per-microbatch collectives.

**Action:** assert the inner length.

```suggestion
    assert (
        len(
            {
                len(microbatch_lengths)
                for shard in shards
                for microbatch_lengths in shard.extra_info[MICRO_BATCH_LENGTHS]
            }
        )
        == 1
    )
```

Worth adding the same corrected form to the sequence-packing test, which has no cross-DP assert at all ([`test_teacher_worker_group.py:326`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/tests/unit/models/policy/test_teacher_worker_group.py#L326)) — and packing is the branch the shipped `-pack` nightly actually runs.

Not run here — no linux/GPU environment.

nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -827,6 +841,14 @@ async def commit(
                 tags=[dict(t) for t in tags],
             )
 
+            if self._post_write_enricher is not None:
+                try:
+                    meta = await self._post_write_enricher(meta, record)
+                except Exception as error:
+                    raise PostWriteEnrichmentError(
+                        f"post-write enrichment failed for group_id={group_id!r}"
+                    ) from error
```
terrykong, line 887: **No action needed in this PR — tracked in #3815 for follow-up.**

Noting it here because this is where the seam is. `set_post_write_enricher` and `PostWriteEnrichmentError` are new in this PR — nothing on `main` had them — so `TQReplayBuffer`, which every SC algorithm uses, gained an extension point with one implementation, and `commit()` went from a bounded write to awaiting a GPU forward pass on another node.

The same placement also sets the batch size. A commit only sees its own group, so the teacher is called with 4 rows against a data-parallel width of 4 — one sequence per rank — and all 8 groups in a step queue on one teacher lock. That is where the 16.44 s of lock wait in the 5-step comparison comes from.

#3815 carries a full design sketch rather than just a complaint: `commit(defer_ready=True)` plus `mark_ready()` as the only buffer surface, teacher scoring as a third task beside the rollout and train pumps, batched by teacher. It also lists the decisions that would need making (deadlock bound on batch size, failure blast radius across a batch, drain at shutdown, checkpoint interaction), and asks for one measurement first — a 32-row call vs eight 4-row calls — because the batching win only exists if that forward is latency-bound.

Nothing to change here; the hook is ~25 lines and easy to move later.

tests/unit/algorithms/test_opd.py
terrykong, line 735: **1 action item (test).**

**TL;DR — nothing proves distinct teachers run *concurrently*, so collapsing the per-teacher locks into one shared lock would pass every test in this PR while silently serializing independent teachers.**

[`test_opd.py:675-735`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/tests/unit/algorithms/test_opd.py#L675-L735) — PR-introduced.

This test puts two aliases behind **one** physical group and asserts `max_active == 1`, which pins the serialization half well. The other half — the PR description's "while allowing distinct teachers to run concurrently" — has no test at all. `_teacher_locks` is deliberately keyed on the physical group ([`opd.py:243-246`](https://github.com/NVIDIA-NeMo/RL/blob/359c895809a32982e700a4a6f80877c36b5cecec/nemo_rl/algorithms/opd.py#L243-L246)); a regression to a single shared `asyncio.Lock` is invisible to the whole suite.

**Action:** add the complementary test. Use a `threading.Barrier`, not `sleep` — asserting a *positive* overlap needs both threads to genuinely coexist, and a sleep-based version can flake on a loaded CI box. With a barrier, a shared-lock regression fails crisply with `BrokenBarrierError` instead of intermittently.

`dp_size=1` with `batch_size=1` gives `remainder == 0`, so `_enrich_sync` skips the padding block entirely and never touches `dp_client` — the same reason the neighbouring test gets away with `object()`.

```suggestion
    assert max_active == 1


def test_tq_teacher_enrichment_runs_distinct_teachers_concurrently():
    """Distinct physical teachers hold distinct locks, so they overlap."""
    import asyncio
    import threading

    from nemo_rl.algorithms import opd

    barrier = threading.Barrier(2)

    class BarrierTeacher(_MockTeacherWorkerGroup):
        def get_logprobs_from_meta(self, meta):
            del meta
            # Blocks until BOTH teachers are inside inference at once. One
            # shared lock would serialize them and trip BrokenBarrierError.
            barrier.wait(timeout=5)

    coordinator = opd.TQTeacherLogprobCoordinator(
        dp_client=object(),
        teacher_worker_groups={
            "primary": BarrierTeacher(dp_size=1),
            "secondary": BarrierTeacher(dp_size=1),
        },
        alias_to_group_alias={"math": "primary", "code": "secondary"},
        on_policy_distillation_cfg={
            "teacher_model_by_agent_name": {
                "math": "/ckpt/math",
                "code": "/ckpt/code",
            }
        },
    )

    async def run_both():
        await asyncio.gather(
            coordinator.enrich(_teacher_meta("math", 1, 4), _teacher_record("math")),
            coordinator.enrich(_teacher_meta("code", 1, 4), _teacher_record("code")),
        )

    asyncio.run(run_both())

    metrics = coordinator.drain_metrics()
    assert metrics["on_policy_distillation/teacher_model_unique"] == 2.0
```

Not run here — no linux/GPU environment, so please confirm it passes before merging.

terrykong (commented):
> [!IMPORTANT]
> **This PR now has merge conflicts with `main` and needs a rebase.** It was `MERGEABLE` when this review started and went `CONFLICTING` while the review was being written — four commits landed on `main` in between, including #3771 (`fix(sc): skip reference logprobs when KL is disabled`), which is the most likely overlap. None of them touched `opd.py` or `tests/unit/algorithms/test_opd.py`, so the findings below still apply as written, but the line numbers will shift after you rebase.

Re-review at `359c8958` by a team of 6 coordinated agents (rl-expert, bug-finder, test-agent, design-reviewer, an architecture tracer, and a devil's advocate).

**All 9 findings from the previous review are verified genuinely fixed in b0b91a302** — three lanes checked each fix independently, including sibling branches, and found no partial fixes. Nothing below re-raises them. The 5-step legacy-vs-SC convergence comparison is accepted; **nothing below re-requests performance or convergence evidence.**

Overall this is a well-shaped port, and the hard parts are handled carefully: post-write TQ enrichment ordering, per-physical-teacher serialization with cross-teacher concurrency, DP padding with rollback under cancellation, driver-side global packing plans that keep DP collectives in sync, and exact pooled metrics across unequal streaming chunks. 4 low-severity items follow.

**Design choices we examined and think are right — please don't churn these:**

- **`set_post_write_enricher` is the right seam and the right size.** A single optional callback invoked inside `commit()`'s existing `try`, so the enricher inherits the `put_samples` rollback path for free. We looked for a second post-write consumer and could not name one (multimodal MOPD is this same enricher widened, not a second one; reference logprobs are a train-pump stage). Turning it into a stage list would force ordering and partial-failure semantics that nothing today can pin down.
- **The metric accumulators should stay inline.** They score 2 of 3 on extraction — closed state and low reach-back, but no independent lifecycle. A `TeacherMetrics` class here would be an anemic helper. The coordinator *itself* earns being a class: it owns the per-teacher locks and a real stateful protocol.
- **The setup ordering is pinned executably**, not just in comments: `assert events == ["reserve_teacher", "build_generation", "create_teacher"]` turns "claim teacher GPUs before NeMo-Gym can take them, but load checkpoints after the student to avoid the shared HF-to-Megatron cache race" into a regression guard.

Two findings were raised and then **killed** by the adversarial pass, noted so they don't resurface: a proposed consistency guard in `SingleControllerActor.__init__` (unreachable — validation enforces OPD-implies-teachers three times over) and a concern about the `prev_logprobs`/`logprobs_policy` kwarg branch (safe — validation is bidirectional and no estimator consumes both).

**Caveat on verification:** this review ran on macOS with no GPU, and the repo's `uv.lock` is linux-only, so **no test and no linter was executed**. Every suggestion below is marked unverified and should be run before merging.

_Generated by Claude Code_

tests/test_suites/llm/mopd-qwen3-1.7b-3n8g-megatron-pack-single-controller.sh
yfw: Addressed in 112afd1e55e0cadaa25aa8f185b09398aef7eed6

nemo_rl/algorithms/opd.py
yfw: Addressed in 112afd1e55e0cadaa25aa8f185b09398aef7eed6

tests/unit/models/policy/test_teacher_worker_group.py
yfw: Addressed in 112afd1e55e0cadaa25aa8f185b09398aef7eed6

tests/unit/algorithms/test_opd.py
yfw: Addressed in 112afd1e55e0cadaa25aa8f185b09398aef7eed6

nemo_rl/algorithms/async_utils/replay_buffer.py
yfw: ack, agree about the design proposed in #3815

tests/test_suites/llm/mopd-qwen3-1.7b-3n8g-megatron-pack-single-controller.sh
terrykong: fixed in 112afd1e55e0cadaa25aa8f185b09398aef7eed6

nemo_rl/algorithms/opd.py
terrykong: fixed in 112afd1e55e0cadaa25aa8f185b09398aef7eed6

tests/unit/models/policy/test_teacher_worker_group.py
terrykong: fixed in 112afd1e55e0cadaa25aa8f185b09398aef7eed6

tests/unit/algorithms/test_opd.py
terrykong: fixed in 112afd1e55e0cadaa25aa8f185b09398aef7eed6

terrykong (approved):

PR #2770, opened 2026-06-10
feat: add effort levels to nemo gym

ashors1:
# What does this PR do ?
Adds a cleaned-up version of https://github.com/NVIDIA-NeMo/RL/commit/1693cc1baf84d8eb1d8ee91a83e461fed4031e54. A few notes:
- Effort levels are configured via `nemo_gym.effort_levels`, whereas in the original commit, each effort level key was simply present in the top-level `master_config`
- The original commit had a typo (`low_penlty` rather than `low_penalty`). This PR has fixed that typo

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #3770, opened 2026-08-23
feat(gym): apply effort-level reward shaping in the SingleController

RayenTian:
PR #2770 only wired length-based effort reward shaping into the run_grpo.py (v1) synchronous paths. The SingleController (v2) path runs its own NeMo-Gym rollout via RolloutManager/AsyncNemoGymRolloutImpl, which never applied it.

- Thread EffortLevelsConfig through RolloutManager into AsyncNemoGymRolloutImpl and call the shared _apply_effort_shaping before completions are built, so the shaped reward flows into Completion.reward.
- Fold the effort length/reward metrics into _compute_rollout_metrics.
- setup.py resolves the config via the shared _get_effort_config helper.

No behaviour change when env.nemo_gym.effort_levels is unset (_get_effort_config returns None -> shaping is a no-op).

# result
image
image

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/single_controller_utils/setup.py
+++ b/nemo_rl/algorithms/single_controller_utils/setup.py
@@ -38,6 +38,7 @@ from nemo_rl.algorithms.async_utils.replay_buffer import TQReplayBuffer
 from nemo_rl.algorithms.grpo import (
     GRPOSaveState,
     _create_advantage_estimator,
+    _get_effort_config,
     _get_grpo_save_state,
 )
 from nemo_rl.algorithms.grpo import MasterConfig as GrpoMasterConfig
@@ -810,6 +811,7 @@ def setup_single_controller(
             env_s=master_config.async_rl.rollout_failure.native.env_timeout_s,
         ),
         retry_policy=_build_retry_policy(master_config),
+        effort_config=_get_effort_config(cast(GrpoMasterConfig, master_config)),
     )
 
     # Print setup timing metrics
--- a/nemo_rl/experience/rollout_manager.py
+++ b/nemo_rl/experience/rollout_manager.py
@@ -16,6 +16,7 @@ import asyncio
 import copy
 import enum
 import json
+import statistics
 from dataclasses import dataclass, field
 from typing import Any, Optional
 
@@ -41,8 +42,11 @@ from nemo_rl.experience.failures import (
 from nemo_rl.experience.interfaces import Completion, PromptGroupRecord
 from nemo_rl.experience.metric_utils import calculate_single_metric, pct
 from nemo_rl.experience.rollouts import (
+    EffortLevelsConfig,
+    _apply_effort_shaping,
     _attach_routed_experts_to_message_log_prefix,
     _dummy_routed_experts_for_tokens,
+    _EffortShapingMetrics,
     _find_routed_experts_template,
     _tensorize_by_key,
     attach_static_multimodal_payload,
@@ -749,6 +753,8 @@ class AsyncNemoGymRolloutImpl:
         # Shared with the owning RolloutManager so row-level re-dispatches are visible
         # in the same counters as everything else. None when constructed directly.
         stats: Optional[RolloutStats] = None,
+        # Length-based reward shaping for low-effort prompts; None disables it.
+        effort_config: Optional[EffortLevelsConfig] = None,
         **kwargs: Any,
     ) -> None:
         self._tokenizer = tokenizer
@@ -765,6 +771,7 @@ class AsyncNemoGymRolloutImpl:
             else RolloutRetryPolicy.single_attempt()
         ).max_gym_row_attempts
         self._stats = stats
+        self._effort_config = effort_config
 
         self._validate_init_params()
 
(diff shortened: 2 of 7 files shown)


RayenTian:
/ok to test [316182c](https://github.com/NVIDIA-NeMo/RL/pull/3770/commits/316182c4051a7883783c6ebc780398a1bcb5d642)

nemo_rl/experience/rollout_manager.py
```diff
@@ -998,6 +1008,8 @@ async def _run_rollouts(
             rollout_metrics = self._compute_rollout_metrics(
                 completions, inputs[0]["agent_ref"]["name"]
             )
+            # Same helper the batched path uses, so the two cannot drift apart.
+            rollout_metrics.update(_effort_shaping_metrics(shaping))
```
yuki-97, line 1012: The shaped reward reaches training, but these six keys never reach a logger on the SC path — `run_rollout` puts them on [`PromptGroupRecord.rollout_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/316182c4051a7883783c6ebc780398a1bcb5d642/nemo_rl/experience/rollout_manager.py#L808), and the record's only consumer, [`record_to_train_batch`](https://github.com/NVIDIA-NeMo/RL/blob/316182c4051a7883783c6ebc780398a1bcb5d642/nemo_rl/experience/payload.py#L55), reads `completions` and `prompt` only; [the TODO](https://github.com/NVIDIA-NeMo/RL/blob/316182c4051a7883783c6ebc780398a1bcb5d642/nemo_rl/algorithms/single_controller.py#L1198) still lists `rollout_metrics` as unwired.

It predates this PR (the whole `_compute_rollout_metrics` dict is already discarded in SC), so there's no need to remove the `rollout_metrics.update(...)` call — a short comment here plus a caveat in the PR description would do, since the new tests assert on the returned dict and pass either way.

nemo_rl/experience/rollouts.py
```diff
@@ -466,6 +466,38 @@ def _apply_effort_shaping(
...
+    Shared by the batched v1 path and the SingleController rollout manager so the
+    two cannot drift apart.
+
+    Args:
+        shaping: Per-sample tracking lists returned by ``_apply_effort_shaping``.
+
+    Returns:
+        Metric name to value. Empty when shaping was disabled or when no prompt in
+        the group matched ``low_string`` -- callers ``update`` an existing dict, so
+        an absent key leaves the metric unreported rather than reporting a zero.
```
yuki-97: The second empty-case is wrong: with shaping enabled and no prompt matching `low_string`, every sample lands in [`high_lengths`](https://github.com/NVIDIA-NeMo/RL/blob/316182c4051a7883783c6ebc780398a1bcb5d642/nemo_rl/experience/rollouts.py#L461), so the helper returns `mean_length_high` + `median_length_high` rather than `{}` — [this PR's own test](https://github.com/NVIDIA-NeMo/RL/blob/316182c4051a7883783c6ebc780398a1bcb5d642/tests/unit/test_effort_shaping.py#L367) asserts that. Only a disabled config yields an empty dict.

```suggestion
        Metric name to value. Empty only when shaping was disabled; callers
        ``update`` an existing dict, so an absent key leaves the metric unreported
        rather than reporting a zero.
```

nemo_rl/experience/rollout_manager.py
yuki-97: pre-exist issue, feels ok to do in a subsequent PR.

RayenTian: Yes, logging for rollout metrics has never been supported in V2. We could add support for it, but we need to design how to define a "rollout step," since in a streaming scenario, there isn't really a step in the strict sense.

RayenTian:
/ok to test [bbc8191](https://github.com/NVIDIA-NeMo/RL/pull/3770/commits/bbc819125ff985f5c0306fca28fb6896e2df6c16)

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -810,6 +811,7 @@ def _build_generation_then_trainer(
             env_s=master_config.async_rl.rollout_failure.native.env_timeout_s,
         ),
         retry_policy=_build_retry_policy(master_config),
+        effort_config=_get_effort_config(cast(GrpoMasterConfig, master_config)),
```
yuki-97, line 814: This wires the SC path, but `grpo.async_grpo` still doesn't get it — [`trajectory_collector.py:924`](https://github.com/NVIDIA-NeMo/RL/blob/bbc819125ff985f5c0306fca28fb6896e2df6c16/nemo_rl/algorithms/async_utils/trajectory_collector.py#L924) calls `run_async_nemo_gym_rollout` without `effort_config=`, though the parameter exists at [`rollouts.py:2304`](https://github.com/NVIDIA-NeMo/RL/blob/bbc819125ff985f5c0306fca28fb6896e2df6c16/nemo_rl/experience/rollouts.py#L2304).

Nine of the ten in-tree configs that set `env.nemo_gym.effort_levels` also set `async_grpo.enabled: true`, so they run on the one gym path that drops shaping. And since [`validate()`](https://github.com/NVIDIA-NeMo/RL/blob/bbc819125ff985f5c0306fca28fb6896e2df6c16/nemo_rl/algorithms/grpo.py#L3883) does pass it, those runs shape validation rewards but not training rewards — no guard or warning fires.

yuki-97: not a blocker for this PR since this PR support it for SC, but worth to notice that legacy async path silently skip it.

yuki-97 (approved):

PR #2794, opened 2026-06-12
perf: Fix DSV3 perf regression & boost weight loading time (dsv3 MTP)

youngeunkwon0405:
# What does this PR do ?

1. Fix the performance regression caused by #2672
2. Enhance the weight loading time when the MTP spec dec is enabled. Mostly targeting for dsv3 recipe update.(#2779)

Without this fix, MTP-enabled dsv3 takes 30-40mins for the model loading to load the entire ~1300 GB models from the storage.  This fix makes the code to follow the `load_format == "dummy"` trick that relies on the refit to fill the model weights and perform the load from storage only for the MTP layers.

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #2672, opened 2026-06-02
fix(grpo): improve non-colocated refit handling

macandro96:
# What does this PR do ?

Offload policy optimizer state before NCCL refit and restore training state after weight sync. Also skip reference policy logprobs when KL is disabled.

# Issues
Closes https://github.com/NVIDIA-NeMo/RL/issues/1978

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #3771, opened 2026-08-23
fix(sc): skip reference logprobs when KL is disabled

RayenTian:
# What does this PR do?

Ports the still-applicable reference-policy logprob behavior from #2672 to the SingleController GRPO path (`examples/run_grpo_single_controller.py`).

When `loss_fn.reference_policy_kl_penalty == 0`, SingleController now skips reference-policy logprob computation even when `grpo.skip_reference_policy_logprobs_calculation` is left at its default value. This aligns the runtime decision with trainer setup, which does not initialize a reference model when the KL penalty is disabled.

## Scope note

This PR intentionally does **not** port the non-colocated refit changes from #2672 (`offload_before_refit()` / `prepare_for_training()`). Those changes were [reverted](https://github.com/NVIDIA-NeMo/RL/pull/2794/changes#r3465670441) by #2794 because they caused a performance regression, so only the reference-logprob portion is migrated to SingleController.

# Testing

- Added coverage for KL disabled with explicit skip both false and true, and KL enabled with skip false.
- Targeted test: 3 passed.
- Ruff checks passed for the changed files.
- Commit-time pre-commit hooks passed, including Ruff, Ruff format, and Pyrefly.

# Related PRs

- #2672
- #2794

### DIFF
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -131,8 +131,9 @@ class SingleControllerActor:
             master_config.loss_fn.force_on_policy_ratio
             and master_config.grpo.seq_logprob_error_threshold is None
         )
-        self._reference_logprobs_required = not bool(
-            master_config.grpo.skip_reference_policy_logprobs_calculation
+        self._reference_logprobs_required = bool(
+            master_config.loss_fn.reference_policy_kl_penalty > 0
+            and not master_config.grpo.skip_reference_policy_logprobs_calculation
         )
         self._dp_client = actor_args.dp_client
         self._gen: Generation = actor_args.gen_handle
--- a/nemo_rl/algorithms/single_controller_utils/config.py
+++ b/nemo_rl/algorithms/single_controller_utils/config.py
@@ -758,6 +758,16 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
             "loss_fn.reference_policy_kl_penalty=0."
         )
 
+    if (
+        reference_policy_kl_penalty <= 0
+        and not master_config.grpo.skip_reference_policy_logprobs_calculation
+    ):
+        print(
+            "Reference policy logprob calculation will be skipped since "
+            "`loss_fn.reference_policy_kl_penalty` is 0, so no reference "
+            "model was initialized."
+        )
+
     _validate_failure_settings(async_config, num_prompts_per_step)
 
     # Nesting says which knob applies to which path, but nothing stops an operator
--- a/nemo_rl/algorithms/single_controller_utils/config.py
+++ b/nemo_rl/algorithms/single_controller_utils/config.py
@@ -745,6 +745,14 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
     reference_policy_kl_penalty = getattr(
         master_config.loss_fn, "reference_policy_kl_penalty", 0
     )
+
+    if reference_policy_kl_penalty < 0:
+        raise ValueError(
+            "loss_fn.reference_policy_kl_penalty="
+            f"{reference_policy_kl_penalty} must not be negative; "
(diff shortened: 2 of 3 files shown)


RayenTian:
/ok to test [77ad70a](https://github.com/NVIDIA-NeMo/RL/pull/3771/commits/77ad70a027e143e1dd3bd03eec8947eca786a1ea)

RayenTian:
/ok to test [68fffae](https://github.com/NVIDIA-NeMo/RL/pull/3771/commits/68fffae7277f434eb360040aeef78b9565488362)

nemo_rl/algorithms/single_controller.py
```diff
@@ -131,8 +131,9 @@ def __init__(
             master_config.loss_fn.force_on_policy_ratio
             and master_config.grpo.seq_logprob_error_threshold is None
         )
-        self._reference_logprobs_required = not bool(
-            master_config.grpo.skip_reference_policy_logprobs_calculation
+        self._reference_logprobs_required = bool(
```
yuki-97, line 134: don't we need to align here? if no need to calculate, then there's no need to init reference model?
https://github.com/NVIDIA-NeMo/RL/blob/ffbf33f3099847b6685072d4eb251bae605ff8de/nemo_rl/algorithms/single_controller_utils/setup.py#L299

nemo_rl/algorithms/single_controller.py
```diff
@@ -131,8 +131,9 @@ def __init__(
...
+            master_config.loss_fn.reference_policy_kl_penalty > 0
```
yuki-97, line 135: not blocker for this PR, but @yfw just curious is there some case that `kl_penalty == 0` but don't `skip_reference_policy_logprobs_calculation`?
if not, why don't we drop `skip_reference_policy_logprobs_calculation` and only use `kl_penalty == 0` to control whether to skip?

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -758,6 +758,16 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
             "loss_fn.reference_policy_kl_penalty=0."
         )
 
+    if (
+        reference_policy_kl_penalty <= 0
```
yuki-97: nit: `=0` to align with the message
```suggestion
        reference_policy_kl_penalty == 0
```

nemo_rl/algorithms/single_controller.py
RayenTian: No need change. `skip = true`, but the case where `skip_reference_policy_logprobs_calculation > 0` has already been caught [here](https://github.com/NVIDIA-NeMo/RL/pull/3771/changes#diff-8105859df792b2077a44f1ce124e9f4846a879de918c6057a5a9b698ade6fd28L749).

nemo_rl/algorithms/single_controller_utils/config.py
RayenTian: Done in [e09f8db](https://github.com/NVIDIA-NeMo/RL/pull/3771/commits/e09f8db504bf0898b7ffe670b8ce0728169f890a)

RayenTian:
/ok to test [e09f8db](https://github.com/NVIDIA-NeMo/RL/pull/3771/commits/e09f8db504bf0898b7ffe670b8ce0728169f890a)

nemo_rl/algorithms/single_controller.py
yfw: It's a good point, I don't think there's a case where kl_penalty is 0, but we should still compute the reference logprobs. I don't remember the history of that flag, but I imagine it may be so that we could opt-in to skipping it when it was first introduced. I agree we can remove this in a follow-up PR.

yuki-97 (approved):

PR #3774, opened 2026-08-23
fix(vllm): patch flashinfer_trtllm moe backend to support refit

guyueh1:
Summary:
- Patch vLLM flashinfer_trtllm MoE weight processing to preserve refit buffers.
- Add coverage for the monkeypatch path.

Verification:
- git diff --check HEAD~1..HEAD
- python3 -m py_compile nemo_rl/models/generation/vllm/patches.py nemo_rl/models/generation/vllm/vllm_worker.py tests/unit/models/generation/test_vllm_patches.py

### DIFF
--- a/nemo_rl/models/generation/vllm/patches.py
+++ b/nemo_rl/models/generation/vllm/patches.py
@@ -14,7 +14,29 @@
 
 import os
 from contextlib import contextmanager
+from functools import wraps
 from importlib.util import find_spec
+from typing import Any
+
+G_FLASHINFER_TRTLLM_W13_KERNEL_ATTR = "_nrl_flashinfer_trtllm_w13_weight"
+G_FLASHINFER_TRTLLM_W2_KERNEL_ATTR = "_nrl_flashinfer_trtllm_w2_weight"
+G_FLASHINFER_TRTLLM_PATCH_ATTR = "_nrl_flashinfer_trtllm_refit_patch_applied"
+G_FLASHINFER_TRTLLM_INTERMEDIATE_ALIGNMENT = 128
+G_FLASHINFER_TRTLLM_ENGINE_CORE_PATCH_ATTR = (
+    "_nrl_flashinfer_trtllm_refit_engine_core_patch_applied"
+)
+G_FLASHINFER_TRTLLM_ENGINE_CORE_ORIGINAL_ATTR = "_nrl_original_run_engine_core"
+G_FLASHINFER_TRTLLM_RAY_WORKER_PATCH_ATTR = (
+    "_nrl_flashinfer_trtllm_refit_ray_worker_patch_applied"
+)
+G_FLASHINFER_TRTLLM_RAY_WORKER_ORIGINAL_ATTR = "_nrl_original_initialize_worker"
+G_FLASHINFER_TRTLLM_RAY_EXECUTOR_PATCH_ATTR = (
+    "_nrl_flashinfer_trtllm_refit_ray_executor_patch_applied"
+)
+G_FLASHINFER_TRTLLM_RAY_EXECUTOR_ORIGINAL_ATTR = "_nrl_original_collective_rpc"
+G_FLASHINFER_TRTLLM_RAY_EXECUTOR_WORKERS_PATCHED_ATTR = (
+    "_nrl_flashinfer_trtllm_refit_workers_patched"
+)
 
 
 def _get_vllm_file(relative_path: str) -> str:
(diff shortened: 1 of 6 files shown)


guyueh1:
Validation on Nemotron-3-Super 120B A12B + vLLM `flashinfer_trtllm` MoE backend:

**Without the patch**, the super-model run reaches the refit path and fails while reloading MoE weights into vLLM:

- Job: `6432787`
- Base code: `8863ebeff172fc3e1d14ab6cdbd0af626f477c01`
- Config: `examples/configs/recipes/llm/performance/grpo-nemotron3-super-120BA12B-32n4g-async-1off.yaml`
- Override: `++policy.generation.vllm_kwargs.moe_backend=flashinfer_trtllm`

Relevant log:

```text
Error in VllmInternalWorkerExtension.update_weights_from_collective:
ValueError: shard_dim=0 is not a valid data dimension for a 3D tensor (expected 1 or 2)

... _load_hf_weights(...)
... vllm/model_executor/models/nemotron_h.py, line 748, in load_weights
... vllm/model_executor/layers/fused_moe/routed_experts.py, line 490, in _load_w13
    hidden_dim = self._get_hidden_dim(shard_dim, expert_data.ndim)
... routed_experts.py, line 409, in _get_hidden_dim
    raise ValueError(...)
```

This matches the suspected failure mode: vLLM mutates the MoE parameter tensors into the FlashInfer TRTLLM kernel layout after initial load, so the next NeMo-RL refit/load path no longer sees the canonical loader layout.

**With this patch**, the super-model repro completes 5 GRPO steps cleanly.

- Job: `6471701`
- Code: `f0557321ead671f57995a32ebf16705b1611d1fe`
- Config: same super async recipe
- Overrides: `moe_backend=flashinfer_trtllm`, `expert_parallel_size=4`, `max_num_steps=5`, checkpointing disabled
- W&B: https://wandb.ai/nvidia/nemo-rl/runs/4cek0501

W&B summary at step 5:

```text
_step: 5
train/token_mult_prob_error: 1.0278989381185966
train/gen_kl_error: 0.0028161558283912314
train/reward: 0.5234375
train/loss: -0.008716936157657895
performance/tokens_per_sec: 13616.51111280378
performance/tokens_per_sec_per_gpu: 106.37899306877954
efficiency/efficiency_pct: 89.44657080519458
```

Run log also ends cleanly:

```text
Max number of steps has been reached, stopping training early
Async GRPO training complete!
wandb: View run grpo-nemotron3-super-120BA12B-32n4g-async-1off-trtllm-ep4-main-nightly-wandb at:
https://wandb.ai/nvidia/nemo-rl/runs/4cek0501
```

kajalj22 (dismissed):

nemo_rl/models/generation/vllm/patches.py
```diff
@@ -566,6 +588,469 @@ def _patch_vllm_radio_layerscale_loader(logger) -> None:
...
+        UnquantizedMoeBackend,
+        convert_to_unquantized_kernel_format,
+        make_unquantized_moe_kernel,
+    )
+    from vllm.model_executor.layers.fused_moe.unquantized_fused_moe_method import (
+        UnquantizedFusedMoEMethod,
+    )
+
+    if logger is None:
+        logger = init_logger("vllm_patch")
```
yfw, line 618: **1 action item.**

**TL;DR — the unconditional `from vllm.logger import init_logger` at L606 makes 4 of the 5 new tests fail on a clean checkout (and be order-dependent where vLLM is installed). No production effect; moving it into the `if logger is None:` branch fixes it.**

[`patches.py:606`](https://github.com/NVIDIA-NeMo/RL/blob/58c328993c1c76faaaf78290a09039422e87d88c/nemo_rl/models/generation/vllm/patches.py#L606) — PR-introduced.

`from vllm.logger import init_logger` runs unconditionally, ten lines above the [`if logger is None:`](https://github.com/NVIDIA-NeMo/RL/blob/58c328993c1c76faaaf78290a09039422e87d88c/nemo_rl/models/generation/vllm/patches.py#L616) that is its only consumer — and the only in-tree caller, [`_patch_vllm_flashinfer_trtllm_refit_buffers`](https://github.com/NVIDIA-NeMo/RL/blob/58c328993c1c76faaaf78290a09039422e87d88c/nemo_rl/models/generation/vllm/patches.py#L1048), always passes a logger. So there's no production effect, but it fails 4 of the 5 new tests: `fake_vllm_unquantized_moe_modules` shadows `sys.modules["vllm"]` with a stub whose `__path__ = []` and installs fakes for the two MoE modules, not `vllm.logger`.

```
$ uv run --group test pytest tests/unit/models/generation/test_vllm_patches.py --maxfail=0 -q
nemo_rl/models/generation/vllm/patches.py:606: ModuleNotFoundError: No module named 'vllm.logger'
4 failed, 8 passed
```

(`--maxfail=0` is needed — `addopts` carries `-x`, so a plain run only reports the first failure.)

It's order-dependent rather than reliably red, which is the more awkward part: if anything earlier in the same process already imported `vllm.logger`, Python serves it from `sys.modules` without ever consulting the stub's empty `__path__`, and these pass. So in a container where vLLM is installed, the result depends on shard ordering.

**Action:** defer the import into the `if logger is None:` branch — the same shape [`_patch_vllm_flashinfer_trtllm_refit_from_collective_rpc`](https://github.com/NVIDIA-NeMo/RL/blob/58c328993c1c76faaaf78290a09039422e87d88c/nemo_rl/models/generation/vllm/patches.py#L957) already uses at L957-L960. Verified locally: 12 passed.

```suggestion
    # These imports require vLLM, so keep them out of module import time.
    import torch
    from vllm.model_executor.layers.fused_moe.oracle.unquantized import (
        UnquantizedMoeBackend,
        convert_to_unquantized_kernel_format,
        make_unquantized_moe_kernel,
    )
    from vllm.model_executor.layers.fused_moe.unquantized_fused_moe_method import (
        UnquantizedFusedMoEMethod,
    )

    if logger is None:
        from vllm.logger import init_logger

        logger = init_logger("vllm_patch")
```

yfw (approved):

PR #3778, opened 2026-08-23
fix(loss): mask filtered tokens in force-on-policy reductions

tianyi-zhang-02:
# What does this PR do?

This follows up on #3551. That PR now owns the filtered actor mask and the actor-side retained-token aggregation; this PR extends the same protocol to the `force_on_policy_ratio` path, where `prev_logprobs` aliases `curr_logprobs.detach()`.

When a current logprob is filtered, its finite substitute must not leak into quantities that read `prev_logprobs`. This change:

- narrows the four train/generation mismatch diagnostics to the retained tokens;
- narrows token- and sequence-level importance-sampling inputs, including TIS/ICEPOP and `seq-mask-tis`;
- emits raw numerator fragments with their matching retained-token counts, then divides once after microbatch and data-parallel aggregation;
- leaves the normal, separately computed `prev_logprobs` path unchanged.

The actor loss still uses the original global-token denominator, so filtering does not increase the effective step size. Reference-policy KL also keeps the full token mask.

| Quantity | Mask / denominator |
|---|---|
| actor diagnostics | retained actor tokens / retained actor-token count (#3551) |
| prev-derived diagnostics with `force_on_policy_ratio` | retained prev tokens / retained prev-token count |
| actor gradient | retained actor tokens / original global-token count |
| reference-policy KL | full token mask / original global-token count |

# Stack

- `main`: `e831622435180c88d6adcc9fce46bfd88a85f3e5`
- #3551 head: `cbfd57d454fdb8e6f963b4d241bece9c2b6f2df7`
- this PR head: `69b9b424876a53900054d8e9ce8c2e616074339b`

The Files changed view includes #3551 until that PR lands. This branch merges both the current #3551 head and the current `main` and is conflict-free at the SHA above.

# Validation

Final-head CPU validation on macOS arm64, Python 3.12.2, PyTorch 2.8.0, Ray 2.51.1:

| Suite | Result |
|---|---:|
| `test_loss_functions.py`, `test_utils.py`, `test_sequence_packing_fusion.py`, `single_controller/test_utils.py` | 117 passed, 45 CUDA-only skipped |
| `test_grpo.py`, `test_ppo.py` | 259 passed, 7 CUDA-only skipped |
| Ruff check + format check on all 11 touched Python files | passed |

The tests cover uneven microbatches/ranks, zero retained tokens, all affected diagnostics, token- and sequence-level IS, TIS/ICEPOP normalization, and classic/async/sync/Single Controller caller aggregation.

The final head was not rerun on CUDA because the private 4090 host became unavailable. No functional training run is claimed here.

- [x] Added the necessary unit and caller-level tests
- [x] Aligned the branch with current `main`
- [x] Ran lint and formatting checks
- [ ] Final-head CUDA / functional run


tianyi-zhang-02:
Closing to avoid duplicating the active top-k/top-p masking work in #4154 and the masked-value safety fix in #4120. This branch also still needs a denominator redesign, so keeping a parallel implementation open would make the intended loss semantics less clear.

PR #3787, opened 2026-08-24
fix(sc): apply reward scaling before advantage estimation

tianyi-zhang-02:
## What does this PR do?

Makes Single Controller apply the existing `grpo.reward_scaling` setting before advantage estimation, using the shared `scale_rewards` helper.

This PR is intentionally limited to reward scaling. Current main already provides advantage clipping, overlong filtering, and environment sample masking. Reward shaping and message-level penalties remain out of scope because Single Controller does not yet carry the response-length/message fields they need.

The shipped default remains a no-op, and algorithms without a reward-scaling block remain unchanged.

## Validation

Current head `90d294e9df9376dec52d71d94488d814a8d20b72`, rebased onto upstream `main` at `90a2a212d503455d8590be5c8de3cb989d3425b0`.

- full `tests/unit/single_controller/test_single_controller_actor.py`: **61 passed**
- caller-level coverage verifies scaling happens before advantage estimation and that logged rewards use the scaled values

An earlier pre-refresh head completed a real 2×H100 SC run with Qwen3-0.6B, Megatron policy/reference, vLLM, and TQ. With scaling enabled, rewards were mapped from `0..1` to `-1..1`; training completed with loss `0.1040` and grad norm `2.6518`. The current head was shrunk against main and rerun through the full SC actor suite; the 2-GPU functional run was not repeated after that refresh.

### DIFF
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -87,6 +87,7 @@ from nemo_rl.algorithms.grpo import (
     _write_latest_checkpoint_status,
     aggregate_rollout_metrics,
     compute_and_apply_seq_logprob_error_masking,
+    scale_rewards,
 )
 from nemo_rl.algorithms.metric_utils import SetupTimingMetrics
 from nemo_rl.algorithms.ppo import _compute_critic_metrics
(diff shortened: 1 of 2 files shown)


tianyi-zhang-02:

```python
    #
    unsupported = (
        []
        else [
            name
            for name, enabled in (
            )
            if enabled
        ]
    )
```

PR #3788, opened 2026-08-24
fix(dtensor): count TP-replicated gradients once in get_grad_norm

khazic:
## Problem

`get_grad_norm` accumulates each rank's local `sum(|g|^p)` and then reduces it with `SUM` twice, first over `dp_cp_group` and then over `tp_group`:

```python
for grad in grads_for_norm:
    grad_norm = torch.linalg.vector_norm(grad, ord=norm_type, dtype=dtype)
    total_norm_cuda += torch.pow(grad_norm, norm_type)

torch.distributed.all_reduce(total_norm_cuda, op=SUM, group=dp_cp_group)
torch.distributed.all_reduce(total_norm_cuda, op=SUM, group=tp_group)
```

The second reduction is only valid for gradients that are genuinely sharded across the TP group. Tensor parallelism shards only the parameters named in the parallel plan (attention and MLP projections, embeddings, `lm_head`). Everything else stays replicated: RMSNorm/LayerNorm weights, biases, router weights, and any module the plan does not cover. Every TP rank computes an identical gradient for those parameters, so summing their squared norms over `tp_group` counts each of them `tp_size` times.

Two consequences:

1. The logged `grad_norm` is systematically over-reported, and the error grows with `tp_size`.
2. `clip_grad_by_total_norm_` is driven by that same value, so gradient clipping fires more aggressively than `policy.max_grad_norm` asks for. The effective clip threshold shrinks by the same factor.

Megatron's `clip_grads.py`, which this function is adapted from, excludes duplicates before the reduction via `param_is_not_tensor_parallel_duplicate`. That filter was not carried over when the function was ported.

## Fix

Split the gradients by whether they are sharded on the mesh dimension backed by `tp_group`:

- both halves are reduced over `dp_cp_group`, since FSDP shards every parameter across DP and each DP rank therefore owns a distinct piece of both;
- only the sharded half is reduced over `tp_group`; the replicated half is added once.

The TP mesh dimension is located by comparing process-group ranks rather than object identity, so a mesh slice such as `mesh["tp"]` resolves correctly. A plain `torch.Tensor` gradient, or a `DTensor` that does not live on the TP mesh, is treated as replicated.

The infinity-norm path is left as it was, because a `MAX` reduction is idempotent and duplicates are harmless there. The new test covers that path too so the behaviour stays locked in.

## Evidence

`tests/unit/models/dtensor/test_grad_norm_tp.py` builds one TP-replicated gradient (4096 elements, identical on every rank) and one TP-sharded gradient (8192 elements) on a 2-rank TP mesh, then compares `get_grad_norm` against the exact norm of the concatenated full gradients.

On the current `main` the test fails:

```
tests/unit/models/dtensor/test_grad_norm_tp.py::test_get_grad_norm_counts_tp_replicated_grads_once[2]

>           assert result["actual"] == pytest.approx(result["expected"], rel=1e-5)
E           assert 130.01957484836657 == 112.023248812...9 +- 0.00112023
E
E             comparison failed
E             Obtained: 130.01957484836657
E             Expected: 112.02324881294129 +- 0.00112023

1 failed
```

That is a 16.1% over-report at `tp_size=2`, consistent with `sqrt((2*||g_repl||^2 + ||g_shard||^2) / (||g_repl||^2 + ||g_shard||^2))` for this split.

With the fix applied the same test passes:

```
tests/unit/models/dtensor/test_grad_norm_tp.py::test_get_grad_norm_counts_tp_replicated_grads_once[2] PASSED

1 passed
```

A reduced form of the same check, run directly against `get_grad_norm` on a 2-rank TP mesh with a single parameter, isolates the factor exactly:

```
before:
  true ||g||                       = 31.032419
  get_grad_norm, SHARDED param     = 31.032418  ratio=1.0000
  get_grad_norm, REPLICATED param  = 43.886468  ratio=1.4142   <- sqrt(2)

after:
  true ||g||                       = 31.032419
  get_grad_norm, SHARDED param     = 31.032418  ratio=1.0000
  get_grad_norm, REPLICATED param  = 31.032419  ratio=1.0000
```

The sharded parameter was already correct in both cases, which confirms the regression is confined to replicated parameters.

## Note on the extra file

`tests/unit/models/dtensor/__init__.py` is added because the test spawns a Ray actor defined in the test module and Ray resolves the actor by its fully qualified module name. Without it the module is collected under a bare top-level name that the worker process cannot import. Every sibling test directory already ships one.

## Placement handling

Only `Replicate` counts as a duplicate. A `Partial` gradient on the TP mesh dimension is neither sharded nor duplicated, since each rank holds a different addend of the true gradient, so bucketing it either way would be wrong and would leave TP ranks with different clip coefficients. `_is_tp_duplicate` raises on it instead. It is not reachable on the current path, because FSDP redistributes non-DP mesh dimensions back to the parameter placements before exposing `.grad`, but the invariant is now enforced rather than assumed.

## Behaviour change worth calling out

When `tp_group` is `None`, every gradient now lands in the replicated bucket and skips the second all-reduce. Previously `all_reduce(..., group=None)` fell through to the WORLD group. Nothing in the training path passes `None` (`dtensor_policy_worker` always passes a real group), and the one caller that does is `test_get_grad_norm_precision`, which monkeypatches `all_reduce` to a no-op.

## Test run

The new test and the pre-existing `test_get_grad_norm_precision` parametrizations pass together, so the change does not regress the single-process path:

```
tests/unit/models/dtensor/test_grad_norm_tp.py::test_get_grad_norm_counts_tp_replicated_grads_once[2] PASSED
tests/unit/models/dtensor/test_parallelize.py::test_get_grad_norm_precision[grad_dtype0-norm_dtype0-1] PASSED
tests/unit/models/dtensor/test_parallelize.py::test_get_grad_norm_precision[grad_dtype1-norm_dtype1-2] PASSED
tests/unit/models/dtensor/test_parallelize.py::test_get_grad_norm_precision[grad_dtype2-norm_dtype2-inf] PASSED
tests/unit/models/dtensor/test_parallelize.py::test_get_grad_norm_precision[grad_dtype3-norm_dtype3-1] PASSED
tests/unit/models/dtensor/test_parallelize.py::test_get_grad_norm_precision[grad_dtype4-norm_dtype4-2] PASSED
tests/unit/models/dtensor/test_parallelize.py::test_get_grad_norm_precision[grad_dtype5-norm_dtype5-inf] PASSED

7 passed
```

## Scope

This PR only fixes the TP axis, which is the axis that demonstrably double-counts today. The `dp_cp_group` reduction keeps its existing assumption that FSDP shards every parameter across DP.

### DIFF
--- a/nemo_rl/models/dtensor/parallelize.py
+++ b/nemo_rl/models/dtensor/parallelize.py
@@ -813,6 +813,46 @@ def clip_grad_by_total_norm_(
             g.mul_(clip_coeff)
 
 
+def _is_tp_duplicate(
+    grad: Union[torch.Tensor, DTensor],
+    tp_group: torch.distributed.ProcessGroup,
+) -> bool:
+    """Whether every rank of ``tp_group`` holds an identical copy of ``grad``.
+
+    Tensor parallelism only shards the parameters named in the parallel plan
+    (attention/MLP projections, embeddings). Everything else stays replicated:
+    norm weights, biases, router weights, and any module the plan does not
+    cover. Each TP rank therefore computes the very same gradient for those
+    parameters, and that gradient must contribute to the global norm exactly
+    once.
+
+    A plain ``torch.Tensor`` is replicated by definition. A ``DTensor`` is
+    replicated when its placement on the mesh dimension backed by ``tp_group``
+    is not a ``Shard``. The mesh dimension is located by comparing group ranks
+    rather than object identity so that mesh slices (``mesh["tp"]``) resolve
+    correctly.
+
+    Args:
+        grad: The gradient to classify.
+        tp_group: Tensor-parallel process group, or ``None`` when TP is unused.
+
+    Returns:
+        bool: True when the gradient is duplicated across ``tp_group``.
+    """
+    if tp_group is None or not isinstance(grad, DTensor):
+        return True
+
+    mesh = grad.device_mesh
+    tp_ranks = torch.distributed.get_process_group_ranks(tp_group)
+    for dim in range(mesh.ndim):
+        if torch.distributed.get_process_group_ranks(mesh.get_group(dim)) == tp_ranks:
+            return not isinstance(grad.placements[dim], Shard)
+
+    # The gradient does not live on the TP mesh at all, so it is replicated
+    # across the group.
+    return True
+
+
 def get_grad_norm(
     parameters: Union[list[Union[torch.Tensor, DTensor]], Union[torch.Tensor, DTensor]],
     dp_cp_group: torch.distributed.ProcessGroup,
@@ -840,10 +880,17 @@ def get_grad_norm(
     if isinstance(parameters, (torch.Tensor, DTensor)):
         parameters = [parameters]
 
-    # Grads.
-    grads_for_norm = [
-        to_local_if_dtensor(p.grad.detach()) for p in parameters if p.grad is not None
-    ]
+    # Grads, split by whether the TP group holds duplicate copies of them.
+    grads_for_norm = []
+    tp_duplicate_grads = []
+    for p in parameters:
+        if p.grad is None:
+            continue
+        grad = p.grad.detach()
+        if _is_tp_duplicate(grad, tp_group):
+            tp_duplicate_grads.append(to_local_if_dtensor(grad))
+        else:
+            grads_for_norm.append(to_local_if_dtensor(grad))
 
     # Norm parameters.
     norm_type = float(norm_type)
@@ -851,6 +898,8 @@ def get_grad_norm(
 
     # Calculate norm.
     if norm_type == torch.inf:
+        # A max-reduction is idempotent, so duplicates need no special handling.
+        grads_for_norm = grads_for_norm + tp_duplicate_grads
         total_norm = max(grad.abs().max().item() for grad in grads_for_norm)
         total_norm_cuda = torch.tensor([float(total_norm)], dtype=dtype, device="cuda")
         # Take max across all data-parallel GPUs if using FSDP and then all model-parallel GPUs.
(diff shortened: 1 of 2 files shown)


khazic:
I ran this end to end to quantify what the fix changes in a real training loop, and to check the mechanism rather than a single data point. Posting the full result, including the part that argues for a narrower claim than my original description made.

## What the upstream TP plan already says

The strongest argument for this change is not the magnitude, it is the semantics. `transformers` declares the parallel style of these parameters explicitly. From `Qwen3MoeConfig.base_model_tp_plan`, and identically in the dense `Qwen3Config`:

```python
"layers.*.self_attn.q_norm": "replicated_with_grad_allreduce",
"layers.*.self_attn.k_norm": "replicated_with_grad_allreduce",
```

A dedicated parallel style exists for exactly this class of parameter because a replicated parameter cannot be reduced the same way a sharded one is. `get_grad_norm` currently applies one `SUM` over `tp_group` to every parameter regardless of its placement, so parameters the plan marks as replicated are counted `tp_size` times. This PR makes the function honour a distinction the plan already draws.

## Measurement

SFT, Qwen3-0.6B, `dtensor_cfg._v2: false`, `max_grad_norm: 1.0`, fixed seed, 2 steps. Step 1 is the only controlled comparison, since both sides start from identical parameters there; `train/loss` agrees to all reported digits on every pair, confirming the runs are aligned. The patch is applied and reverted around each pair, so within a pair the only difference is this change.

| `tensor_parallel_size` | main | with this PR | ratio |
|---|---|---|---|
| 2 | 407.295990 | 406.975983 | 1.000786 |
| 4 | 416.472839 | 415.752716 | 1.001732 |
| 8 | 410.532776 | 409.184296 | 1.003296 |

The over-report grows with `tensor_parallel_size`, as expected. The magnitude is small: between 0.08% and 0.33%.

## Direct measurement of the replicated share

The ratio alone only supports an inference, so I measured the quantity directly. The patched function already separates the two contributions, which makes this a two line diagnostic on top of the patch:

```
r = duplicate_sq / (sharded_sq_after_tp_reduce + duplicate_sq)
```

| `tensor_parallel_size` | `sharded_sq` | `duplicate_sq` | r measured | r inferred from ratio |
|---|---|---|---|---|
| 2 | 1.6538e+05 | 246.59 | 0.1489% | 0.1573% |
| 4 | 1.7266e+05 | 184.97 | 0.1070% | 0.1156% |
| 8 | 1.6709e+05 | 136.66 | 0.0817% | 0.0943% |

The two agree to within 6% to 15%, which confirms the mechanism: the over-report follows `sqrt(1 + (tp_size - 1) * r)`.

It also corrects an assumption I had made. `r` is not a fixed property of the model. `duplicate_sq` falls by 45% from `tensor_parallel_size` 2 to 8 while `sharded_sq` moves by only 4%, so the set of parameters that end up replicated itself depends on the parallel configuration. The size of the error therefore depends on how much gradient mass a given configuration leaves replicated, not on `tensor_parallel_size` alone.

## Scope, stated plainly

Two limits are worth being explicit about, since they bound how much this matters in practice.

**Only the v1 DTensor worker is affected.** `get_grad_norm` is called from `dtensor_policy_worker.py` and nowhere else; `dtensor_policy_worker_v2.py` does not call it. Most shipped configs set `_v2: true`.

**The measured effect is a fraction of a percent** on a dense model, because the only replicated parameters there are norm scales. I also tried to measure a case with more replicated mass. Qwen3-MoE does not load on this path at all (`ValueError: Unknown parallel style: replicated_with_grad_allreduce`, raised in `dtensor_policy_worker.py`), and its plan shards the experts (`packed_colwise`, `rowwise`, `moe_tp_experts`) anyway, so MoE would not enlarge the effect even if it ran.

The clipping consequence is real at this magnitude but correspondingly small. With `max_grad_norm` below the observed norm every step clips, so the error in the clip coefficient does reach the parameter trajectory: the two runs separate from step 2 onward and stay close through the end of the run.

I would rather present this accurately than oversell it. The case for the change is that the reduction does not respect a placement distinction the TP plan already declares, and that `clip_grad_by_total_norm_` consumes the resulting value as if it were the true norm. The measured magnitude on a dense model is small, and I have adjusted the description to say so.

<details>
<summary>Per-tp step 1 values, both sides</summary>

```
tp   main grad_norm   patched grad_norm       ratio   implied r   loss identical
 2      407.295990         406.975983    1.000786     0.1573%             yes
 4      416.472839         415.752716    1.001732     0.1156%             yes
 8      410.532776         409.184296    1.003296     0.0943%             yes

direct measurement, step 1, patched build:
tp   sharded_sq        duplicate_sq      r
 2   1.653770625e+05   2.465894775e+02   0.001488854
 4   1.726637500e+05   1.849747009e+02   0.001070154
 8   1.670887344e+05   1.366627197e+02   0.000817237
```

</details>

khazic:
Closing: #4198 removed the DTensor v1 worker and this `get_grad_norm` with it. The v2 path clips through Automodel's `_clip_grad_norm_impl`, which skips the all-reduce on `Replicate` mesh dims, so it does not double count TP-replicated gradients.

PR #3798, opened 2026-08-24
feat: Pass tokenizer_kwargs to AutoTokenizer.from_pretrained

pstjohn:
Adds `tokenizer_kwargs` to `TokenizerConfig`

`get_tokenizer()` currently constructs tokenizers with only the model name and `trust_remote_code=True`:

```python
tokenizer = AutoTokenizer.from_pretrained(
    tokenizer_config["name"], trust_remote_code=True
)
```

There is no config route for passing additional `from_pretrained` kwargs — for example `fix_mistral_regex=False` for tokenizer artifacts that already carry a corrected Mistral regex pretokenizer. A caller that needs this must monkeypatch `AutoTokenizer.from_pretrained` in every process, including inside isolated Ray workers where tokenizers are constructed after the actor has started.

This PR adds `tokenizer_kwargs` to `TokenizerConfig` and threads it through both `get_tokenizer()` implementations. When `tokenizer_kwargs` is absent or `None`, behaviour is unchanged: `.get("tokenizer_kwargs") or {}` produces an empty unpacking.

### DIFF
--- a/nemo_rl/algorithms/utils.py
+++ b/nemo_rl/algorithms/utils.py
@@ -300,6 +300,10 @@ def get_tokenizer(
                     - "default": Uses the tokenizer's default template
                     - A custom jinja2 template string
                     If not specified, the tokenizer's default template will be used.
+                - tokenizer_kwargs: Extra keyword arguments forwarded to tokenizer
+                  loading, e.g. {"fix_mistral_regex": False}. When
+                  get_processor=True, these are passed through
+                  AutoProcessor.from_pretrained().
         get_processor: Whether to return a processor (via AutoProcessor) instead of a tokenizer.
 
     Returns:
@@ -359,17 +363,21 @@ def get_tokenizer(
     maybe_patch_fastokens(bool(tokenizer_config.get("use_fastokens")))
 
     processor = None
+    tokenizer_kwargs = dict(tokenizer_config.get("tokenizer_kwargs") or {})
 
     if get_processor:
         processor = AutoProcessor.from_pretrained(
-            tokenizer_config["name"], trust_remote_code=True, use_fast=True
+            tokenizer_config["name"],
+            trust_remote_code=True,
+            use_fast=tokenizer_kwargs.pop("use_fast", True),
+            **tokenizer_kwargs,
         )
         tokenizer = processor.tokenizer
     else:
         tokenizer = AutoTokenizer.from_pretrained(
             tokenizer_config["name"],
             trust_remote_code=True,
-            **(tokenizer_config.get("tokenizer_kwargs") or {}),
+            **tokenizer_kwargs,
         )
 
     if tokenizer.pad_token is None:
--- a/nemo_rl/algorithms/utils.py
+++ b/nemo_rl/algorithms/utils.py
@@ -367,7 +367,9 @@ def get_tokenizer(
         tokenizer = processor.tokenizer
     else:
         tokenizer = AutoTokenizer.from_pretrained(
-            tokenizer_config["name"], trust_remote_code=True
+            tokenizer_config["name"],
+            trust_remote_code=True,
+            **(tokenizer_config.get("tokenizer_kwargs") or {}),
         )
 
     if tokenizer.pad_token is None:
--- a/nemo_rl/models/automodel/setup.py
+++ b/nemo_rl/models/automodel/setup.py
@@ -190,7 +190,9 @@ def get_tokenizer(
         tokenizer = processor.tokenizer
     else:
         tokenizer = NeMoAutoTokenizer.from_pretrained(
-            tokenizer_config["name"], trust_remote_code=True
+            tokenizer_config["name"],
+            trust_remote_code=True,
+            **(tokenizer_config.get("tokenizer_kwargs") or {}),
         )
 
     if tokenizer.pad_token is None:
--- a/nemo_rl/models/automodel/setup.py
+++ b/nemo_rl/models/automodel/setup.py
@@ -176,23 +176,31 @@ def get_tokenizer(
                     - A custom jinja2 template string
                     If not specified, the tokenizer's default template will be used.
                 - chat_template_kwargs: Arguments passed to tokenizer.apply_chat_template()
+                - tokenizer_kwargs: Extra keyword arguments forwarded to
+                  NeMoAutoTokenizer.from_pretrained(), e.g.
+                  {"fix_mistral_regex": False}. When get_processor=True, these
+                  are passed through AutoProcessor.from_pretrained().
         get_processor: Whether to return a processor (via AutoProcessor) instead of a tokenizer.
 
     Returns:
         The configured tokenizer or processor instance.
     """
     processor = None
+    tokenizer_kwargs = dict(tokenizer_config.get("tokenizer_kwargs") or {})
 
     if get_processor:
         processor = AutoProcessor.from_pretrained(
-            tokenizer_config["name"], trust_remote_code=True, use_fast=True
+            tokenizer_config["name"],
+            trust_remote_code=True,
+            use_fast=tokenizer_kwargs.pop("use_fast", True),
+            **tokenizer_kwargs,
         )
         tokenizer = processor.tokenizer
     else:
         tokenizer = NeMoAutoTokenizer.from_pretrained(
             tokenizer_config["name"],
             trust_remote_code=True,
-            **(tokenizer_config.get("tokenizer_kwargs") or {}),
+            **tokenizer_kwargs,
         )
 
     if tokenizer.pad_token is None:
--- a/nemo_rl/models/policy/__init__.py
+++ b/nemo_rl/models/policy/__init__.py
@@ -509,6 +509,8 @@ class TokenizerConfig(TypedDict):
     chat_template: NotRequired[str | None]
     # Arguments to pass to tokenizer.apply_chat_template(...). This can be used to pass kwargs like enable_thinking=true
     chat_template_kwargs: NotRequired[dict[str, Any] | None]
+    # Arguments forwarded to AutoTokenizer.from_pretrained(...) via get_tokenizer.
+    tokenizer_kwargs: NotRequired[dict[str, Any] | None]
     # Multimodal configs
     audio: NotRequired[dict[str, Any]]
     video: NotRequired[dict[str, Any]]
--- a/nemo_rl/models/policy/__init__.py
+++ b/nemo_rl/models/policy/__init__.py
@@ -509,7 +509,7 @@ class TokenizerConfig(TypedDict):
     chat_template: NotRequired[str | None]
     # Arguments to pass to tokenizer.apply_chat_template(...). This can be used to pass kwargs like enable_thinking=true
     chat_template_kwargs: NotRequired[dict[str, Any] | None]
-    # Arguments forwarded to AutoTokenizer.from_pretrained(...) via get_tokenizer.
+    # Arguments forwarded to tokenizer loading via get_tokenizer.
     tokenizer_kwargs: NotRequired[dict[str, Any] | None]
     # Multimodal configs
     audio: NotRequired[dict[str, Any]]
--- a/tests/unit/algorithms/test_utils.py
+++ b/tests/unit/algorithms/test_utils.py
@@ -14,6 +14,7 @@
 
 import math
 from datetime import datetime
+from unittest.mock import patch
 
 import pytest
 import torch
@@ -150,12 +151,37 @@ def test_get_tokenizer_forwards_tokenizer_kwargs():
     """Test get_tokenizer unpacks tokenizer_kwargs into from_pretrained."""
     config = {
         "name": "meta-llama/Llama-3.2-1B-Instruct",
-        "tokenizer_kwargs": {"use_fast": False},
+        "tokenizer_kwargs": {"model_max_length": 123},
     }
-    tokenizer = get_tokenizer(config)
-    # use_fast=False produces the slow Python tokenizer; is_fast is True
-    # only for the HuggingFace fast (Rust-backed) variant.
-    assert not tokenizer.is_fast
+    with patch("nemo_rl.algorithms.utils.AutoTokenizer") as mock_auto_tokenizer:
+        get_tokenizer(config)
+
+    mock_auto_tokenizer.from_pretrained.assert_called_once_with(
+        "meta-llama/Llama-3.2-1B-Instruct",
+        trust_remote_code=True,
+        model_max_length=123,
+    )
+
+
+def test_get_processor_forwards_tokenizer_kwargs():
+    """Test get_tokenizer forwards tokenizer_kwargs through AutoProcessor."""
+    config = {
+        "name": "Qwen/Qwen2.5-VL-3B-Instruct",
+        "tokenizer_kwargs": {"model_max_length": 123, "use_fast": False},
+    }
+    with patch("nemo_rl.algorithms.utils.AutoProcessor") as mock_auto_processor:
+        get_tokenizer(config, get_processor=True)
+
+    mock_auto_processor.from_pretrained.assert_called_once_with(
+        "Qwen/Qwen2.5-VL-3B-Instruct",
+        trust_remote_code=True,
+        use_fast=False,
+        model_max_length=123,
+    )
+    assert config["tokenizer_kwargs"] == {
+        "model_max_length": 123,
+        "use_fast": False,
+    }
 
 
 def test_maybe_pad_last_batch():
--- a/tests/unit/algorithms/test_utils.py
+++ b/tests/unit/algorithms/test_utils.py
@@ -146,6 +146,18 @@ def test_get_tokenizer_custom_jinja_template(conversation_messages):
     assert formatted == expected
 
 
+def test_get_tokenizer_forwards_tokenizer_kwargs():
+    """Test get_tokenizer unpacks tokenizer_kwargs into from_pretrained."""
+    config = {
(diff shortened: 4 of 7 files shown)


rohitrango (dismissed):
LGTM

terrykong (dismissed):

tests/unit/algorithms/test_utils.py
```diff
@@ -146,6 +146,18 @@ def test_get_tokenizer_custom_jinja_template(conversation_messages):
...
+def test_get_tokenizer_forwards_tokenizer_kwargs():
+    """Test get_tokenizer unpacks tokenizer_kwargs into from_pretrained."""
+    config = {
+        "name": "meta-llama/Llama-3.2-1B-Instruct",
+        "tokenizer_kwargs": {"use_fast": False},
+    }
+    tokenizer = get_tokenizer(config)
+    # use_fast=False produces the slow Python tokenizer; is_fast is True
+    # only for the HuggingFace fast (Rust-backed) variant.
+    assert not tokenizer.is_fast
```
terrykong: [`tests/unit/algorithms/test_utils.py:149`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/tests/unit/algorithms/test_utils.py#L149)

**1 action item.**

**TL;DR — this test can never pass: transformers v5 throws `use_fast` away before it reaches any tokenizer, so `is_fast` is always `True`.**

PR-introduced. No unit-test lane has run on this PR yet (only DCO / copyright / secrets / submodule checks), so CI has not flagged it.

`AutoTokenizer.from_pretrained` discards `use_fast` at the top of the function:

```python
# V5: Always use fast tokenizers, ignore use_fast parameter
_ = kwargs.pop("use_fast", None)
```

[`tokenization_auto.py:663-664`](https://github.com/huggingface/transformers/blob/v5.8.1/src/transformers/models/auto/tokenization_auto.py#L663-L664) — identical at [5.5.0:656-657](https://github.com/huggingface/transformers/blob/v5.5.0/src/transformers/models/auto/tokenization_auto.py#L656-L657). This repo pins [`transformers>=5.5.0,<5.9.0`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/pyproject.toml#L57), and every version it resolves to (5.5.0 / 5.5.4 / 5.6.0 / 5.8.1) behaves this way. The slow backend is unreachable too: [`tokenization_auto.py:774-776`](https://github.com/huggingface/transformers/blob/v5.8.1/src/transformers/models/auto/tokenization_auto.py#L774-L776) reroutes `PythonBackend` to `TokenizersBackend`, whose [`is_fast` returns `True`](https://github.com/huggingface/transformers/blob/v5.8.1/src/transformers/tokenization_utils_tokenizers.py#L492-L493).

So `assert not tokenizer.is_fast` is `assert not True`, and it fails on every run.

There is a second problem underneath: `use_fast` is the one kwarg that cannot show this feature works. Deleting the new `**kwargs` unpack from `get_tokenizer` would give an identical result, so the test cannot tell "forwarded" from "dropped".

**Action:** assert the forwarding directly with a mock. That also removes the gated-model download, so the test then needs neither `@pytest.mark.hf_gated` (its [four siblings](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/tests/unit/algorithms/test_utils.py#L94) have it, this one does not) nor an entry in [`excluded_unit_tests.sh:31-34`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/tests/unit/excluded_unit_tests.sh#L31-L34) — and it starts running in FAST mode.

```suggestion
def test_get_tokenizer_forwards_tokenizer_kwargs():
    """Test get_tokenizer unpacks tokenizer_kwargs into from_pretrained."""
    from unittest.mock import patch

    config = {
        "name": "meta-llama/Llama-3.2-1B-Instruct",
        "tokenizer_kwargs": {"model_max_length": 123},
    }
    with patch("nemo_rl.algorithms.utils.AutoTokenizer") as mock_auto:
        get_tokenizer(config)
    mock_auto.from_pretrained.assert_called_once_with(
        "meta-llama/Llama-3.2-1B-Instruct",
        trust_remote_code=True,
        model_max_length=123,
    )
```

The import sits inside the function because `unittest.mock` is not imported at module scope in this file.

If you would rather keep a real download, `model_max_length: 123` plus `assert tokenizer.model_max_length == 123` also works ([it survives into the instance](https://github.com/huggingface/transformers/blob/v5.8.1/src/transformers/tokenization_utils_base.py#L1058-L1059)) — but then please also add `@pytest.mark.hf_gated` and the `excluded_unit_tests.sh` entry.

nemo_rl/algorithms/utils.py
```diff
@@ -367,7 +367,9 @@ def get_tokenizer(
         tokenizer = processor.tokenizer
     else:
         tokenizer = AutoTokenizer.from_pretrained(
-            tokenizer_config["name"], trust_remote_code=True
+            tokenizer_config["name"],
+            trust_remote_code=True,
+            **(tokenizer_config.get("tokenizer_kwargs") or {}),
```
terrykong: [`nemo_rl/algorithms/utils.py:372`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/algorithms/utils.py#L372)

**1 action item.**

**TL;DR — `tokenizer_kwargs` is silently ignored whenever `get_tokenizer` builds a processor, so every VLM / multimodal run drops it, including half of the eval flow.**

PR-introduced, in the sense of incomplete coverage rather than a regression. The new unpack is only in the `else:` branch. The [`if get_processor:` branch just above](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/algorithms/utils.py#L364-L366) still calls `AutoProcessor.from_pretrained(name, trust_remote_code=True, use_fast=True)` and then takes `processor.tokenizer`, so the key is never read there.

Every caller that takes that branch:

- [`run_vlm_grpo.py:92`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/examples/run_vlm_grpo.py#L92) — always
- [`run_eval.py:117`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/examples/run_eval.py#L117) when the dataset is multimodal, so `evals/mmau.yaml` and `evals/daily_omni.yaml` drop it while `math_eval.yaml` / `mmlu*.yaml` honour it
- [`run_sft.py:195`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/examples/run_sft.py#L195) when `is_vlm`, [`run_grpo_nemo_gym.py:163`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/examples/nemo_gym/run_grpo_nemo_gym.py#L163), [`nemo_gym.py:502`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/environments/nemo_gym.py#L502)
- [`dtensor_policy_worker_v2.py:261`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/policy/workers/dtensor_policy_worker_v2.py#L261) — inside every Ray worker, which is the isolated-process case this PR exists to solve

Breaking case, using your own motivating kwarg:

```yaml
policy:
  tokenizer:
    name: mistralai/Mistral-Small-3.1-24B-Instruct-2503
    tokenizer_kwargs: {fix_mistral_regex: false}
```

Run `examples/run_vlm_grpo.py` with that: the processor branch is taken, `fix_mistral_regex` is never passed, and transformers applies its default regex path instead of your opt-out. No error and no warning — just different token ids.

**Action:** forward the kwargs in the `get_processor` branch, here and in the twin at [`automodel/setup.py:188`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/automodel/setup.py#L187-L189). Forwarding does reach the tokenizer: [`processing_utils.py:1543`](https://github.com/huggingface/transformers/blob/v5.8.1/src/transformers/processing_utils.py#L1543-L1544) calls `_load_tokenizer_from_pretrained`, which [passes `**kwargs` straight to `AutoTokenizer.from_pretrained`](https://github.com/huggingface/transformers/blob/v5.8.1/src/transformers/processing_utils.py#L1489-L1491).

One caveat — do not splat naively. `use_fast=True` is hardcoded on that call and, unlike the tokenizer path, it is meaningful for processors, so a user setting `use_fast` would hit `TypeError: got multiple values for keyword argument 'use_fast'`. One way of several:

```python
_tok_kwargs = dict(tokenizer_config.get("tokenizer_kwargs") or {})
processor = AutoProcessor.from_pretrained(
    tokenizer_config["name"],
    trust_remote_code=True,
    use_fast=_tok_kwargs.pop("use_fast", True),
    **_tok_kwargs,
)
```

Not a `suggestion` block because the lines being changed sit outside this hunk. If you would rather keep the PR minimal, raising on that branch when `tokenizer_kwargs` is set is a fine alternative — silently dropping it is the part worth avoiding.

**Context — no action.** `trust_remote_code` has the same collision shape on this line, so `tokenizer_kwargs: {trust_remote_code: false}` raises `TypeError`. It fails at startup rather than silently, so it is probably fine as-is.

nemo_rl/models/automodel/setup.py
```diff
@@ -190,7 +190,9 @@ def get_tokenizer(
         tokenizer = processor.tokenizer
     else:
         tokenizer = NeMoAutoTokenizer.from_pretrained(
-            tokenizer_config["name"], trust_remote_code=True
+            tokenizer_config["name"],
+            trust_remote_code=True,
+            **(tokenizer_config.get("tokenizer_kwargs") or {}),
```
terrykong: [`nemo_rl/models/automodel/setup.py:195`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/automodel/setup.py#L195)

**2 action items.**

**AI-1 — mirror the processor fix.** The [`AutoProcessor` branch above](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/automodel/setup.py#L187-L189) drops `tokenizer_kwargs` exactly like its twin in `nemo_rl/algorithms/utils.py`, including the same hardcoded `use_fast=True` collision hazard. The reasoning and a sketch are in the comment on that file; whatever you land there should land here too. This path serves the automodel and DTensor-v2 workers.

**AI-2 — document the key.** This docstring's "Optional keys" list [already documents `chat_template_kwargs` at line 178](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/automodel/setup.py#L178) but not `tokenizer_kwargs`, so the new key is undiscoverable from the function that consumes it. One line covers it:

```
                - tokenizer_kwargs: Extra keyword arguments forwarded to
                  NeMoAutoTokenizer.from_pretrained(), e.g. {"fix_mistral_regex": False}.
```

Neither is a `suggestion` block — both target lines outside this hunk.

nemo_rl/models/policy/__init__.py
```diff
@@ -509,6 +509,8 @@ class TokenizerConfig(TypedDict):
     chat_template: NotRequired[str | None]
     # Arguments to pass to tokenizer.apply_chat_template(...). This can be used to pass kwargs like enable_thinking=true
     chat_template_kwargs: NotRequired[dict[str, Any] | None]
+    # Arguments forwarded to AutoTokenizer.from_pretrained(...) via get_tokenizer.
+    tokenizer_kwargs: NotRequired[dict[str, Any] | None]
```
terrykong, line 513: [`nemo_rl/models/policy/__init__.py:513`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/policy/__init__.py#L513)

**TL;DR — no YAML in the repo mentions `tokenizer_kwargs`, so users have no way to discover it; both sibling keys are documented in an exemplar.**

**AI-1 — give the new key a documented default in an exemplar YAML.** That is where v1 `TypedDict` defaults live per the repo's `config-conventions` skill, and both nearest siblings follow it: `chat_template_kwargs: null` is in all seven exemplars ([e.g. here](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/examples/configs/grpo_math_1B.yaml#L129)) and `use_fastokens: false` was added to [`grpo_math_1B.yaml:130`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/examples/configs/grpo_math_1B.yaml#L130). `tokenizer_kwargs` is in none, so no YAML in the repo shows the key exists.

Add it beside `chat_template_kwargs` in [`examples/configs/grpo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/examples/configs/grpo_math_1B.yaml#L129):

```yaml
    tokenizer_kwargs: null # extra kwargs forwarded to AutoTokenizer.from_pretrained, e.g. {fix_mistral_regex: false}
```

**and add the same line to [`tests/unit/reference_configs/grpo_math_1B.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/tests/unit/reference_configs/grpo_math_1B.yaml#L135)** — [`test_config_v2.py::test_reference_configs_up_to_date`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/tests/unit/test_config_v2.py#L84) asserts every key in the real exemplar also exists in the reference, so changing only the exemplar turns that test red. One exemplar is enough; that is how `use_fastokens` was done. Not a `suggestion` block — the edit is in two files outside this diff.

**Follow-up — your call, not an ask for this PR.** For completeness on "does this reach everywhere": the Megatron generation backend builds its own tokenizer through Megatron-Bridge with hardcoded [`hf_tokenizer_kwargs={"trust_remote_code": True, "use_fast": True}`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/megatron/setup.py#L2200-L2206), and that tokenizer drives rollouts — it is handed to the [`TextGenerationController`](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/generation/megatron/megatron_worker.py#L217) and to the [text-generation server](https://github.com/NVIDIA-NeMo/RL/blob/1bab67ffc10c834e101d625a53f06cac7be94fb9/nemo_rl/models/generation/megatron/megatron_worker.py#L321). So under `generation.backend: megatron` (6 shipped recipes) a tokenization-changing kwarg would apply to training but not to rollouts. That hardcode is pre-existing and lives in a file this PR does not touch, so it is only worth a tracking issue if you think it matters.

terrykong (commented):
Reviewed with a team of specialised agents plus an adversarial pass. The core change is sound and worth having: `fix_mistral_regex` is a real transformers kwarg with no other route through `get_tokenizer` today, and threading it through config rather than a monkeypatch is the right call, especially for DTensor-v2 workers that rebuild the tokenizer locally from config.

Four comments below: one that blocks (the new test cannot pass on any pinned transformers version), one coverage gap (the processor branch, which is the half of the VLM/multimodal flow the change misses), one convention item, and one optional follow-up.

_Generated by Claude Code_

tests/unit/algorithms/test_utils.py
pstjohn: Good catch. I replaced the `use_fast`/`is_fast` assertion with a direct mock assertion that `model_max_length=123` reaches `AutoTokenizer.from_pretrained`. This removes the gated-model download and ensures the test fails if forwarding is removed. Addressed in `140bdcb6`.

nemo_rl/algorithms/utils.py
pstjohn: Addressed, both processor paths now copy `tokenizer_kwargs`, extract `use_fast` with the existing `True` default, and forward the remaining kwargs through `AutoProcessor.from_pretrained`. Copying the mapping also avoids mutating the caller's config. I added processor-forwarding coverage for both implementations.

nemo_rl/models/automodel/setup.py
pstjohn: Done, I mirrored the processor handling in the automodel implementation and documented `tokenizer_kwargs` in the function's optional-key list, including the `fix_mistral_regex` example. I also added focused tokenizer and processor forwarding tests for the automodel path.

nemo_rl/models/policy/__init__.py
pstjohn: Added `tokenizer_kwargs: null` beside `chat_template_kwargs` in both `examples/configs/grpo_math_1B.yaml` and its reference config.

I left the pre-existing Megatron generation path unchanged for this PR, as suggested; we can track that separately if we decide it needs end-to-end support.

terrykong (approved):

PR #3799, opened 2026-08-24
fix(vllm): replace fastokens-b10 with fastokens

guyueh1:
Summary: move fastokens from the base environment to the vLLM extra, switch to the canonical fastokens package required by vLLM, and translate deprecated NRL_USE_FASTOKENS to VLLM_USE_FASTOKENS.\n\nValidation: uv lock; git commit hooks passed except pyrefly, which cannot run locally because this lockfile supports Linux targets only; python3 py_compile for touched Python files; standalone fastokens env smoke test.


guyueh1:
Context/rationale for reviewers:

1. **Why switch from `fastokens-b10` to `fastokens`:** vLLM 0.25.1 gates `VLLM_USE_FASTOKENS=1` by checking for an installed distribution named `fastokens>=0.2.0` before it calls `fastokens.patch_transformers()`. The current lock resolved `fastokens-b10==0.2.3`, which can make `import fastokens` possible but does not satisfy `importlib.metadata.version("fastokens")`. I reproduced this on HSG with the requested single-node, 4-GPU vLLM serve path: baseline `vllm==0.25.1` + `fastokens-b10==0.2.3` failed in `vllm/tokenizers/fastokens.py::apply_fastokens_patch()` with `ImportError: fastokens >= 0.2.0 is required when VLLM_USE_FASTOKENS=1 (found unknown)`. After uninstalling `fastokens-b10` and installing `fastokens>=0.2.0` (`0.3.1`), the direct vLLM fastokens patch probe succeeded: `[fastokens] patch_transformers: successfully patched transformers ...` and `PATCH_OK`.

2. **Why deprecate `NRL_USE_FASTOKENS=1` in favor of `VLLM_USE_FASTOKENS=1`:** NeMo-RL's helper and vLLM's helper both ultimately perform the same process-local action: call `fastokens.patch_transformers()` before HF fast tokenizers are constructed. The difference is which process runs it. `maybe_patch_fastokens()` runs in NeMo-RL-side tokenizer processes, while vLLM's `apply_fastokens_patch()` runs in vLLM worker/server processes. Since the vLLM worker path is where this compatibility issue surfaced and vLLM already owns the canonical env var, this PR keeps `NRL_USE_FASTOKENS` as a compatibility shim that prints a deprecation note and translates it to `VLLM_USE_FASTOKENS`.

3. **Why move the dependency to the `vllm` extra:** fastokens is only required for the vLLM fast-tokenizer path. The base NeMo-RL environment does not need to carry the wheel by default, especially now that the canonical runtime switch is `VLLM_USE_FASTOKENS`. Moving `fastokens>=0.2.0` into the `vllm` extra keeps the base environment smaller while ensuring vLLM worker environments have the package vLLM expects.

guyueh1:
Follow-up update: restored the pre-existing override semantics in maybe_patch_fastokens. The config value remains the default/fallback, while VLLM_USE_FASTOKENS, when set, overrides it (1 forces on; any other value forces off). The deprecated NRL_USE_FASTOKENS path now only translates into VLLM_USE_FASTOKENS before that same override logic runs.

guyueh1:
Update after dependency placement review: fastokens is now kept in the base dependencies as fastokens>=0.2.0, replacing fastokens-b10. Reason: maybe_patch_fastokens is used by NeMo-RL-side Transformers tokenizer paths, not only by vLLM workers, so the base environment must have the compatible fastokens package when policy.tokenizer.use_fastokens or the env override enables the patch. Because uv extras include the base dependencies, uv run --extra vllm also gets fastokens from base; the vLLM extra no longer needs a duplicate fastokens declaration.

guyueh1:
Update: NRL_USE_FASTOKENS is no longer deprecated and is now the NeMo-RL top-level fastokens override. When NRL_USE_FASTOKENS is set, NeMo-RL mirrors it into VLLM_USE_FASTOKENS so vLLM workers follow the same setting. If only VLLM_USE_FASTOKENS is set, NeMo-RL leaves it alone and does not let it control the NeMo-RL-side transformers patch; it remains a vLLM-only flag in that case.

guyueh1:
@ananthsub @yuki-97 can you help review this? this is for enabling VLLM_FASTOKENS=1 (similar to NRL_USE_FASTOKENS), right now this feature is broken in main

ananthsub (dismissed):

ray.sub
```diff
@@ -124,12 +124,19 @@ fi
...
 if [[ -n "${NRL_USE_FASTOKENS:-}" ]]; then
-  export NRL_USE_FASTOKENS
+  if [[ "${NRL_USE_FASTOKENS}" == "1" ]]; then
+    export VLLM_USE_FASTOKENS=1
+  else
+    export VLLM_USE_FASTOKENS=0
+  fi
+fi
+if [[ -n "${VLLM_USE_FASTOKENS:-}" ]]; then
+  export VLLM_USE_FASTOKENS
```
guyueh1: [`ray.sub:131-139`](https://github.com/NVIDIA-NeMo/RL/blob/5c5ef3987196b2c6249d9d1037ef1a026fe33fc4/ray.sub#L131-L139)

**1 action item.**

The Python mirror (`normalize_fastokens_env()` in [`nemo_rl/utils/fastokens.py`](https://github.com/NVIDIA-NeMo/RL/blob/5c5ef3987196b2c6249d9d1037ef1a026fe33fc4/nemo_rl/utils/fastokens.py#L42-L48)) and this bash block implement the same `NRL_USE_FASTOKENS → VLLM_USE_FASTOKENS` rule independently, and they diverge on `NRL_USE_FASTOKENS=""` (declared-but-empty): Python's `os.environ.get()` returns `""` (not `None`), so it proceeds and forces `VLLM_USE_FASTOKENS="0"`; bash's `[[ -n "${NRL_USE_FASTOKENS:-}" ]]` is false for `""`, so the whole block is skipped and `VLLM_USE_FASTOKENS` is left untouched — inconsistent with the documented "anything else forces off" semantics.

### AI-1
**Action:** use a presence test that matches Python's `is None` check, e.g. `${NRL_USE_FASTOKENS+set}` instead of `-n`, or add a comment cross-referencing `normalize_fastokens_env()` so the two don't re-diverge — this mirroring rule has already changed several times over this PR's review.

nemo_rl/models/generation/vllm/vllm_generation.py
```diff
@@ -210,6 +210,10 @@ def __init__(
         else:
             worker_builder = RayWorkerBuilder(worker_cls, config)
 
+        from nemo_rl.utils.fastokens import normalize_fastokens_env
```
guyueh1: [`vllm_generation.py:213`](https://github.com/NVIDIA-NeMo/RL/blob/5c5ef3987196b2c6249d9d1037ef1a026fe33fc4/nemo_rl/models/generation/vllm/vllm_generation.py#L213)

**1 action item.**

This in-function import has no circular-import or heavy-dependency reason to be deferred — `nemo_rl/utils/fastokens.py` only imports `logging`/`os` at module scope. The file's own convention (see [line 866](https://github.com/NVIDIA-NeMo/RL/blob/5c5ef3987196b2c6249d9d1037ef1a026fe33fc4/nemo_rl/models/generation/vllm/vllm_generation.py#L866)) justifies its deferred import with a one-line comment; this one breaks that precedent.

### AI-1
**Action:** hoist `from nemo_rl.utils.fastokens import normalize_fastokens_env` to the top-level imports of this file.

```suggestion
        normalize_fastokens_env()
```

tests/unit/utils/test_fastokens.py
```diff
@@ -70,26 +72,65 @@ def test_patches_once_when_config_enabled(monkeypatch):
...
-    fastokens_util.maybe_patch_fastokens(True)  # config on, env forces off
+    fastokens_util.maybe_patch_fastokens(True)
 
     assert calls == []
     assert fastokens_util._patched is False
+    assert os.environ["VLLM_USE_FASTOKENS"] == "0"
+    assert capsys.readouterr().err == ""
+
+
+def test_nrl_env_overrides_explicit_vllm_env(monkeypatch):
```
guyueh1, line 147: [`test_fastokens.py:123`](https://github.com/NVIDIA-NeMo/RL/blob/5c5ef3987196b2c6249d9d1037ef1a026fe33fc4/tests/unit/utils/test_fastokens.py#L123)

**No action needed** — informational, three real but non-blocking coverage gaps.

1. No test for `NRL_USE_FASTOKENS` set to a non-"1"/"0" value (e.g. `"true"`, `""`), verifying it forces off and mirrors `VLLM_USE_FASTOKENS="0"`.
2. The new call site at [`vllm_generation.py:213-215`](https://github.com/NVIDIA-NeMo/RL/blob/5c5ef3987196b2c6249d9d1037ef1a026fe33fc4/nemo_rl/models/generation/vllm/vllm_generation.py#L213-L215) has zero test coverage.
3. `ray.sub`'s new bash block has no shell-level test.

(1) is cheap to add:
```suggestion
def test_nrl_env_invalid_value_forces_off_and_mirrors_zero(monkeypatch):
    calls = []
    _install_fake_fastokens(monkeypatch, lambda: calls.append(1))
    monkeypatch.setenv("NRL_USE_FASTOKENS", "true")
    fastokens_util.maybe_patch_fastokens(True)
    assert calls == []
    assert os.environ["VLLM_USE_FASTOKENS"] == "0"
```
(2) and (3) are lower priority — no existing lightweight test pattern to hang them on (`VllmGeneration.__init__` currently requires a real Ray cluster in existing tests; `ray.sub` has no shell test harness). Fine to defer.

pyproject.toml
```diff
@@ -70,7 +70,7 @@ dependencies = [
   "pybase64",                                                                                                 # for sglang refit
   "awscrt>=0.35.0",                                                                                           # for parallel S3 refit transport
   "zstandard",                                                                                                # for sparse refit body compression
-  "fastokens-b10>=0.1.1",                                                                                     # Rust-backed BPE tokenizer (~10x faster encode); enable at runtime with NRL_USE_FASTOKENS=1
+  "fastokens>=0.2.0",                                                                                         # Rust-backed BPE tokenizer (~10x faster encode); enable at runtime with NRL_USE_FASTOKENS=1
```
guyueh1, line 73: [`pyproject.toml:73`](https://github.com/NVIDIA-NeMo/RL/blob/5c5ef3987196b2c6249d9d1037ef1a026fe33fc4/pyproject.toml#L73)

**No action needed** — optional follow-up, not introduced by this PR.

`examples/configs/ppo_math_1B_megatron_single_controller.yaml:140` still says "requires the fastokens-b10 wheel" — that file isn't touched by this PR, but `grpo_math_1B.yaml` and `tests/unit/reference_configs/grpo_math_1B.yaml` got the equivalent comment updated here. Since you're already touching this exact comment pattern in sibling configs, worth a quick follow-up edit if convenient.

guyueh1 (commented):
⚠️ **Merge conflict**: this PR is currently `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY` against `main`. Please rebase before this can be merged.

Reviewed with a 6-agent team (RL expert, bug finder, design reviewer, test agent, comment reviewer, devil's advocate). The root-cause fix (switching `fastokens-b10` → `fastokens>=0.2.0` to satisfy vLLM 0.25.1's `importlib.metadata.version("fastokens")>=0.2.0` gate) was independently verified against vendored vLLM source — confirmed correct. Two low-severity findings below; nothing blocking.

_Generated by Claude Code_

examples/configs/ppo_math_1B_megatron_single_controller.yaml
```diff
@@ -145,8 +145,8 @@ checkpointing:
 
 policy:
   tokenizer:
-    # Opt-in Rust-backed BPE tokenizer (~10x faster encode); requires the
-    # fastokens-b10 wheel. NRL_USE_FASTOKENS overrides at runtime.
+    # Opt-in Rust-backed BPE tokenizer (~10x faster encode); requires the fastokens wheel.
+    # NRL_USE_FASTOKENS overrides at runtime.
```
yuki-97: nit: this comment still describes the pre-PR semantics, unlike the other two surfaces the PR updated.

[`grpo_math_1B.yaml:139`](https://github.com/NVIDIA-NeMo/RL/blob/85e34159f800b71f3f03191f4f495c6258268eaa/examples/configs/grpo_math_1B.yaml#L139) and [`policy/__init__.py:520`](https://github.com/NVIDIA-NeMo/RL/blob/85e34159f800b71f3f03191f4f495c6258268eaa/nemo_rl/models/policy/__init__.py#L520) both now say the env var also sets `VLLM_USE_FASTOKENS`; this one stops at "overrides at runtime".

```suggestion
    # Opt-in Rust-backed BPE tokenizer (~10x faster encode); requires the fastokens wheel.
    # NRL_USE_FASTOKENS overrides this and also sets VLLM_USE_FASTOKENS.
```

tests/unit/utils/test_fastokens.py
```diff
@@ -70,26 +72,89 @@ def test_patches_once_when_config_enabled(monkeypatch):
...
+    calls = []
+    _install_fake_fastokens(monkeypatch, lambda: calls.append(1))
+    monkeypatch.setenv("NRL_USE_FASTOKENS", "1")
+    monkeypatch.setenv("VLLM_USE_FASTOKENS", "0")
+
+    fastokens_util.maybe_patch_fastokens(False)
+
+    assert calls == [1]
+    assert fastokens_util._patched is True
+    assert os.environ["VLLM_USE_FASTOKENS"] == "1"
```
yuki-97: nit: the inverse of this case is uncovered: `NRL_USE_FASTOKENS=0` with `VLLM_USE_FASTOKENS=1` should mirror down to `"0"` and override the user's explicit vLLM opt-in.

That is the direction that silently turns the feature off, so it is the more valuable of the two to pin.

**Suggest fix:** mirror this test with the two values swapped.

ananthsub (approved):

yuki-97 (approved):

PR #3648, opened 2026-08-14
feat(video): add Gym support for sync and async GRPO - v2

ehosseiniasl:
## Summary

Add end-to-end static-prompt video support for NeMo Gym GRPO with synchronous and asynchronous Megatron training.

This PR:

- accepts one local video in the initial Gym prompt, either as a source video or a cached-frame manifest;
- carries video metadata, policy frame tensors, native token IDs, and log probabilities through Gym rollout, batching, generation, and policy training;
- matches stock vLLM 0.25.1 Nemotron video sampling on the policy side and registers the compatible video loader without requiring a custom vLLM fork;
- trains the canonical processor-expanded Nemotron Omni model through the pinned Megatron-Bridge integration;
- supports both synchronous and asynchronous video GRPO;
- adds dataset preparation, product recipes, documentation, focused tests, and nightly coverage.

## Scope

This integration supports static prompt media: one video known before rollout and attached to the initial user turn. Environment-produced video in later trajectory turns is not supported by this PR.

Generic GRPO step-limit semantics, SingleController behavior, async lifecycle/checkpoint changes, and unrelated refit ordering are intentionally outside this PR. It does not patch vLLM source interpolation behavior.

## Validation

- Ruff formatting and lint checks passed.
- Pyrefly passed for supported changed modules.
- Exact rebased candidate: `704 passed, 6 skipped` across recipe/config, dataset conversion, Gym video, multimodal rollout, Nemotron preprocessing, and stock-vLLM compatibility coverage.
- The 2-node synchronous run is logging raw, unmasked sequence probability error and reward to W&B; exact-final 2-node synchronous and 16-node asynchronous evidence will be added before merge.
- No credentials, personal launchers, internal accounts, or debugging-only artifacts are included.

## Related integrations

- [NVIDIA-NeMo/Megatron-Bridge#5304](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5304) `feat(video): enable canonical Nemotron Omni training for V2 MoE checkpoints`
- [NVIDIA-NeMo/Gym#2324](https://github.com/NVIDIA-NeMo/Gym/pull/2324) `feat(video): support video inputs in vLLM responses`

Both upstream integrations are merged, and the RL submodules are pinned to commits containing them. No separate Megatron-LM change is required.

PR #3802, opened 2026-08-24
feat(length-reward): support global reasoning profile bands

ehosseiniasl:
## Summary

- Allow a global reasoning profile band under `grpo.length_bonus.default` when dataset rows do not provide profile metadata.
- Preserve row-level and agent-level overrides.
- Compute reasoning length from generated assistant token IDs only, split at the configured reasoning-end token.
- Clamp the post-band multiplier to its configured floor.
- Add focused utility and async propagation coverage.

## Motivation

Video async GRPO uses one shared reasoning-length profile across the cached dataset. Requiring every row to carry identical profile metadata is unnecessary and makes Arushi-style video recipes harder to configure safely.

## Validation

- 13 focused length-adjustment tests pass.
- Async collector regression confirms `length_adjustment_config` reaches Gym rollouts.
- Ruff checks pass for the changed implementation and tests.

Full 2-node synchronous and 16-node asynchronous video validation is pending; this PR remains draft until those runs complete.

## Dependency

- NVIDIA-NeMo/RL#3604 — generic length adjustments and async Gym propagation.

The Arushi-aligned video recipes are being validated on the separate integration branch `ehsan/adi-video-profile-integration` because the current #3604 branch predates the merged video recipes from NVIDIA-NeMo/RL#3648.


ehosseiniasl:
Closing for now while runtime validation and upstream dependency work continue. I will reopen or create the review PR only when it is ready.

PR #3803, opened 2026-08-24
feat(vllm): generalize multimodal prompt data

pstjohn:
NeMo RL represents vLLM multimodal inputs as four parallel batch fields:

```text
vllm_content
vllm_images
vllm_audios
vllm_videos
```

The processor, collators, rollout transport, evaluation path, payload metrics, and prompt formatter each know some or all of those modality-specific names. The formatter then reconstructs the mapping that vLLM accepts as `multi_modal_data`.

vLLM's input contract is already modality-neutral: its [multimodal inputs documentation](https://docs.vllm.ai/en/stable/features/multimodal_inputs/) defines `multi_modal_data` as a dictionary and shows both singleton values and lists of items under modality keys. The parallel NeMo RL fields therefore make adding another vLLM modality require framework changes even when vLLM already supports it.

In this PR, we migrate nemo-RL to use the vllm contract directly
```python
{
    "vllm_content": prompt,
    "vllm_multi_modal_data": {
        "image": image,
        "audio": (audio, sample_rate),
        "video": frames,
        "conditioning": embedding,
    },
}
```

NeMo-RL currently carries vLLM multimodal inputs in separate image, audio, and video batch fields. This PR replaces them with vllm_multi_modal_data, a per-row mapping forwarded as vLLM's multi_modal_data through processing, collation, rollouts, evaluation, and prompt formatting.

NeMo-RL forwards modality keys without maintaining its own allowlist. The served vLLM model still validates which keys and payloads it supports. The conditioning key in the unit test is a synthetic unknown key used to verify forwarding; it does not demonstrate vLLM support for that modality.

The processor preserves singleton values and lists of multiple items. Formatter tests cover unknown-key forwarding and omission of empty media values; eval tests cover formatted, text-only, placeholder-style, and truncated rows. Dynamo rejects either vLLM prompt column because direct multimodal generation is unsupported.


nemo_rl/evals/eval.py
```diff
@@ -326,44 +326,35 @@ async def _run_env_eval_impl(
...
                 # Note: utils.py's format_prompt_for_vllm_generation uses pre-tokenized
                 # prompt_token_ids instead, since the training pipeline already has
                 # input_ids tensors. Both are valid vLLM inputs but may tokenize
                 # slightly differently.
-                content = [message["content"] for message in message_log]
-                content = "\n".join(content)
-                prompts.append(content)
-                prompts_for_display.append(content)
+                prompts.append(fallback_content)
+                prompts_for_display.append(fallback_content)
```
pstjohn: [`nemo_rl/evals/eval.py:335-336`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/evals/eval.py#L335-L336)

**TL;DR — `fallback_content` is now computed for every row, but VLM rows carry a *list* of typed content parts, so multimodal eval dies with `TypeError: sequence item 0: expected str instance, list found` before the first prompt is built.**

PR-introduced: the `"\n".join(...)` moved out of the text-only `else` branch, so it now runs unconditionally.

1. [`MMAUDataset.format_data`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/datasets/eval_datasets/mmau.py#L77-L80) builds `user_content` as a list of typed parts unconditionally — same for [`daily_omni`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/datasets/eval_datasets/daily_omni.py#L61-L69) (`[video, audio, text]`).
2. [`vlm_hf_data_processor`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/processors.py#L504) keeps it as a list in `user_message["content"]` — nothing on that path reassigns it to a string.
3. [`eval_collate_fn`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/collate_fn.py#L119) passes `message_log` straight through.
4. [`eval.py:336`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/evals/eval.py#L336) joins it → `TypeError: sequence item 0: expected str instance, list found`.

Both shipped multimodal eval configs are affected (`examples/configs/evals/mmau.yaml`, `examples/configs/evals/daily_omni.yaml`), deterministically on the first batch; text-only eval configs are fine. `tests/functional/eval_audio.sh` is [`run_test fast`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/tests/functional/L1_Functional_Tests_Eval.sh#L39), so L1 should surface this.

**Action:** make the join lazy so it only runs on the text-only path. This also drops the `vllm_content is None` + non-empty-media branch, which would re-crash on the same join — and which no shipped producer can reach anyway ([truncation sets both to `None`/`{}`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/processors.py#L665-L669)).

Verified locally: with this applied the regression test below passes; without it, that test fails.

```suggestion
            vllm_content = (
                vllm_content_rows[i] if vllm_content_rows is not None else None
            )
            multi_modal_data = (
                multi_modal_rows[i] if multi_modal_rows is not None else None
            )
            if vllm_content is not None:
                prompt_dict = {"prompt": vllm_content}
                if multi_modal_data:
                    prompt_dict["multi_modal_data"] = multi_modal_data
                prompts.append(prompt_dict)
                prompts_for_display.append(vllm_content)
            else:
                # Text-only fallback: use raw prompt strings (vLLM will tokenize them).
                # Note: utils.py's format_prompt_for_vllm_generation uses pre-tokenized
                # prompt_token_ids instead, since the training pipeline already has
                # input_ids tensors. Both are valid vLLM inputs but may tokenize
                # slightly differently. Only text-only rows have joinable
                # string content; VLM rows carry a list of typed content parts.
                fallback_content = "\n".join(
                    message["content"] for message in message_log
                )
                prompts.append(fallback_content)
                prompts_for_display.append(fallback_content)
```

nemo_rl/evals/eval.py
```diff
@@ -326,44 +326,35 @@ async def _run_env_eval_impl(
         if num_tests_per_prompt > 1:
             batch = batch.repeat_interleave(num_tests_per_prompt)
 
-        # get input prompt from message_log
-        is_multimodal = "vllm_content" in batch
+        # Get input prompts from message_log, preserving any vLLM-ready data.
+        vllm_content_rows = batch.get("vllm_content", None)
+        multi_modal_rows = batch.get("vllm_multi_modal_data", None)
```
pstjohn, line 331: [`nemo_rl/evals/eval.py:331`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/evals/eval.py#L331)

**1 action item.**

This prompt-building loop has no unit coverage — [`test_eval.py:61`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/tests/unit/evals/test_eval.py#L61) monkeypatches `_run_env_eval_impl` out entirely, which is why the crash above ships with a green suite. Pre-existing gap, but this PR rewrote the loop.

**Action:** add a regression test to `tests/unit/evals/test_eval.py` (not in this diff). One important detail — the multimodal row's fixture must use **list-shaped** `content`; a plain-string fixture passes even with the bug present, so it would give false confidence.

```python
def test_run_env_eval_impl_builds_vllm_multi_modal_prompts(monkeypatch):
    """The eval prompt loop must forward per-row ``multi_modal_data`` to vLLM.

    Row 0 is a VLM row: list-shaped ``content`` plus vLLM-ready prompt data.
    Row 1 is text-only and must fall back to the joined message_log text.
    """
    import asyncio

    from nemo_rl.distributed.batched_data_dict import BatchedDataDict
    from nemo_rl.evals.eval import _run_env_eval_impl

    image = object()
    batch = BatchedDataDict(
        {
            "message_log": [
                [
                    {
                        "role": "user",
                        "content": [
                            {"type": "image", "image": image},
                            {"type": "text", "text": "describe the image"},
                        ],
                    }
                ],
                [{"role": "user", "content": "plain"}],
            ],
            "extra_env_info": [{"ground_truth": "a"}, {"ground_truth": "b"}],
            "vllm_content": ["<image> describe the image", None],
            "vllm_multi_modal_data": [{"image": image}, {}],
        }
    )

    captured = {}

    async def fake_generate_texts(vllm_generation, inputs, use_async):
        captured["prompts"] = list(inputs["prompts"])
        return ["out-0", "out-1"]

    monkeypatch.setattr("nemo_rl.evals.eval._generate_texts", fake_generate_texts)
    monkeypatch.setattr("nemo_rl.evals.eval._print_results", lambda *a, **k: None)
    monkeypatch.setattr("nemo_rl.evals.eval.ray.get", lambda value: value)

    env = SimpleNamespace(
        step=SimpleNamespace(
            remote=lambda *args: SimpleNamespace(rewards=torch.tensor([1.0, 0.0]))
        ),
        shutdown=SimpleNamespace(remote=lambda: None),
    )
    master_config = SimpleNamespace(
        generation={"backend": "vllm"},
        eval={
            "metric": "pass@k",
            "num_tests_per_prompt": 1,
            "k_value": 1,
            "save_path": None,
        },
    )

    class _SingleBatchDataloader(list):
        dataset = [0, 1]

    asyncio.run(
        _run_env_eval_impl(
            vllm_generation=SimpleNamespace(shutdown=lambda: None),
            dataloader=_SingleBatchDataloader([batch]),
            env=env,
            master_config=master_config,
        )
    )

    prompts = captured["prompts"]
    assert prompts[0] == {
        "prompt": "<image> describe the image",
        "multi_modal_data": {"image": image},
    }
    assert prompts[1] == "plain"
```

Verified: fails on this commit, passes with the fix above (`12 passed`). `SimpleNamespace`, `torch` and `pytest` are already imported in that file.

tests/unit/data/datasets/test_mmpr_tiny.py
```diff
@@ -233,8 +233,8 @@ def test_processor_produces_valid_datum_spec(self, tiny_image_path):
         assert "ground_truth" in result["extra_env_info"]
         assert result["extra_env_info"]["ground_truth"] == "A"
         assert "vllm_content" in result
-        assert "vllm_images" in result
-        assert len(result["vllm_images"]) == 1
+        assert "vllm_multi_modal_data" in result
+        assert "image" in result["vllm_multi_modal_data"]
```
pstjohn: [`tests/unit/data/datasets/test_mmpr_tiny.py:237`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/tests/unit/data/datasets/test_mmpr_tiny.py#L237)

**1 action item.**

PR-introduced. This replaced `assert len(result["vllm_images"]) == 1`, which dropped the only arity assertion — right as the singleton collapse ([`items[0] if len(items) == 1 else items`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/processors.py#L657-L660)) moved into the processor and became the load-bearing behavior of this PR.

I mutated that collapse to plain `items` and the entire suite still passed (160 tests across the six multimodal test files), so nothing currently pins it.

**Action:** pin the collapse here. Verified: passes on clean source, fails under that mutation.

```suggestion
        assert isinstance(result["vllm_multi_modal_data"]["image"], Image.Image)
```

Worth adding a multi-item case too, asserting a 2-image datum yields a `list` — the existing `_make_stub_nemotron_processor(num_tiles=2)` stub makes that cheap and needs no real HF processor.

nemo_rl/data/collate_fn.py
```diff
@@ -45,21 +45,18 @@ def rl_collate_fn(data_batch: list[DatumSpec]) -> BatchedDataDict[Any]:
     # Extract stop_strings if present
     stop_strings = [datum.get("stop_strings", None) for datum in data_batch]
 
-    # check if any of the data batch has vllm content and images
+    # Preserve vLLM-ready prompt data without interpreting its modality keys.
     extra_args = {}
     if any(
-        [datum_spec.get("vllm_content", None) is not None for datum_spec in data_batch]
+        "vllm_content" in datum_spec or "vllm_multi_modal_data" in datum_spec
```
pstjohn, line 51: [`nemo_rl/data/collate_fn.py:51`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/collate_fn.py#L51)

**1 action item.**

PR-introduced: the gate changed from a value check (`.get("vllm_content") is not None`) to key-presence, here and in the twin [`eval_collate_fn`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/collate_fn.py#L122). That's the *correct* predicate — a truncated VLM row legitimately carries [`vllm_content=None`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/data/processors.py#L665-L669) and must still travel — but nothing tests it: `eval_collate_fn` has zero unit coverage, and `tests/unit/data/test_collate_fn.py` only exercises `preference_collate_fn`.

**Action:** add to `tests/unit/data/test_collate_fn.py` (not in this diff; needs `import pytest` and `from nemo_rl.data.collate_fn import eval_collate_fn, rl_collate_fn`).

```python
def _vllm_datum(idx: int, **vllm_kwargs) -> DatumSpec:
    """Minimal DatumSpec carrying only the fields both collators need."""
    return DatumSpec(
        message_log=[
            {"role": "user", "content": "hi", "token_ids": torch.tensor([1, 2])}
        ],
        length=2,
        loss_multiplier=1.0,
        extra_env_info={"ground_truth": "a"},
        idx=idx,
        task_name="vlm",
        **vllm_kwargs,
    )


@pytest.mark.parametrize("collate_fn", [rl_collate_fn, eval_collate_fn])
def test_collate_fn_omits_vllm_columns_for_text_only_batch(collate_fn):
    batch = collate_fn([_vllm_datum(0), _vllm_datum(1)])
    assert "vllm_content" not in batch
    assert "vllm_multi_modal_data" not in batch


@pytest.mark.parametrize("collate_fn", [rl_collate_fn, eval_collate_fn])
def test_collate_fn_keeps_vllm_columns_for_fully_truncated_batch(collate_fn):
    """Over-length rows set vllm_content=None / data={}, so the gate must key
    off presence rather than value or the vLLM path silently loses its columns."""
    batch = collate_fn(
        [_vllm_datum(i, vllm_content=None, vllm_multi_modal_data={}) for i in range(2)]
    )
    assert batch["vllm_content"] == [None, None]
    assert batch["vllm_multi_modal_data"] == [{}, {}]


@pytest.mark.parametrize("collate_fn", [rl_collate_fn, eval_collate_fn])
def test_collate_fn_pads_rows_missing_vllm_keys(collate_fn):
    image = MagicMock()
    batch = collate_fn(
        [
            _vllm_datum(
                0, vllm_content="<image> q", vllm_multi_modal_data={"image": image}
            ),
            _vllm_datum(1),
        ]
    )
    assert batch["vllm_content"] == ["<image> q", None]
    assert batch["vllm_multi_modal_data"] == [{"image": image}, {}]
```

Verified passing; reverting the gate to the old value check fails the truncated case.

nemo_rl/utils/multimodal_payload_metrics.py
```diff
@@ -200,29 +200,29 @@ def visit(key: str, value: Any) -> None:
...
             )
             totals["logical_media_bytes"] += _typed_content_media_nbytes(value)
             totals["physical_segments"] += _typed_content_media_segment_count(
-                value, seen_native_segments
+                value, seen_vllm_segments
             )
             totals["logical_segments"] += _typed_content_media_segment_count(value)
-        elif key in NATIVE_MULTIMODAL_KEYS:
-            totals["physical_media_bytes"] += _value_nbytes(value, seen_native_leaves)
+        elif key in VLLM_PROMPT_KEYS:
```
pstjohn: [`nemo_rl/utils/multimodal_payload_metrics.py:210`](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/utils/multimodal_payload_metrics.py#L210)

**1 action item — nit.**

Behavior is correct, but the dispatch reads misleadingly: `VLLM_PROMPT_KEYS` contains `vllm_content`, which [line 201](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/utils/multimodal_payload_metrics.py#L201) already catches, so this branch is only ever reachable for `vllm_multi_modal_data`.

**Action:** say what you mean (also needs `VLLM_MULTI_MODAL_DATA_KEY` added to the [existing import](https://github.com/NVIDIA-NeMo/RL/blob/df8af3e3af835daa1659036561869f7f80d5461e/nemo_rl/utils/multimodal_payload_metrics.py#L24-L28)).

```suggestion
        elif key == VLLM_MULTI_MODAL_DATA_KEY:
```

pstjohn (commented):
Reviewed by a team of 5 agents (RL expert, bug finder, test author, design reviewer, plus an adversarial pass that challenged every finding).

**The design is right.** Tracing the extensibility claim: adding a new modality went from ~7 sites across 5 files to 1 file, and `test_vllm_utils_vlm_with_custom_modality` demonstrates a novel modality flowing end-to-end with no GPU, Ray, or model. Unifying was the correct call — the pre-PR code carried three literally-identical 4-line collapse blocks across two files. The rename is complete (zero stale references repo-wide, including the Gym submodule), the metrics layer stayed genuinely modality-blind, and the collate gate change to key-presence is a quiet correctness improvement for truncated rows.

One blocking regression in the eval path, plus three test-coverage asks and one nit. Linters are green (`ruff`, `ruff-format`, `taplo`, `pyrefly`).

Nine further candidate findings were dropped in the adversarial pass — including two that were confidently argued but falsified by direct testing, and several proposing new abstractions on a PR whose goal was to remove them.

_Generated by Claude Code_

nemo_rl/evals/eval.py
pstjohn: fixed in 767f33ffb49a69a5186e3d1f497521a535b7e223

nemo_rl/evals/eval.py
pstjohn: fixed in 767f33ffb49a69a5186e3d1f497521a535b7e223

tests/unit/data/datasets/test_mmpr_tiny.py
pstjohn: fixed in 767f33ffb49a69a5186e3d1f497521a535b7e223

nemo_rl/data/collate_fn.py
pstjohn: fixed in 767f33ffb49a69a5186e3d1f497521a535b7e223

nemo_rl/utils/multimodal_payload_metrics.py
pstjohn: fixed in 767f33ffb49a69a5186e3d1f497521a535b7e223 — with one addition: the suggestion as posted left `VLLM_PROMPT_KEYS` imported but unused, which fails ruff, so that import was dropped too.

terrykong (commented):
@rohitrango can you review?

nemo_rl/evals/eval.py
```diff
@@ -381,11 +369,13 @@ async def _run_env_eval_impl(
...
-                # slightly differently.
-                content = [message["content"] for message in message_log]
-                content = "\n".join(content)
-                prompts.append(content)
-                prompts_for_display.append(content)
+                # slightly differently. Only text-only rows have joinable
+                # string content; VLM rows carry a list of typed content parts.
+                fallback_content = "\n".join(
+                    message["content"] for message in message_log
+                )
```
rohitrango: [`nemo_rl/evals/eval.py:372-376`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/evals/eval.py#L372-L376)

The new comment says "Only text-only rows have joinable string content", but one VLM row shape still reaches this join and raises `TypeError: sequence item 0: expected str instance, list found`.

A truncated VLM row gets both gates falsy at the same time:
- [`processors.py:796-801`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/data/processors.py#L796-L801) sets `vllm_content=None` **and** `vllm_multi_modal_data={}` when `length >= max_seq_length`. It trims `token_ids` and the `PackedTensor`s, but `message["content"]` stays the typed-part list built at [`processors.py:690-716`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/data/processors.py#L690-L716).
- [`eval.py:341`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/evals/eval.py#L341) is `False`; [`eval.py:348`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/evals/eval.py#L348) is `False` because `{}` is falsy. The row lands here.

Nothing upstream drops it — [`eval_collate_fn`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/data/collate_fn.py#L113-L135) never reads `loss_multiplier`, so the processor's masking has no effect in eval and a single over-length sample fails the whole run rather than scoring 0.

This shape is on `main` too, so it is not from this PR. Raising it because the PR rewrote this loop and [`test_collate_fn.py`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/tests/unit/data/test_collate_fn.py#L676) already tests this exact row one layer down. `str(...)` matches the sibling branch at [`eval.py:363`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/evals/eval.py#L363) and gives the "reward 0" outcome that [`daily_omni.yaml:15-17`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/examples/configs/evals/daily_omni.yaml#L15-L17) says is expected.

```suggestion
                # Text-only fallback: use raw prompt strings (vLLM will tokenize them).
                # Note: utils.py's format_prompt_for_vllm_generation uses pre-tokenized
                # prompt_token_ids instead, since the training pipeline already has
                # input_ids tensors. Both are valid vLLM inputs but may tokenize
                # slightly differently. Truncated VLM rows also reach this branch, so
                # stringify: their content is still a list of typed parts.
                fallback_content = "\n".join(
                    str(message["content"]) for message in message_log
                )
```

Aside: the comment on [`eval.yaml:49`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/examples/configs/evals/eval.yaml#L49) calls `max_input_seq_length` "useless since we directly use prompts in evaluation". It is not — [`run_eval.py:80`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/examples/run_eval.py#L80) passes it as `max_seq_length`, which is what drives this truncation branch. Worth correcting while you are here.

tests/unit/models/generation/test_vllm_utils.py
```diff
@@ -87,23 +90,42 @@ def test_vllm_utils_vlm_with_images_and_text():
...
+            "vllm_multi_modal_data": [{"image": "img", "conditioning": embedding}],
+        }
+    )
+
+    prompts = format_prompt_for_vllm_generation(data)
+
+    multi_modal_data = prompts[0]["multi_modal_data"]
+    assert prompts[0]["prompt_token_ids"] == [1, 2]
+    assert multi_modal_data["image"] == "img"
+    assert multi_modal_data["conditioning"] is embedding
```
rohitrango: [`tests/unit/models/generation/test_vllm_utils.py:93-108`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/tests/unit/models/generation/test_vllm_utils.py#L93-L108)

The pass-through logic this asserts is correct, but the test name and the PR description both suggest `"conditioning"` is a modality vLLM accepts. At the pinned vLLM **0.25.1** ([`pyproject.toml:151-164`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/pyproject.toml#L151-L164)) it is rejected:

- [`parse.py:700-710`](https://github.com/vllm-project/vllm/blob/752a3a504485790a2e8491cacbb35c137339ad34/vllm/multimodal/parse.py#L700-L710) — `parse_mm_data` raises for any unregistered key:
  ```python
  for k, v in mm_data.items():
      if k not in subparsers:
          raise ValueError(f"Unsupported modality: {k}")
  ```
- [`parse.py:692-698`](https://github.com/vllm-project/vllm/blob/752a3a504485790a2e8491cacbb35c137339ad34/vllm/multimodal/parse.py#L692-L698) — the base subparsers are exactly `audio`, `image`, `video`, `vision_chunk`.
- Reached via [`renderers/base.py:745`](https://github.com/vllm-project/vllm/blob/752a3a504485790a2e8491cacbb35c137339ad34/vllm/renderers/base.py#L745) → `context.py:442` → the raise above.

We checked the 0.25.1 wheel directly: `_get_subparsers` has zero overrides, and the single `parse_mm_data` override (`terratorch.py:118`) only re-wraps the existing `"image"` key. So no 0.25.1 model registers a key outside those four. A model *can* extend the set by overriding [`get_data_parser()`](https://github.com/vllm-project/vllm/blob/752a3a504485790a2e8491cacbb35c137339ad34/vllm/multimodal/processing/context.py#L353-L365) — that is the documented path, and it is version-specific.

The generalization is a clear win for the registered modalities; only the framing overstates it. Consider renaming to something like `test_vllm_utils_forwards_unknown_modality_keys` and adjusting the PR description to say NeMo-RL is modality-agnostic while vLLM still validates the key against the served model.

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -182,37 +186,26 @@ def _get_regular_prompt(index: int):
...
-            )
-        return multi_modal_data
+        if multi_modal_rows is None or multi_modal_rows[index] is None:
+            return {}
+        return dict(multi_modal_rows[index])
 
-    # Native image, audio, and video side channels share this formatter path.
-    if "vllm_content" in data:
+    # vLLM-ready content and modality data share this formatter path.
+    if content_rows is not None or multi_modal_rows is not None:
```
rohitrango, line 205: [`nemo_rl/models/generation/vllm/utils.py:197-198`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/models/generation/vllm/utils.py#L197-L198)

This gate makes `vllm_multi_modal_data` a prompt shape that stands on its own — a row with media and no `vllm_content` now builds a real multimodal prompt ([`utils.py:201-208`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/models/generation/vllm/utils.py#L201-L208)), which the new `test_vllm_utils_vlm_with_custom_modality` exercises.

The Dynamo guard was not widened to match. [`dynamo_generation.py:377`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/models/generation/dynamo/dynamo_generation.py#L377) still names one key:

```python
if "vllm_content" in data:
    raise NotImplementedError(...)
```

A batch carrying only `vllm_multi_modal_data` passes that check, and Dynamo drops the media and returns a text-only completion instead of raising.

No live trigger today — both [`rl_collate_fn`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/data/collate_fn.py#L50-L59) and [`eval_collate_fn`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/data/collate_fn.py#L121-L130) always write both columns when either key is present, so only a hand-built batch reaches it. Flagging it because the guard should track the same key set the formatter reads, and `VLLM_PROMPT_KEYS` already exists for exactly this. One line in `dynamo_generation.py` (not in this diff):

```python
if VLLM_PROMPT_KEYS & data.keys():
    raise NotImplementedError(
        "DynamoGeneration direct generate() supports token-ID LLM "
        "prompts only; multimodal vLLM prompt data is not supported."
    )
```

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -182,37 +186,26 @@ def _get_regular_prompt(index: int):
...
-            )
-        videos = data.get("vllm_videos", None)
-        if videos is not None and len(videos[index]) > 0:
-            multi_modal_data["video"] = (
-                videos[index][0] if len(videos[index]) == 1 else videos[index]
-            )
-        return multi_modal_data
+        if multi_modal_rows is None or multi_modal_rows[index] is None:
+            return {}
+        return dict(multi_modal_rows[index])
```
rohitrango: [`nemo_rl/models/generation/vllm/utils.py:192-195`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/models/generation/vllm/utils.py#L192-L195)

`dict(...)` forwards every key as-is. The per-modality emptiness check that the old code applied is gone: it only added a modality key when `len(items) > 0`.

The same PR keeps that check in the sibling helper — [`rollouts.py:186-198`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/experience/rollouts.py#L186-L198) skips `None` and requires a non-empty container on the very same dict. So the two files now read one row differently: for `{"image": []}`, `row_has_formatter_consumed_media` says "no media" while this helper sends `{"image": []}` to vLLM.

Not reachable today — [`processors.py:789-792`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/data/processors.py#L789-L792) builds each list via `setdefault(...).append(...)`, so it cannot emit an empty or `None` value. Worth closing anyway since the shapes are now expressible: `{"image": None}` trips [`renderers/base.py:678-681`](https://github.com/vllm-project/vllm/blob/752a3a504485790a2e8491cacbb35c137339ad34/vllm/renderers/base.py#L678-L681), and `{"image": []}` passes validation but sends zero items against a prompt that still has its placeholder.

```suggestion
    def _get_multi_modal_data(index: int) -> dict[str, Any]:
        row = multi_modal_rows[index] if multi_modal_rows is not None else None
        if not row:
            return {}
        return {
            modality: value
            for modality, value in row.items()
            if value is not None
            and (not isinstance(value, (list, tuple)) or len(value) > 0)
        }
```

tests/unit/evals/test_eval.py
```diff
@@ -184,3 +184,82 @@ def test_eval_cons_k_multiple_groups():
     expected = 11 / 20
     assert isinstance(average_score, float)
     assert average_score == pytest.approx(expected, rel=1e-6)
+
+
+def test_run_env_eval_impl_builds_vllm_multi_modal_prompts(monkeypatch):
```
rohitrango: [`tests/unit/evals/test_eval.py:189`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/tests/unit/evals/test_eval.py#L189)

Good addition — this closes the gap where the whole prompt loop was monkeypatched out, and the list-shaped row 0 is what makes it reproduce the original failure.

Two of the three row shapes `vlm_hf_data_processor` emits are still uncovered:

1. **Placeholder rows** — `vllm_content=None` with non-empty media, produced at [`processors.py:815-820`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/data/processors.py#L815-L820) for `<image>`-style processors. This is the [`eval.py:348`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/evals/eval.py#L348) branch, and no test touches it. Needs `token_ids` on each message; asserts a `prompt_token_ids` dict.
2. **Truncated rows** — `vllm_content=None` with `vllm_multi_modal_data={}` and list-shaped content, from [`processors.py:796-801`](https://github.com/NVIDIA-NeMo/RL/blob/6a2a893cc573ff72b226b89cba08f86ba7baed9e/nemo_rl/data/processors.py#L796-L801). That is the crash in the other comment on this PR, so a case here would pin the fix.

A third and fourth row on the existing `BatchedDataDict` covers both.

nemo_rl/evals/eval.py
pstjohn: Fixed in 75bbc8d0c. The fallback now stringifies each message's content before joining, so truncated VLM rows with typed-part lists no longer raise TypeError. I also corrected the max_input_seq_length comment and added the truncated row to the eval prompt-loop regression test.

tests/unit/models/generation/test_vllm_utils.py
pstjohn: Agreed. Renamed this to test_vllm_utils_forwards_unknown_modality_keys and documented that the served vLLM model validates supported modalities. conditioning is only a synthetic key for testing NeMo-RL's forwarding behavior; the test makes no claim that vLLM accepts it. The PR description should use that same distinction.

nemo_rl/models/generation/vllm/utils.py
pstjohn: Fixed in 75bbc8d0c: Dynamo now checks VLLM_PROMPT_KEYS & data.keys() and rejects either column. The regression test covers each key independently and verifies that it raises before making an HTTP request.

nemo_rl/models/generation/vllm/utils.py
pstjohn: Applied the suggested filtering in 75bbc8d0c. None, empty lists, and empty tuples are omitted without mutating the input row. The parameterized test covers both formatted and token-ID prompt paths, including an empty image value alongside valid audio, and checks that an entirely empty media row falls back to token IDs.

tests/unit/evals/test_eval.py
pstjohn: Added both rows to the existing prompt-loop test in 75bbc8d0c. The placeholder row checks concatenated message token IDs and forwarded media; the truncated row checks the string fallback with list-shaped content. The test now exercises formatted multimodal, text-only, placeholder, and truncated inputs together.

pstjohn:
Pushed 75bbc8d0c to address the review feedback:

- Handle truncated VLM content safely in the eval fallback and correct the input-length config comment.
- Filter empty modality values while preserving valid payloads.
- Make Dynamo reject either vLLM prompt column before sending a request.
- Clarify the unknown-modality forwarding test and add coverage for placeholder/truncated eval rows, empty media values, and both Dynamo guard keys.

I also ran the CLEVR recipes with Qwen2.5-VL-3B-Instruct and Nemotron-3-Nano-Omni-30B-A3B on a single Lepton 8×H100 node using the exact source now pushed. Qwen SFT used its recipe's 2-GPU configuration; the other recipes used all 8 GPUs.

| Recipe | Steps | Result |
| --- | ---: | --- |
| Qwen VL GRPO | 10 | TMPE median **1.02359**, maximum **1.02459** |
| Nano Omni GRPO | 10 | TMPE median **1.03265**, maximum **1.03643** |
| Qwen VL SFT | 100 | Loss **7.79264 → 0.00426** |
| Nano Omni SFT | 100 | Loss **8.11099 → 0.00418** |

Every GRPO step stayed below **1.05 TMPE**, and both SFT runs passed the loss-convergence check. The scheduled GRPO validation batches also completed. These are short wiring/TMPE checks, not full GRPO reward-convergence runs.

Two runtime adjustments were needed: `NCCL_NVLS_ENABLE=0` for the node's multicast initialization error, and full optimizer CPU offload for Nano GRPO after the original recipe exhausted H100 memory allocating Adam state. Batch sizes, model parallelism, and generation lengths were preserved. Dataset/model paths were adjusted for the cluster; checkpointing and W&B were disabled, and TensorBoard was enabled.

I agree with the longer-term single-source-of-truth point. This consolidates the modality payload and shared key definitions, but the integration points still need explicit wiring. Centralizing how that payload moves through the pipeline would be a useful follow-up.

rohitrango (approved):
Latest changes address the outstanding review feedback. Approving.

terrykong (approved):

PR #3804, opened 2026-08-24
perf(refit): batch MXFP8 expert prequantization

seonjinn:
## Summary

Batch matching MoE expert weights before trainer-side MXFP8 prequantization. The path quantizes at most 16 experts per call, preserves the original refit wire order, and reuses temporary stacking storage only within one export pass.

Non-expert weights and expert groups that cannot be batched keep the existing per-tensor path. `fp8_param=true` is rejected because this path requires BF16 source weights.

This PR builds on #3294. Async NCCL Reshard receiver-side conversion is separate and is not changed here.

## Why

The existing path launches one quantization operation per expert tensor. Large MoE models repeat this operation hundreds of times per refit. Batching compatible experts reduces launch and synchronization overhead without changing the value/scale payload consumed by vLLM.

## Evidence

Current-head Qwen3-30B-A3B on 16 GB200 GPUs, synchronous colocated GRPO with CUDA Graph, 20 steps. Values are means over steps 2-19.

| Metric | Per expert | Batched | Change |
|---|---:|---:|---:|
| Transfer/update | 6.059 s | 5.283 s | -12.81% |
| Total refit | 9.737 s | 9.219 s | -5.32% |
| E2E step | 189.187 s | 188.483 s | -0.37% |
| E2E throughput | 2,195.58 tok/s/GPU | 2,202.91 tok/s/GPU | +0.33% |

[Per-expert run](https://wandb.ai/nvidia/nemo-rl-pr3804-current-head/runs/8csubf5g) and [batched run](https://wandb.ai/nvidia/nemo-rl-pr3804-current-head/runs/a5oqu3vg). Mean generation KL error was `0.003994` and `0.003978`, respectively. The direct transfer/update reduction is clear; the E2E difference is small enough to treat as near run-to-run noise.

Current head `0606ea87b011451f98c8f86a6c62e1c64706cdad` passed 83 relevant GB200 vLLM/refit tests covering quantization parity, Bridge export order, bounded batching, split payloads, scale shape, CUDA stream handoffs, and recipe scope. Ruff lint, format, and diff checks pass on all changed files.

Qwen3-235B peak-memory and end-to-end validation remain before this PR is ready for review.


tests/unit/models/policy/test_megatron_worker.py
```diff
@@ -468,6 +468,339 @@ def test_megatron_move_model_does_not_serialize_extra_state():
...
+    worker = object.__new__(MegatronPolicyWorkerImpl)
+    name = "model.weight"
+    worker._refit_prequant_names = {name}
+    tensor = torch.ones(2, 2, dtype=torch.float8_e4m3fn)
+
+    with pytest.raises(ValueError, match="BF16 trainer-exported weights"):
+        list(worker._maybe_prequantize_param(name, tensor))
+
+
+def test_iter_params_batches_expert_prequantization(monkeypatch):
```
seonjinn, line 765: **[Medium] The only worker-level test stubs out the real iterator, so batching can be silently disabled with no failing test.**

This test monkeypatches `fp8_train_utils.iter_mxfp8_prequantized_params` with a stub (effective because the call site does a deferred from-import on every call, `megatron_policy_worker.py:2256-2258`), and nothing else composes the worker with the real iterator. The direct unit tests do exercise `_EXPERT_WEIGHT_PATTERN` against realistic HF export names, so the regex isn't untested - but if Megatron-Bridge's export naming ever drifts, every selected tensor falls back to per-tensor `quantize_one`: correctness preserved, the PR's entire perf win silently gone, and no test turns red.

Suggested: one composition test that runs the real `iter_mxfp8_prequantized_params` behind a fake `export_hf_weights` yielding export-shaped expert names and asserts batched (stacked) quantize calls actually occur.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
```diff
@@ -13,6 +13,284 @@
...
+            continue
+
+        prefix = match.group("prefix")
+        projection = match.group("projection")
+        if current_prefix is not None and prefix != current_prefix:
+            yield from flush_pending()
+        current_prefix = prefix
+        group_key = (prefix, projection)
+        group = pending.setdefault(group_key, [])
+        source_stream = (
```
seonjinn, line 303: **[Low] Two ambient-stream assumptions worth a comment - both hold today, neither is checked.**

1. `torch.cuda.current_stream()` sampled at consumption time is assumed to be the stream `export_hf_weights` produced on. Verified true at the pinned Megatron-Bridge commit `8c46dc425` (no stream primitives anywhere in `model_bridge.py`/`param_mapping.py`, synchronous collectives, ambient-stream export loop) - but a future Bridge bump that introduces a side/copy stream would turn every `wait_stream` in this file into a silent race (intermittent bad weights, not a crash).
2. The scratch cache key uses the raw `cudaStream_t` pointer (`int(stack_stream.cuda_stream)` below). If a `torch.cuda.Stream` were created and destroyed within one export pass, CUDA can reissue the pointer and two different streams would alias one scratch entry. No current call site does this, but the failure would be silent corruption rather than an error.

A short comment documenting both ("producer must be the ambient stream; streams must outlive the export pass") makes the contract auditable at the next submodule bump.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
```diff
@@ -13,6 +13,284 @@
...
+        x_q, x_scales = _mxfp8_e4m3_quantize_torch(x)
+    x_scales = torch.squeeze(x_scales, dim=-1)
+    # Match the receiver path's zero-scale clamp: an E8M0 byte of 0 (2^-127)
+    # destabilizes the TRTLLM kernels, and pre-quantized tensors skip the
+    # receiver-side quantize branch where the clamp normally runs.
+    x_scales = torch.where(x_scales == 0, torch.ones_like(x_scales), x_scales)
+    return x_q, x_scales
+
+
+def iter_mxfp8_prequantized_params(
```
seonjinn, line 106: **[Low] `scratch_cache` is per-export-pass by design in production - say so, or the next reviewer re-flags it.**

The only production call site (`megatron_policy_worker.py:2260-2263`) passes no `scratch_cache`, so cross-refit reuse never happens - and that's intentional: `34da20e46` deliberately removed the worker-level cache so the ~192 MiB stacking buffer doesn't stay resident through training. What remains meaningful is in-pass reuse (one buffer shared across all layers/projections of a single export). The parameter is now test-injection-only API, and `test_batched_expert_prequantization_preserves_wire_entries_and_reuses_scratch` still asserts cross-call reuse as if it were a production behavior. A one-line docstring note (or reframing that assertion as a library-capability check) closes the trap. Do not wire a persistent cache back in - that would undo the memory-lifetime fix.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
```diff
@@ -13,6 +13,284 @@
...
+            scale = scale.view(len(chunk), *scale_shape)
+            entries = (
+                entry
+                for index, (_expert_id, name, _tensor, _stream) in enumerate(chunk)
+                for entry in (
+                    (name, value[index]),
+                    (name + "_scale_from_checkpoint", scale[index]),
+                )
+            )
+            yield from yield_on_current_stream(entries, producer_stream)
```
seonjinn: **[Low, latent] Yielded per-expert tensors are views pinning the whole <=16-expert batch.**

`value[index]` / `scale[index]` are views into the single stacked quantize output, so any downstream consumer that retains one expert's tensor keeps the entire batch's value and scale allocations alive. The current receiver copies into vLLM params promptly, so nothing leaks today - but a persistent-IPC or caching consumer would silently hold up to 16x the expected memory per retained entry. A one-line comment is enough; `.contiguous()`/clone per entry would defeat the batching win and is not recommended.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -2154,9 +2249,21 @@ def _iter_params_with_optional_kv_scales(
...
+                iter_mxfp8_prequantized_params,
+            )
+
+            yield from iter_mxfp8_prequantized_params(
+                base_iter,
+                self._refit_prequant_names,
+            )
+        else:
+            for name, tensor in base_iter:
+                yield from self._maybe_prequantize_param(name, tensor)
```
seonjinn, line 2689: **[Info] `_maybe_prequantize_param`'s quantization branch is now dead in production.**

This `else` branch only runs when `self._refit_prequant_names` is empty, and `_maybe_prequantize_param` passes through whenever the name isn't in that set - always true here. All real quantization now flows through `iter_mxfp8_prequantized_params`; the old helper's quantize branch is reachable only from its direct unit tests. Worth a comment (or folding the helper away) so a future reader doesn't assume it's still on the hot path.

seonjinn (commented):
## Self-review at head `1d1970194`

The GitHub diff shows 78 files because the branch stacks on #3294 (`88721ced8` is an ancestor of this head); the actual review surface is **7 commits / 4 files (+574/-2)**: the batched expert prequantization in `fp8_train_utils.py`, its wiring in `megatron_policy_worker.py`, and two test files. Everything below is scoped to that delta - #3294's findings (including the blocking `prepare_refit_info` signature drift and the per-batch scale-check straddle) apply to the base layer and are already posted [there](https://github.com/NVIDIA-NeMo/RL/pull/3294#pullrequestreview-5013003249).

Process: three domain reviewers (numerical correctness, stream/memory lifetime, tests), then an adversarial verification pass that re-read every citation and fetched Megatron-Bridge at the pinned submodule commit: **8 confirmed, 5 downgraded, 2 refuted, 1 new**. No Critical or High correctness bug survived verification - the core batching math, grouping, fallback routing, wire pairing, and all four stream-sync fix commits check out.

### Gates for undrafting

1. **Perf + peak-memory rerun at current head.** The body's prototype table predates the four fix commits and is already labeled as pending rerun - just noting the memory side should be measured too: with interleaved per-expert export order, `pending` extends the lifetime of up to ~46 projection tensors (~540 MiB for 235B-shaped experts) plus one reused ~192 MiB stacking scratch per `(device, dtype, stream)` key. Much of that is memory Megatron-Bridge's grouped export already materializes, so the true delta needs a `torch.cuda.max_memory_allocated()` before/after on a real MoE refit rather than arithmetic.
2. **One worker-composition test** (inline below): the only worker-level test stubs out the real iterator, so an export-naming drift would silently disable batching (correctness preserved via per-tensor fallback, but the PR's entire perf value gone with no failing test).

### Verified clean (adversarial pass, own citations)

- **Batched vs per-tensor parity**: MXFP8 quantization is row-independent, so the flatten-stack-quantize-unflatten path is mathematically identical to per-tensor; the CPU parity test does a genuine bitwise `torch.equal` against the real per-tensor quantizer (not a tautology) and covers both the multi-block and squeezed single-block scale shapes. E8M0 zero-scale clamp and the `squeeze(dim=-1)` behavior are inherited verbatim through the shared `quantize_fn`, not reimplemented.
- **Stream handshake complete at every producer-consumer edge** (pending-input to stack, stack to quantize output, per-tensor fallback), each with a CUDA-gated regression test targeting exactly that edge. The load-bearing assumption - that `export_hf_weights` produces on the ambient stream - was verified against Megatron-Bridge at submodule pin `8c46dc425`: zero stream primitives in `model_bridge.py`/`param_mapping.py`, synchronous collectives, `cpu=False`. Holds today; worth a comment for future Bridge bumps (inline).
- **Grouping/fallback**: layer+projection-scoped grouping with flush-on-prefix-change; shape/dtype/device/layout mismatch, lone remainders, duplicate expert ids, fused `w13`-style names, and `shared_experts` all route to the per-tensor or passthrough path correctly. Chunk arithmetic has no off-by-one; `max_experts_per_batch<=0` raises, `==1` degenerates to exact old-path behavior.
- The scratch-release fix (`34da20e46`) is the right call: the buffer now lives only for the export pass, and `record_stream` bookkeeping makes an explicit sync unnecessary.

### Refuted by the adversarial pass (so nobody re-flags them)

- "Shape-mismatch fallback never actually triggered in tests" - false; the `(3,64)` parametrized case drives a real mismatch through the fallback with value assertions.
- "The grad-enabled fixture wouldn't catch a removed `no_grad` guard" - false; the `torch.stack(..., out=)` in the test path is real production code and would raise `RuntimeError` without the guard, which is exactly what commit `09591f0b2` fixed.

### Notes

- Race-window tests use fixed `torch.cuda._sleep(5_000_000)` cycles; the failure mode on faster future GPUs is a spurious pass (missed detection), not flakiness - fine for CI, worth a comment.
- Test niceties, all Low: a zeroed row through the CPU parity test (clamp through the batched reshape), a dtype-mismatch fallback case, a 16/17-expert default-boundary case.

seonjinn:
Self-review completed on head `cfbe773c70a61f55835e6ed15339c61e3ae844ed`.

Reviewed the six-file change on top of #3294, including numerical behavior, refit wire order, CUDA stream handoffs, temporary-buffer lifetime, configuration guards, and fallback paths. No blocking correctness, compatibility, security, or deadlock issue was found.

Key checks:

- The batched iterator preserves the exact Bridge export order. This is required because the collective sender transfers tensors without parameter names.
- Each quantization call is bounded to 16 compatible experts; incompatible or non-expert tensors keep the per-tensor path.
- Split value/scale batches and singleton scale dimensions retain the existing vLLM payload contract.
- Active `fp8_param=true` is rejected because trainer-side prequantization requires BF16 source weights. A stale `fp8_param=true` value is ignored when FP8 is disabled.
- Producer/consumer CUDA stream edges and temporary scratch lifetime have direct regression coverage.

Validation on GB200:

- `48 passed, 29 deselected` in the relevant vLLM/MXFP8 unit-test selection.
- The focused source-order, bounded-batch, scale-parity, and stream-handoff selection passed `9/9`.
- Ruff lint, format, and `git diff --check` passed for all changed files.
- All commits are signed off; GitHub DCO, title, copyright, secrets, and submodule checks pass.

The current-head Qwen3-30B-A3B 20-step performance A/B and Qwen3-235B peak-memory validation are still pending, so the PR remains a draft.

tests/unit/models/policy/test_megatron_worker.py
seonjinn: Covered by `test_iter_params_preserves_bridge_expert_wire_order`. It uses the real batching iterator behind Bridge-shaped expert export names and asserts that one stacked quantization call handles both experts while preserving wire order.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
seonjinn: Documented on the current branch. Bridge must hand off any private producer stream before yielding, and stream objects must outlive the export pass because the raw stream handle is part of the scratch key.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
seonjinn: Documented on the current branch. The default cache is scoped to one export pass and is released before training resumes; cross-call injection remains only a library-level test capability.

nemo_rl/models/generation/vllm/quantization/fp8_train_utils.py
seonjinn: Documented on the current branch. Per-expert outputs intentionally remain views to avoid copies, and the refit consumer contract requires copying them before advancing beyond the export batch.

nemo_rl/models/policy/workers/megatron_policy_worker.py
seonjinn: Documented on the helper. Normal trainer export uses `iter_mxfp8_prequantized_params`; `_maybe_prequantize_param` is the single-tensor fallback.

PR #3809, opened 2026-08-25
feat(workers): Add config-driven worker extension FQNs

pstjohn:
This change is to make it easier to plug in custom workers and policies.

`Policy.__init__` already accepts a `worker_extension_cls_fqn`, but standard policy configs cannot set it. The vLLM generation path has no equivalent extension hook. A caller that needs custom policy and vLLM workers therefore has to wrap or replace Policy and monkeypatch the vLLM worker resolver.

This PR adds one optional `worker_extension_cls_fqn` field to `PolicyConfig` and `VllmConfig.` 

`Policy` reads the field directly from its config, so normal construction paths honor it without algorithm-specific forwarding. The existing constructor argument remains supported. 

If config and the constructor provide different FQNs, construction fails rather than choosing one silently. `VllmGeneration` applies its configured FQN after the existing worker resolver for both synchronous and asynchronous engines. 

On both paths, `worker_extension_cls_fqn` is mutually exclusive with `quant_cfg`, because both settings replace the backend’s resolved worker. Invalid combinations fail before worker or placement-group allocation. The canonical GRPO exemplar and its reference config include the two optional entries, as required for the existing TypedDict config schema.

### DIFF
--- a/examples/configs/grpo_math_1B.yaml
+++ b/examples/configs/grpo_math_1B.yaml
@@ -393,7 +393,7 @@ policy:
     # null = topology default (IPC colocated, NCCL non-colocated).
     # Non-colocated vLLM also supports vllm_s3_sparse, vllm_zmq_sparse, and nixl.
     refit_transport: null
-    # vLLM backend only - sglang/trtllm/dynamo inherit this exemplar and ignore it.
+    # vLLM backend only - sglang/megatron/trtllm/dynamo require this to remain null.
     # Selects the NeMo-RL Ray actor class, NOT vLLM's own `worker_extension_cls`.
     worker_extension_cls_fqn: null # optional registered vLLM generation-worker subclass; mutually exclusive with quant_cfg
     # Keep real-quant export tensors on CPU by default. Set false only for
--- a/examples/configs/grpo_math_1B.yaml
+++ b/examples/configs/grpo_math_1B.yaml
@@ -393,7 +393,9 @@ policy:
     # null = topology default (IPC colocated, NCCL non-colocated).
     # Non-colocated vLLM also supports vllm_s3_sparse, vllm_zmq_sparse, and nixl.
     refit_transport: null
-    worker_extension_cls_fqn: null # optional registered vLLM-worker subclass; mutually exclusive with quant_cfg
+    # vLLM backend only - sglang/trtllm/dynamo inherit this exemplar and ignore it.
+    # Selects the NeMo-RL Ray actor class, NOT vLLM's own `worker_extension_cls`.
+    worker_extension_cls_fqn: null # optional registered vLLM generation-worker subclass; mutually exclusive with quant_cfg
     # Keep real-quant export tensors on CPU by default. Set false only for
     # colocated CUDA-IPC refit to avoid a GPU -> CPU -> GPU round trip.
     real_quant_export_cpu_offload: true
--- a/examples/configs/grpo_math_1B.yaml
+++ b/examples/configs/grpo_math_1B.yaml
@@ -146,6 +146,7 @@ policy:
   precision: "bfloat16"
   logprob_chunk_size: null
   offload_optimizer_for_logprob: false # Only useful for non-colocated generation since colocated generation will always offload optimizer to cuda before refit
+  worker_extension_cls_fqn: null # optional registered policy-worker subclass; mutually exclusive with quant_cfg
 
   dtensor_cfg:
     _v2: true
@@ -392,6 +393,7 @@ policy:
     # null = topology default (IPC colocated, NCCL non-colocated).
     # Non-colocated vLLM also supports vllm_s3_sparse, vllm_zmq_sparse, and nixl.
     refit_transport: null
+    worker_extension_cls_fqn: null # optional registered vLLM-worker subclass; mutually exclusive with quant_cfg
     # Keep real-quant export tensors on CPU by default. Set false only for
     # colocated CUDA-IPC refit to avoid a GPU -> CPU -> GPU round trip.
     real_quant_export_cpu_offload: true
--- a/nemo_rl/models/generation/vllm/config.py
+++ b/nemo_rl/models/generation/vllm/config.py
@@ -194,6 +194,12 @@ class VllmConfig(GenerationConfig):
     real_quant_export_cpu_offload: NotRequired[bool]
     real_quant_ignore: NotRequired[list[str]]
 
+    # FQN of a worker extension class to use instead of the resolved default
+    # generation worker. Must be a subclass of the resolved worker and cannot
+    # be combined with quant_cfg. Its runtime environment must already be in
+    # ACTOR_ENVIRONMENT_REGISTRY.
+    worker_extension_cls_fqn: NotRequired[str | None]
+
 
 def resolve_vllm_video_config(config: VllmConfig) -> VllmVideoConfig | None:
     """Validate and return the optional vLLM video sampling contract."""
--- a/nemo_rl/models/generation/vllm/vllm_generation.py
+++ b/nemo_rl/models/generation/vllm/vllm_generation.py
@@ -30,6 +30,7 @@ from ray.util.placement_group import PlacementGroup
 
 from nemo_rl.distributed.batched_data_dict import BatchedDataDict, SlicedDataDict
 from nemo_rl.distributed.named_sharding import NamedSharding
+from nemo_rl.distributed.ray_actor_environment_registry import get_actor_python_env
 from nemo_rl.distributed.virtual_cluster import NVLINK_DOMAIN_UNKNOWN, RayVirtualCluster
 from nemo_rl.distributed.worker_groups import RayWorkerBuilder, RayWorkerGroup
 from nemo_rl.models.generation.fleet_health import (
@@ -223,6 +224,10 @@ class VllmGeneration(GenerationInterface):
                 "a custom generation worker cannot be combined with ModelOpt "
                 "quantization"
             )
+        if extension_fqn is not None:
+            # Validate registration before allocating workers or placement groups.
+            get_actor_python_env(extension_fqn)
+
         self.sharding_annotations = NamedSharding(
             layout=np.arange(cluster.world_size()).reshape(
                 self.dp_size, self.pp_size, self.tp_size
--- a/nemo_rl/models/generation/vllm/vllm_generation.py
+++ b/nemo_rl/models/generation/vllm/vllm_generation.py
@@ -216,6 +216,13 @@ class VllmGeneration(GenerationInterface):
 
         assert_reload_refit_config_supported(self.cfg)
 
+        extension_fqn = self.cfg.get("worker_extension_cls_fqn")
+        if extension_fqn is not None and self.cfg.get("quant_cfg") is not None:
+            raise ValueError(
+                "worker_extension_cls_fqn and quant_cfg are mutually exclusive: "
+                "a custom generation worker cannot be combined with ModelOpt "
+                "quantization"
+            )
         self.sharding_annotations = NamedSharding(
             layout=np.arange(cluster.world_size()).reshape(
                 self.dp_size, self.pp_size, self.tp_size
@@ -247,6 +254,8 @@ class VllmGeneration(GenerationInterface):
                 "nemo_rl.models.generation.vllm.vllm_worker.VllmGenerationWorker"
             )
         worker_cls = resolve_generation_worker_cls(worker_cls, self.cfg)
+        if extension_fqn is not None:
+            worker_cls = extension_fqn
         if self.cfg["vllm_cfg"]["async_engine"]:
             worker_builder = RayWorkerBuilder(
                 worker_cls, config, defer_model_load=defer_model_load
--- a/nemo_rl/models/policy/__init__.py
+++ b/nemo_rl/models/policy/__init__.py
@@ -634,3 +634,9 @@ class PolicyConfig(TypedDict):
     disable_modelopt_layer_spec: NotRequired[bool]
 
     is_vlm: NotRequired[bool]
+
+    # FQN of a worker extension class to use instead of the resolved default
+    # policy worker. Must be a subclass of the resolved worker and cannot be
+    # combined with quant_cfg. Its runtime environment must already be in
+    # ACTOR_ENVIRONMENT_REGISTRY.
+    worker_extension_cls_fqn: NotRequired[str | None]
--- a/nemo_rl/models/policy/lm_policy.py
+++ b/nemo_rl/models/policy/lm_policy.py
@@ -105,6 +105,27 @@ class Policy(ColocatablePolicyInterface, GenerationInterface):
         reserved_http_server_port: Optional[int] = None,
     ):
         self.debug_payload_metrics = False
+        configured_extension_fqn = config.get("worker_extension_cls_fqn")
+        if (
+            configured_extension_fqn is not None
+            and worker_extension_cls_fqn is not None
+            and configured_extension_fqn != worker_extension_cls_fqn
+        ):
+            raise ValueError(
+                "worker_extension_cls_fqn was set to different values in the "
+                "policy config and Policy constructor"
+            )
+        extension_fqn = (
+            configured_extension_fqn
+            if configured_extension_fqn is not None
+            else worker_extension_cls_fqn
+        )
+        if extension_fqn is not None and config.get("quant_cfg") is not None:
+            raise ValueError(
+                "worker_extension_cls_fqn and quant_cfg are mutually exclusive: "
+                "a custom policy worker cannot be combined with ModelOpt quantization"
+            )
+
         if weights_path:
             weights_path = os.path.abspath(weights_path)
         if optimizer_path:
@@ -237,11 +258,11 @@ class Policy(ColocatablePolicyInterface, GenerationInterface):
             env_vars = config["dtensor_cfg"].get("env_vars", {})
 
         # If a worker extension class is provided, use it instead of the default worker builder class
-        if worker_extension_cls_fqn is not None:
+        if extension_fqn is not None:
             print(
-                f"Using worker extension class: {worker_extension_cls_fqn}, please make sure it is a subclass of {worker_builder_cls_fqn}."
+                f"Using worker extension class: {extension_fqn}, please make sure it is a subclass of {worker_builder_cls_fqn}."
             )
-            worker_builder_cls_fqn = worker_extension_cls_fqn
+            worker_builder_cls_fqn = extension_fqn
 
         # Validate world_size compatibility with parallelism configuration
         model_parallel_size = pp_size * cp_size * tp_size
--- a/nemo_rl/models/policy/lm_policy.py
+++ b/nemo_rl/models/policy/lm_policy.py
@@ -120,7 +120,12 @@ class Policy(ColocatablePolicyInterface, GenerationInterface):
             if configured_extension_fqn is not None
             else worker_extension_cls_fqn
         )
-        if extension_fqn is not None and config.get("quant_cfg") is not None:
+        # Only the config-supplied FQN is mutually exclusive with quant_cfg: a config
+        # author cannot know which worker quant_cfg resolves to, so silently replacing
+        # it would be a footgun. The constructor argument is exempt because callers
+        # passing it already see the resolved class name and are expected to subclass
+        # it, which has always been supported alongside quantization.
+        if configured_extension_fqn is not None and config.get("quant_cfg") is not None:
             raise ValueError(
                 "worker_extension_cls_fqn and quant_cfg are mutually exclusive: "
                 "a custom policy worker cannot be combined with ModelOpt quantization"
--- a/nemo_rl/models/policy/lm_policy.py
+++ b/nemo_rl/models/policy/lm_policy.py
@@ -31,6 +31,7 @@ from nemo_rl.distributed.batched_data_dict import (
     SlicedDataDict,
 )
 from nemo_rl.distributed.named_sharding import NamedSharding
+from nemo_rl.distributed.ray_actor_environment_registry import get_actor_python_env
 from nemo_rl.distributed.virtual_cluster import RayVirtualCluster
 from nemo_rl.distributed.worker_groups import RayWorkerBuilder, RayWorkerGroup
 from nemo_rl.models.generation.interfaces import (
@@ -130,6 +131,9 @@ class Policy(ColocatablePolicyInterface, GenerationInterface):
                 "worker_extension_cls_fqn and quant_cfg are mutually exclusive: "
                 "a custom policy worker cannot be combined with ModelOpt quantization"
             )
+        if extension_fqn is not None:
+            # Validate registration before allocating workers or placement groups.
+            get_actor_python_env(extension_fqn)
 
         if weights_path:
             weights_path = os.path.abspath(weights_path)
--- a/tests/unit/models/policy/test_policy_validation.py
+++ b/tests/unit/models/policy/test_policy_validation.py
@@ -207,15 +207,20 @@ def test_policy_flops_tracker_uses_hf_config_overrides() -> None:
     ],
 )
 def test_policy_selects_worker_extension_from_config_or_constructor(
-    tiny_llama_model_path,
-    configured_extension_fqn,
-    explicit_extension_fqn,
+    configured_extension_fqn: str | None,
+    explicit_extension_fqn: str | None,
 ) -> None:
-    config = create_dtensor_config(tiny_llama_model_path, tp=1)
+    config = create_dtensor_config("test-model", tp=1)
     if configured_extension_fqn is not None:
         config["worker_extension_cls_fqn"] = configured_extension_fqn
     with (
+        patch("nemo_rl.models.policy.lm_policy.get_hf_config"),
+        patch("nemo_rl.models.policy.lm_policy.FLOPTracker"),
         patch("nemo_rl.models.policy.lm_policy.RayQueue"),
+        patch.dict(
+            "nemo_rl.distributed.ray_actor_environment_registry.ACTOR_ENVIRONMENT_REGISTRY",
+            {"tests.extensions.CustomPolicyWorker": "python"},
+        ),
         patch("nemo_rl.models.policy.lm_policy.RayWorkerBuilder") as worker_builder,
         patch("nemo_rl.models.policy.lm_policy.RayWorkerGroup"),
     ):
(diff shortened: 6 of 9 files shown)


tests/unit/models/policy/test_policy_validation.py
```diff
@@ -171,6 +171,71 @@ def create_megatron_config(
     }
 
 
+@pytest.mark.parametrize(
+    ("configured_extension_fqn", "explicit_extension_fqn"),
+    [
+        ("tests.extensions.CustomPolicyWorker", None),
+        (None, "tests.extensions.CustomPolicyWorker"),
```
pstjohn, line 178: [`tests/unit/models/policy/test_policy_validation.py:178`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/tests/unit/models/policy/test_policy_validation.py#L178)

**1 action item.**

Neither parametrize row sets *both* values, so the `and configured_extension_fqn != worker_extension_cls_fqn` clause at [`lm_policy.py:111`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/nemo_rl/models/policy/lm_policy.py#L111) is dead code with respect to this suite. PR-introduced.

Mutation-verified independently by two reviewers: delete that clause — which turns the supported "both set, identical" configuration into a hard error — and all four new tests still pass. Adding the equal-values row kills the mutant (`1 failed, 3 passed`).

**Action:** add a third parametrize row.

```suggestion
        (None, "tests.extensions.CustomPolicyWorker"),
        ("tests.extensions.CustomPolicyWorker", "tests.extensions.CustomPolicyWorker"),
```

nemo_rl/models/policy/lm_policy.py
```diff
@@ -104,6 +104,27 @@ def __init__(
...
+            raise ValueError(
+                "worker_extension_cls_fqn was set to different values in the "
+                "policy config and Policy constructor"
+            )
+        extension_fqn = (
+            configured_extension_fqn
+            if configured_extension_fqn is not None
+            else worker_extension_cls_fqn
+        )
+        if extension_fqn is not None and config.get("quant_cfg") is not None:
```
pstjohn: [`nemo_rl/models/policy/lm_policy.py:122`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/nemo_rl/models/policy/lm_policy.py#L122)

**1 action item.**

**TL;DR — this narrows the pre-existing constructor argument: `Policy(config=<config with quant_cfg>, worker_extension_cls_fqn=...)` used to work and now hard-raises, so "the existing constructor argument remains supported" isn't quite accurate.**

PR-introduced narrowing of pre-existing behavior (the config path itself is pure addition — this is only about the ctor arg).

Pre-PR ([`10dc8e02` lines 195-200](https://github.com/NVIDIA-NeMo/RL/blob/10dc8e02/nemo_rl/models/policy/lm_policy.py#L195)) the override was unconditional and applied *after* the resolver:

```python
if worker_extension_cls_fqn is not None:
    print(f"Using worker extension class: {worker_extension_cls_fqn}, please make sure it is a subclass of {worker_builder_cls_fqn}.")
    worker_builder_cls_fqn = worker_extension_cls_fqn
```

When `quant_cfg` was set, `worker_builder_cls_fqn` at that point was already the *quant* worker ([`policy/utils.py:120-127`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/nemo_rl/models/policy/utils.py#L120)) — so that print was telling the user to subclass `MegatronQuantPolicyWorker`. "Extension subclasses the resolved quant worker" was a contemplated, working combination, and it now raises with no escape hatch.

The stated rationale (both settings replace the resolved worker) holds cleanly for the **config** path, where the user can't know which worker got resolved. It's weaker for the **ctor** path, where the prior contract was explicitly subclass-the-resolved-class.

Nothing in-tree breaks — the only ctor-arg caller, [`research/template_project/single_update.py:105`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/research/template_project/single_update.py#L105), sets no `quant_cfg` — so this is a speculative out-of-tree break, not an observed one.

**Action:** either scope the exclusion to the config-supplied FQN (one option, of several):

```suggestion
        if configured_extension_fqn is not None and config.get("quant_cfg") is not None:
```

or keep the stricter rule and say so explicitly in the PR description and the field comment, since it's a deliberate narrowing rather than a pure addition.

examples/configs/grpo_math_1B.yaml
```diff
@@ -380,6 +381,7 @@ policy:
     # null = topology default (IPC colocated, NCCL non-colocated).
     # Non-colocated vLLM also supports vllm_s3_sparse, vllm_zmq_sparse, and nixl.
     refit_transport: null
+    worker_extension_cls_fqn: null # optional registered vLLM-worker subclass; mutually exclusive with quant_cfg
```
pstjohn: [`examples/configs/grpo_math_1B.yaml:384`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/examples/configs/grpo_math_1B.yaml#L384)

**1 action item.**

13 exemplars inherit this file via `defaults:`, including `grpo_math_1B_sglang.yaml`, `grpo_math_1B_trtllm.yaml` and `grpo_math_1B_dynamo.yaml`. All three resolve with this key present, and all three ignore it: sglang hardcodes its worker FQN at [`sglang_generation.py:175`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/nemo_rl/models/generation/sglang/sglang_generation.py#L175), trtllm at [`trtllm_generation.py:128`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/nemo_rl/models/generation/trtllm/trtllm_generation.py#L128), and Dynamo has no `RayWorkerBuilder` path at all. Only [`vllm_generation.py:175`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/nemo_rl/models/generation/vllm/vllm_generation.py#L175) reads it.

Not asking for a validator — `refit_transport`, `quant_cfg`, `real_quant` and `real_quant_export_cpu_offload` all already behave exactly this way in these backends, so that's a pre-existing gap rather than something this PR created. Just the comment.

Second nit on the same line: "vLLM-worker subclass" reads like vLLM's own `worker_extension_cls` engine kwarg ([`vllm_worker.py:586`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/nemo_rl/models/generation/vllm/vllm_worker.py#L586), documented in [`checkpoint-engines.md:317`](https://github.com/NVIDIA-NeMo/RL/blob/1be08dda5f5d9caa35d58982bd8acb9bccfaf8e6/docs/design-docs/checkpoint-engines.md#L317)) — that one attaches to the vLLM engine, this one replaces the NeMo-RL Ray actor class.

**Action:**

```suggestion
    # vLLM backend only - sglang/trtllm/dynamo inherit this exemplar and ignore it.
    # Selects the NeMo-RL Ray actor class, NOT vLLM's own `worker_extension_cls`.
    worker_extension_cls_fqn: null # optional registered vLLM generation-worker subclass; mutually exclusive with quant_cfg
```

pstjohn (commented):
Reviewed by a team of 5 agents (RL expert, bug finder, test agent, design reviewer, devil's advocate). Nice, tightly-scoped change — and failing loudly on conflicting config-vs-constructor values instead of silently picking one is the right call.

**Verification run locally:**
- Linters clean: `ruff`, `ruff-format`, `taplo`, whitespace, and `pyrefly check` (0 errors).
- All 4 new tests pass on 8xH100. They were mutation-tested and kill 4 real mutants (dropping the vLLM override; moving the quant check after PG init; the async/sync kwargs split) — these are not tautologies. `cluster._init_placement_groups.assert_not_called()` in particular has teeth: moving the check after the PG call makes the test fail.
- The YAML nesting was checked hard by three agents independently and is **correct** — `VllmGeneration.self.cfg` is the `policy.generation` dict, so the key is read where it's set. Not a silent no-op.

**3 inline comments**, all suggestions, none blocking.

Two lower-confidence items I'm noting here rather than inline, both arguably tracking issues rather than work for this PR:
- `TeacherWorkerGroup` deep-copies the student policy config and hardcodes `MegatronPolicyWorker`, so `policy.worker_extension_cls_fqn` is silently dropped for distillation teachers — the same hole it already warns-and-clears for `quant_cfg`. Debatable whether that's even wrong (a frozen teacher on the stock worker seems fine).
- Nothing in the suite pins an exemplar YAML key to the dict level its consumer reads. Harmless here since the value is `null` and both consumers use `.get()`, but see below.

**Unrelated repo-level find, worth its own issue:** `_collect_mismatched_keys` in `tests/unit/test_config_v2.py:101` is one-directional. Deleting a key from an exemplar while leaving the reference config intact still passes all 14 tests, so exemplar-side drift is invisible in both directions.

_Generated by Claude Code_

tests/unit/models/policy/test_policy_validation.py
pstjohn: fixed in ab2f594ed67db597562226e83997552ffbe7c37a

nemo_rl/models/policy/lm_policy.py
pstjohn: fixed in ab2f594ed67db597562226e83997552ffbe7c37a

examples/configs/grpo_math_1B.yaml
pstjohn: fixed in ab2f594ed67db597562226e83997552ffbe7c37a

nemo_rl/models/generation/vllm/vllm_generation.py
```diff
@@ -172,6 +172,13 @@ def __init__(
             f"Please update your configuration to include all required VLLM parameters."
         )
 
+        extension_fqn = self.cfg.get("worker_extension_cls_fqn")
+        if extension_fqn is not None and self.cfg.get("quant_cfg") is not None:
```
yuki-97: nit: this guard fails early on `quant_cfg`, but the likelier YAML-author mistake — an FQN that isn't in `ACTOR_ENVIRONMENT_REGISTRY` — isn't checked here.

It first surfaces at [`worker_groups.py:458`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/distributed/worker_groups.py#L458) inside `RayWorkerGroup`, after [`cluster._init_placement_groups(...)`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/models/generation/vllm/vllm_generation.py#L200) has already reserved the bundles.

**Suggest fix:** call `get_actor_python_env(extension_fqn)` next to this check — it's already imported transitively via `worker_groups` and is a plain dict lookup. (Same at [`lm_policy.py:127`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/models/policy/lm_policy.py#L127), though nothing expensive is allocated there.)

examples/configs/grpo_math_1B.yaml
```diff
@@ -380,6 +381,9 @@ policy:
     # null = topology default (IPC colocated, NCCL non-colocated).
     # Non-colocated vLLM also supports vllm_s3_sparse, vllm_zmq_sparse, and nixl.
     refit_transport: null
+    # vLLM backend only - sglang/trtllm/dynamo inherit this exemplar and ignore it.
```
yuki-97: nit: `megatron` is the fifth generation backend, and 6 shipped recipes set it while inheriting this exemplar (e.g. [`grpo-llama3.2-1b-instruct-1n8g-megatron_generation.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/examples/configs/recipes/llm/grpo-llama3.2-1b-instruct-1n8g-megatron_generation.yaml#L1)).

It ignores this key too — [`MegatronGeneration`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/models/generation/megatron/megatron_generation.py#L154) builds its workers through `Policy`, which reads the policy-level `worker_extension_cls_fqn` you added at line 140.

```suggestion
    # vLLM backend only - sglang/megatron/trtllm/dynamo inherit this exemplar and ignore it.
```

nemo_rl/models/generation/vllm/vllm_generation.py
```diff
@@ -172,6 +172,13 @@ def __init__(
...
+            raise ValueError(
+                "worker_extension_cls_fqn and quant_cfg are mutually exclusive: "
+                "a custom generation worker cannot be combined with ModelOpt "
+                "quantization"
+            )
```
yuki-97: nit: this guard fails early on `quant_cfg`, but the likelier mistake with a new string config key — an FQN that isn't in `ACTOR_ENVIRONMENT_REGISTRY` — isn't checked here.

It first surfaces at [`worker_groups.py:458`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/distributed/worker_groups.py#L458) inside `RayWorkerGroup`, after [`cluster._init_placement_groups(...)`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/models/generation/vllm/vllm_generation.py#L200) has already reserved the bundles. `get_actor_python_env` is a plain driver-side dict lookup, so checking here is free.

**Suggest fix:**

1. Import `get_actor_python_env` from `nemo_rl.distributed.ray_actor_environment_registry` — `worker_groups` loads the module, but the name isn't bound in this file.
2. Apply the block below.

```suggestion
        extension_fqn = self.cfg.get("worker_extension_cls_fqn")
        if extension_fqn is not None and self.cfg.get("quant_cfg") is not None:
            raise ValueError(
                "worker_extension_cls_fqn and quant_cfg are mutually exclusive: "
                "a custom generation worker cannot be combined with ModelOpt "
                "quantization"
            )
        if extension_fqn is not None:
            # Fail before placement groups are reserved if the FQN was never registered.
            get_actor_python_env(extension_fqn)
```

nemo_rl/models/policy/lm_policy.py
```diff
@@ -104,6 +104,32 @@ def __init__(
...
+        # Only the config-supplied FQN is mutually exclusive with quant_cfg: a config
+        # author cannot know which worker quant_cfg resolves to, so silently replacing
+        # it would be a footgun. The constructor argument is exempt because callers
+        # passing it already see the resolved class name and are expected to subclass
+        # it, which has always been supported alongside quantization.
+        if configured_extension_fqn is not None and config.get("quant_cfg") is not None:
+            raise ValueError(
+                "worker_extension_cls_fqn and quant_cfg are mutually exclusive: "
+                "a custom policy worker cannot be combined with ModelOpt quantization"
+            )
```
yuki-97: nit: same fail-early gap on the policy side — nothing checks `extension_fqn` against `ACTOR_ENVIRONMENT_REGISTRY` before the worker group is built.

This costs more than it looks. In colocated GRPO the generation fleet is constructed before the policy ([`grpo.py:1185`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/algorithms/grpo.py#L1185), calls at [`:1386`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/algorithms/grpo.py#L1386) / [`:1389`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/algorithms/grpo.py#L1389)), so a typo'd `policy.worker_extension_cls_fqn` only surfaces at [`worker_groups.py:458`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/distributed/worker_groups.py#L458) once vLLM is already fully up.

**Suggest fix:**

1. Import `get_actor_python_env` from `nemo_rl.distributed.ray_actor_environment_registry`.
2. Apply the block below — it checks the merged `extension_fqn`, so the constructor argument is covered too.

```suggestion
        if configured_extension_fqn is not None and config.get("quant_cfg") is not None:
            raise ValueError(
                "worker_extension_cls_fqn and quant_cfg are mutually exclusive: "
                "a custom policy worker cannot be combined with ModelOpt quantization"
            )
        if extension_fqn is not None:
            # Fail before the generation fleet is built if the FQN was never registered.
            get_actor_python_env(extension_fqn)
```

nemo_rl/models/generation/vllm/vllm_generation.py
yuki-97, line 226: nit: this guard fails early on `quant_cfg`, but the likelier mistake with a new string config key — an FQN that isn't in `ACTOR_ENVIRONMENT_REGISTRY` — isn't checked here.

It first surfaces at [`worker_groups.py:458`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/distributed/worker_groups.py#L458) inside `RayWorkerGroup`, after [`cluster._init_placement_groups(...)`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/models/generation/vllm/vllm_generation.py#L200) has already reserved the bundles. `get_actor_python_env` is a plain driver-side dict lookup, so checking here is free.

**Suggest fix:**

1. Import `get_actor_python_env` from `nemo_rl.distributed.ray_actor_environment_registry` — `worker_groups` loads the module, but the name isn't bound in this file.
2. Apply the block below.

```suggestion
            )
        if extension_fqn is not None:
            # Fail before placement groups are reserved if the FQN was never registered.
            get_actor_python_env(extension_fqn)
```

nemo_rl/models/policy/lm_policy.py
yuki-97, line 133: nit: same fail-early gap on the policy side — nothing checks `extension_fqn` against `ACTOR_ENVIRONMENT_REGISTRY` before the worker group is built.

This costs more than it looks. In colocated GRPO the generation fleet is constructed before the policy ([`grpo.py:1185`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/algorithms/grpo.py#L1185), calls at [`:1386`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/algorithms/grpo.py#L1386) / [`:1389`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/algorithms/grpo.py#L1389)), so a typo'd `policy.worker_extension_cls_fqn` only surfaces at [`worker_groups.py:458`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/distributed/worker_groups.py#L458) once vLLM is already fully up.

**Suggest fix:**

1. Import `get_actor_python_env` from `nemo_rl.distributed.ray_actor_environment_registry`.
2. Apply the block below — it checks the merged `extension_fqn`, so the constructor argument is covered too.

```suggestion
            )
        if extension_fqn is not None:
            # Fail before the generation fleet is built if the FQN was never registered.
            get_actor_python_env(extension_fqn)
```

nemo_rl/models/generation/vllm/config.py
```diff
@@ -185,6 +185,12 @@ class VllmConfig(GenerationConfig):
     real_quant_export_cpu_offload: NotRequired[bool]
     real_quant_ignore: NotRequired[list[str]]
 
+    # FQN of a worker extension class to use instead of the resolved default
+    # generation worker. Must be a subclass of the resolved worker and cannot
+    # be combined with quant_cfg. Its runtime environment must already be in
+    # ACTOR_ENVIRONMENT_REGISTRY.
+    worker_extension_cls_fqn: NotRequired[str | None]
```
yuki-97, line 201: Setting this on a non-vLLM backend is accepted and then silently dropped — the custom worker never runs, and the job trains to completion looking normal. Only [`vllm_generation.py:175`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/models/generation/vllm/vllm_generation.py#L175) reads it; sglang/trtllm/megatron/dynamo never look, and none of them rejects unknown generation keys.

I'd guard it in [`configure_generation_config`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/nemo_rl/models/generation/__init__.py#L27), which every entrypoint calls before any cluster exists (e.g. [`run_grpo.py:119`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/examples/run_grpo.py#L119)):

```python
if config["backend"] != "vllm" and config.get("worker_extension_cls_fqn") is not None:
    raise ValueError(
        "generation.worker_extension_cls_fqn is only supported by the vLLM backend, "
        f"got backend={config['backend']!r}. Use policy.worker_extension_cls_fqn to "
        "extend the training worker instead."
    )
```

nemo_rl/models/policy/__init__.py
```diff
@@ -606,3 +606,9 @@ class PolicyConfig(TypedDict):
     disable_modelopt_layer_spec: NotRequired[bool]
 
     is_vlm: NotRequired[bool]
+
+    # FQN of a worker extension class to use instead of the resolved default
+    # policy worker. Must be a subclass of the resolved worker and cannot be
+    # combined with quant_cfg. Its runtime environment must already be in
+    # ACTOR_ENVIRONMENT_REGISTRY.
+    worker_extension_cls_fqn: NotRequired[str | None]
```
yuki-97, line 642: **No action needed in this PR.** Filed #4032 to track unifying the three mechanisms that currently decide the worker class — this config field, `quant_cfg`'s resolver override maps, and the `Policy` constructor argument that [`single_update.py:105`](https://github.com/NVIDIA-NeMo/RL/blob/ab2f594ed67db597562226e83997552ffbe7c37a/research/template_project/single_update.py#L105) still uses — onto the config field this PR adds.

yuki-97 (commented):
**Scope:** the two new `worker_extension_cls_fqn` fields and their wiring in `lm_policy.py` / `vllm_generation.py`, plus the exemplar + reference YAMLs and the two new test modules.

**Checked:**

- Config placement — the generation-level key is read at the same level `quant_cfg` and `refit_transport` live at, and the reference config mirrors the exemplar, so `test_reference_configs_up_to_date` stays green.
- Guard ordering — the quant exclusion correctly precedes `cluster._init_placement_groups`; traced where each *other* invalid-FQN case actually surfaces, and how much is allocated by then.
- Backend reach — which of the five generation backends can see the key after `defaults:` inheritance, and which of them read it.
- Test mutation — dropped each new production line in turn and confirmed a new test fails; the default no-extension path is unchanged from `main`.
- Exemplar convention — compared against how other `NotRequired` `PolicyConfig` keys are spread across the five base configs.

**Remaining comments** are three fail-early / validation suggestions, one fix to a backend list in a YAML comment, and a pointer to #4032 for the longer-term consolidation (no action needed here).

**Before merge:** could you rebase and run CI? The branch is 69 commits behind `main`, and `nemo_rl/models/generation/__init__.py` has moved there since it branched — that is the file the `vllm/config.py` comment asks you to add a guard to, so it is worth rebasing before picking that one up. I have added `CI:Lfast`; all unit shards were skipped on ab2f594ed67db597562226e83997552ffbe7c37a, so the four new tests have not run in CI yet.

nemo_rl/models/policy/__init__.py
pstjohn: awesome sounds great 😅

examples/configs/grpo_math_1B.yaml
pstjohn: ok thanks, updated the comment to include Megatron and clarify that non-vLLM backends require this field to remain `null`, matching the new validation.

nemo_rl/models/generation/vllm/vllm_generation.py
pstjohn: sgtm, added `get_actor_python_env(extension_fqn)` before placement-group allocation. Added regression coverage for unregistered FQNs in both synchronous and asynchronous modes, asserting that neither placement groups nor workers are allocated.

nemo_rl/models/policy/lm_policy.py
pstjohn: Ack, added registry validation for the merged config/constructor FQN before policy allocation, with tests covering both sources. One clarification: this fails early within `Policy.__init__`, but cannot prevent generation allocation when the caller constructs the generation fleet first.

nemo_rl/models/generation/vllm/config.py
pstjohn: Added the guard in `configure_generation_config`. Non-null generation worker extensions now raise on sglang, megatron, trtllm, and dynamo, with guidance to use the policy-level field for training workers. Tests cover rejection on all four backends, acceptance of absent/null values, and acceptance of the override on vLLM.

examples/configs/grpo_math_1B.yaml
yuki-97: fixed in f934c26ddd8be0d1b7c30fb9106a879a0eda2f9f

nemo_rl/models/generation/vllm/vllm_generation.py
yuki-97: fixed in f934c26ddd8be0d1b7c30fb9106a879a0eda2f9f

nemo_rl/models/policy/lm_policy.py
yuki-97: fixed in f934c26ddd8be0d1b7c30fb9106a879a0eda2f9f

nemo_rl/models/generation/vllm/config.py
yuki-97: fixed in f934c26ddd8be0d1b7c30fb9106a879a0eda2f9f

yuki-97 (approved):

PR #3410, opened 2026-07-29
feat(ppo): support async ppo

bg51717:
# What does this PR do?

Based on https://github.com/NVIDIA-NeMo/RL/pull/3185.

Adds asynchronous PPO on top of non-colocated PPO, overlapping rollout generation on dedicated vLLM GPUs with policy and value-model training on the training cluster.

This is the asynchronous follow-up to #3262. The existing synchronous colocated and non-colocated PPO paths remain supported.

## Design

The main changes are:

- Extend the shared asynchronous rollout collector and replay buffer to support Async PPO while preserving the existing Async GRPO behavior.
- Generate trajectories continuously on the inference cluster while the policy and value models train on the separate training cluster.
- Recompute values and policy/reference logprobs when a trajectory is sampled, then run GAE and the configured PPO epochs against the current learner state.
- Track the policy version used for generation and bound off-policy staleness with `max_trajectory_age_steps`.
- Allow the collector to bank additional rollouts during critic warmup with `warmup_max_trajectory_age_steps`, while keeping the effective policy-update age within the normal window once policy training starts.
- Publish updated policy weights to vLLM after each policy-training step, with support for in-flight weight updates and explicit synchronization before checkpointing or shutdown.
- Save and restore the replay buffer, collector dataloader state, policy, and value model so asynchronous training can resume without losing completed trajectories.
- Regenerate incomplete rollout targets after restore by default instead of treating partially generated data as complete.
- Cycle the training dataloader until `max_num_steps` is reached so asynchronous progress is controlled by learner steps rather than dataloader exhaustion.
- Migrate `PPOConfig` from the legacy `TypedDict` representation to a Pydantic `BaseModel`, preserving existing defaults and extra-field compatibility while aligning PPO configuration access with GRPO.
- Fail fast for unsupported combinations, including colocated generation, non-vLLM generation, synchronous vLLM engines, dynamic sampling, reward scaling or shaping, multiple dataloaders, NeMo Gym, and FP8 KV-cache scale synchronization.
- Add unit coverage, AutoModel and Megatron functional checkpoint/resume tests, Async GRPO compatibility coverage, 40-step nightly recipes, and PPO documentation.

# Usage

Enable asynchronous PPO together with non-colocated asynchronous vLLM generation and importance-sampling correction:

```yaml
ppo:
  max_num_epochs: -1
  async_ppo:
    enabled: true
    max_trajectory_age_steps: 1
    warmup_max_trajectory_age_steps: null
    in_flight_weight_updates: true
[...]

PR #3582, opened 2026-08-11
feat(sc): add ready-first staleness sampler

haitian-nvidia:
# What does this PR do ?

Ports the staleness-admission policy from #2940 (AReaL) onto the current
SingleController sampler interface as a new built-in `ReadyFirstSampler`
(`name="ready_first"`).

- Admission gates dispatch to at most `max_staleness_versions` (eta) batches
  ahead of the trainer version, reusing the shared `_GatedSampler` dispatch
  counter; the controller keeps no duplicate staleness state.
- Selection is ready-first across weight versions (mixed-version batches)
  instead of draining one version at a time.
- The sampler never evicts: every admitted group stays selectable until it is
  trained on, so late stragglers remain usable and no rollout is discarded.
- `ReadyFirstSamplerConfig` joins the discriminated `SamplerConfig` union,
  `create_sampler()`, and `required_buffer_capacity_for_config()`
  (capacity = `num_prompts_per_step * (eta + 1)`).
- `validate_single_controller_config()` rejects `ready_first` unless
  `loss_fn.use_importance_sampling_correction=true` and
  `loss_fn.force_on_policy_ratio=false`, so `prev_logprobs` are real and used
  as the behavior-policy correction. The check fails before any setup factory
  allocates resources.

# Issues

Related: #2940 (original AReaL PR this admission policy is ported from).

# Usage

```yaml
loss_fn:
  use_importance_sampling_correction: true
  force_on_policy_ratio: false

async_rl:
  sampler:
    name: ready_first
    max_staleness_versions: 1   # eta: at most 1 batch ahead of the trainer
  max_buffered_rollouts: 128    # >= num_prompts_per_step * (eta + 1)
```

See the full runnable recipe at
`examples/configs/recipes/llm/grpo-llama3.1-8b-instruct-2n8g-ready-first-single-controller.yaml`.

# Before your PR is "Ready for review"

**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [x] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information

[...]

PR #3480, opened 2026-08-03
feat(sc): recover replay buffer from native TQ checkpoints

macandro96:
# What does this PR do ?

## Summary

Adds crash-consistent SingleController recovery for two rollout states:

1. **Completed, unconsumed groups**
   - Tensor payloads remain authoritative in the native TransferQueue checkpoint.
   - SingleController stores only a metadata replay index.
   - Recovery rebuilds the local replay buffer without copying tensors back into TQ.

2. **Unfinished prompt groups**
   - A lightweight ownership ledger records prompts that have left the dataloader but have not reached canonical TQ commit.
   - Recovery rehydrates those prompts from the dataset and redispatches them for whole-group regeneration.
   - This prevents dataloader advancement and sampler dispatch state from silently skipping prompts after restart.

This builds on the merged TQ v0.19 bump (#3423) and SingleController checkpoint lifecycle (#3429).

## Motivation

Previously, SC checkpoint recovery either serialized rollout tensors separately in `replay_buffer.pt` or restarted with an empty rollout buffer.

Serializing tensors in both the replay-buffer checkpoint and TQ:

- duplicates potentially large payloads;
- creates two possible sources of truth;
- requires re-putting tensors into TQ during recovery;
- cannot guarantee that the replay index and TQ snapshot describe the same rows.

This change makes the native TQ snapshot authoritative for tensors and keeps only the controller metadata required to resume consuming those rows.

## Checkpoint contents

```text
step_N/
├── policy/
│   ├── weights/
│   └── optimizer/
├── train_dataloader.pt
├── training_info.json
├── data_plane/                   # Native TQ checkpoint; owns tensors
├── replay_buffer_metadata.pt     # Completed-group metadata only
├── rollout_recovery.pt           # Unfinished prompt-group ownership
└── replacement_reserve.pt        # Optional spare prompt pool
```

`replay_buffer_metadata.pt` contains no rollout tensors or fields_data.

`rollout_recovery.pt` stores:
1. stable group and admission IDs;
2. dataset prompt reference and task name;
3. reserved/admitted phase;
4. expected generation count;
5. start policy weight version;
6. optional target training step;
7. batch-shortfall and sampler scheduling state.

## Consistency model
TQ mutations, sampler admission commits, rollout-ledger mutations, and checkpoint capture participate in the same data-plane checkpoint barrier. 

The barrier ensures a checkpoint observes transitions atomically, including:
1. dataloader advancement together with prompt ownership;
[...]

PR #3813, opened 2026-08-25
feat(sc): support drop_incomplete_targets_on_restore in in_order

yuki-97:
# What does this PR do ?

**Ports `drop_incomplete_targets_on_restore` from the legacy async-PPO path to the SingleController path, so a resume can discard a partially restored target step instead of gap-filling it.**

Legacy async PPO/GRPO has had this knob since #3410 (`ppo.async_ppo.drop_incomplete_targets_on_restore`, `nemo_rl/algorithms/async_utils/replay_buffer.py:531`). SC had no equivalent: a resume always gap-filled. This adds `async_rl.drop_incomplete_targets_on_restore` (default `false`, so existing runs are byte-identical).

**What a "partially restored target step" is.** In-flight rollouts are not written to the checkpoint, so a step whose batch was still filling comes back with fewer than `num_prompts_per_step` groups stamped for it.

- **`false` (default) — gap-fill.** `_rollout_pump` dispatches only the missing prompts for that step and drops the rest of that dataloader batch (`nemo_rl/algorithms/single_controller.py:630-632`).
- **`true` — drop.** The restored groups of that step are discarded at load time and the step is dispatched whole from subsequent prompts.

Neither setting regenerates the original prompts, and both consume exactly one dataloader batch per admitted step.

**A complete restored target step is already skipped — no new branch is needed for it.** The same clamp does both jobs: `num_prompts = max(0, prompt_batch.size - buffered)`. When the step is full, `num_prompts` is `0`, the dispatch loop body never runs, and that dataloader batch is consumed without generating anything. This PR only changes what `buffered` is for *incomplete* steps. Two things worth knowing about that skip, both pre-existing: the skipped batch's prompts are discarded (the dataloader is not rewound, matching legacy's `last_target_weight_already_generated` skip), and `admit` is still called and still advances the dispatch index, which it must — otherwise the stamp sequence drifts one step off the trainer version.

**`in_order` only.** `validate_single_controller_config` raises `NotImplementedError` under any other sampler, `custom` included. `in_order` is the only sampler that stamps a `target_step`; under `windowed` / `weight_fifo` / `ready_first` every group is fungible across steps, so no restored group belongs to a step that could be incomplete and the flag would silently change nothing. `custom` is rejected because setup cannot tell whether it stamps without importing it.

**Ordering:** the drop runs before the over-capacity truncation in `load_state_dict`. Truncating a target-stamped envelope raises, and the drop can bring an over-capacity checkpoint back under `max_buffered_rollouts` — so it must come first.

# Issues

None closed. Part of the SingleController parity work tracked in #2625.

# Usage

```yaml
async_rl:
  sampler:
    name: in_order
    max_lookahead_versions: 1
  # Discard a target step restored short of a full batch and dispatch it whole,
  # instead of gap-filling it from subsequent prompts.
  drop_incomplete_targets_on_restore: true
```

**Pre checks**:

# Additional Information

* **Stacked on `yukih/sc-ppo`** — base branch is that PR, not `main`.
* **No new deadlock surface.** Dropped groups never acquire a `_buffer_capacity` permit (`load_state_dict` returns the post-drop count), and per-step capacity pressure is identical either way: gap-fill holds `k` restored permits and dispatches `batch - k`; drop holds `0` and dispatches `batch`. Dataloader consumption is also identical, so `_clamp_max_num_steps`'s step budget is unaffected. Dropping strictly reduces the restored group count, so it can only make the over-capacity raise fire less often.
* **One caveat.** `_batch_shortfall` is not checkpointed, so a GRPO step that was legitimately shortened by `on_dropped_prompt="shrink"` reads as incomplete after a resume; with the flag on it is re-dispatched in full. That is extra rollout work, not a stall. PPO is unaffected — setup rejects drop budgets there.
* Tests: `tests/unit/single_controller/test_tq_replay_buffer.py` (drop / unstamped groups untouched / drop-before-truncate), `test_resiliency_config.py` (accepted under `in_order`, raises under the other four), `test_checkpointing.py` (the flag reaches the buffer).

  ```sh
  uv run --group test pytest tests/unit/single_controller/test_tq_replay_buffer.py tests/unit/single_controller/test_resiliency_config.py tests/unit/single_controller/test_checkpointing.py
  ```
* Test results. Yellow is the legacy path with the same settings, as a baseline. Red is v2. Blue is v2 resumed from checkpoint 20. Green is v2 with this PR, resumed from checkpoint 20 **with `drop_incomplete_targets_on_restore: true`**.
  image

### DIFF
--- a/docs/guides/single-controller.md
+++ b/docs/guides/single-controller.md
@@ -95,6 +95,7 @@ Field definitions:
 - `max_buffered_rollouts` — hard cap on unconsumed rollout groups buffered in the data plane. Validated at setup against the gated sampler's required capacity; a value too small deadlocks the rollout pump, so setup raises instead of silently blocking. Sized from the widest window the run ever uses, so `warmup_lookahead_versions` rather than `max_lookahead_versions` when it is set.
 - `min_groups_for_streaming_train` — minimum ready groups the trainer waits for before dispatching a batch. Set to `num_prompts_per_step` for sync/legacy semantics; lower for streaming. (PPO) Must equal `num_prompts_per_step` — the critic has no split train API, so one `train_from_meta` call is one optimizer step, and streaming a step across chunks would step the critic once per chunk.
 - `sampler.warmup_lookahead_versions` (PPO) — lookahead used while `ppo.policy_training_start_step` critic warmup is in progress, shrinking back to `max_lookahead_versions` afterwards. The SC equivalent of `ppo.async_ppo.warmup_generation_lead_steps`.
+- `drop_incomplete_targets_on_restore` — `in_order` only; setup raises under any other sampler. On resume a target step holding fewer than `num_prompts_per_step` groups is gap-filled by default: the rollout pump dispatches only the missing prompts and drops the rest of that dataloader batch. `true` discards the restored groups and dispatches the step whole instead. Neither regenerates the original prompts. The SC equivalent of `ppo.async_ppo.drop_incomplete_targets_on_restore`.
 
 ## Implementation Structure
 
--- a/docs/guides/single-controller.md
+++ b/docs/guides/single-controller.md
@@ -95,7 +95,7 @@ Field definitions:
 - `max_buffered_rollouts` — hard cap on unconsumed rollout groups buffered in the data plane. Validated at setup against the gated sampler's required capacity; a value too small deadlocks the rollout pump, so setup raises instead of silently blocking. Sized from the widest window the run ever uses, so `warmup_lookahead_versions` rather than `max_lookahead_versions` when it is set.
 - `min_groups_for_streaming_train` — minimum ready groups the trainer waits for before dispatching a batch. Set to `num_prompts_per_step` for sync/legacy semantics; lower for streaming. (PPO) Must equal `num_prompts_per_step` — the critic has no split train API, so one `train_from_meta` call is one optimizer step, and streaming a step across chunks would step the critic once per chunk.
 - `sampler.warmup_lookahead_versions` (PPO) — lookahead used while `ppo.policy_training_start_step` critic warmup is in progress, shrinking back to `max_lookahead_versions` afterwards. The SC equivalent of `ppo.async_ppo.warmup_generation_lead_steps`.
-- `drop_incomplete_targets_on_restore` — `in_order` only; setup raises under any other sampler. On resume a target step holding fewer than `num_prompts_per_step` groups is gap-filled by default: the rollout pump dispatches only the missing prompts and drops the rest of that dataloader batch. `true` discards the restored groups and dispatches the step whole instead. Neither regenerates the original prompts. The SC equivalent of `ppo.async_ppo.drop_incomplete_targets_on_restore`.
+- `drop_incomplete_targets_on_restore` — needs a sampler that stamps the target step: `in_order`, or a `custom` sampler that stamps; setup raises under the other built-ins. On resume a target step holding fewer than `num_prompts_per_step` groups is gap-filled by default: the rollout pump dispatches only the missing prompts and drops the rest of that dataloader batch. `true` discards the restored groups and dispatches the step whole instead. Neither regenerates the original prompts. The SC equivalent of `ppo.async_ppo.drop_incomplete_targets_on_restore`.
 
 ## Implementation Structure
 
@@ -171,6 +171,7 @@ SC reads its async knobs from `async_rl:` and **requires `grpo.async_grpo: null`
 | *(no legacy equivalent — matches legacy full-batch train semantics)* | `min_groups_for_streaming_train: ${grpo.num_prompts_per_step}`, or `${ppo.num_prompts_per_step}` on a PPO run |
 | *(no legacy equivalent — matches legacy `max_trajectory_age + 1` batches in flight)* | `max_inflight_prompts: num_prompts_per_step × (max_lookahead_versions + 1)` |
 | *(no legacy equivalent — legacy sizes its buffer to `num_prompts_per_step × max_trajectory_age_steps × 2`)* | `max_buffered_rollouts: num_prompts_per_step × (max_lookahead_versions + 1)` (tight; see the [Config → behavior map](#config--behavior-map) for per-sampler values) |
+| `drop_incomplete_targets_on_restore` | `drop_incomplete_targets_on_restore` (same; needs a target-stamping sampler such as `in_order`) |
 
 ## Known Missing Features
 
--- a/examples/configs/grpo_math_1B_megatron_single_controller.yaml
+++ b/examples/configs/grpo_math_1B_megatron_single_controller.yaml
@@ -21,6 +21,9 @@ async_rl:
   max_inflight_prompts: ${grpo.num_prompts_per_step}
   # Cap on unconsumed rollout groups buffered in the DataPlane (backpressure).
   max_buffered_rollouts: 64
+  # in_order only. On resume, drop a target step's restored groups when it is
+  # short of a full batch instead of gap-filling it from subsequent prompts.
+  drop_incomplete_targets_on_restore: false
   # Enable per-rollout diagnostic prints (prompt content / completion previews).
   diagnostics: false
 
--- a/examples/configs/ppo_math_1B_megatron_single_controller.yaml
+++ b/examples/configs/ppo_math_1B_megatron_single_controller.yaml
@@ -33,6 +33,9 @@ async_rl:
   max_inflight_prompts: ${ppo.num_prompts_per_step}
   # Cap on unconsumed rollout groups buffered in the DataPlane (backpressure).
   max_buffered_rollouts: 64
+  # in_order only. On resume, drop a target step's restored groups when it is
+  # short of a full batch instead of gap-filling it from subsequent prompts.
+  drop_incomplete_targets_on_restore: false
   # Enable per-rollout diagnostic prints (prompt content / completion previews).
   diagnostics: false
 
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -410,6 +410,9 @@ class SingleControllerActor:
         Skipped with a warning when the checkpoint was written under a
         different sampler: restored groups carry the saving sampler's
         weight/target-step stamps, which another policy may never select.
+
+        A target step restored short of a full batch is gap-filled by the rollout
+        pump, unless async_rl.drop_incomplete_targets_on_restore drops it instead.
         """
         if self._last_checkpoint_path is None:
             return
@@ -442,6 +445,10 @@ class SingleControllerActor:
             max_groups=self._async_cfg.max_buffered_rollouts,
             expected_partition_id=self._partition_id,
             expected_group_size=self._algo_cfg.num_generations_per_prompt,
+            groups_per_step=self._algo_cfg.num_prompts_per_step,
+            drop_incomplete_targets_on_restore=(
+                self._async_cfg.drop_incomplete_targets_on_restore
+            ),
         )
         # Each buffered group holds one _buffer_capacity permit; the load
         # truncation guarantees restored <= capacity, so this never blocks.
--- a/nemo_rl/algorithms/single_controller_utils/config.py
+++ b/nemo_rl/algorithms/single_controller_utils/config.py
@@ -420,6 +420,9 @@ class AsyncRLConfig(BaseModel, extra="allow"):
     max_inflight_prompts: int = 32
     # Cap on unconsumed rollout groups buffered in the DataPlane (backpressure).
     max_buffered_rollouts: int = 64
+    # in_order only (setup raises otherwise): on resume, drop the restored groups
+    # of a target step short of a full batch instead of gap-filling it.
+    drop_incomplete_targets_on_restore: bool = False
     # Enable per-rollout diagnostic prints (prompt content / completion previews).
     diagnostics: bool = False
 
@@ -930,6 +933,21 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
                 "the ready_first sampler requires "
                 "loss_fn.force_on_policy_ratio=false so prev_logprobs are used"
             )
+
+    # in_order is the only sampler that stamps a target step, so nowhere else can
+    # a restored step be short of one. "custom" is rejected too: setup cannot tell
+    # whether it stamps without importing it.
+    if (
+        async_config.drop_incomplete_targets_on_restore
+        and async_config.sampler.name != "in_order"
+    ):
+        raise NotImplementedError(
+            "async_rl.drop_incomplete_targets_on_restore is implemented for "
+            "async_rl.sampler.name='in_order' only, but this run uses "
+            f"{async_config.sampler.name!r}. Only in_order stamps the target step "
+            "this drops, so elsewhere the flag would silently change nothing about "
+            "the resume. Remove it, or switch to the in_order sampler."
+        )
 
     # Top-k retention keys off checkpointing.metric_name, but SC has no
     # validation loop yet (see _save_checkpoint), so a "val:" metric would
--- a/nemo_rl/algorithms/single_controller_utils/config.py
+++ b/nemo_rl/algorithms/single_controller_utils/config.py
@@ -934,19 +934,16 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
                 "loss_fn.force_on_policy_ratio=false so prev_logprobs are used"
             )
 
-    # in_order is the only sampler that stamps a target step, so nowhere else can
-    # a restored step be short of one. "custom" is rejected too: setup cannot tell
-    # whether it stamps without importing it.
+    # Only a stamped target step can be short of a batch. "custom" is left alone
+    # as in _validate_failure_settings: only its author knows whether it stamps.
     if (
         async_config.drop_incomplete_targets_on_restore
-        and async_config.sampler.name != "in_order"
+        and async_config.sampler.name not in ("in_order", "custom")
     ):
         raise NotImplementedError(
-            "async_rl.drop_incomplete_targets_on_restore is implemented for "
-            "async_rl.sampler.name='in_order' only, but this run uses "
-            f"{async_config.sampler.name!r}. Only in_order stamps the target step "
-            "this drops, so elsewhere the flag would silently change nothing about "
-            "the resume. Remove it, or switch to the in_order sampler."
+            f"async_rl.sampler.name={async_config.sampler.name!r} stamps no target "
+            "step, so async_rl.drop_incomplete_targets_on_restore would change "
+            "nothing about the resume. Remove it, or switch to the in_order sampler."
         )
 
     # Top-k retention keys off checkpointing.metric_name, but SC has no
--- a/tests/unit/single_controller/test_resiliency_config.py
+++ b/tests/unit/single_controller/test_resiliency_config.py
@@ -712,3 +712,34 @@ class TestADropBudgetNeedsASamplerThatStamps:
             rollout_failure={"max_consecutive_dropped_prompts": 2},
         )
         validate_single_controller_config(cfg)
+
+
+class TestDropIncompleteTargetsOnRestore:
+    """The knob is in_order-only: nothing else stamps the target step it drops."""
+
+    def test_default_is_off(self):
+        assert AsyncRLConfig().drop_incomplete_targets_on_restore is False
+
+    @pytest.mark.parametrize(
+        "sampler",
+        [
+            {"name": "windowed"},
+            {"name": "weight_fifo"},
+            {"name": "ready_first"},
+            # Rejected even though a custom sampler may well stamp target steps:
+            # setup cannot know that it does without importing it.
+            {"name": "custom", "target": "some_module:SomeSampler"},
+        ],
+        ids=lambda sampler: sampler["name"],
+    )
+    def test_rejected_under_every_sampler_but_in_order(self, sampler):
+        cfg = _master_config(
+            sampler=sampler,
+            drop_incomplete_targets_on_restore=True,
+        )
+        with pytest.raises(NotImplementedError, match="sampler.name='in_order' only"):
+            validate_single_controller_config(cfg)
+
+    def test_accepted_under_in_order(self):
+        cfg = _master_config(drop_incomplete_targets_on_restore=True)
+        validate_single_controller_config(cfg)
--- a/tests/unit/single_controller/test_resiliency_config.py
+++ b/tests/unit/single_controller/test_resiliency_config.py
@@ -715,7 +715,7 @@ class TestADropBudgetNeedsASamplerThatStamps:
 
 
 class TestDropIncompleteTargetsOnRestore:
-    """The knob is in_order-only: nothing else stamps the target step it drops."""
+    """The knob needs a stamping sampler: nothing else has a target step to drop."""
 
     def test_default_is_off(self):
         assert AsyncRLConfig().drop_incomplete_targets_on_restore is False
@@ -726,20 +726,25 @@ class TestDropIncompleteTargetsOnRestore:
             {"name": "windowed"},
             {"name": "weight_fifo"},
             {"name": "ready_first"},
-            # Rejected even though a custom sampler may well stamp target steps:
-            # setup cannot know that it does without importing it.
-            {"name": "custom", "target": "some_module:SomeSampler"},
         ],
         ids=lambda sampler: sampler["name"],
     )
-    def test_rejected_under_every_sampler_but_in_order(self, sampler):
+    def test_rejected_under_every_built_in_but_in_order(self, sampler):
         cfg = _master_config(
             sampler=sampler,
             drop_incomplete_targets_on_restore=True,
         )
-        with pytest.raises(NotImplementedError, match="sampler.name='in_order' only"):
+        with pytest.raises(NotImplementedError, match="stamps no target step"):
             validate_single_controller_config(cfg)
 
     def test_accepted_under_in_order(self):
         cfg = _master_config(drop_incomplete_targets_on_restore=True)
+        validate_single_controller_config(cfg)
+
+    def test_a_custom_sampler_is_not_second_guessed(self):
+        """Whether it stamps is its author's to know, as with the drop budget."""
+        cfg = _master_config(
+            sampler={"name": "custom", "target": "some_module:SomeSampler"},
+            drop_incomplete_targets_on_restore=True,
+        )
         validate_single_controller_config(cfg)
--- a/tests/unit/single_controller/test_tq_replay_buffer.py
+++ b/tests/unit/single_controller/test_tq_replay_buffer.py
@@ -35,6 +35,8 @@ from nemo_rl.experience.interfaces import PromptGroupRecord
 
 # Each record yields _N_GENS training rows.
 _N_GENS = 2
+# Groups that make one target step whole (production: num_prompts_per_step).
+_GROUPS_PER_STEP = 2
 
 
 def _stub_record_to_train_batch(
@@ -618,6 +620,8 @@ def _load(
     max_groups: int = 8,
     expected_partition_id: str = "rollout_data",
     expected_group_size: int = _N_GENS,
+    groups_per_step: int = _GROUPS_PER_STEP,
+    drop_incomplete_targets_on_restore: bool = False,
 ) -> int:
     return _run(
         buf.load_state_dict(
@@ -625,6 +629,8 @@ def _load(
             max_groups=max_groups,
             expected_partition_id=expected_partition_id,
             expected_group_size=expected_group_size,
+            groups_per_step=groups_per_step,
+            drop_incomplete_targets_on_restore=drop_incomplete_targets_on_restore,
         )
     )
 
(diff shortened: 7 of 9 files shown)


nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -917,6 +920,21 @@ def validate_single_controller_config(master_config: MasterConfig) -> None:
                 "loss_fn.force_on_policy_ratio=false so prev_logprobs are used"
             )
 
+    # in_order is the only sampler that stamps a target step, so nowhere else can
+    # a restored step be short of one. "custom" is rejected too: setup cannot tell
+    # whether it stamps without importing it.
+    if (
+        async_config.drop_incomplete_targets_on_restore
+        and async_config.sampler.name != "in_order"
```
yuki-97: [`config.py:928`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/nemo_rl/algorithms/single_controller_utils/config.py#L928)

PR-introduced. This rejects `custom` too, but [`_validate_failure_settings`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/nemo_rl/algorithms/single_controller_utils/config.py#L638-L639) deliberately doesn't -- it leaves `custom` out of *both* `sampler_stamps` and `sampler_never_stamps`, so its hard error at [`:652`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/nemo_rl/algorithms/single_controller_utils/config.py#L652) never fires for a custom sampler, and [`test_a_custom_sampler_is_not_second_guessed`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/tests/unit/single_controller/test_resiliency_config.py#L702) pins exactly that.

That lands the stricter rule on the safer knob: an unstamped drop budget turns a recoverable prompt failure into a stalled run and is permitted, while this flag under an unstamped sampler is an inert no-op and is refused. A custom sampler that *does* stamp is a supported thing to write -- [`admit`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/nemo_rl/algorithms/async_utils/staleness_sampler.py#L79) publishes `target_step` as part of the protocol -- so that user loses the flag with no workaround.

**Action:** match the sibling and let `custom` through.

```suggestion
        and async_config.sampler.name not in ("in_order", "custom")
```

docs/guides/single-controller.md
```diff
@@ -95,6 +95,7 @@ Field definitions:
...
 - `min_groups_for_streaming_train` — minimum ready groups the trainer waits for before dispatching a batch. Set to `num_prompts_per_step` for sync/legacy semantics; lower for streaming. (PPO) Must equal `num_prompts_per_step` — the critic has no split train API, so one `train_from_meta` call is one optimizer step, and streaming a step across chunks would step the critic once per chunk.
 - `sampler.warmup_lookahead_versions` (PPO) — lookahead used while `ppo.policy_training_start_step` critic warmup is in progress, shrinking back to `max_lookahead_versions` afterwards. The SC equivalent of `ppo.async_ppo.warmup_generation_lead_steps`.
+- `drop_incomplete_targets_on_restore` — `in_order` only; setup raises under any other sampler. On resume a target step holding fewer than `num_prompts_per_step` groups is gap-filled by default: the rollout pump dispatches only the missing prompts and drops the rest of that dataloader batch. `true` discards the restored groups and dispatches the step whole instead. Neither regenerates the original prompts. The SC equivalent of `ppo.async_ppo.drop_incomplete_targets_on_restore`.
```
yuki-97: [`docs/guides/single-controller.md:98`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/docs/guides/single-controller.md#L98)

This bullet ends "The SC equivalent of `ppo.async_ppo.drop_incomplete_targets_on_restore`" -- which is the relation the [migration table at `:164-173`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/docs/guides/single-controller.md#L164-L173) exists to encode, but the table has no row for it. It covers 5 of `AsyncPPOConfig`'s 6 fields ([`ppo.py:120-131`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/nemo_rl/algorithms/ppo.py#L120-L131)); the one missing is the knob this PR ports. Someone porting a legacy async config reads that table, not this field list.

[`:169`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/docs/guides/single-controller.md#L169) is already a same-name row, so the shape is established.

**Action:** add a row --

| `drop_incomplete_targets_on_restore` | `drop_incomplete_targets_on_restore` (same; `in_order` only) |

nemo_rl/algorithms/single_controller.py
```diff
@@ -406,6 +409,10 @@ async def _maybe_restore_replay_buffer(self) -> None:
             max_groups=self._async_cfg.max_buffered_rollouts,
             expected_partition_id=self._partition_id,
             expected_group_size=self._algo_cfg.num_generations_per_prompt,
+            groups_per_step=self._algo_cfg.num_prompts_per_step,
```
yuki-97, line 448: [`single_controller.py:412`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/nemo_rl/algorithms/single_controller.py#L412)

PR-introduced. Change this to `num_generations_per_prompt` and every test still passes: [`_actor_master_config`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/tests/unit/single_controller/test_checkpointing.py#L323) fixes `num_prompts_per_step` and `num_generations_per_prompt` both at 2 ([`:333`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/tests/unit/single_controller/test_checkpointing.py#L333), [`:360`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/tests/unit/single_controller/test_checkpointing.py#L360)), and the new `test_the_drop_flag_is_forwarded_to_the_buffer` asserts only the boolean. Shipped recipes differ (e.g. 32 prompts x 8 generations), so a swap here is a live bug class.

**Action:** add a `num_generations_per_prompt` param to `_actor_master_config`, derive `train_global_batch_size` from it ([`:351`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/tests/unit/single_controller/test_checkpointing.py#L351) hardcodes `* 2`), pass 4 in the new test, and assert `groups_per_step == 2` beside `expected_group_size == 4`. Required capacity is unchanged at 2x(1+1)=4 <= 4.

Trap worth knowing: raising `num_prompts_per_step` to 3 instead fails setup -- [`_gated_required_buffer_capacity`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/nemo_rl/algorithms/async_utils/staleness_sampler.py#L312-L318) then wants 3x(1+1)=6 against the hardcoded [`max_buffered_rollouts=4`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/tests/unit/single_controller/test_checkpointing.py#L388).

tests/unit/single_controller/test_tq_replay_buffer.py
```diff
@@ -32,6 +32,8 @@
 
 # Each record yields _N_GENS training rows.
 _N_GENS = 2
+# Groups that make one target step whole (production: num_prompts_per_step).
+_GROUPS_PER_STEP = 2
```
yuki-97: [`test_tq_replay_buffer.py:36`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/tests/unit/single_controller/test_tq_replay_buffer.py#L36)

`_GROUPS_PER_STEP` and [`_N_GENS`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/tests/unit/single_controller/test_tq_replay_buffer.py#L34) are both 2 and no test overrides `groups_per_step`, so the two thresholds collapse: mutating [`replay_buffer.py:1086`](https://github.com/NVIDIA-NeMo/RL/blob/4d45be53210904cf1072ae6d8f6eef32673d5fe6/nemo_rl/algorithms/async_utils/replay_buffer.py#L1086) to `count < expected_group_size` leaves all four new tests green.

**Action:** `_GROUPS_PER_STEP = 3`, with `_partial_envelope` holding three groups at target 1 and **two** at target 2. Correct code then drops target 2 (`2 < 3`) and restores 3; the swapped threshold drops nothing (`2 < 2` is false) and restores 5. Three assertions need the new counts, and `test_dropping_runs_before_truncation` needs `max_groups=3`.

Three-and-one doesn't work -- `1 < 3` and `1 < 2` both mark target 2 incomplete, so the swap survives it.

yuki-97 (commented):
Team review (5 agents: RL expert, bug finder, test agent, design reviewer, devil's advocate).

**No correctness findings.** The deadlock/liveness audit came back clean: peak buffer occupancy is `N*(L+1) <= C` and is *identical* with the flag on or off (gap-fill holds `k` restored permits and dispatches `N-k`; drop holds 0 and dispatches `N`), which rests on `drop_last=True` on the dataloader so every batch is exactly `num_prompts_per_step`. Dropping only ever reduces held permits. The sibling-inheritance check against the legacy async-PPO implementation found three structural deltas, all inherited or outcome-identical. Drop-before-truncate is right and the ordering is enforced by an executable test, not just the docstring.

Four suggestions below, all tests and docs.

Two caveats on this review: nothing was executed (`uv` could not run in this environment), so the mutation-survival claims are hand-traced against the source rather than run; and the perf evidence attached to the PR is a PPO run, where the `shrink` caveat you disclosed cannot fire by construction -- not asking for more, just noting the GRPO x `shrink` path has no run behind it.

_Generated by Claude Code_

nemo_rl/algorithms/single_controller_utils/config.py
yuki-97: fixed in 2fcc765498a3b7679577e76eac5217ed28bb43e2

docs/guides/single-controller.md
yuki-97: fixed in 2fcc765498a3b7679577e76eac5217ed28bb43e2

nemo_rl/algorithms/single_controller.py
yuki-97: fixed in 2fcc765498a3b7679577e76eac5217ed28bb43e2

tests/unit/single_controller/test_tq_replay_buffer.py
yuki-97: fixed in 2fcc765498a3b7679577e76eac5217ed28bb43e2

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -418,6 +418,9 @@ class AsyncRLConfig(BaseModel, extra="allow"):
     max_inflight_prompts: int = 32
     # Cap on unconsumed rollout groups buffered in the DataPlane (backpressure).
     max_buffered_rollouts: int = 64
+    # in_order only (setup raises otherwise): on resume, drop the restored groups
+    # of a target step short of a full batch instead of gap-filling it.
```
terrykong: **1 action item.**

PR-introduced. `2ccfec2b` widened the gate to accept `custom` too — [`:948`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L948) is `not in ("in_order", "custom")` — but this comment did not follow. The prose docs got it right ([`single-controller.md:98`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/docs/guides/single-controller.md#L98)).

A user on `sampler.name: custom` reads "in_order only", concludes the knob is unavailable, and deletes a line that works. Nothing corrects them.

**Action:** say which samplers qualify. Same edit at [`examples/configs/grpo_math_1B_megatron_single_controller.yaml:24`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/examples/configs/grpo_math_1B_megatron_single_controller.yaml#L24) and [`examples/configs/ppo_math_1B_megatron_single_controller.yaml:36`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/examples/configs/ppo_math_1B_megatron_single_controller.yaml#L36).

```suggestion
    # Needs a target-stamping sampler -- in_order, or a custom sampler that
    # stamps; setup raises under the other built-ins. On resume, drop the
    # restored groups of a target step short of a full batch instead of
    # gap-filling it.
```

examples/configs/grpo_math_1B_megatron_single_controller.yaml
```diff
@@ -21,6 +21,9 @@ async_rl:
   max_inflight_prompts: ${grpo.num_prompts_per_step}
   # Cap on unconsumed rollout groups buffered in the DataPlane (backpressure).
   max_buffered_rollouts: 64
+  # in_order only. On resume, drop a target step's restored groups when it is
+  # short of a full batch instead of gap-filling it from subsequent prompts.
```
terrykong: **1 action item.**

PR-introduced. Same stale "in_order only" as [`config.py:421`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L421) — the gate at [`config.py:948`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L948) also accepts `custom`.

**Action:**

```suggestion
  # Needs a target-stamping sampler (in_order, or a custom one that stamps). On
  # resume, drop a target step's restored groups when it is short of a full
  # batch instead of gap-filling it from subsequent prompts.
```

examples/configs/ppo_math_1B_megatron_single_controller.yaml
```diff
@@ -33,6 +33,9 @@ async_rl:
   max_inflight_prompts: ${ppo.num_prompts_per_step}
   # Cap on unconsumed rollout groups buffered in the DataPlane (backpressure).
   max_buffered_rollouts: 64
+  # in_order only. On resume, drop a target step's restored groups when it is
+  # short of a full batch instead of gap-filling it from subsequent prompts.
```
terrykong: **1 action item.**

PR-introduced. Same stale "in_order only" as [`config.py:421`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L421) — the gate at [`config.py:948`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L948) also accepts `custom`.

**Action:**

```suggestion
  # Needs a target-stamping sampler (in_order, or a custom one that stamps). On
  # resume, drop a target step's restored groups when it is short of a full
  # batch instead of gap-filling it from subsequent prompts.
```

tests/unit/single_controller/test_tq_replay_buffer.py
```diff
@@ -832,3 +839,71 @@ def test_target_stamped_groups_within_capacity_load_fine(self):
...
+        # Only a stamped step can be short of a batch; under a sampler that
+        # stamps nothing the knob must leave the whole checkpoint alone.
+        state = _make_envelope(
+            [_make_group_entry(f"g{w}", weight=w) for w in (1, 2, 3)],
+            saved_capacity=8,
+        )
+        buf = _make_buffer(FakeDataPlaneClient())
+
+        assert _load(buf, state, drop_incomplete_targets_on_restore=True) == 3
+        assert buf.target_step_list == [None, None, None]
```
terrykong: **1 action item.**

**TL;DR — [`test_unstamped_groups_are_never_dropped`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_tq_replay_buffer.py#L884-L894) cannot fail: with an all-unstamped envelope the filter it is named for never runs, so two real mutations of the drop block survive the whole new suite.**

PR-introduced. The drop block is guarded twice ([`replay_buffer.py:1112-1139`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/async_utils/replay_buffer.py#L1112-L1139)):

1. the `Counter` skips unstamped groups — [`:1117`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/async_utils/replay_buffer.py#L1117) `if group["target_step"] is not None`
2. so with *only* unstamped groups `target_counts` is empty, `incomplete` is empty
3. so `if incomplete:` is False and the `kept` comprehension at [`:1126`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/async_utils/replay_buffer.py#L1126) — the only code that could ever drop an unstamped group — **never executes**

The test asserts 3 in, 3 out. That is also what you get if the filter is deleted. Two mutations survive it and every other new test:

- drop the `is not None` guard at [`:1117`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/async_utils/replay_buffer.py#L1117) → `Counter` yields `{None: 3}`, `3 < 3` is False, nothing dropped
- tighten [`:1126`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/async_utils/replay_buffer.py#L1126) to `is not None and not in incomplete` → the block is skipped anyway

I ran both against a mixed envelope: correct code restores **5**, either mutation restores **3**. So the docstring's promise at [`:1050`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/async_utils/replay_buffer.py#L1050) — "Groups carrying no target_step are never dropped" — is pinned by nothing.

This stopped being hypothetical in `2ccfec2b`. Under `in_order` every group is stamped, so a mixed envelope cannot arise. But [`config.py:945-955`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L945-L955) now admits the flag under a `custom` sampler, and the stated reason for admitting it is that a custom sampler *may* stamp — including one that stamps only after warmup. That is exactly a mixed envelope.

**Action:** keep the existing test and add one that reaches the filter. Fewer than `_GROUPS_PER_STEP` unstamped groups on purpose, so counting `None` as a step would drop them.

```suggestion
        assert buf.target_step_list == [None, None, None]

    def test_unstamped_groups_survive_a_step_being_dropped(self):
        # The all-unstamped case above never reaches the kept-filter: with
        # nothing incomplete the whole block is skipped. Mix the two so the
        # filter actually runs -- the short step goes, the unstamped stay.
        state = _make_envelope(
            [
                _make_group_entry("g0", weight=1, target_step=1),
                _make_group_entry("g1", weight=1, target_step=1),
                _make_group_entry("g2", weight=1, target_step=1),
                _make_group_entry("g3", weight=2, target_step=2),
                _make_group_entry("g4", weight=2, target_step=2),
                _make_group_entry("g5", weight=3),
                _make_group_entry("g6", weight=3),
            ],
            saved_capacity=8,
        )
        buf = _make_buffer(FakeDataPlaneClient())

        assert _load(buf, state, drop_incomplete_targets_on_restore=True) == 5
        assert buf.target_step_list == [1, 1, 1, None, None]
```

Not run locally — this repo's `uv.lock` is linux-only and I had no GPU, so please run it before committing. It should pass as-is and fail under either mutation above.

tests/unit/single_controller/test_resiliency_config.py
```diff
@@ -712,3 +712,39 @@ def test_a_custom_sampler_is_not_second_guessed(self):
...
+        cfg = _master_config(drop_incomplete_targets_on_restore=True)
+        validate_single_controller_config(cfg)
+
+    def test_a_custom_sampler_is_not_second_guessed(self):
+        """Whether it stamps is its author's to know, as with the drop budget."""
+        cfg = _master_config(
+            sampler={"name": "custom", "target": "some_module:SomeSampler"},
+            drop_incomplete_targets_on_restore=True,
+        )
+        validate_single_controller_config(cfg)
```
terrykong: **1 action item.**

**TL;DR — [`test_rejected_under_every_built_in_but_in_order`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_resiliency_config.py#L729) promises "every built-in" in its name but checks a hardcoded list of three, so the next sampler is silently exempt.**

PR-introduced. `TestDropIncompleteTargetsOnRestore` mirrors [`TestUnstampedDropBudget`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_resiliency_config.py#L640) closely — same `_master_config`, same `match="stamps no target step"`, and `test_a_custom_sampler_is_not_second_guessed` carried over almost verbatim — but omits the one test that class exists to carry: [`test_every_built_in_but_in_order_and_custom_is_covered`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_resiliency_config.py#L659-L675), which reads the sampler list off the `SamplerConfig` union via [`_all_sampler_names()`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_resiliency_config.py#L44).

That test's own docstring says this already happened once:

> "ready_first (#3582) landed between this PR's base and its rebase, inheriting the None stamp like the rest. An allow-list would have let it straight through, and **a test that hardcoded the same list would have stayed green while it did.**"

The new check at [`config.py:948`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L948) uses the identical `not in ("in_order", "custom")` shape, so it inherits the identical exposure — guarded only by the three-name parametrize at [`:724-731`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_resiliency_config.py#L724-L731).

Related, and the reason this test is worth more than a refactor: the same predicate is now spelled out at [`config.py:652-653`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L652-L653) and again at [`:948`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L948), with the "named by exclusion" rationale living at only the first. A union-derived test is what keeps the two honest; extracting a shared `sampler_stamps_target_step()` beside [`required_buffer_capacity_for_config`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/async_utils/staleness_sampler.py#L582) would be a reasonable follow-up, not something I'd ask for here.

**Action:** append the sibling's union-derived test. Note `NotImplementedError` derives from `RuntimeError`, **not** `ValueError`, so the sibling's exception type would not catch this one. No new imports — `_all_sampler_names` is at [`:44`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_resiliency_config.py#L44).

```suggestion
        validate_single_controller_config(cfg)

    def test_every_built_in_but_in_order_and_custom_is_covered(self):
        """Read off the union, so a sampler added later is caught by default."""
        exempt = {"in_order", "custom"}
        covered = [n for n in _all_sampler_names() if n not in exempt]
        assert covered, "the union should carry at least one unstamped sampler"
        for sampler_name in covered:
            cfg = _master_config(
                sampler={"name": sampler_name},
                drop_incomplete_targets_on_restore=True,
            )
            with pytest.raises(NotImplementedError, match="stamps no target step"):
                validate_single_controller_config(cfg)
```

Not run locally (linux-only `uv.lock`, no GPU) — please run before committing.

tests/unit/single_controller/test_ppo_setup.py
```diff
@@ -407,6 +407,19 @@ def test_rejects_colocated_generation(self, make_config):
         with pytest.raises(ValueError, match="colocated.enabled=false"):
             validate_single_controller_config(mc)
 
+    @pytest.mark.parametrize(
+        "make_config", [_ppo_master_config, _make_master_config], ids=["ppo", "grpo"]
+    )
+    @pytest.mark.parametrize("epochs", [-1, 0], ids=["negative", "zero"])
+    def test_rejects_a_non_positive_max_num_epochs(self, make_config, epochs):
```
terrykong: **1 action item** — about [`test_the_exemplars_still_validate`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_ppo_setup.py#L183-L184) at `:183-184`, anchored here because that line is outside this PR's diff.

**TL;DR — the shipped SC exemplars are never run through `validate_single_controller_config`, so a `defaults:`-chain regression in either one goes red only at launch, not in CI.**

Pre-existing gap, newly relevant: this PR adds two cross-field rules — this one, and the drop-flag sampler gate — and neither is exercised against a real shipped config. [`test_the_exemplars_still_validate`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_ppo_setup.py#L183-L184) constructs `MasterConfig` — pydantic field validation only — and stops there.

The sharpest case is not the new `max_num_epochs` rule but buffer capacity. Both exemplars sit at **exactly** the validated minimum: `num_prompts_per_step` 32 x (`max_lookahead_versions` 1 + 1) = 64, and [`max_buffered_rollouts: 64`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/examples/configs/grpo_math_1B_megatron_single_controller.yaml#L23). The check is a strict `<` at [`config.py:605`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L605). So raising `max_lookahead_versions` to 2 without raising `max_buffered_rollouts` ships a config that deadlocks the rollout pump at launch — and nothing in the unit suite goes red. The guide calls this sizing "(tight)" at [`single-controller.md:175`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/docs/guides/single-controller.md#L175) and warns "a value too small deadlocks the rollout pump" at [`:95`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/docs/guides/single-controller.md#L95).

I hand-evaluated every branch of the validator against both resolved exemplars and both pass today, so this should not ship a red test. `validate_single_controller_config` is already imported at [`:54`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_ppo_setup.py#L54).

**Action:** at [`tests/unit/single_controller/test_ppo_setup.py:183-184`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/tests/unit/single_controller/test_ppo_setup.py#L183-L184), have that test run the validator too. Not a `suggestion` block because the target line is not in this PR's diff:

```python
    def test_the_exemplars_still_validate(self, name):
        mc = MasterConfig(**self._resolved(name))
        # Field validation alone would miss a defaults:-chain regression from any
        # cross-field SC rule (buffer capacity, max_num_epochs, sampler-gated knobs).
        validate_single_controller_config(mc)
```

Two caveats, both worth a moment before you commit. I could not execute this (linux-only `uv.lock`, no GPU) — the pass/fail table above is hand-traced across a four-file `defaults:` chain. And the validator calls [`_validate_failure_settings`](https://github.com/NVIDIA-NeMo/RL/blob/85ad93ca74cea3b698afb9d441e64b32c02b4574/nemo_rl/algorithms/single_controller_utils/config.py#L1011), which *warns*; both exemplars use `on_dropped_prompt: shrink` with zero budgets so I expect none, but if this class ever runs under `filterwarnings("error")` it would go red for an unrelated reason.

Extending the same coverage to the seven SC recipes under `examples/configs/recipes/llm/*single-controller*.yaml` would be more valuable still — that is where a `-1` `max_num_epochs` actually lives, survived only by a one-line override. Bigger change; I would not block on it.

terrykong:
note for those reading that we'll probably focus on getting #3480 in so we can restore 100% accurately instead of dropping

PR #3773, opened 2026-08-23
feat(sc): support PPO in single controller

yuki-97:
# What does this PR do ?

Ports PPO — including the critic/value model — to the SingleController (SC) path, so an async PPO run can be driven by `examples/run_grpo_single_controller.py` instead of `run_ppo.py`. Megatron critic only, async only.

## What lands here

**Config.** `MasterConfig.grpo` becomes optional and a `ppo:` block is added beside it; the two are mutually exclusive and every read goes through `algo_config()`. The advantage estimator config moves from `dict` to a BaseModel with `name` pinned to a `Literal`. `_validate_algo_settings` rejects the misconfigurations listed below at setup time, each message naming the fix.

**Value model on the data plane.** New `TQValue` (driver) and `get_values_presharded` (worker), with `TQDriverMixin` extracted from `tq_policy.py` so the policy and the critic share one copy of the packing / data-plane helpers. `PPO_VALUE_FIELDS` is deliberately kept out of `DP_TRAIN_FIELDS` so a GRPO run never fetches a column nobody wrote.

**Train pump.** `_value_stage` (critic forward) and `_value_train` (critic optimizer step) around the existing advantage stage, plus `ppo.ppo_epochs` (N optimizer steps per RL step, the policy offloaded between epochs — closes the "multiple mini-steps inside a single RL step" item in #2625), `ppo.policy_training_start_step` (critic warmup: weight sync skipped, policy optimizer not checkpointed, step excluded from best-checkpoint tracking), and `async_rl.sampler.warmup_lookahead_versions` (the SC equivalent of legacy `warmup_generation_lead_steps`; buffer capacity is sized from the peak window). The critic forward runs after the policy and reference logprobs rather than before as in `ppo.py`, the policy is parked on CPU across both critic stages, and the two checkpoint saves are serialized critic-first.

**Setup.** The train cluster is sized for two worker groups on a PPO run, the critic is built and resumed alongside the policy, and the Megatron tick budget is scaled by `ppo_epochs`.

## Constraints enforced at config time

- `ppo.async_ppo: null` — SC drives the async loop itself (rejected in the entrypoint).
- `ppo` and `grpo` are mutually exclusive, and exactly one must be set.
- `value` and `value_loss_fn` are present iff the `ppo` block is, and `value.megatron_cfg.enabled: true`.
- `ppo.ppo_epochs >= 1`.
- `async_rl.sampler.name: in_order`.
- `min_groups_for_streaming_train == ppo.num_prompts_per_step`.
[...]

PR #3821, opened 2026-08-25
feat(ppo): warm-start the value model from a critic-pretrain checkpoint

yuki-97:
# What does this PR do ?

**Adds `ppo.warm_start_value_checkpoint`: a fresh PPO run seeds its critic from a separately pretrained value checkpoint, so critic warmup can be paid once offline instead of inside every PPO run.**

Today the only way to reach a usable critic is `ppo.policy_training_start_step: N` — N steps during which generation and environment scoring run at full cost while only the critic trains. Where generation dominates the step cost (SWE, long-horizon agentic runs), that is the most expensive way to buy a value function. Pretraining the critic on already-collected rollouts and starting PPO from it moves that cost off the PPO allocation entirely.

**How it works.** The value component's resume paths are resolved independently of the policy's (`CheckpointManager.get_resume_paths(..., model_component="value")`), so the whole feature is a choice of which checkpoint that call reads:

```python
warm_start = ppo_config.warm_start_value_checkpoint
if last_checkpoint_path is None and warm_start is not None:
    validate_warm_start_checkpoint(warm_start)
    print(f"🔥 Warm-starting the value model from {warm_start}")
value_weights_path, value_optimizer_path = checkpointer.get_resume_paths(
    last_checkpoint_path or warm_start,
    model_component="value",
)
```

`validate_warm_start_checkpoint` rejects an empty string and a path with no `value/weights`. `get_resume_paths` never stats what it is given, so without it a typo prints the 🔥 line over a cold critic.

The seed is a `step_<n>` directory holding a `value/` subtree — what a PPO or critic-pretraining run already checkpoints — and the critic restores its weights, plus optimizer moments and LR-scheduler state whenever the seed carries them. On Megatron the seed and this run must then agree on the value scheduler — `docs/guides/ppo.md` lists the fields. Nothing else is read: the policy starts from the base model, the dataloader from the beginning, and the step counter at 0. Those are a fresh run's defaults, so no extra handling is needed for them.

**Fresh runs only, deliberately.** When the run already has a checkpoint of its own, that checkpoint wins and the setting is not read. This is what lets the key stay in the config across resumes: a requeued run restores its own critic rather than re-seeding from the pretrained one, with no config edit between submissions. Making the warm start win instead would silently discard the critic's progress on every resume.

**Both paths.** The knob lives on `PPOConfig`, and SC's `algo_config()` returns that same object on a PPO run, so legacy `ppo.setup` and `setup_single_controller` are both covered by one field.

# Issues

None closed.

# Usage

```yaml
ppo:
  warm_start_value_checkpoint: /path/to/critic_pretrain_run/step_370
  policy_training_start_step: 0  # the critic is already warm
```

`policy_training_start_step: 0` is the natural pairing, but keeping a short online warmup is equally valid — it calibrates the seeded critic on this run's own rollout distribution before the policy starts moving. A nonzero value is also what `async_ppo.warmup_generation_lead_steps` requires (`async_rl.sampler.warmup_lookahead_versions` on the SingleController path); both must be null when it is 0.

**Pre checks**:

# Additional Information

* **Byte-identical on a PPO run when unset.** The default is `null`, so `last_checkpoint_path or warm_start` reduces to `last_checkpoint_path`.
* **One shape change on the SC path.** Value resume-path resolution now sits inside `if is_ppo_run(master_config)`, so a GRPO run no longer resolves value paths it never uses. Side effect on the GRPO path: a GRPO resume previously warned `Optimizer state not found` about a value model it does not have. Warning text only.
* **Tests.** `test_ppo_setup.py::TestValueWarmStart` covers the near end on the SC path — the seeded paths reach `_build_value` while the trainer's `weights_path` stays `None`, so the policy is not seeded, plus `test_ppo.py::test_ppo_setup_rejects_a_warm_start_that_does_not_resolve` for the legacy `ppo.setup` copy. `test_{megatron,dtensor}_value_worker.py::test_value_worker_checkpoint_save_and_load` covers the far end: a restored value worker reproduces the saved model's predictions exactly, so a wrong checkpoint or a partial restore fails.

  ```sh
  uv run --group test pytest tests/unit/single_controller/test_ppo_setup.py tests/unit/algorithms/test_ppo.py tests/unit/utils/test_checkpoint.py
  # 2 GPUs + HF_TOKEN
  uv run --group test --extra mcore pytest tests/unit/models/value/test_megatron_value_worker.py::test_value_worker_checkpoint_save_and_load
  ```
* **One unrelated fix** to resolve [`#3773 (comment)`](https://github.com/NVIDIA-NeMo/RL/pull/3773#discussion_r3850651580). SC has no `-1` convention — the rollout pump gates on `_current_epoch < max_num_epochs`, so an async-PPO user following `docs/guides/ppo.md` and carrying `max_num_epochs: -1` onto SC got zero optimizer steps and exit status 0. `_validate_algo_settings` now rejects a non-positive value and names the remedy; `_clamp_max_num_steps` no longer needs its own `<= 0` early return.
* **No functional test.** A seeded run and a cold control do not share rollouts (async vLLM is not reproducible even at a fixed seed), so a cross-run critic-loss comparison is a statistical claim, not an identity check. The one property e2e would add — a real SC PPO run writing a loadable `step_<n>/value/` subtree — is already asserted by `tests/functional/ppo_async_single_controller.sh`.


examples/configs/ppo_math_1B.yaml
```diff
@@ -10,6 +10,10 @@ ppo:
   max_num_steps: 100000
   ppo_epochs: 4
   policy_training_start_step: 0  # number of PPO steps of critic-only warmup before policy training begins
+  # step_<n> dir of a critic-pretrain run whose value/ seeds the critic.
+  # Only a fresh run reads it; a resume ignores it and restores the critic from
+  # its own checkpoint, so it can stay set.
+  warm_start_value_checkpoint: null
```
yuki-97, line 16: [`examples/configs/ppo_math_1B.yaml:16`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/examples/configs/ppo_math_1B.yaml#L16)

**TL;DR — `tests/unit/test_config_v2.py::test_reference_configs_up_to_date[ppo_math_1B_megatron.yaml]` fails on this branch: the new key reaches the resolved config through `defaults:` but the reference config was not updated.**

[`ppo_math_1B_megatron.yaml:4`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/examples/configs/ppo_math_1B_megatron.yaml#L4) is `defaults: ppo_math_1B.yaml`, and `load_config` is [`return load_config_with_inheritance(config_path)`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/config.py#L163), so the merged real config now carries `ppo.warm_start_value_checkpoint`. [`_collect_mismatched_keys`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/tests/unit/test_config_v2.py#L70) recurses into `ppo`, and the test asserts [`len(missing) == 0`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/tests/unit/test_config_v2.py#L104). The reference file is flat and stops at `policy_training_start_step` ([`:12`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/tests/unit/reference_configs/ppo_math_1B_megatron.yaml#L12)).

```
Keys present in real config .../examples/configs/ppo_math_1B_megatron.yaml but missing from
reference config .../tests/unit/reference_configs/ppo_math_1B_megatron.yaml:
  - ppo.warm_start_value_checkpoint
```

This is what `config-conventions/SKILL.md` § "Documenting New Config Keys" requires: *"If the change affects an exemplar covered by `tests/unit/test_config_v2.py`, also update the matching `tests/unit/reference_configs/*.yaml`."*

**Action:** add `warm_start_value_checkpoint: null` after line 12 of `tests/unit/reference_configs/ppo_math_1B_megatron.yaml`. That is the only affected parametrization — it is the sole reference config with a `ppo` block, and the SC exemplar has no reference twin. (No suggestion block: the edit is in a file outside this diff.)

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -641,10 +641,16 @@ def setup_single_controller(
...
-        model_component="value",
-    )
+    if is_ppo_run(master_config):
+        # Only a fresh run reads this; a resume ignores it and restores the critic
+        # from its own checkpoint, so the key can stay in the config.
+        warm_start = master_config.ppo.warm_start_value_checkpoint
+        if last_checkpoint_path is None and warm_start is not None:
+            print(f"🔥 Warm-starting the value model from {warm_start}")
+        value_weights_path, value_optimizer_path = checkpointer.get_resume_paths(
+            last_checkpoint_path or warm_start,
```
yuki-97, line 803: [`nemo_rl/algorithms/single_controller_utils/setup.py:651`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/algorithms/single_controller_utils/setup.py#L651)

**1 action item, 1 doc fix.** PR-introduced.

**TL;DR — seeding from a critic-pretrain run whose step budget differs from this run's kills the critic at init with a raw mcore assert, and it is the doc's own worked example that triggers it.**

Background: on a valid Megatron seed, `get_resume_paths` returns a non-None `optimizer_path` ([`checkpoint.py:223-229`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/checkpoint.py#L223)), so `load_optim=True` ([`megatron/setup.py:1276`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/models/megatron/setup.py#L1276)). Two defaults make the rest inevitable: `finetune` stays False ([Megatron-LM `training_config.py:464`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/training/config/training_config.py#L464) — `_create_checkpoint_config` never sets it) and `save_optim` is True, so the seed run wrote scheduler state. Bridge therefore restores it: [`checkpointing.py:3115-3143`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/training/checkpointing.py#L3115) → `opt_param_scheduler.load_state_dict(...)`.

That call runs [`_check_and_set`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/optimizer_param_scheduler.py#L332), which **asserts equality** because `use_checkpoint_opt_param_scheduler` defaults False and nothing in this repo sets it.

Which field trips depends on the config, and with the shipped one it is `wd_incr_steps`: `ppo_math_1B.yaml` pins `lr_decay_style: "constant"`, `lr_decay_iters: 1000`, `lr_warmup_iters: 10`, `lr: 2.0e-6`, `min_lr: 1.0e-7` as literals ([`:398-404`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/examples/configs/ppo_math_1B.yaml#L398), [`:381`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/examples/configs/ppo_math_1B.yaml#L381)), identical in both runs, so every earlier check passes. But `scheduler.max_steps` is unset, so [`wd_incr_steps = train_iters * global_batch_size`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/training/config.py#L1776) — and `train_iters` is this run's injected `max_num_steps * ppo_epochs` ([`setup.py:426-445`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/algorithms/single_controller_utils/setup.py#L426)).

```
AssertionError: OptimizerParamScheduler: class input value {X} and checkpointvalue {Y}
for total number of weight decay iterations do not match
```

So `warm_start_value_checkpoint: .../step_370` + `ppo.max_num_steps: 100000` — the pairing in the guide — asserts, whenever the pretrain run was configured short rather than merely stopped early. It passes only when `max_num_steps * ppo_epochs * train_global_batch_size` matches on both sides, which is why an e2e run seeded from the same YAML looks fine.

**Action:** either set `value.megatron_cfg.scheduler.override_opt_param_scheduler: true` on the warm-start path (settable via `SchedulerConfig(**config["megatron_cfg"]["scheduler"])`, so this run's schedule wins and only weights + moments come from the seed), or document that the seed run's scheduler/optimizer settings and step budget must match. Note `self.step(increment=num_steps)` at [`optimizer_param_scheduler.py:390`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/optimizer_param_scheduler.py#L390) is outside `_check_and_set`, so the seed's step count carries over either way — harmless under `lr_decay_style: constant`, not under `cosine`/WSD.

**Doc fix, same root cause:** [`docs/guides/ppo.md:217`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/docs/guides/ppo.md#L217) says the critic "restores its weights, optimizer moments and LR-scheduler state" unconditionally. It degrades to weights-only whenever the seed carries no optimizer state (e.g. `save_optimizer: false`), signalled only by the optimizer-only warning at [`checkpoint.py:232`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/checkpoint.py#L232). The promise holds exactly when the assert above is armed — worth stating both together.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -641,10 +641,16 @@ def setup_single_controller(
...
-    value_weights_path, value_optimizer_path = checkpointer.get_resume_paths(
-        last_checkpoint_path,
-        model_component="value",
-    )
+    if is_ppo_run(master_config):
+        # Only a fresh run reads this; a resume ignores it and restores the critic
+        # from its own checkpoint, so the key can stay in the config.
+        warm_start = master_config.ppo.warm_start_value_checkpoint
+        if last_checkpoint_path is None and warm_start is not None:
+            print(f"🔥 Warm-starting the value model from {warm_start}")
```
yuki-97, line 801: [`nemo_rl/algorithms/single_controller_utils/setup.py:648`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/algorithms/single_controller_utils/setup.py#L648)

**TL;DR — a `warm_start_value_checkpoint` that does not resolve prints the 🔥 line and then trains with a cold, randomly-initialised critic; on the SC path nothing ever raises.**

This is the first checkpoint path here that comes straight from user config. `last_checkpoint_path` could only ever be a directory `get_latest_checkpoint_path()` already found on disk; `warm_start` is an arbitrary string and nothing validates it. [`get_resume_paths`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/checkpoint.py#L207) returns `weights_path` unexamined on every branch — it never stats it.

**Worked example, the quiet one.** A launcher wires the key to a variable, the way every recipe script already passes overrides ([`ppo-…-dynbatch.sh:27-28`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/tests/test_suites/llm/ppo-qwen2.5-1.5b-gsm8k-1n8g-megatron-valuetp2sp-dynbatch.sh#L27) does exactly this with `checkpointing.checkpoint_dir=$CKPT_DIR` followed by `$@`):

```sh
ppo.warm_start_value_checkpoint=$WARM_START   # WARM_START unset on the cold-start run
```

The literal `key=` prefix survives word-splitting, so the token reaching Hydra is `ppo.warm_start_value_checkpoint=`. Hydra parses that to the **empty string**, not to `None` — grammar [`OverrideParser.g4:11`](https://github.com/facebookresearch/hydra/blob/v1.3.2/hydra/grammar/OverrideParser.g4#L11) (`key EQUAL value?  // key=value, key= (for empty value)`) and visitor [`overrides_visitor.py:170-172`](https://github.com/facebookresearch/hydra/blob/v1.3.2/hydra/core/override_parser/overrides_visitor.py#L170) (`if ctx.value() is None: value = ""`). `str | None` accepts it. Then:

- `"" is not None` → the 🔥 line prints with a blank path.
- `None or ""` is falsy at [`checkpoint.py:207`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/checkpoint.py#L207) → `get_resume_paths` returns `(None, None)` **before** reaching the warning at [`:232`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/checkpoint.py#L232).

Net output for the whole failure: one line claiming a warm start, and nothing else.

**The same defect with a typo'd path is louder but still silent enough to finish a run.** `get_resume_paths` returns `(<typo>/value/weights, None)`; `resume_checkpoint_exists` is False ([`megatron/setup.py:1699`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/models/megatron/setup.py#L1699)) but `pretrained_checkpoint_exists` is True — `validate_model_paths` never returns None ([`:479-498`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/models/megatron/setup.py#L479)) — so Bridge falls back to the **base HF weights** ([`checkpointing.py:2349-2363`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/training/checkpointing.py#L2349)) behind a single `print_rank_0`. With the documented pairing `policy_training_start_step: 0` there is no online warmup to fall back on either.

The DTensor critic raises on the same input ([`native_checkpoint.py:229`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/native_checkpoint.py#L229)), so the silent path is exactly the Megatron-only SC path this feature targets. The nearest analog in the repo validates: `checkpointing.pretrained_checkpoint`, also user-supplied, gets an explicit check ([`megatron/setup.py:1471-1475`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/models/megatron/setup.py#L1471)).

**Action:** fail loud at setup. `value/weights` is the right probe — it is the one component both backends share, built before `get_resume_paths` branches. Verified this rejects nothing legitimate: both new tests still pass (the fresh-run test creates `seed/value/weights`; the unset test skips the guard), every shipped recipe has the key `null`, and resumes never reach it.

```suggestion
        if last_checkpoint_path is None and warm_start:
            if not (Path(warm_start) / "value" / "weights").exists():
                raise ValueError(
                    f"ppo.warm_start_value_checkpoint={warm_start!r} has no value/weights "
                    "subtree; the critic would silently start cold."
                )
            print(f"🔥 Warm-starting the value model from {warm_start}")
```

Truthiness in the guard is what closes the `""` case; the existence check is what closes the typo case. `Path` is already imported ([`:28`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/algorithms/single_controller_utils/setup.py#L28)). The twin at [`ppo.py:714`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/algorithms/ppo.py#L714) needs the same guard — that file imports `os` but not `pathlib`.

tests/unit/single_controller/test_ppo_setup.py
```diff
@@ -509,6 +509,42 @@ def test_grpo_run_builds_no_critic(self, patched_ppo_factories):
...
+
+    def test_unset_leaves_the_critic_cold(self, patched_ppo_factories, tmp_path):
+        mc = _ppo_master_config()
+        mc.checkpointing["checkpoint_dir"] = str(tmp_path / "run")
+
+        setup_single_controller(mc, tokenizer=MagicMock(pad_token_id=0))
+
+        value_kwargs = patched_ppo_factories["_build_value"].call_args.kwargs
+        assert value_kwargs["weights_path"] is None
+        assert value_kwargs["optimizer_path"] is None
```
yuki-97, line 619: [`tests/unit/single_controller/test_ppo_setup.py:512`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/tests/unit/single_controller/test_ppo_setup.py#L512)

**TL;DR — resume-wins is the PR's load-bearing invariant and nothing tests it; both new tests stay green if the operands are swapped, which is exactly the regression that would silently discard a critic's progress on every resume.**

The description makes the precedence the reason the key can stay in the config across resubmits — "making the warm start win instead would silently discard the critic's progress on every resume". But both new tests run with a `checkpoint_dir` that does not exist, so `last_checkpoint_path` is `None` and `None or warm_start` is indistinguishable from `warm_start or None`. Rewrite [`:651`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/algorithms/single_controller_utils/setup.py#L651) as `warm_start or last_checkpoint_path` and the suite still passes. The same expression at [`ppo.py:713`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/algorithms/ppo.py#L713) is a second copy with no coverage at all, so a regression in either is invisible to the other.

**Action:** add a resume case. Verified against source — `get_latest_checkpoint_path` globs `step_*` without consulting `checkpointing.enabled` ([`checkpoint.py:560`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/checkpoint.py#L560)); `load_training_info` opens `training_info.json` unguarded ([`:584`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/nemo_rl/utils/checkpoint.py#L584)); and `load_dataloader_state` needs patching because `patched_ppo_factories` does not cover it and it `torch.load`s `train_dataloader.pt`. `sc_setup_mod` and `patch` are already imported ([`:25,29`](https://github.com/NVIDIA-NeMo/RL/blob/6adf1a46d0c07c10e83dee0b1ef36c7e0335224b/tests/unit/single_controller/test_ppo_setup.py#L25)). Not executed here — no runnable env on this host; please run `uv run --group test pytest tests/unit/single_controller/test_ppo_setup.py`.

```suggestion
        assert value_kwargs["optimizer_path"] is None

    def test_resume_ignores_the_warm_start(self, patched_ppo_factories, tmp_path):
        """The run's own checkpoint wins: re-seeding would discard the critic's progress."""
        seed = tmp_path / "critic_pretrain" / "step_370"
        (seed / "value" / "weights").mkdir(parents=True)
        (seed / "value" / "optimizer").mkdir()
        step_5 = tmp_path / "run" / "step_5"
        for component in ("policy", "value"):
            (step_5 / component / "weights").mkdir(parents=True)
            (step_5 / component / "optimizer").mkdir()
        (step_5 / "training_info.json").write_text("{}")
        mc = _ppo_master_config(
            ppo=PPOConfig.model_construct(
                max_num_steps=100,
                warm_start_value_checkpoint=str(seed),
                **_STEP_CONFIG,
            )
        )
        mc.checkpointing["checkpoint_dir"] = str(tmp_path / "run")

        with patch.object(sc_setup_mod, "load_dataloader_state"):
            setup_single_controller(mc, tokenizer=MagicMock(pad_token_id=0))

        value_kwargs = patched_ppo_factories["_build_value"].call_args.kwargs
        assert value_kwargs["weights_path"] == step_5 / "value" / "weights"
        assert value_kwargs["optimizer_path"] == step_5 / "value" / "optimizer"
```

yuki-97 (commented):
Team review of the critic warm-start path, run across five agents (RL/guidelines, bug-finder, test, design, devil's advocate) with an adversarial pass over every finding.

15 raw findings deduped to 11 clusters; 4 survived confidence + impact filtering. Seven were dropped as confirmed-but-not-worth-a-round-trip, and one (a claim that the new test fixture models an unreachable checkpoint layout) was disputed outright — `get_resume_paths` returns the identical tuple on both branches, so the fixture is fine.

No lint/type results: this host has no `uv` or `pre-commit`, and nothing was executed except two standalone stdlib repros. Test suggestions are unverified locally and flagged as such.

_Generated by Claude Code_

examples/configs/ppo_math_1B.yaml
yuki-97: fixed in f4077fa9c19fc3fe6c726c98cbed69fdc32c65fa

tests/unit/single_controller/test_ppo_setup.py
yuki-97: fixed in f4077fa9c19fc3fe6c726c98cbed69fdc32c65fa

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: fixed in c14539c6729d492da9cd02f93d078eb9dbaafd1d

nemo_rl/algorithms/single_controller_utils/setup.py
yuki-97: Intended — the warm start is a resume by design, and carrying the seed's optimizer moments and LR-scheduler state over is the point: it lets the post-warmup LR continue instead of restarting. `prep_warm_start.sh` in the SWE pipeline does the same thing with a fabricated `step_0`; this key just replaces the shell script, so forcing `override_opt_param_scheduler` would defeat it.

Recorded in the docs instead — `docs/guides/ppo.md` now states that the seed and the run must agree on `value.megatron_cfg.scheduler` and on the scheduler-tick budget, shows the assertion you would hit, and points at `value.megatron_cfg.scheduler.override_opt_param_scheduler: true` for anyone who wants this run's schedule to win. The unconditional "restores its weights, optimizer moments and LR-scheduler state" is fixed there too.

documented in c14539c6729d492da9cd02f93d078eb9dbaafd1d

docs/guides/ppo.md
```diff
@@ -202,6 +202,30 @@ ppo:
...
+
+**The seed and this run must agree on the value scheduler.** The seed is restored through the ordinary resume path, so Megatron checks its LR-scheduler state field by field against the one this run builds and raises on the first mismatch — `use_checkpoint_opt_param_scheduler` is off, so `OptimizerParamScheduler._check_and_set` asserts equality. Keep `value.megatron_cfg.scheduler` identical across the two runs, and give them the same scheduler-tick budget: `wd_incr_steps` is derived from `train_iters`, which PPO sets to `max_num_steps × ppo_epochs`, so a critic pretrained under a short budget and reused under a long one fails during critic init with
+
+```
+AssertionError: OptimizerParamScheduler: class input value <X> and checkpoint value <Y> for total number of weight decay iterations do not match
+```
+
+Carrying the seed's schedule over is deliberate — it is what lets the post-warmup LR continue instead of restarting. Set `value.megatron_cfg.scheduler.override_opt_param_scheduler: true` if you would rather this run's schedule win; the seeded LR is then discarded along with the mismatch.
```
terrykong: [`docs/guides/ppo.md:219`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/docs/guides/ppo.md#L219)

**TL;DR — the "keep identical" list leaves out the two settings Megatron checks _first_, so someone can follow this paragraph exactly and still fail at critic init.**

Megatron compares nine scheduler fields on restore and raises on the first mismatch ([`optimizer_param_scheduler.py:341-404`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/optimizer_param_scheduler.py#L341-L404)). Three corrections:

1. **`max_lr`/`min_lr` are checked first and live in the optimizer block**, not the scheduler block ([`optim.py:342-343`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/training/optim.py#L342-L343)). A critic pretrained at a different LR fails on `learning rate` before `scheduler` is read.
2. **`value.train_global_batch_size` multiplies three of the compared fields** ([`config.py:1778-1788`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8c46dc4259080c510b7455f43e836fdff222c5d3/src/megatron/bridge/training/config.py#L1778-L1788)), via the value worker's [`validate()`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/models/value/workers/megatron_value_worker.py#L406). Same scheduler block and same `train_iters` still assert if batch size differs.
3. **On a non-async run `train_iters` is clamped** to `min(max_num_steps, max_num_epochs × len(dataloader)) × ppo_epochs` ([`ppo.py:733-741`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/algorithms/ppo.py#L733-L741); SC clamps in place at [`setup.py:702`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/algorithms/single_controller_utils/setup.py#L702)). Four of the nine PPO recipes under `examples/configs/recipes/llm/` set `max_num_epochs: 15`, so for those the dataset size and `num_prompts_per_step` are in the budget too and matching `max_num_steps` alone is not enough. Async is the exception — it must set `max_num_epochs: -1` and takes the unclamped branch, where the current wording is correct.

Also: the quoted error is not verbatim — the f-strings join with no space, so it prints `checkpointvalue` ([`:333-336`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/optimizer_param_scheduler.py#L333-L336)) — and "the seeded LR is then discarded" overstates the override; [`self.step(increment=num_steps)`](https://github.com/NVIDIA/Megatron-LM/blob/14346b65a2d0790e451919858f7771078105c5f0/megatron/core/optimizer_param_scheduler.py#L390) still restores the seed's position.

**Action:** replace the paragraph with the suggestion, which also scopes it to Megatron — [`ppo_math_1B.yaml:320-322`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/examples/configs/ppo_math_1B.yaml#L320-L322) makes DTensor the default critic, and that path adopts the seed's schedule unchecked.

````suggestion
**The seed and this run must agree on the value scheduler.** The seed is restored through the ordinary resume path, so Megatron compares nine scheduler fields against the ones this run builds and raises on the first mismatch — `use_checkpoint_opt_param_scheduler` is off, so `OptimizerParamScheduler._check_and_set` asserts equality. Match all of these across the two runs:

- `value.megatron_cfg.optimizer.lr` and `.min_lr` — they feed `max_lr`/`min_lr` and are the *first* two fields checked. They live in the optimizer block, not the scheduler block.
- `value.megatron_cfg.scheduler`.
- `value.train_global_batch_size` — it multiplies `lr_decay_steps`, `wd_incr_steps` and `lr_warmup_steps`.
- the tick budget `train_iters`. A synchronous run sets it to `min(max_num_steps, max_num_epochs × len(dataloader)) × ppo_epochs`; an async run sets it to `max_num_steps × ppo_epochs`, since async requires `max_num_epochs: -1`. `len(dataloader)` is prompt batches per epoch, so on a synchronous run the dataset size and `num_prompts_per_step` are part of the budget whenever the epoch term is the smaller one — as it is for the shipped recipes that set `max_num_epochs: 15`. Matching `max_num_steps` and `ppo_epochs` alone is not enough there.

A mismatch fails during critic init. Which field is named depends on which input differs: a batch-size difference reports `warmup iterations`, a learning-rate difference reports `learning rate`.

```
AssertionError: OptimizerParamScheduler: class input value <X> and checkpointvalue <Y> for total number of weight decay iterations do not match
```

(`checkpointvalue` runs together in the upstream message; search for it as written.)

Carrying the seed's schedule over is deliberate — it is what lets the post-warmup LR continue instead of restarting. Set `value.megatron_cfg.scheduler.override_opt_param_scheduler: true` if you would rather this run's settings win; the seed's step position is still restored, so the critic resumes at the seed's tick count rather than at step 0. All of this is Megatron-only — a DTensor critic (the default in `ppo_math_1B.yaml`) loads the seed's scheduler state with no comparison and has no override knob.
````

File: nemo_rl/algorithms/ppo.py [hunk 6]
```diff
@@ -704,8 +712,14 @@ def setup(
     worker_init_timing_metrics = {}
 
     weights_path, optimizer_path = checkpointer.get_resume_paths(last_checkpoint_path)
+    # Only a fresh run reads this; a resume ignores it and restores the critic from
+    # its own checkpoint, so the key can stay in the config.
+    warm_start = ppo_config.warm_start_value_checkpoint
```
terrykong, line 718: [`nemo_rl/algorithms/ppo.py:717`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/algorithms/ppo.py#L717)

**TL;DR — this copy of the warm start has no test at all; the 88 new test lines cover only the SingleController copy, and this is the path the exemplar that received the new key actually runs.**

`grep -rn "warm_start_value_checkpoint" tests/` returns only the reference config and `tests/unit/single_controller/test_ppo_setup.py`. So the validator call, the 🔥 print and the `last_checkpoint_path or warm_start` precedence at [`ppo.py:715-724`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/algorithms/ppo.py#L715-L724) are uncovered, while the near-identical SC block has four tests. The key was added to [`examples/configs/ppo_math_1B.yaml:13-16`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/examples/configs/ppo_math_1B.yaml#L13-L16), which runs through *this* path via [`examples/run_ppo.py:163`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/examples/run_ppo.py#L163).

Since the two copies are deliberately duplicated rather than shared, a regression in either is invisible to the other — which is exactly what the earlier review comment on this PR meant by "a second copy with no coverage at all". That half of it is still open.

The harness already exists and provably reaches this code: [`_run_noncolocated_setup`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/tests/unit/algorithms/test_ppo.py#L1437) runs the real `ppo.setup` on CPU with `Value` monkeypatched, and [`test_megatron_train_iters_matches_ppo_training_limit`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/tests/unit/algorithms/test_ppo.py#L1943) already asserts on `value["megatron_cfg"]["train_iters"]`, which `ppo.py` sets *after* this block — so setup runs straight through it today.

**Action:** add the negative case to `tests/unit/algorithms/test_ppo.py`. It needs no harness change, because `DummyCheckpointer.get_latest_checkpoint_path` already returns `None` (the fresh-run shape):

```python
def test_ppo_setup_rejects_a_warm_start_that_does_not_resolve(monkeypatch, tmp_path):
    config = _make_noncolocated_setup_config()
    config.ppo.warm_start_value_checkpoint = str(tmp_path / "typo")

    with pytest.raises(ValueError, match="would silently start cold"):
        _run_noncolocated_setup(monkeypatch, config)
```

Not a `suggestion` block — it is a new test in a file this diff does not touch. A positive "the seed path reached the value factory" assertion would additionally need [`DummyCheckpointer.get_resume_paths`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/tests/unit/algorithms/test_ppo.py#L1409) to record its argument instead of returning `None, None`; worth having, but the negative test is the zero-friction one.

I could not run this — macOS, no GPU, and this tree's `uv.lock` does not resolve (`Failed to parse entry: nemo-gym`), so please verify it locally.

nemo_rl/utils/checkpoint.py
```diff
@@ -61,6 +61,30 @@ def _load_megatron_common_state_dict(iteration_dir: Path) -> dict[str, Any]:
...
+    get_resume_paths never stats it -- an unresolvable path would train a cold
+    model behind a line claiming a warm start. Callers must gate this on a fresh
+    run, since a resume ignores the warm start entirely.
+
+    Args:
+        warm_start: step_<n> directory the user pointed the warm start at. A Hydra
+            override written as ``key=`` with an unset variable arrives as "".
+        model_component: Subtree the seed must carry, matching get_resume_paths.
+    """
+    if not (Path(warm_start) / model_component / "weights").exists():
```
terrykong, line 85: [`nemo_rl/utils/checkpoint.py:79`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/utils/checkpoint.py#L79)

**1 action item.** PR-introduced, low severity.

**TL;DR — the empty-string case this docstring calls out is caught by accident, not by design: `Path("")` is `Path(".")`, so the check depends on what the launch directory happens to contain.**

`Path("") / "value" / "weights"` is the *relative* path `value/weights`, so for `warm_start=""` this line stats the launch directory. I reproduced it: from a directory that contains `value/weights`, the check returns `True`, validation passes, and then `last_checkpoint_path or warm_start` is falsy — so [`get_resume_paths("")`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/utils/checkpoint.py#L237) returns `(None, None)` and the run prints the 🔥 banner over a cold critic, which is the exact outcome this function was added to prevent.

That cwd is contrived and I am not claiming a user will hit it. Two things are not contrived:

- The guard tests identity ([`warm_start is not None`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/algorithms/ppo.py#L718)) while the resolver tests truthiness ([`last_checkpoint_path or warm_start`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/nemo_rl/algorithms/ppo.py#L722)). `""` passes the first and is discarded by the second, so the validator is the only thing standing between them.
- [`test_rejects_a_warm_start_that_does_not_resolve[empty_string]`](https://github.com/NVIDIA-NeMo/RL/blob/6e8790cf9f7746fbee021a8e29183f3feba6cbf4/tests/unit/single_controller/test_ppo_setup.py#L619) passes only because pytest's cwd has no `value/weights`. It asserts on ambient state rather than on behavior — the `typo` parametrization is hermetic, this one is not.

**Action:** reject the empty value explicitly, in the validator so both call sites get it. The test then passes for the reason its docstring gives, with no test change needed.

```suggestion
    if not str(warm_start).strip():
        raise ValueError(
            "warm_start_value_checkpoint is empty. A Hydra override written as "
            "`ppo.warm_start_value_checkpoint=` with an unset variable produces "
            "this; point it at a step_<n> directory or drop the override."
        )
    if not (Path(warm_start) / model_component / "weights").exists():
```

terrykong (commented):
Team review of the critic warm-start path: seven agents (RL/guidelines, Megatron-Bridge, bug-finder, test, design, comment-audit, devil's advocate), with an adversarial pass over every finding including my own.

Nice feature — moving critic warmup off the PPO allocation is a real saving where generation dominates, and the resume-wins precedence is the right call.

**Three comments below.** Eleven candidate findings were cut to three: four were disputed outright by the adversarial pass, and the rest folded into the surviving comments. Two design calls are worth affirming rather than questioning:

- **Routing the seed through the resume path instead of `checkpointing.pretrained_checkpoint` is correct, and non-obviously so.** That path auto-sets `finetune=True` in megatron-bridge, which suppresses optimizer/RNG state *and* triggers the value worker's `hide_loss_modules` hook that drops `output_layer.*` — so it would restore the trunk and randomly re-initialize the value head, discarding exactly what critic pretraining produces. `megatron_value_worker.py:97` already states the rule.
- **Gating value-path resolution on `is_ppo_run` incidentally fixes a pre-existing GRPO wart** — a GRPO resume previously stat'd `step_N/value/optimizer`, missed, and emitted a spurious "Optimizer state not found" warning about a value model the run does not have. Worth a line in the PR body, since it is a behavior change on the GRPO path in a PR titled `feat(ppo)`.

**Not verified — no runnable environment.** Everything here is static reading plus upstream source at the pinned SHAs (Megatron-LM `14346b65`, Megatron-Bridge `8c46dc42`). No test and no linter was executed: this tree's `uv.lock` is linux-only and `uv run` / `pre-commit` both fail on macOS with `Failed to parse entry: nemo-gym`. I make no claim about lint status, and the suggested test is unrun.

FYI, no action: the PR description's "How it works" snippet is stale — it still shows the call site before `validate_warm_start_checkpoint` was added in c14539c6.

_Generated by Claude Code_

docs/guides/ppo.md
yuki-97: fixed in 95c03dc59a4780258229d2896f9326318b24275d

nemo_rl/algorithms/ppo.py
yuki-97: fixed in 95c03dc59a4780258229d2896f9326318b24275d

nemo_rl/utils/checkpoint.py
yuki-97: fixed in 95c03dc59a4780258229d2896f9326318b24275d

terrykong (approved):

PR #3829, opened 2026-08-25
feat(ppo): decouple actor and critic training epochs

bg51717:
# What does this PR do ?

Adds an independent `ppo.critic_ppo_epochs` setting while preserving the existing behavior by defaulting it to `ppo_epochs`.

Critic updates are grouped before policy updates so each colocated model stays resident across its epochs instead of being repeatedly moved between CPU and GPU. The policy and critic Megatron scheduler budgets are sized independently.

# Issues

None closed.

# Usage

```yaml
ppo:
  ppo_epochs: 2
  critic_ppo_epochs: 3
```

**Pre checks**:
- [x] Did you add or update any necessary documentation?

# Additional Information

* Tests: 177 targeted unit tests passed with 7 GPU-only skips.
* A two-step GB300 functional smoke completed with `ppo_epochs=2` and `critic_ppo_epochs=3`.

### DIFF
--- a/docs/guides/ppo.md
+++ b/docs/guides/ppo.md
@@ -187,12 +187,13 @@ Unlike GRPO, which performs one training update per rollout, PPO can perform mul
 ```yaml
 ppo:
   ppo_epochs: 4          # actor passes over each rollout batch
-  critic_ppo_epochs: 6  # critic passes; null uses ppo_epochs
+  critic_ppo_epochs: ${ppo.ppo_epochs}  # critic passes; follows actor by default
 ```
 
 Each pass uses the same returns and advantage estimates computed from the initial
-rollout. `critic_ppo_epochs` must be greater than or equal to `ppo_epochs`, so the
-critic can be fit harder without adding actor updates.
+rollout. Both epoch counts must be at least 1 and can be configured independently;
+the exemplar uses interpolation so the critic follows the actor unless explicitly
+overridden.
 
 ### Critic Warmup
 
@@ -273,7 +274,7 @@ ppo:
   max_num_epochs: 100000
   max_num_steps: 100000
   ppo_epochs: 4
-  critic_ppo_epochs: null
+  critic_ppo_epochs: ${ppo.ppo_epochs}
   policy_training_start_step: 0
   warm_start_value_checkpoint: null
   val_period: 20
@@ -331,7 +332,7 @@ value_loss_fn:
 
 **PPO-specific parameters:**
 - **`ppo.ppo_epochs`**: Number of actor training updates per rollout batch
-- **`ppo.critic_ppo_epochs`**: Number of critic training updates per rollout batch. `null` uses `ppo_epochs`; an explicit value must be greater than or equal to it.
+- **`ppo.critic_ppo_epochs`**: Number of critic training updates per rollout batch. It can differ from `ppo_epochs`; the exemplar defaults it to `${ppo.ppo_epochs}`.
 - **`ppo.policy_training_start_step`**: Number of critic-only warmup steps before policy training begins
 - **`ppo.warm_start_value_checkpoint`**: Checkpoint step directory whose `value/` seeds the critic on a fresh run. See [Warm-Starting the Critic](#warm-starting-the-critic)
 - **`ppo.seq_logprob_error_threshold`**: Nullable sequence-level multiplicative probability-error threshold. PPO always logs sequence-level train/generation mismatch metrics; when this is set, sequences above the threshold are excluded from advantage and loss computation.
--- a/docs/guides/ppo.md
+++ b/docs/guides/ppo.md
@@ -175,10 +175,10 @@ The PPO training loop, [ppo_train](../../nemo_rl/algorithms/ppo.py), follows thi
 3. **Value inference**: the value model predicts per-token state values
 4. **Logprob computation**: the policy computes log probabilities for advantage estimation
 5. **Advantage estimation**: GAE computes advantages using value predictions and rewards
-6. **Value training**: the critic is updated first (critic-before-actor, following [veRL](https://arxiv.org/abs/2412.09613))
-7. **Policy training**: the actor is updated with the clipped surrogate objective
+6. **Value training**: the critic completes all of its updates first
+7. **Policy training**: the actor completes all of its updates with the clipped surrogate objective
 
-Steps 6–7 repeat `ppo_epochs` times per rollout before generating new responses.
+The critic stays resident for all `critic_ppo_epochs` updates, then the policy stays resident for all `ppo_epochs` updates. This avoids moving the colocated models between CPU and GPU after every epoch.
 
 ### Multiple Training Steps per Rollout
 
@@ -186,10 +186,13 @@ Unlike GRPO, which performs one training update per rollout, PPO can perform mul
 
 ```yaml
 ppo:
-  ppo_epochs: 4   # Train 4 times on each rollout batch
+  ppo_epochs: 4          # actor passes over each rollout batch
+  critic_ppo_epochs: 6  # critic passes; null uses ppo_epochs
 ```
 
-Each step trains both the critic and the actor on the same advantage estimates computed from the initial rollout.
+Each pass uses the same returns and advantage estimates computed from the initial
+rollout. `critic_ppo_epochs` must be greater than or equal to `ppo_epochs`, so the
+critic can be fit harder without adding actor updates.
 
 ### Critic Warmup
 
@@ -270,6 +273,7 @@ ppo:
   max_num_epochs: 100000
   max_num_steps: 100000
   ppo_epochs: 4
+  critic_ppo_epochs: null
   policy_training_start_step: 0
   warm_start_value_checkpoint: null
   val_period: 20
@@ -326,7 +330,8 @@ value_loss_fn:
 ```
 
 **PPO-specific parameters:**
-- **`ppo.ppo_epochs`**: Number of training updates per rollout batch
+- **`ppo.ppo_epochs`**: Number of actor training updates per rollout batch
+- **`ppo.critic_ppo_epochs`**: Number of critic training updates per rollout batch. `null` uses `ppo_epochs`; an explicit value must be greater than or equal to it.
 - **`ppo.policy_training_start_step`**: Number of critic-only warmup steps before policy training begins
 - **`ppo.warm_start_value_checkpoint`**: Checkpoint step directory whose `value/` seeds the critic on a fresh run. See [Warm-Starting the Critic](#warm-starting-the-critic)
 - **`ppo.seq_logprob_error_threshold`**: Nullable sequence-level multiplicative probability-error threshold. PPO always logs sequence-level train/generation mismatch metrics; when this is set, sequences above the threshold are excluded from advantage and loss computation.
--- a/docs/guides/ppo.md
+++ b/docs/guides/ppo.md
@@ -59,7 +59,7 @@ When only one node remains for policy and generation after other resources are r
 
 ### Asynchronous PPO
 
-Set `ppo.async_ppo.enabled: true` to overlap rollout generation with training. A background collector fills a replay buffer on the non-colocated vLLM GPUs while the policy and value model train on their shared cluster. Values and policy/reference log probabilities are recomputed when a trajectory is sampled, then PPO runs GAE and its normal `ppo_epochs` updates before publishing one new policy version to vLLM.
+Set `ppo.async_ppo.enabled: true` to overlap rollout generation with training. A background collector fills a replay buffer on the non-colocated vLLM GPUs while the policy and value model train on their shared cluster. Values and policy/reference log probabilities are recomputed when a trajectory is sampled, then PPO runs GAE, all `critic_ppo_epochs` critic updates, and all `ppo_epochs` policy updates before publishing one new policy version to vLLM.
 
 Async PPO reuses the trajectory collector, replay buffer, and weight-versioning infrastructure described in the [Async GRPO guide](async-grpo.md); this section focuses on PPO-specific behavior and constraints.
 
--- a/docs/guides/ppo.md
+++ b/docs/guides/ppo.md
@@ -225,7 +225,7 @@ The path is a `step_<n>` directory holding a `value/` subtree — the layout a P
 - `value.megatron_cfg.optimizer.lr` and `.min_lr` — they feed `max_lr`/`min_lr` and are the *first* two fields checked. They live in the optimizer block, not the scheduler block.
 - `value.megatron_cfg.scheduler`.
 - `value.train_global_batch_size` — it multiplies `lr_decay_steps`, `wd_incr_steps` and `lr_warmup_steps`.
-- the tick budget `train_iters`. A synchronous run sets it to `min(max_num_steps, max_num_epochs × len(dataloader)) × ppo_epochs`; an async run sets it to `max_num_steps × ppo_epochs`, since async requires `max_num_epochs: -1`. `len(dataloader)` is prompt batches per epoch, so on a synchronous run the dataset size and `num_prompts_per_step` are part of the budget whenever the epoch term is the smaller one — as it is for the shipped recipes that set `max_num_epochs: 15`. Matching `max_num_steps` and `ppo_epochs` alone is not enough there.
+- the tick budget `train_iters`. A synchronous run sets it to `min(max_num_steps, max_num_epochs × len(dataloader)) × critic_ppo_epochs`; an async run sets it to `max_num_steps × critic_ppo_epochs`, since async requires `max_num_epochs: -1`. `len(dataloader)` is prompt batches per epoch, so on a synchronous run the dataset size and `num_prompts_per_step` are part of the budget whenever the epoch term is the smaller one — as it is for the shipped recipes that set `max_num_epochs: 15`. Matching `max_num_steps` and `critic_ppo_epochs` alone is not enough there.
 
 A mismatch fails during critic init. Which field is named depends on which input differs: a batch-size difference reports `warmup iterations`, a learning-rate difference reports `learning rate`.
 
--- a/examples/configs/ppo_math_1B.yaml
+++ b/examples/configs/ppo_math_1B.yaml
@@ -9,9 +9,8 @@ ppo:
   max_num_epochs: 100000
   max_num_steps: 100000
   ppo_epochs: 4  # actor passes over each rollout batch
-  # Critic passes over each rollout batch. null preserves the historical
-  # behavior by coupling this to ppo_epochs; an explicit value must be >= it.
-  critic_ppo_epochs: null
+  # Critic passes over each rollout batch; follows ppo_epochs unless overridden.
+  critic_ppo_epochs: ${ppo.ppo_epochs}
   policy_training_start_step: 0  # number of PPO steps of critic-only warmup before policy training begins
   # step_<n> dir of a critic-pretrain run whose value/ seeds the critic.
   # Only a fresh run reads it; a resume ignores it and restores the critic from
--- a/examples/configs/ppo_math_1B.yaml
+++ b/examples/configs/ppo_math_1B.yaml
@@ -8,7 +8,10 @@ ppo:
   max_rollout_turns: 1
   max_num_epochs: 100000
   max_num_steps: 100000
-  ppo_epochs: 4
+  ppo_epochs: 4  # actor passes over each rollout batch
+  # Critic passes over each rollout batch. null preserves the historical
+  # behavior by coupling this to ppo_epochs; an explicit value must be >= it.
+  critic_ppo_epochs: null
   policy_training_start_step: 0  # number of PPO steps of critic-only warmup before policy training begins
   # step_<n> dir of a critic-pretrain run whose value/ seeds the critic.
   # Only a fresh run reads it; a resume ignores it and restores the critic from
--- a/nemo_rl/algorithms/ppo.py
+++ b/nemo_rl/algorithms/ppo.py
@@ -763,7 +763,7 @@ def setup(
 
     # train_iters is the total scheduler-tick budget. Each Megatron worker
     # ticks once per train() call, so policy and value need separate budgets
-    # when the critic trains for more epochs than the actor.
+    # when their epoch counts or training start steps differ.
     ppo_epochs = ppo_config.ppo_epochs
     critic_ppo_epochs = ppo_config.resolved_critic_ppo_epochs
     async_config = ppo_config.async_ppo
@@ -775,7 +775,16 @@ def setup(
             ppo_config.max_num_epochs * len(dataloader),
         )
     if policy_config.get("megatron_cfg", {}).get("enabled", False):
-        policy_config["megatron_cfg"]["train_iters"] = outer_training_steps * ppo_epochs
+        policy_training_steps = max(
+            outer_training_steps - ppo_config.policy_training_start_step,
+            0,
+        )
+        # Megatron-Bridge requires a positive scheduler horizon at setup. The
+        # scheduler is never advanced when critic warmup spans the whole run.
+        policy_config["megatron_cfg"]["train_iters"] = max(
+            policy_training_steps * ppo_epochs,
+            1,
+        )
 
     if value_config.get("megatron_cfg", {}).get("enabled", False):
         value_config["megatron_cfg"]["train_iters"] = (
--- a/nemo_rl/algorithms/ppo.py
+++ b/nemo_rl/algorithms/ppo.py
@@ -192,11 +192,8 @@ class PPOConfig(BaseModel, extra="allow"):
     batch_multiplier: float = 1.0
     # Number of actor (policy) passes over each rollout batch.
     ppo_epochs: int = 4
-    # Number of critic (value) passes over each rollout batch. None preserves
-    # the historical behavior by coupling it to ppo_epochs. An explicit value
-    # may only increase the number of critic passes, never reduce it below the
-    # actor's count.
-    critic_ppo_epochs: int | None = None
+    # Number of critic (value) passes over each rollout batch.
+    critic_ppo_epochs: int = 4
     reward_shaping: RewardShapingConfig = Field(default_factory=RewardShapingConfig)
     reward_scaling: RewardScalingConfig = Field(default_factory=RewardScalingConfig)
     adv_estimator: GAEConfig = Field(default_factory=GAEConfig)
@@ -220,9 +217,16 @@ class PPOConfig(BaseModel, extra="allow"):
     # Legacy async config block; SC reads its async knobs from `async_rl` instead.
     async_ppo: AsyncPPOConfig | None = Field(default_factory=AsyncPPOConfig)
 
+    def validate_epoch_settings(self) -> None:
+        """Validate the actor and critic update counts."""
+        if self.ppo_epochs < 1:
+            raise ValueError("ppo.ppo_epochs must be at least 1")
+        if self.critic_ppo_epochs < 1:
+            raise ValueError("ppo.critic_ppo_epochs must be at least 1")
+
     @model_validator(mode="after")
     def validate_settings(self) -> "PPOConfig":
-        _ = self.resolved_critic_ppo_epochs
+        self.validate_epoch_settings()
         if (
             self.async_ppo is not None
             and self.async_ppo.enabled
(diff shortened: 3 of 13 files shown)


bg51717:
/ok to test [b008666](https://github.com/NVIDIA-NeMo/RL/commit/b0086668a0da8287d6d6563d47cd759660598135)

tests/unit/algorithms/test_ppo.py
```diff
@@ -1067,6 +1096,46 @@ def test_ppo_train_noncolocated_refit_offload_lifecycle(monkeypatch):
...
+    policy_prep_index = harness.events.index("policy_train_prep")
+    policy_train_indices = [
+        index for index, event in enumerate(harness.events) if event == "policy_train"
+    ]
+    assert harness.events[policy_prep_index : policy_train_indices[-1] + 1] == [
+        "policy_train_prep",
+        "policy_train",
+        "policy_train",
+    ]
+
```
bg51717, line 1153: **[TEST]** [`tests/unit/algorithms/test_ppo.py:1137`](https://github.com/NVIDIA-NeMo/RL/blob/b0086668a0da8287d6d6563d47cd759660598135/tests/unit/algorithms/test_ppo.py#L1137)

`critic_ppo_epochs` and `policy_training_start_step` are each covered individually (see the neighboring tests), but there's no test pinning down their interaction at the mock-call-order level. This PR is what makes the critic loop unconditional while the actor loop stays gated behind warmup — exactly the kind of seam a future refactor could silently re-couple. Suggested addition (verified by hand-tracing `ppo_train`/`async_ppo_train` against `_run_mock_ppo_train`, not executed):

```python
@pytest.mark.parametrize("async_mode", [False, True])
def test_ppo_train_critic_keeps_extra_epochs_during_policy_warmup(
    monkeypatch, async_mode
):
    harness = _run_mock_ppo_train(
        monkeypatch,
        async_mode=async_mode,
        max_num_steps=2,
        ppo_epochs=2,
        critic_ppo_epochs=3,
        seq_logprob_error_threshold=None,
        policy_training_start_step=1,
    )

    # Step 0 is critic-only warmup; step 1 trains both. The critic always
    # runs all 3 epochs, independent of the policy warmup gate.
    assert harness.value_model.train.call_count == 6
    assert harness.policy.train.call_count == 2
    assert harness.value_model.prepare_for_training.call_count == 2
    assert harness.policy.prepare_for_training.call_count == 1
```

tests/unit/algorithms/test_ppo.py
```diff
@@ -645,6 +645,26 @@ def test_ppo_schema_rejects_unsupported_estimator_name():
...
+    config = PPOConfig(ppo_epochs=1, critic_ppo_epochs=3)
+
+    assert config.resolved_critic_ppo_epochs == 3
+
+
+def test_ppo_config_rejects_fewer_critic_epochs_than_actor_epochs():
+    from pydantic import ValidationError
+
+    with pytest.raises(ValidationError, match="critic_ppo_epochs"):
+        PPOConfig(ppo_epochs=2, critic_ppo_epochs=1)
```
bg51717: **[TEST]** [`tests/unit/algorithms/test_ppo.py:665`](https://github.com/NVIDIA-NeMo/RL/blob/b0086668a0da8287d6d6563d47cd759660598135/tests/unit/algorithms/test_ppo.py#L665)

`resolved_critic_ppo_epochs`'s `ppo_epochs < 1` guard is currently only exercised via `PPOConfig.model_construct(...)` in the single-controller tests, which bypasses pydantic validation entirely. This PR is what newly centralizes the check into `PPOConfig`'s own `model_validator` (`validate_settings`), so the path users actually hit — the normal constructor — isn't covered yet. Suggested addition, mirroring the sibling test right above it:

```python
def test_ppo_config_rejects_zero_ppo_epochs():
    from pydantic import ValidationError

    with pytest.raises(ValidationError, match="ppo_epochs must be at least 1"):
        PPOConfig(ppo_epochs=0)
```

nemo_rl/algorithms/single_controller.py
```diff
@@ -1932,19 +1939,23 @@ async def _value_stage(self, meta: KVBatchMeta) -> KVBatchMeta:
         await asyncio.to_thread(self._value.finish_inference)
         return meta.with_fields([self._advantage_cfg.values_field])
 
-    async def _value_train(self, meta: KVBatchMeta) -> dict[str, Any]:
-        """Run one value model optimizer step against this chunk's GAE returns.
+    async def _value_train_epochs(
+        self, meta: KVBatchMeta, *, num_epochs: int
+    ) -> dict[str, Any]:
+        """Run consecutive critic epochs under one model onload/offload cycle."""
```
bg51717: **[DOCSTRING]** [`nemo_rl/algorithms/single_controller.py:1942-1945`](https://github.com/NVIDIA-NeMo/RL/blob/b0086668a0da8287d6d6563d47cd759660598135/nemo_rl/algorithms/single_controller.py#L1942-L1945)

`_value_train_epochs` dropped the `Returns:` section that `_value_train` (the method it replaces) had, making it the outlier among its sibling private helpers in the same class — `_value_stage`, `_advantage_stage`, and `_sync_weights` all document their return value. `Args:` for the new `num_epochs` kwarg can stay omitted (siblings don't document Args either), just the Returns section:

```suggestion
        """Run consecutive critic epochs under one model onload/offload cycle.

        Returns:
            The aggregated value model train result, shaped like Value.train's
            (only the last epoch's result is kept).
        """
```

bg51717 (commented):
Reviewed with a 4-agent pass (rl-expert, test-agent, bug-finder, devil's-advocate) covering config validation fail-fast paths, GPU residency correctness across all 3 restructured PPO epoch loops (sync/async/single-controller), 3-way consistency between them, and the two independent Megatron `train_iters` formulas (cross-checked against Megatron-Bridge's scheduler consumption at the pinned submodule SHA).

**No bugs found.** 4 low/medium-severity suggestions below (2 test-coverage gaps, 1 docstring gap, 1 process ask around memory-fragmentation evidence) survived an adversarial devil's-advocate pass that disputed and discarded a 5th (a doc-page omission that turned out to match that page's own established convention).

_Generated by Claude Code_

bg51717:
/ok to test [a39596a](https://github.com/NVIDIA-NeMo/RL/pull/3829/commits/a39596a8029b874a6ea6c35c74b2338382d76795)

tests/unit/algorithms/test_ppo.py
bg51717: Added in a39596a8 as suggested. Counts verified locally.

tests/unit/algorithms/test_ppo.py
bg51717: Added in a39596a8 as suggested.

nemo_rl/algorithms/single_controller.py
bg51717: Added in a39596a8. Reworded slightly: result is overwritten each epoch, not aggregated, and the call is train_from_meta.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -571,23 +571,33 @@ def _clamp_max_num_steps(
...
+    policy_epochs = ppo_config.ppo_epochs if ppo_config is not None else 1
+    policy_training_steps = algo_cfg.max_num_steps
+    if ppo_config is not None:
+        policy_training_steps = max(
+            policy_training_steps - ppo_config.policy_training_start_step,
+            0,
+        )
+    # Megatron-Bridge requires a positive scheduler horizon at setup. A PPO
+    # policy scheduler is never advanced when critic warmup spans the whole run.
+    policy_train_iters = max(policy_training_steps * policy_epochs, 1)
```
yuki-97: **[TEST]** This changes a shipped nightly's policy LR horizon, and the only evidence in the description is a two-step smoke.

The SC PPO nightly sets [`policy_training_start_step: 10`](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-2n8g-megatron-valuetp2sp-dynbatch-noncolocated-async-single-controller.yaml#L24) with [`ppo_epochs: 4`](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/examples/configs/recipes/llm/ppo-qwen2.5-1.5b-gsm8k-2n8g-megatron-valuetp2sp-dynbatch-noncolocated-async-single-controller.yaml#L23) and its [driver runs `MAX_STEPS=40`](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/tests/test_suites/llm/ppo-qwen2.5-1.5b-gsm8k-2n8g-megatron-valuetp2sp-dynbatch-noncolocated-async-single-controller.sh#L8), so the policy budget drops 160 -> 120 while the critic stays at 160 (`policy.megatron_cfg.scheduler.lr_warmup_iters: 0`, so this is pure decay horizon). It is in [nightly.txt](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/tests/test_suites/nightly.txt#L179) and gates on [`mean(data["train/reward"], range_start=-10) > 0.75`](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/tests/test_suites/llm/ppo-qwen2.5-1.5b-gsm8k-2n8g-megatron-valuetp2sp-dynbatch-noncolocated-async-single-controller.sh#L49).

**Suggest fix:** run that nightly and post the reward curve before taking this out of draft.

nemo_rl/algorithms/ppo.py
```diff
@@ -205,6 +212,21 @@ def validate_async_warmup_settings(self) -> "PPOConfig":
...
         return self
 
+    @property
+    def resolved_critic_ppo_epochs(self) -> int:
+        """Resolve and validate the number of critic passes per rollout batch."""
+        if self.ppo_epochs < 1:
+            raise ValueError("ppo.ppo_epochs must be at least 1")
+        if self.critic_ppo_epochs is None:
+            return self.ppo_epochs
+        if self.critic_ppo_epochs < self.ppo_epochs:
```
yuki-97: Nothing in the implementation requires `critic_ppo_epochs >= ppo_epochs`.

The critic and actor epoch loops ([ppo.py#L1633](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/nemo_rl/algorithms/ppo.py#L1633), [#L1664](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/nemo_rl/algorithms/ppo.py#L1664)) and their scheduler budgets ([setup.py#L586](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/nemo_rl/algorithms/single_controller_utils/setup.py#L586), [#L599](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/nemo_rl/algorithms/single_controller_utils/setup.py#L599)) are computed independently, so `critic_ppo_epochs=1` with `ppo_epochs=4` would run correctly. Each critic epoch is a full forward/backward, so a deliberately cheaper critic is as legitimate a config as a harder-fit one.

Either drop the check, or name in the message what it protects.

examples/configs/ppo_math_1B.yaml
```diff
@@ -8,7 +8,10 @@ ppo:
   max_rollout_turns: 1
   max_num_epochs: 100000
   max_num_steps: 100000
-  ppo_epochs: 4
+  ppo_epochs: 4  # actor passes over each rollout batch
+  # Critic passes over each rollout batch. null preserves the historical
+  # behavior by coupling this to ppo_epochs; an explicit value must be >= it.
+  critic_ppo_epochs: null
```
yuki-97: what about just default to `ppo_epochs` and always set it?
```suggestion
  critic_ppo_epochs: ${ppo.ppo_epochs}
```

nemo_rl/algorithms/ppo.py
```diff
@@ -177,7 +177,13 @@ class PPOConfig(BaseModel, extra="allow"):
     # When using dynamic sampling, generation prompt batch size will equal
     # num_prompts_per_step * batch_multiplier
     batch_multiplier: float = 1.0
+    # Number of actor (policy) passes over each rollout batch.
     ppo_epochs: int = 4
+    # Number of critic (value) passes over each rollout batch. None preserves
+    # the historical behavior by coupling it to ppo_epochs. An explicit value
+    # may only increase the number of critic passes, never reduce it below the
+    # actor's count.
+    critic_ppo_epochs: int | None = None
```
yuki-97: same here
```suggestion
    critic_ppo_epochs: int = 4
```

nemo_rl/algorithms/ppo.py
```diff
@@ -205,6 +212,21 @@ def validate_async_warmup_settings(self) -> "PPOConfig":
             )
         return self
 
+    @property
+    def resolved_critic_ppo_epochs(self) -> int:
```
yuki-97: this can also be removed after https://github.com/NVIDIA-NeMo/RL/pull/3829/changes#r3883214508.

nemo_rl/algorithms/ppo.py
```diff
@@ -711,12 +733,10 @@ def setup(
...
     # train_iters is the total scheduler-tick budget. Each Megatron worker
-    # ticks once per train() call (matching upstream main's per-rollout
-    # convention), and PPO calls each worker's train() `ppo_epochs` times
-    # per outer step. So total ticks = (outer steps) * ppo_epochs.
-    # Scale train_iters accordingly so the configured warmup/decay horizon
-    # matches the actual scheduler-step count.
+    # ticks once per train() call, so policy and value need separate budgets
+    # when their epoch counts or training start steps differ.
     ppo_epochs = ppo_config.ppo_epochs
+    critic_ppo_epochs = ppo_config.resolved_critic_ppo_epochs
```
yuki-97:
```suggestion
    critic_ppo_epochs = ppo_config.critic_ppo_epochs
```

nemo_rl/algorithms/single_controller.py
```diff
@@ -155,8 +156,16 @@ def __init__(
         self._algo_cfg = algo_config(master_config)
         self._async_cfg = master_config.async_rl
         self._is_ppo: bool = is_ppo_run(master_config)
-        # GRPO has no epoch knob: it makes one optimizer step per RL step.
-        self._ppo_epochs: int = self._algo_cfg.ppo_epochs if self._is_ppo else 1
```
yuki-97, line 204: what about just keep this line and add a same one for `_critic_ppo_epochs`?

nemo_rl/algorithms/single_controller.py
```diff
@@ -155,8 +156,16 @@ def __init__(
...
+        self._ppo_epochs: int
+        self._critic_ppo_epochs: int
+        if self._is_ppo:
+            assert master_config.ppo is not None
```
yuki-97: the assert here is no-op, this should already be guard at setup.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1932,19 +1939,28 @@ async def _value_stage(self, meta: KVBatchMeta) -> KVBatchMeta:
...
+        self, meta: KVBatchMeta, *, num_epochs: int
+    ) -> dict[str, Any]:
+        """Run consecutive critic epochs under one model onload/offload cycle.
 
         Returns:
-            The aggregated value model train result, shaped like Value.train's.
+            The final epoch's ``train_from_meta`` output; earlier epochs'
+            results are discarded.
         """
+        if num_epochs < 1:
```
yuki-97: let move this to setup.

docs/guides/ppo.md
```diff
@@ -59,7 +59,7 @@ When only one node remains for policy and generation after other resources are r
 
 ### Asynchronous PPO
```
yuki-97, line 60: One item outside the diff: [docs/about/algorithms/ppo.md#L71](https://github.com/NVIDIA-NeMo/RL/blob/a39596a8029b874a6ea6c35c74b2338382d76795/docs/about/algorithms/ppo.md#L71) still lists only ppo_epochs and policy_training_start_step under "Key PPO-specific parameters" -- adding critic_ppo_epochs: null there would keep it in sync with docs/guides/ppo.md.

yuki-97 (commented):
Reviewed the epoch-decoupling change across `ppo.py`, `single_controller.py`, the SC setup/config validators, the docs and the tests, along four lenses: correctness of the regrouped train loop, the scheduler-budget contract, config/doc parity, and test coverage. The regrouping itself looks numerically equivalent to the old interleaving (advantages and returns are computed once before the block and the policy loss never reads the critic), and the offload lifecycle is preserved.

bg51717:
/ok to test [bf66a15](https://github.com/NVIDIA-NeMo/RL/commit/bf66a15b53770131246ae9cc36c14d9da74e4323)

examples/configs/ppo_math_1B.yaml
bg51717: Makes sense. I'll use `${ppo.ppo_epochs}` as the recipe default and update the reference config, docs, and config-resolution coverage with it.

nemo_rl/algorithms/ppo.py
bg51717: Makes sense, I'll make `critic_ppo_epochs` a concrete `int`. One question: with `critic_ppo_epochs: int = 4`, the recipe path preserves the inherited behavior through `${ppo.ppo_epochs}`, but direct construction such as `PPOConfig(ppo_epochs=N)` no longer makes the critic follow `N`. Is the intent that this default coupling is guaranteed only by the exemplar config, or should direct `PPOConfig` construction preserve it as well?

nemo_rl/algorithms/ppo.py
bg51717: Agreed. With a concrete field, the property no longer adds value. I'll validate `ppo_epochs` and `critic_ppo_epochs` independently in `PPOConfig` (both must be at least 1), reuse the same validation helper from single-controller setup for `model_construct` callers, and replace the resolved-property call sites with direct field access.

nemo_rl/algorithms/ppo.py
bg51717: Agreed — I'll use the validated concrete field directly here and at the remaining legacy/single-controller call sites.

nemo_rl/algorithms/single_controller.py
bg51717: Yes, that is simpler. I'll restore the existing ternary assignment and add the matching `_critic_ppo_epochs` assignment, then verify the type check.

nemo_rl/algorithms/single_controller.py
bg51717: Agreed. `validate_single_controller_config` already establishes this invariant before these assignments, so I'll remove the duplicate assert.

nemo_rl/algorithms/single_controller.py
bg51717: Agreed. I'll enforce the epoch constraints in `PPOConfig` and single-controller setup validation (including the `model_construct` path), then remove the repeated helper-level check.

docs/guides/ppo.md
bg51717: Good catch. I'll add it there as well. With the interpolation change above, I'll show it as `${ppo.ppo_epochs}` and describe that it inherits the actor epoch count unless explicitly overridden, so the two docs stay consistent.

nemo_rl/algorithms/ppo.py
```diff
@@ -209,8 +212,16 @@ class PPOConfig(BaseModel, extra="allow"):
     # Legacy async config block; SC reads its async knobs from `async_rl` instead.
     async_ppo: AsyncPPOConfig | None = Field(default_factory=AsyncPPOConfig)
 
+    def validate_epoch_settings(self) -> None:
```
yuki-97: both have `validate_epoch_settings` and `validate_settings` seems not good. let's split to separate `@model_validator(mode="after")` to make it clear.
```suggestion
    @model_validator(mode="after")
    def validate_epoch(self) -> "PPOConfig":
```

nemo_rl/algorithms/ppo.py
```diff
@@ -209,8 +212,16 @@ class PPOConfig(BaseModel, extra="allow"):
...
+        """Validate the actor and critic update counts."""
+        if self.ppo_epochs < 1:
+            raise ValueError("ppo.ppo_epochs must be at least 1")
+        if self.critic_ppo_epochs < 1:
+            raise ValueError("ppo.critic_ppo_epochs must be at least 1")
+
     @model_validator(mode="after")
-    def validate_async_warmup_settings(self) -> "PPOConfig":
+    def validate_settings(self) -> "PPOConfig":
+        self.validate_epoch_settings()
```
yuki-97: remove `validate_epoch_settings` as https://github.com/NVIDIA-NeMo/RL/pull/3829/changes#r3891987530.
```suggestion
    def validate_async_warmup(self) -> "PPOConfig":
```

nemo_rl/algorithms/single_controller_utils/config.py
```diff
@@ -854,21 +854,24 @@ def _validate_algo_settings(master_config: MasterConfig) -> None:
             "carry TQWorkerMixin, so it has no data-plane setup to call (#2625)."
         )
 
-    if algo_cfg.ppo_epochs < 1:
-        raise ValueError("ppo.ppo_epochs must be at least 1")
+    # model_construct-based callers bypass PPOConfig's model validator, so run
+    # the same epoch validation at setup to keep invalid counts fail-fast.
+    assert master_config.ppo is not None
+    master_config.ppo.validate_epoch_settings()
```
yuki-97: `validate_epoch_settings` already checked in ppo config, and `master_config.ppo is not None` is already checked above, both no need to keep.
```suggestion
```

tests/unit/single_controller/test_single_controller_actor.py
```diff
@@ -1459,11 +1460,15 @@ def _ppo_train_pump_controller(
...
     ppo_epochs: int = 1,
+    critic_ppo_epochs: int | None = None,
 ) -> tuple[object, _NoOpValue]:
     ctrl = _train_pump_controller(sampler=sampler)
     value = _NoOpValue() if value is None else value
     ctrl._is_ppo = True
     ctrl._ppo_epochs = ppo_epochs
+    ctrl._critic_ppo_epochs = (
+        ppo_epochs if critic_ppo_epochs is None else critic_ppo_epochs
+    )
```
yuki-97, line 1496: This default collapses the two counts, so the SC warmup test can't tell them apart: [test_train_pump_freezes_the_policy_during_critic_warmup](https://github.com/NVIDIA-NeMo/RL/blob/bf66a15b53770131246ae9cc36c14d9da74e4323/tests/unit/single_controller/test_single_controller_actor.py#L1606-L1618) (line 1606) passes no `critic_ppo_epochs`, and its assertion at [line 1626](https://github.com/NVIDIA-NeMo/RL/blob/bf66a15b53770131246ae9cc36c14d9da74e4323/tests/unit/single_controller/test_single_controller_actor.py#L1626) reads `== ppo_epochs`. The one SC test with distinct counts runs at `policy_training_start_step=0`.

Swapping `num_epochs=self._critic_ppo_epochs` at [single_controller.py:1853](https://github.com/NVIDIA-NeMo/RL/blob/bf66a15b53770131246ae9cc36c14d9da74e4323/nemo_rl/algorithms/single_controller.py#L1853) for `self._ppo_epochs` on the warmup branch keeps the whole suite green -- the seam [the previous comment](https://github.com/NVIDIA-NeMo/RL/pull/3829#discussion_r3882921150) asked to pin, now covered on `ppo_train` and `async_ppo_train` but not on SC.

**Suggest fix:** pass `critic_ppo_epochs=3` into that test and assert the count is 3.

docs/guides/ppo.md
```diff
@@ -270,6 +274,7 @@ ppo:
   max_num_epochs: 100000
   max_num_steps: 100000
   ppo_epochs: 4
+  critic_ppo_epochs: ${ppo.ppo_epochs}
```
yuki-97, line 277: The value tick-budget bullet at [docs/guides/ppo.md:228](https://github.com/NVIDIA-NeMo/RL/blob/bf66a15b53770131246ae9cc36c14d9da74e4323/docs/guides/ppo.md#L228) goes stale with this key: it sizes the critic budget as `x ppo_epochs`, but setup now uses `outer_training_steps x critic_ppo_epochs` at [ppo.py:774-776](https://github.com/NVIDIA-NeMo/RL/blob/bf66a15b53770131246ae9cc36c14d9da74e4323/nemo_rl/algorithms/ppo.py#L774-L776).

Decoupling the counts is exactly what makes that section's assert fire -- someone who matches `ppo_epochs` across the seed and this run, as written, still trips `_check_and_set` on the value scheduler.

**Suggest fix:** `ppo_epochs` -> `critic_ppo_epochs` in that bullet.

nemo_rl/algorithms/ppo.py
yuki-97: Exemplar-only, and that's the right call for everything in-tree -- nothing in `nemo_rl/` constructs `PPOConfig` directly, and all 12 PPO configs reach the interpolation, 10 of them through a `defaults:` chain rooted at [ppo_math_1B.yaml](https://github.com/NVIDIA-NeMo/RL/blob/bf66a15b53770131246ae9cc36c14d9da74e4323/examples/configs/ppo_math_1B.yaml#L13).

The case it doesn't cover is upgrade migration: that file has no `defaults:` key of its own, so a copy taken before this PR carries no `critic_ppo_epochs` at all and now picks up the class default [4](https://github.com/NVIDIA-NeMo/RL/blob/bf66a15b53770131246ae9cc36c14d9da74e4323/nemo_rl/algorithms/ppo.py#L196) -- a `ppo_epochs: 1` config goes from 1 critic pass to 4 with nothing to warn on it.

**Suggest fix:** a `model_validator(mode="after")` copying `ppo_epochs` when `"critic_ppo_epochs" not in self.model_fields_set` keeps the concrete `int` and restores the coupling. Leaving it as-is is defensible too -- your call.

yuki-97: good catch! I think we can copy `ppo_epochs` when `"critic_ppo_epochs" not in self.model_fields_set`, and worth adding a comment above `critic_ppo_epochs: int = 4` and in the yaml to say this behavior.

RayenTian:
/ok to test [a440755](https://github.com/NVIDIA-NeMo/RL/pull/3829/commits/a440755d229414871ee261b5e4bc3b7ffdea9a46)

docs/guides/ppo.md
RayenTian: Done in [a440755](https://github.com/NVIDIA-NeMo/RL/pull/3829/commits/a440755d229414871ee261b5e4bc3b7ffdea9a46)

tests/unit/single_controller/test_single_controller_actor.py
RayenTian: Done in [a440755](https://github.com/NVIDIA-NeMo/RL/pull/3829/commits/a440755d229414871ee261b5e4bc3b7ffdea9a46)

nemo_rl/algorithms/single_controller_utils/config.py
RayenTian: Done in [a440755](https://github.com/NVIDIA-NeMo/RL/pull/3829/commits/a440755d229414871ee261b5e4bc3b7ffdea9a46)

nemo_rl/algorithms/ppo.py
RayenTian: Done in [a440755](https://github.com/NVIDIA-NeMo/RL/pull/3829/commits/a440755d229414871ee261b5e4bc3b7ffdea9a46)

nemo_rl/algorithms/ppo.py
RayenTian: Done in [a440755](https://github.com/NVIDIA-NeMo/RL/pull/3829/commits/a440755d229414871ee261b5e4bc3b7ffdea9a46)

nemo_rl/algorithms/ppo.py
RayenTian: Done in [a440755](https://github.com/NVIDIA-NeMo/RL/pull/3829/commits/a440755d229414871ee261b5e4bc3b7ffdea9a46)

yuki-97 (approved):

PR #3834, opened 2026-08-25
feat(multimodal): Support multimodal Megatron generation.

cspades:
# What does this PR do ?

Depends on https://github.com/NVIDIA/Megatron-LM/pull/6809 and https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5799

- Integrates multimodal Megatron Inference into NeMo-RL using Nemotron Omni as a start, and having functional / image (`CLEVR`) and video (`VSTAT`) accuracy parity with vLLM: https://wandb.ai/adlr/mllm-rl-dev?nw=nwusercye_nv
- Refactors some multimodal utils after an earlier video GRPO PR: https://github.com/NVIDIA-NeMo/RL/pull/3648

Screenshot 2026-08-25 at 1 26 50 PM

Screenshot 2026-08-26 at 1 40 41 AM

# Usage

- Many YAML configs like `RL/examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-16n8g-megatron-tp4ep4-async-gym-video.v1.yaml` have been updated to support multimodal generation (if provided a multimodal dataset). Some example MLLM and vLLM parity scripts I used to test with are in a side-branch: https://github.com/cspades/RL/tree/cye/rl_mllm_omni_multimodal_scripts/scripts
  - To run these scripts: `GENERATION_BACKEND={vllm,megatron} scripts/submit_nemotron_omni_vstat_megatron_8n4g.sh` on the Slurm login node or `run_nemotron_omni_clevr_megatron_1n4g.sh` in an interactive node.

# Cause I keep forgetting where everything is...

## Non-Gym RL Multimodal Data Pipeline
```
Dataset preprocessing:

vlm_hf_data_processor → resolve_to_image / load_video → Hugging Face Omni processor → produces token_ids, pixel_values, imgs_sizes, optionally num_frames → stores media as PackedTensor

Files:

nemo_rl/data/processors.py
nemo_rl/data/multimodal_utils.py

Rollout generation:

RolloutManager.generate_and_push → AsyncRolloutImpl._generate_response → batched_message_log_to_flat_message → MegatronGeneration.generate_async → MegatronWorker._prepare_data_for_generation → _sample_vision_tensors → _build_prompt_and_multimodal_data → MCore InferenceClient.add_request

Files:

nemo_rl/experience/rollout_manager.py
nemo_rl/data/llm_message_utils.py
nemo_rl/models/generation/megatron/megatron_worker.py

Megatron inference:

resolve_multimodal_data_for_engine → tensor payload passes through without decoding → DynamicInferenceEngine._build_vlm_request → NemotronOmniInferenceWrapper._forward_vision_encoder → NemotronOmniModel._encode_images

Here media_tokens_preexpanded=True, because NeMo-RL already expanded the image placeholders.
```
## Gym RL Multimodal Data Pipeline
```
Dataset-side preprocessing for training/recompute:

nemo_gym_data_processor → nemo_gym_example_to_video_datum_spec → load_video_frames_with_metadata → process_nemotron_video_frames → _expand_nemotron_video_placeholders → produces PackedTensor(pixel_values, imgs_sizes, num_frames)

Files:

nemo_rl/data/processors.py
nemo_rl/environments/nemo_gym_multimodal.py
nemo_rl/environments/nemotron_utils.py

Gym rollout generation:

AsyncNemoGymRolloutImpl → NemoGym.run_rollouts → normalize_media_in_examples → Gym RolloutCollectionHelper.run_examples → Gym agent → SimpleResponsesAPIModel.responses_dispatch → Megatron OpenAI HTTP server

Files:

nemo_rl/experience/rollout_manager.py
nemo_rl/environments/nemo_gym.py
3rdparty/Gym-workspace/Gym/nemo_gym/rollout_collection.py
3rdparty/Gym-workspace/Gym/nemo_gym/base_responses_api_model.py

HTTP-side preprocessing for generation:

chat_completions → _extract_multimodal_from_messages → raw image/video bytes → resolve_multimodal_data_for_engine → preprocess_video_bytes_list → DynamicInferenceEngine._build_vlm_request → NemotronOmniInferenceWrapper.expand_image_tokens → _forward_vision_encoder → NemotronOmniModel._encode_images

Files:

Megatron-LM/.../endpoints/chat_completions.py
Megatron-LM/.../image_preprocessing.py
Megatron-LM/.../dynamic_engine.py
Megatron-LM/.../nemotron_omni_inference_wrapper.py

Here media_tokens_preexpanded=False; Megatron-LM expands placeholders itself.

After Gym returns:

NemoGym._postprocess_nemo_gym_to_nemo_rl_result → attach generated token IDs/logprobs → attach_static_multimodal_payload → restore the original NeMo-RL PackedTensors for policy recompute and training.
```

**Pre checks**:

# Additional Information
* ...


cspades:
/ok to test [63f916f](https://github.com/NVIDIA-NeMo/RL/pull/3834/commits/63f916fc6933a1bd3f93e7cd315ae34a6c0c8dac)

cspades:
Lockfile test is not gonna be fixed until we bump MBridge and MCore. 😩

pyproject.toml
```diff
@@ -437,7 +437,7 @@ constraint-dependencies = [
   # whose [tool.uv.sources] pins them to git. uv requires such URL deps to also appear as a
   # direct requirement or constraint of the root project. Keep these URLs/revs in sync with
   # Megatron-LM's [tool.uv.sources] (uv errors loudly on mismatch after a submodule bump).
-  "emerging-optimizers @ git+https://github.com/NVIDIA-NeMo/Emerging-Optimizers.git@v0.2.0",
+  "emerging-optimizers @ git+https://github.com/NVIDIA-NeMo/Emerging-Optimizers.git@v0.3.0",
```
cspades, line 459: [`pyproject.toml:440`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/pyproject.toml#L440)

**1 action item.**

This bumps `emerging-optimizers` to `@v0.3.0`, but the nested Megatron-LM submodule pin at `3rdparty/Megatron-Bridge-workspace/Megatron-Bridge/3rdparty/Megatron-LM/pyproject.toml:244` still pins `v0.2.0`. Reproduced directly: `uv lock --check` fails with

```
Failed to resolve dependencies for `megatron-core`: Requirements contain conflicting URLs for package `emerging-optimizers`: v0.2.0 vs v0.3.0
```

The comment directly above this line documents the exact invariant this violates: "Keep these URLs/revs in sync with Megatron-LM's `[tool.uv.sources]` (uv errors loudly on mismatch after a submodule bump)." Every contributor's `uv sync --extra mcore` is currently broken by this.

### AI-1
Revert this to `v0.2.0` until the companion Megatron-LM/Megatron-Bridge submodule bump lands together, or land the submodule bump in the same PR.

tests/unit/models/generation/test_megatron_generation.py
```diff
@@ -37,6 +39,167 @@
 
 model_name = "Qwen/Qwen3-0.6B"
 
+
+def test_direct_megatron_media_request_preserves_preexpanded_prompt():
```
cspades: [`tests/unit/models/generation/test_megatron_generation.py:43`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/tests/unit/models/generation/test_megatron_generation.py#L43)

**1 action item.**

This test (and the two below it, `test_direct_megatron_video_request_marks_preexpanded_prompt` at [L71](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/tests/unit/models/generation/test_megatron_generation.py#L71) and `test_direct_megatron_multimodal_generate_round_trip` at [L107](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/tests/unit/models/generation/test_megatron_generation.py#L107)) is missing `@pytest.mark.mcore` — every other test/class in this file (from [L441](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/tests/unit/models/generation/test_megatron_generation.py#L441) onward) carries it.

This matters because none of the other L0 shards ever collect this file's content: `Models_*.sh` explicitly `--ignore`s this directory, and `Vllm_*.sh`/`Sglang.sh` filter by their own backend markers. Only `L0_Unit_Tests_Megatron.sh` (`--mcore-only`) would collect it, and marker-filtering means these 3 unmarked tests are excluded there too. **These 3 tests are not executed by any currently-configured CI shard.**

### AI-1
Add `@pytest.mark.mcore` above each of the 3 new tests.

```suggestion
@pytest.mark.mcore
def test_direct_megatron_media_request_preserves_preexpanded_prompt():
```

nemo_rl/algorithms/grpo.py
```diff
@@ -1349,41 +1375,55 @@ def initialize_generation_with_policy(
...
+                    policy_generation, megatron_gen_time = generation_future.result()
+                    if not colocated_inference:
+                        setup_timing_metrics.parallel_wall_time_s = (
+                            time.perf_counter() - init_tasks_t0
+                        )
+                        setup_timing_metrics.parallel_init_enabled = 1.0
+                        # NeMo Gym probes the pre-published endpoint before its future completes.
+                        # A skip-load Megatron endpoint starts only during this initial refit,
+                        # so it must happen while Gym is waiting rather than after it resolves.
+                        init_megatron_weight_synchronizer(policy, policy_generation)
```
cspades, line 1449: [`nemo_rl/algorithms/grpo.py:1402-1423`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/algorithms/grpo.py#L1402)

**1 action item.**

**Background:** for non-colocated (skip-load) Megatron generation, the inference engine's HTTP server does not start at construction — only during `sync_weights()` inside `init_megatron_weight_synchronizer` (defined at [L1275](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/algorithms/grpo.py#L1275)). `_spinup_nemo_gym` (submitted as `nemo_gym_future` in the same `ThreadPoolExecutor`) polls that endpoint and blocks until it responds.

**What breaks:** `init_megatron_weight_synchronizer(policy, policy_generation)` is called inline (not as its own future) at [L1420](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/algorithms/grpo.py#L1420), between `generation_future.result()` and `nemo_gym_future.result()`, still inside the `with ThreadPoolExecutor(...)` block. If `init_communicator()` raises before `sync_weights()` runs, the exception unwinds through `ThreadPoolExecutor.__exit__` → `shutdown(wait=True)`, which now waits on `nemo_gym_future` — still polling an endpoint that will never come up. Traced through `nemo_gym`'s own readiness wait (`RunHelper.start()` → `_wait_for_model_endpoints`): it's bounded by `model_endpoint_readiness_timeout_seconds`, defaulting to **600s** and unset by every recipe in this repo. So the real error (and the outer `finally: ray.kill(port_holder)` cleanup at [L1423](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/algorithms/grpo.py#L1423)) is delayed by up to 10 minutes behind a confusing secondary `ConfigError` about unreachable model endpoints — a control-plane failure now looks like a hang instead of failing fast. (If a recipe ever sets that timeout to `0`, this becomes a true infinite hang.)

### AI-1
Don't let a raise from `init_megatron_weight_synchronizer` transit `ThreadPoolExecutor.__exit__` while `nemo_gym_future` can still block on it. Either submit it as its own executor task and use `concurrent.futures.wait(..., return_when=FIRST_EXCEPTION)` + `cancel_futures=True`, or wrap the inline call in `try/except` and explicitly cancel/bound `nemo_gym_future` (or at least call `ray.kill(port_holder)`) before re-raising.

nemo_rl/models/generation/megatron/config.py
```diff
@@ -44,17 +44,31 @@ class MCoreGenerationSpecificArgs(TypedDict):
...
+    # Required for EP>1 + local CUDA graphs.
+    moe_pad_experts_for_cuda_graph_inference: NotRequired[bool]
 
     materialize_only_last_token_logits: bool
     enable_chunked_prefill: bool
     enable_prefix_caching: bool
+    async_sched_mode: NotRequired[Literal["legacy", "async"]]
+    vision_embedding_cache_max_bytes: NotRequired[int]
+    allow_stale_multimodal_embeddings: NotRequired[bool]
+    logprobs_mode: NotRequired[Literal["raw_logprobs", "processed_logprobs"]]
```
cspades: [`nemo_rl/models/generation/megatron/config.py:56`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/models/generation/megatron/config.py#L56)

**1 action item.**

`logprobs_mode` is declared twice in `MCoreGenerationSpecificArgs`: `NotRequired[...]` here, and required (`Literal["processed_logprobs", "raw_logprobs"]`, pre-existing) at [L85](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/models/generation/megatron/config.py#L85). Since `TypedDict` class bodies are executed like a normal namespace, the later (required) declaration wins — this new `NotRequired` entry is dead.

Consistent with that, [`megatron_worker.py:391-393`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/models/generation/megatron/megatron_worker.py#L391) was changed from a plain `mcore_generation_config["logprobs_mode"]` subscript to `.get("logprobs_mode", "processed_logprobs")` — a hidden default at the call site that contradicts both the still-required TypedDict declaration and the config-conventions rule that defaults belong in exemplar YAML, not `.get(k, default)` in code. Every exemplar/recipe YAML checked already sets this key explicitly.

### AI-1
Pick one: delete the stale required declaration at L85 to make the field genuinely optional (keep the `.get()` fallback), or delete the dead `NotRequired` entry here and revert `megatron_worker.py:391-393` back to the plain subscript.

nemo_rl/evals/eval.py
```diff
@@ -350,10 +348,29 @@ async def _run_env_eval_impl(
                     multi_modal_data["video"] = (
                         videos[i][0] if len(videos[i]) == 1 else videos[i]
                     )
-                if multi_modal_data:
-                    prompt_dict["multi_modal_data"] = multi_modal_data
+
+            if multi_modal_data:
```
cspades: [`nemo_rl/evals/eval.py:352-374`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/evals/eval.py#L352)

**1 action item** (medium confidence — code path confirmed, haven't confirmed a specific recipe hits it).

For a native-content (non-placeholder) VLM processor, `vllm_content[i]` is populated with the chat-template-formatted dialog regardless of whether that particular row has any media (`vlm_hf_data_processor` sets it unconditionally when `not uses_placeholder`). The old gate used `vllm_content` whenever it was non-`None`; this PR's gate is `if multi_modal_data:` ([L352](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/evals/eval.py#L352)) — so a text-only row inside an otherwise-multimodal batch now falls into the "Text-only fallback" branch ([L374-383](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/evals/eval.py#L374)), which naively joins raw `message["content"]` strings instead of using the already-formatted `vllm_content[i]`.

### AI-1
Confirm whether mixed (some-rows-have-media, some-don't) multimodal eval batches are a real path for any current recipe. If so, use `vllm_content[i]` whenever it's non-`None`, independent of whether `multi_modal_data` is empty for that row.

nemo_rl/models/generation/megatron/utils.py
```diff
@@ -12,10 +12,107 @@
...
+    target_num_patches = generation_config.get("video_target_num_patches")
+    if target_num_patches is not None:
+        image_config = replace(
+            image_config,
+            dynamic_resolution_max_patches=int(target_num_patches),
+        )
+
+    return VideoProcessingConfig(
+        image_config=image_config,
+        num_frames=int(generation_config["video_num_frames"]),
```
cspades: [`nemo_rl/models/generation/megatron/utils.py:109`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/models/generation/megatron/utils.py#L109)

**1 action item**, low severity — no exemplar/recipe currently triggers this (all set both fields together).

`build_video_preprocessing_config` gates only on `video_temporal_patch_size` being set, then unconditionally does `generation_config["video_num_frames"]`. Both fields are independently `NotRequired` in `MCoreGenerationConfig`, so a config setting only `video_temporal_patch_size` gets a raw `KeyError` instead of a clear error.

### AI-1
```suggestion
    video_num_frames = generation_config.get("video_num_frames")
    if video_num_frames is None:
        raise ValueError(
            "video_num_frames must be set when video_temporal_patch_size is set."
        )
    return VideoProcessingConfig(
        image_config=image_config,
        num_frames=int(video_num_frames),
```

nemo_rl/algorithms/grpo.py
```diff
@@ -1349,41 +1375,55 @@ def initialize_generation_with_policy(
...
-            print(f"  ⚡ Init tasks: {', '.join(init_tasks.keys())}", flush=True)
+            # Exactly one task adopts the reserved port: the policy when colocated
+            # (generation wraps it), else the dedicated generation policy.
+            policy_port, generation_port = (
+                (reserved_http_server_port, None)
+                if colocated_inference
+                else (None, reserved_http_server_port)
+            )
+
+            def init_megatron_generation_task(policy_future):
```
cspades, line 1419: [`nemo_rl/algorithms/grpo.py:1362-1433`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/algorithms/grpo.py#L1362)

**No action needed** — non-blocking design suggestion, not a blocker for this PR.

Megatron's nemo_gym-parallel-init path here diverges from the shared `init_tasks` dict/loop shape vLLM's branch still uses ([L1506-1577](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/algorithms/grpo.py#L1506)) into 3 hand-named futures (`policy_future`, `generation_future`, `nemo_gym_future`) with the weight-sync ordering baked in procedurally, to satisfy the same "sync must precede gym resolving" constraint the bug above is about. A named plausible extension: sglang/trtllm adding a similar deferred-load + parallel-gym-init mode would have no shared primitive to extend and would likely copy this hand-unrolled block a third time.

### Follow-up
Consider a shared `run_parallel_init_tasks(tasks, after={name: hook})` primitive — an optional `after` hook keyed by task name that fires once that task's future resolves, still inside the same executor block — so Megatron could rejoin the same dict/loop shape as vLLM instead of a bespoke block.

tests/unit/utils/test_venvs.py
```diff
@@ -16,7 +16,10 @@
 from tempfile import TemporaryDirectory
 from unittest.mock import patch
 
-from nemo_rl.utils.venvs import create_local_venv
+from nemo_rl.utils.venvs import (
+    add_hf_modules_cache_to_pythonpath,
```
cspades, line 20: [`nemo_rl/utils/venvs.py:233-235`](https://github.com/NVIDIA-NeMo/RL/blob/faede41c4de1325010a1a0d6a141be9d54e578b8/nemo_rl/utils/venvs.py#L233)

**1 action item** — new test, verified passing locally.

`make_actor_runtime_env`'s `"uv"`-prefixed executable branch (the local per-node venv creation path, previously inlined in `grpo.py` and now centralized here) has no direct unit test — `tests/unit/algorithms/test_grpo.py` mocks `make_actor_runtime_env` wholesale instead of exercising it.

### AI-1
```suggestion
def test_make_actor_runtime_env_builds_local_venv_for_uv_python_executable():
    """Mirrors the inline venv-creation logic that used to live in grpo.py."""
    with (
        patch(
            "nemo_rl.distributed.ray_actor_environment_registry.get_actor_python_env",
            return_value="uv run --group vllm",
        ) as mock_get_env,
        patch(
            "nemo_rl.utils.venvs.create_local_venv_on_each_node",
            return_value="/fake/venv/bin/python",
        ) as mock_create_venv,
    ):
        runtime_env = make_actor_runtime_env("some.module.SomeActor")

    mock_get_env.assert_called_once_with("some.module.SomeActor")
    mock_create_venv.assert_called_once_with(
        "uv run --group vllm", "some.module.SomeActor"
    )
    assert runtime_env["py_executable"] == "/fake/venv/bin/python"
    assert runtime_env["env_vars"]["VIRTUAL_ENV"] == "/fake/venv"
    assert runtime_env["env_vars"]["UV_PROJECT_ENVIRONMENT"] == "/fake/venv"
```
(needs `make_actor_runtime_env` added to this file's imports)

cspades (commented):
Team review by 5 agents (rl-expert, bug-finder, test-agent, design-reviewer, devil-advocate). 8 findings staged as inline comments: 3 critical (a submodule/pyproject version mismatch that breaks `uv sync --extra mcore` today, 3 new tests never collected by any CI shard, and a bounded ~600s setup stall on Megatron weight-sync failure with NeMo Gym), 5 suggestions (a duplicate/hidden-default config field, a medium-confidence eval.py regression for mixed multimodal batches, a low-severity KeyError gap, a non-blocking design note, and a verified-passing test-coverage addition). Local `pre-commit` could not run in this environment (Python 3.13.14 interpreter unavailable) — not attributable to this PR's diff.

_Generated by Claude Code_

nemo_rl/algorithms/grpo.py
cspades: Disputing my own finding here: @tdene reviewed this exact restructured code on the sibling PR #3864 and confirmed the 600s `wait_for_model_endpoints` timeout is intentional, working-as-designed behavior, not a leak: https://github.com/NVIDIA-NeMo/RL/pull/3864#discussion_r3874954084 ("there is already a 600s timeout on `wait_for_model_endpoints`. Nothing leaks, things time out as designed."). Withdrawing the suggested code change — no action needed here.

nemo_rl/algorithms/grpo.py
cspades: Related to the other issue that isn't a bug.

nemo_rl/evals/eval.py
cspades: This is a nice QoL for mixed batches.

nemo_rl/models/generation/megatron/utils.py
```diff
@@ -12,10 +12,130 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from dataclasses import replace
+from typing import Any
+
 import torch
+from megatron.core.inference.config import ImageProcessingConfig, VideoProcessingConfig
```
terrykong, line 19: [`utils.py:19`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/utils.py#L19)

Placeholder comment so this does not get lost: Megatron-Bridge needs to be bumped, along with the matching Megatron-LM carrying [#6809](https://github.com/NVIDIA/Megatron-LM/pull/6809), before this merges — `ImageProcessingConfig` and `VideoProcessingConfig` do not exist at the current pin.

examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n4g-megatron_generation.v1.yaml
```diff
@@ -0,0 +1,72 @@
+defaults: ../../vlm_grpo_3B_megatron.yaml
```
terrykong, line 1: [`examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n4g-megatron_generation.v1.yaml:1`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n4g-megatron_generation.v1.yaml#L1)

**1 action item.**

**TL;DR — the three new recipe YAMLs have no matching test-suite driver, so `test_all_recipe_yamls_accounted_for_in_test_suites` fails in every unit lane.**

`tests/unit/test_recipes_and_test_suites.py` compares the set of recipe YAMLs against the set of scripts listed in `nightly.txt`, `release.txt`, the two `*_gb200.txt` files, both `performance*.txt` files and `disabled.txt`. I reproduced its set math on this head with plain `python3`:

```
recipes: 266   suites: 263
in recipes not in suites:
  vlm/vlm_grpo-nemotron-omni-30ba3b-circle-count-1n4g-megatron_generation.v1
  vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n4g-megatron_generation.v1
  vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-8n4g-megatron_generation.v1
in suites not in recipes: []
```

The test asserts the two lengths are equal before it does anything else, so it fails with `AssertionError: Recipe YAMLs should be accounted for in the test suites`. The file sits in `tests/unit/` and is not excluded, so it runs in every unit shard.

The three new `tests/functional/*.sh` scripts do not satisfy this — that is a different lane.

**Action:** add `tests/test_suites/vlm/<same-basename>.sh` for each of the three recipes and list them in `tests/test_suites/nightly.txt`; or, if they are not meant to run nightly yet, add them to `tests/test_suites/disabled.txt` with a reason, the way the other Omni entries are handled. Not a suggestion block — it spans three new files plus a suite list.

tests/functional/L1_Functional_Tests_Megatron_Omni.sh
```diff
@@ -0,0 +1,38 @@
...
+if (( GPU_COUNT < 4 )); then
+    echo "SKIP: Nemotron Omni functional tests require at least four GPUs"
+    exit 0
+fi
+
+# The recipes are intentionally 1n4g even when the CI runner exposes eight GPUs.
+export CUDA_VISIBLE_DEVICES="${CUDA_VISIBLE_DEVICES:-0,1,2,3}"
+
+time uv run --no-sync bash ./tests/functional/nemotron_omni_clevr_megatron_1n4g.sh
+time uv run --no-sync bash ./tests/functional/nemotron_omni_gym_video_megatron_1n4g.sh
```
terrykong: [`L1_Functional_Tests_Megatron_Omni.sh:23-26`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/L1_Functional_Tests_Megatron_Omni.sh#L23)

**1 action item.**

**TL;DR — this job asks for 4 GPUs, the functional shard has 2, so the guard below trips and the job exits 0. It will report green in CI without ever running.**

Both inner scripts are sized for 4 ([`clevr:84`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/nemotron_omni_clevr_megatron_1n4g.sh#L84), [`video:72`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/nemotron_omni_gym_video_megatron_1n4g.sh#L72): `cluster.gpus_per_node=4`, with TP=4 and EP=4 on both the train and generation side). The wrapper then guards on count:

```bash
GPU_COUNT=$(nvidia-smi --query-gpu=index --format=csv,noheader | wc -l)
if (( GPU_COUNT < 4 )); then
    echo "SKIP: Nemotron Omni functional tests require at least four GPUs"
    exit 0
fi
```

The repo already records that the shard has 2 GPUs, and disabled its only other 4-GPU test for exactly this reason — [`L1_Functional_Tests_GB200_MXFP8.sh:38-40`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/L1_Functional_Tests_GB200_MXFP8.sh#L38) says the test is *"sized for 4 GPUs (2 train + 2 inference, EP=2 per side)"* but the runner has 2, "so the inference placement group can never be satisfied", tracked in [#3731](https://github.com/NVIDIA-NeMo/RL/issues/3731).

Supporting count: of the functional scripts that set `cluster.gpus_per_node`, 73 use 2. The only three that use 4 are the two added here and that disabled one.

An `exit 0` skip is worse than no test here, because the job is wired into CI and will show a green check for coverage that never ran.

**Action:** one of —
1. Scale both scripts to 2 GPUs (TP=2, EP=2, `cluster.gpus_per_node=2`) so they match every other functional test and actually execute. Whether Nemotron Omni 30B-A3B fits at TP=2 is the open question, and you are better placed to answer it than I am.
2. Move this to nightly, where multi-GPU topologies live, and drop it from L1.
3. Keep it as-is but make the skip loud, and accept that it is inert in presubmit.

No suggestion block — the fix is a parallelism retune across two files, not a single-region edit.

(Separately, and only if this ends up running on Blackwell: the sibling `L1_Functional_Tests_Megatron_4.sh` wraps every megatron-generation script in a `compute_cap >= 100` skip, with the note "Megatron Inference currently hits an IMA on Blackwell tests." That guard is missing here, but it is academic while the GPU-count check trips first.)

nemo_rl/data/processors.py
```diff
@@ -676,8 +678,9 @@ def vlm_hf_data_processor(
         loss_multiplier = 0.0
     else:
         # get the prompt content! (use this for vllm-backend that needs formatted dialog and list of images/audios) for the entire conversation
+        # Placeholder-style processors set vllm_content to None so vLLM uses expanded input_ids.
         vllm_kwargs = {
-            "vllm_content": string_formatted_dialog,
+            "vllm_content": None if uses_placeholder else string_formatted_dialog,
```
terrykong: [`nemo_rl/data/processors.py:683`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/data/processors.py#L683)

**1 action item.**

This gate checks `uses_placeholder` on its own, but the two sibling branches in the same function check `uses_placeholder and images` — [`processors.py:562`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/data/processors.py#L562) and [`processors.py:589`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/data/processors.py#L589).

For a placeholder-style processor (Nemotron Omni) on a row with no media, `images` is empty, so both siblings correctly take the non-placeholder path and build `string_formatted_dialog` at [`processors.py:583`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/data/processors.py#L583). This line still sets `vllm_content` to `None`.

That row then has `vllm_content = None` and no media, so [`eval.py`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/evals/eval.py#L379) sends it down the text-only fallback, which joins the raw `message["content"]` strings with no chat template and no generation prompt. Before this PR, `vllm_content` was set unconditionally and the row kept its formatted dialog. The eval score moves and nothing raises.

The `eval.py` restructure in `73f6117` fixed the consumer side, but it cannot recover this row, because `vllm_content` is already `None` by the time `eval.py` sees it.

**Action:** match the siblings.

```suggestion
            "vllm_content": None if (uses_placeholder and images) else string_formatted_dialog,
```

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -179,8 +180,9 @@ def __init__(
         )
 
         # `self.cfg` exposes the `generation` that matches the `GenerationInterface` contract.
-        # `self._policy_config` keeps a reference to the full PolicyConfig.
-        self._policy_config = config
+        # `self._policy_config` keeps a reference to the full PolicyConfig. Dedicated
+        # inference receives a copy because worker setup may modify it.
+        self._policy_config = config if policy is not None else deepcopy(config)
```
terrykong: [`nemo_rl/models/generation/megatron/megatron_generation.py:185`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_generation.py#L185)

**1 action item.**

The `deepcopy` here is never read. On the only path that computes it, [`L203-206`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_generation.py#L203) reassigns `self._policy_config` to `{**config, ...}` — spreading the **original** `config`, not the copy.

1. `policy is not None` takes `config` directly and returns at [`L199`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_generation.py#L199).
2. So the deepcopy branch is the only way to reach L203.
3. Nothing between L185 and L203 reads `self._policy_config`. [`effective_megatron_cfg`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_generation.py#L47) is a `@staticmethod` that takes `config` as an argument, so it does not read it either.

The intent looks right — give dedicated inference its own config so worker setup can mutate freely. The protection just is not complete. `{**config}` is a shallow copy: the top level is safe and `megatron_cfg` is safe (`effective_megatron_cfg` returns a fresh dict), but every other nested dict — `config["generation"]` among them — is still shared with the caller. So worker setup mutating a nested value still reaches the config the caller passed in, which is what the comment above says will not happen.

**Action:** wire the copy into the dict that is actually used, so the isolation the comment promises is real.

```suggestion
        self._policy_config = {
            **deepcopy(config),
            "megatron_cfg": self.effective_megatron_cfg(config),
        }
```

If the isolation is not actually needed, the other valid fix is to delete the copy at L185 and the `from copy import deepcopy` at [`L15`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_generation.py#L15), and reword the comment to say the config is shared. Either way, the current shape pays for a deep copy and throws it away.

nemo_rl/environments/nemo_gym_request.py
```diff
@@ -0,0 +1,91 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+import json
+from typing import Any
+
+
+def _json_mapping(value: Any, *, field_name: str) -> dict[str, Any]:
```
terrykong, line 22: [`nemo_rl/environments/nemo_gym_request.py:20`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/environments/nemo_gym_request.py#L20)

**2 action items.**

This new module has no module docstring and none of its four functions have one, yet three of them are imported by other modules ([`nemo_gym_multimodal.py:44-46`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/environments/nemo_gym_multimodal.py#L44)). All four raise typed errors that callers need to know about.

### AI-1
Add a module docstring and Google-style docstrings to the four functions. Starting with this one:

```suggestion
def _json_mapping(value: Any, *, field_name: str) -> dict[str, Any]:
    """Coerce a Responses-API metadata value into a plain dict.

    Responses `metadata` values must be strings, so callers may pass either a
    JSON object or a JSON string holding one.

    Args:
        value: A mapping, or a JSON string that decodes to one.
        field_name: Name used in the error message when `value` is neither.

    Returns:
        The decoded mapping, or an empty dict when `value` is None.

    Raises:
        ValueError: If `value` is not a mapping and does not decode to one.
    """
```

### AI-2
`_chat_template_kwargs_for_processor` ([`L51`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/environments/nemo_gym_request.py#L51)) and `_deep_merge_dict` ([`L84`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/environments/nemo_gym_request.py#L84)) have no direct tests, while `_metadata_extra_body` does. `_chat_template_kwargs_for_processor` is the reader for the new JSON-string `chat_template_kwargs` contract that `examples/nemo_gym/prepare_video_dataset.py` now writes, so it is the one most worth pinning. Please add cases to `tests/unit/environments/test_nemo_gym.py` covering a dict value, a JSON-string value, a missing value, and an invalid-JSON value.

tests/unit/data/test_multimodal_image_encoding.py
```diff
@@ -117,31 +119,33 @@ def test_encode_images_deduplicates_sources_and_uses_a_bounded_thread_pool(
...
+    assert observed_max_workers == [
+        nemo_gym_multimodal._NEMO_GYM_IMAGE_ENCODE_MAX_WORKERS
+    ]
     for example in examples:
         parts = example["responses_create_params"]["input"][0]["content"]
         assert parts[0]["image_url"] == expected[first]
-        assert parts[1]["image"] == expected[first]
-        assert parts[2]["url"] == expected[second]
+        assert parts[1]["image_url"] == expected[first]
+        assert parts[2]["image_url"] == expected[second]
```
terrykong: [`test_multimodal_image_encoding.py:147-148`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/unit/data/test_multimodal_image_encoding.py#L147)

**1 action item.**

These two assertions were updated to read the new `image_url` key, which records a real behavior change: `normalize_media_in_examples` now rewrites `part["type"]` to a canonical `input_image` / `input_video` and drops the original `image` / `url` key ([`nemo_gym_multimodal.py:177-180`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/environments/nemo_gym_multimodal.py#L177)).

Nothing asserts either half. A regression that left the stale `image` key in place holding the original local path would still pass this test.

**Action:** assert the whole part, so both the canonical type and the dropped key are covered.

```suggestion
        assert parts[1] == {"type": "input_image", "image_url": expected[first]}
        assert parts[2] == {"type": "input_image", "image_url": expected[second]}
```

Separately, five branches added to the same function have no test at all: the two `ValueError`s for a part with the wrong number of source keys and for an empty URL, the `file_id` skip, the nested-`detail` promotion, and the whole video branch through `video_path_to_data_url`.

nemo_rl/models/generation/megatron/utils.py
```diff
@@ -12,10 +12,130 @@
...
+        pixel_shuffle=merge_size > 1,
+        spatial_merge_size=merge_size,
+        dynamic_resolution_min_patches=int(min_patches),
+        dynamic_resolution_max_patches=int(max_patches),
+        pixel_mean=[float(value) for value in pixel_mean],
+        pixel_std=[float(value) for value in pixel_std],
+    )
+
+
+def build_video_preprocessing_config(
```
terrykong, line 213: [`nemo_rl/models/generation/megatron/utils.py:98`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/utils.py#L98)

**1 action item.**

Neither `build_video_preprocessing_config` nor [`build_image_preprocessing_config`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/utils.py#L23) is referenced anywhere under `tests/`. They are the highest-risk logic this PR adds and the cheapest to test: both take a duck-typed HF processor plus a plain dict and return a dataclass, so a `SimpleNamespace` fake is enough — no GPU, no Ray, no model.

`build_image_preprocessing_config` reads five fields under alternate names (`patch_size` or `patch_dim`, `norm_mean` or `image_mean`, and so on), unwraps a dict-valued `patch_size`, and derives `merge_size` from `downsample_ratio` via `round(1/x)` with a fallback chain. Each of those is a place where a second VLM family quietly gets the wrong number instead of an error.

**Action:** add `tests/unit/models/generation/test_megatron_generation_utils.py` (or extend `test_megatron_generation_parse.py`, which is already `mcore`-marked and collected by the Megatron lane) covering:

- `build_image_preprocessing_config`: dict-valued `patch_size`; `downsample_ratio` present, then absent with `merge_size` as fallback, then neither; and that the error names every missing field.
- `build_video_preprocessing_config`: returns `None` when `image_config is None`; returns `None` when `video_temporal_patch_size` is absent; raises the new named `ValueError` when `video_num_frames` is absent; and that `video_target_num_patches` overrides `dynamic_resolution_max_patches` **without mutating the caller's image config**.

No suggestion block — this is a new test file rather than an edit to a line here.

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -641,10 +784,115 @@ def _merge_stop_strings(
                     stop_set.update(sample_ss)
         return list(stop_set) if stop_set else None
 
+    def _sample_vision_tensors(self, data, index: int):
```
terrykong, line 795: [`megatron_worker.py:787`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_worker.py#L787)

**No action needed** — design suggestion for a follow-up, not a blocker.

This PR adds two per-sample helpers to `MegatronGenerationMixin`, and they are a pair:

- `_sample_vision_tensors(data, index)` (this line) pulls one row's `pixel_values` / `imgs_sizes` / `num_frames` out of the batch.
- [`_build_prompt_and_multimodal_data(data, index)`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_worker.py#L856) builds one request — the token prompt plus the media payload MCore receives — and [`calls the first at L860`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_worker.py#L860).

Neither touches worker state. The first uses only its two arguments. The second adds calls to `self._sample_vision_tensors`, `self._get_megatron_inference_wrapper_cls` and `self._wrapper_supports_modality` — and nothing else. No `self.model`, no `self.inference_context`, no CUDA.

That is why the new tests have to build a worker that was never constructed ([`test_megatron_generation.py:45-59`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/unit/models/generation/test_megatron_generation.py#L45)):

```python
worker = object.__new__(MegatronGenerationMixin)
worker.cfg = {"generation": {"mcore_generation_config": {}}}
worker._sample_vision_tensors = sample_vision_tensors
worker._get_megatron_inference_wrapper_cls = get_wrapper_cls
worker._wrapper_supports_modality = wrapper_supports_modality
```

`object.__new__` plus three monkeypatched methods is the price of putting logic that needs no worker on a worker class.

To be clear about what this is **not**: it is not a case for a new class. The mixin has no `__init__` and no `__getstate__`, and the rest of it does reach into `self.model`, `self.processor` and `self.rank`, so there is no separable state here. A new class would just be a wrapper around its own arguments.

**Follow-up (fine to defer):** move both into [`megatron/utils.py`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/utils.py#L23), next to the two free functions this PR already put there, taking `supported_modalities: frozenset[str]` instead of reaching back for the wrapper. Keep the mixin methods as one-line delegations so nothing else changes. The three new tests then drop the `object.__new__` line and all three monkeypatches, and `build_prompt_and_multimodal_data` becomes callable with a plain dict.

nemo_rl/models/generation/megatron/utils.py
```diff
@@ -12,10 +12,130 @@
...
+            "pipeline does."
+        )
+
+    downsample_ratio = read("downsample_ratio")
+    if downsample_ratio is not None:
+        merge_size = int(round(1.0 / float(downsample_ratio)))
+    else:
+        merge_size = int(read("merge_size", "spatial_merge_size") or 1)
+
+    return ImageProcessingConfig(
```
terrykong, line 191: [`nemo_rl/models/generation/megatron/utils.py:81`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/utils.py#L81)

**1 action item**, low severity.

`ImageProcessingConfig.vision_model_type` defaults to `"radio"` upstream and this builder never sets it. That is correct for Nemotron Omni, whose vision tower is RADIO, but it is correct by accident — the value comes from an upstream default this code does not state.

A second VLM family with a different tower would silently pick up `"radio"` and preprocess images the wrong way, with no error.

**Action:** set it explicitly.

```suggestion
    return ImageProcessingConfig(
        patch_dim=int(patch_dim),
        vision_model_type="radio",
```

If it should vary per model, read it off the HF config instead and raise when it cannot be determined — but please do not leave it implicit.

nemo_rl/environments/nemo_gym_multimodal.py
```diff
@@ -28,9 +31,19 @@
...
     extract_multimodal_model_inputs,
     get_dim_to_pack_along,
+    get_responses_content_part_url,
+    image_to_data_url,
+    media_sources_equal,
     resolve_to_image,
+    video_path_to_data_url,
+)
+from nemo_rl.environments.nemo_gym_request import (
+    _chat_template_kwargs_for_processor,
```
terrykong: [`nemo_gym_multimodal.py:44-46`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/environments/nemo_gym_multimodal.py#L44)

**1 action item**, low severity.

A leading underscore means "private to this module", but all four functions in `nemo_gym_request.py` are underscore-prefixed and three of them are imported here. The module ends up with no public surface at all, so someone reading `nemo_gym_request.py` has no way to tell these names are used elsewhere and that renaming one is not safe.

The same pattern appears at [`nemo_gym.py:42`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/environments/nemo_gym.py#L42), which imports `_index_per_turn_images` and `_without_initial_media_sources` from `nemo_gym_multimodal`.

**Action:** drop the leading underscore from the names that cross a module boundary (`json_mapping`, `deep_merge_dict`, `chat_template_kwargs_for_processor`, `index_per_turn_images`, `without_initial_media_sources`) and update the three import sites. No suggestion block — it is a rename across several files.

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -112,6 +117,85 @@ def _init_inference_engine_state(self) -> None:
...
+        ):
+            return model, None
+        return model.language_model, model
+
+    def _build_image_preprocessing_config(self, generation_config: dict[str, Any]):
+        """Build raw-image preprocessing settings."""
+        inference_wrapper_cls = self._get_megatron_inference_wrapper_cls()
+        if not self._wrapper_supports_modality(inference_wrapper_cls, "image"):
+            return None
+        processor = getattr(self, "processor", None)
```
terrykong: [`nemo_rl/models/generation/megatron/megatron_worker.py:182`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_worker.py#L182)

**1 action item**, low severity.

`self.processor` is never set by `MegatronGenerationMixin`. It is set by the class that mixes it in, at [`megatron_policy_worker.py:522`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/policy/workers/megatron_policy_worker.py#L522). Reading it through `getattr(..., None)` hides that dependency: nothing in the mixin declares that it needs a `processor`, so the requirement only shows up as a runtime `ValueError` two lines down.

**Action:** initialize `self.processor: Optional[Any] = None` in the mixin's own state setup ([`_init_inference_engine_state`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_worker.py#L106)) and read `self.processor` directly, so the attribute the mixin depends on is declared in one place.

nemo_rl/models/generation/megatron/config.py
```diff
@@ -48,13 +48,36 @@ class MCoreGenerationSpecificArgs(TypedDict):
...
+    video_target_num_patches: NotRequired[int]  # Overrides the image max-patch budget.
+    video_maintain_aspect_ratio: NotRequired[bool]
+
+    # Fully-qualified class path of the MCore inference wrapper, e.g.
+    # "megatron.core.inference.model_inference_wrappers.multimodal.
+    # nemotron_omni_inference_wrapper.NemotronOmniInferenceWrapper".
+    # Resolved by `_get_megatron_inference_wrapper_cls`; its `supports_*`
+    # attributes gate which modalities are preprocessed. Not media preprocessing
+    # itself, and used on the direct generate path as well as the HTTP endpoint.
+    megatron_inference_wrapper: NotRequired[str]
```
terrykong, line 77: [`nemo_rl/models/generation/megatron/config.py:79`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/config.py#L79)

**1 action item.**

`megatron_inference_wrapper` is the key that turns multimodal Megatron generation on, and it is not mentioned in the docs. `docs/design-docs/generation.md` already documents `mcore_generation_config`, and `docs/guides/nemo-gym-integration.md` covers the Gym HTTP path; neither says Megatron can now serve image or video requests. The PR checklist leaves the documentation box unchecked.

**Action:** add a short section to `docs/design-docs/generation.md` covering the multimodal Megatron path — which keys enable it (`megatron_inference_wrapper`, `image_dynamic_resolution`, and the `video_*` group), that `video_temporal_patch_size` is what switches the video path on and `video_num_frames` is then required, and that `expose_http_server` is what Gym dispatch uses. Not a suggestion block — the change is in a file outside this diff.

While you are there, two knobs are worth a sentence each because their defaults are easy to get wrong: `vision_embedding_cache_max_bytes` ships at `0`, which disables the cache that lets a prompt group reuse one vision-encoder pass, and `allow_stale_multimodal_embeddings` should stay `False` for RL, since the cache is otherwise served from pre-refit weights.

nemo_rl/models/generation/megatron/megatron_generation.py
terrykong: [`megatron_generation.py:161-162`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_generation.py#L161)

**1 action item**, docs only.

The docstring says [`"Exactly one of cluster or policy must be provided"`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/models/generation/megatron/megatron_generation.py#L156), but not which mode each one selects. That is the thing a reader actually needs, because the two arguments are how the class picks between colocated and dedicated inference — and the whole constructor branches on it.

The call site is the proof ([`grpo.py:1268`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/nemo_rl/algorithms/grpo.py#L1268)):

```python
policy=policy if colocated_inference else None,
```

So `policy` set means colocated (reuse the training model, share its GPUs), and `cluster` set means dedicated inference (stand up a separate Policy with an inference-shaped `megatron_cfg`). Naming that in the docstring saves the next reader a trip to `grpo.py`.

**Action:** say the mode in both lines:

```
    cluster: Cluster to deploy a dedicated inference Policy on
        (non-colocated inference).
    policy: Existing training Policy to reuse for generation
        (colocated inference).
```

No suggestion block — the docstring lines are not in this PR's diff, so there is nothing to attach one to.

tests/functional/nemotron_omni_clevr_megatron_1n4g.sh
```diff
@@ -0,0 +1,145 @@
...
+PY
+
+uv run --no-sync python examples/run_vlm_grpo.py \
+    --config examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n4g-megatron_generation.v1.yaml \
+    cluster.num_nodes=1 \
+    cluster.gpus_per_node=4 \
+    policy.megatron_cfg.env_vars.TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST}" \
+    policy.megatron_cfg.tensor_model_parallel_size=4 \
+    policy.megatron_cfg.expert_model_parallel_size=4 \
+    policy.megatron_cfg.expert_tensor_parallel_size=1 \
```
terrykong: [`nemotron_omni_clevr_megatron_1n4g.sh:84-88`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/nemotron_omni_clevr_megatron_1n4g.sh#L84)

**1 action item.**

This is the config side of the skip on [`L1_Functional_Tests_Megatron_Omni.sh:23-26`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/L1_Functional_Tests_Megatron_Omni.sh#L23). The functional shard has 2 GPUs, so at `gpus_per_node=4` this script never runs — the wrapper prints SKIP and exits 0, and CI goes green.

A test that never runs also rots quietly: nothing here is checked against the code, so it can drift out of date for months and the first person to get 4 GPUs inherits the debugging.

The repo already hit this once and gave up rather than retune — [`L1_Functional_Tests_GB200_MXFP8.sh:38-40`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/L1_Functional_Tests_GB200_MXFP8.sh#L38) disables its 4-GPU test with the note that the shard has 2 so "the inference placement group can never be satisfied" ([#3731](https://github.com/NVIDIA-NeMo/RL/issues/3731)). Worth not adding a second one.

**Action:** retune to 2 GPUs so it actually executes. Train side:

```suggestion
    cluster.gpus_per_node=2 \
    policy.megatron_cfg.env_vars.TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST}" \
    policy.megatron_cfg.tensor_model_parallel_size=2 \
    policy.megatron_cfg.expert_model_parallel_size=2 \
    policy.megatron_cfg.expert_tensor_parallel_size=1 \
```

and the same on the generation side — [`L100`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/nemotron_omni_clevr_megatron_1n4g.sh#L100) `colocated.resources.gpus_per_node=2`, [`L102`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/nemotron_omni_clevr_megatron_1n4g.sh#L102) and [`L103`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/nemotron_omni_clevr_megatron_1n4g.sh#L103) to `tensor_model_parallel_size=2` / `expert_model_parallel_size=2`. [`nemotron_omni_gym_video_megatron_1n4g.sh:72`](https://github.com/NVIDIA-NeMo/RL/blob/73f611768086558b1d51031d71159809b44963fc/tests/functional/nemotron_omni_gym_video_megatron_1n4g.sh#L72) needs the same treatment. The suggestion only covers the contiguous train-side block; the rest is not adjacent so it cannot go in one.

If 30B-A3B does not fit at TP=2, then this belongs in nightly rather than L1 — you are better placed to call that than I am.

terrykong (commented):
Re-review of the new commits since `63c67a0e` (this branch was rebased and gained [`73f6117`](https://github.com/NVIDIA-NeMo/RL/commit/73f611768086558b1d51031d71159809b44963fc) "Address Claude and Fable reviews"). Reviewed at head `73f61176`, with every finding below re-anchored and re-verified against this head.

**Credit where it is due — `73f6117` fixed a lot**, several exactly as they should have been: the duplicate `logprobs_mode` declaration is gone and the direct `[...]` read is back; the `.get(k, default)` call-site defaults were replaced with a conditional override dict; the three new unit tests got `@pytest.mark.mcore`; `build_video_preprocessing_config` now raises a named error instead of a bare `KeyError`; `install_audio_deps.sh` now uses `uv pip install --python` and fails loudly when the worker venv is missing; and `_sample_vision_tensors` gained an explicit guard against de-duplicated `PackedTensor` media. The `emerging-optimizers` revert plus re-lock is correct and consistent with the rebased submodule pin — I checked the lock against Megatron-LM `731b7914` and it matches.

Two things stand out in what is left:

- **The submodule bump is still the gate.** `ImageProcessingConfig` / `VideoProcessingConfig` do not exist at the current pin, so Megatron cannot import until Megatron-Bridge moves. Left as a short placeholder on `utils.py:19` so it does not get lost.
- **The new functional test cannot run on the shard it targets.** It asks for 4 GPUs, the functional shard has 2, so the wrapper's own guard trips and the job exits 0 — green in CI, having executed nothing. Two comments cover this, with a suggested 2-GPU retune.

Merge state: the branch is currently **conflicting with `main`** and needs a rebase. The `/ok to test` comment points at `63f916f`, which is no longer in the branch, so a fresh `/ok to test 73f6117` is worth running once the rebase lands — no unit or functional CI has run on this head.

Findings that did not survive checking are not included. Notably, the concern that `multimodal_prompt_config` would raise on a plain `GPTInferenceWrapper` is **not** real — it is declared as a `ClassVar` defaulting to `None` on `AbstractModelInferenceWrapper`, so every wrapper inherits it.

_Generated by Claude Code_

tests/functional/nemotron_omni_clevr_megatron_1n4g.sh
cspades: Yeah, I updated the functional test to just do colocated 2 GPUs. It'll fit GB200.

nemo_rl/models/generation/megatron/megatron_worker.py
cspades: Yeah, it's been promoted to a member of the mixin instead of hacked in.

nemo_rl/models/generation/megatron/utils.py
cspades: Nope, don't like this. The wrapper and/or MCore is expected to know the multimodal data format, and really RL has no part in touching this specific part of the MCore config, unless it wants to strongly imply a vision encoding setup. Let's not hard-code a specific model just because it's the only one we support at the moment.

nemo_rl/models/generation/megatron/megatron_worker.py
cspades: This was a good change, I just refactored the functions from the self methods.

tests/functional/L1_Functional_Tests_Megatron_Omni.sh
cspades: Dupe of another comment below, I've changed them to use 2 GPUs.

examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n4g-megatron_generation.v1.yaml
cspades: I don't have bandwidth to get a non-SFT config and dataset so that the training loss isn't effectively a flat line. Added a `TODO(@cspades)` and `nightly_disable.txt` with temporary scripts that have no check and I'll add these functional tests later.

cspades:
/ok to test [7ef4ec3](https://github.com/NVIDIA-NeMo/RL/pull/3834/commits/7ef4ec31657979d558a5fe93245d7169bffc05dd)

nemo_rl/models/generation/megatron/utils.py
cspades: Oh wait, this is actually supported across a variety of encoders. I'll expose this in the RL mcore gen config.

cspades:
Migrating to NRL local branches.

cspades:
Moved to: https://github.com/NVIDIA-NeMo/RL/pull/3920

nemo_rl/models/generation/megatron/utils.py
terrykong: fixed in a0edb2761c4d93125c54bd2fc56177d4a1b99b06

The pin now resolves through Megatron-Bridge b11414c7 to Megatron-LM 7c9c3a02, which carries both classes. Landed on main via #3906 rather than in the multimodal branch itself.

(Carried over in #3920, which supersedes this PR.)

examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-clevr-1n4g-megatron_generation.v1.yaml
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

tests/functional/L1_Functional_Tests_Megatron_Omni.sh
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

nemo_rl/data/processors.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

nemo_rl/models/generation/megatron/megatron_generation.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

nemo_rl/environments/nemo_gym_request.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

tests/unit/data/test_multimodal_image_encoding.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

nemo_rl/models/generation/megatron/utils.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

nemo_rl/models/generation/megatron/megatron_worker.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

nemo_rl/models/generation/megatron/utils.py
terrykong: fixed in 1d8ee81a93269ae3c61f194435547561fc02331a

(Carried over in #3920, which supersedes this PR.)

nemo_rl/models/generation/megatron/megatron_worker.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

nemo_rl/models/generation/megatron/config.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

nemo_rl/models/generation/megatron/megatron_generation.py
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

tests/functional/nemotron_omni_clevr_megatron_1n4g.sh
terrykong: fixed in 1a207e5abf55d8c1046ec08273cffa6816d20ff0

(Carried over in #3920, which supersedes this PR.)

PR #3839, opened 2026-08-26
feat(vllm): pause generation during in-flight refit

youngeunkwon0405:
## Summary

- route in-flight refit coordination through vLLM native keep-mode pause/resume
- pause every async vLLM engine before weight transfer while preserving request state
- map recompute_kv_cache_after_weight_updates to the native clear_cache option
- keep trajectory admission closed when native resume fails

## Implementation

The trajectory collector still stops new rollout batches first. For async vLLM with in_flight_weight_updates enabled, VllmGeneration broadcasts pause_generation_async to every DP engine leader. Each worker calls AsyncLLM.pause_generation(mode="keep", clear_cache=...). After refit, the collector broadcasts resume_generation_async before admitting new rollout work.

Other backends and the drain-before-refit path retain their existing behavior.

## Validation
Post-refactor real-GPU validation completed successfully for both KV-cache modes.

Common setup:

- Recipe: `grpo-qwen3-30ba3b-4n8g-async-1off` (non-Gym perf path)
- Scope: 10 training steps with generation logging enabled
- Recipe dimensions unchanged: `train_global_batch_size=2048`, `max_total_sequence_length=4096`
- Cluster: cw-dfw, 4 nodes / 32 GPUs
- Slurm account: `coreai_dlalgo_genai`

### Preserve existing KV cache (`recompute_kv_cache_after_weight_updates=false`)

- Slurm: job `16780048`, `COMPLETED` (`0:0`) in `00:33:54`
- Refit lifecycle: 10 pause requests, 10 `Ready for refit` transitions, and 10 resume requests
- Completion: reached step 10/10 and logged `Async GRPO training complete!`
- Step 10: loss `0.0077`, average reward `0.5942`, validation accuracy `0.5430`
- Failure markers: none
- W&B: https://wandb.ai/nvidia/async-grpo-pr-validation/runs/woa745di

### Recompute KV cache after weight updates (`recompute_kv_cache_after_weight_updates=true`)

- Slurm: job `16783855`, `COMPLETED` (`0:0`) in `00:34:20`
- Resolved mode: `kv_cache_management_mode` changed from `persist` to `recompute`
- Refit lifecycle: 10 pause requests, 10 `Ready for refit` transitions, and 10 resume requests
- Completion: reached step 10/10 and logged `Async GRPO training complete!`
- Step 10: loss `-0.0018`, average reward `0.5942`, validation accuracy `0.5430`
- Failure markers: none
- W&B: https://wandb.ai/nvidia/async-grpo-pr-validation/runs/j872mzhi

Together these runs exercise the final backend-agnostic collector contract and the vLLM native pause/resume implementation with both `clear_cache=false` and `clear_cache=true`.


youngeunkwon0405:
## GPU validation

Validated exact PR head `abf3b2a37416a56bfc469d25b0dffafd84024d15` on CW-DFW with `/lustre/fs1/portfolios/coreai/projects/coreai_dlalgo_nemorl/users/youngeunk/sqsh/nemo_rl.0825.sqsh` and the `coreai_dlalgo_llm` account.

### Full non-Gym performance recipe (10 steps, generation logging enabled)

- Recipe/entrypoint: `examples/configs/recipes/llm/performance/grpo-qwen3-30ba3b-4n8g-async-1off.yaml` via `examples/run_grpo.py`
- Slurm job: `16641972` — `COMPLETED`, exit `0:0`, elapsed `00:33:48` (4 nodes / 32 H100s)
- W&B: https://wandb.ai/nvidia/async-grpo-pr-validation/runs/hm6vr3e8
- Kept the recipe's full workload unchanged: Qwen3-30B-A3B, 64 prompts/step × 32 generations/prompt, global batch size 2048, max sequence length 4096, two 8-GPU generation nodes plus two 8-GPU training nodes, async vLLM, and in-flight weight updates
- No batch-size, sequence-length, model, topology, validation, or checkpointing reductions. Overrides were limited to `grpo.max_num_steps=10`, explicitly enabling the existing vLLM metrics logger, and setting log/W&B destinations.
- Completed 10/10 training steps and the unchanged step-10 validation (`256` samples, accuracy `0.5508`), followed by `Async GRPO training complete!` and `ray.sub exiting (exit_code=0)`
- Observed 10 matched native lifecycle cycles: 10 `Pausing vLLM generation with in-flight request state preserved` and 10 `Resuming vLLM generation after refit`
- In-flight preservation was exercised repeatedly: the first pause reported 0 ongoing batches; the other 9 pauses each reported `1 ongoing generation batches paused`
- vLLM generation metrics were emitted for all 10 steps
- Full generation logging produced `train_data_step1.jsonl` through `train_data_step10.jsonl` (10 non-empty files, 3,435,333,916 bytes / 3.20 GiB total), plus `val_data_step10.jsonl`
- No `Traceback`, `Exception`, or `ERROR` entries were present in the driver log

### Additional integration coverage

- NeMo Gym async 10-step smoke: job `16641492`, `COMPLETED`/`0:0`; W&B: https://wandb.ai/nvidia/async-grpo-pr-validation/runs/wwpnjrp7
- Earlier full-size 2-step confirmation: job `16639865`, `COMPLETED`/`0:0`; W&B: https://wandb.ai/nvidia/async-grpo-pr-validation/runs/i49wh310

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -973,15 +982,24 @@ def prepare_for_refit(self) -> None:
...
-            # async engines support in-flight weight updates
-            # Ongoing generations will continue with their current KV caches
-            # New generations (after weight update) will use the updated weights
-            print(
-                f"🚀 Using {backend} in-flight weight update - skipping wait for pending generations"
-            )
-            print(
-                f"   {len(self._inflight_threads)} ongoing generations will complete with current weights"
-            )
+            if self._vllm_inflight_weight_update():
```
youngeunkwon0405: [`trajectory_collector.py:985`](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/algorithms/async_utils/trajectory_collector.py#L985)

**[PERF-EVIDENCE]** The 10-step 32xH100 validation run is great evidence the path works end-to-end. Since this changes the performance shape of an existing feature — engines now pause for the transfer window instead of generating through it, and `recompute_kv_cache_after_weight_updates=true` goes from a silent no-op for in-flight requests ([vLLM `reset_prefix_cache` only resets when no running request holds KV](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/core/sched/scheduler.py#L2199-L2204)) to a full re-prefill of every preserved request — it would be helpful to also share:
- before/after step time & generation throughput on the same recipe vs `main` (ideally with the weight-sync / refit-bubble timer breakdown, so the new pause window is visible), and
- a convergence sanity check for both `recompute_kv_cache_after_weight_updates` settings, since both settings' semantics change on this path.

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -929,12 +929,21 @@ def resume(self) -> None:
...
+
     def prepare_for_refit(self) -> None:
         """Pause new generation starts and optionally wait for pending generations.
 
-        For backends with an async engine in-flight weight updates allows ongoing generations
-        to continue with their current KV caches while weights are updated.
-        This significantly improves async performance.
+        For async vLLM, native keep-mode pause preserves in-flight request state
+        across the weight update. Other async backends keep their existing in-flight
+        refit behavior. This significantly improves async performance.
```
youngeunkwon0405: [`trajectory_collector.py:944-946`](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/algorithms/async_utils/trajectory_collector.py#L944-L946)

**[DOC]** [`docs/guides/async-grpo.md` tips 4-5](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/docs/guides/async-grpo.md#L208-L210) still describe the old vLLM semantics: tip 4 says in-flight updates "update engine weights during generation" (vLLM engines now pause, with requests preserved rather than running), and tip 5 says recompute "invalidate[s] ... after weight updates ... independent of in-flight updates" (on this path it now happens at pause time via `clear_cache`, and also clears in-flight requests' KV). Consider updating both bullets alongside this PR.

Minor: the retained sentence below now dangles between the two divergent behaviors it used to describe:
```suggestion
        For async vLLM, native keep-mode pause preserves in-flight request state
        across the weight update. Other async backends keep their existing in-flight
        refit behavior. Both paths avoid stalling the training pipeline on the
        longest in-flight generation.
```

nemo_rl/models/generation/interfaces.py
```diff
@@ -487,6 +487,17 @@ def attach_fleet_health(self, monitor: Any, selector: Any) -> None:
     def invalidate_kv_cache(self) -> bool:
         return False
 
+    def pause_generation(self, *, clear_cache: bool) -> bool:
+        """Pause in-flight generation while preserving request state.
+
+        Backends with native in-flight refit support may override this hook.
+        """
+        return False
```
youngeunkwon0405: [`interfaces.py:490-495`](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/models/generation/interfaces.py#L490-L495)

**[DOCSTRING]** New public hook whose `clear_cache` kwarg and bool return drive collector behavior (a `False` becomes a `RuntimeError` in `prepare_for_refit`) — worth full Args/Returns per the Google-style guideline, like sibling `attach_fleet_health`. Wording below is verified against vLLM v0.25.1 [pause semantics](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/engine/async_llm.py#L750-L793):
```suggestion
    def pause_generation(self, *, clear_cache: bool) -> bool:
        """Pause in-flight generation while preserving request state.

        Backends with native in-flight refit support may override this hook.
        In-flight requests are frozen rather than aborted and resume from
        :meth:`resume_generation`; new requests queue until then.

        Args:
            clear_cache: Also clear the engine's reusable caches (KV/prefix)
                at pause time so preserved requests recompute their KV after
                the weight update instead of resuming with pre-update cache.

        Returns:
            True if every engine paused; False when the backend has no
            native pause support (callers treat False as a failed pause).
        """
        return False
```

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -929,12 +929,21 @@ def resume(self) -> None:
...
         print("Trajectory collection resumed")
 
+    def _vllm_inflight_weight_update(self) -> bool:
+        """Whether refit uses vLLM's in-flight weight-update path."""
+        generation_cfg = self.master_config.policy.get("generation", {})
+        return (
+            generation_cfg.get("backend") == "vllm"
+            and generation_cfg.get("vllm_cfg", {}).get("async_engine", False)
+            and self.async_config.in_flight_weight_updates
+        )
```
youngeunkwon0405: [`trajectory_collector.py:932-939`](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/algorithms/async_utils/trajectory_collector.py#L932-L939)

**[GUIDELINE]** `generation_cfg.get("vllm_cfg", {}).get("async_engine", False)` is the config-conventions "chained `.get()` with hidden defaults" pattern — `async_engine` is a required key of [`VllmSpecificArgs`](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/models/generation/vllm/config.py#L52) — and this predicate now exists in three places (here, `prepare_for_refit`, and [`should_use_async_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/models/generation/interfaces.py#L243-L244)). Reusing the shared helper centralizes it; the `backend == "vllm"` guard must stay since `should_use_async_rollouts` is also true for megatron/trtllm/dynamo. Needs `should_use_async_rollouts` added to the existing `interfaces` import at L61. (The similar inline derivation in `prepare_for_refit` is pre-existing — low severity.)
```suggestion
    def _vllm_inflight_weight_update(self) -> bool:
        """Whether refit uses vLLM's in-flight weight-update path."""
        generation_cfg = self.master_config.policy.get("generation")
        return (
            generation_cfg is not None
            and generation_cfg.get("backend") == "vllm"
            and should_use_async_rollouts(generation_cfg)
            and self.async_config.in_flight_weight_updates
        )
```

tests/unit/algorithms/test_async_utils.py
```diff
@@ -2383,6 +2393,66 @@ def test_async_trajectory_collector_prepare_for_refit(self):
...
+        collector.wait_for_pending_generations.assert_not_called()
+        assert not collector._refit_pause_cleared.is_set()
+
+        collector.resume_after_refit()
+
+        assert collector.policy_generation.resume_generation_calls == 1
+        collector.policy_generation.invalidate_kv_cache.assert_not_called()
+        assert collector._refit_pause_cleared.is_set()
+
+    def test_vllm_refit_drains_without_native_pause_when_in_flight_disabled(
```
youngeunkwon0405, line 2460: [`test_async_utils.py:2425`](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/tests/unit/algorithms/test_async_utils.py#L2425)

**[TEST]** The branch this PR deliberately preserves — non-vLLM async backend with `in_flight_weight_updates=True` (shipping configs, e.g. [grpo-nanov3-30BA3B-2n8g-megatron_generation-async-gym.yaml](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/examples/configs/recipes/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation-async-gym.yaml)) — has no pin, so a future gating regression (e.g. the vLLM predicate accidentally capturing megatron) would be silent. A 10-line sibling of this test covers it (assertions verified against `prepare_for_refit`; passes against the extracted source, not yet run under real pytest):
```python
    def test_non_vllm_async_backend_keeps_legacy_in_flight_path(self) -> None:
        collector = self.create_local_collector()
        collector.master_config.policy["generation"] = {"backend": "megatron"}
        collector.master_config.grpo.async_grpo.in_flight_weight_updates = True
        collector.wait_for_pending_generations = mock.Mock()

        collector.prepare_for_refit()

        assert collector.policy_generation.pause_generation_calls == []
        collector.wait_for_pending_generations.assert_not_called()
        assert not collector._refit_pause_cleared.is_set()
```

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -996,6 +1014,13 @@ def resume_after_refit(self) -> None:
         """Resume new generation starts after refit is complete."""
         print("🔄 Resuming generation starts after refit")
 
+        if self._vllm_inflight_weight_update():
```
youngeunkwon0405: [`trajectory_collector.py:1017`](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1017)

**[INFO — no change requested]** Async-PPO critic-warmup steps call `prepare_for_refit`/`resume_after_refit` with the policy transfer skipped ([ppo.py:2617-2651](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/algorithms/ppo.py#L2617-L2651)): previously effectively a no-op on this path, each warmup step now pauses+resumes every engine — and with `recompute=true`, [preempts all in-flight requests and clears caches](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/core/sched/scheduler.py#L2196-L2245) against unchanged weights (wasted re-prefill). Bounded to warmup steps, and a fix would need an is-warmup signal plumbed into the collector — flagging for awareness / possible follow-up. (GRPO has no analogous call site.)

youngeunkwon0405 (commented):
Team review by 5 coordinated agents (RL-code, vLLM-upstream verification, tests, independent bug scan, adversarial verification). **No critical findings** — 5 suggestions and 1 informational note posted inline. No merge conflicts; CI already triggered on this head.

**Upstream verification (vLLM v0.25.1, pinned):** `AsyncLLM.pause_generation(*, mode="keep", clear_cache=...)` / `resume_generation()` usage is exactly right — [signature](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/engine/async_llm.py#L750-L756), [`PauseMode = Literal["abort","wait","keep"]`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/engine/__init__.py#L26). Keep-mode pause [quiesces the engine before returning](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/engine/core.py#L1665-L1713) (idle-callback future), and weight-update `collective_rpc` utility calls [still execute while paused](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/engine/core.py#L1259-L1298) — so the weight write can no longer race an in-flight forward pass. The broadcast axes (`run_rank_0_only_axes=["tensor_parallel", "pipeline_parallel"]`) match every sibling engine-wide op.

**Nice side effect worth calling out:** the old `recompute_kv_cache_after_weight_updates=true` in-flight path was effectively a silent no-op — [`reset_prefix_cache` refuses to reset while requests hold KV blocks](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/core/block_pool.py#L656-L672), [`reset_prefix_cache_async` discards the returned False](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/models/generation/vllm/vllm_worker_async.py#L1556-L1569), and [`invalidate_kv_cache`'s `all()` over the filtered results](https://github.com/NVIDIA-NeMo/RL/blob/abf3b2a37416a56bfc469d25b0dffafd84024d15/nemo_rl/models/generation/vllm/vllm_generation.py#L1260-L1280) then reported success. This PR is the first path that actually delivers in-flight KV recompute, via pause-time `clear_cache` ([preempts running requests and frees their KV](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/core/sched/scheduler.py#L2196-L2245)).

Minor non-actionable note: `pause_generation`/`resume_generation` raise on misconfig and propagate `RayTaskError` on worker failure, so the interface's `False` contract (and the collector's "Failed to pause" branch) is unreachable for the real backend — fail-loud is right per the error-handling guidelines; just a small signaling inconsistency with `invalidate_kv_cache`'s catch-and-return-False.

Tests: all 5 new tests verified against the extracted source (simulation harness; runtime confirmation left to CI as noted in the PR description) and are correctly bucketed (Algorithms + vLLM base shards; `pytest.mark.asyncio` strict-mode marker correct). Linter: ruff format + check pass on all 6 changed files.

_Generated by Claude Code_

nemo_rl/algorithms/async_utils/trajectory_collector.py
youngeunkwon0405: Acknowledged. I agree this is a bounded critic-warmup inefficiency. I am leaving it unchanged here because avoiding the pause requires plumbing an explicit no-policy-refit or warmup signal through the PPO control flow, which is broader than this PR.

nemo_rl/algorithms/async_utils/trajectory_collector.py
youngeunkwon0405: Addressed in c815f32a8. The async-GRPO guide now documents preserved-request pause semantics and pause-time cache clearing, and the collector docstring now describes the two backend behaviors without the dangling performance sentence.

nemo_rl/algorithms/async_utils/trajectory_collector.py
youngeunkwon0405: Agreed that matched before/after performance and convergence runs would be useful characterization, but I am keeping them outside this correctness-focused PR. The existing full 10-step recipe run validates the native pause path end to end; matched main and both recompute settings would require a separate controlled performance study. No code change for this suggestion.

tests/unit/algorithms/test_async_utils.py
youngeunkwon0405: Addressed in c815f32a8. Added the non-vLLM async in-flight regression test. It passed in the provided CW-DFW container as part of Slurm job 16663421.

nemo_rl/models/generation/interfaces.py
youngeunkwon0405: Addressed in c815f32a8. Added Args and Returns documentation for pause_generation and a matching Returns section for resume_generation.

nemo_rl/algorithms/async_utils/trajectory_collector.py
youngeunkwon0405: Addressed in c815f32a8. The predicate now reuses should_use_async_rollouts while retaining the explicit vLLM backend guard; generation is a required normalized config entry, so it uses direct access.

youngeunkwon0405:
Claude review follow-up is in `c815f32a847f3d2b9a7c4eaad32fc8b7299023a6`.

Addressed:
- updated async-GRPO documentation for keep-mode pause and pause-time cache clearing;
- expanded public pause/resume hook docstrings;
- reused `should_use_async_rollouts` in the vLLM in-flight predicate;
- added non-vLLM legacy-path and vLLM pause-failure regression coverage.

Validation on CW-DFW:
- Slurm job `16663421`: `COMPLETED`, exit `0:0`;
- focused pytest: 6 passed, 101 deselected;
- Ruff check passed; 3 Python files already formatted;
- `git diff --check` passed.

Scope decisions are recorded in the inline threads: matched-main performance/convergence comparisons are useful follow-up characterization rather than a merge gate for this correctness PR, and the bounded PPO critic-warmup pause inefficiency needs broader control-flow signaling. The existing full 10-step recipe runtime validation was run on `abf3b2a37416a56bfc469d25b0dffafd84024d15`: https://wandb.ai/nvidia/async-grpo-pr-validation/runs/hm6vr3e8.

youngeunkwon0405:
Follow-up design update in `74b9063cbf29db2272e4aec52b1a916e50985285`:

- `AsyncTrajectoryCollector` now invokes `pause_generation(clear_cache=...)` and `resume_generation()` for every async backend using in-flight weight updates.
- Backend dispatch stays in the generation interface: vLLM overrides the hooks with native keep-mode pause/resume; the default implementation warns and returns unsupported, so Megatron/TRT-LLM retain their existing overlap behavior without an error.
- The collector records whether a pause actually succeeded. Unsupported (`False`) continues as requested; a backend that reports a successful pause must resume successfully, otherwise the collector remains paused and fails loudly.
- Managed Dynamo keeps its existing drain path because its configuration rejects in-flight refit.
- Updated async-GRPO documentation and added coverage for unsupported-backend warnings, successful-pause/resume pairing, propagated pause errors, and partial vLLM worker failures.

CW-DFW validation: Slurm `16664018` completed with exit `0:0`; 13 focused tests passed, 149 deselected; Ruff 0.9.9 check/format and `git diff --check` passed.

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -943,24 +949,14 @@ def prepare_for_refit(self) -> None:
...
-        elif backend == "megatron":
-            is_async_engine = True
-        elif backend == "trtllm":
-            assert generation_cfg.get("trtllm_cfg", {}).get("async_engine", False), (
-                "TRT-LLM backend requires trtllm_cfg.async_engine=true; the "
-                "synchronous engine path (async_engine=false) is no longer supported."
-            )
-            is_async_engine = True
-        elif backend == "dynamo":
+        generation_cfg = self.master_config.policy["generation"]
```
youngeunkwon0405, line 959: [`trajectory_collector.py:957`](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/algorithms/async_utils/trajectory_collector.py#L957)

**[TEST-REGRESSION]** The direct `policy["generation"]` access is right for production (grpo.py [reads it before the collector exists](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/algorithms/grpo.py#L4224), run_ppo.py [validates it](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/examples/run_ppo.py#L52-L60)) — but it breaks the one pre-existing refit test whose fixture lacks the key: [`create_mock_config`](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/tests/unit/algorithms/test_async_utils.py#L2189-L2192) builds `policy` without `"generation"`, so [`test_async_trajectory_collector_prepare_for_refit`](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/tests/unit/algorithms/test_async_utils.py#L2396) now fails with `KeyError: 'generation'` inside the Ray actor (execution-confirmed: fails at head, passes at base; the focused Slurm runs deselected this test, so a full `tests/unit` run will go red). Suggested fix in `tests/unit/algorithms/test_async_utils.py` — keeps the fail-fast access and the test's original drain-path behavior (verified):
```python
            policy={
                "max_total_sequence_length": 512,
                "make_sequence_length_divisible_by": 1,
                "generation": {"backend": "vllm", "vllm_cfg": {"async_engine": False}},
            },
```
Cosmetic side note: [`resume_after_refit`](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1035) still uses the tolerant `"generation" in self.master_config.policy` guard while this line assumes presence — worth picking one style.

docs/guides/async-grpo.md
```diff
@@ -205,9 +205,9 @@ If no `replay_buffer.pt` file is found in the latest checkpoint directory, train
 
 3. **Resource Allocation**: Ensure sufficient GPU memory for both the training and generation clusters
 
-4. **In-Flight Weight Updates**: Enable `in_flight_weight_updates: true` to update engine weights during generation; with vLLM this requires `async_engine: true`, while the Megatron backend is always async-engine. This prevents stalling the training pipeline until the longest generation finishes and provides significant performance benefits.
+4. **In-Flight Weight Updates**: Enable `in_flight_weight_updates: true` to refit without waiting for the longest in-flight generation to finish. The collector requests a generation pause and resume from every async backend around the weight transfer. Async vLLM implements this contract while preserving request state. Backends without native pause/resume support emit a warning and retain their existing in-flight update behavior. vLLM requires `async_engine: true`; the Megatron backend is always async-engine.
```
youngeunkwon0405: [`async-grpo.md:208`](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/docs/guides/async-grpo.md#L208)

**[DOC]** Tip 4 says the collector requests pause/resume "from every async backend", but Dynamo — async per `should_use_async_rollouts` — is [deliberately carved out and force-drained](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/algorithms/async_utils/trajectory_collector.py#L959-L966). Same imprecision in the [`prepare_for_refit` docstring](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/algorithms/async_utils/trajectory_collector.py#L941) ("Every async backend is asked to pause"), though the in-function comment explains the carve-out.
```suggestion
4. **In-Flight Weight Updates**: Enable `in_flight_weight_updates: true` to refit without waiting for the longest in-flight generation to finish. The collector requests a generation pause and resume from every async backend around the weight transfer (Managed Dynamo excepted: it always drains in-flight requests before refit because its layerwise reload cannot run concurrently with generation). Async vLLM implements this contract while preserving request state. Backends without native pause/resume support emit a warning and retain their existing in-flight update behavior. vLLM requires `async_engine: true`; the Megatron backend is always async-engine.
```

nemo_rl/models/generation/interfaces.py
```diff
@@ -487,6 +487,45 @@ def attach_fleet_health(self, monitor: Any, selector: Any) -> None:
...
+                preserved requests recompute their KV after the weight update.
+
+        Returns:
+            True if every engine paused; False when the backend has no native pause
+            support. Backends with native support raise when pausing fails.
+        """
+        print(
+            f"⚠️ {type(self).__name__} does not support pausing generation during "
+            "in-flight weight updates; continuing without native pause"
+        )
```
youngeunkwon0405: [`interfaces.py:507-510`](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/models/generation/interfaces.py#L507-L510)

**[GUIDELINE, low]** These defaults print a ⚠️ on **every refit** for megatron/trtllm in-flight configs that [the guide describes as supported expected behavior](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/docs/guides/async-grpo.md#L208) — and colocated megatron calls `prepare_for_refit` twice per non-save step ([`blocks_training` site](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/algorithms/grpo.py#L4947-L4950) + [refit site](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/algorithms/grpo.py#L5132)), so with the [resume twin](https://github.com/NVIDIA-NeMo/RL/blob/74b9063cbf29db2272e4aec52b1a916e50985285/nemo_rl/models/generation/interfaces.py#L523-L526) that's 3 warning lines per training step, phrased like a misconfiguration alarm. Consider neutral phrasing (below) and/or warn-once-per-process; same for the resume hook.
```suggestion
        print(
            f"{type(self).__name__} has no native pause support; refit continues "
            "with this backend's existing in-flight update behavior"
        )
```

nemo_rl/algorithms/async_utils/trajectory_collector.py
youngeunkwon0405: Confirmed and fixed in f94fd34b1. The shared mock config now includes the required synchronous vLLM generation entry, and the managed-Dynamo invalidation check now uses the same direct required-config access. The previously failing test_async_trajectory_collector_prepare_for_refit passed in CW-DFW Slurm job 16757596.

docs/guides/async-grpo.md
youngeunkwon0405: Addressed in f94fd34b1. Both the async-GRPO guide and the collector docstring now explicitly carve out managed Dynamo and document that it drains active trajectories before refit.

nemo_rl/models/generation/interfaces.py
youngeunkwon0405: Addressed in f94fd34b1. Unsupported backends now emit one neutral fallback warning per backend type and process; pause and resume share the same cached warning. The regression test verifies that a full pause/resume cycle emits it only once.

youngeunkwon0405:
Addressed the second Claude review in f94fd34b1:

- repaired the pre-existing Ray refit fixture by adding its required generation config
- documented the managed-Dynamo drain exception in both the guide and collector docstring
- changed unsupported-backend pause/resume logging to one neutral warning per backend type/process
- made the managed-Dynamo invalidation check consistently use the required normalized config

Validation: CW-DFW Slurm job 16757596 completed 0:0 in the requested nemo_rl.0825.sqsh image. All 14 focused tests passed, including test_async_trajectory_collector_prepare_for_refit, and Ruff 0.9.9 check plus format verification passed.

youngeunkwon0405:
Updated the branch to current main to clear the freshness gate. The previous failure was because the PR was 12 commits behind while CI allows at most 10.\n\nCurrent main introduced a separate pause_generation(mode) / continue_generation() contract for the SGLang WeightSynchronizer. To avoid a semantic signature collision, this PR now uses the refit-specific hooks pause_generation_for_refit(clear_cache=...) and resume_generation_after_refit(), leaving the SGLang WeightSynchronizer API unchanged.\n\nValidation on current main: CW-DFW Slurm job 16764115 completed 0:0; 14 focused async-refit/vLLM tests and all 57 WeightSynchronizer tests passed; Ruff 0.9.9 check and format verification passed.

youngeunkwon0405:
Hi @terrykong, I did the self-review for this PR line by line. Can I ask for your final review for the merge?

youngeunkwon0405:
Post-refactor real-GPU validation completed successfully for both KV-cache modes.

Common setup:

- Recipe: `grpo-qwen3-30ba3b-4n8g-async-1off` (non-Gym perf path)
- Scope: 10 training steps with generation logging enabled
- Recipe dimensions unchanged: `train_global_batch_size=2048`, `max_total_sequence_length=4096`
- Cluster: cw-dfw, 4 nodes / 32 GPUs
- Slurm account: `coreai_dlalgo_genai`

### Preserve existing KV cache (`recompute_kv_cache_after_weight_updates=false`)

- Slurm: job `16780048`, `COMPLETED` (`0:0`) in `00:33:54`
- Refit lifecycle: 10 pause requests, 10 `Ready for refit` transitions, and 10 resume requests
- Completion: reached step 10/10 and logged `Async GRPO training complete!`
- Step 10: loss `0.0077`, average reward `0.5942`, validation accuracy `0.5430`
- Failure markers: none
- W&B: https://wandb.ai/nvidia/async-grpo-pr-validation/runs/woa745di

### Recompute KV cache after weight updates (`recompute_kv_cache_after_weight_updates=true`)

- Slurm: job `16783855`, `COMPLETED` (`0:0`) in `00:34:20`
- Resolved mode: `kv_cache_management_mode` changed from `persist` to `recompute`
- Refit lifecycle: 10 pause requests, 10 `Ready for refit` transitions, and 10 resume requests
- Completion: reached step 10/10 and logged `Async GRPO training complete!`
- Step 10: loss `-0.0018`, average reward `0.5942`, validation accuracy `0.5430`
- Failure markers: none
- W&B: https://wandb.ai/nvidia/async-grpo-pr-validation/runs/j872mzhi

Together these runs exercise the final backend-agnostic collector contract and the vLLM native pause/resume implementation with both `clear_cache=false` and `clear_cache=true`.

docs/guides/async-grpo.md
```diff
@@ -205,9 +205,9 @@ If no `replay_buffer.pt` file is found in the latest checkpoint directory, train
 
 3. **Resource Allocation**: Ensure sufficient GPU memory for both the training and generation clusters
 
-4. **In-Flight Weight Updates**: Enable `in_flight_weight_updates: true` to update engine weights during generation; with vLLM this requires `async_engine: true`, while the Megatron backend is always async-engine. This prevents stalling the training pipeline until the longest generation finishes and provides significant performance benefits.
+4. **In-Flight Weight Updates**: Enable `in_flight_weight_updates: true` to refit without waiting for the longest in-flight generation to finish. Except for managed Dynamo, the collector requests a generation pause and resume from every async backend around the weight transfer. Async vLLM implements this contract while preserving request state. Backends without native pause/resume support emit a once-per-backend warning and retain their existing in-flight update behavior. Managed Dynamo always drains active trajectories before refit. vLLM requires `async_engine: true`; the Megatron backend is always async-engine.
```
terrykong: [`docs/guides/async-grpo.md:208`](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/docs/guides/async-grpo.md#L208)

**1 action item.**

The sentence "Backends without native pause/resume support emit a once-per-backend warning and retain their existing in-flight update behavior" is inaccurate for SGLang — the one backend whose behavior [this PR changes](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/nemo_rl/algorithms/async_utils/trajectory_collector.py#L969), so it is the case a reader most needs the docs to get right.

Both halves are off:
- SGLang *does* have native pause/resume — [`pause_generation(mode)` / `continue_generation()`](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/nemo_rl/models/generation/sglang/sglang_generation.py#L329-L345). It just does not implement the new refit hook, so it takes the warning path anyway.
- Its "existing in-flight update behavior" *was* draining through the collector, which is exactly what it no longer does.

**Action:** replace the sentence. One way to do it — of several — keeping the rest of the tip as-is:

```suggestion
4. **In-Flight Weight Updates**: Enable `in_flight_weight_updates: true` to refit without waiting for the longest in-flight generation to finish. Except for managed Dynamo, the collector requests a generation pause and resume from every async backend around the weight transfer. Async vLLM implements this contract while preserving request state. A backend that does not implement the hook emits a warning once per backend type per process and refits without a collector-side pause or drain; SGLang is in this group today and instead relies on the pause its own weight synchronizer performs around the transfer. Managed Dynamo always drains active trajectories before refit. vLLM requires `async_engine: true`; the Megatron backend is always async-engine.
```

terrykong: [`docs/guides/async-grpo.md:208`](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/docs/guides/async-grpo.md#L208)

**1 action item.**

The sentence "Backends without native pause/resume support emit a once-per-backend warning and retain their existing in-flight update behavior" is inaccurate for SGLang, on both halves:

- SGLang *does* have native pause/resume — [`pause_generation(mode)` / `continue_generation()`](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/nemo_rl/models/generation/sglang/sglang_generation.py#L329-L345). It just does not implement the new refit hook, so it takes the warning path anyway.
- Its "existing in-flight update behavior" was to drain via [`wait_for_pending_generations()`](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1051), because the old `if/elif` chain sent every unlisted backend to `is_async_engine = False`. Under [the new predicate](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/nemo_rl/algorithms/async_utils/trajectory_collector.py#L969) SGLang no longer drains once `use_async_rollouts` is on.

To be clear this is a docs fix, not a bug report: [`should_use_async_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/nemo_rl/models/generation/interfaces.py#L250-L251) reads `use_async_rollouts` with a `False` default and the only config that mentions it sets it [explicitly to `false`](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/examples/configs/grpo_math_1B_sglang.yaml#L9), so nothing in the repo reaches that path today and no default changes. The transfer itself is also safe either way, since [`SGLangWeightSynchronizer._refit`](https://github.com/NVIDIA-NeMo/RL/blob/cb8948d375e943962467a5b9e82584b00987d5c6/nemo_rl/weight_sync/sglang_weight_synchronizer.py#L150-L180) brackets it with `finally: continue_generation()`. It is only worth correcting because this sentence is exactly where a future reader would go to find out what SGLang does.

**Action:** replace the sentence. One way to do it — of several — keeping the rest of the tip as-is:

```suggestion
4. **In-Flight Weight Updates**: Enable `in_flight_weight_updates: true` to refit without waiting for the longest in-flight generation to finish. Except for managed Dynamo, the collector requests a generation pause and resume from every async backend around the weight transfer. Async vLLM implements this contract while preserving request state. A backend that does not implement the hook emits a warning once per backend type per process and refits without a collector-side pause or drain; SGLang is in this group today and instead relies on the pause its own weight synchronizer performs around the transfer. Managed Dynamo always drains active trajectories before refit. vLLM requires `async_engine: true`; the Megatron backend is always async-engine.
```

terrykong (commented):
Third-round review by a team of six agents (RL expert, bug finder, test reviewer, design reviewer, comment reviewer, devil's advocate), focused on what changed since the last round.

One docs correction below. No blockers, and no code changes requested.

It is worth listing what we chased down and cleared, so it does not get re-raised:

- **vLLM API usage is correct.** `AsyncLLM.pause_generation(mode="keep", clear_cache=...)` and `resume_generation()` match [v0.25.1](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/engine/async_llm.py#L750-L797) exactly, including `"keep"` being a valid `PauseMode`.
- **Skipping `invalidate_kv_cache()` on the native path is right, and stronger than what it replaces.** `clear_cache=True` reaches [`reset_prefix_cache(reset_running_requests=True)`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/core/sched/scheduler.py#L2196-L2232), which preempts running requests and drops their blocks, so preserved requests really do re-prefill. The old `reset_prefix_cache()` was a no-op while running requests still held KV.
- **The broadcast is correct.** `run_rank_0_only_axes=["tensor_parallel", "pipeline_parallel"]` matches every sibling call and reaches every DP rank, which matters because `resume_scheduler` does an all-reduce across DP ranks.
- **The predicate swap changes no default.** `should_use_async_rollouts` matches the old chain for vLLM, Megatron, TRT-LLM and Dynamo. SGLang is the only backend whose result can differ, and only when `use_async_rollouts: true`, which defaults to `False` and is set to `false` in the one config that mentions it. Hence the docs note rather than a code comment.
- **No deadlock, no missing state reset, no threading issue.** The two new flags are only touched on the collector's actor thread; rollout threads use the `threading.Event`.
- **The trtllm assert was not lost** — it lives inside `should_use_async_rollouts`.
- Linters pass on all changed files.

Thanks for the pause/resume approach here — routing through vLLM's native keep-mode is a real correctness improvement over writing weights into an actively decoding engine, and the 10-step 32xH100 run with 10 matched pause/resume cycles is good evidence it works end to end.

_Generated by Claude Code_

youngeunkwon0405:
/ok to test [83e7e5b](https://github.com/NVIDIA-NeMo/RL/pull/3839/commits/83e7e5bc72f78557dd5c2074a199c583e00b8462)

terrykong (approved):
nit on the docs but overall lgtm. thanks for finally

PR #2580, opened 2026-05-27
feat(distillation): add TransferQueue support for On-Policy Distillation

pthombre:
# What does this PR do ?

Adds TransferQueue-backed on-policy distillation so teacher top-k logits can stay in the data plane instead of being returned through the driver.

This PR extends the data-plane support that already exists for GRPO to the distillation workflow. When `data_plane.enabled=true`, `examples/run_distillation.py` now builds the student policy as a `TQPolicy` and dispatches to a new TQ-aware distillation trainer. The legacy distillation path remains the default when the data plane is disabled.

Key pieces added:

- A new `distillation_train_sync` implementation that orchestrates rollout, teacher top-k inference, and student training through TransferQueue metadata.
- A `DistillationRolloutActor` that runs generation/rollout processing and seeds the TQ partition with row-aligned distillation training tensors.
- Worker-side `get_topk_logits_presharded` support so teacher workers write `teacher_topk_logits` and `teacher_topk_indices` back to TQ directly.
- Distillation-specific data-plane schema fields for teacher top-k seed inputs and train-time top-k columns.
- `TQPolicy` extensions for custom partition fields, consumer task names, and custom train fetch fields so the same policy wrapper can serve GRPO and distillation.
- Transport-volume metrics for teacher top-k payloads, valid-token payloads, padding overhead, TQ write volume, and driver bytes avoided.
- A TQ exemplar config and a runnable recipe/test-suite script for Qwen3 32B-to-1.7B distillation.

# Issues

None.

# Usage

Enable the TQ distillation path by using the new exemplar or recipe:

```bash
uv run examples/run_distillation.py \
  --config examples/configs/distillation_math_tq.yaml \
  data_plane.enabled=true
```

The recipe added in this PR can be launched through the test-suite wrapper:

```bash
tests/test_suites/llm/distillation-qwen3-32b-to-1.7b-base-1n8g-fsdp2tp1-tq.v1.sh
```

# Before your PR is "Ready for review"

**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
[...]

PR #3843, opened 2026-08-26
feat(sc): add the teacher top-k forward for distillation on SingleController

tianyi-zhang-02:
## What does this PR do?

Adds the teacher top-k forward primitive needed by distillation on SingleController:

- `TQWorkerMixin.get_topk_logits_presharded` runs the rank-local forward and writes both top-k tensors back to TransferQueue.
- `TQPolicy.get_topk_logits_from_meta` dispatches that work from the driver.
- The tests pin both output columns, the `[B, S, k]` shape, leader-only write-back, and batch-size validation.

This is the first PR in the #3843 → #3846 → #3849 stack. Nothing calls the new entry point until #3846.

## Ownership

#2580 already contains the same `get_topk_logits_presharded` entry point and the same `teacher_topk_logits` / `teacher_topk_indices` columns as part of a larger TQ distillation implementation. That PR is still open, so ownership is unresolved. I am holding this stack here and will not add more downstream work until the maintainers choose which direction to keep.

## Validation

- Current live head: `744463629719d1e33d49d85ff5f959ac1c4865de`.
- Focused policy tests cover the worker and driver entry points.
- The full stack at #3849 head `907606048aa8c950027dbc199bff1495035e862c` passed 765 unit/config cases (6 GPU-only skips) and a real 2-GPU teacher/student checkpoint-restore run.

This base branch is intentionally not being presented as current-main-ready while #2580 ownership remains unresolved; the two downstream branches are kept tested, but no further stack is being built on it.

### DIFF
--- a/nemo_rl/data_plane/worker_mixin.py
+++ b/nemo_rl/data_plane/worker_mixin.py
@@ -561,6 +561,50 @@ class TQWorkerMixin:
         )
         del result
 
+    @wrap_with_nvtx_name("policy_worker/get_topk_logits_presharded")
+    def get_topk_logits_presharded(
+        self,
+        meta: "KVBatchMeta",
+        k: int,
+        micro_batch_size: Optional[int] = None,
+    ) -> None:
+        """Per-rank teacher top-k forward entrypoint.
+
+        Same contract as ``get_logprobs_presharded``, with one difference worth
+        naming: ``get_topk_logits`` returns two tensors rather than one, so this
+        writes back twice. Both are ``[B, S, k]`` -- the write-back only checks
+        the batch dimension, so the extra axis passes through unchanged.
+
+        Only a distillation teacher mixes this in. The teacher never trains, so
+        there is no split begin/microbatch/finish counterpart: one call is one
+        forward.
+
+        Args:
+            meta: Per-rank ``KVBatchMeta`` for this slice.
+            k: Number of top logits to keep per position.
+            micro_batch_size: Inference micro batch size; None uses the config default.
+        """
+        data = self._fetch(meta)
+        data = self._attach_or_repack_pack_metadata(data, meta)
+        result: BatchedDataDict[Any] = self.get_topk_logits(  # type: ignore[attr-defined]
+            data=data,
+            k=k,
+            micro_batch_size=micro_batch_size,
+        )
+        self._write_back_result_field(
+            meta,
+            result,
+            result_key="topk_logits",
+            tq_field="teacher_topk_logits",
+        )
+        self._write_back_result_field(
+            meta,
+            result,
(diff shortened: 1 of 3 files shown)


tianyi-zhang-02:
#3768 landed while this was open, so worth saying up front how the two relate — they both put a teacher on SingleController and they are not the same thing.

**#3768 (MOPD)**: several teachers, each on its own reserved nodes (`non_colocated_teachers`), routed per NeMo-Gym agent. Their **logprobs** become a token-level *advantage* via `adv_estimator.name='opd'`. Entrypoint is `get_teacher_logprobs_presharded`, config lives under `on_policy_distillation`.

**This stack**: one frozen teacher, sharing the training GPUs as a second worker group and offloaded between forwards. Its **top-k logits** feed `DistillationLossFn` as a *loss*. Entrypoint is `get_topk_logits_presharded`, config is the `distillation` + `teacher` blocks — the SC port of `examples/run_distillation.py`.

So: different signal, different resource model, different objective. Both entrypoints now sit next to each other in `TQWorkerMixin` and the two config paths reject each other explicitly (a config setting both is an error, same as `ppo` + `on_policy_distillation`).

Rebased onto #3768. One real fix fell out of it: its new validation reads `algo_cfg.adv_estimator`, which `DistillationConfig` does not have, so a distillation config died with an `AttributeError` inside the validator. That is in #3849.

The one thing I would like a steer on: #3768 established `teacher_worker_group` as an SC teacher abstraction, and this stack predates it — it reuses `TQPolicy` instead, since a colocated teacher is a policy that never trains and needed no new driver. If you would rather have one teacher abstraction on this path I am happy to converge on it, but the resource models are different enough (`teacher_worker_group` reserves nodes; this one shares the trainer's) that I did not want to force it without asking. :)

tianyi-zhang-02:
**Duplicate flag, found while sweeping the open PRs.** @pthombre's #2580 ("feat(distillation): add TransferQueue support for On-Policy Distillation") already adds `get_topk_logits_presharded` with the same signature and the same two column names, `teacher_topk_logits` / `teacher_topk_indices`. The only difference is the return value — theirs returns a transport ack, this returns `None`.

That PR is much larger (it also adds the write-back dispatcher, payload sizing and transport metrics), has been open since 2026-05-27, last touched 2026-06-26, and is CONFLICTING against `main`.

I've asked there whether it's being picked back up. If it is, I'll close this — no reason to have two. Flagging it here so a reviewer doesn't find the collision cold.

tianyi-zhang-02:
Closing to avoid competing with the existing TransferQueue distillation work in #2580 and the newer distillation roadmap. The teacher top-k/TQ entry point is ownership-sensitive, and a parallel SingleController stack would make that direction harder to review.

PR #2924, opened 2026-06-24
fix(topology): exclude unknown-NVLink-domain nodes from segment selection

terrykong:
# What does this PR do?

Fixes a placement bug in `select_segment_nodes` (topology-aware NVLink-domain placement, from #2612). Nodes with no NVLink-domain info — e.g. a GPU-less or unprobed Ray head node — collapse into a single `NVLINK_DOMAIN_UNKNOWN` pseudo-domain with `topo_rank = -1`, so they sort first and get selected once there are >= `segment_size` of them. The resulting placement-group constraint `{"unknown": 0.001}` names a Ray resource that `ray.sub` never registers, so the bundle can never schedule (retry loop / hard failure on heterogeneous or partial-probe clusters).

The fix skips unknown-domain nodes when building the candidate set, so we only ever pin to real, registered NVLink domains; skipped nodes fall through to `remaining_node_ids`. No-op on a homogeneous all-GPU cluster.

## Testing

**Unit tests** — added regression tests in `tests/unit/distributed/test_topology_placement.py::TestSelectSegmentNodes`:
- `test_unknown_domain_nodes_excluded_from_selection` and `test_unknown_domain_only_raises_instead_of_unschedulable_pg` — both **fail on the pre-fix code, pass after**.
- Full file: **32 passed** on a GB200 node (HEAD `e2b6b2a2`).

**Functional `--segment` run on GB200** — confirms the fix does not regress topology-aware placement on real hardware. `grpo-llama3.2-1b-instruct-1n8g-megatron.yaml`, **2 nodes × 4 GPU, `cluster.segment_size=2`**, 5 steps:
- Topology-aware placement engaged: `✓ Topology-aware allocation: 2 training nodes in 1 NVLink domains (segment_size=2)`
- Reached **Step 5/5**, exited cleanly (`---EXIT:0---`), with no unschedulable-PG / `ResourceInsufficientError`.

(The unknown-domain case itself cannot occur on a clean homogeneous GB200 rack — it is covered by the unit tests; this functional run verifies no regression to the normal placement path.)

## Stack
Bottom of a stack — **#2613 (feat: Numa aware binding)** stacks on top of this.

## Before your PR is "Ready for review"
- [x] Did you write any new necessary tests?
- [x] Did you run the unit tests locally?

PR #3844, opened 2026-08-26
fix(vllm): return HTTP 400 for chat completion context overflow

jepio:
# What does this PR do ?

This PR fixes a contract mismatch between NeMo-RL and NeMo-Gym: when `_clamp_max_tokens` detects an overflow the ValueError exception results in an HTTP 500 error that Gym retries. Turn it into an VLLMValidationError that Gym sees as HTTP 400 and can immediately treat as a context overflow.

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/models/generation/vllm/vllm_worker_async.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker_async.py
@@ -768,11 +768,16 @@ class VllmAsyncGenerationWorkerImpl(
                 remaining = self.model_config.max_model_len - len(prompt_token_ids)
                 if remaining <= 0:
                     # preserve the literal "context length" in this message to match Gym's overflow handling
-                    raise VLLMValidationError(
+                    message = (
                         f"Prompt length ({len(prompt_token_ids)}) fills or exceeds "
                         f"this model's maximum context length ({self.model_config.max_model_len}). "
-                        f"No room for output tokens.",
-                        parameter="max_model_len",
+                        f"No room for output tokens."
+                    )
+                    LOGGER.warning("Prompt exceeds max_model_len: %s", message)
+                    raise VLLMValidationError(
+                        message,
+                        parameter="input_tokens",
+                        value=len(prompt_token_ids),
                     )
                 max_tokens = min(request_max_tokens, remaining)
                 self._set_max_tokens(request, max_tokens)
@@ -1155,6 +1160,7 @@ class VllmAsyncGenerationWorkerImpl(
                         "error": {
                             "message": str(e),
                             "type": "invalid_request_error",
+                            "param": e.parameter,
                             "code": 400,
                         }
                     },
--- a/nemo_rl/models/generation/vllm/vllm_worker_async.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker_async.py
@@ -767,10 +767,12 @@ class VllmAsyncGenerationWorkerImpl(
                 """Clamp the request's max output tokens so that input + output <= max_model_len."""
                 remaining = self.model_config.max_model_len - len(prompt_token_ids)
                 if remaining <= 0:
-                    raise ValueError(
+                    # preserve the literal "context length" in this message to match Gym's overflow handling
+                    raise VLLMValidationError(
                         f"Prompt length ({len(prompt_token_ids)}) fills or exceeds "
-                        f"max_model_len ({self.model_config.max_model_len}). "
-                        f"No room for output tokens."
+                        f"this model's maximum context length ({self.model_config.max_model_len}). "
+                        f"No room for output tokens.",
+                        parameter="max_model_len",
                     )
                 max_tokens = min(request_max_tokens, remaining)
                 self._set_max_tokens(request, max_tokens)
--- a/tests/unit/models/generation/test_vllm_generation.py
+++ b/tests/unit/models/generation/test_vllm_generation.py
@@ -546,10 +546,23 @@ def _install_fake_vllm_openai_modules(monkeypatch):
             self.kwargs = kwargs
             self.instances.append(self)
 
-    class VLLMValidationError(Exception):
-        def __init__(self, message="", parameter=None):
+    class VLLMValidationError(ValueError):
+        def __init__(self, message, *, parameter=None, value=None):
             super().__init__(message)
             self.parameter = parameter
+            self.value = value
+
+        def __str__(self):
+            base = super().__str__()
+            extras = [
+                f"{name}={value}"
+                for name, value in (
+                    ("parameter", self.parameter),
+                    ("value", self.value),
+                )
+                if value is not None
+            ]
+            return f"{base} ({', '.join(extras)})" if extras else base
 
     class ToolParserManager:
         import_tool_parser = MagicMock()
--- a/tests/unit/models/generation/test_vllm_generation.py
+++ b/tests/unit/models/generation/test_vllm_generation.py
@@ -547,7 +547,9 @@ def _install_fake_vllm_openai_modules(monkeypatch):
             self.instances.append(self)
 
     class VLLMValidationError(Exception):
-        pass
+        def __init__(self, message="", parameter=None):
+            super().__init__(message)
+            self.parameter = parameter
 
     class ToolParserManager:
         import_tool_parser = MagicMock()
@@ -658,6 +660,75 @@ def test_vllm_async_http_server_loads_reasoning_parser_plugin(monkeypatch):
     assert openai_serving_chat.instances[0].kwargs["reasoning_parser"] == "nano_v3"
     # make sure that the config attribute does not leak into `http_server_serving_chat_kwargs`
     assert "reasoning_parser_plugin" not in openai_serving_chat.instances[0].kwargs
+
+
+def _nemo_gym_recognizes_context_overflow(
+    *, status: int, response_content: str
+) -> bool:
+    """Mirror responses_api_models/vllm_model/app.py overflow detection."""
+    return status == 400 and (
+        "context length" in response_content or "max_tokens" in response_content
+    )
+
+
+@pytest.mark.asyncio
+async def test_context_overflow_returns_http_400_for_nemo_gym(monkeypatch):
+    """Overflow from _clamp_max_tokens must be HTTP 400 that NeMo Gym recognizes."""
+    _, _, openai_serving_chat = _install_fake_vllm_openai_modules(monkeypatch)
+
+    worker = VllmAsyncGenerationWorkerImpl.__new__(VllmAsyncGenerationWorkerImpl)
+    worker.cfg = {
+        "temperature": 1.0,
+        "top_p": 1.0,
+        "val_temperature": 0.0,
+        "val_top_p": 1.0,
+        "vllm_cfg": {},
+    }
+    worker.llm = MagicMock(model_config="model-config", renderer="renderer")
+    worker._http_engine_client = worker.llm
+    worker._capture_calls = {}
+    worker.token_capture = None
+    worker.llm_async_engine_args = MagicMock()
+    worker.llm_async_engine_args.create_model_config.return_value = MagicMock(
+        served_model_name="served-model", model="model-path"
+    )
+
+    app = _FakeFastAPIApp()
+    worker._setup_vllm_openai_api_server(app)
+
+    max_model_len = 128
+    renderer = openai_serving_chat.instances[0].kwargs["online_renderer"]
+    renderer.model_config = types.SimpleNamespace(max_model_len=max_model_len)
+    serving_chat = openai_serving_chat.instances[0]
+    chat_handler = next(
+        handler for path, handler in app.routes if path == "/v1/chat/completions"
+    )
+    overflow_prompt = [0] * max_model_len
+
+    async def create_chat_completion(request, _raw_request):
+        renderer._clamp_max_tokens(request, request.max_tokens, overflow_prompt)
+
+    serving_chat.create_chat_completion = create_chat_completion
+    response = await chat_handler(
+        types.SimpleNamespace(
+            top_k=-1,
+            top_p=1.0,
+            temperature=1.0,
+            max_tokens=1,
+            max_completion_tokens=None,
+        ),
+        MagicMock(),
+    )
+
+    response_content = response.body.decode()
+    assert response.status_code == 400
+    assert _nemo_gym_recognizes_context_overflow(
+        status=response.status_code,
+        response_content=response_content,
+    )
+    error = json.loads(response_content)["error"]
+    assert error["type"] == "invalid_request_error"
+    assert error["code"] == 400
 
 
 def test_nano_v3_reasoning_parser_swaps_reasoning_when_thinking_disabled(
--- a/tests/unit/models/generation/test_vllm_generation.py
+++ b/tests/unit/models/generation/test_vllm_generation.py
@@ -735,12 +735,15 @@ async def test_context_overflow_returns_http_400_for_nemo_gym(monkeypatch):
 
     response_content = response.body.decode()
     assert response.status_code == 400
+    assert "maximum context length" in response_content
+    assert "(parameter=input_tokens, value=128)" in response_content
     assert _nemo_gym_recognizes_context_overflow(
         status=response.status_code,
         response_content=response_content,
     )
     error = json.loads(response_content)["error"]
     assert error["type"] == "invalid_request_error"
+    assert error["param"] == "input_tokens"
     assert error["code"] == 400
 
 


nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -440,10 +440,12 @@ def _clamp_max_tokens(
...
-                    raise ValueError(
+                    # preserve the literal "context length" in this message to match Gym's overflow handling
+                    raise VLLMValidationError(
                         f"Prompt length ({len(prompt_token_ids)}) fills or exceeds "
-                        f"max_model_len ({self.model_config.max_model_len}). "
-                        f"No room for output tokens."
+                        f"this model's maximum context length ({self.model_config.max_model_len}). "
+                        f"No room for output tokens.",
+                        parameter="max_model_len",
                     )
```
jepio, line 457: [`vllm_worker_async.py:442-449`](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/nemo_rl/models/generation/vllm/vllm_worker_async.py#L442)

**2 low-severity nits, PR-introduced.**

### AI-1 — overflow is no longer logged anywhere (consider, not a break)
Both `_clamp_max_tokens` raise sites ([L515](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/nemo_rl/models/generation/vllm/vllm_worker_async.py#L515), [L570](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/nemo_rl/models/generation/vllm/vllm_worker_async.py#L570)) sit outside the `preprocess_chat` try/except ([L489-507](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/nemo_rl/models/generation/vllm/vllm_worker_async.py#L489)) that would otherwise log a one-line warning for this case, so this path was never logged by that handler either before or after this PR. What the PR does change: before, the escaping plain `ValueError` was uncaught by our code and surfaced as a 500 (loud, logged by the ASGI layer); after, it's caught cleanly at L787-796 and returned as a 400 with no log line anywhere. That's plausibly intentional — Gym now handles overflow gracefully instead of retrying — but worth a deliberate call: log before raising if overflow events should stay visible in server logs.

### AI-2 — `parameter="max_model_len"` names the wrong thing
vLLM's `VLLMValidationError.__str__` ([pinned v0.25.1](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/exceptions.py)) appends `(parameter=...)` to the message, and the HTTP-400 handler at [L787-796](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/nemo_rl/models/generation/vllm/vllm_worker_async.py#L787) emits `str(e)` — so this is user-visible text, not dropped metadata. `parameter` should name the offending *request* field (per the sibling call at [L611-615](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/nemo_rl/models/generation/vllm/vllm_worker_async.py#L611), which correctly uses `parameter="top_logprobs"`), not a server config the client can't act on. Cosmetic only — all 3 Gym matchers pass either way. Note: if you take this suggestion, the `assert "maximum context length" in response_content` proposed on the test below becomes load-bearing rather than optional — today the wording literal is pinned only incidentally (nothing else in the body contains "max_tokens"); adding `max_tokens` to the body via this change makes Gym's `or "max_tokens"` branch fire unconditionally, so a future wording regression would silently stop being caught without that explicit assertion.

```suggestion
                if remaining <= 0:
                    # preserve the literal "context length" in this message to match Gym's overflow handling
                    message = (
                        f"Prompt length ({len(prompt_token_ids)}) fills or exceeds "
                        f"this model's maximum context length ({self.model_config.max_model_len}). "
                        f"No room for output tokens."
                    )
                    LOGGER.warning("Prompt exceeds max_model_len: %s", message)
                    raise VLLMValidationError(
                        message,
                        parameter="max_tokens",
                    )
```

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -401,6 +403,72 @@ def test_vllm_async_http_server_loads_reasoning_parser_plugin(monkeypatch):
...
+        ),
+        MagicMock(),
+    )
+
+    response_content = response.body.decode()
+    assert response.status_code == 400
+    assert _nemo_gym_recognizes_context_overflow(
+        status=response.status_code,
+        response_content=response_content,
+    )
```
jepio, line 546: [`test_vllm_generation.py:462-466`](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/tests/unit/models/generation/test_vllm_generation.py#L462)

**No action needed — optional robustness note.** [`_nemo_gym_recognizes_context_overflow`](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/tests/unit/models/generation/test_vllm_generation.py#L406)'s `or "max_tokens"` disjunct means a future response-body change could satisfy the helper without the wording regression being caught — mutation testing confirms the current OR still catches it today, but a direct assertion pins the literal that Gym's `responses_api_models/vllm_model/app.py:682-683` actually checks, independent of the helper's leniency.

```suggestion
    response_content = response.body.decode()
    assert response.status_code == 400
    assert "maximum context length" in response_content
    assert _nemo_gym_recognizes_context_overflow(
        status=response.status_code,
        response_content=response_content,
    )
```

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -289,7 +289,9 @@ def __init__(self, **kwargs):
             self.instances.append(self)
 
     class VLLMValidationError(Exception):
-        pass
+        def __init__(self, message="", parameter=None):
+            super().__init__(message)
+            self.parameter = parameter
```
jepio: [`test_vllm_generation.py:291-294`](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/tests/unit/models/generation/test_vllm_generation.py#L291)

**1 action item, low severity.** This fake `VLLMValidationError` is what production actually gets (`vllm.exceptions` is stubbed at test:343, and prod lazy-imports it at [vllm_worker_async.py:377](https://github.com/NVIDIA-NeMo/RL/blob/7cb254027df88a48e532b35d2776e8e149967db9/nemo_rl/models/generation/vllm/vllm_worker_async.py#L377)), but it omits vLLM's `__str__` override, subclasses `Exception` instead of `ValueError`, and takes `parameter` positionally instead of keyword-only. Consequence: the double's message shape never occurs in production — the real body ends `"...(parameter=max_tokens)"` (see the sibling comment on `_clamp_max_tokens`) but the double's `str(e)` won't include that suffix.

**Action:** mirror the real class.

```suggestion
    class VLLMValidationError(ValueError):
        def __init__(self, message="", *, parameter=None, value=None):
            super().__init__(message)
            self.parameter = parameter
            self.value = value

        def __str__(self):
            base = super().__str__()
            extras = [
                f"{name}={val}"
                for name, val in (("parameter", self.parameter), ("value", self.value))
                if val is not None
            ]
            return f"{base} ({', '.join(extras)})" if extras else base
```

jepio (commented):
Reviewed by a coordinated team of agents (rl-expert, bug-finder, test-agent, design-reviewer, comment-reviewer) with an adversarial devil's-advocate verification pass. The fix itself is correct and necessary — confirmed that pre-PR, the plain `ValueError` from `_clamp_max_tokens` escaped the existing `except VLLMValidationError` HTTP-400 handler and produced an uncaught 500 that Gym retried as an infra failure instead of a data/context-length failure. All findings below are low-severity polish, none blocking. Note: this repo's `uv.lock` only supports linux platforms, so lint/pre-commit could not run on the reviewing host; the new test was executed directly (via an ad-hoc CPU-only env) and passes, and was mutation-tested to confirm it actually pins the regression.

_Generated by Claude Code_

tests/unit/models/generation/test_vllm_generation.py
jepio: fixed in 34c81388a19bf45bf9630597056bcc4fc3659216

tests/unit/models/generation/test_vllm_generation.py
jepio: fixed in 99d92bd7a5b151627f4a6d4504c73fa6d55c968a

nemo_rl/models/generation/vllm/vllm_worker_async.py
jepio: fixed in 34c81388a19bf45bf9630597056bcc4fc3659216

jepio: Verified: `_clamp_max_tokens` now logs `LOGGER.warning("Prompt exceeds max_model_len: %s", message)` before raising, and `parameter="max_tokens"` matches the sibling `top_logprobs` pattern. Matches the suggestion exactly. Resolving.

tests/unit/models/generation/test_vllm_generation.py
jepio: Verified: both `assert "maximum context length" in response_content` and `assert "(parameter=max_tokens)" in response_content` are now present, pinning the wording independent of the `_nemo_gym_recognizes_context_overflow` helper's `or "max_tokens"` leniency. Resolving.

tests/unit/models/generation/test_vllm_generation.py
jepio: Verified: the test double now subclasses `ValueError`, takes `parameter`/`value` keyword-only, and implements `__str__` to append `(parameter=..., value=...)` — mirrors production `vllm.exceptions.VLLMValidationError`. Resolving.

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -466,6 +481,74 @@ def test_vllm_async_http_server_loads_reasoning_parser_plugin(monkeypatch):
...
+    *, status: int, response_content: str
+) -> bool:
+    """Mirror responses_api_models/vllm_model/app.py overflow detection."""
+    return status == 400 and (
+        "context length" in response_content or "max_tokens" in response_content
+    )
+
+
+@pytest.mark.asyncio
+async def test_vllm_http_server_context_overflow_matches_nemo_gym(monkeypatch):
```
yuki-97: This test never runs in CI: pytest's `--deselect` matches by node-id **prefix** ([`_pytest/main.py:489`](https://github.com/pytest-dev/pytest/blob/9.1.1/src/_pytest/main.py#L489)), and [`excluded_unit_tests.sh:93`](https://github.com/NVIDIA-NeMo/RL/blob/34c81388a19bf45bf9630597056bcc4fc3659216/tests/unit/excluded_unit_tests.sh#L93) deselects `...::test_vllm_http_server`, which is a prefix of this name.

In the CI:Lfast run on 34c81388a19bf45bf9630597056bcc4fc3659216 it appears in no `L0_Unit_Tests_Vllm_*` summary, `Coverage (unit-test)` reports [`vllm_worker_async.py` 445-459](https://github.com/NVIDIA-NeMo/RL/blob/34c81388a19bf45bf9630597056bcc4fc3659216/nemo_rl/models/generation/vllm/vllm_worker_async.py#L445-L459) as missing, and `codecov/patch` is red at 0.00% of diff hit. A rename is enough:

```suggestion
async def test_vllm_context_overflow_matches_nemo_gym(monkeypatch):
```

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -466,6 +481,74 @@ def test_vllm_async_http_server_loads_reasoning_parser_plugin(monkeypatch):
...
+    serving_chat = openai_serving_chat.instances[0]
+    chat_handler = next(
+        handler for path, handler in app.routes if path == "/v1/chat/completions"
+    )
+    overflow_prompt = [0] * max_model_len
+
+    async def create_chat_completion(request, _raw_request):
+        renderer._clamp_max_tokens(request, request.max_tokens, overflow_prompt)
+
+    serving_chat.create_chat_completion = create_chat_completion
```
yuki-97: nit: worth pointing this stub at the second `_clamp_max_tokens` call site — [L578](https://github.com/NVIDIA-NeMo/RL/blob/34c81388a19bf45bf9630597056bcc4fc3659216/nemo_rl/models/generation/vllm/vllm_worker_async.py#L578), after `replace_prefix_tokens` — since that is the only site where the new raise is reachable.

At [L523](https://github.com/NVIDIA-NeMo/RL/blob/34c81388a19bf45bf9630597056bcc4fc3659216/nemo_rl/models/generation/vllm/vllm_worker_async.py#L523) `preprocess_chat` has already set the request's max tokens to 1, so `max_input_tokens` is `max_model_len - 1` and vLLM's own [`_token_len_check`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/params.py#L434) raises on the same condition first. As written the stub swaps out `create_chat_completion` wholesale, so deleting both calls would leave this test green.

yuki-97 (commented):
@ananthsub could you help to review?

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -466,6 +481,74 @@ def test_vllm_async_http_server_loads_reasoning_parser_plugin(monkeypatch):
...
+def _nemo_gym_recognizes_context_overflow(
+    *, status: int, response_content: str
+) -> bool:
+    """Mirror responses_api_models/vllm_model/app.py overflow detection."""
+    return status == 400 and (
+        "context length" in response_content or "max_tokens" in response_content
+    )
+
+
+@pytest.mark.asyncio
```
ananthsub: **1 action item — please fix in this PR.**

Codecov reports 0% diff coverage and marks the new production `_clamp_max_tokens` branch uncovered. Green L0 shards show that selected tests pass, but this regression is not participating in aggregated patch coverage.

**Action:** ensure this test runs in the coverage job and directly cover `max_model_len - 1`, `max_model_len`, and `max_model_len + 1`; rerun until [Codecov](https://app.codecov.io/gh/NVIDIA-NeMo/RL/pull/3844) passes.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -444,11 +444,17 @@ def _clamp_max_tokens(
...
+                    message = (
                         f"Prompt length ({len(prompt_token_ids)}) fills or exceeds "
-                        f"max_model_len ({self.model_config.max_model_len}). "
+                        f"this model's maximum context length ({self.model_config.max_model_len}). "
                         f"No room for output tokens."
                     )
+                    LOGGER.warning("Prompt exceeds max_model_len: %s", message)
+                    raise VLLMValidationError(
+                        message,
+                        parameter="max_tokens",
```
ananthsub: **1 action item.**

This reports the deprecated alias even when the request supplied only `max_completion_tokens`: the preprocessing path prefers that modern field, but every overflow body says `(parameter=max_tokens)`.

**Action:** select `max_completion_tokens` when populated, otherwise `max_tokens`, and parameterize the boundary test over both request forms.

tests/unit/models/generation/test_vllm_generation.py
```diff
@@ -466,6 +481,74 @@ def test_vllm_async_http_server_loads_reasoning_parser_plugin(monkeypatch):
     assert "reasoning_parser_plugin" not in openai_serving_chat.instances[0].kwargs
 
 
+def _nemo_gym_recognizes_context_overflow(
```
ananthsub: **1 action item, low severity.**

This helper no longer mirrors Gym #2924, which moves classification into a shared regex classifier, and its result is already guaranteed by the direct status and message assertions below. Keeping a copied private predicate creates cross-repository drift without testing the real integration.

**Action:** remove this helper and retain assertions on RL's own stable HTTP status, envelope, and context-length wording; cover actual cross-repo compatibility in a dedicated integration test.

nemo_rl/models/generation/vllm/vllm_worker_async.py
ananthsub: **1 action item — please fix in this PR.**

This always attributes overflow to `max_tokens`, even when the request supplied only `max_completion_tokens`, and the resulting JSON exposes the parameter only inside message text. Please select the populated request field, pass that name to `VLLMValidationError`, include it as structured `error.param` as supported by [vLLM's error envelope](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/openai/engine/protocol.py#L55-L63), and parameterize the test over both request forms.

ananthsub, line 456: This always attributes overflow to `max_tokens`, even when the request supplied only `max_completion_tokens`, and the resulting JSON exposes the parameter only inside message text. Please select the populated request field, pass that name to `VLLMValidationError`, include it as structured `error.param` as supported by [vLLM's error envelope](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/openai/engine/protocol.py#L55-L63)

jepio: Thanks for the review @ananthsub, this conditional branch in this function is actually unaffected by the request supplied `max_tokens` or `max_completion_tokens`, it fires when the input prompt overflows the server side `max_model_len` setting. The more appropriate parameter is `input_tokens`, i've switched to that parameter and included the value as `error.param`. See https://github.com/NVIDIA-NeMo/RL/pull/3844/changes/0c3eda7ef84c4481495af968836b4ce5ae2e0a06

yuki-97: fixed in 0c3eda7ef84c4481495af968836b4ce5ae2e0a06

yuki-97 (approved):

PR #3846, opened 2026-08-26
feat(sc): run the distillation teacher in the SingleController train pump

tianyi-zhang-02:
## What does this PR do?

Wires distillation into the SingleController train pump. It builds a frozen teacher on the training cluster, parks it between forwards, writes teacher top-k logits and indices through TransferQueue, fetches only the distillation loss fields, and skips reward/advantage work that distillation does not use.

The current version also keeps the newer mainline SC behavior intact: dynamic GRPO/PPO train fields, message-level penalties, token-capture schema, route fields, PPO critic epochs, and current setup validation.

This is the middle PR in #3843 → #3846 → #3849. The runnable recipe and checkpoint/restore functional test remain test-only in #3849.

## Ownership

The teacher top-k primitive overlaps #2580. I disclosed the duplicate there and in #3843; #2580 remains open and conflicting, with no author response since August 27. I am keeping this stack current and tested, but not extending the overlap further.

## Validation

- Head: `80966321d71eb204c42d3b51a322640c2f1964a6`
- Merged main: `e831622435180c88d6adcc9fce46bfd88a85f3e5`
- Linux x86_64, Python 3.13.14, CUDA disabled: **285 passed** across teacher top-k, SC distillation, entrypoint, setup, PPO setup, checkpointing, and controller actor suites.
- Ruff check and format: clean for all 13 changed Python files.
- The 2-GPU H100 checkpoint/restore run is recorded in #3849. It was run on the pre-refresh stack head `90760604`; the current main refresh was validated with CPU/config tests, not rerun on H100.


tianyi-zhang-02:

```python
    #
    unsupported = (
        []
        else [
            name
            for name, enabled in (
            )
            if enabled
        ]
    )
```

tianyi-zhang-02:
Closing this dependent stack item together with #3843. The active distillation work now has separate ownership and direction in #2580 and the newer distillation series, so I do not want to maintain a competing SingleController integration.

PR #3849, opened 2026-08-26
test(sc): add the distillation recipe and functional test for SingleController

tianyi-zhang-02:
## What does this PR do?

Adds the native SingleController distillation recipe and its checkpoint/restore functional test. Relative to #3846, this PR is test-only:

- one 2-GPU Megatron + vLLM recipe;
- one functional script that trains steps 1–2, checks policy/metadata/TransferQueue checkpoints, restores, and finishes steps 3–4;
- config and recipe tests that fail before model loading if the shell overrides or SC constraints drift;
- L1 suite registration.

This is the top of #3843 → #3846 → #3849. GitHub compares it to `main`, but its unique delta from #3846 is exactly five test/recipe files (408 added lines).

## Ownership

The lower stack's teacher top-k primitive overlaps #2580. The duplicate is disclosed on both sides; #2580 remains open and conflicting, with no author response since August 27. This PR adds coverage only and does not extend the overlapping production surface.

## Validation

- Head: `50ca2117216b1c9bb689d8638354f192b13f6529`
- Parent #3846: `80966321d71eb204c42d3b51a322640c2f1964a6`
- Merged main: `e831622435180c88d6adcc9fce46bfd88a85f3e5`
- Current head, Linux x86_64, Python 3.13.14, CUDA disabled: both functional shell scripts pass `bash -n`; **21 passed** across functional-config, teacher-stage, field-schema, validation, and shipped-recipe tests.
- Ruff check and format: clean.

The full functional script previously passed on a Runpod 4×H100 80 GB host using 2 GPUs at pre-refresh head `90760604`:

- run 1 trained steps 1–2 and wrote metadata, data-plane, and policy checkpoints;
- run 2 restored step 2 and completed steps 3–4;
- losses: `0.387430`, `0.479430`, `0.405156`, `0.432410`;
- both metric files passed the loss-count, finite/non-negative, `< 20`, and teacher-timing gates.

That H100 result is historical evidence for the unchanged functional path; it has not been rerun after the current main refresh.


tianyi-zhang-02:
Update: the 2-GPU functional test has now run successfully on the current stack.

- live head: `907606048aa8c950027dbc199bff1495035e862c`
- host: Runpod 4×H100 80 GB; test used 2 GPUs
- run 1 trained steps 1–2 and wrote metadata, data-plane, and policy checkpoints
- run 2 restored step 2 and completed steps 3–4
- losses: `0.387430`, `0.479430`, `0.405156`, `0.432410`
- unit/config sweep: 765 passed, 6 GPU-only skips

The exact configuration, metric gates, and ownership hold are in the updated PR body.

tianyi-zhang-02:

```python
    #
    unsupported = (
        []
        else [
            name
            for name, enabled in (
            )
            if enabled
        ]
    )
```

tianyi-zhang-02:
Closing this dependent stack item together with #3843/#3846. The active distillation work now has separate ownership and direction in #2580 and the newer distillation series, so I do not want to maintain a competing functional-test layer.

PR #3852, opened 2026-08-26
feat: super v3.5 posttraining with flexible length penalties

arendu:
# What does this PR do ?

  Adds configurable length penalties/bonuses for rollouts (reward-mutation and GDPO reward-feature modes) to reduce verbosity in RLVR training — including group-relative length scaling, zMAD outlier penalties, profile-band multipliers with per-prompt or global `{a, b, f}` bands, and per-prompt profile-gated group-relative scaling — applied
  in both sync and async GRPO paths.

  # Issues

  None. This is the `super-v3.5-posttraining` port of the length-adjustment work targeting `main` in #3604, plus fixes made since (async rollout propagation, profile-band clamp at `f`, global profile-band defaults, unit tests).

  # Usage

  ```yaml
  # Reward-mutation mode: dense group-relative length pressure, gated per prompt
  grpo:
    length_bonus:
      default:
        enabled: true
        length_type: tokens
        group_total_length_penalty_coeff: 0.1
        group_length_penalty_profile_gate: true          # penalty only when group mean length
        group_length_penalty_profile_gate_channel: total #   exceeds the prompt's profiled target
        group_length_penalty_profile_gate_field: a
      profile_band:            # global {a,b,f} fallback for datasets without per-prompt bands
        enabled: true
        defaults:
          total: {a: 18000, b: 24000, f: 0.9}
      agent_overrides:
        instruction_following_simple_agent:
          enabled: false

  # GDPO feature mode: same signals as a separate reward feature, env reward untouched
  grpo:
    adv_estimator:
      name: gdpo
      reward_features:
        default:
          env_reward: 1.0
          length_adjusted_reward:
            group_total_length_penalty_coeff: 0.1
          think_count_delta: 1.0

  Full algorithm reference: nemo_rl/utils/length_adjustments.md.

  Before your PR is "Ready for review"

  Pre checks:
  - [x] Make sure you read and followed Contributor guidelines
  - [x] Did you run the unit tests and functional tests locally? Visit our Testing Guide for how to run tests
  - [x] Did you add or update any necessary documentation? Visit our Document Development Guide for how to write, build and test the docs.
# Additional Information
* ...


nemo_rl/algorithms/advantage_estimator.py
```diff
@@ -133,6 +216,14 @@ def compute_advantage(
         Returns:
             Advantages tensor of shape [batch_size, seq_len].
         """
+        if "gdpo_reward_features" in repeated_batch:
```
yfw: **1 action item.**

**TL;DR — this branch keys on a dict key that the gym rollout path now always writes, so every NeMo-Gym GDPO run silently switches to single-feature `env_reward`: the `reward/*` components and `reward_weights` both stop having any effect, and the "needs ≥2 components" guard is bypassed. No config opt-in required.**

PR-introduced (`gdpo_reward_features` has zero hits at the merge-base).

The dispatch is on **key presence**, not on configuration:
1. [`rollouts.py:2772`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/experience/rollouts.py#L2772) `_record_gdpo_env_reward_features(results)` runs unconditionally — the `if length_adjustment_config is not None:` gate is later, at [:2788](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/experience/rollouts.py#L2788), and covers only `apply_group_length_adjustments`.
2. [`rollouts.py:2961`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/experience/rollouts.py#L2961) writes `gdpo_reward_features` into `final_batch` unconditionally.
3. That batch becomes `repeated_batch`, so this `in` test is always True and the legacy multi-component path at [:227](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/advantage_estimator.py#L227) never runs.
4. [`_resolve_gdpo_reward_feature_selection`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/advantage_estimator.py#L97) is `estimator_config.reward_features or ["env_reward"]`, and **no checked-in config sets `reward_features`** — so it resolves to one feature.

**Your own test suite already asserts the property this breaks.** [`test_gdpo_advantage_estimator_reward_weights`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/tests/unit/algorithms/test_grpo.py#L4816) — docstring: *"non-uniform differ"* — ends with:

```python
# Non-uniform weights change the advantages.
assert not torch.allclose(default, run([1.0, 0.25]), atol=1e-3)
```

On the gym path that is now false. It stays green only because its `repeated_batch` ([:4821-4824](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/tests/unit/algorithms/test_grpo.py#L4821)) carries `reward/*` keys and no `gdpo_reward_features`, so it exercises the pre-PR branch. Measured, same batch, only the weights changed:

```
reward_weights [1.0, 1.0]   legacy [0.0, -1.0, 1.0]           feature [0.0, -1.0, 1.0]
reward_weights [5.0, 0.1]   legacy [0.5601, -1.1545, 0.5944]  feature [0.0, -1.0, 1.0]
```

`self.reward_weights` is read only at L195/L236/L241 — all after this early return — so [`gdpo_example_tool_call_multireward.yaml`'s](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/examples/nemo_gym/gdpo_example_tool_call_multireward.yaml) `reward_weights: [1.0, 0.5, 0.25]` is silently inert. (Divergence needs the components to differ across rollouts; a symmetric fixture shows no change, which is part of why this is easy to miss.)

Affected: `gdpo_example_tool_call_multireward.yaml`, `gdpo_multireward.yaml`, and `recipes/llm/gdpo-qwen2.5-1.5b-1n8g-gym-multireward.yaml` — the last is in `tests/test_suites/nightly.txt:310`. `examples/configs/gdpo_math_1B.yaml` is unaffected (native path). Note `gdpo_multireward.yaml`'s own header documents that GDPO *should* raise "requires multiple reward components" for its single-reward env — after this PR it trains silently instead, so a documented safety net is gone.

Also worth knowing why the suite didn't catch it: of the four `test_gdpo_advantage_estimator_*` tests, only the new one ([:4722](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/tests/unit/algorithms/test_grpo.py#L4722)) supplies the key; the three pre-existing ones don't, so they keep exercising the old branch.

**Action:** gate on configuration rather than key presence, and stop populating `gdpo_reward_features` unconditionally in `_postprocess_single_nemo_gym_group` (both halves, or the rollout and estimator sides drift again). Once gated, L4841's assertion should hold on the gym path too.

```suggestion
        if (
            self.estimator_config.reward_features is not None
            and "gdpo_reward_features" in repeated_batch
        ):
```

nemo_rl/utils/length_adjustments.py
```diff
@@ -0,0 +1,1236 @@
...
+                for c in content:
+                    t = c.get("text", "") if isinstance(c, dict) else getattr(c, "text", "")
+                    answer_text += t
+            elif isinstance(content, str):
+                answer_text += content
+
+    return reasoning_text, answer_text
+
+
+def _extract_gdpo_length_feature_params(feature_cfg: Any) -> dict[str, Any]:
```
yfw: **1 action item.**

**TL;DR — GDPO feature mode produces no length pressure in the documented config shape, and zero advantages in the flat shape. Neither form works; the fix is one level of dict descent here.**

PR-introduced.

This reads param keys **flat** out of the per-agent feature block, but [`length_adjustments.md:130-140`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/utils/length_adjustments.md) and `:145-156` — and the PR's own test at `test_grpo.py:4725-4732` — nest them under the feature name:

```yaml
reward_features:
  default:
    env_reward: 1.0
    length_adjusted_reward:
      group_total_length_penalty_coeff: 0.1
```

Executed against the real functions:

```
merged length_cfg (documented nested form): {}                     <- no length pressure
merged length_cfg (flat form)             : {'_gdpo_feature_only': True, 'default': {...}}
estimator features (nested): ['env_reward', 'length_adjusted_reward', 'think_count_delta']
estimator features (flat)  : ['group_total_length_penalty_coeff']    <- never a recorded feature
```

- **Nested (documented):** the loop sees feature *names*, none of which are in [`_GDPO_LENGTH_FEATURE_PARAM_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/utils/length_adjustments.py#L90), so `params == {}` and the merge returns early. Meanwhile [`rollouts.py:2331`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/experience/rollouts.py#L2331) has already `setdefault`-ed `length_adjusted_reward` to the raw env reward — so **`env_reward` is counted twice at weight 1.0** and length pressure is zero.
- **Flat (the workaround):** the length config builds, but the estimator then resolves the feature name to `group_total_length_penalty_coeff`, which is never recorded → `_feature_reward(None)` → `0.0` for every row → **every advantage is exactly zero**.

The doc and the test agree on the nested shape, so the code is the side to change.

**Action:** apply the extractor to the nested block (`default_features["length_adjusted_reward"]`, ignoring a `weight` key) rather than to `default_features`, and raise when a `reward_features` entry names `length_adjusted_reward` but no length params survive the merge — this must not fail quiet. Nothing currently covers the merge path (`grep "_merge_gdpo_reward_features_into_length_cfg\|_gdpo_feature_only" tests/` → no hits); a test starting from the YAML in the design doc would catch it, whereas one that builds the config programmatically will not.

nemo_rl/algorithms/grpo.py
```diff
@@ -3938,6 +3939,7 @@ def validate(
                     greedy=False,
                     effort_config=_get_effort_config(master_config),
                     reward_penalty_config=master_config.reward_penalties,
+                    length_adjustment_config=master_config.grpo.model_dump(),
```

**TL;DR — a group-relative shaper cannot be correctly strided on the validation path, so distinct prompts get scored against each other and against prompt 0's profile band; the mutated reward then flows into `accuracy`/`pass_k`, which drive checkpoint selection and early stopping.**

PR-introduced (this line is the only addition in the `validate()` hunk).

**The stride is wrong here, independent of any policy question.** [`rollouts.py:2719`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/experience/rollouts.py#L2719) passes `num_generations=input_batch.size`, so the whole batch arrives as one group; [`length_adjustments.py:288`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/utils/length_adjustments.py#L288) then re-splits it with `master_config["grpo"]["num_generations_per_prompt"]` — the **training** stride. At train time the batch layout genuinely uses that stride, so [:3018](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/grpo.py#L3018) is fine. At validation the layout uses `val_num_generations_per_prompt` ([`grpo.py:3904`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/grpo.py#L3904)).

`examples/nemo_gym/grpo_nanov3.yaml` (touched by this PR) has `num_generations_per_prompt: 16` and `val_num_generations_per_prompt: 4` — so four distinct prompts are bundled into one pseudo-group, all scored against `results[g]`'s `profile_band` and against each other's lengths. Measured with four prompts each carrying their own band: `[0.5, 0.75, 0.75, 0.75]` expected, `[0.5, 0.5, 0.5, 0.5]` actual.

**Consequence:** rewards are mutated in place at [`length_adjustments.py:600`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/utils/length_adjustments.py#L600) with no clamp, and flow to `accuracy = rewards.mean()` ([:4002](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/grpo.py#L4002)) and `pass_k` ([:4009](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/grpo.py#L4009)) → `grpo_save_state.val_reward` ([:3591](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/grpo.py#L3591)) → `checkpointing.metric_name: val:accuracy` top-k selection and `stop_at_validation_metric`. Measured with all rollouts correct: accuracy 1.0 → 0.25 and pass_k 1.0 → 0.5 under `longest_total_penalty: 1.5`, because a correct rollout is driven to −0.5.

**Action:** drop this line. If shaped validation rewards are wanted for logging, record them under a separate metric key and leave `accuracy`/`pass_k` on the raw env reward.

*(Noting the sibling `reward_penalty_config=` on the line above is pre-existing — I'm not asking you to change that; the argument here is the stride, which is specific to a group-relative shaper.)*

nemo_rl/utils/length_penalty.py
```diff
@@ -0,0 +1,1236 @@
...
+            continue
+        ch_total = band.get("total") if use_total else None
+        ch_rsn = band.get("reasoning") if use_rsn else None
+        ch_ans = band.get("answer") if use_ans else None
+        for k in range(group_size):
+            idx = g + k
+            # Gate on the env reward (correct rollouts only).
+            if original_rewards[idx] <= 0:
+                continue
+            current_reward = base_rewards[idx]
```
yfw, line 697: **1 action item.**

PR-introduced. The gate reads `original_rewards[idx]` but the multiplication is applied to `base_rewards[idx]` — the value *after* additive penalties. Once additive penalties push the base negative, `base * m - base` with `m < 1` is **positive** and grows as `m` shrinks, so among negative-base rows **longer ⇒ higher reward**.

Reproduced with ordinary documented knobs (`longest_total_penalty: 1.2`, `longest_reasoning_penalty: 1.2`, `profile_band_total: true`, band `{a: 100, b: 1000, f: 0.6}`), group of 4, all correct:

```
idx=0 len= 100  additive=+0.000  mult=1.000  band_delta=+0.000  final=+1.000
idx=1 len= 400  additive=+0.000  mult=0.867  band_delta=-0.133  final=+0.867
idx=2 len=1000  additive=-1.200  mult=0.600  band_delta=+0.080  final=-0.120
idx=3 len=1400  additive=-1.200  mult=0.600  band_delta=+0.080  final=-0.120
```

The two longest rollouts receive a **positive** band contribution while the shorter, un-penalised one is docked.

**Action:** gate on the value actually being scaled. (Clamping `additive_base_rewards` at 0 at [:600](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/utils/length_adjustments.py#L600) would also fix it, and would additionally bound the unclamped-negative reward that reaches `pass_k` on the validation path.)

```suggestion
            if original_rewards[idx] <= 0 or base_rewards[idx] <= 0:
                continue
            current_reward = base_rewards[idx]
```

nemo_rl/utils/length_adjustments.py
```diff
@@ -0,0 +1,1236 @@
...
+            defaults[k] = default_cfg.get(k, True)
+        elif k in _BOOL_PARAM_KEYS:
+            defaults[k] = default_cfg.get(k, False)
+        elif k == "profiled_length_min_samples":
+            defaults[k] = default_cfg.get(k, 2)
+        elif k == "profiled_length_n_std":
+            defaults[k] = default_cfg.get(k, 1.0)
+        else:
+            defaults[k] = default_cfg.get(k, 0.0)
+    defaults.setdefault("top_percentile", 0.2)
```
yfw: **2 action items.**

**TL;DR — `length_bonus` is never validated (it rides in via `GRPOConfig`'s `extra="allow"`), so a dropped key, a typo or an unsupported key is a silent no-op. Three live instances below; the ask is to fail loudly, not to build a typed model.**

PR-introduced. Three concrete cases:

1. **This line is dead code, and the effective default is `0.0`.** `top_percentile` is in [`_PARAM_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/utils/length_adjustments.py#L43), so the loop above always assigns it via the final `else` (`default_cfg.get(k, 0.0)`); the key exists by the time `setdefault` runs, so `0.2` never applies. With `0.0`, `threshold_idx = max(0, int(n*0.0) - 1) = 0` → `top_threshold = max(pos_rewards)`, collapsing the "top percentile" band to the best scorer. Three different numbers are in play: the doc says `0.5`, this dead line says `0.2`, behaviour is `0.0`. (Inert under binary rewards — every positive ties at the max — so it only bites with graded/judge rewards and group size ≥ 10, which this PR's own `math_with_judge_simple_agent` example hits.)

2. **Seven keys are silently dropped in GDPO feature mode.** [`_GDPO_LENGTH_FEATURE_PARAM_KEYS`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/utils/length_adjustments.py#L90) has 20 entries against `_PARAM_KEYS`' 28. `enabled` is deliberate (forced True at [:234](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/utils/length_adjustments.py#L234)), but the other seven are lost: the three `*_zmad_threshold`, `top_percentile`, `profiled_length_n_std`, `profiled_length_min_samples`, `length_type`. The zMAD case is the damaging one — `*_zmad_penalty` is **accepted** while its threshold is **dropped**, and the channel only arms `if threshold > 0`, so all three zMAD penalties are permanently dead in GDPO mode. `length_type` means a user writing `length_type: chars` under `reward_features` silently gets **tokens**, and every threshold (`a`, `b`, the zMAD median, the gate limit) is then read in the wrong unit.

3. **An explicit `profile_band_total: false` is overridden.** Lines 312-313 (`for _ch in global_band: defaults[f"profile_band_{_ch}"] = True`) run *after* the loop that honoured the user's value, so any global `profile_band.defaults` block silently re-enables a channel the user turned off. `setdefault` instead of `=` fixes it.

### AI-1
Reject unknown keys under `length_bonus.default` / `agent_overrides` at setup time (and warn on `reward_features` keys that don't survive the merge). The failure mode today is a multi-day run that trains with no length shaping and looks healthy.

### AI-2
Align the three `top_percentile` numbers and delete the dead `setdefault`; add the seven missing keys to `_GDPO_LENGTH_FEATURE_PARAM_KEYS` (or derive it from `_PARAM_KEYS` minus a commented exclusion list, so this can't silently recur); make line 313 a `setdefault`.

### Follow-up
A typed `LengthBonusConfig` mirroring `RewardPenaltyConfig` would retire this whole class — worth a tracking issue, not this PR.

examples/nemo_gym/grpo_nanov3.yaml
```diff
@@ -312,7 +312,7 @@ data:
 env:
   should_use_nemo_gym: true
   # true: skip expensive train_data_step*.jsonl; false: write full jsonl.
-  should_log_nemo_gym_responses: true
+  should_log_nemo_gym_responses: false
```
yfw: **1 action item.** Looks unrelated to length penalties — possibly missed by your own `50244bcf5` ("drop example/recipe configs from the lenpen PR") cleanup.

The polarity is inverted from a perf saving: per [`grpo.py:2230-2238`](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/grpo.py#L2230), *"When **True**, skip the expensive per-step `train_data_step*.jsonl` dump. When **False** (the default if unset), write the local JSONL file"* — and both use sites are `if not _should_log_nemo_gym_responses(...)` ([:3690](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/grpo.py#L3690), [:5456](https://github.com/NVIDIA-NeMo/RL/blob/50244bcf5753665c58f639b9e72d1718716e8459/nemo_rl/algorithms/grpo.py#L5456)). So `true → false` makes this recipe **start** writing the per-step dump it previously skipped.

Three recipes inherit this file via `defaults:` and pick the change up, **two of them nightly**: `recipes/llm/grpo-nanov3-30BA3B-2n8g-megatron_generation-async-gym.yaml` (`nightly.txt:119`), `recipes/llm/grpo-nanov3-30ba3b-3n8g-megatron-dynamo-swe1.yaml` (`nightly.txt:226`), and `recipes/vlm/vlm_grpo-nemotron-super-omni-120ba12b-…-async-gym.v1.yaml` (currently in `disabled.txt`).

A fourth inheritor, `recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-…-async-gym-video.v1.yaml`, sets `should_log_nemo_gym_responses: false` locally at line 93 and is unaffected — which is itself the argument: this flag is deliberately managed per-recipe, so flipping the shared base default changes it for everyone who *hasn't* opted in.

**Action:** revert if unintended.

```suggestion
  should_log_nemo_gym_responses: true
```

yfw (commented):
Local review by a multi-agent pass (RL/config, reward-math, bug-scan, tests, design, plus an adversarial verification round). Every numeric claim below was executed, not inferred.

**The feature's `length_bonus` half looks sound** — I verified that group-relative scaling preserves the group mean exactly, that degenerate groups (size 1, all-equal lengths, all-zero or all-negative rewards, zero lengths) are all safe with no NaN created and no division by zero, that `_band_multiplier`'s boundaries are correct including the past-`b` clamp from `25df5bfcc`, that the zMAD statistic uses the correct Iglewicz–Hoaglin constant, and that a config which doesn't opt in is an exact no-op on rewards. The 661-line design doc is genuinely above the bar for this repo.

**The GDPO feature half is where the problems are**, and two of them need fixing before merge: the dispatch hijack (which affects existing configs with no opt-in) and the fact that the documented config shape produces no length pressure at all. Details inline.

**Follow-ups — not asking for these in this PR:**
- `_ZMAD_MIN_MAD_REL = 0.015` gates the whole channel rather than suppressing noise: lengths `[20000, 20100, 19900, 20050, 19950, 32768]` give MAD/median = 0.0037 < 0.015, so a modified-Z of **114.6** is missed — precisely the tight-distribution-with-one-runaway case the channel exists for. The justifying comment cites `flag_reasoning_length_outliers`, which doesn't exist anywhere in the repo.
- `profiled_length_penalty` has no positive-reward gate, unlike every other channel, so an all-wrong group gains within-group variance (and a gradient) where GRPO previously produced none.
- `_apply_profile_band_multipliers` never checks `enabled`, unlike Phase 1 at L353. Currently masked — skipped groups leave lengths at 0 and `_band_multiplier(0, ch)` returns 1.0 — but it's a live guard gap. Same root as Phase 2 printing coefficients the profile gate had already zeroed: `_resolve_agent_params` is re-resolved independently in three places and the copies have drifted.
- `sync_rollout_actor.py:262` passes `reward_penalty_config` but not `length_adjustment_config` — the same omission `b8ca5c331` fixed for the async collector. No shipped config reaches it today (nothing sets `data_plane.enabled` together with `nemo_gym`), so no impact, but a parametrised "every rollout producer threads the config" test would stop this recurring a third time.
- `scripts/build_profile_band_dataset.py`: `ZeroDivisionError` on empty input (after the output file is written), `TypeError` on an explicit `"pass_rate": null`, and a comment promising a "tiny parser" fallback that doesn't exist wrapped around an unreachable `except ImportError`. Also `tools/` is the established home for standalone utilities.
- `_resolve_agent_params` can never return `None`, so three `if params is None` guards and the verbose `[skipped]` branch are unreachable; and its `print()` warning fires once per group per step (128 unsuppressable prints/step at `num_prompts_per_step: 128`) while the module already has a `logger`.

_Generated by Claude Code_

nemo_rl/algorithms/advantage_estimator.py
arendu: striped GDPO feature from the PR.

nemo_rl/utils/length_penalty.py
arendu: Fixed — we took both suggestions. bbcc4d9df added the gate on the value being scaled (base_rewards[idx] <= 0 → skip), and after discussion we also adopted the clamp (7a4d88897): the additive stage now floors adjusted rewards at 0 for originally-positive rollouts, so stacked flat penalties can wipe a reward out but never flip its sign.
Envs' own negative rewards pass through unclamped. Regression tests cover the inversion scenario, the clamp, and the negative-env-reward passthrough.

nemo_rl/utils/length_adjustments.py
arendu: Addressed by removal: 541be2010 strips GDPO feature mode from this PR entirely (estimator and its tests reverted to base; the merge/extractor and the rollouts.py feature recording — including the length_adjusted_reward setdefault you traced — are deleted). Length adjustments here are grpo.length_bonus reward-mutation only. When feature mode returns as its own PR, it'll take this comment as the spec: nested-block descent, loud failure when a named length feature yields no params, and YAML-driven tests over the documented shape.

nemo_rl/algorithms/grpo.py
yfw: fixed in 87bea4fc8ba109878c3f47896714179ab4f9286b — verified the call site is gone and the comment left in its place names both reasons (raw env reward for metrics, and the train-vs-val stride mismatch).

examples/nemo_gym/grpo_nanov3.yaml
yfw: fixed in f02100ff034052495b5d4b6b77f695144bad8f1c

nemo_rl/utils/length_adjustments.py
yfw: fixed in bbcc4d9df5ca2ec02dfc9269bc8a0ee40260c32f — verified all three: `_reject_unknown_length_penalty_keys` (length_penalty.py:140) now raises on unknown keys, `top_percentile` defaults to 0.5 in the loop (:222) with the dead `setdefault` gone, and the global-band loop (:229-231) no longer overrides an explicitly-set channel flag. The GDPO key-drop instance is moot after 541be2010.

nemo_rl/utils/length_penalty.py
```diff
@@ -0,0 +1,1158 @@
...
+    length_cfg = dict(grpo_config.get("length_penalty", {}) or {})
+    if not length_cfg:
+        return
+
+    _reject_unknown_length_penalty_keys(length_cfg)
+    default_cfg = length_cfg.get("default", {})
+    agents_cfg = length_cfg.get("agent_overrides")
+    global_band = _resolve_global_profile_band(length_cfg.get("profile_band"))
+    verbose = bool(length_cfg.get("verbose", False))
+    if not default_cfg.get("enabled", False) and not agents_cfg and not global_band:
```
yfw: **1 action item.**

`enabled` is read twice with **opposite defaults**, 13 lines apart — `False` here, [`True` at L234](https://github.com/NVIDIA-NeMo/RL/blob/c89b83f7169d3ac3bb6616ed9823742aa40682f9/nemo_rl/utils/length_penalty.py#L234). This early return wins, so a `default:` block that omits `enabled` is a silent no-op.

```yaml
grpo:
  length_penalty:
    default: {group_total_length_penalty_coeff: 0.2}    # no `enabled`
```
→ rewards unchanged `[1.0, 1.0, 1.0, 1.0]`.

The part that makes it a genuine trap rather than a missing-key nit: adding an **empty** `agent_overrides` entry makes `agents_cfg` truthy, skips this return, and L234's `True` then applies — so the *same* `default:` block becomes active:

```yaml
    default: {group_total_length_penalty_coeff: 0.2}
    agent_overrides: {some_agent: {}}                  # empty!
```
→ `[1.1, 1.0333, 0.9667, 0.9]`.

One block's behaviour is decided by an unrelated key. Every example in the design doc writes `enabled: true`, so the docs don't lead anyone into this — but a user trimming a config to its essentials would hit it, and the symptom is "my penalty silently does nothing".

**Action:** make the two reads agree. Matching L234 is the smaller change and treats a configured `default:` block as intent-to-enable:

```suggestion
    if not default_cfg.get("enabled", True) and not agents_cfg and not global_band:
```

One way of several — requiring `enabled` explicitly (and raising when a `default:` block omits it) would also work and is arguably louder.

arendu: Fixed in 164360066, taking your smaller option — enabled now defaults True at both read sites, with a comment pinning the invariant. Added two regression tests: default: {coeff} without enabled applies (your [1.1, ..., 0.9] case), and the same block behaves identically with/without an empty agent_overrides entry.

PR #3854, opened 2026-08-26
fix(algorithms): raise instead of asserting on the skipped-reference-KL pairing

tianyi-zhang-02:
## What does this PR do?

Replaces skipped-reference-logprob configuration asserts in synchronous PPO and GRPO with `ValueError`s, so the safety check still runs under `python -O`.

The GRPO validation also runs at the start of setup, before logger, dataloader, checkpoint, or cluster initialization. The train-loop check remains as a defensive guard.

## Validation

Current head `53d5cd8acfe0282b28172e8184d7b50a5b1e85c5`, rebased onto upstream `main` at `90a2a212d503455d8590be5c8de3cb989d3425b0`.

- `tests/unit/algorithms/test_reference_kl_guard.py`: **5 passed**
- setup coverage confirms the user-facing error occurs before cluster-side work
- refreshed the GRPO save-state fixture for current main's `total_steps` access

This is setup/config validation, so no GPU workload is required.

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -539,6 +539,16 @@ def setup(
     cluster_config = master_config.cluster
     checkpointing_config = master_config.checkpointing
 
+    # Validate this before logger, checkpoint, dataloader, or cluster setup can
+    # produce side effects for a configuration that cannot train correctly.
+    if (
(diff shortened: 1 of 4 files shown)


tianyi-zhang-02:
Closing to reduce the review queue. This is a defensive error-type cleanup rather than a user-visible correctness fix, so it is not worth asking maintainers to review as a standalone PR.

PR #3858, opened 2026-08-26
fix(vllm): realign multimodal placeholders after prefix replacement

aroshanghias-nvd:
# What does this PR do?

Realigns vLLM multimodal placeholder offsets after NeMo-RL replaces re-tokenized conversation history with the exact model-generated token prefix.

In the OpenAI-compatible multi-turn generation path, vLLM first renders the chat template and computes `mm_placeholders` against that token sequence. NeMo-RL then calls `replace_prefix_tokens()` so prior turns use the exact token IDs originally consumed and sampled by the model. When the exact and re-tokenized prefixes have different lengths, replacing only `prompt_token_ids` leaves the multimodal offsets in the old template coordinates. vLLM can then apply media embeddings at incorrect token positions.

This PR:

- relocates each existing media-token span into the final exact-token prompt;
- processes image, audio, and video ranges in global prompt order so per-item shifts are preserved;
- preserves mapping and frozen-dataclass range representations;
- fails closed if an expected media span cannot be located;
- keeps an inexpensive no-op path when the prompt tokens are unchanged or there are no multimodal placeholders.

# Issues

None.

# Usage

No user-facing configuration or API changes are required.

**Pre checks**:

- [x] Added focused unit coverage for accumulated per-item shifts, mixed modalities, mapping/dataclass preservation, and missing-span failure.
- [x] Focused unit validation completed: `3 passed, 42 deselected`.
- [x] No documentation changes are required because this is an internal token/media-alignment correction with no public API or configuration changes.

# Additional Information

The repair belongs in NeMo-RL rather than Gym or vLLM: vLLM's offsets are correct for the prompt it originally preprocesses, and NeMo-RL subsequently changes that token sequence to preserve exact multi-turn rollout tokens.

A 20-turn multimodal no-compaction causal validation after applying the repair completed with TMPE `1.030905` ([W&B run](https://wandb.ai/nvidia/nemo-rl-context-compaction/runs/9nax1zne)).

### DIFF
--- a/nemo_rl/models/generation/vllm/utils.py
+++ b/nemo_rl/models/generation/vllm/utils.py
@@ -50,8 +50,36 @@ def remap_multimodal_placeholders(
     multimodal features remain aligned.
 
     Ranges are matched in global prompt order because different media items can
-    accumulate different shifts. A missing span fails closed rather than
-    submitting token IDs with incorrect multimodal positions.
+    accumulate different shifts. Coincident ranges resolve to the same offset:
+    Qwen2.5-Omni derives an audio range from its paired video range with an
+    identical ``(offset, length)``, so a span is only consumed once.
+
+    A span that cannot be located fails closed rather than submitting token IDs
+    with incorrect multimodal positions. Note that an *ambiguous* match is not
+    detected: media spans are bare runs of one repeated pad token, so if a run
+    is longer in ``final_token_ids`` than in the template, a later item can
+    match inside an earlier item's run. Locating spans by search is only
+    necessary because the splice boundary computed by ``replace_prefix_tokens``
+    is not threaded through to here; passing it would make the suffix region
+    pure arithmetic and remove that ambiguity entirely.
+
+    Args:
+        template_token_ids: The chat-template token sequence vLLM used to
+            compute ``mm_placeholders``.
+        final_token_ids: The exact-token prompt produced by
+            ``replace_prefix_tokens``, which will be submitted to the engine.
+        mm_placeholders: vLLM's per-modality placeholder ranges, in
+            ``template_token_ids`` coordinates.
+
+    Returns:
+        A new per-modality mapping with the same item ordering, whose ranges are
+        expressed in ``final_token_ids`` coordinates.
+
+    Raises:
+        ValueError: If an input range is out of bounds for
+            ``template_token_ids``, or if a media span cannot be relocated in
+            ``final_token_ids``.
+        TypeError: If a range is neither a mapping nor a dataclass instance.
     """
     if template_token_ids == final_token_ids or not mm_placeholders:
         return {modality: list(ranges) for modality, ranges in mm_placeholders.items()}
@@ -76,25 +104,33 @@ def remap_multimodal_placeholders(
             entries.append((offset, modality, item_index, placeholder_range, length))
 
     search_start = 0
+    resolved: dict[tuple[int, int], int] = {}
     for old_offset, modality, item_index, placeholder_range, length in sorted(
         entries, key=lambda entry: entry[0]
     ):
-        expected = template_token_ids[old_offset : old_offset + length]
-        max_start = len(final_token_ids) - length
-        new_offset = next(
-            (
-                candidate
-                for candidate in range(search_start, max_start + 1)
-                if final_token_ids[candidate] == expected[0]
-                and final_token_ids[candidate : candidate + length] == expected
-            ),
-            None,
-        )
+        # Two modalities can describe the same span, so a resolved offset is
+        # reused instead of scanning past it. Only a newly located span
+        # advances the cursor.
+        new_offset = resolved.get((old_offset, length))
         if new_offset is None:
-            raise ValueError(
-                f"Could not locate {modality} placeholder range {item_index} "
-                f"from template offset {old_offset} in the final exact-token prompt"
+            expected = template_token_ids[old_offset : old_offset + length]
+            max_start = len(final_token_ids) - length
+            new_offset = next(
+                (
+                    candidate
+                    for candidate in range(search_start, max_start + 1)
+                    if final_token_ids[candidate] == expected[0]
+                    and final_token_ids[candidate : candidate + length] == expected
+                ),
+                None,
             )
+            if new_offset is None:
+                raise ValueError(
+                    f"Could not locate {modality} placeholder range {item_index} "
+                    f"from template offset {old_offset} in the final exact-token prompt"
+                )
+            resolved[(old_offset, length)] = new_offset
+            search_start = new_offset + length
 
         if isinstance(placeholder_range, Mapping):
             updated_range = dict(placeholder_range)
@@ -108,7 +144,6 @@ def remap_multimodal_placeholders(
             )
 
         remapped[modality][item_index] = updated_range
-        search_start = new_offset + length
 
     return remapped
 
--- a/nemo_rl/models/generation/vllm/utils.py
+++ b/nemo_rl/models/generation/vllm/utils.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 from collections import defaultdict
+from collections.abc import Mapping
+from dataclasses import is_dataclass, replace
 from typing import Any, Optional
 
 import torch
@@ -31,6 +33,84 @@ VLLM_LOGPROB_FLOOR = -9999.0
 # The expert-id range vs carry dtype is model-constant, so it is verified on the
 # first non-empty routed-experts tensor per process and skipped afterwards.
 G_ROUTED_EXPERTS_RANGE_CHECKED = False
+
+
+def remap_multimodal_placeholders(
+    *,
+    template_token_ids: list[int],
+    final_token_ids: list[int],
+    mm_placeholders: Mapping[str, list[Any]],
+) -> dict[str, list[Any]]:
+    """Move vLLM multimodal ranges into the final prompt coordinates.
+
+    vLLM computes multimodal placeholder offsets while preprocessing the
+    chat-template token sequence. NeMo-RL can subsequently replace the
+    re-tokenized history with the exact model-generated token prefix. Locate
+    each unchanged media-token span in that final sequence so its associated
+    multimodal features remain aligned.
+
+    Ranges are matched in global prompt order because different media items can
+    accumulate different shifts. A missing span fails closed rather than
+    submitting token IDs with incorrect multimodal positions.
+    """
+    if template_token_ids == final_token_ids or not mm_placeholders:
+        return {modality: list(ranges) for modality, ranges in mm_placeholders.items()}
+
+    entries: list[tuple[int, str, int, Any, int]] = []
+    remapped = {modality: list(ranges) for modality, ranges in mm_placeholders.items()}
+    for modality, ranges in mm_placeholders.items():
+        for item_index, placeholder_range in enumerate(ranges):
+            if isinstance(placeholder_range, Mapping):
+                offset = int(placeholder_range["offset"])
+                length = int(placeholder_range["length"])
+            else:
+                offset = int(placeholder_range.offset)
+                length = int(placeholder_range.length)
+
+            if offset < 0 or length <= 0 or offset + length > len(template_token_ids):
+                raise ValueError(
+                    f"Invalid {modality} placeholder range {item_index}: "
+                    f"offset={offset}, length={length}, "
+                    f"template_length={len(template_token_ids)}"
+                )
+            entries.append((offset, modality, item_index, placeholder_range, length))
+
+    search_start = 0
+    for old_offset, modality, item_index, placeholder_range, length in sorted(
+        entries, key=lambda entry: entry[0]
+    ):
+        expected = template_token_ids[old_offset : old_offset + length]
+        max_start = len(final_token_ids) - length
+        new_offset = next(
+            (
+                candidate
+                for candidate in range(search_start, max_start + 1)
+                if final_token_ids[candidate] == expected[0]
+                and final_token_ids[candidate : candidate + length] == expected
+            ),
+            None,
+        )
+        if new_offset is None:
+            raise ValueError(
+                f"Could not locate {modality} placeholder range {item_index} "
+                f"from template offset {old_offset} in the final exact-token prompt"
+            )
+
+        if isinstance(placeholder_range, Mapping):
+            updated_range = dict(placeholder_range)
+            updated_range["offset"] = new_offset
+        elif is_dataclass(placeholder_range):
+            updated_range = replace(placeholder_range, offset=new_offset)
+        else:
+            raise TypeError(
+                "Multimodal placeholder ranges must be mappings or dataclass "
+                f"instances, got {type(placeholder_range).__name__}"
+            )
+
+        remapped[modality][item_index] = updated_range
+        search_start = new_offset + length
+
+    return remapped
 
 
 def _as_routed_experts_tensor(
--- a/nemo_rl/models/generation/vllm/vllm_worker_async.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker_async.py
@@ -49,6 +49,7 @@ from nemo_rl.models.generation.vllm.utils import (
     format_prompt_for_vllm_generation,
     model_dump_chat_response_with_dynamic_message_fields,
     pad_and_align_routed_expert_indices,
+    remap_multimodal_placeholders,
 )
 from nemo_rl.models.generation.vllm.vllm_worker import BaseVllmGenerationWorker
 from nemo_rl.models.generation.openai_server_utils import (
@@ -560,6 +561,13 @@ class VllmAsyncGenerationWorkerImpl(
                     template_prefix_token_ids=actual_corresponding_token_ids,
                     template_token_ids=engine_prompt["prompt_token_ids"],
                 )
+
+                if mm_placeholders := engine_prompt.get("mm_placeholders"):
+                    engine_prompt["mm_placeholders"] = remap_multimodal_placeholders(
+                        template_token_ids=engine_prompt["prompt_token_ids"],
+                        final_token_ids=final_prompt_token_ids,
+                        mm_placeholders=mm_placeholders,
+                    )
 
                 engine_prompt["prompt_token_ids"] = final_prompt_token_ids
 
--- a/nemo_rl/models/generation/vllm/vllm_worker_async.py
+++ b/nemo_rl/models/generation/vllm/vllm_worker_async.py
@@ -562,6 +562,10 @@ class VllmAsyncGenerationWorkerImpl(
                     template_token_ids=engine_prompt["prompt_token_ids"],
                 )
 
+                # Must precede the reassignment below: the remap reads the
+                # current value of prompt_token_ids as the template-coordinate
+                # sequence the placeholders were computed against. Reordering
+                # these two statements silently degrades to a no-op.
                 if mm_placeholders := engine_prompt.get("mm_placeholders"):
                     engine_prompt["mm_placeholders"] = remap_multimodal_placeholders(
                         template_token_ids=engine_prompt["prompt_token_ids"],
--- a/tests/unit/models/generation/test_vllm_utils.py
+++ b/tests/unit/models/generation/test_vllm_utils.py
@@ -55,7 +55,7 @@ def test_remap_multimodal_placeholders_handles_accumulated_per_item_shift():
     final_token_ids = [1, *media_tokens, 2, *media_tokens, 3, 21, 21, 22]
     placeholders = {
         # Deliberately list modalities in the opposite order from prompt order.
-        "audio": [{"offset": 13, "length": 3}],
+        "audio": [_PlaceholderRange(offset=13, length=3)],
         "image": [
             _PlaceholderRange(offset=1, length=4),
             _PlaceholderRange(offset=7, length=4, is_embed=embed_mask),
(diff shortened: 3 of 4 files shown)


tests/unit/models/generation/test_vllm_utils.py
```diff
@@ -35,10 +36,54 @@
...
+
+    assert [item.offset for item in result["image"]] == [1, 6]
+    assert result["image"][1].is_embed is embed_mask
+    assert result["audio"] == [{"offset": 11, "length": 3}]
+
+
+def test_remap_multimodal_placeholders_fails_closed_when_span_is_missing():
+    with pytest.raises(
+        ValueError, match="Could not locate image placeholder range 0"
+    ):
```
aroshanghias-nvd: **1 action item.**

[`tests/unit/models/generation/test_vllm_utils.py:77`](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/tests/unit/models/generation/test_vllm_utils.py#L77-L79) — PR-introduced. This file is not ruff-formatted, so the `pre-commit` gate will fail CI. Verified with the repo-pinned `ruff==0.9.9` ([`.pre-commit-config.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/.pre-commit-config.yaml#L11)): `ruff format --diff` reports `1 file would be reformatted`. `ruff check`, import-sort, and both source files are clean.

### AI-1

```suggestion
    with pytest.raises(ValueError, match="Could not locate image placeholder range 0"):
```

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -33,6 +35,84 @@
...
+        elif is_dataclass(placeholder_range):
+            updated_range = replace(placeholder_range, offset=new_offset)
+        else:
+            raise TypeError(
+                "Multimodal placeholder ranges must be mappings or dataclass "
+                f"instances, got {type(placeholder_range).__name__}"
+            )
+
+        remapped[modality][item_index] = updated_range
+        search_start = new_offset + length
```
aroshanghias-nvd: **1 action item.**

**TL;DR — Qwen2.5-Omni emits an audio range with byte-identical `(offset, length)` to its video range, and the monotonic `search_start` makes the second one unfindable, so the request dies with a spurious `ValueError`.**

PR-introduced. Not reachable today, but it blocks the backlog item already recorded as `BL-20260428-omni-use-audio-in-video` ([`intent.py:26-29`](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/data/datasets/response_datasets/intent.py#L26-L29)).

**How it breaks.** With `use_audio_in_video=True`, vLLM derives the audio placeholder from the video placeholder using the *same* `start_idx` and the *same* `tokens`, differing only in `is_embed` — [`qwen2_5_omni_thinker.py` L552-566](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/models/qwen2_5_omni_thinker.py#L552-L566) — and [`to_range()`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/multimodal/processing/processor.py#L687-L694) sets `offset=start_idx, length=len(tokens)`. So both `PlaceholderRange`s are identical.

1. Both entries sort to the same `old_offset`.
2. The first resolves and sets [`search_start = new_offset + length`](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/utils.py#L111) — past the only occurrence.
3. The second searches from there, finds nothing, and [raises](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/utils.py#L94-L97).

Reproduced against this commit: `template=[1,2,VS,S,VE]`, `final=[1,VS,S,VE]`, `{"video":[PR(2,3)],"audio":[PR(2,3)]}` → `ValueError: Could not locate audio placeholder range 0 from template offset 2`.

It fails closed, so no silent corruption — but the `ValueError` is raised outside the [`except (ValueError, VLLMValidationError)`](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/vllm_worker_async.py#L488-L498) above, and vLLM's route doesn't guard the handler, so it surfaces as an opaque HTTP 500 rather than a structured 400. Per the comment at [`nemo_rl/models/generation/vllm/vllm_worker_async.py`](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/vllm_worker_async.py#L585) that hits `/tokenize` as well as `/v1/chat/completions`.

### AI-1

Memoize resolved offsets and advance `search_start` only for newly-resolved spans. Verified to fix both the one-video and two-video coincident cases:

```suggestion
    search_start = 0
    resolved: dict[tuple[int, int], int] = {}
    for old_offset, modality, item_index, placeholder_range, length in sorted(
        entries, key=lambda entry: entry[0]
    ):
        # Modalities can share an identical span -- Qwen2.5-Omni derives the audio
        # range from the video range, differing only in `is_embed` -- so reuse a
        # resolved offset instead of scanning past it.
        cached_offset = resolved.get((old_offset, length))
        if cached_offset is not None:
            new_offset = cached_offset
        else:
            expected = template_token_ids[old_offset : old_offset + length]
            max_start = len(final_token_ids) - length
            new_offset = next(
                (
                    candidate
                    for candidate in range(search_start, max_start + 1)
                    if final_token_ids[candidate] == expected[0]
                    and final_token_ids[candidate : candidate + length] == expected
                ),
                None,
            )
            if new_offset is None:
                raise ValueError(
                    f"Could not locate {modality} placeholder range {item_index} "
                    f"from template offset {old_offset} in the final exact-token prompt"
                )
            resolved[(old_offset, length)] = new_offset

        if isinstance(placeholder_range, Mapping):
            updated_range = dict(placeholder_range)
            updated_range["offset"] = new_offset
        elif is_dataclass(placeholder_range):
            updated_range = replace(placeholder_range, offset=new_offset)
        else:
            raise TypeError(
                "Multimodal placeholder ranges must be mappings or dataclass "
                f"instances, got {type(placeholder_range).__name__}"
            )

        remapped[modality][item_index] = updated_range
        search_start = max(search_start, new_offset + length)
```

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -33,6 +35,84 @@
...
+
+    vLLM computes multimodal placeholder offsets while preprocessing the
+    chat-template token sequence. NeMo-RL can subsequently replace the
+    re-tokenized history with the exact model-generated token prefix. Locate
+    each unchanged media-token span in that final sequence so its associated
+    multimodal features remain aligned.
+
+    Ranges are matched in global prompt order because different media items can
+    accumulate different shifts. A missing span fails closed rather than
+    submitting token IDs with incorrect multimodal positions.
```
aroshanghias-nvd: **1 action item.**

**TL;DR — "fails closed" is not guaranteed: if a media run is longer in the exact prefix than in the template, the next item silently matches *inside* the previous item's run.**

PR-introduced. Distinct from the coincident-range issue above — that one is a duplicate-key collision, this one is an unanchored first-match. Confirmed by measurement: the memoization fix above resolves the coincident case and leaves this one bit-identical.

**Repro on this commit** (executed, no exception raised):

```python
template_token_ids = [1, *[pad] * 4, 2, *[pad] * 4, 3]
final_token_ids    = [1, *[pad] * 8, 2, *[pad] * 4, 3]
mm_placeholders    = {"video": [PR(offset=1, length=4), PR(offset=6, length=4)]}
# -> offsets [1, 5];  correct answer [1, 10]
```

Item 0 matches at the run start, but [`search_start = new_offset + length`](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/utils.py#L111) stops *inside* the grown run, so item 1 matches item 0's own pad tokens. The general form is the same: [Qwen2-VL placeholder spans are bare runs of one repeated token](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/model_executor/models/qwen2_vl.py#L1130-L1137) with no `<|vision_start|>`/`<|vision_end|>` bracketing, so a first-match search has nothing anchoring it.

I could not demonstrate a trigger on any in-repo path — still images have deterministic pad counts, `nemotron_vl` samples deterministically, and the other video path sends `video_as_images=True`. So this is hardening plus docstring accuracy, not a live break.

### AI-1

Either add a run-boundary check (reject a match whose neighbouring tokens disagree with the template) or drop the fail-closed claim. One way to do it — of several:

```suggestion
    Ranges are matched in global prompt order because different media items can
    accumulate different shifts. A span that cannot be located fails closed, but
    note that an ambiguous match is not currently detected: if a media run is
    longer in ``final_token_ids`` than in the template, a later item can match
    inside an earlier item's run.
```

Please don't take the ambiguity-rejection route (raise if a second occurrence exists) — two identical images legitimately produce two identical spans, so that would fail closed on the most ordinary multi-image prompt.

nemo_rl/models/generation/vllm/vllm_worker_async.py
```diff
@@ -561,6 +562,13 @@ async def preprocess_chat(
                     template_token_ids=engine_prompt["prompt_token_ids"],
                 )
 
+                if mm_placeholders := engine_prompt.get("mm_placeholders"):
+                    engine_prompt["mm_placeholders"] = remap_multimodal_placeholders(
+                        template_token_ids=engine_prompt["prompt_token_ids"],
+                        final_token_ids=final_prompt_token_ids,
+                        mm_placeholders=mm_placeholders,
+                    )
+
```
aroshanghias-nvd, line 575: **1 action item.**

PR-introduced. This block has a load-bearing order dependency that nothing encodes: `remap_multimodal_placeholders` must read `engine_prompt["prompt_token_ids"]` while it still holds the **template** ids, and [line 572](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/vllm_worker_async.py#L572) overwrites that key with `final_prompt_token_ids`.

Swap the two statements — the natural "assign the result where you compute it" refactor — and `template_token_ids == final_token_ids`, so the function takes its [early-return fast path](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/utils.py#L56), placeholders stay in template coordinates, and no exception fires. That silently reintroduces the exact bug this PR fixes. `preprocess_chat` is defined inside a method body ([L456](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/vllm_worker_async.py#L456)) so it can't be imported, and no unit test references it — nothing would go red.

### AI-1

```suggestion
                # Must run before `prompt_token_ids` is reassigned below: the remap
                # reads the current value as the template-coordinate token sequence.
                if mm_placeholders := engine_prompt.get("mm_placeholders"):
                    engine_prompt["mm_placeholders"] = remap_multimodal_placeholders(
                        template_token_ids=engine_prompt["prompt_token_ids"],
                        final_token_ids=final_prompt_token_ids,
                        mm_placeholders=mm_placeholders,
                    )
```

### Follow-up

Optionally fold both statements into one `apply_exact_prefix_token_ids(engine_prompt, final_token_ids)` helper in `vllm/utils.py`, which would make the contract testable against a plain dict. Worth it for the test, though it moves the ordering hazard inside the helper rather than removing it — your call.

tests/unit/models/generation/test_vllm_utils.py
```diff
@@ -35,10 +36,54 @@
...
+        remap_multimodal_placeholders(
+            template_token_ids=[1, 18, 18, 2],
+            final_token_ids=[1, 2],
+            mm_placeholders={"image": [_PlaceholderRange(offset=1, length=2)]},
+        )
```
aroshanghias-nvd, line 121: **1 action item.**

Four of the new function's branches are unreached by any test. All four tests below were run against this commit and pass. `SimpleNamespace` is already imported at [line 17](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/tests/unit/models/generation/test_vllm_utils.py#L17), so no new imports are needed.

Worth noting one asymmetry they expose: the [fast path](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/vllm/utils.py#L56) returns *before* range validation, so an invalid range is silently accepted when the prompt is unchanged but rejected when it changed.

### AI-1

```python
def test_remap_multimodal_placeholders_leaves_an_unchanged_prompt_alone():
    ranges = [_PlaceholderRange(offset=1, length=2)]
    placeholders = {"image": ranges}
    token_ids = [1, 18, 18, 2]

    result = remap_multimodal_placeholders(
        template_token_ids=token_ids,
        final_token_ids=list(token_ids),
        mm_placeholders=placeholders,
    )

    assert result == placeholders
    assert result is not placeholders
    assert result["image"] is not ranges


def test_remap_multimodal_placeholders_without_media_returns_empty():
    assert (
        remap_multimodal_placeholders(
            template_token_ids=[1, 2, 3],
            final_token_ids=[9, 9, 1, 2, 3],
            mm_placeholders={},
        )
        == {}
    )


@pytest.mark.parametrize("offset, length", [(-1, 2), (1, 0), (1, -2), (2, 3)])
def test_remap_multimodal_placeholders_rejects_out_of_range_input(offset, length):
    with pytest.raises(ValueError, match="Invalid image placeholder range 0"):
        remap_multimodal_placeholders(
            template_token_ids=[1, 18, 18, 2],
            final_token_ids=[9, 1, 18, 18, 2],
            mm_placeholders={
                "image": [_PlaceholderRange(offset=offset, length=length)]
            },
        )


def test_remap_multimodal_placeholders_rejects_an_unknown_range_type():
    with pytest.raises(TypeError, match="got SimpleNamespace"):
        remap_multimodal_placeholders(
            template_token_ids=[1, 18, 18, 2],
            final_token_ids=[9, 1, 18, 18, 2],
            mm_placeholders={"image": [SimpleNamespace(offset=1, length=2)]},
        )
```

### Follow-up

A red test documenting the grown-run case from the docstring comment (currently returns `[1, 5]` instead of `[1, 10]`). Land it with whichever fix you choose, and assert bare `pytest.raises(ValueError)` rather than matching a message — the natural fix would emit a different string than the existing "could not locate" template.

tests/unit/models/generation/test_vllm_utils.py
```diff
@@ -35,10 +36,54 @@
...
+
+    result = remap_multimodal_placeholders(
+        template_token_ids=template_token_ids,
+        final_token_ids=final_token_ids,
+        mm_placeholders=placeholders,
+    )
+
+    assert [item.offset for item in result["image"]] == [1, 6]
+    assert result["image"][1].is_embed is embed_mask
+    assert result["audio"] == [{"offset": 11, "length": 3}]
```
aroshanghias-nvd: **1 action item.** Low severity.

PR-introduced. This assertion certifies an output shape the engine cannot consume. vLLM only ever emits `PlaceholderRange` — the engine-prompt TypedDict pins it as [`Mapping[str, Sequence[PlaceholderRange]]`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/inputs/engine.py#L153) — and every consumer uses attribute access: [`x[2].offset` in `argsort_mm_positions`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/multimodal/utils.py#L155), [`mm_position.get_num_embeds()`](https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/engine/input_processor.py#L458). A plain dict would `AttributeError` inside the engine.

### AI-1

Change the `audio` fixture at [line 65](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/tests/unit/models/generation/test_vllm_utils.py#L65) to a `_PlaceholderRange` and assert its `.offset`, so the test only certifies shapes the engine accepts.

**Context — no action.** The `Mapping` branch itself is harmless: it costs four lines and cannot fire at the pinned version. Keeping it for cross-version tolerance is defensible — the ask here is only about the test.

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -33,6 +35,84 @@
...
+        max_start = len(final_token_ids) - length
+        new_offset = next(
+            (
+                candidate
+                for candidate in range(search_start, max_start + 1)
+                if final_token_ids[candidate] == expected[0]
+                and final_token_ids[candidate : candidate + length] == expected
+            ),
+            None,
+        )
```
aroshanghias-nvd: **No action needed in this PR** — raising it as a follow-up because it subsumes the two search-related findings above.

The remap searches for information the caller already has exactly. [`replace_prefix_tokens`](https://github.com/NVIDIA-NeMo/RL/blob/ff3560d4faac4ed97795754916194f7e103a6f75/nemo_rl/models/generation/openai_server_utils.py#L127-L129) is a splice at a known boundary — it computes `template_cut_start` and `model_cut_end`, returns only the concatenation, and discards both indices. Given them, every span in the suffix region (`old_offset >= template_cut_start`) is pure arithmetic:

```
new_offset = old_offset - template_cut_start + model_cut_end
```

with a slice-equality assertion turning "fails closed" from a hope into a verified postcondition. Only spans landing inside the replaced prefix would still need a search, bounded to `final[:model_cut_end]`. That removes the ambiguity class entirely rather than hardening the scan against it.

It touches a second file and widens a shared signature, so it's genuinely a follow-up rather than something to do here.

aroshanghias-nvd (commented):
Team review of #3858 — 5 agents (RL expert, bug finder, test agent, design reviewer, devil's advocate), each finding independently challenged before staging.

**The change is needed and correctly layered.** `mm_placeholders` carries offsets into the *chat-template* token sequence, and the existing code then replaces `prompt_token_ids` with a different-length splice — so media embeddings land on the wrong tokens with no error at all. Nice catch, and the fail-closed instinct is the right one. The two correctness findings below are both about the search *technique*, not the premise.

**One CI blocker:** the new test file isn't ruff-formatted, so `pre-commit` will fail.

**Verified and cleared,** so nobody re-litigates them: the call-site statement ordering is correct as written; the walrus guard's empty/missing cases are right; out-of-order ranges within a modality write back correctly; `dataclasses.replace` on the frozen `PlaceholderRange` handles the `cached_property` cleanly; the `mm_placeholders` key and its consumption by `InputProcessor` are correct at the pinned vLLM 0.25.1; and the three other `replace_prefix_tokens` call sites (TRT-LLM, Dynamo, the sync worker) genuinely need no remap. `mm_hashes` correctly needs no remapping either — hashes are over media content, not positions. No performance evidence is needed here: the call-site guard, the first-token short-circuit, and the monotonic `search_start` bound this to a single left-to-right pass.

**Two things one field over, both latent, neither for this PR.** `assistant_tokens_mask` on the same `MultiModalInput` is a per-token mask over `prompt_token_ids` and goes stale on the same splice — NeMo-RL never sets `return_assistant_tokens_mask`, so it's dormant. And `super().preprocess_chat` is called twice in this override (L487 and L544), the second re-decoding every image in the history; pre-existing, but it's the real cost on this path.

_Generated by Claude Code_

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -33,6 +35,119 @@
...
+
+    Returns:
+        A new per-modality mapping with the same item ordering, whose ranges are
+        expressed in ``final_token_ids`` coordinates.
+
+    Raises:
+        ValueError: If an input range is out of bounds for
+            ``template_token_ids``, or if a media span cannot be relocated in
+            ``final_token_ids``.
+        TypeError: If a range is neither a mapping nor a dataclass instance.
```
yfw: **1 action item (optional).**

**TL;DR — `Raises:` documents a `TypeError` that only fires for attribute-bearing non-dataclasses; a range without `offset`/`length` hits `AttributeError` first, and the unchanged-prompt fast path raises nothing at all.**

[`nemo_rl/models/generation/vllm/utils.py:82`](https://github.com/NVIDIA-NeMo/RL/blob/5282b117ea7f05a8bb0e331ec71897d417ba314b/nemo_rl/models/generation/vllm/utils.py#L82) — PR-introduced. `Raises:` promises `TypeError` when a range is neither a mapping nor a dataclass, but the first loop (<a href="https://github.com/NVIDIA-NeMo/RL/blob/5282b117ea7f05a8bb0e331ec71897d417ba314b/nemo_rl/models/generation/vllm/utils.py#L94-L96">L94-L96</a>) reads `.offset` on every non-mapping object before any dataclass check, so a range without those attributes raises `AttributeError` — `{"image": [(1, 2)]}` gives `AttributeError: 'tuple' object has no attribute 'offset'`. The `TypeError` at <a href="https://github.com/NVIDIA-NeMo/RL/blob/5282b117ea7f05a8bb0e331ec71897d417ba314b/nemo_rl/models/generation/vllm/utils.py#L140-L144">L140-L144</a> only fires for attribute-bearing non-dataclasses, and the fast path at <a href="https://github.com/NVIDIA-NeMo/RL/blob/5282b117ea7f05a8bb0e331ec71897d417ba314b/nemo_rl/models/generation/vllm/utils.py#L84">L84</a> returns before either raise. Not reachable from vLLM 0.25.1, which only emits <a href="https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/inputs/engine.py#L153"><code>PlaceholderRange</code></a> — accuracy only.

### AI-1
Docstring-only (ruff-format clean). If you would rather make the code match the docstring: `elif is_dataclass(placeholder_range): offset = int(placeholder_range.offset)` fails pyrefly (`is_dataclass` narrows to `DataclassInstance`, +4 missing-attribute errors); a guard-first `if not is_mapping and not is_dataclass(placeholder_range): raise TypeError(...)` before the reads type-checks clean.

```suggestion
        TypeError: If a range that exposes ``offset`` and ``length`` is neither
            a mapping nor a dataclass instance. A range without those
            attributes fails earlier with ``AttributeError``.
```

nemo_rl/models/generation/vllm/utils.py
```diff
@@ -33,6 +35,119 @@
...
+    identical ``(offset, length)``, so a span is only consumed once.
+
+    A span that cannot be located fails closed rather than submitting token IDs
+    with incorrect multimodal positions. Note that an *ambiguous* match is not
+    detected: media spans are bare runs of one repeated pad token, so if a run
+    is longer in ``final_token_ids`` than in the template, a later item can
+    match inside an earlier item's run. Locating spans by search is only
+    necessary because the splice boundary computed by ``replace_prefix_tokens``
+    is not threaded through to here; passing it would make the suffix region
+    pure arithmetic and remove that ambiguity entirely.
```
yfw: **1 action item (optional).**

[`nemo_rl/models/generation/vllm/utils.py:64`](https://github.com/NVIDIA-NeMo/RL/blob/5282b117ea7f05a8bb0e331ec71897d417ba314b/nemo_rl/models/generation/vllm/utils.py#L64) — PR-introduced. <a href="https://github.com/NVIDIA-NeMo/RL/blob/5282b117ea7f05a8bb0e331ec71897d417ba314b/nemo_rl/models/generation/openai_server_utils.py#L111-L129"><code>replace_prefix_tokens</code></a> appends the template suffix verbatim after an EOS cut, so a media run can only grow inside the replaced prefix, i.e. for earlier-turn media. Threading the splice boundary would make suffix spans pure arithmetic, but earlier-turn spans would still be located by a (bounded) search — it confines the ambiguity rather than removing it. Worth stating precisely, since the follow-up will be built from this sentence.

### AI-1
```suggestion
    pure arithmetic and confine that ambiguity to earlier-turn media.
```

PR #3860, opened 2026-08-26
build: mount token secret for apptainer dependency download

kajalj22:
## Summary

sbsa (arm64) builds fail with HTTP 429 fetching Apptainer's e2fsprogs patches during the custom-setup step, due to GitHub's unauthenticated per-IP rate limit (60/hr) on the shared CI runner.

Apptainer's own `scripts/download-dependencies` (run by `install_apptainer.sh`'s arm64 source build) supports an authenticated fallback via `$GITHUB_TOKEN`. This mounts that secret into the custom-setup `RUN` step.

The secret `id` is `APPTAINER_GITHUB_TOKEN` (not the generic `GITHUB_TOKEN`), so it's structurally clear this slot expects a dedicated, minimally-scoped credential (public-repo, read-only) rather than a broader-scope org token. The env var exposed *inside* the RUN step stays `GITHUB_TOKEN`, since that's the literal name Apptainer's upstream script checks for.

Pairs with dl/JoC/nemo-ci!2850, which forwards `APPTAINER_GITHUB_TOKEN` as this buildx secret.

## Verification

Manual pipeline run on both paired branches (single nightly test, `cw_dfw`, forced fresh build): https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65781582

## Test plan

- [ ] Verify arm64 image build succeeds with the secret supplied
- [ ] Confirm no HTTP 429 in the apptainer dependency download step


chtruong814 (approved):

kajalj22:
Tests also failing on the main branch - not due to this PR

PR #3864, opened 2026-08-27
perf: overlap non-colocated MInf + polict init

### LINKED ISSUE #3777: Megatron generation init is serialized behind policy init; other backends parallelize it
Megatron is the only generation backend whose policy init and generation init run one after the other. Every other backend runs them in parallel for non-colocated setups. The result is that megatron's setup wall clock is `policy + generation` where vLLM's is `max(policy, generation)`.

Permalinks pinned at `f0557321e` (main at time of writing).

## Every other backend parallelizes; megatron does not

`initialize_generation_with_policy` ([`grpo.py:1217`](https://github.com/NVIDIA-NeMo/RL/blob/f0557321e/nemo_rl/algorithms/grpo.py#L1217)) runs generation and policy init concurrently when inference is not colocated:

```python
use_parallel_init = not colocated_inference
...
with ThreadPoolExecutor(max_workers=2) as executor:
    generation_future = executor.submit(init_generation_fn)
    policy_future = executor.submit(init_policy)
```

It is used by vLLM, SGLang, TRT-LLM and Dynamo ([L1417](https://github.com/NVIDIA-NeMo/RL/blob/f0557321e/nemo_rl/algorithms/grpo.py#L1417), [L1435](https://github.com/NVIDIA-NeMo/RL/blob/f0557321e/nemo_rl/algorithms/grpo.py#L1435), [L1459](https://github.com/NVIDIA-NeMo/RL/blob/f0557321e/nemo_rl/algorithms/grpo.py#L1459), [L1490](https://github.com/NVIDIA-NeMo/RL/blob/f0557321e/nemo_rl/algorithms/grpo.py#L1490)).

The megatron branch does not call it ([`grpo.py:1275-1280`](https://github.com/NVIDIA-NeMo/RL/blob/f0557321e/nemo_rl/algorithms/grpo.py#L1275-L1280)):

```python
if backend == "megatron":
    # Initialize training first so checkpoint conversion completes before inference starts.
    policy, policy_time = init_policy()
    ...
    policy_generation, megatron_gen_time = init_megatron_generation(policy)
```

## Not fixed by the in-flight work

#3569 overlaps NeMo-Gym spinup with the megatron load, which is a real improvement, but it keeps the inner serialization — its `init_megatron_stack` calls `init_policy()` and then `init_megatron_generation()` inside a single thread. So after #3569:

## Is the stated reason still binding?

The comment says inference must wait so "checkpoint conversion completes". That is clearly true for **colocated**, where generation wraps the training policy. For **non-colocated** it is less obvious, and worth checking before assuming this cannot be parallelized:

- `weights_path` comes from the checkpointer at [`grpo.py:1121`](https://github.com/NVIDIA-NeMo/RL/blob/f0557321e/nemo_rl/algorithms/grpo.py#L1121) (`checkpointer.get_resume_paths(...)`), not from `init_policy()`.
- Non-colocated passes `policy=None` and `skip_weight_load=True` ([`grpo.py:1206-1213`](https://github.com/NVIDIA-NeMo/RL/blob/f0557321e/nemo_rl/algorithms/grpo.py#L1206-L1213)), so the generation workers do not load the converted checkpoint at all — refit delivers the weights.

If there is no remaining dependency in the non-colocated path, megatron could use `initialize_generation_with_policy` like the other backends. If there *is* one (e.g. something in the conversion writes state the generation workers read at build time), it would be worth stating it in that comment, since the current wording does not say what the dependency actually is.

Someone who knows the megatron checkpoint-conversion path should confirm which it is — that is the main question here, not the scheduling change itself.

## Context

Surfaced while reviewing #3731. Not caused by that PR, and not something it should fix.

tdene:
Closes #3777.

# What does this PR do ?

# Issues

# Usage

```python
```

**Pre checks**:

# Additional Information
* ...

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1292,8 +1292,8 @@ def setup(
         setup_timing_metrics.collective_init_time_s = time.perf_counter() - t0
         if not colocated_inference:
             # The skip-load inference engine gets its final weight buffers here.
-            # Its first prepare_for_generation also starts the HTTP server only
-            # after the refit, so CUDA graphs capture those persistent buffers.
+            # Its first prepare_for_generation also starts the HTTP server only after the refit,
+            # so CUDA graphs can capture those persistent buffers.
             t0 = time.perf_counter()
             weight_synchronizer.sync_weights()
             setup_timing_metrics.weight_sync_time_s = time.perf_counter() - t0
@@ -1412,10 +1412,9 @@ def setup(
                             time.perf_counter() - init_tasks_t0
                         )
                         setup_timing_metrics.parallel_init_enabled = 1.0
-                        # NeMo Gym probes the pre-published endpoint before its
-                        # future completes. A skip-load Megatron endpoint starts
-                        # only during this initial refit, so it must happen while
-                        # Gym is waiting rather than after its future resolves.
+                        # NeMo Gym probes the pre-published endpoint before its future completes.
+                        # A skip-load Megatron endpoint starts only during this initial refit,
+                        # so it must happen while Gym is waiting rather than after it resolves.
                         init_megatron_weight_synchronizer(policy, policy_generation)
                     nemo_gym_actor, nemo_gym_time = nemo_gym_future.result()
             finally:
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1277,7 +1277,10 @@ def setup(
         policy: ColocatablePolicyInterface,
         policy_generation: MegatronGeneration,
     ) -> None:
-        """Initialize Megatron refit and load non-colocated generation weights."""
+        """Initialize Megatron weight synchronizer.
+
+        For non-colocated inference, also performs the initial weight sync.
+        """
         t0 = time.perf_counter()
         weight_synchronizer = create_weight_synchronizer(
             policy=policy,
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1273,6 +1273,31 @@ def setup(
         )
         return mg, time.perf_counter() - t0
 
+    def init_megatron_weight_synchronizer(
+        policy: ColocatablePolicyInterface,
+        policy_generation: MegatronGeneration,
+    ) -> None:
+        """Initialize Megatron refit and load non-colocated generation weights."""
+        t0 = time.perf_counter()
+        weight_synchronizer = create_weight_synchronizer(
+            policy=policy,
+            generation=policy_generation,
+            generation_backend="megatron",
+            colocated=colocated_inference,
+            train_cluster=train_cluster,
+            inference_cluster=None if colocated_inference else inference_cluster,
+        )
+        policy_generation.weight_synchronizer = weight_synchronizer
+        weight_synchronizer.init_communicator()
+        setup_timing_metrics.collective_init_time_s = time.perf_counter() - t0
+        if not colocated_inference:
+            # The skip-load inference engine gets its final weight buffers here.
+            # Its first prepare_for_generation also starts the HTTP server only
+            # after the refit, so CUDA graphs capture those persistent buffers.
+            t0 = time.perf_counter()
+            weight_synchronizer.sync_weights()
+            setup_timing_metrics.weight_sync_time_s = time.perf_counter() - t0
+
     def initialize_generation_with_policy(
         init_generation_fn,
         colocated_inference: bool,
@@ -1382,16 +1407,21 @@ def setup(
                     nemo_gym_future = executor.submit(init_nemo_gym)
                     policy, policy_time = policy_future.result()
                     policy_generation, megatron_gen_time = generation_future.result()
+                    if not colocated_inference:
+                        setup_timing_metrics.parallel_wall_time_s = (
+                            time.perf_counter() - init_tasks_t0
+                        )
+                        setup_timing_metrics.parallel_init_enabled = 1.0
+                        # NeMo Gym probes the pre-published endpoint before its
+                        # future completes. A skip-load Megatron endpoint starts
+                        # only during this initial refit, so it must happen while
+                        # Gym is waiting rather than after its future resolves.
+                        init_megatron_weight_synchronizer(policy, policy_generation)
                     nemo_gym_actor, nemo_gym_time = nemo_gym_future.result()
             finally:
                 ray.kill(port_holder)
 
-            if not colocated_inference:
-                setup_timing_metrics.parallel_wall_time_s = (
-                    time.perf_counter() - init_tasks_t0
-                )
-                setup_timing_metrics.parallel_init_enabled = 1.0
-            else:
+            if colocated_inference:
                 setup_timing_metrics.parallel_init_enabled = 0.0
             setup_timing_metrics.policy_init_time_s = policy_time
             setup_timing_metrics.generation_init_time_s = (
@@ -1659,22 +1689,8 @@ def setup(
         )
 
     if backend == "megatron":
-        t0 = time.perf_counter()
-        policy_generation.weight_synchronizer = create_weight_synchronizer(
-            policy=policy,
-            generation=policy_generation,
-            generation_backend=backend,
-            colocated=colocated_inference,
-            train_cluster=train_cluster,
-            inference_cluster=None if colocated_inference else inference_cluster,
-        )
-        policy_generation.weight_synchronizer.init_communicator()
-        setup_timing_metrics.collective_init_time_s = time.perf_counter() - t0
-        if not colocated_inference:
-            # Load the model weights now.
-            t0 = time.perf_counter()
-            policy_generation.weight_synchronizer.sync_weights()
-            setup_timing_metrics.weight_sync_time_s = time.perf_counter() - t0
+        if policy_generation.weight_synchronizer is None:
+            init_megatron_weight_synchronizer(policy, policy_generation)
         if enable_nemo_gym:
             served_urls = policy_generation.dp_openai_server_base_urls
             if served_urls != [reserved_url]:
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1268,7 +1268,6 @@ def setup(
             cluster=None if colocated_inference else inference_cluster,
             policy=policy if colocated_inference else None,
             processor=processor,
-            weights_path=weights_path,
             skip_weight_load=not colocated_inference,
             reserved_http_server_port=reserved_http_server_port,
         )
@@ -1349,41 +1348,51 @@ def setup(
             setup_timing_metrics.generation_init_reserve_time_s = reserve_time
             print(f"  ✓ Reserved Megatron server URL: {reserved_url}", flush=True)
 
-            def init_megatron_stack():
-                """Init policy then generation; rank 0 holds the reserved port."""
-                p, policy_t = init_policy(
-                    reserved_http_server_port=reserved_http_server_port
-                    if colocated_inference
-                    else None
-                )
-                pg, gen_t = init_megatron_generation(
-                    p,
-                    reserved_http_server_port=None
-                    if colocated_inference
-                    else reserved_http_server_port,
-                )
-                return p, policy_t, pg, gen_t
-
             def init_nemo_gym():
                 """Spin up NeMo Gym servers against the reserved URL."""
                 return _spinup_nemo_gym([reserved_url], generation_config["model_name"])
 
-            init_tasks = {
-                "megatron": init_megatron_stack,
-                "nemo_gym": init_nemo_gym,
-            }
-            print(f"  ⚡ Init tasks: {', '.join(init_tasks.keys())}", flush=True)
+            # Exactly one task adopts the reserved port: the policy when colocated
+            # (generation wraps it), else the dedicated generation policy.
+            policy_port, generation_port = (
+                (reserved_http_server_port, None)
+                if colocated_inference
+                else (None, reserved_http_server_port)
+            )
+
+            def init_megatron_generation_task(policy_future):
+                """Colocated generation waits; non-colocated inits in parallel."""
+                if colocated_inference:
+                    p, _ = policy_future.result()
+                    return init_megatron_generation(p)
+                return init_megatron_generation(
+                    reserved_http_server_port=generation_port
+                )
+
+            print("  ⚡ Init tasks: policy, megatron_generation, nemo_gym", flush=True)
+            init_tasks_t0 = time.perf_counter()
             try:
-                with ThreadPoolExecutor(max_workers=len(init_tasks)) as executor:
-                    submitted = {k: executor.submit(fn) for k, fn in init_tasks.items()}
-                    results = {k: f.result() for k, f in submitted.items()}
+                with ThreadPoolExecutor(max_workers=3) as executor:
+                    policy_future = executor.submit(
+                        init_policy, reserved_http_server_port=policy_port
+                    )
+                    generation_future = executor.submit(
+                        init_megatron_generation_task, policy_future
+                    )
+                    nemo_gym_future = executor.submit(init_nemo_gym)
+                    policy, policy_time = policy_future.result()
+                    policy_generation, megatron_gen_time = generation_future.result()
+                    nemo_gym_actor, nemo_gym_time = nemo_gym_future.result()
             finally:
                 ray.kill(port_holder)
 
-            policy, policy_time, policy_generation, megatron_gen_time = results[
-                "megatron"
-            ]
-            nemo_gym_actor, nemo_gym_time = results["nemo_gym"]
+            if not colocated_inference:
+                setup_timing_metrics.parallel_wall_time_s = (
+                    time.perf_counter() - init_tasks_t0
+                )
+                setup_timing_metrics.parallel_init_enabled = 1.0
+            else:
+                setup_timing_metrics.parallel_init_enabled = 0.0
             setup_timing_metrics.policy_init_time_s = policy_time
             setup_timing_metrics.generation_init_time_s = (
                 reserve_time + megatron_gen_time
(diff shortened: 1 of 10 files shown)


nemo_rl/algorithms/grpo.py
```diff
@@ -1349,41 +1373,55 @@ def initialize_generation_with_policy(
...
+                    policy_generation, megatron_gen_time = generation_future.result()
+                    if not colocated_inference:
+                        setup_timing_metrics.parallel_wall_time_s = (
+                            time.perf_counter() - init_tasks_t0
+                        )
+                        setup_timing_metrics.parallel_init_enabled = 1.0
+                        # NeMo Gym probes the pre-published endpoint before its future completes.
+                        # A skip-load Megatron endpoint starts only during this initial refit,
+                        # so it must happen while Gym is waiting rather than after it resolves.
+                        init_megatron_weight_synchronizer(policy, policy_generation)
```
cspades, line 1421: [`nemo_rl/algorithms/grpo.py:1418`](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1418)

**2 action items.** **PR-introduced.**

**TL;DR — if `init_megatron_weight_synchronizer` raises before `sync_weights()`, `nemo_gym_future` blocks indefinitely and `ray.kill(port_holder)` can never fire.**

**Background.** The skip-load Megatron HTTP server only starts during `sync_weights()` inside `init_megatron_weight_synchronizer` — [the comment at line 1415 confirms this](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1415). `_spinup_nemo_gym` probes that endpoint and blocks until it responds.

**What breaks** (trigger: any exception during [`init_communicator()`](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1291)):
1. `generation_future.result()` returns normally.
2. `init_megatron_weight_synchronizer` raises — `sync_weights()` never called, HTTP server never starts.
3. Exception exits the `with ThreadPoolExecutor(max_workers=3)` block.
4. `executor.__exit__` calls `shutdown(wait=True)` — waits for all futures.
5. `nemo_gym_future` polls forever for the server → **hard hang**.
6. `finally: ray.kill(port_holder)` at line 1421 never fires — port reservation actor leaked.

### AI-1
Catch the exception from `init_megatron_weight_synchronizer` and call `executor.shutdown(wait=False, cancel_futures=True)` (Python 3.9+) before re-raising, so the gym future is cancelled and `ray.kill(port_holder)` can fire promptly.

### AI-2
Alternatively, add a bounded timeout to `_spinup_nemo_gym` (or its caller) so it fails fast with a clear error rather than blocking indefinitely when the Megatron endpoint never starts.

nemo_rl/models/megatron/community_import.py
```diff
@@ -62,6 +65,67 @@ def to_torch_dtype(dtype: str | torch.dtype) -> torch.dtype:
...
+                args=(staging_path,),
+                kwargs={"ignore_errors": True},
+                daemon=True,
+            ).start()
+            break
+        try:
+            os.rename(staging_path, pretrained_path)
+            break
+        except OSError:
+            if attempt == 1:
```
cspades, line 105: [`nemo_rl/models/megatron/community_import.py:105`](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/models/megatron/community_import.py#L105)

**1 action item.** **PR-introduced.**

In [`publish_megatron_conversion`](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/models/megatron/community_import.py#L75): on attempt 0, if `os.rename(staging_path, pretrained_path)` fails with `OSError`, the code displaces `pretrained_path` to a UUID-named sibling (line 113). On attempt 1, if the rename fails again (e.g., staging dir on a different filesystem — `EXDEV`), `raise` at line 105 exits before the cleanup block at line 119. The displaced directory is left on disk.

### AI-1
Wrap the loop in `try/finally` so the displaced copy is always scheduled for deletion:
```python
try:
    for attempt in range(2):
        ...
finally:
    if displaced is not None:
        threading.Thread(
            target=shutil.rmtree, args=(displaced,),
            kwargs={"ignore_errors": True}, daemon=True
        ).start()
```

nemo_rl/algorithms/grpo.py
```diff
@@ -1268,12 +1268,36 @@ def init_megatron_generation(
...
             skip_weight_load=not colocated_inference,
             reserved_http_server_port=reserved_http_server_port,
         )
         return mg, time.perf_counter() - t0
 
+    def init_megatron_weight_synchronizer(
+        policy: ColocatablePolicyInterface,
+        policy_generation: MegatronGeneration,
+    ) -> None:
+        """Initialize Megatron refit and load non-colocated generation weights."""
```
cspades: [`nemo_rl/algorithms/grpo.py:1280`](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1280)

**1 action item.** **PR-introduced.**

[`init_megatron_weight_synchronizer`](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1276) is called from two sites: [line 1418](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1418) (non-colocated only) and [line 1692](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1692) (both colocated and non-colocated via the `weight_synchronizer is None` guard). In the colocated path, `sync_weights()` is skipped — the current docstring's "load non-colocated generation weights" is inaccurate for that call.

### AI-1
```suggestion
        """Initialize Megatron weight synchronizer; for non-colocated inference, also performs the initial weight sync."""
```

nemo_rl/algorithms/grpo.py
```diff
@@ -1392,13 +1430,20 @@ def init_nemo_gym():
             setup_timing_metrics.nemo_gym_init_time_s = nemo_gym_time
 
         else:
-            # Initialize training first so checkpoint conversion completes before inference starts.
-            policy, policy_time = init_policy()
-            setup_timing_metrics.policy_init_time_s = policy_time
+            if not colocated_inference:
```
cspades, line 1436: [`nemo_rl/algorithms/grpo.py:1433`](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1433)

**1 action item.** **PR-introduced new path.**

This `else` branch (non-colocated megatron without NeMo Gym) calls `initialize_generation_with_policy(init_megatron_generation, …)` followed by the `weight_synchronizer is None` guard at [line 1692](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/nemo_rl/algorithms/grpo.py#L1692). The only new unit test covers the NeMo Gym + non-colocated path — this path has no test.

### AI-1
Consider adding a test with `backend="megatron"`, `should_use_nemo_gym=False`, `colocated.enabled=False` that verifies: (a) `MegatronGeneration` is called with `skip_weight_load=True`, and (b) `create_weight_synchronizer` + `sync_weights` are called via the post-init guard. `test_setup_refits_noncolocated_megatron_while_nemo_gym_waits` in [`tests/unit/algorithms/test_grpo.py`](https://github.com/NVIDIA-NeMo/RL/blob/93682b9cf82ff7b098c1ae422f1318e63c24d377/tests/unit/algorithms/test_grpo.py#L3149) is a close template — remove the NeMo Gym mocks and assert the weight sync fires post-`setup()`.

cspades (commented):
This PR addresses the performance gap documented in #3777 — megatron was the only non-colocated backend running policy and generation init serially. The fix is architecturally sound (3-way parallel ThreadPoolExecutor for the NeMo Gym path; reuse of `initialize_generation_with_policy` for the non-gym path) and the atomic checkpoint publication in `publish_megatron_conversion` is a welcome improvement.

Four findings below: one medium-severity hang on a failure path, and three low-severity suggestions.

**Note:** The PR description was not filled in — the motivation and approach are clear from the diff and issue #3777, but a one-line summary would help future readers.

_Reviewed by 5 specialized agents (rl-expert, bug-finder, test-agent, design-reviewer, comment-reviewer) with devil's advocate pass. Pre-commit: PASS._

_Generated by Claude Code_

nemo_rl/algorithms/grpo.py
tdene: This is incorrect: there is already a 600s timeout on `wait_for_model_endpoints`. Nothing leaks, things time out as designed.

nemo_rl/models/megatron/community_import.py
tdene: If we were working with different filesystems, we would fail from the beginning and no part of the function ever works, so there is no need to apply extra guarding.

The suggested fix does not help; it cleans up the leaked file, but the original file has already been renamed so it no longer exists - this is worse.

nemo_rl/algorithms/grpo.py
tdene: Agreed, fixed!

nemo_rl/algorithms/grpo.py
tdene: There is already such a test: `grpo-llama3.2-1b-instruct-2n8g-megatron_generation-noncolocated.sh`

nemo_rl/models/megatron/community_import.py
```diff
@@ -225,8 +292,35 @@ def import_model_from_hf_name(
...
+        torch.distributed.is_available() and torch.distributed.is_initialized()
+    )
+    staging_token = uuid.uuid4().hex[:8]
+    if dist_active:
+        token_box = [staging_token]
+        torch.distributed.broadcast_object_list(token_box, src=0)
+        staging_token = token_box[0]
+    staging_path = os.path.join(
+        output_parent, f".{os.path.basename(output_path)}.staging-{staging_token}"
+    )
```
terrykong: [`community_import.py:309-311`](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L309-L311)

**1 action item.** **PR-introduced.**

**TL;DR — an interrupted conversion now leaves a full-size checkpoint in `.<name>.staging-<uuid>` that nothing ever deletes, and every retry picks a fresh uuid, so N interrupted attempts leave N orphan copies in the shared cache.**

Before this PR the save went straight to `output_path`, so an interrupted run left one directory that the next attempt reused or overwrote — at most one partial copy. Now the save goes to a uuid-named hidden sibling and only becomes visible at the rename. If the job is preempted, OOMs, or loses a node between [the save at L314](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L314) and [the publish at L321](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L321), that directory stays forever.

Nothing else cleans it up: the only two `rmtree` calls in this file are both inside `publish_megatron_conversion`, and neither runs if we never reach it. A `try/finally` would not close this either — preemption and SIGKILL skip it.

It matters because `get_megatron_checkpoint_dir()` is normally a shared, quota'd filesystem, and the orphans are dot-prefixed, so nobody notices until it fills.

### AI-1
Have rank 0 reap stale staging siblings on entry, right after `staging_path` is computed. Sketch — **one way to do it, of several**:

```python
if not dist_active or torch.distributed.get_rank() == 0:
    # Nothing else deletes these; an interrupted conversion leaves one behind.
    # Age-gate so a concurrent producer's live staging dir is never removed.
    cutoff = time.time() - STALE_STAGING_AGE_S
    prefix = os.path.join(output_parent, f".{os.path.basename(output_path)}.staging-")
    for stale in glob.glob(prefix + "*"):
        if stale != staging_path and os.path.getmtime(stale) < cutoff:
            shutil.rmtree(stale, ignore_errors=True)
```

Not a committable `suggestion` block for two reasons: it needs new imports (`glob`, `time`) at the top of the file, which is outside this line range; and the age gate is needed for correctness — a bare glob-and-delete would destroy a concurrent producer's live staging directory — but the right cutoff is your call.

nemo_rl/models/megatron/community_import.py
```diff
@@ -62,6 +65,67 @@ def to_torch_dtype(dtype: str | torch.dtype) -> torch.dtype:
...
+                kwargs={"ignore_errors": True},
+                daemon=True,
+            ).start()
+            break
+        try:
+            os.rename(staging_path, pretrained_path)
+            break
+        except OSError:
+            if attempt == 1:
+                raise
```
terrykong, line 115: [`community_import.py:104-106`](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L104-L106)

**1 action item.** **PR-introduced. Low severity** — a rare startup race, and a rerun succeeds because the conversion is complete by then.

**TL;DR — this is the only exit in `publish_megatron_conversion` that gives up instead of conceding, so a job can die with a raw `OSError: Directory not empty` while a perfectly good conversion is sitting at `pretrained_path` that it could have used.**

Every other exit re-probes and concedes: [L87](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L87) at the top of each attempt, and [L107](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L107) after a failed rename. Only this one raises.

Breaking case — two jobs share `NEMO_RL_MEGATRON_CHECKPOINT_DIR` and convert the same model, over a stale partial conversion (a bare `iter_0000000/` with no `run_config.yaml`, left by an interrupted run — the case your own test parametrizes as `("partial", False, True)`):

1. Both probe at [L87](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L87): the occupant is partial, so not complete — both proceed.
2. Both [`os.rename` at L102](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L102) fail with `ENOTEMPTY` (renaming onto an existing non-empty directory).
3. Job A displaces `pretrained_path`; job B's displace hits `FileNotFoundError` and resets `displaced = None` ([L116-118](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L116-L118)).
4. Attempt 1: B probes at [L87](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L87) again, but the path is now *absent*, so still "not complete" — B falls through to the rename. **This is the step that succeeds when it should not.**
5. A's rename lands a complete conversion in the freed slot. B's rename hits `ENOTEMPTY`, `attempt == 1`, and raises here.

Symptom: job B dies during worker init with `OSError: [Errno 39] Directory not empty`. The same `raise` also strands the `.displaced-<uuid>` copy, because [the cleanup at L119-126](https://github.com/NVIDIA-NeMo/RL/blob/7463c341c9e93f198848ba1913d94bd7cfbb2606/nemo_rl/models/megatron/community_import.py#L119-L126) sits after the loop.

### AI-1
Re-probe before giving up:

```suggestion
        except OSError:
            if attempt == 1:
                if not overwrite and megatron_conversion_is_complete(pretrained_path):
                    # A peer published into the slot we freed; use its artifact.
                    threading.Thread(
                        target=shutil.rmtree,
                        args=(staging_path,),
                        kwargs={"ignore_errors": True},
                        daemon=True,
                    ).start()
                    break
                raise
```

terrykong (commented):
Thanks for this — the change hangs together well, and the part that makes it *safe* is easy to miss, so worth calling out: the overlap only works because two other things landed with it. `megatron_policy_worker.py` now skips `handle_model_import` entirely for refit-fed policies, and `publish_megatron_conversion` makes the HF->Megatron publish atomic. Without both, the two worker groups would race on the same cache directory — which is exactly what the old "initialize training first" ordering existed to prevent. Switching the completion marker to `iter_0000000/run_config.yaml` also fixes a real pre-existing bug where an interrupted conversion counted as complete.

One suggestion below, no blockers — a narrow failure in the new atomic-publish path.

Two notes on scope:

- PR #3873 (@wdykas) is merged into this branch rather than into `main`, so `init_megatron_weight_synchronizer`, the `grpo_megatron_mxfp8_refit_gb200.sh` assertion changes and the new `test_setup_refits_noncolocated_megatron_while_nemo_gym_waits` all land on `main` through this PR. Flagging it so the right person sees any feedback on those hunks.
- One thing we checked and are explicitly **not** asking you to change: that script tightens `token_mult_prob_error` from 2.0 to 1.5, but it is commented out at `L1_Functional_Tests_GB200_MXFP8.sh:40` on this branch and on `main`, so the new bound has never run. Re-enabling needs a 4-GPU runner this PR does not provide, so there is nothing to do here — just be aware the bound is unverified for whoever re-enables it. The metric swap in the same hunk (`generation_init_load_time_s` -> `weight_sync_time_s`) is a genuine fix, since the old field is never set on the non-colocated non-Gym path.

We also looked hard at whether the new HF-derived provider for `skip_weight_load` policies could disagree with the training provider built from the conversion cache, and concluded it cannot — both routes end up building from the same `AutoConfig.from_pretrained(...)` result. No action needed; noting it so the question does not get re-opened.

Reviewed by 7 agents (RL expert, Megatron-Bridge expert, bug finder, test agent, design reviewer, comment reviewer, devil's advocate). `pre-commit` could not run in this environment, so no lint results are reported.

_Generated by Claude Code_

tdene:
> * One thing we checked and are explicitly **not** asking you to change: that script tightens `token_mult_prob_error` from 2.0 to 1.5, but it is commented out at `L1_Functional_Tests_GB200_MXFP8.sh:40` on this branch and on `main`, so the new bound has never run. Re-enabling needs a 4-GPU runner this PR does not provide, so there is nothing to do here — just be aware the bound is unverified for whoever re-enables it. The metric swap in the same hunk (`generation_init_load_time_s` -> `weight_sync_time_s`) is a genuine fix, since the old field is never set on the non-colocated non-Gym path.

As with several other test-related issues, I'm choosing to defer this to #3570.

Once everything settles down, I vow that #3570 will fix all these assorted test issues.

nemo_rl/models/megatron/community_import.py
tdene: Suggestion applied.

And I guess GitHub auto-resolves threads where you apply suggestions? I didn't expect that.

terrykong (approved):

PR #2872, opened 2026-06-18
fix(megatron): honor policy.logprob_chunk_size in the training loss path

kaloyan-inherent:
# What does this PR do ?

This PR closes https://github.com/NVIDIA-NeMo/RL/issues/2871 by passing `chunk_size` for the logprobs calculation.

# Issues

Closes https://github.com/NVIDIA-NeMo/RL/issues/2871

PR #2823, opened 2026-06-15
feat: video + audio understanding GRPO training recipe 

yuekaizhang:
The PR supports audio visual GRPO training using qwen2.5-omni-7B.

Based on [[HumanOmniV2](https://github.com/HumanMLLM/HumanOmniV2)](https://arxiv.org/abs/2506.21277) training set, the recipe could improve qwen-omni's video reasoning perfomance a lot.

#### Curve
image

#### Results on DailyOmni testset:

Daily-Omni accuracy (1197 questions, greedy decoding) for the base Qwen2.5-Omni-7B versus the GRPO-trained checkpoint:

| Question type | Base | After GRPO |
| --- | --- | --- |
| **Overall** | **0.498** | **0.590** |
| AV Event Alignment | 0.353 | 0.450 |
| Comparative | 0.618 | 0.725 |
| Context understanding | 0.446 | 0.534 |
| Event Sequence | 0.395 | 0.490 |
| Inference | 0.714 | 0.760 |
| Reasoning | 0.651 | 0.766 |

PR #3869, opened 2026-08-27
feat(rollout): integrate MInf ledger capture with checkpointing

lauradang:
## Summary

- unify vLLM and Megatron token capture around Gym's `CaptureAdmission`, token-free lineage ledger, and the shared TQ staging contract
- resolve staged parent tokens before generation so both backends continue agent rollouts from the exact captured token history
- carry capture state through Megatron generation and stage the exact `OffloadedRequestPayload` through `RolloutTokenCapture` and `TQTokenSink`
- wait for durable TQ staging before returning commit coordinates; Gym writes a token-free `CallRecord` only after that acknowledgement
- assemble a token-free `RolloutReceipt` at rollout end, then fetch, verify, and linearize staged records into the canonical GRPO sample during finalization
- support both colocated and non-colocated SingleController Megatron-Inference deployments while retaining the vLLM capture path

## Token capture flow

The two serving backends use different prompt-preparation and generation hooks, but converge at the same durability boundary. A model call becomes eligible for lineage resolution only after its canonical token delta is staged in TQ. Capture failures write a failure row instead of a committed call record, causing finalization to reject or mask the rollout rather than train on incomplete lineage.

Token capture custody — vLLM and Megatron function map

The source diagram is tracked in `docs/assets/token-capture-ledger-queue-data-flow.dot`; the full invariants, terminal-selection rules, cleanup behavior, and fail-closed semantics are documented in `docs/design-docs/token-capture-ledger.md`.

## Dependencies

- Base stack: `amahishi/partial-rollout-telemetry-v3`
- Gym capture contract and unified Megatron staging: [`NVIDIA-NeMo/Gym@c3d7cf4`](https://github.com/NVIDIA-NeMo/Gym/commit/c3d7cf4ff484d0f4f07898aedeaba05eb4821438)
- No separate Megatron metadata-ledger dependency: Megatron and vLLM now stage canonical records through the same TQ path

## Testing

- Ruff check and format verification pass for the newly touched RL files
- Gym scoped pre-commit passes
- Gym token-capture and focused model-server suites: 268 passed

## Local test limitation

The RL lockfile supports Linux x86_64/aarch64 only, so the focused RL pytest files cannot run natively on this macOS host. An ephemeral dependency run progressed through PyTorch, Transformers, and pyzmq before reaching the NVIDIA-only `pynvml` runtime dependency; Linux CI provides the authoritative RL test coverage.


nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -424,6 +431,88 @@ def report_dp_openai_server_base_url(self) -> Optional[str]:
...
+            for name in (
+                "local_metadata_ledger_enabled",
+                "local_metadata_ledger_offload_enabled",
+                "fetch_from_metadata_ledger",
+            )
+            if not hasattr(engine, name)
+        ]
+        if missing:
+            raise RuntimeError(
+                "Megatron token capture requires the MInf ledger-capture API; "
```
lauradang, line 627: **[DEP/blocker]** The pinned Megatron-Bridge (`0c565c9a0` → Megatron-LM `d12f6c8c9`) has no `local_metadata_ledger_*` / `fetch_from_metadata_ledger` API (verified by git-grep at the pinned commits; only `set_generation_epoch` exists there). So with this PR as-is, enabling megatron token capture always hits this raise at setup. Failing loudly here is right, and the PR body declares the `tdene:Megatron-LM:tde/ledger_capture` dependency — but the PR doesn't bump the Megatron-Bridge pin, so this needs the MInf ledger to land and the submodule bump to be part of (or precede) the merge. Flagging so the merge order is explicit.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -256,6 +337,199 @@ def finish_generation(self, *args: Any, **kwargs: Any) -> bool:
...
+                    parent_chain_hash=parent_chain_hash,
+                )
+                weight_version = adapter.extract_weight_version(ledger_payload)
+                prepared_calls.append(
+                    (record, ledger_payload, admission, weight_version)
+                )
+
+            for record, ledger_payload, admission, weight_version in prepared_calls:
+                call = capture.begin_call(admission, weight_version=weight_version)
+                coords = capture.complete_call_from_response(call, ledger_payload)
```
lauradang, line 502: **[Suggestion — custody gap]** The flush joins HTTP lineage to the MInf ledger purely by `ledger_request_uid` and cross-checks lengths only. The `chain_hash` that Gym makes *required* on every `PendingCallRecord` precisely to prove token custody is never compared to the ledger-recomputed value, and the prefix check inside Gym's `complete_call` is tautological here (the admission prefix is sliced from the same ledger payload it validates). An equal-length content divergence (uid collision, ledger corruption, engine bug) would be durably staged and trained on silently — nothing downstream catches it either, since the finalizer compares TQ snapshots to a manifest derived from the same ledger payload.

One-line fix, and it is apples-to-apples (both sides are `compute_chain_hash` over the same parent anchor and the same intended delta):

```python
coords = capture.complete_call_from_response(call, ledger_payload)
if coords.disposition != "staged":
    ...
if coords.chain_hash != record.chain_hash:
    raise RuntimeError(
        f"MInf request {record.ledger_request_uid!r} token content diverges from the HTTP lineage"
    )
```

(Hold off on comparing `cumulative_hash` — serve-side currently hashes delta-only for children due to the Gym-side `required_prefix_token_ids=[]` issue noted in the review summary, so it wouldn't match until that's fixed.)

nemo_rl/environments/nemo_gym.py
```diff
@@ -555,13 +555,19 @@ def _spinup(self) -> None:
         # same rollout ids — the retry's calls would resolve against the
         # first attempt's ledger rows — so the NaN retry must be exactly 1.
         token_capture = self.cfg.get("token_capture") or None
+        generation_backend = (
+            token_capture.pop("generation_backend", "vllm")
```
lauradang: **[Suggestion — low severity]** `generation_backend` is smuggled as an undeclared key into the dumped `TokenCaptureConfig` dict in `_spinup_gym`, then popped back out here with a hidden call-site default. Besides the side-channel shape (the dict no longer matches its documented schema, `NemoGymConfig.token_capture` = "the dumped TokenCaptureConfig"), the `"vllm"` fallback silently picks the wrong staging backend if any caller dumps the config without the key. `TokenCaptureConfig` already has derived fields filled at setup time (`control_auth_token`, `capture_dir`) — declaring `generation_backend: Optional[Literal["vllm", "megatron"]] = None` there and setting it alongside them would make this typed and loud.

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -1194,8 +1394,8 @@ def _build_generation_then_trainer(
                 # (nemo_rl/utils/venvs.py), so a venv prebuilt before token
                 # capture predates the nemo_gym extra and is reused as-is.
                 raise RuntimeError(
-                    "token_capture.enabled requires nemo_gym inside the vLLM "
-                    "worker venv, but the cached worker venv predates it. "
+                    "token_capture.enabled requires nemo_gym in the capture "
```
lauradang: **[Nit — misleading remediation for megatron]** For the vllm backend the `nemo_gym` import indeed happens inside the worker venv, so this remediation is right. But for megatron, `MegatronGeneration.setup_token_capture` imports `nemo_gym` in the *driver* process — rebuilding/deleting the vLLM worker venv can't fix that ImportError. Consider branching the message on `generation_config["backend"]` (megatron → "install the nemo_gym extra in the driver environment").

lauradang (commented):
Review of the lauradang commits only (`08faaee37`, `0c5e033ec`, `0ac170793`). Verified against the pinned Gym submodule (`5cd5c0d`), the pinned Megatron-Bridge → Megatron-LM (`0c565c9a0` → `d12f6c8c9`), and the `tdene:Megatron-LM:tde/ledger_capture` branch. 4 inline comments below, plus one Gym-side issue on the submodule bump that I can't anchor inline:

Also verified OK, for the record: the exactly-one-rank expectation in `flush_token_capture` matches the MInf branch (ledger is populated only on the MP-coordinator rank per DP engine; fan-out-and-union is the documented pattern), and `set_generation_epoch` from global rank 0 propagates to all engines via the DP coordinator.

nemo_rl/models/generation/megatron/megatron_worker.py
lauradang: Pending merge: https://github.com/tdene/Megatron-LM/tree/tde/ledger_capture

tdene:
For other reviewers, note that NVIDIA/Megatron-LM#7015 is the "MInf metadata ledger" PR linked to this.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -256,6 +337,209 @@ def finish_generation(self, *args: Any, **kwargs: Any) -> bool:
...
+        ray.get(futures)
+
+    def set_rollout_weight_version(self, version: int) -> None:
+        """Rotate the policy epoch stamped by MInf on subsequent requests."""
+        futures = self._policy.worker_group.run_all_workers_single_data(
+            "set_rollout_weight_version", version=version
+        )
+        ray.get(futures)
+
+    def flush_token_capture(self, receipt: dict[str, Any]) -> dict[str, Any]:
```
tdene, line 393: We enter this from `_flush_and_seal` after checking only `pending_manifest`, but we should also be checking `capture_poisoned`. Right now, poisoned rollouts that are still carrying `pending_manifest` will get flushed and sealed, even though it will just be thrown away in the end.

I think the correct fix is to early exit out of `flush_token_capture` when `receipt["capture_poisoned"]` or `receipt.get("capture_poisoned")`.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -256,6 +337,209 @@ def finish_generation(self, *args: Any, **kwargs: Any) -> bool:
...
+            futures = self._policy.worker_group.run_all_workers_single_data(
+                "fetch_token_capture_records", request_uids=request_uids
+            )
+            worker_records = ray.get(futures)
+            records_by_uid: dict[str, dict[str, Any]] = {}
+            for uid in request_uids:
+                matches = [records[uid] for records in worker_records if uid in records]
+                if len(matches) != 1:
+                    raise RuntimeError(
+                        f"MInf request {uid!r} was present on {len(matches)} engine ranks; expected 1"
```
tdene, line 434: I don't think this should be handled right now, but I just want to bring attention to it.

Today, rollout checkpointing only happens on training boundaries. We lose everything within one training step.

But MLM and MRL have the ability to checkpoint continuously. Once that ability comes into play, we will have situations where `matches==0` is valid, because we're loading checkpointed requests.

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -424,6 +431,88 @@ def report_dp_openai_server_base_url(self) -> Optional[str]:
...
+            raise RuntimeError("Megatron token capture is not initialized")
+        setter = getattr(self.inference_client, "set_generation_epoch", None)
+        if not callable(setter):
+            raise RuntimeError(
+                "Megatron token capture requires InferenceClient.set_generation_epoch"
+            )
+        setter(version)
+
+    @staticmethod
+    def _serialize_token_capture_record(record) -> dict:
```
tdene, line 653: This serializes both `routing_indices` and `prompt_log_probs`, but neither is consumed. In particular, `routing_indices` is absolutely huge (order of magnitude of 1KB per token).

nemo_rl/experience/rollout_manager.py
```diff
@@ -1775,24 +1776,36 @@ async def _record_streamed_completion(
...
             else:
                 async with barrier.mutation("sibling_seals"):
-                    self._recovery_ledger.mark_sibling_sealed(
-                        group_id,
-                        generation_index=generation_index,
-                        gate_rollout_id=gate_rollout_id,
-                        receipt=receipt,
-                        reward=completion.reward,
-                    )
+                    await _flush_and_seal()
```
tdene: This technically holds the barrier for longer than it strictly needs to. I'm not sure it matters though. `verify` is probably going to be the slowest thing that can be taken out of the barrier, but I really don't know.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -256,6 +337,209 @@ def finish_generation(self, *args: Any, **kwargs: Any) -> bool:
...
+        from nemo_gym.token_id_capture.staging.capture import RolloutTokenCapture
+
+        from nemo_rl.data_plane import build_data_plane_client
+        from nemo_rl.data_plane.tq_token_sink import TQTokenSink
+
+        dp_client = build_data_plane_client(dp_cfg, bootstrap=False)
+        sink = TQTokenSink(dp_client, staging_partition=staging_partition)
+        self._token_capture = RolloutTokenCapture(
+            sink=sink,
+            weight_version_fn=lambda: 0,
```
tdene, line 378: This kind of "default to 0" has caused us trouble before in Megatron RL. I don't think we should be nice. I think we should raise if we don't have an actual `weight_version`. So, hah, maybe `lambda: raise Error` would be more appropriate.

tdene (commented):
A functional test is also important to add.

3rdparty/Gym-workspace/Gym
```diff
@@ -1 +1 @@
-Subproject commit fd5e84d6b1c485c80e7ae61553bbd485611c03b4
+Subproject commit c3d7cf4ff484d0f4f07898aedeaba05eb4821438
```
lauradang, line 1: [`3rdparty/Gym-workspace/Gym:1`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/3rdparty/Gym-workspace/Gym#L1)

**1 action item — merge-blocking; please fix in this PR.** PR-introduced.

**TL;DR — the Gym pin `c3d7cf4` is a fork branch with no shared history with the base branch's pin; merging it reverts ~215 upstream Gym commits (730 deleted files).**

Evidence (`gh api repos/NVIDIA-NeMo/Gym/compare/...`):
- `main...c3d7cf4` → `diverged`, ahead 65, **behind 215** (merge base `3a5afd9`, 2026-08-19)
- `fd5e84d...c3d7cf4` (base pin → PR pin) → `diverged`, ahead 65, behind 183; `main...fd5e84d` → `behind 32`, so the base pin is on main.
- `git diff --diff-filter=D fd5e84d c3d7cf4 | wc -l` → **730** deleted files, `+7365/−134155`: `nemo_gym/telemetry`, `nemo_gym/sandbox/providers/{e2b,local}`, `nemo_gym/health`, `nemo_gym/exporters`, a dozen `resources_servers/*`, `tests/unit_tests`, …
- `c3d7cf4` is only on `lauradang/Gym:laurad/megatron-adapter` (Gym PR #2823, which targets `pthombre/tq-tokidcap-capture` — already squash-merged to main as #2872 = the base pin).

This is what the "Submodule Fast-Forward Check: Commits have DIVERGED" failure on this PR is reporting. RL is also coupled to fork-only surface (`external_staging_backend` — main's `TokenIdCaptureSettings` is `extra="forbid"`; `begin_call(…, weight_version=)`; the `megatron_worker` handler in [`external_capture.py`](https://github.com/NVIDIA-NeMo/Gym/blob/c3d7cf4ff484d0f4f07898aedeaba05eb4821438/nemo_gym/token_id_capture/external_capture.py#L265); `CallRecord.logical_request_id`), so re-pinning to main alone would break the RL side.

### AI-1
Re-port Gym #2823 onto Gym `main`, merge it, then pin RL to a SHA where `gh api repos/NVIDIA-NeMo/Gym/compare/main...<sha>` reports `identical` or `behind`. Consider making the fast-forward check a required status.

nemo_rl/data_plane/tq_token_sink.py
```diff
@@ -331,7 +544,7 @@ class TQTokenSource:
     """
 
     def __init__(self, dp_client: Any, *, staging_partition: str) -> None:
-        self._dp_client = dp_client
+        self._store = TQStagingStore(dp_client, staging_partition=staging_partition)
```
lauradang, line 548: [`nemo_rl/data_plane/tq_token_sink.py:547`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/data_plane/tq_token_sink.py#L547)

**1 action item — please fix in this PR.** PR-introduced.

**TL;DR — `TQTokenSource.__init__` no longer sets `_dp_client`, but `fetch_for_finalization` still reads it, so every finalizer fetch fails and every rollout (vLLM included) is rejected as `missing_staging_row`.**

1. `__init__` now sets only `_store`/`_staging_partition` (this line).
2. [`fetch_for_finalization`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/data_plane/tq_token_sink.py#L601-L628) (unchanged by the diff) still calls `_call_dp(self._dp_client, "get_samples", …)` at L607/L615/L623 → `AttributeError`.
3. The broad `except Exception` at [L628](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/data_plane/tq_token_sink.py#L628) re-wraps it as `KeyError("staged rows … could not be fetched … 'TQTokenSource' object has no attribute '_dp_client'")`.
4. [`rollout_reassembler.py:210`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/experience/rollout_reassembler.py#L210) maps that to `missing_staging_row` → the rollout becomes a masked placeholder, silently, for every rollout.

Existing `test_tq_sink_source_passes_conformance` / `test_fetch_for_finalization_*` and the new `test_megatron_stager_writes_canonical_row_and_returns_coords` (`:256`, calls `TQTokenSource.fetch`) cannot pass at this head (consistent with RL pytest not being runnable on the macOS host).

### AI-1
Finish the `TQStagingStore` migration: replace the three `_call_dp(self._dp_client, "get_samples", sample_ids=…, partition_id=…, select_fields=X)` calls with `self._store.get(staging_keys, select_fields=X)`. Please run `tests/unit/data_plane/test_tq_token_sink.py` in the nemo_gym lane before merge.

nemo_rl/environments/nemo_gym.py
```diff
@@ -475,6 +488,13 @@ def _spinup(self) -> None:
         self._control_headers: Dict[str, str] = {}
         self._control_timeout_s = 60.0
         if self._token_capture_enabled:
+            assert token_capture is not None
+            if self.rollout_max_attempts_to_avoid_lp_nan != 1:
```
lauradang, line 496: [`nemo_rl/environments/nemo_gym.py:492`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/environments/nemo_gym.py#L492)

**1 action item — please fix in this PR.** PR-introduced.

`self.rollout_max_attempts_to_avoid_lp_nan` is never assigned anywhere at HEAD — `git grep` finds only this line and the message below it; it is not in `NemoGymConfig`, `__init__`, or Gym at the pin. Every `token_capture.enabled=true` run (both functional recovery scripts) raises `AttributeError` here at Gym spin-up.

### AI-1
Delete the guard (the NaN-retry knob it references no longer exists at HEAD), or if a retry knob exists under another name, read it from `self.cfg` and validate it in `setup_single_controller` so it fails at setup rather than inside the actor. A unit test that drives this branch of `_spinup` would have caught it (the existing tests use `object.__new__` and never reach `_spinup`).

nemo_rl/environments/nemo_gym.py
```diff
@@ -1239,16 +1261,19 @@ def validate_reward_components_match_scalar(nemo_gym_results: List[dict]) -> Non
...
-    backend = generation_config.get("backend")
-    if backend == "vllm":
-        # Enable the http server. Requires both async engine and the expose_http_server flag
+    # Enable the backend's OpenAI-compatible server.
+    if generation_config["backend"] == "vllm":
         generation_config["vllm_cfg"]["async_engine"] = True
         generation_config["vllm_cfg"]["expose_http_server"] = True
-    elif backend == "megatron":
-        # Enable the http server for Gym dispatch over the Megatron generation backend.
+    elif generation_config["backend"] == "megatron":
```
lauradang, line 1272: [`nemo_rl/environments/nemo_gym.py:1268`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/environments/nemo_gym.py#L1268)

**No action needed** — for the record: at `1de0d2895` this branch set `mcore_generation_config["async_engine"] = True`, which [`should_use_async_rollouts`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/models/generation/interfaces.py#L268-L273) asserts must be absent, so every Megatron+Gym SC run died at setup. Fixed in a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e with a regression test.

nemo_rl/models/generation/megatron/megatron_worker.py
```diff
@@ -758,6 +764,69 @@ def report_dp_openai_server_base_url(self) -> Optional[str]:
...
+    ) -> bool:
+        """Install canonical TQ capture on each MInf model-parallel leader."""
+        engine = self.dynamic_inference_engine
+        if engine is None:
+            raise RuntimeError(
+                "Megatron token capture requires an initialized inference engine"
+            )
+        missing = [
+            name
+            for name in ("payload_stager", "prompt_preparer")
```
lauradang, line 778: [`nemo_rl/models/generation/megatron/megatron_worker.py:778`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/models/generation/megatron/megatron_worker.py#L778)

**1 action item, 1 follow-up.** PR-introduced.

**TL;DR — the pinned Megatron-LM has none of the engine hooks this path needs; they exist only in the still-open NVIDIA/Megatron-LM#7015, so Megatron capture is non-functional as pinned and the PR body understates the dependency.**

Megatron-Bridge `5ed97996` pins Megatron-LM `1e7598cb`. At that SHA [`dynamic_engine.py`](https://github.com/NVIDIA/Megatron-LM/blob/1e7598cbfae888cdd3d741a351aae588d56f66c0/megatron/core/inference/engines/dynamic_engine.py#L968) has `is_mp_coordinator` and [`inference_client.py`](https://github.com/NVIDIA/Megatron-LM/blob/1e7598cbfae888cdd3d741a351aae588d56f66c0/megatron/core/inference/inference_client.py#L599) has `set_generation_epoch`, but there is no `payload_stager`, `prompt_preparer`, `request_metadata`, or `OffloadedRequestPayload`, and `chat_completions.py` never forwards `ng_capture` / `ng_prompt_suffix_token_ids` or emits `ng_commit_coords`. All of those come from [NVIDIA/Megatron-LM#7015](https://github.com/NVIDIA/Megatron-LM/pull/7015) (OPEN). Result today: `RuntimeError("… missing payload_stager, prompt_preparer")` right here. This is the same merge-order blocker as the earlier `local_metadata_ledger_*` thread, under new names.

### AI-1
State the dependency in the PR description (replace "No separate Megatron metadata-ledger dependency") and in the design doc's MInf section, and name #7015 in this `RuntimeError` so the remediation is one click away.

### Follow-up
Gate merge on #7015 landing and a Megatron-Bridge bump whose `3rdparty/Megatron-LM` includes it, then re-lock RL.

tests/functional/grpo_async_gym_single_controller_sibling_recovery.sh
```diff
@@ -23,7 +23,7 @@ COMMON_OVERRIDES=(
     checkpointing.checkpoint_dir="$CHECKPOINT_DIR"
     checkpointing.metric_name=null
     checkpointing.save_period=1
-    +checkpointing.save_data_plane=true
+    checkpointing.save_data_plane=true
```
lauradang: [`tests/functional/grpo_async_gym_single_controller_sibling_recovery.sh:26`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/tests/functional/grpo_async_gym_single_controller_sibling_recovery.sh#L26)

**1 action item — please fix in this PR.** PR-introduced.

Dropping `+` requires the key to already exist in the loaded config. This script runs `grpo_async_gym_single_controller.sh` → `examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml` → `defaults: grpo_workplace_assistant_nemotron_nano_v2_9b.yaml`; neither `checkpointing:` block declares `save_data_plane`, and [`parse_hydra_overrides`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/utils/config.py#L186) sets struct mode before applying overrides. Reproduced with the locked hydra-core 1.3.2: plain form → `ConfigCompositionException: Could not override 'checkpointing.save_data_plane'`; `+`/`++` OK. (The other scripts using the plain form load `grpo_math_1B_megatron_single_controller.yaml`, which declares the key; the PR's own mirror at [`test_setup.py:582`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/tests/unit/single_controller/test_setup.py#L582) keeps `+`.)

### AI-1
```suggestion
    ++checkpointing.save_data_plane=true
```
(Alternative: add `save_data_plane: false` to the workplace exemplar's `checkpointing:` block and keep the plain override.)

tests/functional/grpo_async_gym_single_controller_streaming_recovery.sh
```diff
@@ -52,7 +52,7 @@ COMMON_OVERRIDES=(
     checkpointing.checkpoint_dir="$CHECKPOINT_DIR"
     checkpointing.save_period=1
     checkpointing.metric_name=null
-    +checkpointing.save_data_plane=true
+    checkpointing.save_data_plane=true
```
lauradang, line 55: [`tests/functional/grpo_async_gym_single_controller_streaming_recovery.sh:55`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/tests/functional/grpo_async_gym_single_controller_streaming_recovery.sh#L55)

**1 action item — please fix in this PR.** PR-introduced.

Dropping `+` requires the key to already exist in the loaded config. This script runs `grpo_async_gym_single_controller.sh` → `examples/nemo_gym/grpo_qwen3_30ba3b_instruct.yaml` → `defaults: grpo_workplace_assistant_nemotron_nano_v2_9b.yaml`; neither `checkpointing:` block declares `save_data_plane`, and [`parse_hydra_overrides`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/utils/config.py#L186) sets struct mode before applying overrides. Reproduced with the locked hydra-core 1.3.2: plain form → `ConfigCompositionException: Could not override 'checkpointing.save_data_plane'`; `+`/`++` OK. (The other scripts using the plain form load `grpo_math_1B_megatron_single_controller.yaml`, which declares the key; the PR's own mirror at [`test_setup.py:582`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/tests/unit/single_controller/test_setup.py#L582) keeps `+`.)

### AI-1
```suggestion
    ++checkpointing.save_data_plane=true
```
(Alternative: add `save_data_plane: false` to the workplace exemplar's `checkpointing:` block and keep the plain override.)

tests/functional/grpo_async_gym_single_controller_streaming_recovery.sh
```diff
@@ -52,7 +52,7 @@ COMMON_OVERRIDES=(
     checkpointing.checkpoint_dir="$CHECKPOINT_DIR"
```
lauradang, line 52: [`tests/functional/grpo_async_gym_single_controller_streaming_recovery.sh:52`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/tests/functional/grpo_async_gym_single_controller_streaming_recovery.sh#L52)

**1 follow-up.** Pre-existing gap this PR doesn't close (tdene asked for a functional test).

Both touched functional scripts wrap `grpo_async_gym_single_controller.sh`, which pins vLLM. No CI scenario runs `backend=megatron` with `token_capture.enabled`, so `TQMegatronPromptPreparer`, `TQMegatronTokenStager`, the worker hooks and the `megatron_worker` Gym backend have no end-to-end exercise.

### Follow-up
Once the Megatron-LM pin allows it, add to `L1_Functional_Tests_SingleController.sh` next to the vLLM smoke:
```bash
run_test uv run --no-sync bash ./tests/functional/grpo_megatron_generation_gym_single_controller.sh ++token_capture.enabled=true
```
(`grpo_megatron_generation_gym_single_controller.sh` already forwards `$@` and its yaml has `expose_http_server: true`), and port the `EXTRA_CHECKS` finalize assertion from `grpo_async_gym_single_controller.sh:112-116` so the run proves capture ran.

nemo_rl/experience/rollout_reassembler.py
```diff
@@ -175,9 +178,11 @@ def rejected(reason: str, staging_keys: list[str]) -> FinalizedRollout:
             return rejected("missing_receipt", [])
         try:
             parsed = RolloutReceipt.model_validate(receipt)
-        except ValueError as error:
+            staging_keys = [record.staging_key for record in parsed.manifest]
+        except KeyError as error:
+            return rejected(f"missing_staging_row:{error}", [])
+        except (TypeError, ValueError) as error:
             return rejected(f"invalid_receipt:{error}", [])
```
lauradang, line 185: [`nemo_rl/experience/rollout_reassembler.py:180`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/experience/rollout_reassembler.py#L180-L185)

The new `except KeyError` is unreachable: `RolloutReceipt.model_validate` raises pydantic `ValidationError` (a `ValueError` subclass), Gym's validators raise only `ValueError` ([`records.py:318-325`](https://github.com/NVIDIA-NeMo/Gym/blob/c3d7cf4ff484d0f4f07898aedeaba05eb4821438/nemo_gym/token_id_capture/staging/records.py#L318-L325)), and `record.staging_key` is a required attribute ([`records.py:243`](https://github.com/NVIDIA-NeMo/Gym/blob/c3d7cf4ff484d0f4f07898aedeaba05eb4821438/nemo_gym/token_id_capture/staging/records.py#L243)). If it ever fired it would reuse `missing_staging_row`, the reason reserved for real TQ fetch misses at L210.

### AI-1
```suggestion
        try:
            parsed = RolloutReceipt.model_validate(receipt)
        except (TypeError, ValueError) as error:
            return rejected(f"invalid_receipt:{error}", [])
        staging_keys = [record.staging_key for record in parsed.manifest]
```

docs/design-docs/token-capture-ledger.md
```diff
@@ -0,0 +1,156 @@
...
+the token-free record of rollout capture state. There is no separate gate
+state machine; serving workers coordinate only through the ledger, and NeMo RL
+(the rollout owner) assembles the `RolloutReceipt` itself at rollout end.
+
+The external staging contract (`StagingSink` / `StagingSource`), the vLLM
+worker capture path, and the `verify_and_linearize()` trust boundary are
+unchanged from the worker-custody design. The recipe exercising this path end
+to end is [Nano SWE with Token Capture](../guides/nano-swe-token-capture.md);
+the verification trust boundary is described in
+[Rollout Verification Boundary](rollout-verification-boundary.md).
```
lauradang, line 15: [`docs/design-docs/token-capture-ledger.md:13`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/docs/design-docs/token-capture-ledger.md#L13-L15)

**3 action items.** PR-introduced.

### AI-1
Neither linked page exists at HEAD (`docs/guides/nano-swe-token-capture.md`, `docs/design-docs/rollout-verification-boundary.md`) — drop the links or add the docs.

### AI-2
Add this page to the `design-docs` toctree in [`docs/index.md`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/docs/index.md#L386-L389) (nemo-rl-docs convention).

### AI-3
`docs/assets/token-capture-ledger-queue-data-flow.{dot,png}` are referenced only from the PR body — embed the PNG here or drop the assets.

docs/design-docs/token-capture-ledger.md
```diff
@@ -0,0 +1,156 @@
...
+deleted.
+
+## The ledger
+
+`FileLineageStore` writes one locked, fsynced JSONL row per committed call.
+In external-staging mode (`token_id_capture.external_staging: true`) each row
+additionally carries the token-free `CallRecord` custody columns —
+`parent_call_id`, `staging_key`, `weight_version`, `prev_len` / `delta_len` /
+`cum_len`, the staged record's `digest` and `extras_digest`, `mode`, and
+`logical_request_id` (the client header when present, else the vLLM response
```
lauradang, line 41: [`docs/design-docs/token-capture-ledger.md:41`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/docs/design-docs/token-capture-ledger.md#L41)

**2 action items.** PR-introduced.

### AI-1
This line and L111-112 describe a `logical_request_id`-based terminal join with a response-id fallback. At the Gym pin `logical_request_id` is header-only — [`records.py:252-255`](https://github.com/NVIDIA-NeMo/Gym/blob/c3d7cf4ff484d0f4f07898aedeaba05eb4821438/nemo_gym/token_id_capture/staging/records.py#L252-L255) says "Attribution never reads it" — the ledger writes it with no fallback ([`external_capture.py:249`](https://github.com/NVIDIA-NeMo/Gym/blob/c3d7cf4ff484d0f4f07898aedeaba05eb4821438/nemo_gym/token_id_capture/external_capture.py#L249)), and the join is [`resolve_terminal`](https://github.com/NVIDIA-NeMo/Gym/blob/c3d7cf4ff484d0f4f07898aedeaba05eb4821438/nemo_gym/token_id_capture/staging/attribution.py#L76-L81) on `response_id` (declared › response_id › content witness, then `select_terminal_call`), which RL calls at [`nemo_gym.py:840`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/environments/nemo_gym.py#L840). Please rewrite these two passages to match.

### AI-2
The MInf section (L87-104) should name the requirement on [NVIDIA/Megatron-LM#7015](https://github.com/NVIDIA/Megatron-LM/pull/7015) (`RequestPayloadStager` / `RequestPromptPreparer` hooks, `ng_commit_coords`), since the pinned Megatron-LM lacks them.

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -389,6 +395,29 @@ def finish_generation(self, *, release_gpu: bool = True) -> bool:
         ray.get(futures)
         return True
 
+    def setup_token_capture(
```
lauradang, line 397: [`nemo_rl/models/generation/megatron/megatron_generation.py:398`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/models/generation/megatron/megatron_generation.py#L398)

**1 action item.** PR description.

The PR body says this supports "both colocated and non-colocated SingleController Megatron-Inference deployments", but [`config.py:1009-1015`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/algorithms/single_controller_utils/config.py#L1009-L1015) unconditionally rejects `colocated.enabled=true` on the SC path, and [`setup.py:547`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/algorithms/single_controller_utils/setup.py#L547) says "Non-colocated only". 

### AI-1
Correct the description to non-colocated only (or, if colocated is intended, that validation and a test need to change — a much bigger scope).

nemo_rl/models/generation/megatron/megatron_generation.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
+import threading
```
lauradang: [`nemo_rl/models/generation/megatron/megatron_generation.py:15`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/models/generation/megatron/megatron_generation.py#L15-L16)

`threading` is never used in this file (ruff F401 in CI), and `logging`/`LOGGER` ([L49](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/models/generation/megatron/megatron_generation.py#L49)) are defined but never referenced.

### AI-1
```suggestion
```
and delete the `LOGGER = logging.getLogger(__name__)` line.

tests/unit/data_plane/test_tq_token_sink.py
```diff
@@ -218,3 +222,159 @@ def test_fetch_prefix_token_ids_rejects_duplicates(tq_client, staging_partition)
...
+            prompt_token_ids=[10],
+            generated_token_ids=[11],
+            generated_log_probs=[-0.1],
+        ),
+        finished_metadata=SimpleNamespace(policy_epoch=[(0, 7)]),
+    )
+    assert result is None
+
+
+def test_megatron_stager_declines_mixed_policy_epochs(tq_client, staging_partition):
```
lauradang: [`tests/unit/data_plane/test_tq_token_sink.py:361`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/tests/unit/data_plane/test_tq_token_sink.py#L361)

`stage()` catches every exception and returns `None` ([`tq_token_sink.py:487-491`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/data_plane/tq_token_sink.py#L487-L491)), so `assert result is None` also passes for a bad admission, a sink failure, or a typo in `_weight_version` — it doesn't pin the mixed-epoch rejection. None of the other `_weight_version` branches (missing/empty/malformed/negative) are covered.

### AI-1
Pin the logged reason and test the static method directly (one way to do it; not run here — no Linux env):
```python
    with caplog.at_level(logging.ERROR, logger="nemo_rl.data_plane.tq_token_sink"):
        result = stager.stage(...)
    assert result is None
    assert "spans policy epochs [7, 8]" in caplog.text


@pytest.mark.parametrize(
    ("policy_epoch", "match"),
    [
        (None, "no policy_epoch boundaries"),
        ([], "no policy_epoch boundaries"),
        ([(0,)], "invalid policy_epoch metadata"),
        ([(0, 7), (1, 8)], r"spans policy epochs \[7, 8\]"),
        ([(0, -1)], "negative policy epoch -1"),
    ],
)
def test_megatron_stager_weight_version_rejects_invalid_metadata(policy_epoch, match):
    with pytest.raises(ValueError, match=match):
        TQMegatronTokenStager._weight_version(SimpleNamespace(policy_epoch=policy_epoch))
```

nemo_rl/algorithms/single_controller_utils/setup.py
```diff
@@ -1087,22 +1108,51 @@ def setup_single_controller(
...
+            if router_replay_enabled(master_config.policy):
+                raise NotImplementedError(
+                    "Megatron token capture does not yet support router replay: "
+                    "the canonical MInf stager does not yet normalize routed experts"
+                )
+            ACTOR_ENVIRONMENT_REGISTRY[
+                "nemo_rl.models.policy.workers.megatron_policy_worker.MegatronPolicyWorker"
+            ] = PY_EXECUTABLES.MCORE_GYM
+        else:
+            ACTOR_ENVIRONMENT_REGISTRY[
```
lauradang, line 1195: [`nemo_rl/algorithms/single_controller_utils/setup.py:1148`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/algorithms/single_controller_utils/setup.py#L1148)

**1 action item — low severity.** PR-introduced.

This `else:` re-registers `VllmAsyncGenerationWorker → VLLM_GYM`, which [`ray_actor_environment_registry.py:19`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/distributed/ray_actor_environment_registry.py#L19) already sets unconditionally at import (at base and head). It is the only registry write outside the registry module and reads as if the vLLM capture path depends on it. (The megatron branch's `MCORE_GYM` registration just above, added in a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e, is the one doing real work — that one should stay.)

### AI-1
Delete the vLLM `else:` block, or turn it into an assert on `get_actor_python_env(...) == PY_EXECUTABLES.VLLM_GYM` if the intent was a guard.

tests/unit/single_controller/test_checkpointing.py
```diff
@@ -438,6 +438,12 @@ def sync_weights(self, *, kv_scales: Any = None) -> None:
     def shutdown(self) -> None:
         self.shutdown_count += 1
 
+    def blocks_training(self) -> bool:
+        return False
+
+    def wake_carries_weight_updates(self) -> bool:
+        return False
```
lauradang: [`tests/unit/single_controller/test_checkpointing.py:441`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/tests/unit/single_controller/test_checkpointing.py#L441-L445)

**1 action item — trivial.** PR-introduced.

`blocks_training` / `wake_carries_weight_updates` are only ever called on `policy_generation` ([`grpo.py:5215`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/algorithms/grpo.py#L5215), [`:5391`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/nemo_rl/algorithms/grpo.py#L5391)) and are declared on `GenerationInterface`, not `WeightSynchronizer`; this fake is passed only as `weight_synchronizer=` (L694). Dead code.

### AI-1
```suggestion
```

tests/unit/environments/test_nemo_gym_token_capture.py
```diff
@@ -36,9 +36,34 @@ def _digest(label: str) -> str:
...
+    token_capture: dict,
+) -> None:
+    with pytest.raises(ValueError, match="setup-derived generation_backend"):
+        _external_staging_backend(token_capture)
+
+
 def _manifest_record(
     call_id: str,
     *,
+    logical_request_id: str | None = None,
```
lauradang, line 66: [`tests/unit/environments/test_nemo_gym_token_capture.py:66`](https://github.com/NVIDIA-NeMo/RL/blob/a7cea3c8fd31d5da2811bd0f1939d63a73d1c06e/tests/unit/environments/test_nemo_gym_token_capture.py#L66)

**1 action item — trivial.** PR-introduced.

No test passes `logical_request_id=`; the helper always emits `None`. Either drop the parameter and the dict entry, or add one assertion that a non-`None` value survives `_assemble_receipt`.

PR #3873, opened 2026-08-27
  fix(megatron): refit before awaiting NeMo Gym startup

wdykas:
Description

  ## Summary

  Restore the correct initialization order for non-colocated Megatron generation with NeMo Gym, including MXFP8 rollouts.

  ## Problem

  The combination of #3569 and #3731 introduced a circular dependency:

  - NeMo Gym starts concurrently and waits for the reserved generation endpoint to become ready.
  - Megatron generation using `skip_weight_load` intentionally delays engine startup until after the initial refit. This ensures MXFP8 persistent buffers are installed before
  CUDA-graph capture.
  - The initialization path waited for NeMo Gym before initializing weight synchronization and performing that refit.

  As a result, Gym waited for the generation endpoint while generation waited for the refit that could only happen after Gym returned.

  Starting the generation engine before refit is not safe for MXFP8 because CUDA graphs may capture the wrong parameter buffers.

  ## Fix

  - Initialize the weight synchronizer and perform the initial policy-to-generation refit after the policy and generation workers are ready, but while NeMo Gym is still
  waiting for endpoint readiness.
  - Await NeMo Gym only after the initial refit has allowed the generation engine and HTTP endpoint to start.
  - Reuse the existing initialization helper for other configurations and avoid duplicate synchronization.
  - Add a regression test that verifies the required ordering:
    `Gym started → initial refit → generation endpoint ready`.
  - Repair the GB200 MXFP8 functional assertions:
    - Check `weight_sync_time_s`, which is where the initial refit is recorded.

  No dependency pins or submodule revisions are included.

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -1273,6 +1273,31 @@ def setup(
         )
         return mg, time.perf_counter() - t0
 
+    def init_megatron_weight_synchronizer(
+        policy: ColocatablePolicyInterface,
+        policy_generation: MegatronGeneration,
+    ) -> None:
+        """Initialize Megatron refit and load non-colocated generation weights."""
+        t0 = time.perf_counter()
+        weight_synchronizer = create_weight_synchronizer(
+            policy=policy,
+            generation=policy_generation,
+            generation_backend="megatron",
+            colocated=colocated_inference,
+            train_cluster=train_cluster,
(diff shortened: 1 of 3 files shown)


wdykas:
/ok to test [6a0961e](https://github.com/NVIDIA-NeMo/RL/pull/3873/commits/6a0961e18aa9bcf6df257fbf0243721d98481a87)

PR #3874, opened 2026-08-27
feat: add lora warm-start

pstjohn:
# What does this PR do?

Adds LoRA warm-start support so a new training run can initialize its adapters from an existing checkpoint—for example, continuing from an SFT LoRA checkpoint with GRPO.

This supports both training backends:

- **DTensor/AutoModel:** loads a PEFT adapter directory containing `adapter_model.safetensors` and `adapter_config.json`.
- **Megatron Core:** loads adapters from a native Megatron `iter_*` checkpoint or checkpoint root.

The implementation:

- Adds `restore_from` to the AutoModel LoRA and Megatron PEFT configurations.
- Validates donor paths, adapter configuration, and adapter tensor coverage before loading.
- Supports custom AutoModel state-dict adapters and expert-parallel MoE models.
- Loads only adapter weights; optimizer, RNG, train, and rerun state start fresh.
- Gives normal resume checkpoints precedence over `restore_from`.
- Anchors the KL reference model to the warm-started adapters.
- Adds defaults to example and reference configurations.
- Documents the workflow and backend-specific checkpoint formats.

# Issues

N/A

# Usage

## DTensor/AutoModel

```yaml
policy:
  dtensor_cfg:
    _v2: true
    lora_cfg:
      enabled: true
      restore_from: /path/to/sft-run/step_100/policy/weights
```

`restore_from` may point directly to the adapter directory or its `model/` subdirectory.

## Megatron Core

```yaml
policy:
  megatron_cfg:
    peft:
      enabled: true
      restore_from: /path/to/sft-run/step_100/policy/weights
```

`restore_from` may point to an `iter_XXXXXXX` directory or a checkpoint root that resolves to one.

The donor LoRA configuration must be compatible with the new run. In particular, adapter rank and scaling must match.

# Testing

- AutoModel adapter validation and loading: **11 passed**
- Megatron PEFT warm-start suite: **24 passed**
- Configuration validation tests: **14 passed**
- Targeted pre-commit checks: passed
- `git diff --check`: passed

The backend-specific tests were run in the prebuilt DTensor and Megatron container runtimes.

- [x] Added unit tests covering both backends, invalid configurations, resume behavior, relative paths, expert parallelism, and failure cleanup
- [x] Ran the relevant unit tests locally
- [x] Added documentation and updated example configurations
- [ ] Full functional test suite will run in CI

# Additional Information

Warm-starting is distinct from resuming:

- `restore_from` initializes a new run from donor adapter weights.
- Optimizer and training progress are not restored.
- If a normal NeMo RL resume checkpoint exists, its policy weights take precedence.
- For RL algorithms using a reference policy, the reference remains anchored to the warm-started initial policy.


nemo_rl/models/automodel/setup.py
```diff
@@ -586,6 +593,168 @@ def _noop(*args, **kwargs) -> None:  # pragma: no cover
...
+    """Warm-start the model's LoRA adapters from a donor PEFT adapter checkpoint.
+
+    The load goes through the Automodel checkpointer's PEFT path (each rank
+    reads adapter_model.safetensors, then
+    ``set_model_state_dict(broadcast_from_rank0=True)`` places it with
+    DTensor/EP awareness), the same machinery used to resume NeMo RL PEFT
+    checkpoints. Optimizer state is not loaded: warm starts begin with a
+    fresh optimizer.
+    """
+    adapter_dir = _resolve_lora_adapter_dir(restore_from)
```
pstjohn: **1 action item. Please fix this in this PR.** [`setup.py:725`](https://github.com/NVIDIA-NeMo/RL/blob/cd5a3d9e66f435f3a8d8e3364bed3c46e595ed5a/nemo_rl/models/automodel/setup.py#L725)

This is PR-introduced. With a relative `restore_from`, [the symlink](https://github.com/NVIDIA-NeMo/RL/blob/cd5a3d9e66f435f3a8d8e3364bed3c46e595ed5a/nemo_rl/models/automodel/setup.py#L737) stores a relative target inside the temporary directory, so it resolves beneath that directory and dangles. A relative directory named `model` also skips staging, while AutoModel selects PEFT loading using the literal [`"/model" in path` check](https://github.com/NVIDIA-NeMo/Automodel/blob/24b47e856263d313b942f0ed666c63fff83306b4/nemo_automodel/components/checkpoint/checkpointing.py#L806-L819).

### AI-1

Normalize the resolved directory before basename/symlink handling and cover both relative-path cases.

```suggestion
    adapter_dir = os.path.abspath(_resolve_lora_adapter_dir(restore_from))
```

nemo_rl/models/automodel/setup.py
```diff
@@ -586,6 +593,168 @@ def _noop(*args, **kwargs) -> None:  # pragma: no cover
...
+            f"dtensor_cfg.lora_cfg.restore_from: {adapter_dir}/"
+            "adapter_model.safetensors contains no tensors."
+        )
+    # HF PEFT exports prefix keys with "base_model.model."; the loader strips
+    # the prefix when present, so accept either form here.
+    prefix = "base_model.model."
+    stripped_donor_keys = {
+        key[len(prefix) :] if key.startswith(prefix) else key for key in donor_keys
+    }
+    expected_keys = {
```
pstjohn: **1 action item. Please fix this in this PR.** [`setup.py:691`](https://github.com/NVIDIA-NeMo/RL/blob/cd5a3d9e66f435f3a8d8e3364bed3c46e595ed5a/nemo_rl/models/automodel/setup.py#L691)

This is PR-introduced. For Qwen3-MoE with EP=1, `donor_keys` are HF-format names while `expected_keys` are native parameter names. AutoModel maps native `lora_gate_and_up_A` to PEFT [`base_layer.lora_B.weight`](https://github.com/NVIDIA-NeMo/Automodel/blob/24b47e856263d313b942f0ed666c63fff83306b4/nemo_automodel/components/models/qwen3_moe/state_dict_adapter.py#L146-L166), then performs the [reverse mapping during load](https://github.com/NVIDIA-NeMo/Automodel/blob/24b47e856263d313b942f0ed666c63fff83306b4/nemo_automodel/components/models/qwen3_moe/state_dict_adapter.py#L192-L203). This validator runs first and raises the new “donor adapter key mismatch” error for a valid donor.

### AI-1

Compare keys in the same post-conversion namespace and add a Qwen3-MoE EP=1 save/warm-start round-trip test.

nemo_rl/models/automodel/setup.py
```diff
@@ -586,6 +593,168 @@ def _noop(*args, **kwargs) -> None:  # pragma: no cover
...
+
+    The underlying PEFT load is unconditionally non-strict (a key mismatch is
+    only a warning), so this check is the only guarantee that the donor covers
+    every LoRA parameter -- it must run for all models, including those with a
+    custom state_dict_adapter. The sole exception is expert-parallel models:
+    their LoRA params get custom state_dict key names (e.g.
+    ``gate_up_linear.weight0``) that do not appear in ``named_parameters()``,
+    so their coverage cannot be checked against parameter names.
+    """
+    if _has_expert_parallelism(model):
```
pstjohn: **1 action item. Please fix this in this PR.** [`setup.py:674`](https://github.com/NVIDIA-NeMo/RL/blob/cd5a3d9e66f435f3a8d8e3364bed3c46e595ed5a/nemo_rl/models/automodel/setup.py#L674)

This is PR-introduced. With EP>1, a QKV-only donor and a current QKV+FC configuration pass here because validation returns immediately. AutoModel’s EP loader then [continues past missing or unmatched donor keys and only logs counts](https://github.com/NVIDIA-NeMo/Automodel/blob/24b47e856263d313b942f0ed666c63fff83306b4/nemo_automodel/components/checkpoint/stateful_wrappers.py#L121-L150), leaving the missing FC adapters freshly initialized while warm-start reports success.

### AI-1

Validate normalized complete adapter-key sets for EP checkpoints and add the QKV-only donor versus QKV+FC regression case.

nemo_rl/models/megatron/setup.py
```diff
@@ -485,6 +495,170 @@ def _resolve_iter_dir_from_root(path: str, not_found_msg: str) -> str:
...
+    def peft_warm_start_hook(model: list[MegatronModule]) -> list[MegatronModule]:
+        ckpt_cfg = megatron_cfg.checkpoint
+        original_load = ckpt_cfg.load
+        original_finetune = ckpt_cfg.finetune
+        original_load_optim = ckpt_cfg.load_optim
+        original_load_rng = ckpt_cfg.load_rng
+        ckpt_cfg.load = restore_dir
+        # finetune=False is required for the adapter-only filter, but
+        # optimizer/RNG state must not come from the donor run.
+        ckpt_cfg.finetune = False
```
pstjohn, line 632: **1 action item. Please fix this in this PR.** [`setup.py:629`](https://github.com/NVIDIA-NeMo/RL/blob/cd5a3d9e66f435f3a8d8e3364bed3c46e595ed5a/nemo_rl/models/megatron/setup.py#L629)

This is PR-introduced. Disabling optimizer and RNG restore does not make this donor load weights-only: Megatron-Bridge independently [restores the singleton rerun state](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/d352aceda8ed4136f1db787bcf449c1b210a2438/src/megatron/bridge/training/checkpointing.py#L3224-L3230), and its PEFT filtering [preserves non-model metadata](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/d352aceda8ed4136f1db787bcf449c1b210a2438/src/megatron/bridge/training/checkpointing.py#L3366-L3397). MCore then [overwrites mode, iteration, state, and pending-rerun flags](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/rerun_state_machine.py#L890-L936); resetting `TrainState` does not undo that.

### AI-1

Preserve/reset the current `RerunStateMachine` around donor loading, or use a true weights-only path, and add a donor-with-pending-rerun regression test.

tests/unit/models/megatron/test_megatron_setup.py
```diff
@@ -3625,3 +3626,518 @@ def test_megatron_internals_have_not_drifted(self):
...
+        even if Bridge stops honouring it -- degrading the warm start into a
+        strict full-model load. Skips when mcore/Bridge is unavailable.
+        """
+        import inspect
+
+        checkpointing = pytest.importorskip(
+            "megatron.bridge.training.checkpointing",
+            reason="requires the mcore extra (Megatron-Bridge)",
+        )
+        src = inspect.getsource(checkpointing)
```
pstjohn: **1 action item. Please fix this in this PR.** [`test_megatron_setup.py:3997`](https://github.com/NVIDIA-NeMo/RL/blob/cd5a3d9e66f435f3a8d8e3364bed3c46e595ed5a/tests/unit/models/megatron/test_megatron_setup.py#L3997)

This test gap is PR-introduced. The five searched strings occur together in both the upstream [adapter-filtering predicate](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/d352aceda8ed4136f1db787bcf449c1b210a2438/src/megatron/bridge/training/checkpointing.py#L3050-L3061) and [non-strict-load predicate](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/d352aceda8ed4136f1db787bcf449c1b210a2438/src/megatron/bridge/training/checkpointing.py#L3151-L3163). Removing either required block leaves every string elsewhere, so this tripwire still passes; the other warm-start tests mock the loader boundary.

### AI-1

Replace this module-wide source search with a focused loader/storage-mocked behavior test that proves both adapter filtering and non-strict loading remain active.

pstjohn:
/ok to test [cc2a58a](https://github.com/NVIDIA-NeMo/RL/pull/3874/commits/cc2a58a0637339d90f8e106a1197484dbcf57fe6)

nemo_rl/models/automodel/setup.py
pstjohn: Resolved in cc2a58a0.

nemo_rl/models/automodel/setup.py
pstjohn: Resolved in cc2a58a0.

nemo_rl/models/automodel/setup.py
pstjohn: Resolved in cc2a58a0.

nemo_rl/models/megatron/setup.py
pstjohn: Resolved in cc2a58a0.

tests/unit/models/megatron/test_megatron_setup.py
pstjohn: Resolved in cc2a58a0.

yuki-97 (commented):
@RayenTian could you help review this one?

nemo_rl/models/megatron/setup.py
```diff
@@ -587,6 +598,177 @@ def _resolve_iter_dir_from_root(path: str, not_found_msg: str) -> str:
...
+                "same LoRA rank and scaling; train a new adapter instead."
+            )
+    for key in ("target_modules", "exclude_modules"):
+        if key not in saved_peft:
+            raise ValueError(
+                f"megatron_cfg.peft.restore_from={restore_dir!r}: the donor "
+                f"checkpoint's run_config.yaml peft section has no {key!r} key; "
+                "cannot verify compatibility with this run's peft config."
+            )
+        if set(saved_peft[key]) != set(peft_cfg[key]):
```
RayenTian: [`nemo_rl/models/megatron/setup.py:680`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/megatron/setup.py#L680)

`set()` raises on `None`, so a donor whose `run_config.yaml` carries `target_modules: null` fails here with a bare `TypeError: 'NoneType' object is not iterable` instead of the actionable `ValueError` this function produces everywhere else.

The trigger is narrow: NeMo RL's own configs always write `[]`, so a NeMo RL-produced donor is unaffected. It only comes up for a native Megatron-Bridge donor trained in the exclude-modules style — which the comment just below at L687-L690 says this validator is meant to accommodate. Bridge's [`LoRA` docstring](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/5ed97996cc2b422904d18179375b6d7366915097/src/megatron/bridge/peft/lora.py#L110-L112) recommends exactly that pairing (*"will require target_modules to be empty list or None"*), [`ModuleMatcher` tolerates it](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/5ed97996cc2b422904d18179375b6d7366915097/src/megatron/bridge/peft/module_matcher.py#L91) via `self.target_modules or []`, and neither class defines a `__post_init__` that would normalize `None` to `[]` before the config is serialized.

`None` and `[]` are equivalent in Bridge's matcher, so this keeps the semantics unchanged:

```suggestion
        if set(saved_peft[key] or []) != set(peft_cfg[key] or []):
```

nemo_rl/models/megatron/setup.py
```diff
@@ -587,6 +598,177 @@ def _resolve_iter_dir_from_root(path: str, not_found_msg: str) -> str:
...
+    # for MoE expert layers. NeMo RL never sets them (a run always uses the
+    # bridge defaults), but a native Megatron-Bridge donor checkpoint may
+    # have; a mismatch would restore onto a different adapter layout.
+    moe_shaping_defaults = {
+        "normalize_moe_lora": False,
+        "share_expert_adapters": True,
+        "experts_shared_outer_loras": False,
+    }
+    for key, default in moe_shaping_defaults.items():
+        if key in saved_peft and saved_peft[key] != peft_cfg.get(key, default):
```
RayenTian: [`nemo_rl/models/megatron/setup.py:691`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/megatron/setup.py#L691)

Would it be worth reading these three defaults off the bridge dataclass rather than restating them? They match [`LoRA`'s current defaults](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/5ed97996cc2b422904d18179375b6d7366915097/src/megatron/bridge/peft/lora.py#L155-L157) exactly today, so this is about the upgrade path rather than a present-day bug.

The part I'm unsure about: since NeMo RL never passes these three to `LoRA(...)`, a run's effective value always tracks whatever bridge's default is, while `peft_cfg.get(key, default)` here keeps comparing against the literal. If a future bridge bump flipped one, this would reject a valid donor and the error message would report a mismatch against a value the run isn't actually using — which would be a confusing one to chase.

`fields` and `LoRA` are both already imported in this module, so it stays local:

```suggestion
    # Read the defaults off the bridge LoRA dataclass so a future bridge
    # default change can't silently invalidate this check.
    lora_field_defaults = {f.name: f.default for f in fields(LoRA)}
    for key in (
        "normalize_moe_lora",
        "share_expert_adapters",
        "experts_shared_outer_loras",
    ):
        default = lora_field_defaults[key]
        if key in saved_peft and saved_peft[key] != peft_cfg.get(key, default):
```

nemo_rl/models/automodel/setup.py
```diff
@@ -526,6 +536,186 @@ def setup_distributed(
...
+        )
+    # HF PEFT exports prefix keys with "base_model.model."; the loader strips
+    # the prefix when present, so accept either form here.
+    prefix = "base_model.model."
+    normalized_donor_keys = {
+        key[len(prefix) :] if key.startswith(prefix) else key for key in donor_keys
+    }
+
+    expected_state_dict = {
+        f"{prefix}{name.replace('_checkpoint_wrapped_module.', '')}": (
```
RayenTian: [`nemo_rl/models/automodel/setup.py:634`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/setup.py#L634)

Automodel's save path normalizes this same FQN through its public [`canonical_parameter_fqn`](https://github.com/NVIDIA-NeMo/Automodel/blob/1814c6c93a66b9d59d254960ef6a99a64249b671/nemo_automodel/shared/parameter_names.py#L20-L34) helper ([called at `stateful_wrappers.py#L154`](https://github.com/NVIDIA-NeMo/Automodel/blob/1814c6c93a66b9d59d254960ef6a99a64249b671/nemo_automodel/components/checkpoint/stateful_wrappers.py#L154)), which resolves the prefix from torch rather than restating it:

```python
return name.replace(_CHECKPOINT_PREFIX, "")
```

Since this key set has to stay byte-identical to what the saver writes, calling the same helper keeps the two from drifting. If the normalization ever gains a step (say, stripping `_orig_mod.` for `torch.compile`), the hardcoded literal would silently stop matching and every warm start would fail with a "donor adapter key mismatch" pointing at the donor rather than at the real cause. Behaviour is identical today.

```suggestion
        f"{prefix}{canonical_parameter_fqn(name)}": (
```

This needs `from nemo_automodel.shared.parameter_names import canonical_parameter_fqn` added to the import block — it sorts directly after `nemo_automodel.components.distributed.tensor_utils`.

docs/guides/lora.md
```diff
@@ -126,6 +130,23 @@ policy:
...
+`restore_from` initializes this run's adapters from a donor PEFT checkpoint, e.g. to carry an
+SFT LoRA into GRPO. The accepted path format differs by backend:
+
+- **DTensor (Automodel)**: `policy.dtensor_cfg.lora_cfg.restore_from` — a directory containing
+  `adapter_model.safetensors` + `adapter_config.json` (a previous run's `step_*/policy/weights`
+  or its `model/` subdirectory).
+- **Megatron Core**: `policy.megatron_cfg.peft.restore_from` — a native Megatron `iter_XXXXXXX`
+  directory, or a checkpoint root containing one.
+
+The donor's `dim`/`alpha` must match this run's; setup fails loudly otherwise. Optimizer, RNG,
```
RayenTian: [`docs/guides/lora.md:146`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/docs/guides/lora.md#L146)

Could this mention the base-model check as well? On the DTensor backend `_validate_lora_adapter_config` also requires the donor's `base_model_name_or_path` to equal this run's `model_name` as an exact string ([`automodel/setup.py#L597-L604`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/setup.py#L597-L604)). Automodel writes that field from `model_part.config.name_or_path` ([`addons.py#L332`](https://github.com/NVIDIA-NeMo/Automodel/blob/1814c6c93a66b9d59d254960ef6a99a64249b671/nemo_automodel/components/checkpoint/addons.py#L332)) — i.e. whatever string was handed to `from_pretrained` — so an HF repo id and a local path for the *same* model are rejected, as is the same model at a new path after a cluster move.

The error message is clear when it fires, so nobody is stuck; it's just the one rejection a user can hit with `dim`/`alpha` already matching, which is what the current sentence points them at.

```suggestion
The donor's `dim`/`alpha` must match this run's; on DTensor the donor's
`base_model_name_or_path` must also equal this run's `policy.model_name` as an exact string
(an HF repo id and a local path for the same model will not match). Setup fails loudly
otherwise. Optimizer, RNG,
```

nemo_rl/models/automodel/setup.py
```diff
@@ -526,6 +536,186 @@ def setup_distributed(
...
+        model,
+        moe_mesh=getattr(checkpoint_manager, "moe_mesh", None),
+    )
+    staging_dir = None
+    load_dir = adapter_dir
+    if os.path.basename(adapter_dir.rstrip(os.sep)) != "model":
+        # Automodel's checkpointer selects its PEFT safetensors read by a
+        # substring test on the path ("/model" in path); a bare adapter
+        # directory would fall through to the DCP/HF-storage-reader branch
+        # instead. Expose it under a temporary "model" path component.
```
RayenTian: [`nemo_rl/models/automodel/setup.py:692`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/setup.py#L692)

At the Automodel commit this repo pins, the selector is a basename comparison rather than a substring test — [`_is_model_checkpoint_path`](https://github.com/NVIDIA-NeMo/Automodel/blob/1814c6c93a66b9d59d254960ef6a99a64249b671/nemo_automodel/components/checkpoint/checkpointing.py#L2038-L2040):

```python
def _is_model_checkpoint_path(path: str) -> bool:
    return Path(path.rstrip("/")).name == "model"
```

The guard on the line above is exactly its complement, so the code is correct — it's the comment that describes a looser rule than upstream applies. Under a substring test `/foo/model_v2/adapter` would skip staging; under the basename rule it doesn't. The substring form does exist at older Automodel commits, so this looks like it simply drifted as the submodule pin moved. The same wording appears in `test_bare_adapter_dir_loaded_via_model_path`'s docstring.

```suggestion
        # Automodel's checkpointer selects its PEFT safetensors read by a
        # basename test on the path (_is_model_checkpoint_path:
        # Path(path).name == "model"); a bare adapter directory would fall
        # through to the DCP/HF-storage-reader branch instead. Expose it
        # under a temporary "model" path component.
```

tests/unit/models/automodel/test_automodel_setup.py
```diff
@@ -2245,3 +2274,590 @@ def test_compatible_adapter_no_change(self):
...
+        adapter_dir = tmp_path / "adapter"
+        _write_adapter_checkpoint(adapter_dir, keys=_QWEN3_MOE_HF_LORA_KEYS[:-1])
+        with pytest.raises(ValueError, match="Missing from donor"):
+            _validate_lora_adapter_keys(
+                str(adapter_dir), _TinyQwen3MoeLoraModel(ep_size=2)
+            )
+
+
+@pytest.mark.automodel
+class TestLoadInitialLoraAdapter:
```
RayenTian, line 2621: [`tests/unit/models/automodel/test_automodel_setup.py:2557`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/tests/unit/models/automodel/test_automodel_setup.py#L2557)

Two gaps worth considering. Describing rather than suggesting code here, since I couldn't run pytest in my environment to verify a suggestion would pass.

**1. Resume-vs-`restore_from` precedence isn't covered on the DTensor side.** The tests in this class call `_load_initial_lora_adapter` directly, so the dispatch at [`automodel/setup.py#L1003-L1011`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/setup.py#L1003-L1011) (`if weights_path: ... elif lora_enabled and restore_from is not None: ...`) has no coverage. That branch is what encodes the "resumed weights take precedence" guarantee from the PR description, and the Megatron side pins it with `test_policy_warm_start_hook_skipped_on_resume`. A refactor turning that `elif` into an `if` would let a warm start overwrite resumed weights with no visible symptom.

**2. The Megatron hook tests don't pin hook ordering.** `test_policy_warm_start_hook_appended_on_fresh_run` asserts only `mock_hook.assert_called_once()`. Since `get_model` is patched, nothing checks that the returned hook actually landed in `pre_wrap_hook`, nor its position relative to `composed_peft_hook` — which [`_create_peft_warm_start_hook`'s docstring](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/megatron/setup.py#L714) states must be immediately after, and the ordering is load-bearing (too early and the adapters aren't attached; too late and the module is wrapped, changing the sharded state-dict keys). `get_model.call_args.kwargs["pre_wrap_hook"]` is already reachable in those tests, so asserting the index would be cheap.

nemo_rl/models/automodel/setup.py
```diff
@@ -526,6 +536,186 @@ def setup_distributed(
...
+            "dtensor_cfg.lora_cfg.restore_from: donor adapter key mismatch "
+            f"against this run's LoRA parameters. Missing from donor: "
+            f"{missing[:5]}{' ...' if len(missing) > 5 else ''}; unexpected in "
+            f"donor: {unexpected[:5]}{' ...' if len(unexpected) > 5 else ''}. "
+            "The donor adapter must target the same modules as this run's "
+            "lora_cfg."
+        )
+
+
+def _load_initial_lora_adapter(
```
RayenTian, line 650: [`nemo_rl/models/automodel/setup.py:666`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/setup.py#L666)

Would this fit better as a method on `AutomodelCheckpointManager`? It's close to a line-for-line variant of the existing [`load_checkpoint`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/checkpoint.py#L385-L432) — same `checkpointer is not None` assert, same three `config_updates` keys (down to the "already dequantized" comment), same `checkpoint_root` derivation, same "resolve the `model` subdirectory" step — but written as a free function that reaches through the manager into `checkpoint_manager.checkpointer`.

Making it `AutomodelCheckpointManager.load_lora_adapter(...)` would put it beside the sibling it mirrors and keep the checkpointer's config lifecycle owned by the class that owns the checkpointer. As written, the config mutation at [L704-L712](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/setup.py#L704-L712) is left in place after the call, with `checkpoint_root` pointing into the staging directory that the `finally` then removes — so correctness depends on every later caller re-setting those fields before use, an invariant spread across two modules rather than held by the class that owns the checkpointer.

Relatedly, [`_resolve_lora_adapter_dir`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/setup.py#L539) is the same shape as [`_infer_checkpoint_root`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/checkpoint.py#L482) and [`detect_checkpoint_format`](https://github.com/NVIDIA-NeMo/RL/blob/041a89c9433b57e6b5aac9b6717be5ab91c67f86/nemo_rl/models/automodel/checkpoint.py#L447) — "given a user-supplied path, work out what to actually read" — which already live in `checkpoint.py`. The two `_validate_lora_adapter_*` functions read more like config validation, so leaving those in `setup.py` seems reasonable either way.

RayenTian (commented):
Reviewed the warm-start implementation on both backends, focusing on the upstream Automodel / Megatron-Bridge API usage and the KL reference-anchoring behaviour.

The feature looks solid. Validation is genuinely fail-closed on both paths; the Megatron hook mirrors megatron-bridge's own PEFT-resume hook parameter-for-parameter; and the reference-anchoring invariant holds on both backends — on DTensor via the existing `defer_checkpoint_load` mechanism, which is subtle but correct. `test_donor_adapter_weights_land_in_model` driving a real `Checkpointer` end-to-end is a nice touch, and much stronger than mock-only coverage.

Seven comments below: one on where `_load_initial_lora_adapter` sits relative to `AutomodelCheckpointManager`, plus a small robustness fix, two "read the value from upstream rather than restate it" suggestions, a doc addition, a stale comment, and two test gaps.

_Generated by Claude Code_

RayenTian:
Hi @pstjohn. Thanks for adding this. Left some comments. Besides, would you please run pre-commit for these commits to fix Link check?

nemo_rl/models/megatron/setup.py
pstjohn: great, makes sense, changed. Thanks!

nemo_rl/models/megatron/setup.py
pstjohn: Sure that makes sense. The validator now reads these defaults from `fields(LoRA)`. Added a regression test that changes a Bridge default and verifies that validation follows it.

nemo_rl/models/automodel/setup.py
pstjohn: Sure, changed. Thanks!

docs/guides/lora.md
pstjohn: changed!

nemo_rl/models/automodel/setup.py
pstjohn: changed, thanks

tests/unit/models/automodel/test_automodel_setup.py
pstjohn: Added DTensor setup coverage for cold start, warm start, and resume with `restore_from` configured, explicitly asserting that resume skips the warm-start load. The Megatron policy test now inspects the hooks passed to `get_model` and asserts that the warm-start hook immediately follows `composed_peft_hook`.

nemo_rl/models/automodel/setup.py
pstjohn: Yup, moved the loading and staging logic into `AutomodelCheckpointManager.load_lora_adapter()` and the path resolver into `checkpoint.py`; validation stays in `setup.py`. The manager now restores the previous checkpointer configuration in `finally`, before removing the staging directory. Added tests covering configuration restoration and staging cleanup on both success and load failure.

nemo_rl/models/megatron/setup.py
yuki-97: fixed in eb303685acec82e91804ece779c01470bf36faa3

nemo_rl/models/megatron/setup.py
yuki-97: fixed in eb303685acec82e91804ece779c01470bf36faa3

nemo_rl/models/automodel/setup.py
yuki-97: fixed in eb303685acec82e91804ece779c01470bf36faa3

docs/guides/lora.md
yuki-97: fixed in eb303685acec82e91804ece779c01470bf36faa3

nemo_rl/models/automodel/setup.py
yuki-97: fixed in eb303685acec82e91804ece779c01470bf36faa3

tests/unit/models/automodel/test_automodel_setup.py
yuki-97: fixed in eb303685acec82e91804ece779c01470bf36faa3

nemo_rl/models/automodel/setup.py
yuki-97: fixed in eb303685acec82e91804ece779c01470bf36faa3

yuki-97 (approved):

PR #3879, opened 2026-08-27
feat(ppo): add Perfetto workload tracing

youngeunkwon0405:
## What does this PR do?

Adds opt-in Chrome JSON/Perfetto workload tracing for synchronous and asynchronous PPO.

- Mirrors the existing PPO driver timers into nested Perfetto spans, with explicit step and PPO-epoch markers.
- Adds native per-sample rollout tracks with turn, generation, reward, tokenization, and outcome metadata.
- Merges asynchronous trajectory-collector events into the driver trace after midpoint clock alignment.
- Records async replay-buffer size and average trajectory age on step-complete events.
- Keeps tracing disabled by default and treats trace failures as non-fatal warnings.
- Documents usage, trace layout, and current limitations.

## Usage

```yaml
logger:
  perfetto:
    enable: true
    name: ppo_trace.json
```

Relative names are written under the resolved experiment `logger.log_dir`;
absolute paths are preserved. The same values can be supplied as CLI overrides:

```bash
uv run examples/run_ppo.py --config <ppo-recipe.yaml> \
  logger.perfetto.enable=true \
  logger.perfetto.name=ppo_trace.json
```

Open the resulting JSON directly in https://ui.perfetto.dev/.

## Example traces

NVIDIA access is required for these shared files:

- [Synchronous PPO trace](https://drive.google.com/file/d/1V0vVve322gtlVGQYxbjawzcKkVYYtrTY/view?usp=drive_link)
- [Asynchronous PPO trace](https://drive.google.com/file/d/19m7fVbL95miPmBcOr49egDBshmeImR59/view?usp=drive_link)

## Validation

- Ruff 0.9.9 format and lint checks passed.
- Focused unit/rollout/config suite: 114 passed (`test_trace.py`, `test_timer.py`, `test_rollouts.py`, `test_config_v2.py`; CW-DFW Slurm job 16755447).
- Sync PPO: 2 steps completed (`16749554`, exit 0); validated 550 events, nested driver phases, step/epoch markers, and 64 finite-duration sample rollouts.
- Async PPO: 2 steps completed (`16751626`, exit 0); validated 1,012 merged events, driver and collector processes, 4 collector batch spans, 128 finite-duration sample rollouts, and step metadata with buffer size plus trajectory ages 0 and 1.

## Notes

The trace is intended for CPU-side workload and lifecycle analysis. Nsight Systems remains the appropriate tool for CUDA kernels, streams, and NCCL detail.


nemo_rl/algorithms/ppo.py
```diff
@@ -2011,6 +2067,7 @@ def ppo_train(
     # so without this the daemon finalization thread could be killed before the
     # final tmp_step_N is renamed.
     checkpointer.shutdown()
+    _save_sync_trace()
```
youngeunkwon0405: [`nemo_rl/algorithms/ppo.py:2070`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/algorithms/ppo.py#L2070)

**[BUG]** Sync `ppo_train` writes the trace only on the timeout / max-steps / normal-completion paths, so any training exception discards the entire trace — exactly the run you'd most want to inspect. Async PPO already saves in a `finally` via `_shutdown_workers` ([ppo.py:3122-3129](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/algorithms/ppo.py#L3122)), and [docs/perfetto-tracing.md](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/docs/perfetto-tracing.md#L76) says "handled training failures write the trace", which doesn't hold for sync today.

Consider wrapping the sync training loop in `try: ... finally: _save_sync_trace()` — that also collapses the three explicit call sites into one. While touching this, moving `driver_trace.finalize_open_spans()` inside the existing `try` here and in `_shutdown_workers` ([ppo.py:2206](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/algorithms/ppo.py#L2206)) would guarantee span finalization can never raise past the non-fatal boundary.

nemo_rl/algorithms/async_utils/trajectory_collector.py
```diff
@@ -249,7 +252,18 @@ def __init__(
         self._outstanding_task_indices: set[int] = set()
 
         # Timer for efficiency metrics
-        self._efficiency_timer = ThreadSafeTimer(context={"worker": "collector"})
+        trace_enabled, _ = resolve_trace_config(master_config.logger)
```
youngeunkwon0405, line 295: [`trajectory_collector.py:255`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/algorithms/async_utils/trajectory_collector.py#L255)

**[BUG]** The collector enables tracing for the GRPO branch too (`algorithm_name = "grpo"`, and `LoggerConfig` is shared so `logger.perfetto.enable=true` is settable from GRPO YAMLs), but `grpo.py` has no `collect_trace`/`save_trace` consumer. A GRPO user who enables perfetto would get unbounded per-sample event accumulation in the actor for the whole run with no output file and no warning.

Consider forcing `trace_enabled = False` (or emitting a loud warning) for the GRPO branch until GRPO harvesting exists, so misconfiguration fails visibly rather than silently.

nemo_rl/utils/trace.py
```diff
@@ -0,0 +1,454 @@
...
+
+    enabled = bool(perfetto_config.get("enable", False))
+    name = str(perfetto_config.get("name", _DEFAULT_TRACE_NAME))
+    if enabled and not name:
+        raise ValueError("logger.perfetto.name must be non-empty when enabled")
+
+    output_path = Path(name)
+    if not output_path.is_absolute():
+        output_path = Path(logger_config.get("log_dir", ".")) / output_path
+    return enabled, output_path
```
youngeunkwon0405, line 58: [`nemo_rl/utils/trace.py:46-58`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/utils/trace.py#L46)

**[GUIDELINE]** `enable`/`name` are required keys of `PerfettoConfig` and `log_dir` is required in `LoggerConfig`, so the inner `.get(key, default)` calls are the call-site-default pattern the config-conventions skill forbids — the exemplar YAML should be the only source of defaults, with missing keys failing loudly. (The outer `logger_config.get("perfetto")` + `isinstance` check is right and should stay, since the block is `NotRequired`.) All three PPO exemplars either carry the full block or omit it entirely, so direct indexing breaks nothing:

```suggestion
    perfetto_config = logger_config.get("perfetto")
    if not isinstance(perfetto_config, Mapping):
        return False, Path(logger_config["log_dir"]) / _DEFAULT_TRACE_NAME

    enabled = bool(perfetto_config["enable"])
    name = str(perfetto_config["name"])
    if enabled and not name:
        raise ValueError("logger.perfetto.name must be non-empty when enabled")

    output_path = Path(name)
    if not output_path.is_absolute():
        output_path = Path(logger_config["log_dir"]) / output_path
    return enabled, output_path
```

nemo_rl/experience/rollouts.py
```diff
@@ -1275,21 +1324,37 @@ async def run_sample_multi_turn_rollout(
                 turn_multimodal_data = dict(sample_multimodal_data)
                 turn_multimodal_data["vllm_content"] = None
 
+            if tracer is None:
```
youngeunkwon0405, line 1343: [`nemo_rl/experience/rollouts.py:1327`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/experience/rollouts.py#L1327)

**[GUIDELINE]** This `if tracer is None / else` duplicates a 10-argument call verbatim just to omit three kwargs; `async_generate_response_for_sample_turn` already defaults `tracer=None, trace_sample_id=None, turn=0` and guards internally ([rollouts.py:1195](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/experience/rollouts.py#L1195)). Always passing `tracer=tracer, trace_sample_id=trace_sample_id, turn=turn` and deleting the branch is behavior-identical.

Minor related note: `initial_input_tokens` at [line 1265](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/experience/rollouts.py#L1265) is summed for every sample even when `tracer is None` (all GRPO paths); it could move under the `if tracer is not None` guard. (`trace_sample_id` needs to stay where it is, since turn spans reuse it.)

nemo_rl/utils/trace.py
```diff
@@ -0,0 +1,454 @@
...
+                ]
+                _shift_timestamps(actor_events, offset_us)
+                merged.extend(actor_events)
+            except Exception as error:
+                warnings.warn(
+                    f"Could not collect Perfetto events from Ray actor: {error}",
+                    stacklevel=2,
+                )
+
+    merged.sort(key=lambda event: (event.get("ts", -1), event.get("ph", "")))
```
youngeunkwon0405: [`nemo_rl/utils/trace.py:443`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/utils/trace.py#L443)

**[BUG]** Since `"B" < "E"`, this key reorders a same-microsecond `E`-then-`B` pair on the same tid into `B`-then-`E`, which mis-nests the viewer's span stack (reachable when `timer.stop()` is immediately followed by `timer.start()`). Python's sort is stable, so sorting by `ts` alone preserves each stream's correct append order and keeps cross-stream ties deterministic; metadata events sort to the front via the `-1` default either way.

```suggestion
    merged.sort(key=lambda event: event.get("ts", -1))
```

nemo_rl/algorithms/ppo.py
```diff
@@ -2085,7 +2142,17 @@ def async_ppo_train(
     # Import async utilities only when needed (heavy Ray actors).
     from nemo_rl.algorithms.async_utils import AsyncTrajectoryCollector, ReplayBuffer
 
-    timer = Timer(context={"worker": "driver"})
+    trace_enabled, trace_output_path = resolve_trace_config(master_config.logger)
+    driver_trace = Tracer(
+        "driver",
+        virtual_process_name="ppo_async_driver_events",
```
youngeunkwon0405: [`nemo_rl/algorithms/ppo.py:2148`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/algorithms/ppo.py#L2148)

**[GUIDELINE]** The async driver tracer never opens an async span (only `instant`/`counter`/timer spans), yet `virtual_process_name="ppo_async_driver_events"` unconditionally emits process metadata — producing a permanently empty process group in the Perfetto UI that the docs don't mention (they list only `ppo_sync_rollouts`/`ppo_async_rollouts`). Passing `virtual_process_name=None` here (or documenting the group) would keep the trace layout matching the docs.

nemo_rl/utils/logger.py
```diff
@@ -79,6 +79,11 @@ class GPUMonitoringConfig(TypedDict):
     flush_interval: int | float
 
 
+class PerfettoConfig(TypedDict):
```
youngeunkwon0405, line 83: [`nemo_rl/utils/logger.py:82`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/utils/logger.py#L82)

**[GUIDELINE]** (low severity) `PerfettoConfig` is a new TypedDict config class; the config-conventions skill asks that new config classes be `pydantic.BaseModel`, reserving TypedDict edits for extending existing legacy classes. Nesting inside the legacy `LoggerConfig` and matching siblings like `MLflowConfig` is a fair consistency argument, so this may be intentional — flagging for an explicit decision either way.

tests/unit/utils/test_trace.py
```diff
@@ -0,0 +1,114 @@
...
+
+    assert enabled is True
+    assert output_path == tmp_path / "exp_001" / "ppo_trace.json"
+
+
+def test_resolve_trace_config_disabled_when_block_is_missing(tmp_path: Path):
+    enabled, output_path = resolve_trace_config({"log_dir": str(tmp_path)})
+
+    assert enabled is False
+    assert output_path == tmp_path / "nemo_rl_perfetto_trace.json"
```
youngeunkwon0405, line 164: [`tests/unit/utils/test_trace.py`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/tests/unit/utils/test_trace.py#L114)

**[TEST]** The suite here is solid (verified 7/7 passing locally). A few small additions would pin the contracts default runs rely on — all four verified passing locally:

```python
def test_disabled_tracer_async_spans_are_noop():
    tracer = Tracer("driver", virtual_process_name="rollouts")

    tracer.start_async_span("rollout", "sample", track_name="sample 0")
    tracer.end_async_span("sample", args={"total_reward": 1.0})
    tracer.finalize_open_spans()

    assert tracer.events() == []


def test_resolve_trace_config_preserves_absolute_name(tmp_path: Path):
    absolute_name = tmp_path / "shared" / "trace.json"
    enabled, output_path = resolve_trace_config(
        {
            "log_dir": str(tmp_path / "exp_001"),
            "perfetto": {"enable": True, "name": str(absolute_name)},
        }
    )

    assert enabled is True
    assert output_path == absolute_name


def test_resolve_trace_config_rejects_empty_name_when_enabled():
    with pytest.raises(ValueError, match="non-empty"):
        resolve_trace_config(
            {"log_dir": "logs", "perfetto": {"enable": True, "name": ""}}
        )


def test_save_trace_skips_empty_event_list(tmp_path: Path):
    output_path = tmp_path / "trace.json"

    assert save_trace([], output_path=output_path) is None
    assert not output_path.exists()
```

The first one matters most: every default (tracing-disabled) async run relies on `start_async_span`/`end_async_span`/`finalize_open_spans` silently no-op'ing, and only `span()` is covered today. The absolute-path and empty-name cases pin behaviors the docs explicitly promise.

nemo_rl/utils/timer.py
```diff
@@ -132,6 +174,8 @@ def stop(self, label: str, should_log: bool = True) -> float:
             self._timers[label] = []
         self._timers[label].append(elapsed)
         del self._start_times[label]
+        if self._trace is not None:
```
youngeunkwon0405, line 180: [`nemo_rl/utils/timer.py:177`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/utils/timer.py#L177)

**[NOTE]** (latent, no action required) `Timer.stop(label)` is label-addressed and legal to call non-LIFO (`start("a"); start("b"); stop("a")`), but the mirrored `end_span()` closes the most recent `B` event on the thread, so out-of-order stops would silently mis-attribute trace spans. All current call sites are LIFO-safe (audited independently three times), so nothing is wrong today — a short docstring note on `stop()` documenting the LIFO expectation would protect future callers.

nemo_rl/utils/trace.py
```diff
@@ -0,0 +1,454 @@
...
+            open_span_ids = list(self._open_async_spans)
+        for span_id in open_span_ids:
+            self.end_async_span(span_id, args={"incomplete": True})
+
+    def events(self) -> list[dict[str, Any]]:
+        """Return a shallow copy of all collected trace events."""
+        with self._lock:
+            return [dict(event) for event in self._events]
+
+    def collect_trace(self, timing: bool = False) -> int | list[dict[str, Any]]:
```
youngeunkwon0405: [`nemo_rl/utils/trace.py:392`](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/utils/trace.py#L392)

**[NOTE]** (low severity) `Tracer.collect_trace` appears to have no caller — `save_trace` invokes the collector actor's own `collect_trace` ([trajectory_collector.py:942](https://github.com/NVIDIA-NeMo/RL/blob/31b71da71affa2c9c39d4fda852d54d9852bcb8e/nemo_rl/algorithms/async_utils/trajectory_collector.py#L942)), which deliberately does *not* finalize open spans while this one does. Also, the `timing: bool` flag switching the return type between `int` and `list[dict]` reads less clearly than two methods (e.g. `trace_clock_us()` and `collect_trace()`). Consider removing the unused method or splitting the two concerns.

youngeunkwon0405 (commented):
This review was produced by a coordinated multi-agent pass (RL-codebase expert, test reviewer, independent bug scan, and an adversarial verification round; ruff format/lint clean). The feature is a valuable addition — the trace layout is well designed (consistent monotonic-µs clocks, midpoint actor alignment, atomic tmp-file writes, thread-safe collection), the docs are accurate on the sync path, and the tests pass locally.

The inline comments cover: two behavioral gaps (sync-path trace loss on exceptions; GRPO configs silently accumulating events with no writer), a few config-convention and cleanup suggestions, one sort-order edge in the merge path, and some verified optional test additions. Nothing blocking beyond the two behavioral items.

_Generated by Claude Code_

nemo_rl/algorithms/ppo.py
youngeunkwon0405: Fixed in c0e184966. ppo_train now owns the trace lifecycle in an outer finally, and both finalization and writing are inside the non-fatal boundary. I also added test_sync_ppo_writes_trace_when_training_raises.

nemo_rl/algorithms/async_utils/trajectory_collector.py
youngeunkwon0405: Addressed by stacked PR #3887. GRPO now harvests and saves traces for legacy sync, TQ sync, and async collector paths; its exemplar Perfetto config is introduced in that GRPO-specific PR rather than this PPO base PR.

nemo_rl/utils/trace.py
youngeunkwon0405: Fixed in c0e184966. Required logger.log_dir, perfetto.enable, and perfetto.name keys now use direct indexing; only the optional outer perfetto block uses get().

nemo_rl/experience/rollouts.py
youngeunkwon0405: Partially applied and intentionally kept the branch. Always passing the three tracing kwargs breaks existing monkeypatched generation implementations: job 16763675 reproduced unexpected keyword argument tracer in test_async_vlm_multiturn_drops_stale_native_content. I did move the initial-token sum under the tracer guard. Pyrefly, Ruff, and all 52 rollout tests pass.

nemo_rl/utils/trace.py
youngeunkwon0405: Fixed with a regression test in c0e184966. The merge now uses a stable timestamp-only sort, preserving source-stream order for equal timestamps so an end event remains before the next begin event.

nemo_rl/algorithms/ppo.py
youngeunkwon0405: Fixed in c0e184966. The async PPO driver no longer declares the unused virtual process, so Perfetto does not show an empty lane.

nemo_rl/utils/logger.py
youngeunkwon0405: I am intentionally retaining PerfettoConfig as a TypedDict: it is nested in the legacy LoggerConfig TypedDict and mirrors the neighboring logger config blocks. Making only this nested block a BaseModel would introduce a mixed serialization and access contract without providing additional validation at this boundary.

tests/unit/utils/test_trace.py
youngeunkwon0405: Added all four suggested tests, plus regression coverage for equal-timestamp ordering and sync-PPO exception-path trace writing. The final focused PPO gate completed with 810 passed / 6 skipped; the refreshed stacked GRPO gate completed with 969 passed / 6 skipped.

nemo_rl/utils/timer.py
youngeunkwon0405: Added the required LIFO expectation to the Timer.stop() docstring in c0e184966.

nemo_rl/utils/trace.py
youngeunkwon0405: Removed the unused generic Tracer.collect_trace() in c0e184966. In stacked PR #3887, SyncRolloutActor owns the Ray timing/events RPC directly, which keeps transport behavior at the actor boundary.

youngeunkwon0405:
/ok to test [b15e4a7](https://github.com/NVIDIA-NeMo/RL/pull/3879/commits/b15e4a7a65753b20d060512564d077a2504b6961)

PR #2175, opened 2026-03-30
fix: aggregate rollout metrics by semantic type (min/max/sum/mean)

yfw:
Previously all numeric rollout metrics were averaged when aggregating across trajectory groups, giving misleading values for min/max/total metrics. Now min-prefixed/suffixed metrics use min(), max uses max(), total_turns is summed, and the rest are averaged.

Extracts `aggregate_rollout_metrics()` helper for testability and adds unit tests covering each aggregation path.

Closes https://github.com/NVIDIA-NeMo/RL/issues/1955

# What does this PR do ?

**Add a one line overview of what this PR aims to accomplish.**

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #3880, opened 2026-08-27
fix(sc): retain rollout metrics through training

yfw:
## Summary

Ports the rollout-metric aggregation behavior from #2175 to the SingleController async/TransferQueue path.

- retain per-group rollout metrics when prompt groups are committed to `TQReplayBuffer`
- carry metrics only from prompt groups selected by the sampler for the current training step
- aggregate selected metrics with the shared `aggregate_rollout_metrics` semantic rules and log them with the corresponding train step
- honor the existing W&B opt-in before retaining NeMo Gym full-result tables, keeping disabled payloads out of replay checkpoints

No new configuration or migration changes are required.

## Validation

- focused SingleController, replay-buffer, sampler, rollout-manager, setup, and aggregation tests: 151 passed
- `ruff check` and `ruff format --check`: passed
- `pyrefly check` on changed production files: passed


nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -875,6 +875,7 @@ async def commit(
                 sample_ids=list(sample_ids),
                 fields=list(fields.keys()),
                 sequence_lengths=[int(s) for s in lengths.tolist()],
+                extra_info={ROLLOUT_METRICS: [dict(record.rollout_metrics)]},
```
yfw: [`replay_buffer.py:878`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/async_utils/replay_buffer.py#L878)

**1 action item.**

**TL;DR — this line activates a full-result `wandb.Table` that `AsyncNemoGymRolloutImpl` builds ungated, so on an SC + NeMo-Gym run the complete Gym result payloads are logged and pickled into every `replay_buffer.pt` even when `logger.wandb.log_nemo_gym_full_result_tables` is false.**

Adjacent rather than PR-introduced — the offending line is pre-existing — but it was *dead* until this PR: on `main` the SC path discarded `record.rollout_metrics` entirely ([`single_controller.py`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/single_controller.py#L1401) mentions it only inside a TODO, and `staleness_sampler.py` not at all).

[`AsyncNemoGymRolloutImpl._compute_rollout_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/experience/rollout_manager.py#L1125) builds the table with no guard:

```python
rollout_metrics[f"{agent_name}/full_result"] = Table(
    data=[[json.dumps(r, separators=(",", ":"))] for r in agent_extras],
    columns=["Full result"],
)
```

Its twin at [`rollouts.py:2770`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/experience/rollouts.py#L2770) wraps the identical two lines in `if log_full_result_tables:`, and every other producer passes the gate — [sync_rollout_actor](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/experience/sync_rollout_actor.py#L252), [trajectory_collector](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1282), [grpo](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/grpo.py#L3085), [ppo](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/ppo.py#L1388), [distillation](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/distillation.py#L815). `rollout_manager.py` never imports [`should_log_nemo_gym_full_result_tables`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/utils/logger.py#L97). The flag is opt-in [because "These payloads can be very large, so the recommended default is false"](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/utils/logger.py#L54). Note the gate is `wandb_enabled and flag`, so the other paths skip the table even when W&B is off entirely; this path builds it regardless.

What this line now does with it:
1. this line stamps it into `extra_info`
2. [staleness_sampler.py:245](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/async_utils/staleness_sampler.py#L245) carries it onto the selected meta
3. [single_controller.py:1037](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/single_controller.py#L1037) collects it
4. [single_controller.py:1256](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/single_controller.py#L1256) feeds `aggregate_rollout_metrics`, whose non-numeric branch [grpo.py:4174](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/grpo.py#L4174) forwards the value untouched
5. it is not a `histogram/` key, so [`Logger.log_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/utils/logger.py#L1061) hands `train/<agent>/full_result` to the backends as a list of Tables

Independently of logging, the meta is snapshotted by [`TQReplayBuffer.state_dict`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/async_utils/replay_buffer.py#L1018) and [`torch.save`d to `replay_buffer.pt`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/single_controller.py#L1849) on every checkpoint. (Your new round-trip test is what proves `extra_info` survives that path.)

Reachable by config alone: [`should_use_nemo_gym`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/environments/nemo_gym.py#L85) reads `env.should_use_nemo_gym`, SC threads it at [setup.py:801](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/single_controller_utils/setup.py#L801), and [setup.py:954](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/single_controller_utils/setup.py#L954) actively tells users to set it (*"Set env.should_use_nemo_gym=true, or disable the router."*).

**Action:** gate it the way the other producers do — add `log_full_result_tables: bool = False` to `AsyncNemoGymRolloutImpl.__init__` and `RolloutManager.__init__`, forward it alongside the existing `effort_config=`, pass `log_full_result_tables=should_log_nemo_gym_full_result_tables(wandb_enabled=master_config.logger["wandb_enabled"], wandb_config=master_config.logger["wandb"])` from [setup.py:1109](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/single_controller_utils/setup.py#L1109), and wrap rollout_manager.py:1125 in `if self._log_full_result_tables:`. Spans four regions, so no suggestion block.

Filtering the key at the SC consumption point instead would be worse: it would also drop the legitimate `histogram/*` and `per_worker_token_counts` pass-throughs, and it would leave the table in every checkpoint.

One thing I did not check: what `wandb.Run.log` does with a `list[Table]` value. The finding does not rest on it — it rests on the documented opt-out being bypassed and the payload landing in every `replay_buffer.pt`.

tests/unit/single_controller/test_sampler_interface.py
```diff
@@ -63,11 +63,13 @@ def add(
         *,
         ready: bool = True,
         target_step: int | None = None,
+        rollout_metrics: dict | None = None,
     ) -> None:
         meta = KVBatchMeta(
             partition_id=self._partition_id,
             task_name=None,
             sample_ids=[f"{group_id}_g0"],
+            extra_info={"rollout_metrics": [dict(rollout_metrics or {})]},
```
yfw: [`test_sampler_interface.py:72`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/tests/unit/single_controller/test_sampler_interface.py#L72)

**1 action item.** Low priority — nothing is broken.

This is the only `KVBatchMeta(` construction in the file, and it now stamps `extra_info` unconditionally. So no sampler test produces a meta without the key, and the missing-key default at [staleness_sampler.py:242](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/async_utils/staleness_sampler.py#L242) is unexercised.

That default is a live safety valve rather than dead code: this PR does not touch `state_dict`/`load_state_dict`, and `load_state_dict` validates only the envelope/group keys and sizes — never `extra_info` — so a group checkpointed *before* this PR restores with `extra_info == {}` ([`field(default_factory=dict)`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/data_plane/interfaces.py#L194)) and reaches :242 on resume. Uncommon — there is no version stamp and the checkpoint is treated as a same-job artifact — but unguarded.

**Action:** make the fake stop stamping when no metrics were passed. One line, and it exercises the default across every existing `buf.add(...)` call that omits metrics — including the `ReadyFirst` and `InOrder` cases, not just `Windowed`.

I measured this exact text on a scratch copy: `ruff format --check` and `ruff check` clean, **49 passed** unchanged; then with the `.get(..., [])` default removed from staleness_sampler.py:242, the current fixture still gives **49 passed / 0 detection**, while this one gives **6 failed**, each `KeyError: 'rollout_metrics'`.

```suggestion
            extra_info=(
                {}
                if rollout_metrics is None
                else {"rollout_metrics": [dict(rollout_metrics)]}
            ),
```

tests/unit/single_controller/test_single_controller_actor.py
```diff
@@ -1366,6 +1366,57 @@ def test_train_pump_logs_nonzero_stale_group_metrics(monkeypatch) -> None:
...
+    train_call = ctrl._logger.log_metrics.call_args_list[0]
+    train_metrics = train_call.args[0]
+    assert train_metrics["gen_tokens/min"] == 3
+    assert train_metrics["gen_tokens/max"] == 20
+    assert train_metrics["total_turns"] == 7
+    assert train_metrics["accuracy"] == pytest.approx(0.5)
+    assert train_metrics["trajectory_duration_s"] == pytest.approx(2.0)
+    assert train_metrics["trajectory_duration_s/max"] == 3.0
+    assert train_metrics["trajectory_duration_s/p95"] == 3.0
+    assert train_call.kwargs == {"step": 1, "prefix": "train"}
```
yfw, line 1675: [`test_single_controller_actor.py:1417`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/tests/unit/single_controller/test_single_controller_actor.py#L1417)

**1 action item.** Low priority.

The `pop` at [single_controller.py:1038](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/algorithms/single_controller.py#L1038) is deliberate: [`shard_meta_for_dp`](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/data_plane/preshard.py#L136) copies `extra_info` verbatim into [every per-rank meta](https://github.com/NVIDIA-NeMo/RL/blob/a57599f10b0798bbc5386f8efeedc00bead12c11/nemo_rl/data_plane/preshard.py#L168), so leaving the key in would replicate the per-group payload to every DP rank. Nothing pins that choice — changing it to `.get(...)` leaves all 46 tests in this file green.

To be clear on severity: the mutant is **not** a correctness bug. No path re-consumes `train_meta`, so the metric values are identical either way. This just pins the hygiene property that keeps rollout-metric-shaped data off the trainer ranks.

Verified: passes as written, and goes red under the `.get` mutant with this line as the failing assertion — `_SequenceSampler.select` hands back the very objects the test holds, so the pop is observable by identity.

**Action:** one line appended to this test.

```suggestion
    assert train_call.kwargs == {"step": 1, "prefix": "train"}
    # pop, not get: shard_meta_for_dp copies extra_info onto every rank meta.
    assert all("rollout_metrics" not in meta.extra_info for meta in metas)
```

yfw (commented):
Rollout metrics reaching the SingleController train logs is a real gap: on `main` the SC path computed `record.rollout_metrics` and discarded it, with no other route to those numbers. The implementation matches its nearest analogs (`grpo.py:4845-4849`, `ppo.py:2432`) almost line for line and reuses the shared `aggregate_rollout_metrics` rather than reimplementing its min/max/sum/histogram semantics, which is the right call. The re-attach in `_finalize_selection` is genuinely necessary — `concat` keeps only the first meta's `extra_info` — and the `pop` targets a fresh concat-produced meta, so it cannot corrupt buffered state. Multi-chunk accumulation is correctly scoped, and resume from a pre-PR checkpoint degrades gracefully.

One inline comment is a real bug on an adjacent line this PR activates; the other two are low-priority test pins.

Minor, no action required: the TODO at `single_controller.py:1401-1403` still lists `rollout_metrics` as outstanding — this PR implements it. The `histogram log` item in that same TODO is a different thing (the explicit `log_histogram` calls that `grpo.py:2562-2582` makes for baseline/advantage) and is still outstanding.

_Generated by Claude Code_

nemo_rl/algorithms/async_utils/replay_buffer.py
yfw: Addressed in 14a80c8f1. `AsyncNemoGymRolloutImpl` and `RolloutManager` now default `log_full_result_tables` to false, SingleController setup derives it through the shared `should_log_nemo_gym_full_result_tables` gate, and the table is only constructed when enabled. Added coverage for default-off/opt-in behavior and setup propagation, so disabled full payloads no longer enter rollout metrics or replay checkpoints.

tests/unit/single_controller/test_sampler_interface.py
yfw: Addressed in 14a80c8f1. The sampler fake now leaves `extra_info` empty when no rollout metrics are supplied, so the existing no-metrics sampler cases exercise the missing-key compatibility path while the selected-metrics test still covers retention.

tests/unit/single_controller/test_single_controller_actor.py
yfw: Addressed in 14a80c8f1. The train-pump test now asserts that `rollout_metrics` is popped from every selected meta, pinning the invariant that this payload is removed before `shard_meta_for_dp` copies metadata to trainer ranks.

nemo_rl/algorithms/async_utils/replay_buffer.py
```diff
@@ -1130,6 +1130,7 @@ async def commit(
                     sample_ids=list(sample_ids),
                     fields=list(fields.keys()),
                     sequence_lengths=[int(s) for s in lengths.tolist()],
+                    extra_info={ROLLOUT_METRICS: [dict(record.rollout_metrics)]},
```
mehraakash, line 1145: Blocking: `commit()` now retains `record.rollout_metrics` in `KVBatchMeta.extra_info`, but this function canonicalizes all of `extra_info` and requires every mapping key to be a string. Native rollout metrics contain `per_worker_token_counts: dict[int, int]`, so buffered checkpoint creation fails.

The `fast_L1_Functional_Tests_SingleController` check reproduces this directly:

```text
TypeError: Replay metadata at groups[0].meta.extra_info.rollout_metrics[0].per_worker_token_counts has non-string key 0
```

Please rebase onto current `main` and preserve `_canonical_manifest_extra_info` from #3883, which excludes advisory rollout metrics from the replay identity digest. Please also retain a regression test covering checkpoint creation with integer-keyed rollout metrics.

nemo_rl/experience/rollout_manager.py
```diff
@@ -1156,10 +1158,11 @@ def _compute_rollout_metrics(
                 rollout_metrics.update(
                     calculate_single_metric(values, n, f"{agent_name}/{key}")
                 )
-        rollout_metrics[f"{agent_name}/full_result"] = Table(
-            data=[[json.dumps(r, separators=(",", ":"))] for r in agent_extras],
-            columns=["Full result"],
-        )
+        if self._log_full_result_tables:
```
mehraakash, line 1161: Please add a checkpoint serialization round-trip test with `log_nemo_gym_full_result_tables=True`. When enabled, this `wandb.Table` is retained inside rollout metrics and can remain in buffered replay metadata until training selects the group. The current tests verify construction and configuration propagation, but not the actual `torch.save`/`torch.load` path that checkpointing uses.

mehraakash (changes requested):
The full-result-table opt-in is useful, but the branch is no longer safe to merge as-is.

The rollout-metric retention and aggregation portion has already landed on `main` through #3883, together with speculative-decoding metrics and replay-manifest handling. This PR now conflicts with current `main`, and taking the PR versions during conflict resolution could regress those newer changes.

Please rebase and reduce this PR to the remaining unique change:

- propagate `log_nemo_gym_full_result_tables` into the SingleController rollout manager;
- construct full-result tables only when enabled;
- retain focused default-off, opt-in, setup-propagation, and checkpoint-serialization tests.

The current SingleController functional failure is a real checkpointing regression caused by hashing `rollout_metrics`, so I don’t think this is ready until that path is fixed and green.

nemo_rl/algorithms/async_utils/replay_buffer.py
yfw: Addressed in b49bd8f45

nemo_rl/experience/rollout_manager.py
yfw: Addressed in 782dff619

mehraakash (approved):

PR #2801, opened 2026-06-12
feat: add MTP (multi-token prediction) training support for nemotron

yfw:
# What does this PR do ?

Adds training-side Megatron Multi-Token Prediction support, rebased onto the Megatron-Bridge 79c8957 bump which includes needed Megatron-LM changes. Part of Nemotron 3 Super upstreaming.

- megatron_policy_worker: compute mtp_loss_mask from token_mask × sample_mask (with optional mtp_positive_only gating on advantages), set mtp_grad_scale_func for MTP loss scaling, and collect MTP metrics.
- data.py / train.py: plumb mtp_loss_mask through microbatch processing/packing into model_forward's loss_mask.
- common.py: get_mtp_metrics (per-layer MTP loss + acceptance rate).
- grpo.py / lm_policy.py: surface mtp/* metrics in training logs.

Scope: training only. The MTP inference path (vLLM drafter weight load / speculative-config handling) will be handled separately as part of the ultra upstreaming.

# Issues
List issues that this PR closes ([syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)):

# Usage
* **You can potentially add a usage example below**

```python
# Add a code snippet demonstrating how to use this
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](/NVIDIA-NeMo/RL/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you run the unit tests and functional tests locally? Visit our [Testing Guide](/NVIDIA-NeMo/RL/blob/main/docs/testing.md) for how to run tests
- [ ] Did you add or update any necessary documentation? Visit our [Document Development Guide](/NVIDIA-NeMo/RL/blob/main/docs/documentation.md) for how to write, build and test the docs.

# Additional Information
* ...

PR #3881, opened 2026-08-27
feat(megatron): support MTP in split training

yfw:
# What does this PR do ?

Ports the Megatron multi-token prediction (MTP) training lifecycle from #2801 into the SingleController split training path.

- Derives the MTP loss mask from the existing token and sample masks before microbatch packing.
- Leaves `mtp_grad_scale_func` unset so MCore inherits the optimizer's loss scale, including fp16 dynamic scaling.
- Normalizes detached MTP auxiliary gradients by the global valid-token count independently of the main loss denominator.
- Rejects nonzero attached-head MTP with a sequence-level main loss because that mixed backbone gradient cannot be corrected after split-step accumulation.
- Collects the independently clipped MTP gradient norm plus per-layer loss and acceptance metrics after the optimizer step.
- Propagates worker-reduced MTP metrics through TQPolicy to the training logger.
- Forwards model-owned sequence/MTP-mask packing and context-parallel slicing flags through split training.

No TQ schema or new user configuration fields are required.

# Issues

Follow-up to #2801.

- #3889 tracks support for attached-head MTP with a sequence-level main loss.
- #3888 tracks the analogous pre-existing MoE auxiliary-loss denominator issue.

# Usage

Existing MTP-enabled Megatron recipes now use the same MTP training behavior when driven through the SingleController split API.

**Pre checks**:
- [x] Added focused lifecycle and TQ propagation tests.
- [x] Ran the available focused unit tests and static checks locally.
- [x] No documentation update is needed because this restores parity for existing configuration.

# Additional Information

Validation:

- Focused unit tests: 11 passed, 1 skipped.
- Ruff lint and formatting checks passed.
- Repository pre-commit hooks, including Pyrefly, passed.
- Python compilation and git diff checks passed.

The MCore-specific state-machine tests were skipped locally because Transformer Engine is unavailable on this host. Installing the full MCore extras is also blocked here by the missing `infiniband/mlx5dv.h` system header.

Sample run with single controller and mtp training enabled:
Screenshot 2026-09-03 at 2 11 22 PM

wandb: https://wandb.ai/nvidia/nano35-rlvr-convergence/runs/nk99s0qe


nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1497,6 +1517,13 @@ def _train_microbatch_body(
         # hooks. The 1/N rescale happens once at finish.
         placeholder_n = torch.tensor(1.0, device="cuda")
 
+        # MTP is an auxiliary autograd branch, so Megatron needs an explicit
+        # scale matching the main loss. Each streaming chunk accumulates raw
+        # summed gradients (N=1); finish_train_step applies the true 1/N once
+        # to the complete main_grad buffer, including the MTP parameters.
+        if state["mtp_enabled"]:
+            self._set_mtp_grad_scale_func(lambda: placeholder_n)
```
yfw: **1 action item.**

**TL;DR — pinning the MTP scale to a constant `1.0` overrides Megatron's documented fallback, so under `policy.precision: float16` the MTP gradient alone ends up S× too small; leaving the hook unset is what actually implements the "scale matching the main loss" this comment describes.**

[`nemo_rl/models/policy/workers/megatron_policy_worker.py:1520-1525`](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1520-L1525) — PR-introduced, latent (bf16 unaffected).

### AI-1

Megatron documents the unset behaviour on the field itself — [`ModelParallelConfig.mtp_grad_scale_func`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/model_parallel_config.py#L205-L208):

> `mtp_grad_scale_func: Optional[Callable] = None`
> *"If using loss scaling for MTP ... **If None, falls back to `grad_scale_func` with `torch.ones(1)`.**"*

and <a href="https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/pipeline_parallel/schedules.py#L569-L570"><code>grad_scale_func</code></a> is exactly the factor the main loss backward is scaled by (`output_tensor[0] = config.grad_scale_func(output_tensor[0])`), bound to `optimizer.scale_loss` by Megatron-Bridge's [`_update_model_config_funcs`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/d352aceda8ed4136f1db787bcf449c1b210a2438/src/megatron/bridge/training/setup.py#L638). So `None` inherits the main loss's scale *by construction*; a constant `1.0` equals it only when the loss scale is 1.

Breaking case — `policy.precision: float16`, a supported branch with its own asserts ([setup.py:1353-1361](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/nemo_rl/models/megatron/setup.py#L1353-L1361)):

1. `scale_loss` returns the dynamic loss scale S — <a href="https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/optimizer/optimizer.py#L701-L707"><code>get_loss_scale</code></a> returns `grad_scaler.scale`.
2. The main loss backward is seeded with S, while the MTP branch is seeded with `mtp_grad_scale_func()` = 1.0 — <a href="https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L734-L748"><code>MTPLossAutoScaler.backward</code></a> returns `ones_like(mtp_loss) * scale`.
3. The optimizer unscales everything by 1/S, so the MTP contribution lands S× too small. No error — the MTP head just stops learning.

bf16 is unaffected (`grad_scaler is None` → `get_loss_scale()` returns `_scale_one` = 1.0), and all seven MTP-enabled recipes are bf16, so nothing shipped is broken today.

**Action:** leave the hook unset rather than pinning a constant.

```suggestion
        # Leave ``mtp_grad_scale_func`` unset: Megatron documents None as
        # "falls back to grad_scale_func with torch.ones(1)", so the MTP branch
        # inherits exactly the scale the main loss backward receives. Pinning a
        # constant 1.0 would drop the fp16 dynamic loss scale from the MTP
        # gradient only. Each streaming chunk still accumulates raw summed
        # gradients (N=1); finish_train_step applies the true 1/N once to the
        # complete main_grad buffer, including the MTP parameters.
        if state["mtp_enabled"]:
            self._set_mtp_grad_scale_func(None)
```

**Context — no action.** This flips two of the new tests: `test_builds_mtp_mask_and_sets_raw_gradient_scale` (asserts the scale is ≈1.0) and `test_clears_mtp_gradient_scale` (asserts non-`None` after `train_microbatch`). The first is folded into the suggestion on `tests/unit/models/policy/test_megatron_split_state.py` so the two edits don't collide. It also makes the clears at :1420, :1603 and :1890 redundant — but :1643 still earns its place, since `begin_train_step` → `finish_train_step` with zero `train_microbatch` calls is reachable. No need to remove any of them.

tests/unit/models/policy/test_megatron_split_state.py
```diff
@@ -427,6 +444,31 @@ def test_does_not_call_optimizer_step(self, mock_module_symbols):
...
+        w = _make_worker(LossType.TOKEN_LEVEL)
+        w.model.config.mtp_num_layers = 2
+        batch = _fake_batch()
+        w.begin_train_step(loss_fn=w._test_loss_fn)
+        w.train_microbatch(batch)
+
+        expected_mask = batch["token_mask"] * batch["sample_mask"].unsqueeze(-1)
+        torch.testing.assert_close(batch["mtp_loss_mask"], expected_mask)
+        mtp_scale = w.model.config.mtp_grad_scale_func()
+        assert float(mtp_scale.item()) == pytest.approx(1.0)
```
yfw: **1 action item.**

**TL;DR — `_fake_batch()` is all-ones, so this mask assertion passes for four semantically different implementations, including one that ignores `sample_mask` entirely.**

[`tests/unit/models/policy/test_megatron_split_state.py:451-460`](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/tests/unit/models/policy/test_megatron_split_state.py#L451-L460) — PR-introduced.

### AI-1

[`_fake_batch()`](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/tests/unit/models/policy/test_megatron_split_state.py#L265-L266) returns `sample_mask = torch.ones(8)` and `token_mask = torch.ones(8, 257)`, and `expected_mask` re-derives the production expression verbatim. Running the candidates against `assert_close` with that fixture:

| implementation | passes today |
|---|---|
| `token_mask * sample_mask.unsqueeze(-1)` (correct) | yes |
| `token_mask.clone()` | yes |
| `torch.ones_like(token_mask)` | yes |
| `token_mask * 1.0` (`sample_mask` ignored entirely) | yes |

The only variant it catches is a dropped `.unsqueeze(-1)`, and that surfaces as a broadcast `RuntimeError` rather than an assertion failure. Zeroing one token and one sample makes all three wrong variants fail — verified.

**Action:** perturb the fixture locally and assert positionally instead of re-deriving the formula. The scale assertion also flips to `is None`, pairing with the `megatron_policy_worker.py:1520` comment — both edits are in this one block so they cannot collide.

```suggestion
        w = _make_worker(LossType.TOKEN_LEVEL)
        w.model.config.mtp_num_layers = 2
        batch = _fake_batch()
        # _fake_batch is all-ones, so comparing against a re-derived
        # token_mask * sample_mask cannot tell the product apart from either
        # factor alone. Zero one token and one whole sample so the assertions
        # below discriminate. (_fake_batch builds fresh tensors per call, so
        # this does not leak into other tests.)
        batch["token_mask"][0, 5] = 0.0
        batch["sample_mask"][3] = 0.0
        w.begin_train_step(loss_fn=w._test_loss_fn)
        w.train_microbatch(batch)

        mask = batch["mtp_loss_mask"]
        assert mask.shape == (8, 257)
        assert float(mask[3].sum()) == 0.0  # masked sample zeroes its whole row
        assert float(mask[0].sum()) == 256.0  # masked token drops exactly one
        assert float(mask[1].sum()) == 257.0  # untouched row keeps every token
        # Pairs with the megatron_policy_worker.py change: the split path leaves
        # mtp_grad_scale_func unset so mcore falls back to grad_scale_func.
        assert w.model.config.mtp_grad_scale_func is None
```

**Context — no action.** Please run this before accepting: the PR's "10 passed, 1 skipped" is the whole of this module skipping ([`pytest.importorskip("megatron.bridge")`](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/tests/unit/models/policy/test_megatron_split_state.py#L61) plus `pytestmark = pytest.mark.mcore`), so the new MCore tests here have not executed anywhere yet. I could not run them either — `transformer-engine` will not build on this host — so the suggestion above is verified numerically in plain torch, not through pytest.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1605,6 +1638,9 @@ def _finish_train_step_body(self, state: dict[str, Any]) -> dict[str, Any]:
         # global mean grad; for reduce_scatter (dist-opt) it's the shard.
         # Either way, opt.step sees the right-normalized gradient.
         self.model.scale_gradients(inv_n)
+        # No more forward/backward calls remain in this step. Clear the
+        # callable before optimizer/scheduler/checkpoint state can serialize it.
+        self._set_mtp_grad_scale_func(None)
```
yfw, line 1689: **1 action item, 1 follow-up.**

**TL;DR — the uniform finish-time rescale normalizes the MTP branch by the main loss's denominator, so under a `SEQUENCE_LEVEL` loss the MTP gradient lands on the sequence count instead of the token count.**

[`nemo_rl/models/policy/workers/megatron_policy_worker.py:1641-1643`](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1641-L1643) — the MTP half is PR-introduced; the same defect on the MoE branch is pre-existing.

### AI-1

[`self.model.scale_gradients(inv_n)`](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1640) is uniform, so auxiliary branches inherit the main loss's denominator: [`inv_n`](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/nemo_rl/models/policy/workers/megatron_policy_worker.py#L1629-L1634) is `1/global_valid_toks` for `TOKEN_LEVEL` but `1/global_valid_seqs` for `SEQUENCE_LEVEL`.

MTP enters as a token **sum** — <a href="https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L734-L748"><code>MTPLossAutoScaler.backward</code></a> seeds every element of the `[b, s]` loss tensor with the scale, and <a href="https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L875-L888"><code>process_mtp_loss</code></a> documents the expected divide as `total_num_tokens` — so it needs `1/global_valid_toks`, which sync `train()` applies for every loss type at [:913-915](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/nemo_rl/models/policy/workers/megatron_policy_worker.py#L913-L915). Otherwise `mtp_loss_scaling_factor` scales with mean response length rather than being a constant.

`TOKEN_LEVEL` is already exactly equivalent between the two paths; the correction below is a no-op there.

**Action:** give the MTP branch the token denominator, and refuse the case where it cannot be given.

- **`mtp_detach_heads=True`** — mcore tags every MTP parameter <a href="https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L1727-L1730"><code>grad_norm_group='mtp'</code></a> (only under this flag), and the detaches at [:819-823](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L819-L823) / [:1144-1145](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L1144-L1145) / [:1855](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L1855) keep the MTP loss off every other parameter. Scale just those parameters' `main_grad` by `n_true / global_valid_toks` after the uniform rescale and before the reduce.
- **`mtp_detach_heads=False`** — the MTP gradient is summed into shared backbone parameters and cannot be separated, and the compensating factor is a ratio of globals that do not exist until the finish-time all-reduce. Raise in `begin_train_step` instead. This is also the case with no separate grad-norm group, so MTP is clipped jointly with the policy.

### Follow-up

Making `mtp_detach_heads=False` + `SEQUENCE_LEVEL` actually work is tracked in **#3889** — that issue covers only the deferred case, not the action item above.

**Context — no action.** The MoE auxiliary branch has the identical defect on this path, since `_set_moe_grad_scale_func` is only ever called inside sync `train()` ([:910](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/nemo_rl/models/policy/workers/megatron_policy_worker.py#L910), [:953](https://github.com/NVIDIA-NeMo/RL/blob/b0dd48137db9913db3196a13f69bbd1ea10572cb/nemo_rl/models/policy/workers/megatron_policy_worker.py#L953)). It is strictly harder — the aux gradient routes through the router and everything upstream, so no parameter subset carries only the aux term. Tracked in **#3888**.

yfw (commented):
The MTP lifecycle port looks sound. `scale_gradients` does reach the MTP parameters (it iterates `buffers + expert_parallel_buffers` and scales each whole buffer, with no MTP or expert exclusion), so the raw-accumulate-per-chunk then normalize-once-at-finish contract holds. The loss mask is load-bearing too — without it `process_mtp_loss` gets `loss_mask=None` and trains MTP over prompt and padding tokens.

Three inline comments: a latent fp16 issue in the gradient-scale hook, a test assertion that cannot currently fail, and one intent question about auxiliary-branch normalization that is pre-existing on this path.

_Generated by Claude Code_

nemo_rl/models/policy/workers/megatron_policy_worker.py
yfw: Fixed in 60c55c89e. For detached heads, finish now corrects only MTP-tagged main_grad views by n_true/global_valid_toks before final gradient reduction. Nonzero sequence-level MTP with attached heads now fails early because that mixed backbone gradient cannot be separated; zero-weight MTP remains allowed. #3889 continues to track full attached-head support, and the analogous pre-existing MoE case remains scoped to #3888.

nemo_rl/models/policy/workers/megatron_policy_worker.py
yfw: Fixed in 60c55c89e. The split step now clears mtp_grad_scale_func at begin and leaves it unset during forward/backward, so MCore falls back to optimizer.scale_loss and preserves the fp16 dynamic loss scale. The lifecycle test now verifies that the hook remains None through begin, train, and abort.

tests/unit/models/policy/test_megatron_split_state.py
yfw: Fixed in 60c55c89e. The test now zeros one token and one whole sample, asserts the affected rows positionally, and checks that mtp_grad_scale_func stays None so MCore uses the main loss-scale fallback.

nemo_rl/models/policy/workers/megatron_policy_worker.py
```diff
@@ -1455,6 +1497,15 @@ def _train_microbatch_body(
         state["local_valid_seqs"] = state["local_valid_seqs"] + call_local_seqs
         state["local_valid_toks"] = state["local_valid_toks"] + call_local_toks
 
+        # Match the synchronous Megatron path: derive the mask on the worker
+        # immediately before microbatch processing so sequence packing applies
+        # the same layout transformation to tokens and the MTP loss mask. The
+        # mask is worker-local derived data; it does not need a TQ schema field.
+        if state["mtp_enabled"] and "token_mask" in data and "sample_mask" in data:
+            data["mtp_loss_mask"] = data["token_mask"] * data["sample_mask"].unsqueeze(
```
mehraakash, line 1507: Could the `get_microbatch_iterator` call below forward the same model-owned packing arguments used by the synchronous training path?

```python
delegate_pack_to_model=self.delegate_pack_to_model,
delegate_mtp_loss_mask_to_model=self.delegate_mtp_loss_mask_to_model,
model_slices_context_parallel_inputs=self.model_slices_context_parallel_inputs,
```

Without them, the defaults are all `False`, which could affect models that own sequence packing or CP slicing—particularly the alignment of the newly added `mtp_loss_mask` with packed tokens. A small regression test for the delegated-packing case would also be useful.

mehraakash (commented):
The core MTP lifecycle and scaling logic looks consistent with the synchronous path: mask construction, fp16 loss-scale inheritance, detached-head token normalization, metric reduction, and TQ propagation all look sound.

I left one compatibility question about forwarding the model-owned packing and context-parallel flags through the split microbatch path.

Could we also run `CI:L1` against the exact current head and include a real MCore split-MTP functional test? The current head has only metadata, security, and submodule checks. The previous Lfast PPO failure used a non-MTP configuration and appears unrelated, but there is not yet a clean functional signal for this training-path change.

nemo_rl/models/policy/workers/megatron_policy_worker.py
yfw: Fixed in b93b84cfc. The split-training iterator now forwards delegate_pack_to_model, delegate_mtp_loss_mask_to_model, and model_slices_context_parallel_inputs, matching the synchronous training and log-probability paths. I also added regression coverage for both valid capability combinations: model-owned sequence/MTP-mask packing, and caller-owned packing with model-owned CP slicing.

yfw:
I triggered CI:L1 for the exact updated head, b93b84cfc3754a6e712675e8aed9291da9ddc024. I did not add a nominal MTP L1 script because the current SingleController L1 recipe uses Qwen3-0.6B with mtp_num_layers=0 and no MTP weights, so it would not exercise the behavior under review. The repository’s actual Megatron MTP functional recipe requires the private Super Omni checkpoint/training data and 16 nodes, and is currently disabled. The added MCore regression instead covers propagation for both supported model-owned packing/CP-slicing combinations; a true end-to-end CI lane will require a small CI-accessible MTP checkpoint.

yfw:
CI:L1 is explicitly dispatched against b93b84cfc3754a6e712675e8aed9291da9ddc024: https://github.com/NVIDIA-NeMo/RL/actions/runs/33715206186

mehraakash (approved):

PR #3882, opened 2026-08-27
fix(grpo): group advantages by rollout identity

aroshanghias-nvd:
# What does this PR do ?

Fix GRPO credit assignment by grouping completions by their rollout sampling identity instead of tokenized prompt equality.

## Problem

GRPO baselines are defined over the completions sampled together for one prompt instance. NeMo RL previously used the tokenized initial prompt as that instance's group identifier. Independent rollout groups were therefore merged whenever their initial token rows were identical.

This is especially likely for multimodal and stateful environments. Different images can share the same text and image-placeholder token sequence, while text-only environment instances can share initial instructions but differ in reward-relevant state such as their seed, tool state, or later observations.

For example, consider two independent 16-generation groups with identical initial token rows:

```text
Group A: image/state A, rewards [0, 0, ..., 0]
Group B: image/state B, rewards [1, 1, ..., 1]
```

The correct within-group RLOO advantages are zero for both groups. Token-based grouping instead forms one 32-sample group. A zero-reward sample receives a baseline of `16/31`, and a one-reward sample receives a baseline of `15/31`, creating false negative and positive advantages.

The issue is often not visible in normal training because most text datasets have unique prompt tokens. Even when collisions occur, pooling has no numerical effect if the colliding groups have the same reward distribution. A limited amount of corrupted credit assignment can also look like ordinary optimization noise rather than an obvious failure.

## Fix

- Assign explicit sampling-group IDs in legacy synchronous and asynchronous GRPO.
- Recover the existing group UUID from `{group_id}_g{generation_index}` sample IDs in TransferQueue and Single Controller paths.
- Preserve group identity through filtering, caching, concatenation, and slicing.
- Rebuild TransferQueue group IDs after dynamic-sampling caches from multiple generation batches are combined.
- Validate malformed, duplicate, and incomplete unfiltered groups while permitting legitimate partial groups after dynamic sampling.

This does not change the sample-ID or checkpoint schema; it consumes group identity already present in rollout metadata.

# Issues

None.

# Usage

No user-facing configuration or API changes are required.

**Pre checks**:
- [x] Added regression tests for the affected execution paths.
- [x] Ran 17 targeted unit tests in the project container.
- [x] No documentation update is required because this is an internal correctness fix with no user-facing behavior or configuration change.

# Additional Information

The regression coverage includes:

- identical token prompts belonging to different reward groups;
- rollout-group ID construction and validation;
- Single Controller advantage computation;
- legacy dynamic-sampling cache concatenation;
- TransferQueue caching across two generation batches whose temporary dense IDs both begin at zero;
- expected RLOO baselines after cached groups are separated correctly.

This fixes incorrect GRPO credit assignment. It does not claim to resolve entropy collapse or other policy-degeneration behavior generally.

Full repository CI remains the merge gate.

### DIFF
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -57,6 +57,7 @@ from nemo_rl.algorithms.reward_functions import (
     apply_reward_shaping,
 )
 from nemo_rl.algorithms.utils import (
+    build_rollout_group_ids,
     calculate_baseline_and_std_per_prompt,
     get_gdpo_reward_component_keys,
     log_generation_metrics,
@@ -2937,6 +2938,7 @@ def grpo_train(
         batch_cache: BatchedDataDict[DatumSpec] = None
         # This is the number of batches we processed so far at each step to generate responses whose std is non-zero. Maximum threshold is set by dynamic_sampling_max_gen_batches. Used in the case of dynamic sampling.
         dynamic_sampling_num_gen_batches = 0
+        next_rollout_group_id = 0
 
         # Run grpo/dapo training loop (single-turn)
         for batch in wrapped_dataloader:
@@ -3181,12 +3183,25 @@ def grpo_train(
                         and "unshaped_total_reward" in repeated_batch
                         else None
                     )
+                    reward_group_ids = build_rollout_group_ids(
+                        repeated_batch.size,
+                        master_config.grpo.num_generations_per_prompt,
+                        start_group_id=next_rollout_group_id,
+                    )
+                    next_rollout_group_id += (
+                        repeated_batch.size
+                        // master_config.grpo.num_generations_per_prompt
+                    )
+                    # Dynamic sampling may cache and concatenate survivors from
+                    # multiple generation batches. Carry the explicit identity
+                    # through those transformations instead of rebuilding it.
+                    repeated_batch["rollout_group_ids"] = reward_group_ids
                     if master_config.grpo.calculate_advantages_on_gpu:
                         print("Computing advantages on GPU!")
                         # Just fix the device id for now
                         device_id = 0
                         baseline, std = calculate_baseline_and_std_per_prompt(
-                            input_ids.cuda(device_id),
+                            reward_group_ids.cuda(device_id),
                             rewards.cuda(device_id),
                             torch.ones_like(rewards).cuda(device_id),
                             leave_one_out_baseline=master_config.grpo.use_leave_one_out_baseline,
@@ -3200,7 +3215,7 @@ def grpo_train(
                         std = std.cpu()
                     else:
                         baseline, std = calculate_baseline_and_std_per_prompt(
-                            input_ids,
+                            reward_group_ids,
                             rewards,
                             torch.ones_like(rewards),
                             leave_one_out_baseline=master_config.grpo.use_leave_one_out_baseline,
@@ -3243,23 +3258,12 @@ def grpo_train(
                     # Save baseline for logging (before deletion)
                     baseline_for_log = baseline.clone()
 
-                    # Must precede prompt extraction: it reuses the same message
-                    # dicts, so this also protects the prompt flatten below.
+                    # Backfill before flattening the full rollout for training.
                     backfill_missing_routed_experts(repeated_batch["message_log"])
 
-                    # Extract original prompt messages using the length field
-                    # This correctly handles multi-turn prompts that contain assistant messages
-                    initial_prompt_message_logs = extract_initial_prompt_messages(
-                        repeated_batch["message_log"],
-                        repeated_batch["length"],
-                    )
-                    prompt_batched_flat, _ = batched_message_log_to_flat_message(
-                        initial_prompt_message_logs,
-                        pad_value_dict={"token_ids": tokenizer.pad_token_id},
-                    )
-                    prompt_ids_for_adv = prompt_batched_flat["token_ids"]
-                    del initial_prompt_message_logs
-                    del prompt_batched_flat
+                    # Use the sampling group itself as the GRPO identity. Distinct
+                    # media-conditioned prompts can have identical text tokens.
+                    prompt_ids_for_adv = repeated_batch.pop("rollout_group_ids")
                     del input_ids
                     del baseline
                     del std
@@ -3892,6 +3896,7 @@ def grpo_train(
             # Reset the batch and set dynamic_sampling_num_gen_batches to 0
             batch_cache = None
             dynamic_sampling_num_gen_batches = 0
+            next_rollout_group_id = 0
 
             # Clear mem
             memory_tracker.snapshot_start_of_stage("After CPU memory clear", dir())
@@ -4877,24 +4882,14 @@ def async_grpo_train(
 
                 print("▶ Processing rewards...")
                 with timer.time("reward_calculation"):
-                    # Must precede prompt extraction: it reuses the same message
-                    # dicts, so this also protects the prompt flatten below.
+                    # Backfill before flattening the full rollout for training.
                     backfill_missing_routed_experts(repeated_batch["message_log"])
-
-                    # Extract original prompt messages using the length field
-                    # This correctly handles multi-turn prompts that contain assistant messages
-                    initial_prompt_message_logs = extract_initial_prompt_messages(
-                        repeated_batch["message_log"],
-                        repeated_batch["length"],
+                    # Each replay entry is one complete prompt group and the
+                    # concatenation above preserves group-contiguous ordering.
+                    prompt_ids_for_adv = build_rollout_group_ids(
+                        repeated_batch.size,
+                        master_config.grpo.num_generations_per_prompt,
                     )
-
-                    prompt_batched_flat, _ = batched_message_log_to_flat_message(
-                        initial_prompt_message_logs,
-                        pad_value_dict={"token_ids": tokenizer.pad_token_id},
-                    )
-                    prompt_ids_for_adv = prompt_batched_flat["token_ids"]
-                    del initial_prompt_message_logs
-                    del prompt_batched_flat
 
                     rewards = repeated_batch["total_reward"]
 
--- a/nemo_rl/algorithms/grpo.py
+++ b/nemo_rl/algorithms/grpo.py
@@ -2992,12 +2992,6 @@ def grpo_train(
                             enabled=master_config.grpo.debug_payload_metrics,
                         )
                     )
-                    # Convert LLMMessageLogType to FlatMessagesType for generation
-                    batched_flat, input_lengths = batched_message_log_to_flat_message(
-                        repeated_batch["message_log"],
-                        pad_value_dict={"token_ids": tokenizer.pad_token_id},
-                    )
-                    input_ids = batched_flat["token_ids"]
 
                 # Generate responses - this updates the LLMMessageLogType in repeated_batch
                 memory_tracker.snapshot_start_of_stage("Generation", dir())
@@ -3103,7 +3097,6 @@ def grpo_train(
                                 master_config.grpo.debug_payload_metrics
                             ),
                         )
-                        input_ids = nemo_gym_rollout_result.input_ids
                         repeated_batch = nemo_gym_rollout_result.final_batch
                         rollout_metrics = nemo_gym_rollout_result.rollout_metrics
                         del nemo_gym_rollout_result
@@ -3264,7 +3257,6 @@ def grpo_train(
                     # Use the sampling group itself as the GRPO identity. Distinct
                     # media-conditioned prompts can have identical text tokens.
                     prompt_ids_for_adv = repeated_batch.pop("rollout_group_ids")
-                    del input_ids
                     del baseline
                     del std
 
--- a/nemo_rl/algorithms/grpo_sync.py
+++ b/nemo_rl/algorithms/grpo_sync.py
@@ -234,8 +234,8 @@ def _apply_dynamic_sampling(
     # Rebuild them after cache concatenation so groups from different generation
     # batches cannot collide. Dynamic sampling may retain only part of a group,
     # so validate the sample-ID format without requiring a complete group here.
-    pending_carry["prompt_ids_for_adv"] = (
-        build_rollout_group_ids_from_sample_ids(pending_meta.sample_ids)
+    pending_carry["prompt_ids_for_adv"] = build_rollout_group_ids_from_sample_ids(
+        pending_meta.sample_ids
     )
 
     unfiltered_for_log = torch.cat(pending_unfiltered_rewards)[:train_prompts_size]
--- a/nemo_rl/algorithms/grpo_sync.py
+++ b/nemo_rl/algorithms/grpo_sync.py
@@ -66,6 +66,7 @@ from nemo_rl.algorithms.loss import (
 from nemo_rl.algorithms.loss.interfaces import LossFunction
 from nemo_rl.algorithms.reward_functions import apply_reward_shaping
 from nemo_rl.algorithms.utils import (
+    build_rollout_group_ids_from_sample_ids,
     calculate_baseline_and_std_per_prompt,
     get_gdpo_reward_component_keys,
     log_generation_metrics,
@@ -228,6 +229,14 @@ def _apply_dynamic_sampling(
         ds_metrics["dynamic_sampling_num_discarded_valid_samples"] = (
             n - train_prompts_size
         )
+
+    # Each generation batch initially assigns dense group IDs starting at zero.
+    # Rebuild them after cache concatenation so groups from different generation
+    # batches cannot collide. Dynamic sampling may retain only part of a group,
+    # so validate the sample-ID format without requiring a complete group here.
+    pending_carry["prompt_ids_for_adv"] = (
+        build_rollout_group_ids_from_sample_ids(pending_meta.sample_ids)
+    )
 
     unfiltered_for_log = torch.cat(pending_unfiltered_rewards)[:train_prompts_size]
     return pending_meta, pending_carry, [], True, ds_metrics, unfiltered_for_log
@@ -688,6 +697,14 @@ def grpo_train_sync(
                 # now back on the driver where they belong (no bulk
                 # touched by any of these ops).
                 with timer.time("reward_calculation"):
+                    # TQ sample IDs encode the rollout-group UUID. Use that
+                    # explicit identity instead of media-blind prompt tokens.
+                    driver_carry["prompt_ids_for_adv"] = (
+                        build_rollout_group_ids_from_sample_ids(
+                            meta.sample_ids,
+                            expected_group_size=master_config.grpo.num_generations_per_prompt,
+                        )
+                    )
                     driver_carry = scale_rewards(
                         driver_carry,
                         master_config.grpo.reward_scaling,
--- a/nemo_rl/algorithms/single_controller.py
+++ b/nemo_rl/algorithms/single_controller.py
@@ -76,6 +76,7 @@ from nemo_rl.algorithms.single_controller_utils.utils import (
     squeeze_trailing_unit_dim,
     tensor_field,
 )
+from nemo_rl.algorithms.utils import build_rollout_group_ids_from_sample_ids
 from nemo_rl.data.interfaces import DatumSpec
 from nemo_rl.data_plane import KVBatchMeta
 from nemo_rl.data_plane.schema import DP_CALIB_INPUT_FIELDS
@@ -1971,7 +1972,12 @@ class SingleControllerActor:
             select_fields=self._advantage_input_fields(),
         )
 
-        prompt_ids = tensor_field(data, adv_cfg.prompt_ids_field)
+        # The selected metadata retains each prompt group's UUID in its sample
+        # IDs. Text-token equality is insufficient for multimodal prompts.
+        prompt_ids = build_rollout_group_ids_from_sample_ids(
+            meta.sample_ids,
+            expected_group_size=self._algo_cfg.num_generations_per_prompt,
+        )
         rewards = squeeze_trailing_unit_dim(
             tensor_field(data, adv_cfg.reward_field)
         ).float()
@@ -2110,7 +2116,6 @@ class SingleControllerActor:
     def _advantage_input_fields(self) -> list[str]:
         adv_cfg = self._advantage_cfg
         fields = [
-            adv_cfg.prompt_ids_field,
             adv_cfg.reward_field,
             adv_cfg.token_mask_field,
             adv_cfg.sample_mask_field,
--- a/nemo_rl/algorithms/utils.py
+++ b/nemo_rl/algorithms/utils.py
@@ -123,9 +123,7 @@ def build_rollout_group_ids(
             f"group_size={group_size}"
         )
     if start_group_id < 0:
-        raise ValueError(
-            f"start_group_id must be non-negative, got {start_group_id}"
-        )
+        raise ValueError(f"start_group_id must be non-negative, got {start_group_id}")
     num_groups = batch_size // group_size
     return torch.arange(
         start_group_id,
--- a/nemo_rl/algorithms/utils.py
+++ b/nemo_rl/algorithms/utils.py
@@ -101,6 +101,86 @@ def calculate_kl(
     return kl
 
 
+def build_rollout_group_ids(
+    batch_size: int,
+    group_size: int,
+    *,
+    start_group_id: int = 0,
+    device: torch.device | str | None = None,
+) -> torch.Tensor:
+    """Return an explicit identifier for every rollout's sampling group.
+
+    GRPO baselines are defined over the completions sampled together for one
+    prompt instance. Prompt token IDs are not a valid group identifier for
+    multimodal prompts because different media can have identical text and
+    therefore identical token rows.
+    """
+    if group_size <= 0:
+        raise ValueError(f"group_size must be positive, got {group_size}")
+    if batch_size < 0 or batch_size % group_size != 0:
+        raise ValueError(
+            f"batch_size={batch_size} must be non-negative and divisible by "
+            f"group_size={group_size}"
+        )
+    if start_group_id < 0:
+        raise ValueError(
+            f"start_group_id must be non-negative, got {start_group_id}"
+        )
+    num_groups = batch_size // group_size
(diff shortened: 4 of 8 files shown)


nemo_rl/algorithms/utils.py
```diff
@@ -101,6 +101,86 @@ def calculate_kl(
...
+        raise ValueError(f"group_size must be positive, got {group_size}")
+    if batch_size < 0 or batch_size % group_size != 0:
+        raise ValueError(
+            f"batch_size={batch_size} must be non-negative and divisible by "
+            f"group_size={group_size}"
+        )
+    if start_group_id < 0:
+        raise ValueError(
+            f"start_group_id must be non-negative, got {start_group_id}"
+        )
```
aroshanghias-nvd: **1 action item.** `ruff format` (pinned v0.9.9) collapses this to one line.

```suggestion
        raise ValueError(f"start_group_id must be non-negative, got {start_group_id}")
```

Only offending hunk in this file.

nemo_rl/algorithms/grpo_sync.py
```diff
@@ -229,6 +230,14 @@ def _apply_dynamic_sampling(
             n - train_prompts_size
         )
 
+    # Each generation batch initially assigns dense group IDs starting at zero.
+    # Rebuild them after cache concatenation so groups from different generation
+    # batches cannot collide. Dynamic sampling may retain only part of a group,
+    # so validate the sample-ID format without requiring a complete group here.
+    pending_carry["prompt_ids_for_adv"] = (
+        build_rollout_group_ids_from_sample_ids(pending_meta.sample_ids)
+    )
```
aroshanghias-nvd: **1 action item.** `ruff format` drops the redundant parens and moves the argument to its own line.

```suggestion
    pending_carry["prompt_ids_for_adv"] = build_rollout_group_ids_from_sample_ids(
        pending_meta.sample_ids
    )
```

Only offending hunk in this file.

tests/unit/algorithms/test_grpo.py
```diff
@@ -2231,6 +2231,9 @@ def test_dapo_dynamic_sampling_batch_caching(mock_grpo_components):
 
     repeated_batch = create_mock_batch(batch_size, task_names, message_logs)
     repeated_batch["total_reward"] = torch.tensor([1.0, 0.0, 0.5])  # Non-zero std
+    repeated_batch["rollout_group_ids"] = torch.zeros(
+        (batch_size, 1), dtype=torch.long
+    )
```
aroshanghias-nvd: **1 action item.** `ruff format` collapses this to one line.

```suggestion
    repeated_batch["rollout_group_ids"] = torch.zeros((batch_size, 1), dtype=torch.long)
```

Same fix needed at `2279-2281` in this file (`torch.ones(...)`); `uv run ruff format .` does both.

tests/unit/algorithms/test_utils.py
```diff
@@ -35,6 +37,63 @@
...
 
+def test_rollout_group_ids_keep_identical_multimodal_prompts_separate():
+    """Different media states must not share a GRPO reward baseline."""
+    group_ids = build_rollout_group_ids(batch_size=32, group_size=16)
+    rewards = torch.cat((torch.ones(16), torch.zeros(16)))
+    valid_mask = torch.ones_like(rewards)
+
+    baseline, _ = calculate_baseline_and_std_per_prompt(
+        group_ids, rewards, valid_mask
+    )
```
aroshanghias-nvd: **1 action item.** `ruff format` collapses this to one line.

```suggestion
    baseline, _ = calculate_baseline_and_std_per_prompt(group_ids, rewards, valid_mask)
```

Two more in this file at `79-81` and `92-94`; `uv run ruff format .` does all three.

nemo_rl/algorithms/grpo.py
```diff
@@ -3243,23 +3258,12 @@ def grpo_train(
...
-                        initial_prompt_message_logs,
-                        pad_value_dict={"token_ids": tokenizer.pad_token_id},
-                    )
-                    prompt_ids_for_adv = prompt_batched_flat["token_ids"]
-                    del initial_prompt_message_logs
-                    del prompt_batched_flat
+                    # Use the sampling group itself as the GRPO identity. Distinct
+                    # media-conditioned prompts can have identical text tokens.
+                    prompt_ids_for_adv = repeated_batch.pop("rollout_group_ids")
                     del input_ids
```
aroshanghias-nvd: **1 action item.** This `del` is now the only reader of `input_ids`, and that makes an upstream flatten dead work.

Before this PR, `input_ids` was the GRPO group key — it was the first argument to both `calculate_baseline_and_std_per_prompt` calls. This PR replaces both with `reward_group_ids`, which leaves the whole chain unreferenced:

1. `grpo.py:2996` — `batched_flat, input_lengths = batched_message_log_to_flat_message(...)` over the **full** pre-generation message log
2. `:3000` — `input_ids = batched_flat["token_ids"]`
3. `:3106` — reassigned on the NeMo-Gym branch
4. `:3267` — this line, the only remaining reference

`batched_flat` has no other reader, and `input_lengths` from `:2996` is unconditionally reassigned at `:3291` before its first read. The FP8 calibration branch builds its own `calib_flat`/`calib_input_lengths`.

So every generation batch — and every *extra* dynamic-sampling generation batch — pays a full flatten and materialises an unused `[B, S_prompt]` int64 tensor (~16 MB at B=512, S=4096). On VLM recipes it also re-runs `PackedTensor.flattened_concat` over every image in the batch.

**Action:** delete `:2995-3000`, `:3106`, and this line. `batched_message_log_to_flat_message` and its callee were read end-to-end and are side-effect-free over the message dicts (fresh lists, `torch.cat`/`torch.stack`, `_pad_tensor` returns a new tensor), so removing the call changes nothing else.

⚠️ Deleting `2995-3000` **without** also deleting this line raises `NameError: input_ids` on every non-NeMo-Gym path.

aroshanghias-nvd (commented):
Self-review of my own PR, run with this repo's `/review-pr-team` skill (rl-expert, bug-finder, test-agent, design-reviewer, comment-reviewer, devil-advocate). Posting the surviving findings for transparency. The adversarial pass cut or downgraded 12 of 15 clusters and disputed 1 outright, including two of my own — what's left is what survived it.

**2 action items, 3 follow-ups.**

### AI-1 — `ruff format` is not clean on 4 of the 8 changed files

Every offending hunk is a line this PR adds:

| File | Lines |
|---|---|
| `nemo_rl/algorithms/utils.py` | 126-128 |
| `nemo_rl/algorithms/grpo_sync.py` | 237-239 |
| `tests/unit/algorithms/test_grpo.py` | 2234-2236, 2279-2281 |
| `tests/unit/algorithms/test_utils.py` | 46-48, 79-81, 92-94 |

Reproduced three times independently, twice at the v0.9.9 pinned in `.pre-commit-config.yaml`. `ruff check` passes; only `format` fails. **Action:** `uv run ruff format .`. Inline suggestions cover the first hunk in each file.

This is invisible right now because `Request NVSkills CI` is **skipped**, not passing — the `copy-pr-bot` gate means the lint job has never run on this branch.

### AI-2 — this PR orphans a full-batch prompt flatten in `grpo_train`

`input_ids` *was* the GRPO group key. Both consumers are now `reward_group_ids`, so `grpo.py:2995-3000` builds an unused `[B, S_prompt]` tensor every generation batch. Details and the required deletion set inline — note it is **not** a single-block delete.

### Follow-up 1 — attach `CI:L2`

`labels` is currently empty. The `cicd` skill maps "changes that could affect convergence" to `CI:L2`, which this is by definition — it changes the GRPO baseline denominator.

### Follow-up 2 — the "usually not visible" framing in the description is too narrow

The description argues collisions mainly matter for multimodal and stateful environments. That understates it: **any text-only dataset containing duplicate prompt strings** previously pooled those rows into one baseline group and will now get separate per-instance baselines. Advantages change on those runs too, so this is not a text-recipe no-op.

Worth one cheap number in the description instead of a full A/B: for one step on one text and one VLM recipe, report `len(torch.unique(prompt_ids_for_adv, dim=0))` against `num_prompts_per_step`. Equal for the first, strictly less for the second, is the whole empirical case for the PR.

### Follow-up 3 — `prompt_ids_for_adv` is now a write-only column

Removing the last readers (`_advantage_input_fields`, and the overwrite at `grpo_sync.py:702`) leaves `payload.py:91`, `sync_rollout_actor.py:363` and `data_plane/schema.py:54` still producing and shipping the column, and `AdvantageConfig.prompt_ids_field` with zero readers repo-wide. Retaining the write is defensible for a correctness-only PR that promises no schema change — but a follow-up to drop the column, or a comment saying why it stays, would save the next reader from re-deriving that it is dead. Not asking for it here.

---

**Context — no action.**

The core fix was checked rather than assumed. Invariants confirmed independently by more than one agent:

- **Row order is preserved on all three legacy rollout branches**, which positional grouping now depends on and token grouping did not: the sync path re-sorts by original index (`rollouts.py:836-860`), the async path relies on `asyncio.gather` ordering (`:1573`), and the NeMo-Gym path returns in input-row order (`:2587`).
- `repeated_batch` is `batch.repeat_interleave(G)`, so it is group-contiguous and divisible at every `build_rollout_group_ids` call site.
- `rollout_group_ids` survives `select_indices` / `from_batches` / `slice`, and `next_rollout_group_id` resets in lockstep with `batch_cache`, so cached generation batches cannot collide.
- Multimodal dedup does not corrupt `repeated_batch.size` — `PackedTensor.__len__` returns logical rows.
- `[B,1]` is correct at every consumer, including the `calculate_advantages_on_gpu` branch where group IDs, rewards and mask all move to the same device together.
- SC cannot see a partial group: dynamic sampling and overlong filtering are rejected outright at `single_controller_utils/config.py:739`.
- Sample-ID producers all mint `f"{uuid4()}_g{i}"`, and uuid4 contains no underscore, so `rsplit("_g", 1)` is unambiguous.

`test_apply_dynamic_sampling_rebuilds_group_ids_across_generation_batches` is the one test here that genuinely fails on the old behaviour.

Two design notes worth recording: carrying identity as a batch key in legacy `grpo.py` rides `select_indices`/concat for free — the TQ path has to *rebuild* IDs at `grpo_sync.py:237` precisely because it lacks that — and keeping both new helpers as pure functions is what lets the bug be pinned in six lines with no fixtures.

Merge is currently blocked by the `copy-pr-bot` runner gate, which needs a vetter and is not self-serviceable. All checks that did run (DCO, copyright, secrets-detector, submodule fast-forward, semantic title) pass.

_Findings dropped after adversarial review: a sample-ID-grammar refactor (its motivating scenario turned out to be invented — the redispatch path reuses the same `group_id`), two suggested tests (one cosmetic, one whose `pytest.raises(match=...)` would have failed on arrival — `[0, 1]` is a regex character class), a docstring-completeness ask (`D417` is in this repo's ruff ignore list), and an exception-type nit (the friendly error already fires, via the unpack `ValueError`)._

nemo_rl/algorithms/utils.py
aroshanghias-nvd: Fixed in 15081250 (`ruff format`, v0.9.9). `ruff format --check` and `ruff check` both clean at fc1a3c19.

nemo_rl/algorithms/grpo_sync.py
aroshanghias-nvd: Fixed in 15081250 (`ruff format`, v0.9.9). `ruff format --check` and `ruff check` both clean at fc1a3c19.

tests/unit/algorithms/test_grpo.py
aroshanghias-nvd: Fixed in 15081250 (`ruff format`, v0.9.9). `ruff format --check` and `ruff check` both clean at fc1a3c19.

tests/unit/algorithms/test_utils.py
aroshanghias-nvd: Fixed in 15081250 (`ruff format`, v0.9.9). `ruff format --check` and `ruff check` both clean at fc1a3c19.

nemo_rl/algorithms/grpo.py
aroshanghias-nvd: Fixed in fc1a3c19 — removed the flatten at `2995-3000`, the NeMo-Gym branch assignment at `:3106`, and this `del` together, so no `NameError`.

Verified in the CI container (`rl-gym.63363213.sqsh`) rather than by inspection: ran the PR's four touched test files against the PR head and against the fixed tree, with `PYTHONPATH` pinned to each worktree. **242 passed in both**, no failures either side (job 16811398).

nemo_rl/algorithms/single_controller.py
```diff
@@ -2110,7 +2116,6 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> tuple[KVBatchMeta, bool]:
     def _advantage_input_fields(self) -> list[str]:
         adv_cfg = self._advantage_cfg
         fields = [
-            adv_cfg.prompt_ids_field,
             adv_cfg.reward_field,
```
yfw, line 2119: [`single_controller.py:2119`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/single_controller.py#L2119)

**1 action item.** *(PR-caused, but the edits land in adjacent files.)*

**TL;DR — dropping `prompt_ids_field` here removes the last reader of `prompt_ids_for_adv`, leaving a column that every producer still writes and nobody ever reads.**

Three pieces, all downstream of this one deletion:

1. **The SC data-plane column.** [`payload.py:91`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/experience/payload.py#L91) still jagged-packs `prompt_ids_for_adv` into every rollout group's TQ put, and it is still registered in `SC_ROLLOUT_SCHEMA_FIELDS` at [`schema.py:54`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/data_plane/schema.py#L54). With this line gone, `_advantage_input_fields()` never fetches it — a fully-written, never-read per-row column.
2. **The sync actor's flatten.** [`sync_rollout_actor.py:93`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/experience/sync_rollout_actor.py#L93) runs `extract_initial_prompt_messages` + `batched_message_log_to_flat_message` every generation batch, ships the result across the Ray boundary at [`sync_rollout_actor.py:363`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/experience/sync_rollout_actor.py#L363), and [`grpo_sync.py:702`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/grpo_sync.py#L702) overwrites it before any read. Nothing touches `driver_carry` in between, it is not in `bulk_batch`, and validation (`carry_keys=["total_reward", ...]`) pays the flatten and discards it too.
3. **The knob.** `AdvantageConfig.prompt_ids_field` now has exactly one reference repo-wide — its own definition at [`config.py:1065`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/single_controller_utils/config.py#L1065). It went 2 readers to 0 in this PR. Setting it now silently does nothing.

Your Follow-up 3 covers (1) and reasonably defers it as a schema change. **The part that isn't covered is (2)**: `driver_carry` is a Ray return value, not a data-plane column, so it carries no schema commitment — it is the same orphan you already deleted from `grpo.py` in `fc1a3c193`, just on the TQ actor. Note it is not a single-block delete: you'd also narrow the return type at `sync_rollout_actor.py:76`, drop the now-unused `prompt_lengths` param and the `extract_initial_prompt_messages` local import, and update the `:210` docstring.

**Action:** delete the `prompt_ids_for_adv` production in `_flatten_rollout_message_log_for_tq` / `rollout_to_tq`, and delete `AdvantageConfig.prompt_ids_field` (a plain internal `@dataclass`, constructed no-args, no YAML or checkpoint path — deletion is safe). Leave the schema column for the follow-up, but consider a one-line comment at `config.py:1065` so the next reader doesn't re-derive that it's dead.

nemo_rl/algorithms/grpo.py
```diff
@@ -3181,12 +3176,25 @@ def grpo_train(
                         and "unshaped_total_reward" in repeated_batch
                         else None
                     )
+                    reward_group_ids = build_rollout_group_ids(
+                        repeated_batch.size,
+                        master_config.grpo.num_generations_per_prompt,
+                        start_group_id=next_rollout_group_id,
```
yfw: [`grpo.py:3182`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/grpo.py#L3182)

**1 action item.** *(PR-introduced.)*

**TL;DR — mutation testing says both `grpo.py` call sites are unpinned: deleting `start_group_id=next_rollout_group_id` leaves the entire unit suite green.**

We ran 13 mutations against the real source (control: 248 passed). Five were killed by your new tests; **five survived**, and two of those are the `grpo.py` half of this diff:

| Mutation | Result |
|---|---|
| drop `start_group_id=next_rollout_group_id` (this line) | no test fails |
| `async_grpo_train`'s `build_rollout_group_ids` to all zeros ([`grpo.py:4881`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/grpo.py#L4881)) | no test fails |

To be precise about what *is* covered: `build_rollout_group_ids`'s `start_group_id` **mechanism** is pinned by `test_rollout_group_ids_support_namespacing_across_cached_batches`. What's unpinned is the **call-site wiring** — that `grpo_train` actually passes the counter. That wiring is the entire mechanism stopping two dynamic-sampling generation batches from both starting at group 0, i.e. re-introducing exactly the bug class this PR fixes, silently and with no error.

The new `test_dapo_dynamic_sampling_batch_caching` assertion doesn't cover it — `dynamic_sampling` has no hunk in this PR, so that assertion runs byte-identical pre-PR code. It's still a worthwhile regression guard for the *new* dependency on `from_batches` carrying a 2-D int column, just not evidence this change works.

**Action:** add a short test that drives `grpo_train` through the existing `mock_grpo_components` fixture with `use_dynamic_sampling=True`, spies `calculate_baseline_and_std_per_prompt`, and asserts the second generation batch sees group id `[[1]]` rather than `[[0]]`. The nearest analog for style and length is `test_async_grpo_propagates_main_loop_collector_failure` (~43 lines, 2 patches) — we drafted a much larger monkeypatch-heavy version and are deliberately not proposing it, since it doesn't match this file's conventions.

nemo_rl/algorithms/utils.py
```diff
@@ -101,6 +101,84 @@ def calculate_kl(
...
+                raise ValueError(
+                    f"Rollout group {group_id!r} has generation indices "
+                    f"{sorted(generation_indices)}, expected "
+                    f"{list(range(expected_group_size))}"
+                )
+    return torch.tensor(group_indices, device=device, dtype=torch.long)[:, None]
+
+
 def calculate_baseline_and_std_per_prompt(
     prompts: torch.Tensor,
```
yfw, line 183: [`utils.py:183`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/utils.py#L183)

**1 action item.** *(PR-introduced — this PR is what makes the line below wrong.)*

The docstring of the parameter you're redefining still says the opposite of this PR's thesis. At [`utils.py:194`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/utils.py#L194):

```
    prompts:    tensor (b, s)     Tensor of prompts the model used. May be on any device
```

That's the one line in the tree still asserting prompt token rows are the group key — eight lines below the new helper whose whole purpose is to replace them. Highest-probability spot for the next reader to reintroduce the bug.

**Action:** reword line 194 to describe a group key rather than prompts:

```
    prompts:    tensor (b, k)     Per-row group key; rows with an identical key row
                                  share a baseline. Production passes a (b, 1)
                                  rollout-group-id column (see build_rollout_group_ids).
                                  May be on any device
```

No `suggestion` block because line 194 sits just outside this hunk (it ends at 184).

Deliberately **not** asking you to rename the `prompts` parameter — it's a misnomer now, but all six call sites pass it positionally and it reaches untouched PPO files. And the wording above stays generic on purpose: this helper legitimately accepts any `(b, k)`, and 8 existing tests in `test_utils.py` pass `(6, 3)`.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1971,7 +1972,12 @@ async def _advantage_stage(self, meta: KVBatchMeta) -> tuple[KVBatchMeta, bool]:
             select_fields=self._advantage_input_fields(),
         )
 
-        prompt_ids = tensor_field(data, adv_cfg.prompt_ids_field)
+        # The selected metadata retains each prompt group's UUID in its sample
+        # IDs. Text-token equality is insufficient for multimodal prompts.
+        prompt_ids = build_rollout_group_ids_from_sample_ids(
```
yfw, line 1977: [`single_controller.py:1977`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/single_controller.py#L1977)

**1 action item** — low severity. *(PR-introduced.)*

This runs **unconditionally, before the `_is_ppo` branch**, so it fires on paths that have no use for the value:

- PPO runs reach it, but their only reachable estimators are `gae` / `raw_reward` ([`ppo.py:1135`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/ppo.py#L1135) restricts the factory; `GAEConfig.name` is `Literal["gae", "raw_reward"]`), and **neither reads `prompt_ids` at all** — we checked by AST rather than docstring, since `**kwargs` hides unused params. Zero load references in either body.
- It also runs when `has_valid_training_tokens` is False and `compute_advantage` is never called.

So on those paths this is both an unconditional cost and, more importantly, an unconditional **new** `ValueError` — the strict `expected_group_size` check can now fail a run that previously trained fine, for a value that would have been discarded.

To be clear this is a robustness nit, not a live bug: we could not construct a config where a real SC run produces incomplete groups (the sole SC producer mints `f"{group_id}_g{i}"` at [`payload.py:126`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/experience/payload.py#L126), each `PromptGroupRecord` holds exactly `num_generations_per_prompt` completions, and [`config.py:736`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/single_controller_utils/config.py#L736) rejects dynamic sampling and overlong filtering outright for both PPO and GRPO).

**Action:** move the `build_rollout_group_ids_from_sample_ids(...)` call so it only runs when the estimator actually consumes `prompt_ids` — or at minimum below the early-return for `has_valid_training_tokens`.

nemo_rl/algorithms/utils.py
```diff
@@ -101,6 +101,84 @@ def calculate_kl(
...
+        raise ValueError(f"group_size must be positive, got {group_size}")
+    if batch_size < 0 or batch_size % group_size != 0:
+        raise ValueError(
+            f"batch_size={batch_size} must be non-negative and divisible by "
+            f"group_size={group_size}"
+        )
+    if start_group_id < 0:
+        raise ValueError(f"start_group_id must be non-negative, got {start_group_id}")
+    num_groups = batch_size // group_size
+    return torch.arange(
```
yfw: [`utils.py:128`](https://github.com/NVIDIA-NeMo/RL/blob/fc1a3c193357ee7a64374141a7a0c5ec8e597ed0/nemo_rl/algorithms/utils.py#L128)

**No action needed on the code** — this is about the evidence plan in your description, and one CI fact worth knowing before you rely on it.

**TL;DR — the collision metric proposed in your self-review can't ever show a collision, because on this branch it's measuring the fixed output rather than the old key.**

`build_rollout_group_ids` returns `arange(start, start + num_groups).repeat_interleave(group_size)`, so `len(torch.unique(prompt_ids_for_adv, dim=0))` is identically `batch_size // group_size` — i.e. exactly `num_prompts_per_step` on every recipe, text and VLM alike. It's a tautology at head; it can never come out "strictly less."

**The probe has to run against the old token key, on the base commit** (`af0a11fac`), one line before `calculate_baseline_and_std_per_prompt`:

```python
print(f"[collisions] unique_prompt_rows={len(torch.unique(prompt_ids_for_adv, dim=0))} "
      f"num_prompts_per_step={master_config.grpo.num_prompts_per_step}")
```

`unique_prompt_rows < num_prompts_per_step` means the old and new partitions differ, which means this PR changes that run's advantages. One step, one line, and if it comes back `<` on a VLM recipe that is the whole empirical case — no A/B needed.

**Two things that make this worth doing rather than skipping:**

1. **`CI:L2` will not exercise the motivating case.** We checked `cicd-main.yml` and there is no L2-only gate anywhere — every occurrence is `contains('L1 L2', ...)`, and pre-flight collapses `L1 | L2)` into one branch, so `CI:L2` runs exactly what `CI:L1` runs. The `vlm_grpo-*clevr*` recipes appear only in `nightly*.txt`. So nothing before merge touches the multimodal collision this PR is named for; if you want a VLM signal you'd name `vlm_grpo-qwen2.5-vl-3b-instruct-clevr-1n8g-megatrontp2.v1.sh` explicitly.
2. **Text recipes are not a no-op either.** Any text dataset with duplicate prompt strings in a step previously pooled those rows into one baseline and now gets separate per-instance baselines. That's the correct behaviour, but it's a real numerical delta — worth one sentence in the description so reviewers of a future convergence diff don't chase it.

PR #3883, opened 2026-08-27
feat(sc): log rollout and speculative decoding metrics

yfw:
## Summary

Ports the remaining v2 SingleController metrics behavior from #2175:

- preserve per-group rollout metrics through the TransferQueue replay buffer and staleness samplers
- aggregate metrics from the groups selected for each training step using the existing semantic min/max/sum/mean and histogram handling
- snapshot generation metrics before the first training chunk and collect them after optimizer completion, so vLLM speculative-decoding deltas are logged with the correct step

Generation backends without per-step metrics remain compatible through default no-op interface methods. There are no configuration or migration changes.

## Validation

- 134 focused SingleController replay-buffer, sampler, and actor tests passed
- 13 rollout-aggregation and speculative-decoding counter tests passed
- targeted pre-commit hooks passed, including Ruff, Ruff format, and Pyrefly


tests/unit/single_controller/test_single_controller_actor.py
```diff
@@ -1366,6 +1402,90 @@ def test_train_pump_logs_nonzero_stale_group_metrics(monkeypatch) -> None:
...
+    train_call = ctrl._logger.log_metrics.call_args_list[0]
+    train_metrics = train_call.args[0]
+    assert train_metrics["gen_tokens/min"] == 3
+    assert train_metrics["gen_tokens/max"] == 20
+    assert train_metrics["total_turns"] == 7
+    assert train_metrics["accuracy"] == pytest.approx(0.5)
+    assert train_metrics["trajectory_duration_s"] == pytest.approx(2.0)
+    assert train_metrics["trajectory_duration_s/max"] == 3.0
+    assert train_metrics["trajectory_duration_s/p95"] == 3.0
+    assert train_call.kwargs == {"step": 1, "prefix": "train"}
```
yfw, line 1577: [`tests/unit/single_controller/test_single_controller_actor.py:1453`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/tests/unit/single_controller/test_single_controller_actor.py#L1453)

**1 action item.**

**TL;DR - whether line 1038 is `pop` or `get` decides if the whole per-group metrics payload gets pickled to every DP rank on every dispatch of the step, and no test catches the difference.**

The [`.pop()`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/single_controller.py#L1038) is load-bearing rather than incidental, and nothing pins it - this test stays green if it becomes `.get()`.

Why it matters:
1. [`_finalize_selection`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/async_utils/staleness_sampler.py#L241) merges each selected group's metrics onto `selected_meta.extra_info[ROLLOUT_METRICS]`.
2. The train pump [pops them straight back off](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/single_controller.py#L1037-L1039) before `train_meta` goes anywhere else.
3. That ordering is what keeps them off the wire: `train_meta` then flows to [`get_logprobs_from_meta`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/single_controller.py#L1064), `_value_stage`, `_advantage_stage` and `train_microbatches_from_meta`, all of which fan out through [`shard_meta_for_dp`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/data_plane/preshard.py#L136), which copies `extra_info` into [every per-rank meta](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/data_plane/preshard.py#L145) - [documented here](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/data_plane/README.md#L390).
4. With `.get()`, the whole per-group payload - several histogram lists per group, plus a [`wandb.Table`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/experience/rollout_manager.py#L1125) on the Gym path - would be pickled once per DP rank on every logprob and training dispatch of the step. Silent; it would only surface as throughput.

I verified this: changing `pop` to `get` leaves all eight existing assertions in this test green, and only the added line below goes red.

**Action:** pin it with a one-line assertion (verified passing; the whole file is 47 passed).

```suggestion
    assert train_call.kwargs == {"step": 1, "prefix": "train"}
    assert all(ROLLOUT_METRICS not in meta.extra_info for meta in metas)
```

nemo_rl/algorithms/single_controller.py
```diff
@@ -1384,8 +1403,8 @@ async def _train_pump(self) -> None:
                 print(f"  • {k}: {v:.2f}s ({percent:.1f}%)")
 
             # TODO: per-step train_data jsonl dump, vllm metrics logger,
-            #   histogram log, rollout_metrics, pretty-print "Training Results"
-            #   block, print_performance_metrics.
+            #   histogram log, pretty-print "Training Results" block,
+            #   print_performance_metrics.
             print(f"step_metrics={step_metrics}", flush=True)
```
yfw: [`nemo_rl/algorithms/single_controller.py:1408`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/single_controller.py#L1408)

**1 action item.**

**TL;DR - this print now dumps every raw histogram observation to the driver log, order 10 KB of numbers per step, because `step_metrics` carries flattened histogram lists post-PR.**

PR-introduced log noise. Now that rollout metrics land in `step_metrics`, this line dumps every raw histogram observation to the driver log each step.

Each group contributes 5 `{key}/histogram` keys - [`calculate_single_metric`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/experience/metric_utils.py#L48) returns `list(values)` and is called 5x in [`_aggregate_rollout_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/experience/rollout_manager.py#L700) - plus 3 [`histogram/*` keys](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/experience/rollout_manager.py#L729-L737), and [`aggregate_rollout_metrics` flattens them across groups](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/grpo.py#L4172-L4173). At `grpo-qwen3-30ba3b-10n8g-megatron-cp2-r3-async-single-controller.yaml`'s [16 prompts x 8 generations](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/examples/configs/recipes/llm/grpo-qwen3-30ba3b-8n8g-megatron-cp2-r3.yaml#L3-L4) that is 5+ lists of 128 observations on a single stdout line, order 10 KB/step, scaling with turn count on multi-turn Gym recipes.

Logging itself is unaffected - [`Logger.log_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/utils/logger.py#L1061-L1062) routes histogram keys to `log_histogram` before any backend sees them. Only this raw print is, and it is the only whole-metrics-dict print in `nemo_rl/algorithms/`.

**Action:** filter list-valued entries out of the print. One option of several (verified `ruff format` / `ruff check` clean) - [`is_histogram_metric`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/experience/metric_utils.py#L22) is the more precise filter if you would rather add the import:

```suggestion
            print(
                "step_metrics="
                + str(
                    {k: v for k, v in step_metrics.items() if not isinstance(v, list)}
                ),
                flush=True,
            )
```

tests/unit/single_controller/test_single_controller_actor.py
```diff
@@ -1366,6 +1402,90 @@ def test_train_pump_logs_nonzero_stale_group_metrics(monkeypatch) -> None:
...
+
+
```
yfw: [`tests/unit/single_controller/test_single_controller_actor.py:1454`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/tests/unit/single_controller/test_single_controller_actor.py#L1454)

**1 action item.**

**TL;DR - the list-returning `is_histogram_metric` branch of `aggregate_rollout_metrics` is always taken in production but never in these tests, since every fixture here is scalar-only.**

Coverage gap, not a defect: production rollout metrics *always* contain histogram keys, but every fixture in this PR is scalar-only, so the one branch of `aggregate_rollout_metrics` that returns a **list** instead of a scalar is never taken on the SingleController route.

[`calculate_single_metric`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/experience/metric_utils.py#L48) emits a `{key}/histogram` list for every metric, and [`rollout_manager`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/experience/rollout_manager.py#L729-L737) adds three more `histogram/*` lists per group. Those hit the [`is_histogram_metric` branch](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/grpo.py#L4172-L4173), which concatenates across groups rather than averaging. Nothing currently breaks - [`Logger.log_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/utils/logger.py#L1061) expects the list - so this pins behaviour rather than fixing a bug.

**Action:** add a test for the flattening branch. Verified passing in place (47 passed in this file), and teeth-checked: it goes red under `monkeypatch.setattr(grpo, "is_histogram_metric", lambda name: False)`.

```suggestion


def test_train_pump_flattens_histogram_rollout_metrics(monkeypatch) -> None:
    metas = [
        KVBatchMeta(
            partition_id="rollout_data",
            task_name="train",
            sample_ids=[f"sample-{index}"],
            fields=[],
            sequence_lengths=[1],
            extra_info={ROLLOUT_METRICS: [metrics]},
            tags=[{"weight_version": 0}],
        )
        for index, metrics in enumerate(
            [
                {"histogram/gen_tokens_length": [1, 2]},
                {"histogram/gen_tokens_length": [3]},
            ]
        )
    ]
    ctrl = _train_pump_controller(sampler=_SequenceSampler(metas))
    ctrl._sync_weights = AsyncMock(return_value=0)
    ctrl._logger = MagicMock()
    monkeypatch.setattr(single_controller.ray, "cluster_resources", lambda: {})

    asyncio.run(asyncio.wait_for(ctrl._train_pump(), timeout=1.0))

    train_metrics = ctrl._logger.log_metrics.call_args_list[0].args[0]
    # Histogram observations are concatenated into one step-level distribution,
    # not averaged; Logger.log_metrics routes the list to log_histogram.
    assert train_metrics["histogram/gen_tokens_length"] == [1, 2, 3]


```

tests/unit/single_controller/test_sampler_interface.py
```diff
@@ -354,6 +357,29 @@ def test_selects_ready_groups_in_window(self):
...
+        )
+
+        assert num_groups == 2
+        assert meta is not None
+        assert meta.extra_info[ROLLOUT_METRICS] == [
+            {"metric": 1.0},
+            {"metric": 2.0},
+        ]
+        assert buf.start_weight_list == [5]
+
```
yfw, line 456: [`tests/unit/single_controller/test_sampler_interface.py:382`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/tests/unit/single_controller/test_sampler_interface.py#L382)

**1 action item.**

**TL;DR - the code correctly defaults the missing key, but nothing pins that, so a later bare-subscript refactor would pass every test and fail only on resume from a pre-PR checkpoint.**

Coverage gap, not a defect. A checkpoint written before this PR restores metas whose `extra_info` has no `ROLLOUT_METRICS` key at all - [`state_dict` pickles the whole `KVBatchMeta`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/async_utils/replay_buffer.py#L1017) and `load_state_dict` never touches `extra_info`. Your [`.get(ROLLOUT_METRICS, [])`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/async_utils/staleness_sampler.py#L238) and [`.pop(ROLLOUT_METRICS, [])`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/single_controller.py#L1038) both already handle it correctly, so there is no bug here.

The gap is that the defaulted accessor is now unpinned: post-PR [`FakeBuffer.add`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/tests/unit/single_controller/test_sampler_interface.py#L67) always sets the key, so the key-less shape is unreachable in this suite. A later refactor to a bare `extra_info[ROLLOUT_METRICS]` subscript would pass every test and fail only on resume.

**Action:** add a regression test. Verified passing in place (98 passed across both suites) and teeth-checked. The key-less group is deliberately `selected_metas[0]` - the one whose `extra_info` [`concat` copies forward](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/data_plane/interfaces.py#L253) - which is the harder ordering.

```suggestion

    def test_carries_metrics_for_groups_restored_without_them(self):
        buf = FakeBuffer()
        buf.add("restored", weight=5)
        buf.add("fresh", weight=5, rollout_metrics={"metric": 2.0})
        # A checkpoint written before per-group metrics existed restores metas
        # whose extra_info has no ROLLOUT_METRICS key at all.
        buf.meta_list[0].extra_info.pop(ROLLOUT_METRICS)
        sampler = WindowedSampler(buf, max_staleness_versions=1)

        meta, num_groups = _run(
            sampler.select(
                current_train_weight=5,
                min_prompt_groups=1,
                max_prompt_groups=2,
            )
        )

        assert num_groups == 2
        assert meta is not None
        assert meta.extra_info[ROLLOUT_METRICS] == [{"metric": 2.0}]

```

yfw (commented):
Nice change - the SingleController metrics now line up with the async GRPO path, and the shape is right: declaring `snapshot_step_metrics`/`get_step_metrics` on [`GenerationInterface`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/models/generation/interfaces.py#L537-L548) and calling them directly is better than adding a seventh `hasattr` guard, and it mirrors the [`clear_logger_metrics`/`get_logger_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/models/generation/interfaces.py#L518-L535) pair already on that ABC. The gather-before-concat in [`_finalize_selection`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/async_utils/staleness_sampler.py#L235-L241) is also load-bearing and correct, since [`concat` keeps only the first meta's `extra_info`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/data_plane/interfaces.py#L253); routing it through the single shared `_finalize_selection` covers all four samplers.

One side effect worth knowing about: because rollout metrics now land in `step_metrics`, `checkpointing.metric_name: "train:<rollout metric>"` starts working on SingleController. Previously [`_save_checkpoint`](https://github.com/NVIDIA-NeMo/RL/blob/0cdf117b943a1bca7becd9c80ad55753df41804e/nemo_rl/algorithms/single_controller.py#L1780-L1781) would raise `ValueError: Metric ... not found in train metrics` for any rollout-derived name.

Three low-severity comments inline - one coverage addition and two nits. None of them blocks.

_Generated by Claude Code_

tests/unit/single_controller/test_sampler_interface.py
yfw: Addressed in 4b7c974e9. Added a regression test with the first selected meta shaped like a pre-PR checkpoint (no rollout_metrics key) and a fresh second meta, verifying selection preserves only the available metrics.

nemo_rl/algorithms/single_controller.py
yfw: Addressed in 4b7c974e9. The diagnostic print now omits list-valued payloads, while the complete step_metrics dictionary still goes to Logger.log_metrics. The controller test verifies the flattened histogram reaches the logger but not stdout.

tests/unit/single_controller/test_single_controller_actor.py
yfw: Addressed in 4b7c974e9. The controller test now asserts that the rollout sidecar is removed from every selected meta after the train pump consumes it, pinning the pop-before-dispatch behavior.

nemo_rl/algorithms/single_controller.py
```diff
@@ -1240,6 +1249,19 @@ async def _train_pump(self) -> None:
...
                 )
+                per_group_rollout_metrics: dict[str, list[Any]] = {}
+                for group_metrics in selected_rollout_metrics:
+                    for metric_name, value in group_metrics.items():
+                        per_group_rollout_metrics.setdefault(metric_name, []).append(
+                            value
+                        )
+                step_metrics.update(
+                    aggregate_rollout_metrics(per_group_rollout_metrics)
+                )
```
terrykong: [`single_controller.py:1258-1260`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/algorithms/single_controller.py#L1258-L1260)

**1 action item.**

**TL;DR — `per_worker_token_counts` is a dict of per-worker counts, and this line is what first lets it reach `wandb.run.log()` from SingleController. Every other training loop deletes it before logging.**

PR-introduced. Nothing about the aggregation is wrong — the key just never used to get this far on this path.

How it gets there:
1. Each prompt group's rollout metrics carry `per_worker_token_counts` as a `dict[int, int]` — [`rollouts.py:1500`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/experience/rollouts.py#L1500), on the async multi-turn path SingleController uses ([`trajectory_collector.py:1310`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/algorithms/async_utils/trajectory_collector.py#L1310)).
2. `aggregate_rollout_metrics` has no rule for dict values, so it falls into the "not a number" branch and passes the whole per-group list straight through — [`grpo.py:4174-4175`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/algorithms/grpo.py#L4174-L4175). You get `[{0: 100, 1: 50}, {0: 20}]`.
3. This line merges that into `step_metrics`.
4. `Logger.log_metrics` only reroutes histogram keys; everything else goes to [`WandbLogger.log_metrics` → bare `self.run.log(metrics, step=step)`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/utils/logger.py#L392).

Every other loop avoids step 4. `print_performance_metrics` deletes the key, with the reason written next to it — [`utils.py:900-902`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/algorithms/utils.py#L900-L902):

> `# Dict structures cannot be logged to wandb`

Async GRPO [calls that function](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/algorithms/grpo.py#L5557) — which edits `metrics` in place — and only then [logs](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/algorithms/grpo.py#L5602). `_train_pump` never calls it; it is still on the TODO at [`single_controller.py:1408-1410`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/algorithms/single_controller.py#L1408-L1410). TensorBoard and MLflow are fine either way — both drop values they can't handle — so this shows up only in W&B.

Being straight with you about the limit of this: I could not run W&B here, so I don't know whether `run.log` raises or just writes a junk series. What I can point at is that the repo says this value can't be logged and deletes it everywhere else.

**Action:** drop the key before logging in `_train_pump` (`nemo_rl/algorithms/single_controller.py`). Wiring up `print_performance_metrics`, already on your TODO, would cover this too and is the better long-term fix — this is the one-line version until then.

```suggestion
                step_metrics.update(
                    aggregate_rollout_metrics(per_group_rollout_metrics)
                )
                # Dict structures cannot be logged to wandb. Every other loop drops
                # this inside print_performance_metrics, which _train_pump does not
                # call yet (see the TODO below and algorithms/utils.py:899-902).
                step_metrics.pop("per_worker_token_counts", None)
```

nemo_rl/data_plane/schema.py
```diff
@@ -23,6 +23,7 @@
 MICRO_BATCH_LENGTHS = "micro_batch_lengths"
 ELEM_COUNTS_PER_GB = "elem_counts_per_gb"
 GLOBAL_FORWARD_PAD_SEQLEN = "global_forward_pad_seqlen"
+ROLLOUT_METRICS = "rollout_metrics"
```
terrykong, line 31: [`schema.py:26`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/data_plane/schema.py#L26)

**1 action item.**

**TL;DR — this constant is filed under "per-shard packing keys", but it is the one key that must NOT be copied to each shard; the comment above it tells the next reader the opposite of the truth.**

The new constant sits under the header [`# Per-shard packing metadata keys in KVBatchMeta.extra_info.`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/data_plane/schema.py#L21), but it does not behave like the four keys above it. Those get copied onto every rank's shard by [`shard_meta_for_dp`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/data_plane/preshard.py#L145). `ROLLOUT_METRICS` must do the opposite — the train pump [pops it off before the meta is dispatched](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/algorithms/single_controller.py#L1038), which is exactly what keeps it off the wire. Filing it under that header tells the next reader it follows the shard-copy rules.

One thing worth writing down while you're there: [`sync_rollout_actor.py:406`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/experience/sync_rollout_actor.py#L406) already writes the same `"rollout_metrics"` string, but as a single flat dict rather than a list of per-group dicts. Nothing breaks today — that meta goes through `grpo_sync`, not `_train_pump` — and if the two ever did meet it would fail loudly on the first step (`.extend()` over a flat dict yields its keys, then `.items()` on a `str` raises `AttributeError`), so there is no silent-corruption risk to guard against. It just means the constant is not a drop-in for the sync path, and that is cheaper to say than to rediscover.

**Action:** in `nemo_rl/data_plane/schema.py`, move the constant out of the packing block and say what it holds.

```suggestion

# Per-prompt-group rollout metrics: a list of one metrics dict per group.
# Unlike the packing keys above, this is NOT copied to each shard -- the train
# pump pops it off the meta before dispatch. Note sync_rollout_actor.py writes
# the same string with a flat-dict shape, so this constant is not a drop-in there.
ROLLOUT_METRICS = "rollout_metrics"
```

nemo_rl/models/generation/interfaces.py
```diff
@@ -533,3 +533,16 @@ def get_logger_metrics(self) -> dict[str, Any]:
...
+
+        Backends without per-step generation metrics may use this default no-op.
+        """
+
+    def get_step_metrics(self) -> dict[str, float]:
+        """Finish the current metric window and return generation metrics.
+
+        Backends without per-step generation metrics return an empty dictionary.
+        """
+        return {}
```
terrykong, line 652: [`interfaces.py:543-548`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/models/generation/interfaces.py#L543-L548)

**1 action item.** Small one.

The method returns a value but has no `Returns:` section, while its neighbour [`get_logger_metrics`](https://github.com/NVIDIA-NeMo/RL/blob/3357deb6954dc14ace2b6abb8d971fdc1b7bbf21/nemo_rl/models/generation/interfaces.py#L532) eleven lines up has one. Worth adding here in particular because the return is a *delta* since the matching snapshot, not a running total — a backend author implementing this needs to know that, and right now the only place it is written down is the vLLM implementation.

**Action:**

```suggestion
    def get_step_metrics(self) -> dict[str, float]:
        """Finish the current metric window and return generation metrics.

        Returns:
            Metrics accumulated since the matching ``snapshot_step_metrics``
            call, not running totals. Backends without per-step generation
            metrics return an empty dictionary.
        """
        return {}
```

tests/unit/single_controller/test_single_controller_actor.py
terrykong: fixed in 4b7c974e9ec95632a688fb40f2b65ad46779da55

nemo_rl/algorithms/single_controller.py
terrykong: fixed in 4b7c974e9ec95632a688fb40f2b65ad46779da55

tests/unit/single_controller/test_sampler_interface.py
terrykong: fixed in 4b7c974e9ec95632a688fb40f2b65ad46779da55

terrykong (commented):
Reviewed by a team of six agents (RL codebase, bug hunt, tests, design, comment threads, plus an adversarial pass). The change looks sound — I went looking for problems in the four places this kind of change usually breaks and found none of them:

- **Snapshot/collect pairing is balanced.** `snapshot_step_metrics` fires exactly once per step (`groups_dispatched` is only incremented after the snapshot site), and every path that leaves the chunk loop earlier either returns before the snapshot fires or kills the pump. No window is left open into the next step, so vLLM's `RuntimeWarning` guards can't trip.
- **`asyncio.to_thread(self._gen.…)` is right.** `_gen` is a driver-side generation wrapper, not a Ray actor handle, and its methods block on `ray.get` internally — same shape as the existing `asyncio.to_thread(self._gen.invalidate_kv_cache)` in this file.
- **No aliasing through `concat`.** `_replace` builds a fresh `extra_info` dict, so writing the merged list onto the concatenated meta doesn't reach back into the buffered ones. All four samplers share the one `_finalize_selection`, so none is silently left out.
- **Not a hook with no implementation.** vLLM implements both methods on `main` already, so the spec-decode metrics land as soon as this runs.

The three comments from the earlier round are all genuinely fixed in 4b7c974e9. I didn't take that on trust — for each one we flipped the behaviour back in the source and confirmed the new assertion goes red, then restored it. Replies are on the threads.

Ran `tests/unit/single_controller/` locally: **463 passed**. The one failure is a missing `HF_HOME` on my machine, unrelated to this PR. Repo linters could not run here (the lockfile is Linux-only).

Two small inline comments below. Neither blocks.

**On CI:** thanks for re-triggering on `3357deb69` — that answers the one thing I was going to ask for, since the previous run only covered `4b7c974e9`. The one red is `Check if PR branch is up to date`: the branch is 12 commits behind `main` (merge-base `6ae035784`). `main` has since rewritten `test_setup.py` and `single_controller_utils/setup.py`, but neither collides with this diff, so a rebase should be mechanical.

Two observations with no action attached:

- `step_metrics` now carries `per_worker_token_counts`, which SingleController has no consumer for yet — the `performance/average_token_imbalance` number the other four loops derive from it comes along when `print_performance_metrics` does, already on your TODO. Nothing to do here; noting it so the extra key isn't a surprise in W&B.
- The per-step generation windows don't quite tile — generation keeps running through refit, checkpointing, and the next step's wait for its first chunk, and none of that falls inside a window. So the raw `vllm/spec_num_*` counters undercount slightly while `spec_acceptance_rate`, the ratio people actually watch, stays right. The async GRPO path has the same shape, so this isn't something you introduced, and the current vLLM API can't fully close it anyway.

_Generated by Claude Code_

nemo_rl/data_plane/schema.py
yfw: Addressed in 3bfcb3179. `ROLLOUT_METRICS` is now separated from the per-shard packing keys, with its per-group list shape, pop-before-dispatch behavior, and the sync path's flat-dict shape documented.

nemo_rl/models/generation/interfaces.py
yfw: Addressed in 3bfcb3179. `get_step_metrics()` now documents that it returns metrics accumulated since the matching snapshot rather than running totals, while unsupported backends return an empty dictionary.

terrykong (approved):

sharonyu-115:
hi @yfw I'd like to flag a likely SingleController regression exposed after this PR merged.

  PR #3911 now fails the same `grpo_dp_single_controller_chaos.sh` test on both platforms:

  - [H100 failure](https://github.com/NVIDIA-NeMo/RL/actions/runs/33601678101/job/100196988365)
  - [GB200 failure](https://github.com/NVIDIA-NeMo/RL/actions/runs/33601678101/job/100196988021)

  The test intentionally SIGKILLs an idle `VllmAsyncGenerationWorker` and expects the typed rollout/fleet failure contract established by #3589. Instead, the job exits with a raw
  `ActorDiedError` and the harness reports:

      [chaos] FAIL: job stopped but no typed rollout failure was reported

  Both traces terminate through the metrics call added by this PR:

      SingleControllerActor._train_pump
        -> self._gen.get_step_metrics()
        -> VllmGeneration._get_raw_spec_counters()
        -> ray.get(futures)
        -> ray.exceptions.ActorDiedError

  `get_step_metrics()` queries all generation workers, including the intentionally killed actor. The resulting exception escapes before the existing fleet-health/rollout-failure path can
  report a typed failure.

  The failing test uses Qwen3, and #3911 does not modify `single_controller.py`, `vllm_generation.py`, or the chaos test, so this does not appear related to its Gemma4 changes. The failure
  reproduced identically on H100 and GB200.

  I have not yet run the chaos test against this commit’s parent or with the new metrics calls reverted, so this is strongly indicated by the diff and stack rather than confirmed by a
  completed bisect.

PR #3884, opened 2026-08-27
fix: isolate SingleController service port ranges

yfw:
## Summary

- port the remaining service-range isolation from #3103 to SingleController and Nemotron recipes
- reserve `[1202, 1300)` for the SingleController generation router so it cannot overlap Ray management, Gym, sandbox, or generation services
- remove unsafe above-9000 recipe overrides and validate resolved service ranges across all example configs that pin ports

## Why

The affected recipes still selected service ports in the observed ephemeral range. A port can be free during probing and then be claimed as a kernel source port before the service binds, causing intermittent startup collisions. The previous router default also overlapped sandbox ports on the driver node.

## Validation

- 45 targeted config, router, virtual-cluster, and recipe validation tests
- FAST-CI collection: 60/62 virtual-cluster tests collected, with only the two slow tests deselected
- `ruff check` and `ruff format --check` on touched Python files
- Pyrefly: 0 errors
- `bash -n ray.sub` and `bash -n tests/unit/excluded_unit_tests.sh`
- `git diff --check`

## Not run

- representative multi-node cluster startup smoke; this requires a remote cluster allocation


nemo_rl/distributed/virtual_cluster.py
```diff
@@ -87,20 +87,20 @@ class PY_EXECUTABLES:
 #
 # Python port-range bounds below are half-open: [low, high).
 #
+#   [1202, 1300) SingleController gen. router    (driver-local allocation)
 #   1313-1399    Dynamo etcd/NATS control plane  (driver-local allocation)
 #   1400-1999    Master address / TCPStore       (cluster.master_port_range_low/high)
 #   [3000, 4999) Shared NeMo RL generation range (policy.generation.port_range_low/high)
 #     [3000, 4000) Dynamo frontend/token-wrapper HTTP endpoints
 #     [4000, 4100) Dynamo worker system endpoints (node-local free-port selection)
 #   5000-5999    NeMo Gym HTTP servers           (env.nemo_gym.port_range_low/high)
```
yfw, line 96: [`nemo_rl/distributed/virtual_cluster.py:96`](https://github.com/NVIDIA-NeMo/RL/blob/605e14fc68975c1c0f335b1f5dd340cf87c0e6e7/nemo_rl/distributed/virtual_cluster.py#L96)

**1 action item.**

**TL;DR — the NeMo-Skills sandbox at 6000-6999 is the one non-Ray occupant missing from this table, so the gap this PR opens at 6000 reads as free space.**

Adjacent rather than PR-introduced — but this PR rewrites this table and removes the row that occupied 6000, so the hole is newly visible.

Relative to [`ray.sub:158-173`](https://github.com/NVIDIA-NeMo/RL/blob/605e14fc68975c1c0f335b1f5dd340cf87c0e6e7/ray.sub#L158-L173), this table omits six bands: `1200-1201` Ray GCS/client, `1301-1312` Ray management, `2000-2999` Ray worker gRPC, `8265` dashboard, and `6000` / `6001-6999` the NeMo-Skills sandbox. Five of the six are Ray's, which the header at [`:83-86`](https://github.com/NVIDIA-NeMo/RL/blob/605e14fc68975c1c0f335b1f5dd340cf87c0e6e7/nemo_rl/distributed/virtual_cluster.py#L83-L86) explicitly disclaims — *"See ray.sub for the full layout including Ray's own GCS / worker gRPC ports."* The sandbox is the only non-Ray omission, so it is the only occupied band a reader of this file alone is left to infer is free.

It is a real occupant, though only on sandbox-enabled runs: [`ray.sub:1014-1015`](https://github.com/NVIDIA-NeMo/RL/blob/605e14fc68975c1c0f335b1f5dd340cf87c0e6e7/ray.sub#L1014-L1015) starts one sidecar per allocated node (`--nodes="$SLURM_JOB_NUM_NODES" --ntasks-per-node=1 --overlap`), driver included, gated on `SANDBOX_CONTAINER`/`SANDBOX_COMMAND` at [`ray.sub:991`](https://github.com/NVIDIA-NeMo/RL/blob/605e14fc68975c1c0f335b1f5dd340cf87c0e6e7/ray.sub#L991).

**Action:** add the sandbox rows so the next band-picker sees them. Comments-only — `ruff check` and `ruff format --check` pass and the parsed AST is unchanged. (If you'd rather state the rule once instead, widening the header disclaimer to name the sandbox band works too.)

```suggestion
#   5000-5999    NeMo Gym HTTP servers           (env.nemo_gym.port_range_low/high)
#   6000         NeMo-Skills sandbox Nginx       (NEMO_SKILLS_SANDBOX_PORT; ray.sub starts one
#                                                 sidecar per allocated node, driver included)
#   6001-6999    NeMo-Skills sandbox uWSGI       (SANDBOX_BASE_PORT)
```

tests/unit/distributed/test_virtual_cluster.py
```diff
@@ -637,6 +639,14 @@ def test_no_vllm_port_without_bundle_indices(self):
...
     # Lowest observed ephemeral floor on some GB200 nodes; stock Linux is 32768.
     EPHEMERAL_FLOOR = 9000
+    # The driver-local generation router fits between Ray's client port (1201)
+    # and its first management port (1301).
+    assert 1201 < DEFAULT_GENERATION_ROUTER_PORT_RANGE_LOW
+    assert (
+        DEFAULT_GENERATION_ROUTER_PORT_RANGE_LOW
+        < DEFAULT_GENERATION_ROUTER_PORT_RANGE_HIGH
+        <= 1301
+    )
```
yfw: [`tests/unit/distributed/test_virtual_cluster.py:642-649`](https://github.com/NVIDIA-NeMo/RL/blob/605e14fc68975c1c0f335b1f5dd340cf87c0e6e7/tests/unit/distributed/test_virtual_cluster.py#L642-L649)

**1 optional action item.**

**TL;DR — `1201` and `1301` are hardcoded copies of `ray.sub`'s `RAY_CLIENT_SERVER_PORT` and `NODE_MANAGER_PORT`, so renumbering `ray.sub` would silently invalidate the whole justification for `[1202, 1300)` with nothing going red.**

PR-introduced. These two literals — and the same pair at [`tests/unit/utils/test_config.py:345`](https://github.com/NVIDIA-NeMo/RL/blob/605e14fc68975c1c0f335b1f5dd340cf87c0e6e7/tests/unit/utils/test_config.py#L345) — carry the entire argument for the new band, and they duplicate values that actually live in [`ray.sub:105-116`](https://github.com/NVIDIA-NeMo/RL/blob/605e14fc68975c1c0f335b1f5dd340cf87c0e6e7/ray.sub#L105-L116).

**Action (optional):** derive them from `ray.sub` instead. The test below passes today, and I mutation-tested it: **5/5 mutants red** (band low → 1201 or 1200, high → 1302 or 1303, band moved into the 2000-2999 worker range), baseline and restore green. Gates measured against an unpatched control copy: `ruff check` +0 errors, `ruff format --check` pass, `pyrefly` +0. Not a `suggestion` block because it is a new module-scope function rather than an edit to these lines.

```python
def test_router_band_does_not_collide_with_ray_sub_ports():
    """Derive Ray's reserved ports from ray.sub instead of hardcoding 1201/1301."""
    import re
    from pathlib import Path

    ray_sub = (Path(__file__).resolve().parents[3] / "ray.sub").read_text()

    def default_for(name: str) -> int:
        match = re.search(rf"^{name}=\$\{{{name}:-(\d+)\}}", ray_sub, re.MULTILINE)
        assert match is not None, f"{name} default not found in ray.sub"
        return int(match.group(1))

    reserved = {default_for("PORT"), default_for("RAY_CLIENT_SERVER_PORT")}
    # ray.sub gives the head node these ports + 1 (workers get the odd values).
    for name in (
        "NODE_MANAGER_PORT",
        "OBJECT_MANAGER_PORT",
        "RUNTIME_ENV_AGENT_PORT",
        "DASHBOARD_AGENT_GRPC_PORT",
        "METRICS_EXPORT_PORT",
        "DASHBOARD_AGENT_LISTEN_PORT",
    ):
        reserved |= {default_for(name), default_for(name) + 1}
    reserved |= set(
        range(default_for("MIN_WORKER_PORT"), default_for("MAX_WORKER_PORT") + 1)
    )

    band = set(
        range(
            DEFAULT_GENERATION_ROUTER_PORT_RANGE_LOW,
            DEFAULT_GENERATION_ROUTER_PORT_RANGE_HIGH,
        )
    )
    assert not band & reserved, sorted(band & reserved)
```

yfw (commented):
Verified locally at `605e14f` on a clean tree:

- `pre-commit run --all-files` passes.
- `tests/unit/utils/test_config.py` — 43 passed (26 are the new parametrized port cases).
- Router + fleet-health + `tests/unit/single_controller/` — 526 passed.
- One unrelated failure, `test_virtual_cluster.py::test_mcore_py_executable`, which dies because `uv run --locked --extra mcore` can't build transformer-engine in my worktree. Environment, not this PR.

Things I specifically checked and found correct:

- **The band is genuinely free.** Ray takes 1200, 1201 and 1301-1312; the next NeMo-RL band up is Dynamo at 1313-1399, so 1202-1300 is a real gap. No constant in Ray's own `ray_constants.py` falls in it, and any port Ray auto-assigns is structurally above the ephemeral floor — `Node._get_unused_port` seeds from `bind(("", 0))` and only searches *upward*. That also covers the k8s/KubeRay path, which pins no Ray ports at all. 98 ports vs the old 99 makes no practical difference to the 50-draw retry budget.
- **"Driver-local" is accurate** — `_get_free_port_local` runs on the driver and the actor is pinned with `NodeAffinitySchedulingStrategy(..., soft=False)`, so allocate-node == bind-node.
- **The new test catches every regression it guards.** Against the pre-PR tree it fails on exactly the three configs this PR edits — naming which assertion fires for each (generation 11001/15000, master 25000/28000, gym 15001/20000) — and with the old 6000/6099 router default it fails 26/26. On the PR tree, 0/26. The constant-level assertions are live too: mutating the router LOW to 1201 or 6000, or HIGH to 1400, each turns them red.
- **No case is vacuous** — every one of the 26 configs supplies at least one of the four bands from YAML rather than falling through to the constants.
- **Both new YAML comments are accurate**, and precisely so: grpo genuinely inherits 3000/4999 from `grpo_math_1B.yaml`, while ppo's key is absent and falls through to the Python default — and the two comments say exactly that. Every consumer reads these keys with `.get()`, and `RayVirtualCluster` normalizes `None` back to the defaults, so the removals carry no `KeyError` risk.
- **`[1202, 1300)` vs `5000-5999` notation is right.** `_bind_socket_in_range` is exclusive-high, Gym's `_find_open_port_using_range` is inclusive-high, and the table's half-open/inclusive notation matches each. This is the easy thing to get wrong here, and it isn't wrong.
- **Completeness holds** — sorting every `port_range_*` value in every YAML repo-wide gives a global max of 5999. Nothing above 9000 remains anywhere, not just under `examples/`.
- **swe_teacher's gym shrink (4999 → 999 ports) is safe.** Gym allocates one port per *server instance*, never per rollout, so swe_teacher needs ~4 ports — the `concurrency: 1024` in that file is aiohttp request concurrency and consumes none. Its closest live analog, `nemotron-3-super/stage2_swe2.yaml` (same `swebench_openhands_training.yaml`, same train/val split), already ships 5000-5999 on main; and after this PR swe_teacher matches all six of its `nemotron-3-ultra/` siblings exactly. It was the sole outlier in its own directory, and its old range was already tripping the repo's own out-of-band warning in `nemo_gym.py`.

Two low-severity suggestions inline; neither blocks.

_Generated by Claude Code_

nemo_rl/distributed/virtual_cluster.py
yfw: Addressed in f20abcb41: the Python-side port map now documents sandbox Nginx on 6000 and uWSGI on 6001-6999, including the per-node sidecar behavior.

tests/unit/distributed/test_virtual_cluster.py
yfw: Addressed in f20abcb41. The test now parses Ray service defaults directly from ray.sub, including the dashboard port; recipe checks use the declared router constants, and the constant-level test also guards the adjacent Dynamo boundary.

tests/unit/distributed/test_virtual_cluster.py
```diff
@@ -45,6 +50,13 @@
 from tests.unit.conftest import TEST_ASSETS_DIR
 
 
+def _ray_sub_default(name: str) -> int:
+    ray_sub = (Path(__file__).resolve().parents[3] / "ray.sub").read_text()
+    match = re.search(rf"^{name}=\$\{{{name}:-(\d+)\}}", ray_sub, re.MULTILINE)
```
terrykong: [`tests/unit/distributed/test_virtual_cluster.py:55`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/tests/unit/distributed/test_virtual_cluster.py#L55)

**1 action item.**

**TL;DR — `_ray_sub_default` cannot read `ray.sub`'s two sandbox port defaults, so anyone extending this guard to cover them hits a false "default not found in ray.sub" assert.**

PR-introduced. `_ray_sub_default` needs an unquoted `NAME=${NAME:-N}`, so it cannot read the two sandbox ports `ray.sub` actually declares — the same two this PR just added to the port map in [`virtual_cluster.py:97-99`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/nemo_rl/distributed/virtual_cluster.py#L97-L99):

- [`ray.sub:988`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/ray.sub#L988) — `SANDBOX_PORT="${NEMO_SKILLS_SANDBOX_PORT:-6000}"` (quoted, and the name inside differs from the name outside)
- [`ray.sub:990`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/ray.sub#L990) — `SANDBOX_BASE_PORT="${SANDBOX_BASE_PORT:-6001}"` (quoted)

I ran the current pattern against the real `ray.sub`: both return `None`, so `_ray_sub_default("SANDBOX_BASE_PORT")` fails its own `"default not found in ray.sub"` assert even though the default is right there on the line.

Nothing is wrong today. The router band is held inside (1024, 1313] by [`test_default_port_ranges_ordered_and_below_ephemeral_floor`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/tests/unit/distributed/test_virtual_cluster.py#L684-L688), so the sandbox band at 6000+ is out of reach and adding it to `reserved` would be a check that can never fire. This is only about the parser being narrower than the file it reads.

### AI-1

Let the name inside be any variable, and allow the quotes. I checked this against the real `ray.sub`: all 11 names the test looks up today resolve to the same values, and both sandbox ports start resolving.

```suggestion
    match = re.search(rf'^{name}="?\$\{{[A-Z0-9_]+:-(\d+)\}}"?', ray_sub, re.MULTILINE)
```

tests/unit/utils/test_config.py
```diff
@@ -32,6 +32,18 @@
...
+        str(path.relative_to(REPO_ROOT))
+        for pattern in ("*single-controller*.yaml", "*single_controller*.yaml")
+        for path in (REPO_ROOT / "examples/configs").rglob(pattern)
+    }
+    | {
+        str(path.relative_to(REPO_ROOT))
+        for family in (REPO_ROOT / "examples/nemo_gym").glob("nemotron-*")
+        for path in family.rglob("*.yaml")
+    }
+)
```
terrykong, line 42: [`tests/unit/utils/test_config.py:35-46`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/tests/unit/utils/test_config.py#L35-L46)

**1 action item.**

**TL;DR — the guard picks configs by filename, so 12 example configs that pin port ranges are never checked, including two with "nemotron" in the name.**

PR-introduced. `PORT_ISOLATION_CONFIG_PATHS` matches on filename (`*single[-_]controller*.yaml`) plus everything under `examples/nemo_gym/nemotron-*`. I ran that selection and a content-based one over the repo. Both pick 26 files — but they are different sets.

Unguarded today, all of them pinning a port range:

- [`examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-circle-click-2n8g-megatron-tp2ep8.v1.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/examples/configs/recipes/vlm/vlm_grpo-nemotron-omni-30ba3b-circle-click-2n8g-megatron-tp2ep8.v1.yaml#L141) and [`examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/examples/nemo_gym/grpo_workplace_assistant_nemotron_nano_v2_9b.yaml#L312) — both have "nemotron" in the name, and a test called `..._and_nemotron_..._are_isolated` does not collect either
- [`examples/nemo_gym/grpo_nanov3.yaml`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/examples/nemo_gym/grpo_nanov3.yaml#L208) — pins generation, gym **and** master
- `grpo_qwen3_30ba3b_thinking_swe1.yaml`, `swe2.yaml`, `distillation_qwen3_0_6b.yaml`, `examples/configs/distillation_math.yaml`, and the three `prefetch_*.yaml`

Meanwhile 12 of the 26 currently selected pin no port key at all, so they only compare the defaults against themselves.

All of these conform right now, so nothing is broken — this is about the next one. `swe_teacher.yaml` was exactly this class of config before this PR fixed it.

How it gets through: someone adds `examples/nemo_gym/grpo_newmodel.yaml` and copies the `port_range_low: 11001 / port_range_high: 15000` block this PR is removing. The filename does not match `*single_controller*` and it is not under `nemotron-*`, so the guard never sees it, CI stays green, and the collision turns up on a cluster.

### AI-1

Pick configs by what they contain, not what they are called. Nothing is lost: the 12 files this drops pin no port key, and each inherits from a parent the new selection does pick up. I also added the non-empty assert — an empty `parametrize` list reports `1 skipped` and exits 0, so a rename would drain this guard with CI still green.

```suggestion
# Selected by content, not filename: any example config that pins a service port
# range has to stay inside the layout in virtual_cluster.py. A config that pins
# none inherits the defaults and cannot fail the invariant.
PORT_ISOLATION_CONFIG_PATHS = sorted(
    str(path.relative_to(REPO_ROOT))
    for path in (REPO_ROOT / "examples").rglob("*.yaml")
    if "port_range" in path.read_text()
)
assert PORT_ISOLATION_CONFIG_PATHS, "port-isolation config selection matched nothing"
```

tests/unit/excluded_unit_tests.sh
```diff
@@ -34,8 +34,10 @@ EXCLUDED_UNIT_TESTS=(
     # DISTRIBUTED
     ###########################################################################
 
-    # test_virtual_cluster.py — Ray cluster infrastructure tests (~58s each)
-    --ignore=unit/distributed/test_virtual_cluster.py
+    # test_virtual_cluster.py — exclude only the two tests over 1s; the rest are
```
terrykong, line 40: [`tests/unit/excluded_unit_tests.sh:37`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/tests/unit/excluded_unit_tests.sh#L37)

**1 action item** — low, and pre-existing rather than introduced here. Raising it because this PR adds the first two `--deselect` entries directly under the comment that gets it wrong.

**TL;DR — the header tells you to write `--deselect` paths relative to `tests/`; written that way a `--deselect` matches nothing and is ignored silently, so the expensive test just runs.**

[`tests/unit/excluded_unit_tests.sh:5`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/tests/unit/excluded_unit_tests.sh#L5) says *"All paths are relative to tests/ (run_unit.sh cwd)."* That holds for `--ignore`, but not for `--deselect`, and following it makes an exclusion do nothing at all.

- `--ignore` takes a path, resolved from pytest's working directory, which [`tests/run_unit.sh`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/tests/run_unit.sh#L5) sets to `tests/`. So `--ignore=unit/...` is right.
- `--deselect` takes a node ID, matched as a plain string prefix against IDs that are relative to the pytest **rootdir** — the repo root, because [`pyproject.toml`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/pyproject.toml#L676) carries `[tool.pytest.ini_options]`. So it needs the `tests/` prefix.

Run from `tests/`, the two forms behave like this:

- `--deselect=tests/unit/test_d.py::test_a` → `1 passed, 1 deselected`
- `--deselect=unit/test_d.py::test_a` → `2 passed` — deselects nothing, exit 0, no warning

Your two new entries use the correct form, so the PR is fine. The risk is the header telling the next editor to use the other one, which fails silently.

### AI-1

Anchored here because line 5 sits outside this PR's diff, so a `suggestion` block cannot attach to it. Replacement for lines 4-5:

```bash
# Supports: --ignore=<path>, --ignore-glob=<pattern>, --deselect=<node_id>
# --ignore/--ignore-glob paths are relative to tests/ (run_unit.sh cwd).
# --deselect node IDs are relative to the pytest rootdir (the repo root), so
# they start with "tests/". A node ID without that prefix matches nothing and
# is ignored silently.
```

terrykong (commented):
Reviewed by a team of six agents. **The change is correct and worth landing** — no blocking issues, and the three comments below are all low-severity polish on the tests.

What we checked and found right:

- The old `6000-6099` router default really did sit on the NeMo-Skills sandbox — [`ray.sub:988-990`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/ray.sub#L988-L990) puts nginx on 6000 and uWSGI from 6001, and [`ray.sub:1003-1016`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/ray.sub#L1003-L1016) starts one sidecar per allocated node, driver included. Real collision, real fix.
- `[1202, 1300)` is free: nothing `ray.sub` pins (1200/1201, 1301-1312, 2000-2999, 8265) falls in it, and the k8s path pins no Ray ports there either.
- "Driver-local" is accurate — [`_maybe_start_generation_router`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/nemo_rl/algorithms/single_controller_utils/setup.py#L665) allocates on the driver and pins the actor with `NodeAffinitySchedulingStrategy(..., soft=False)`.
- The `--ignore` → `--deselect` swap uses the right prefix (rootdir is the repo root), and the asymmetric deselect is principled, not an oversight: [`_get_sorted_bundle_indices()`](https://github.com/NVIDIA-NeMo/RL/blob/b529d0f9b144dbb2c1c076da4bd3e1fad5b83575/nemo_rl/distributed/virtual_cluster.py#L876-L877) — the expensive topology probe — runs only under `use_unified_pg`, which is the twin you did deselect.
- Both new YAML inheritance comments resolve as written, and the three removed overrides were all above the 9000 floor.

One request, not about the code:

**Could you re-issue `/ok to test b529d0f9b144dbb2c1c076da4bd3e1fad5b83575`?** The existing `/ok to test` is pinned to `f20abcb4`, and `gh pr checks` currently shows no unit-test job at any state. `b529d0f9` is the commit that swaps a whole-file `--ignore` for two `--deselect`s, moving roughly 36 tests in `test_virtual_cluster.py` into the FAST lane — so the change most likely to affect CI is the one CI has not run.

_Generated by Claude Code_

tests/unit/distributed/test_virtual_cluster.py
yfw: Addressed in 1ffab974c. The parser now accepts quoted defaults and aliased inner variable names, with direct cases covering SANDBOX_PORT=6000 and SANDBOX_BASE_PORT=6001.

tests/unit/utils/test_config.py
yfw: Addressed in 1ffab974c. The selection now scans all example YAMLs by port_range content, asserts that the selection is non-empty, and the test name/docstring reflect the broader example-config scope.

tests/unit/excluded_unit_tests.sh
yfw: Addressed in 1ffab974c. The header now distinguishes tests-relative ignore paths from repo-root-relative deselect node IDs and calls out the silent no-match behavior.

terrykong (approved):

PR #3886, opened 2026-08-27
fix(grpo): apply effort shaping to async Gym rollouts

yfw:
## Problem

Legacy async GRPO calls `run_async_nemo_gym_rollout` without `effort_config`. Runs configured with `env.nemo_gym.effort_levels` therefore shape synchronous and validation rewards but silently skip the same shaping during async training.

## Change

- resolve the existing validated effort config in `AsyncTrajectoryCollector`
- pass it to every NeMo Gym stream attempt for GRPO
- keep the shared PPO collector behavior unchanged
- preserve the Super branch flexible length-penalty wiring at the same rollout boundary
- cover the collector-to-rollout wiring, including a partial-stream retry

This is the `super-v3.5-posttraining` counterpart to the main fix and addresses the gap identified in https://github.com/NVIDIA-NeMo/RL/pull/3770#discussion_r3852293228.

## Validation

- `pytest -q tests/unit/algorithms/test_async_utils.py -k nemo_gym_batch_retry_forwards_effort_config_without_duplicates`
- `ruff check nemo_rl/algorithms/async_utils/trajectory_collector.py tests/unit/algorithms/test_async_utils.py`
- `ruff format --check nemo_rl/algorithms/async_utils/trajectory_collector.py tests/unit/algorithms/test_async_utils.py`

### DIFF
--- a/nemo_rl/algorithms/async_utils/trajectory_collector.py
+++ b/nemo_rl/algorithms/async_utils/trajectory_collector.py
@@ -30,6 +30,7 @@ from transformers import PreTrainedTokenizerBase
 from nemo_rl.algorithms.grpo import (
     AsyncGRPOConfig,
     GRPOConfig,
+    _get_effort_config,
 )
 from nemo_rl.algorithms.grpo import (
     MasterConfig as GRPOMasterConfig,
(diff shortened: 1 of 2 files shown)


yfw:
will cp from main instead

PR #3890, opened 2026-08-27
ci: bump Gym to fa0d2515 for image-tools support

DanialTaheri:
## What

`super-v3.5-posttraining` pins Gym at `2251ef7f`, which predates the image tools: that commit carries **zero** files under `resources_servers/image_tools` or `responses_api_agents/image_tools_agent`, so the tooling is unavailable on this branch.

This bumps the Gym submodule to `fa0d2515`.

## Why a pointer bump and not a Gym PR

Both image-tools commits are **already merged into Gym `main`**:

- `4a58218` Add image tools agent and PivotRL verifier (NVIDIA-NeMo/Gym#2624)
- `0951c9b` fix(ci): make image tools agent discoverable (NVIDIA-NeMo/Gym#2643)

So nothing needs to be contributed to Gym — this branch is simply pointing at an older snapshot of it. `2251ef7f` is an ancestor of `fa0d2515`, so the bump is a **fast-forward with no divergence**.

## Why `fa0d2515` rather than Gym main's tip

`fa0d2515` is the exact Gym revision used for the Super Omni image-tools GRPO runs, so it is the one that has actually been exercised end to end.

## What it brings

19 files — 12 under `resources_servers/image_tools`, 7 under `responses_api_agents/image_tools_agent` — including the ten image tool executors and the tool-call parser in `base.py`, plus the intervening Gym progress between the two revisions.

## Note

This makes the tooling available. It does not by itself wire up a runnable multi-turn GRPO config: the `datasets:` block in `image_tools_agent.yaml` and the corresponding dataset files are not part of this change.


yfw:
Bumped in https://github.com/NVIDIA-NeMo/RL/pull/3958

PR #3903, opened 2026-08-28
fix(megatron): disable router replay in MTP layers

guyueh1:
## Summary
- Exclude routers under MCore MTP layers from router replay by default.
- Keep NRL_ROUTER_REPLAY_EXCLUDE_MTP=0 as an escape hatch for legacy/debug behavior.
- Document the default and add unit coverage for default exclusion, explicit inclusion, and MTP-only no-op behavior.

## Testing
- python3 py_compile syntax check for router_replay.py and test_router_replay.py
- uv run pytest tests/unit/models/megatron/test_router_replay.py -q -k mtp could not run locally because the lockfile supports Linux platforms only.


nemo_rl/models/megatron/router_replay.py
```diff
@@ -83,6 +84,37 @@ def _iter_model_modules(model: Any) -> Iterable[Any]:
...
+
+    beneath_mtp = beneath_mtp or bool(getattr(model, "is_mtp_layer", False))
+    yield model, beneath_mtp
+
+    children = getattr(model, "children", None)
+    if callable(children):
+        for child in children():
+            yield from _iter_model_modules_with_mtp_ancestry(
+                child, beneath_mtp=beneath_mtp
+            )
```
guyueh1, line 93: [`nemo_rl/models/megatron/router_replay.py:87-106`](https://github.com/NVIDIA-NeMo/RL/blob/80b0f78a23803fb31622d0d876ff4a450898e0bb/nemo_rl/models/megatron/router_replay.py#L87-L106)

**1 action item.**

### AI-1
Now that both call sites use this function, the old [`_iter_model_modules`](https://github.com/NVIDIA-NeMo/RL/blob/80b0f78a23803fb31622d0d876ff4a450898e0bb/nemo_rl/models/megatron/router_replay.py#L74-L84) (L74-84, pre-existing, unchanged by this PR) is dead code. A repo-wide grep for `_iter_model_modules\b` confirms no other references (definition + its own recursive call only). Per CLAUDE.md: "If you are certain that something is unused, you can delete it completely."

(Minor aside, not a separate ask: this replacement recurses via `.children()` without the id-memo that `nn.Module.modules()` used internally for dedup/cycle-safety. Both current call sites already dedup downstream so this is benign today — no action needed unless a third caller is added later.)

**Action:** delete `_iter_model_modules` at [L74-84](https://github.com/NVIDIA-NeMo/RL/blob/80b0f78a23803fb31622d0d876ff4a450898e0bb/nemo_rl/models/megatron/router_replay.py#L74-L84).

nemo_rl/models/megatron/router_replay.py
```diff
@@ -83,6 +84,37 @@ def _iter_model_modules(model: Any) -> Iterable[Any]:
...
+
+
+def _router_replay_exclude_mtp_enabled() -> bool:
+    return os.getenv(_ROUTER_REPLAY_EXCLUDE_MTP_ENV, "1").lower() in {
+        "1",
+        "true",
+        "yes",
+        "on",
+    }
```
guyueh1: [`nemo_rl/models/megatron/router_replay.py:109-115`](https://github.com/NVIDIA-NeMo/RL/blob/80b0f78a23803fb31622d0d876ff4a450898e0bb/nemo_rl/models/megatron/router_replay.py#L109-L115)

**1 action item.**

### AI-1
`_router_replay_exclude_mtp_enabled` fails **open**: only `{"1","true","yes","on"}` (after `.lower()`, no `.strip()`) enable exclusion. Any typo (`=ture`, `=enabled`, a trailing space from a launcher/YAML) silently falls through to `False` and re-enables the exact MTP/decoder `layer_number` collision this PR exists to fix — no error, no log.

The sibling `_router_replay_validation_enabled` (L229-235) has the same permissive parse, but its fail-direction is safe (unrecognized → validation off, the conservative default). This flag's fail-direction is unsafe (unrecognized → MTP routers re-included → silently wrong training data), which is the "silently produces wrong results" case the error-handling guideline flags for a setup-time raise.

**Action:** `.strip()` before `.lower()`, and raise on values outside the truthy/falsy sets instead of defaulting to False.

```suggestion
def _router_replay_exclude_mtp_enabled() -> bool:
    value = os.getenv(_ROUTER_REPLAY_EXCLUDE_MTP_ENV, "1").strip().lower()
    if value in {"1", "true", "yes", "on"}:
        return True
    if value in {"0", "false", "no", "off"}:
        return False
    raise ValueError(
        f"Invalid {_ROUTER_REPLAY_EXCLUDE_MTP_ENV}={value!r}; expected one of "
        "{'1','true','yes','on','0','false','no','off'}."
    )
```

guyueh1 (commented):
## Agent team review

5 agents (rl-expert, test-agent, bug-finder, design-reviewer, comment-reviewer) reviewed this PR, with devil's-advocate stress-testing every finding before posting.

**PR necessity: confirmed, real fix.** MCore gives MTP inner layers un-offset `layer_number`s starting at 1 ([transformer_layer.py](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/transformer_layer.py#L357) — verified against the pinned nested checkout at `3rdparty/Megatron-Bridge-workspace/Megatron-Bridge/3rdparty/Megatron-LM`), colliding with decoder layer 1 — pre-PR, MTP routers silently got decoder-layer routes replayed into them. The fix, doc update, and 3 new unit tests are correct.

2 actionable findings below. 5 additional candidate findings were disputed/downgraded by devil's advocate after verification and are not posted: tests not covering the self-flagged-router case (premise was factually wrong — the ancestor and self cases execute the identical line), test Dummy-class duplication (matches 8 pre-existing occurrences in this file), doc table placement (earlier prose already gives correct context), the new `.children()` traversal dropping `nn.Module.modules()`'s id-memo (no live bug, no named future caller), and `build_router_replay_tensors` being an unused-function trap (pre-existing, fails loud not silently).

Note: `mergeStateStatus=BLOCKED` only because a vetter hasn't run `/ok to test` yet on this fork PR — not a review blocker. Linter (`pre-commit --all-files`) could not run in this review environment (macOS host; repo lockfile is Linux-only) — not a reflection of PR quality.

_Generated by Claude Code_

guyueh1:
@zyzhou5 @yfw can you review this PR
Basically we hit an issue that when rollout doesn't use MTP but policy trains MTP, MTP layers do not have router replay info; and because Mcore names MTP layer_number 0-index the same as regular layers, so MTP layer 1 will silently use backbone layer 1's router info and mess up the MTP loss. 

This bug is a temporal fix that by default disables router replay for MTP layers; it can still be turned on R3 in MTP with NRL_ROUTER_REPLAY_EXCLUDE_MTP=1 but i think at the moment this path is broken (for both rollout using MTP and not using MTP cases).

zyzhou5 (commented):
This temporary fix looks good to me.

guyueh1:
@yfw any comments? @zyzhou5 can you approve the PR?

zyzhou5 (approved):

guyueh1:
@terrykong  can i ask for final review?

guyueh1:
@yfw  can you do final review?

nemo_rl/models/megatron/router_replay.py
```diff
@@ -70,17 +71,38 @@ def validate_router_replay_config(config: PolicyConfig) -> None:
...
+                item, beneath_mtp=beneath_mtp
+            )
         return
 
-    modules = getattr(model, "modules", None)
-    if callable(modules):
-        yield from modules()
-    else:
-        yield model
+    beneath_mtp = beneath_mtp or bool(getattr(model, "is_mtp_layer", False))
```
yfw: [`nemo_rl/models/megatron/router_replay.py:85`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/nemo_rl/models/megatron/router_replay.py#L85)

**1 action item** (low severity, latent today) — please fix in this PR together with the fixture change in `tests/unit/models/megatron/test_router_replay.py`.

**TL;DR — MCore's MTP wrapper owns the un-offset `layer_number` but never carries `is_mtp_layer`, so on an MTP-only pipeline stage the exclusion leaves MTP layer numbers in the "local MoE layers" set and `build_router_replay_assignments` raises instead of no-op'ing.**

PR-introduced (the new walker), but unreachable from any shipped NeMo-RL config today.

1. [`MultiTokenPredictionLayer.__init__`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L950) sets `self.layer_number = layer_number + get_mtp_layer_offset(...)` (offset 0 without a custom PP layout) and never sets `is_mtp_layer`; the flag first appears on its child [`mtp_model_layer`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L1047) (or the [`HybridStack`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L1030) on the hybrid path).
2. So the walker yields the wrapper with `beneath_mtp=False`, and [`_local_layer_numbers_for_model`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/nemo_rl/models/megatron/router_replay.py#L165-L175) still collects MTP depths `1..k`.
3. On a rank with no decoder replay instances, the [no-instances branch](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/nemo_rl/models/megatron/router_replay.py#L464-L475) intersects `{1..k}` with the global MoE layers and raises `Could not find any model-owned RouterReplay instances for local MoE layers [1]. Ensure Megatron was initialized with moe_enable_routing_replay=True.` — measured on a real MCore `GPTModel` (num_layers=4 all-MoE, mtp_num_layers=2, exclusion on): local layer numbers `[1, 2]`, raise.

Reachability: an MTP-only stage needs `pipeline_model_parallel_layout`, which `setup.py` does not forward to `model_cfg`, and default placement puts MTP on the last stage with decoder layers, so this cannot fire from NeMo-RL today. The PR's `test_router_replay_excluded_mtp_only_rank_noops` is green only because its `DummyMTPStack` puts the flag on the wrapper, unlike the real class.

**Action:** classify the wrapper by the child MCore assigns only in that constructor (MCore itself duck-types on it in [`cuda_graphs.py`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/cuda_graphs.py#L2989)). Verified on a copy: full test file 33 passed, `ruff format`/`ruff check` clean; escape-hatch behavior (`=0`) unchanged.

```suggestion
    beneath_mtp = (
        beneath_mtp
        or bool(getattr(model, "is_mtp_layer", False))
        # MCore stamps is_mtp_layer on the inner TransformerLayer/HybridStack and on
        # the Router, but not on the MultiTokenPredictionLayer wrapper that owns the
        # un-offset layer_number; classify the wrapper by the inner block it owns.
        or getattr(model, "mtp_model_layer", None) is not None
    )
```

**Context — no action.** The ancestry walk itself is load-bearing, not cosmetic: [`HybridStack`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/models/hybrid/hybrid_block.py#L103) flags itself but forwards `is_mtp_layer` only to its [attention/MLA layers](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/models/hybrid/hybrid_block.py#L143-L174), not to the [MOE/MLP branches](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/models/hybrid/hybrid_block.py#L175-L191). A Nemotron-3 `*E` MTP router therefore has `is_mtp_layer=False` on itself and is caught only through the flagged ancestor — exactly the shape the PR's `DummyMTPStack` fixture models. Worth a one-line note in the walker docstring so nobody later "simplifies" it to a per-router flag check.

tests/unit/models/megatron/test_router_replay.py
```diff
@@ -151,6 +151,217 @@ def __init__(self):
...
+            self.config = SimpleNamespace(num_layers=1, moe_layer_freq=[1])
+            self.mtp = DummyMTPStack()
+
+    try:
+        model = DummyModel()
+        routed_experts = torch.tensor([[[1, 2]], [[3, 4]]], dtype=torch.int32)
+
+        assert build_router_replay_assignments(model, routed_experts) == []
+        set_router_replay_forward(model, routed_experts)
+        assert model.mtp.nested[0].router_replay.router_replay_action is None
```
yfw: [`tests/unit/models/megatron/test_router_replay.py:340-360`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/tests/unit/models/megatron/test_router_replay.py#L340-L360)

**1 action item** — companion to the walker fix in `nemo_rl/models/megatron/router_replay.py` L85; **apply both together** (this test is red at head until the walker fix lands — that red is the gap it pins).

The fixture places `is_mtp_layer` on the wrapper that holds the router list, but MCore's [`MultiTokenPredictionLayer`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L950) owns `layer_number` and has no flag; the flag lives on its [`mtp_model_layer`](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/multi_token_prediction.py#L1047) child. With MCore's placement this test raises `Could not find any model-owned RouterReplay instances for local MoE layers [1]` on head.

**Action:** mirror MCore's attribute placement (one spanning edit so the fixture and the assertion move together; ruff-formatted). Measured: red on head, green with the L85 fix, the other 33 tests unchanged. The two byte-identical sibling fixtures at L219-L225 and L286-L292 stay green either way; the same shape there is optional but keeps the three consistent.

```suggestion
    class DummyMTPInner(torch.nn.Module):
        def __init__(self):
            super().__init__()
            self.is_mtp_layer = True
            self.layer_number = 1
            self.nested = torch.nn.ModuleList(
                [DummyRouter(RouterReplay(), layer_number=1)]
            )

    class DummyMTPStack(torch.nn.Module):
        # Mirrors MCore MultiTokenPredictionLayer: owns the un-offset layer_number
        # but carries no is_mtp_layer; the flag lives on the inner layer/router.
        def __init__(self):
            super().__init__()
            self.layer_number = 1
            self.mtp_model_layer = DummyMTPInner()

    class DummyModel(torch.nn.Module):
        def __init__(self):
            super().__init__()
            self.config = SimpleNamespace(num_layers=1, moe_layer_freq=[1])
            self.mtp = DummyMTPStack()

    try:
        model = DummyModel()
        routed_experts = torch.tensor([[[1, 2]], [[3, 4]]], dtype=torch.int32)

        assert build_router_replay_assignments(model, routed_experts) == []
        set_router_replay_forward(model, routed_experts)
        assert (
            model.mtp.mtp_model_layer.nested[0].router_replay.router_replay_action
            is None
        )
```

docs/guides/router-replay.md
```diff
@@ -28,6 +28,12 @@ return routed expert indices by setting `enable_return_routed_experts=True` in
 the vLLM kwargs. The generation payload is then carried through the normal
 rollout and policy data path as the `routed_experts` field.
 
+For models that also train MoE-based MTP heads, Router Replay skips MTP
+routers by default. This keeps MTP routers on their native routing decisions
+while replaying vLLM routes only in the decoder layers. Set
+`NRL_ROUTER_REPLAY_EXCLUDE_MTP=0` only when intentionally debugging the legacy
+behavior that replays MTP routers too.
```
yfw: [`docs/guides/router-replay.md:33-35`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/docs/guides/router-replay.md#L33-L35)

**1 action item.** PR-introduced wording.

**TL;DR — `=0` does not "replay MTP routers too"; it feeds decoder layer *k*'s routes into MTP layer *k* in every rollout mode, so the paragraph presents a known-wrong path as a debug mode.**

`=0` cannot replay MTP routes: [`_payload_indices_for_moe_layers`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/nemo_rl/models/megatron/router_replay.py#L195-L216) only admits payload layouts of `len(global_moe_layers)` or `num_layers`, so there is no MTP slot, and vLLM 0.25.1's routed-experts capture buffer is sized to `num_hidden_layers` and skips higher layer ids. With `=0`, an MTP router with un-offset [`layer_number` k](https://github.com/NVIDIA/Megatron-LM/blob/731b791469004f8fdcb896e65d610d6b1b0ebb32/megatron/core/transformer/transformer_layer.py#L357-L364) is handed [decoder layer k's routes](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/nemo_rl/models/megatron/router_replay.py#L479-L487) in every rollout mode (measured on a real MCore model: instances `[1, 2, 3, 4, 1, 2]`). "Debugging the legacy behavior that replays MTP routers too" reads as a supported mode. The paragraph also omits that other values raise.

**Action:** say what `=0` actually does (matching row edit suggested on L65).

```suggestion
while replaying vLLM routes only in the decoder layers.
`NRL_ROUTER_REPLAY_EXCLUDE_MTP=0` restores the previous behavior, which fed
decoder layer *k*'s routes into MTP layer *k*; the rollout payload carries no
MTP routes, so use it only to reproduce that behavior. Other values raise
`ValueError`.
```

docs/guides/router-replay.md
```diff
@@ -56,6 +62,7 @@ is intended for correctness debugging, not long training runs.
 
 | Environment variable | Default | Meaning |
 | --- | --- | --- |
+| `NRL_ROUTER_REPLAY_EXCLUDE_MTP` | `1` | Skip routers under MCore MTP layers. Set to `0` to include MTP routers in replay. |
```
yfw: [`docs/guides/router-replay.md:65`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/docs/guides/router-replay.md#L65)

**1 action item.** Same fact as the paragraph above: `0` does not "include MTP routers in replay", it feeds decoder layer *k*'s routes into MTP layer *k*; and other values raise.

```suggestion
| `NRL_ROUTER_REPLAY_EXCLUDE_MTP` | `1` | Skip routers under MCore MTP layers. `0` restores the previous behavior, which fed decoder layer *k*'s routes into MTP layer *k* (the rollout payload carries no MTP routes); use it only to reproduce that behavior. Other values raise `ValueError`. |
```

nemo_rl/models/megatron/router_replay.py
```diff
@@ -70,17 +71,38 @@ def validate_router_replay_config(config: PolicyConfig) -> None:
     _install_missing_route_fallback_patch()
```
yfw: [`nemo_rl/models/megatron/router_replay.py:71`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/nemo_rl/models/megatron/router_replay.py#L71)

**1 action item** (low severity). PR-introduced.

[`_router_replay_exclude_mtp_enabled`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/nemo_rl/models/megatron/router_replay.py#L96-L105) raises on a bad value (good — the sibling `_router_replay_validation_enabled` silently treats junk as off), but it is only reached from the two walkers, first via [`set_router_replay_forward` in `forward_step`](https://github.com/NVIDIA-NeMo/RL/blob/723335d8992f7d15cff8c734d7c55f0e0fbc9b29/nemo_rl/models/megatron/train.py#L274-L279) — after model load and the first rollout. This setup-time gate never parses it, so `NRL_ROUTER_REPLAY_EXCLUDE_MTP=ture` passes setup and dies on the first training microbatch.

**Action:** parse once here so it fails at setup. Verified on a copy: full test file still 33 passed (the validate tests run with the var unset), and a test calling `validate_router_replay_config` with `=ture` goes red at head and green with this line.

```suggestion
    _router_replay_exclude_mtp_enabled()
    _install_missing_route_fallback_patch()
```

yfw (approved):

PR #3905, opened 2026-08-29
fix(ci): set TMS_CUDA_MAJOR via uv extra-build-variables

kajalj22:
## Summary
* torch-memory-saver's build backend needs `TMS_CUDA_MAJOR` when uv builds it during `uv lock`. Declare it once via `[tool.uv.extra-build-variables]` in `pyproject.toml` so every caller of `uv lock`/`uv sync` (GitHub Actions, nemo-ci, local dev) picks it up automatically — no per-CI-system shell/sed logic needed.
* Reverts the GitHub Actions-only workaround from #3862 (sed-derive + export before `uv lock`), which this supersedes.

## Test plan
- [ ] `uv lock` succeeds and lockfile includes torch-memory-saver without manually exporting `TMS_CUDA_MAJOR`
- [ ] CI "Generate lockfile" step passes

### DIFF
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -477,10 +477,7 @@ jobs:
           enable-cache: true
 
       - name: Generate lockfile
-        run: |
-          export TMS_CUDA_MAJOR="$(sed -nE 's/^ARG[[:space:]]+BASE_IMAGE=.*cuda([0-9]+)\..*/\1/p' docker/Dockerfile | head -1)"
-          test -n "$TMS_CUDA_MAJOR"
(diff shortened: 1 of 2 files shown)


terrykong (commented):
does the env need to be re locked?

chtruong814 (approved):

terrykong (approved):
approved. i locked locally and i guess the env doesn't need to be re-locked. looks like env vars don't get saved in lock so technically aren't part of what it determines as equivalency

kajalj22:
> does the env need to be re locked?

UV.lock is valid - checked

PR #3907, opened 2026-08-29
feat(refit): support ordered NCCL Reshard components

seonjinn:
## Summary

NCCL Reshard currently treats each logical parameter as one tensor. Native quantized weights need more than one tensor, such as FP8 values plus scales.

This PR adds an ordered component contract:

```text
logical parameter
  - weight
  - optional weight_scale
```

Legacy BF16 and single-tensor FP8 parameters are normalized to one `weight` component, so their transfer path stays unchanged. The generation side verifies that the canonical component plan survives wire serialization before either NCCL communicator enters warm-up.

## What changes

- Add typed metadata for ordered refit components.
- Carry component shape, dtype, placements, mesh geometry, and misc metadata through wire serialization.
- Reject missing, duplicated, reordered, or shape-incompatible components during setup.
- Validate the complete plan before NCCL communicator initialization.

This layer does not import Transformer Engine or vLLM. Producer and consumer support are in #3908 and #3909.

## Tests

- Added unit coverage for legacy one-component plans, FP8 value/scale plans, grouped experts, placement validation, digest mismatch, and communicator setup order.
- 161 focused GB200 tests for component metadata, NCCL Reshard metadata, and weight synchronizers pass.
- Ruff lint, Ruff format, and `git diff --check` pass locally.


seonjinn (commented):
## Review: transport contract for native MXFP8 refit (stack 1/3)

Reviewed as the base of the three-PR stack (#3907 contract -> #3908 Megatron producer -> #3909 vLLM consumer). Scope was taken as the `347fcbcc3..HEAD` range (7 files, 827 insertions), not the cumulative file list that `gh pr view --files` reports.

Before the findings, a note on what was and was not verifiable: this review was performed without a GPU and without a working `uv` environment (the repo lockfile is Linux-only), so **no repo test was executed**. Every claim below is either a direct source read, an inspection of the pinned upstream ref, or a small standalone Python execution, and each is labelled. Claims that could not be settled are posted as questions rather than assertions.

---

### CRITICAL - the new synchronizer test cannot pass as written

The new test asserts on an event ordering by calling `.index("generation.init_collective")` on the recorded event list, but the code path under test (`_build`) never invokes any generation-side method, so that string is never recorded and `list.index` raises `ValueError`. This is a hard failure, not a soft assertion miss - the test does not silently pass, it errors.

Please re-run this test in isolation before merge. If it is passing in your environment, that itself is worth understanding, because it would mean the recorder is being populated from somewhere the review could not see.

### The plan digest cannot distinguish two fleet shapes it is specifically meant to distinguish

`component_plan_digest` hashes role, dtype, global_shape, src/dst placements, pp_stage, and grouped_expert_proj. It does **not** hash mesh extents. And mesh extents are not recoverable from the placements, because `build_mesh_info`'s own docstring (`nccl_reshard_utils.py:377-379`) states that size-1 dims are dropped - the extents live only in `MeshInfo.mesh`.

Consequence, constructed and executed: a 24-worker fleet at tp=8 (mesh `[3, 8]`) and a 16-worker fleet at tp=8 (mesh `[2, 8]`) produce **byte-identical digests**. That is verbatim the mismatch class the guard at `nccl_reshard_weight_synchronizer.py:362` warns about.

In fairness, coverage here is incidental rather than absent: a 16-to-8 worker drop at tp=8 removes `dp` from `dim_map` entirely, which changes the placement list length and therefore does change the digest. The collision requires the new dp degree to stay above the `>1` threshold. So this is a gap in a guard that mostly works, not a guard that never works. Suggested fix is to fold `MeshInfo.mesh` into the hashed tuple.

### The digest round-trip assertion is close to a tautology

`restore_refit_info_placements` has exactly one production call site: `vllm_backend.py:1124`, on the generation side. Train workers never compute or check a digest. So the assertion being made is effectively `digest(restore(wire_safe(P))) == digest(P)` - and the test at `tests/unit/weight_sync/test_weight_synchronizer.py:762` compares the shipped object against itself rather than against an independently constructed expectation.

To be fair to the design: combined with the `raise` in `_canonical_placements` (`refit_components.py:219`), the round-trip check does real work - it converts `_restore_placement`'s catch-all `return Replicate()` (`nccl_reshard_utils.py:272`) from a silent wrong-slice into a loud failure. That is worth keeping. The question is scoping: is a producer-side digest (or a test that builds the expected plan independently) in scope for this PR, or deliberately deferred?

Related scoping question: `misc_meta` rides the same wire (`megatron_policy_worker.py:2957`) with **zero** digest coverage, while producer/consumer ordering agreement is explicitly relied upon (comment at `:2871`). Is the digest intended to cover the bulk path only, by design, or is misc an omission?

### `K % 32` is validated globally but sharding is per-role and local

The MXFP8 block constraint is checked against the **global** K, but each role is sharded independently, so what actually has to be divisible by 32 is the **local** K on each rank. Three separate lines of analysis converged on this same point from three different vantage points - the plan builder, the consumer's `_expected_local_component_shapes`, and the padded-scale slicing in #3908's `mxfp8_refit_source.py`. Any K where `K/32` is not a multiple of the shard count lands here.

Suggested: validate divisibility against the post-shard local K, at plan-build time, with the rank and role in the message.

### Error contract: documented `ValueError`, actual bare `KeyError`

Executed. The docstring declares `Raises: ValueError`, but `metadata["shape"]`, `metadata["dtype"]`, and `item["role"]/["shape"]/["dtype"]` escape as bare `KeyError` carrying no parameter name - across a Ray plus `collective_rpc` boundary, where an unnamed `KeyError` is close to undiagnosable. Every *other* error in the module names the offending param. Cheap fix, real diagnosability win.

A forward-looking companion: the dtype fallback uses the **logical** `param_info["dtype"]`. That is benign today, but for an MXFP8 param the logical dtype differs from the weight component's dtype, so the equivalence this relies on dies in #3908.

### Structural: the component-plan logic exists twice

`refit_adapter.py:745-816` (introduced in #3909) is a near-verbatim reimplementation of `refit_components.py:84-115`. This is the most significant structural finding in the stack, and it is not an isolated instance - the same "two independent implementations of the same native-detection concept" pattern shows up again in #3909, where `_native_component_names` and `_native_mxfp8_param_names` classify the same plan differently. Three reviewers reached that second instance independently.

Since #3907 owns the contract, this is the right PR to decide where the canonical implementation lives.

### Performance / memory - one open question for the stack

The native path halves the wire bytes, which is the headline win. But it also **adds** a per-layer materialize-and-copy-back cycle that the receiver-quantize path does not pay. Peak device memory during refit is therefore a genuine open question, not a formality. Could you share a peak-memory measurement for the native path against the receiver-quantize baseline on the same model?

### Lint

`tests/unit/weight_sync/test_weight_synchronizer.py:17` fails `ruff check --select I` (I001, import ordering). Note this is invisible to a plain `ruff check`, because `[tool.ruff.lint] select` at `pyproject.toml:755` is `["D", "F"]` and excludes `I` - the `I` rule only runs via `.pre-commit-config.yaml:16`. Worth knowing, since two of the three PRs in this stack have the same violation.

---

Overall the contract design reads well and the digest guard is a real improvement over trusting ordering. The blocking item is the test that raises `ValueError`; the digest mesh-extent gap is the one substantive design issue.

seonjinn:
/ok to test [a5edfd0](https://github.com/NVIDIA-NeMo/RL/pull/3907/commits/a5edfd0e32617abce7c5cb3fa46d2cac3a71c213)

seonjinn:
Follow-up self-review completed on `12738f54`.

- The plan digest now includes component placements, source/destination meshes, and misc metadata.
- Malformed placement metadata fails with a clear `ValueError` instead of reaching digest generation with an invalid type.
- Local MXFP8 K-shard divisibility and missing metadata paths fail before communicator setup.
- Generation collective tests now follow the rebuilt communicator object.
- The previous CI `pyrefly` finding for placement metadata is fixed.

Verification:
- Focused Linux component/transport/synchronizer tests passed.
- `ruff check`, `ruff format --check`, and `git diff --check` passed.
- No unresolved review threads remain.

seonjinn:
ok to test

seonjinn:
/ok to test [34fc34f](https://github.com/NVIDIA-NeMo/RL/pull/3907/commits/34fc34f380e542dd4987259100e3ba420af5ceb8)

seonjinn:
## Self-review pass 2 (`/review-pr-team`) — delta since the first self-review

Re-ran the skill against `c5bc3d66..34fc34f3`, the five commits pushed after review 5057345747. Scope was the delta only; the first review still stands for everything below it.

Same environment caveat as last time: **no repo test was executed.** The lockfile resolves only for `linux/x86_64` and `linux/aarch64`, so `uv run` cannot resolve on this machine and there is no GPU here. Every closure below is verified by reading source and matching assertions to the code path they claim to exercise, never by observing a test pass. Per the new self-review policy I will re-run this on a GPU box before asking for human review if that is the expectation.

### Closed

| Finding from review 1 | Evidence |
|---|---|
| Synchronizer test raised `ValueError` from `events.index("generation.init_collective")` | `tests/unit/weight_sync/test_weight_synchronizer.py:677-684` now mocks `generation.rebuild_collective` / `generation.rebuild_nccl_reshard_comm_group`, which is what `nccl_reshard_weight_synchronizer.py:294` and `:331` actually call. The mock matches the real call rather than the assertion being relaxed. |
| Digest collided across mesh extents (24@tp8 `[3,8]` vs 16@tp8 `[2,8]`) | `refit_components.py:170-171` adds `src_mesh` / `dst_mesh` via `_canonical_mesh`, which `.tolist()`s the rank tensor. `_canonical_mesh` handles both the `MeshInfo` object and the post-`make_nccl_reshard_refit_info_wire_safe` dict form, which matters because `component_plan_digest` runs on both sides of that conversion. |
| `misc_meta` outside digest coverage | The payload is now `{"bulk": ..., "misc": ...}` with `misc_meta` name/shape/dtype folded in. This answers the question by inclusion. |
| `K % 32` validated globally while sharding is local | `_validate_native_local_block_width` (`nccl_reshard_utils.py:1083-1107`) derives `shard_count` from mesh extents on the last dim and checks `local_k % 32`. Called for **both** source and destination at `:1027-1041`, so it is not a single-sided fix, and the global check is retained alongside it. |
| Bare `KeyError` escaped where the docstring documents `ValueError` | `_required()` replaces all five bare subscripts in `normalize_refit_components`. `_positive_shape(None, ...)` also raises `ValueError` rather than `TypeError` when a key is present but null. |
| Duplication between the adapter and `refit_components` | Consolidated into `native_mxfp8_param_names`. See the note below about where its callers live. |
| ruff `I001` | `uvx ruff@0.9.9 check --select I`, `check`, and `format --check` all pass against the repo's own `pyproject.toml` config for every `.py` file this PR touches. |

### Still open

**The near-tautological round-trip assertion is unaddressed in this delta.** `test_wire_safe_then_restore_reproduces_placements_and_meshes`, `test_wire_safe_then_restore_reproduces_component_placements`, and `test_restore_refit_info_rejects_plan_digest_mismatch` (`tests/unit/weight_sync/test_nccl_reshard_utils.py:763-826`) are byte-identical to their pre-delta state. The objection stands: re-deriving the expected value with the same code under test means the assertion cannot fail for the class of bug it looks like it is guarding. A literal expected structure, or a fixture captured from a known-good run, would make it load-bearing.

### Note for reviewers, not a defect

`native_mxfp8_param_names` has **zero production callers within #3907**. Its consumers are `refit_adapter.py:33,802` and `vllm_backend.py:44,1112,1166,1248,1632`, all of which land in #3909. That is a defensible shape for a transport-contract PR at the base of a stack, but it does mean this function ships exercised only by unit tests until #3909 merges. Worth stating in the PR description so a reviewer does not read it as dead code.

One cosmetic redundancy: the trailing `if weight.role == "weight" and scale.role == "weight_scale":` in `native_mxfp8_param_names` cannot be false, because `normalize_refit_components` already constrains the role sequence before returning two components. Harmless.

